PR #134, opened 2011-08-07
DDS plugin updates

inequation:
This is already good enough to be pulled into the trunk I think. Commits constitute a functional DDS writer for plain and volume textures and a handful of reader fixes.


lgritz:
This has languished long enough. Closing. OIIO may be ubiquitous in film & animation pipelines, but I guess not so much for games or whoever it is that uses DDS files. I don't think anybody has complained about the current level of dds support in many years. We can always try to revive this if it becomes important.

PR #272, opened 2012-04-02
XPM reader support

sagittarius89:
Request contains 4 commits:
a) 3 commits from last GSoC (2011) with my XPM input plugin (It was not accepted because i didnt write test)
b) 4th commit for current GSoC. It contains some improvements for xpm input plugin (cosmetic changes, and also added code to support reading 3byte and 12byte RGB color from file). I attached also test for my plugin.


lgritz:
I don't honestly know why this was not committed. Possibly my fault letting it slip through the cracks while (mistakenly?) believing we were waiting for more changes? I don't know. But I think the effort it would take to rebase to the modern code is probably not worth it, considering that in the years since it was written, I don't recall a single person asking for xpm support. But if a time traveller from 1992 shows up and wants to read xpm files with OIIO, we could revive this from the archives, I guess.

PR #379, opened 2012-06-18
maketx.cpp modified to filter in linear colorspace

sjparker:
I've added a new flag --colorspace, which should be used when no color conversion is taking place. This makes it a bit more obvious that we're not changing the colorspace but merely defining the input colorspace for conversion to linear. Otherwise, we'd have to issue '--colorconvert sRGB sRGB' to get the same effect and that's a little too implicit to get what we're after.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,13 @@ branches/
 build/
 dist/
 testsuite/runtest.pyc
+
+Makefile.mine
+
+test.tif
+
+test.tx
+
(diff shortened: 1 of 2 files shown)


lgritz:
I'm going to close this very old PR. It seems to me that everything described here can be done, if not with maketx directly, than certainly with a modern oiiotool that can even output textures directly with `-otex`.

PR #402, opened 2012-07-13
Review: sockets.imageio and rendering to iv

chadrik:
The socket branch is good enough for a review.  There are still lots of debug statements and little issues to fix, but I thought it would be good to get some feedback on the overall design before I go any further.

Broad overview:

A SocketServer class listens for incoming connections on a specified port (the port is not yet configurable in iv UI).  When a new connection is made, the server calls a callback with the incoming file name (the file name should end in ".socket").  In the case of iv, this callback is used to create a new SocketInput by way of the ImageCache.  

The ImageCache does an initial iteration over the "tiles" of the SocketInput which is wasteful because the tiles have not arrived yet.  These queries are simply ignored, but it might be nice if we could somehow convey to the ImageCache that certain image plugins should skip the initial iteration.  

When a tile arrives, the SocketInput class calls a tile_changed_callback, which triggers the ImageCache to call get_native_tile with a data pointer for the SocketInput class to fill.

I still need to implement multi-layer images.  On the driver side I will need to convert from the tile-per-layer format that arnold provides to OIIO's layering at the pixel level.  are there any utility functions for this conversion?

I've also glazed over 3-dimensional images for now in the code that handles converting "cropped tiles" to full, native tiles.


lgritz:
This seems to have gotten dropped on the floor long ago, no reply to the last part of the conversation.
I'd still like some kind of socket protocol, but the code has diverged so much over the years, even if this was the right design, it would probably be better to start over than to try to rebase. Closing with apologies.

chadrik:
Hi Larry, sorry I’ve not had time to revive this. It was a side project in
my spare time which has since been displaced by other side projects (like 2
kids!).
Hopefully it’ll be useful as reference to someone in the future.

PR #419, opened 2012-08-06
Contrast

StefanStavrev:
There are 4 public functions, to allow
the user maximum control over how he
applies contrast. He can specify one pivot
and one contrast value for all channels,
one pivot for all channels and contrast
values per channel, etc.

Each of these 4 functions only prepare
the arguments, nothing else, and then
they call another function that does
input validation and that function
then calls contrast_impl.

It only makes sense to apply contrast
to the luminance channel if only
one pivot and one contrast value is
specified, that is why one of the public
functions has an additional parameter luminance.


lgritz:
Closing as obsolete. ImageBufAlgo has been completely overhauled since this was submitted. Not a reflection on Stefan... we just ended up going in a quite different direction for the technical design.

PR #545, opened 2013-03-24
Proposed merge of contrast function from Stefan Stavrev

lgritz:
All my fault, I've let some of Stefan's work from last summer languish. (So sorry, Stefan.)

Here is his work (from #419) that introduced an ImageBufAlgo::contrast() function to adjust image contrast, exposed as oiiotool --contrast.

I've added documentation and testsuite cases, but additionally made the following changes to Stefan's code:

* Pare down the number of contrast() variants to only two: one that takes the full per-channel contrast and pivot arrays, and one that takes scalars and adds contrast to the luminance (we call this contrast_lum).
* Use more modern idioms for type dispatch and parallelism in IBA::contrast.
* Don't clamp to 1 -- allow high dynamic range input and output. Add a parameter to the contrast functions specifying whether or not they should clamp to 0 as a lower bound.
* Changed luminance to assume Rec709 primaries and linear response; the previous one is a commonly-used formula that is also commonly misunderstood as it has 2.2 gamma correction built into it.
* In oiiotool, simplify the error checking by assuming that if not enough contrast or pivot values are supplied, they will default to 1 and 0.5, respectively; and if too many, the extras will be ignored.  Also simplify the handling of alpha channels by assuming that all channels have pivots and contrast scales, and that non-color channels like alpha will be given a scale of 1.0, or users gets what they deserve.

I want to emphasize that these changes are not meant to imply any shortcomings in Stefan's work; for the most part, they merely reflect a more modern style, more consistent with other image processing functions we've added in recent months, and evolution in my thinking about the general aesthetics of such functions in our codebase.

I've rearranged a bit, but really changed very little overall. My total work on this is about 2 hours, Stefan did all the heavy lifting, I just updated with modern OIIO ImageBufAlgo idioms.

(Fixes #419)


lgritz:
Withdrawing. This old PR is not indicative of the eventual direction that we took with the organization and implementation of ImageBufAlgo.

PR #630, opened 2013-06-28
Avoid float conversion in ImageOutput::to_native_rectangle

danwexler:


lgritz:
Dan has been busy since 2013, but his idea was good, so I reimplemented as #1836 (and in fact, it can be achieved with a smaller patch now). Closing this PR.

PR #1015, opened 2014-12-24
"open with configuration" extended to ImageBuf, ImageCache

lgritz:
This has never been a high priority, but it comes up as a request from time to time, so I thought I'd try to squeeze it in before we lock down the API for 1.5 release.

History: ImageInput has a variety of the open() method that takes an ImageSpec whose metadata can communicate special options or hints that a reader would need to know when the file is first opened. It's a rare case, but sometimes is used to solve tricky problems, allowing an app to talk to a particular format reader in custom ways that are not easily accommodated by the main ImageInput API. Occasionally people have noticed that this does not allow the same options to be passed to the readers when using the higher-level ImageBuf or ImageCache APIs, which don't directly expose the underlying ImageInput.

This patch tries to address this limitation by providing ways to pass the optional config ptr to ImageBuf ctr and reset(), and to ImageCache::add_file. It's not super elegant, but it fits, and it's rarely used in the vast majority of cases.

This preserves "source compatibility" but breaks link compatibility on the ImageBuf and ImageCache APIs, which I why I wanted to do it just prior to branching for a release, not within a locked release branch.

### DIFF
--- a/src/doc/imagebuf.tex
+++ b/src/doc/imagebuf.tex
@@ -163,7 +163,8 @@ Constructing a readable \ImageBuf that will hold an image to be read
 from disk.
 
 \apiitem{ImageBuf (string_view name, int subimage=0, int miplevel=0, \\
-\bigspc\bigspc ImageCache *imagecache=NULL)}
+\bigspc\bigspc ImageCache *imagecache = NULL, \\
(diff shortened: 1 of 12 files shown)


lgritz:
No objections -> merging

PR #1021, opened 2015-01-03
ImageBuf: Improve interppixel_NDC and add bicubic interpolation

lgritz:
1. For a couple major releases now, we've considered interppixel_NDC to be deprecated, in the sense that it considers the NDC coordinates to be relative to the data window, which we realized too late was stupid and provided the interppixel_NDC_full to do the "right" thing by considering NDC space to be relative to the "display" window (or "full" image). With this patch, we change the behavior of interppixel_NDC to be the way it should always have been, i.e., identical to interppixel_NDC_full. Now we consider interppixel_NDC_full to be deprecated, and will eventually remove it.

2. Added interppixel_bicubic and interppixel_bicubic_NDC, which do the same things as interppixel and interppixel_NDC, but using b-spline bicubic interpolation, rather than bilinear interpolation. The bicubic is smoother and may be better suited for some purposes.


sobotka:
On the subject of cubic interpolation, perhaps Mr. Gritz, you would consider adding the prefilter pass to the interpolation.

For those that aren't aware, a cubic interpolation will not pass through the exact points of the curve for a given value, leaving values slightly under. In imaging, this reduces overall intensities towards the median, decreasing perceptual sharpness generally.

The prefilter pass considers the image and augments the values such that values are perfectly achieved when performing the cubic interpolation.

Remarkable byproducts are:
  1. Maintains perceptual sharpness on a scale.
  2. Does not deviate to a sharper (such as SinC or Lanczos) or blurrier (Cubic) result over a stair / multiple interpolations.
  3. Results in a much more organic scale.

To the best of my knowledge, no imaging applications have harnessed the prefilter variation.

Turn on HD when viewing the [sample here](http://youtu.be/nfhTET86kdE).

[Further information here](http://wiki.blender.org/index.php/User:Sobotka/Cubic_B-Spline_Scaling_Algorithm).

lgritz:
That looks really nice. I can see this being helpful in a number of situations.
I didn't find the code super clear in that example link. What is the prefilter? Do you have a reference explaining what's going on here in more detail?

lgritz:
Never mind, I've found a bunch of references. Looks like a great idea, though I think it's all implemented as a pre-pass before doing this kind of single-point bicubic lookup. I'll experiment with an implementation and see how it compares to the other interpolation methods we use. In particular, while it's clearly an improvement over straight-up B-spline cubic interpolation, I'm interested if it presents any improvement over Lanczos or the other "better" filters we already support, or if it's roughly equivalent.

sobotka:
It destroys Lanczos and SinC because those err on the sharpening side with their weights.

As said, you can do a 10 step stair interpolation and begin to see nasty halos from the over-sharpening with either of the above window functions.

With the prefilter, there is none.

lgritz:
Nice, I will definitely try it out. The catch, I suppose, is that you always need to pre-filter the whole image, yes? There's not an especially good way to compute it locally.

Seems orthogonal to this review, I'll open a new PR when I've got something working.

sobotka:
Yes to whole image, but as an arty fatty fluff, I will say beyond a shadow of a doubt the prefilter scale destroys all other forms.

I was considering trying to add this myself, and decided to make a note here when your commit popped up. Rather serendipitous.

lgritz:
Nah, I'll take a whack at it today. Sounds like it's not too hard, and I have some specific ideas about where to fit it in.

PR #1028, opened 2015-01-17
Compiler fixes for VS2010 & 2013

darkhorse64:
The first commit has been checked with Visual Studio 2010 and Visual Studio 2013 (Win32 build only).

The second commit is more involved. There are a lot of changes but they are all of the same kind: every SIMD input to a function is changed to a const reference.
The annoying thing is that this change affect all platforms and that I am unable to say whether it has an adverse effect on performances or not. On the positive side, Embree does the same thing and nobody can really complain of its speed !

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -219,7 +219,7 @@ inline T madd (const T& a, const T& b, const T& c) {
 /// This is a template, and so should work for any types.
 template <class T, class Q>
 inline T
-lerp (T v0, T v1, Q x)
+lerp (const T& v0, const T& v1, const Q& x)
 {
     // NOTE: a*(1-x) + b*x is much more numerically stable than a+x*(b-a)
     return v0*(Q(1)-x) + v1*x;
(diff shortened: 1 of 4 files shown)


lgritz:
I think that since the functions end up fully inlined, it should generate the same code for 'float4' versus 'const float4 &'. But I don't understand why the compiler would complain... do you know what's wrong with passing these by value? What was the error message that it gave you?

I'm happy to integrate this, but I want to run some benchmarks first and be super sure that there is no performance issue. I'm 90% confident, but it would be unwise not to test it.

darkhorse64:
The error message was:
error C2719: '(name of an argument)': formal parameter with __declspec(align('16')) won't be aligned. This is how the __m128 type is qualified.

Googling a bit, I have found that "the align __declspec modifier is not permitted on function parameters" (see http://msdn.microsoft.com/en-us/library/373ak2y1%28v=vs.120%29.aspx) and that the workaround is to use const reference. The problem occurs only on the Win32 platform.

lgritz:
Interesting. OK, seems to generate the same code in the end anyway, so const ref it is!
I'll merge this into master and RB-1.5.

PR #1031, opened 2015-01-20
SIMD improvements: partial load and store

lgritz:
Add partial-load SIMD functions, and make the partial-load and partial-store use SSE ops for 1, 2, and 3-wise loads/stores (previously reverted to scalar ops for those cases).

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -625,13 +625,43 @@ public:
     /// Load from an array of 4 values
     OIIO_FORCEINLINE void load (const int *values) {
 #if defined(OIIO_SIMD_SSE)
-        m_vec = _mm_loadu_si128 ((simd_t *)values);
+        m_vec = _mm_loadu_si128 ((const simd_t *)values);
 #else
         m_val[0] = values[0];
         m_val[1] = values[1];
         m_val[2] = values[2];
         m_val[3] = values[3];
 #endif
+    }
+
+    OIIO_FORCEINLINE void load (const int *values, int n) {
+#if defined(OIIO_SIMD_SSE)
+        switch (n) {
+        case 1:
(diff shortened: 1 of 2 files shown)


src/include/OpenImageIO/simd.h
```diff
@@ -634,6 +634,34 @@ class int4 {
 #endif
     }
 
+    OIIO_FORCEINLINE void load (const int *values, int n) {
+#if defined(OIIO_SIMD_SSE)
+        switch (n) {
+        case 1:
+            m_vec = _mm_castps_si128 (_mm_load_ss ((const float *)values));
```
fpsunflower: Are you sure you don't get any type aliasing warnings here? (on gcc -- I don't think clang is as strict about this)

lgritz: Here's my favorite web site: https://software.intel.com/sites/landingpage/IntrinsicsGuide/

Type "_mm_load_" into the search bar, and you'll see that, strangely, _mm_load_ps takes a float*, whereas the analogous _mm_load_si128 takes a __mi128i* rather than an int*.

lgritz: Mis-typed: of course _mm_load_ps takes a float*, not a float.

lgritz: Ugh, and of course I meant _mm_load_ss. Coffee!

lgritz: Oh, and that's not what you meant at all, is it? You meant the cast of int* to float*. 
I'll double check on gcc. Sorry for all the spam.

lgritz: I've verified on multiple gcc versions that it's ok (though I did find an entirely different warning that I didn't get on clang, and which I have fixed and updated this PR).

PR #1033, opened 2015-01-21
SPI only: make COMPILER=clang or =gcc use the default clang or gcc in execution path

lgritz:
This won't affect anybody outside of SPI.


fpsunflower:
LGTM

lgritz:
OK, and I'll make precisely the same change in OSL land, without a separate review.

PR #1035, opened 2015-01-21
Fix broken bicubic texture sampling with non-power-of-two sized tiles

brechtvl:
It was using the wrong variable.


lgritz:
Ick!  Good catch, LGTM, I will add immediately.

Out of curiosity, where are you seeing non-pow2 tiles?

brechtvl:
The non-power-of-two tiles are coming from autoscanline, where each tile has the width of the image.

lgritz:
aha! I will merge shortly to both branches. (Do you also want 1.4?)

brechtvl:
This bug was introduced with the SIMD optimizations, so it doesn't exist in 1.4.

lgritz:
Oops. Doubly sorry, then.

lgritz:
Merged.

PR #1040, opened 2015-01-22
Bug fixes to deep tiled exr input/output

brianhall77:
Input: Tiles were always being read starting at 0,0 rather than the requested offset.  I matched variable names with similar existing logic in the deep tile output.

Output: Within spec_to_header, the call to put_parameter does a sanity check to make sure the image is tiled if setting lineOrder to randomY.  The problem in the case of deep exrs (any multipart exr, really) is that the sanity check was being done with m_spec before it was initialized.  This change instead gets the tile info from the header that is currently being built, with a comment to ensure that the tile description is set *before* the call to put_parameter.


brianhall77:
Please also include these commits in the 1.5 branch.

lgritz:
LGTM. I will merge into both branches.

brianhall77:
Thanks much!

lgritz:
Merged.

PR #1041, opened 2015-01-22
Sync get_texture_info() with the documentation

darkhorse64:
The documentation says: 

Supported dataname values include:

exists Return 1 if the file exists and is an image format that OpenImageIO knows how to read, otherwise return 0. The data pointer is not used.

Actually, it always returns true and write the result in data. 

This pull request syncs the code with the doc. Alternatively, the doc can be changed not to break compatibility


lgritz:
Hmmm... I need to think about which one should be "correct."

darkhorse64:
Cough ... "improved" fix

lgritz:
Ha, yeah, I was wondering about that line.

I'm torn about this. For all other queries, the return value of true means "I recognized the query and found the information you requested", and the result of the query is place in *data. I can't remember now why I did this differently, but it does seem strange that for this one query, it returns false even though it's perfectly capable of answering your question. But on the other hand, for all other queries, a nonexistent file causes a false return value.

I have gone back and forth on this a few times today, sometimes toward keeping the existing code and just changing the documentation. I want to sleep on it and maybe ask a couple people tomorrow.

fpsunflower:
It seems that the behavior for other `gettextureinfo` calls returning false is to issue an error. If we did want `exists` to return `false`, I think it should fill in the error string with what went wrong.

I guess the question to think about is -- when the user queries `"exists"` for a file, do they expect to get an error message about the reason why the file may not exist? It could be a missing file, improper permissions, unknown format, etc ...

It seems like a useful thing to provide. On the other hand, it makes the `exists==false` case more expensive.

lgritz:
If it doesn't exist, we don't really have much of an explanation.

I continue to struggle with this, but my current thinking is that a return value of false should mean "I can't answer the query". If asking if the file exists and is readable, we certainly can answer the query, and the answer (which goes in the data field) is 0 (no). This is the same thing as saying that we should keep the code as it is, but change the docs, and also has the advantage that it can't break anybody who is depending on the current behavior.

lgritz:
I propose instead #1046, just change the docs to match the behavior we have had all along.

darkhorse64:
LGTM !

PR #1042, opened 2015-01-23
Smarter handling of X/Y resolution (i.e. density), resunit, aspect

lgritz:
Smarter handling of X/Y resolution (i.e. density), resunit, aspect for JPEG, TIFF, OpenEXR.

* We weren't reading or writing density for JPEG at all. The xdensity and ydensity in the JPEG header should turn into the standard OIIO attributes "XResolution" and "YResolution".

* JPEG and TIFF both don't have pixel aspect ratio metadata, it's implied by the densities when reading. When writing, if a pixel aspect ratio is requested, you may need to make up densities in order to imply the right aspect ratio.

* OpenEXR only has xDensity (aka XResolution), which we didn't read properly, and aspect ratio, and YResolution (y density) is implied by these when present. Also, OpenEXR dictates that densities are in pixels per inch, so set that upon input, and upon output, if we are given densities in pixels/cm, silently convert to inches to conform to OpenEXR spec.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,10 +3,49 @@ Changes:
 Release 1.6 (in progress) -- compared to 1.5.x
 ----------------------------------------------
 Major new features and improvements:
+* New oiiotool functionality:
+    * --absdiff, --absdiffc compute the absolute difference (abs(A-B)) of
+      two images, or between an image and a constant color. #1029 (1.6.0)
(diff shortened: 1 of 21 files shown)


lgritz:
I looked over this a bit more and realized that TIFF and OpenEXR could also use a touch-up in how they deal with density and aspect ratio. New fixes pushed that makes a more compete fix across all of these plugins.

Issue #1037, opened 2015-01-22
oiio Python binding is misbuilt as a shared library instead of a plugin


nieder:
The python plugin is being built as a SHARED library according to src/python/CMakeLists.txt.  The proper type should be MODULE, since the file OpenImageIO.so is (presumably) not meant to be linked into by other files.

On OS X, shared libraries and loadable modules are different and the current code is problematic because the file is installed into $PREFIX/lib/python/site-packages/OpenImageIO.so, but is given an install name of $PREFIX/lib/OpenImageIO.so.  This mismatch can cause problems.

The patch below changes the oiio python file to the right file type.  It also removes the "if (APPLE)" hack that was needed to give the file the correct file extension so it could be loaded.

https://developer.apple.com/library/mac/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-TPXREF103

````
From 79f446e8f0ff56836a66e642b544b392bd6325ed Mon Sep 17 00:00:00 2001
From: Hanspeter Niederstrasser <niederstrasser@gmail.com>
Date: Thu, 22 Jan 2015 06:04:01 -0600
Subject: [PATCH] python plugin should be built as a MODULE, not a SHARED
 library

---
 src/python/CMakeLists.txt | 7 +------
 1 file changed, 1 insertion(+), 6 deletions(-)

diff --git a/src/python/CMakeLists.txt b/src/python/CMakeLists.txt
index 003fc14..ab583d1 100644
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -83,7 +83,7 @@ if (BOOST_CUSTOM OR Boost_FOUND AND PYTHONLIBS_FOUND)
     message (STATUS "Python to include SO version: ${PYLIB_INCLUDE_SONAME}")
 
     include_directories (${PYTHON_INCLUDE_PATH} ${Boost_INCLUDE_DIRS})
-    add_library (${target_name} SHARED ${python_srcs})
+    add_library (${target_name} MODULE ${python_srcs})
     target_link_libraries (${target_name} OpenImageIO ${Boost_LIBRARIES} ${Boost_PYTHON_LIBRARIES} ${PYTHON_LIBRARIES} ${CMAKE_DL_LIBS})
 
     # Exclude the 'lib' prefix from the name
[...]

PR #1043, opened 2015-01-23
python plugin should be built as a MODULE, not a SHARED library

nieder:
From issue #1037, build the python module as a MODULE, not SHARED library.


lgritz:
LGTM, thanks! I will merge immediately.

lgritz:
Fixes #1037

PR #1048, opened 2015-01-27
Make simd.h safe for older Ilmbase.h.

lgritz:
We have simd::float4 conversions from and to Imath::V4f. Turns out that
prior to OpenEXR 1.7-ish, there was no V4f. To make things extra fun,
back then OpenEXR/IlmBase did not have any handy #defines to let you
conditionally compile based on the Ilmbase version. Now it does.  So we
punt and only define the (rarely used?) V4f conversions when Ilmbase is
safely >= 2.0, which seems safe moving forward, but this header will
still compile without errors for somebody stuck with Ilmbase 1.0.1
(albeit without the ability to convert to and from V4f, which is moot
anyway for those without such a definition).

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -1198,12 +1198,6 @@ public:
 #endif
     }
 
-    /// Construct from a Imath::V4f
-    OIIO_FORCEINLINE float4 (const Imath::V4f &v) { load ((const float *)&v); }
-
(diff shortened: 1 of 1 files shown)


mfvescovi:
Yay! That would be awesome!
In the meanwhile, I'll try to give a huge refresh to IlmBase and OpenEXR in Debian.
Hope it won't be too hard ;-)

PR #1050, opened 2015-01-28
Use empty() to check for emptiness.

mlang:


lgritz:
I don't necessarily object to the change, but I do like to avoid "style churn" -- rewriting of somebody else's working and clear code just because you would have written it using a different, but functionally equivalent, idiom.

Can you explain why you are requesting this change? Is there a difference in efficiency, correctness, or clarity of using vector_or_string.empty() versus vector_or_string.size() == 0?

lgritz:
I see, also, that you have made pull requests to several seemingly unrelated projects recently, all changing .size()==0 to .empty(). This makes me suspicious.

mlang:
Larry Gritz <notifications@github.com> writes:

> I don't necessarily object to the change, but I do like to avoid "style churn" -- rewriting of somebody else's working and clear code just because you would have written it using a different, but functionally equivalent, idiom.

I can see that point.  If you object to the change, simply close the
pull request.

> Can you explain why you are requesting this change? Is there a difference in efficiency, correctness, or clarity of using vector_or_string.empty() versus vector_or_string.size() == 0?

IMO, empty() is indeed more clear to read then size()==0.

Note that this has been found by clang-tidy's
readability-container-size-empty check.

-- 
CYa,
  ⡍⠁⠗⠊⠕

lgritz:
Closing. I just don't think it's any more clear.

PR #1052, opened 2015-01-28
OIIO fixes for compiling OSL on Win32

darkhorse64:
While compiling OSL, I found that my commits from two weeks ago needed to be extended.


lgritz:
LGTM, thanks.

lgritz:
Also back-ported to RB-1.5.

PR #1053, opened 2015-01-29
simd.h refinement: found faster load(...,3) and store(...,3) idioms.

lgritz:
It looks like we're trading simd for serial ops here, but looking at the assembly output shows that gcc and clang (especially clang!) recognize this pattern and turn it into very efficient code. Furthermore, I benchmarked it very carefully, and this makes a definite speedup (significantly so for the 3-float load). I still don't fully understand why my trickery wasn't the fastest method, but the timings could not be more clear, on multiple platforms.

I should clarify: My artificial benchmark that times the load and store of 3-vecs shows marked improvement with this patch. The 3-vec load is recently added and not yet used at all in the code base (except for unit tests), and the 3-vec store is only used in a couple places so far, so this does not yet show any improvement in overall rendering speed. But it will be helpful as we start to use the 3-way load and store in more situations.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -1340,10 +1340,14 @@ public:
             m_vec = _mm_castpd_ps (_mm_load_sd ((const double*)values));
             break;
         case 3:
-            // Trickery: load one double worth of bits, then a float,
-            // and combine.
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

PR #1054, opened 2015-02-06
Another round of fixes for Win32

darkhorse64:
Actually, these fixes are needed to compile OSL, hence the late discovery


lgritz:
LGTM, thanks for the fixes.

lgritz:
Merged into 1.5 as well

PR #1055, opened 2015-02-07
OpenEXR write_tile with AutoStride calculated the wrong strides.

lgritz:
When strides are automatically computed rather than explicitly supplied,
it should have assumed data layout of a "full tile", even if it was the
last tile in a row in an image whose width was not an integral multiple
of the tile width. Instead, it was clamping the width and height by the
size of the remaining data. That didn't match the behavior of read_tile
(which assumed full-tile-size layout, even for edge tiles), nor the
behavior of write_tile for other tiled formats such as TIFF.

Used the occasion to tighten up the imageio.h comments and OIIO
documention on this matter, spelling out more carefully what happens
with AutoStride in write_tile versus write_tiles.

### DIFF
--- a/src/doc/imageinput.tex
+++ b/src/doc/imageinput.tex
@@ -981,7 +981,8 @@ in its native format (including per-channel formats, if applicable).
 The stride values
 give the data spacing of adjacent pixels, scanlines, and volumetric
 slices, respectively (measured in bytes).  Strides set to the special
-value of {\kw AutoStride} imply contiguous data, i.e., \\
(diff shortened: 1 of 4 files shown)


brianhall77:
Thanks, Larry.  I confirmed that this fixes the problem in my test case using AutoStride when writing single tiles to an OpenEXR file that isn't perfectly divided by the tile size.

lgritz:
Merged into master. I'll give this a few days to get tested in the wild, and then back-port to 1.5 if there are no complaints.

PR #1056, opened 2015-02-07
Introduce a CMake option to specify a plugin search path

luhk:
The new PLUGIN_SEARCH_PATH option allows for specifying a default plugin
search path at build time. More precisely, the plugin_searchpath
attribute is assigned the value of PLUGIN_SEARCH_PATH by default.

This makes it possible for example to simple install custom plugins to
PLUGIN_SEARCH_PATH without having to fiddle around with environment
variables (i. e. OIIO_LIBRARY_PATH).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,6 +278,7 @@ if (UNIX)
     set (MAN_INSTALL_DIR ${DEFAULT_MAN_INSTALL_DIR} CACHE STRING
          "Install location for manual pages (relative to CMAKE_INSTALL_PREFIX or absolute)")
 endif()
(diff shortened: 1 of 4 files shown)


lgritz:
LGTM. I will merge this.

lgritz:
Is this ok for master, or do you need it backported to 1.5?

luhk:
A backport to 1.5 would be very nice.

lgritz:
Done.

luhk:
Thanks!

lgritz:
I am going to rename the internal name of the definition, however, to prefix it with OIIO rather than PVT.

PR #1057, opened 2015-02-07
ImageCache / TextureSystem handle-based methods

lgritz:
I'm not really sure if I like this or not, but I'm submitting it for comment and discussion. I'm prepared to either merge it, or largely abandon it, depending on your opinions.

TextureSystem already has a way to retrieve an opaque pointer that serves as a "handle" to a texture, and varieties of texture(), texture3d(), and environment() that take a handle rather than a ustring name. This is intended to allow a slightly faster way to access texture (by bypassing the name->handle lookup that would otherwise be done) if the calling app itself is able to keep track of the handle for the texture needed.

This patch:

1. Extends similar functionality to the TextureSystem methods that don't already have handle-based varieties, such as get_texels(), get_texture_info(), get_imagespec().

2. Extends the approach to ImageCache, allowing IC clients to retrieve a handle and then optionally use that rather than the filename as a file specifier for various IC methods.

The best thing about this patch is that it fixes the asymmetry of having a few TS/IC methods that have both filename-based and handle-based varieties, while others only exist in a form that takes a filename. And presumably, these new versions are slightly faster (5-10%, in previous timings I'd done for the texture lookup functions). But the question is: is this extra clutter worth it? After all, it's important to have a 5% faster version of texture() because it's called so often. But how many times would you realistically call get_texture_info on the same file? Not millions. So I'm unsure if this patch is helpful, or just needless tinkering. I don't see how it would *hurt*, other than making these classes slightly more complicated. Feedback is appreciated.


lgritz:
Chris Kulla approved this over IM.

PR #1058, opened 2015-02-10
oiiotool: gracefully handle impossible requests to save too many channels.

lgritz:
Before saving and encountering an error, check that the output file type supports the number of channels in the image being saved. If not, don't consider it an error, but rather just give a warning, and then do the best we can. What we try to do is: (a) if "R", "G", and "B" (and also "A", if the format supports it) are all in the list of channel names for the image that is being saved, save just those channels; (b) if those names are not found, just punt and save the first three channels (or 4, if the output supports alpha).

In order to facilitate this, we add two new tokens to ImageOutput::supports(): supports("alpha") if the file format supports saving an alpha channel, and supports("nchannels") if the file format supports saving an arbitrary number of channels. (We assume that ALL formats we care about can save a minimum of 3 channels, for RGB.)

As an example, we go from this:

    $ oiiotool fivechannels.exr -o out.jpg
    oiiotool ERROR: output (jpeg does not support 5-channel images)

and no output saved, to this:

    $ oiiotool fivechannels.exr -o out.jpg
    oiiotool WARNING: output (Can't save 5 channels to jpeg... saving only R,G,B)

and with a JPEG file saved (albeit with only the R,G,B channels of the original).

### DIFF
--- a/src/bmp.imageio/bmp_pvt.h
+++ b/src/bmp.imageio/bmp_pvt.h
@@ -169,7 +169,7 @@ class BmpOutput : public ImageOutput {
     BmpOutput () { init (); }
     virtual ~BmpOutput () { close (); }
     virtual const char *format_name (void) const { return "bmp"; }
-    virtual bool supports (const std::string &feature) const {return false;}
+    virtual bool supports (const std::string &feature) const;
     virtual bool open (const std::string &name, const ImageSpec &spec,
                        OpenMode mode);
     virtual bool close (void);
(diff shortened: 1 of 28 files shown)


lgritz:
Anybody object to this?

justinfx:
I was able to give this a test and it does seem to be solving the issue! It is definitely nice to have some default no-fail behavior that gives you a valid image. I tried it out with both a N-channel exr that does have proper R,G,B labels, and one that has completely arbitrary names, or just an alpha. Does exactly what I had hoped. Thanks!

lgritz:
Great, thanks much for the test.

You probably want a backport to 1.5, yes?

justinfx:
If it's not too much trouble :-)

lgritz:
No problem, coming right up.

lgritz:
Done.

justinfx:
You rock man!

PR #1059, opened 2015-02-11
maketx: Fix botched filter with crop windows that need resize

lgritz:
There was one code path, that only comes into play when the source image
is a crop (data window != display window) that didn't take into
consideration the special names we use for requesting MIP map
sharpening, and it was issuing an "unknown filter name" error.


fpsunflower:
LGTM

PR #1060, opened 2015-02-13
Bug fix for oiiotool frame sequences -- passed wrong variable

lgritz:
As reported on the mail list, an occasional crash with file sequences appears to boil down to this.

If I change the 'DASSERT' to 'ASSERT', the old code hits it, so that was certainly wrong. New code is fine. I'm keeping the assert, it keeps us honest.


lgritz:
The patch was verified to solve the problem. I've merged into master, RB-1.5, and RB-1.4.

PR #1063, opened 2015-02-15
ImageSpec minor additions & fixes

lgritz:
* Add ImageSpec::find_attribute() that works for fixed fields in the spec like "width" (in addition to named metadata in extra_attribs).  But needs to be passed a temporary ImageIOParameter as scratch. By using this, we make get_int_attribute, get_float_attribute, and get_string_attribute also able to retrieve those fields. For example, imagespec.get_int_attribute("nchannels") might return 3, or imagespec.get_string_attribute("geom") might return "640x480+0+0".

* ImageSpec::metadata_val should be static (not merely const), since it doesn't need 'this' at all.

* ImageSpec::erase_attribute() fixups:
    1. Document it (had somehow gone undocumented in the PDF)
    2. Set up Python bindings for it (was missing)
    3. Minor touch-ups in describing the workings of find_attribute.

### DIFF
--- a/src/doc/imageioapi.tex
+++ b/src/doc/imageioapi.tex
@@ -588,10 +588,10 @@ function for when you know you are just looking for a simple string value.
 \apiend
 
 
-\apiitem{std::string {\ce metadata_val} (const ImageIOParamaeter \&p,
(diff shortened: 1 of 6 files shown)


lgritz:
No objections -> merge

PR #1064, opened 2015-02-17
oiiotool: expression evaluation and substitution for command line arguments

lgritz:
I've been thinking about this for a long time, and finally did it as a weekend project.

Motivating examples that illustrate how this can be used for fun and profit:

    # create a gradient the same size as fg.exr, composite fg over it
    oiiotool fg.exr --pattern fill:top=0,0,0,0:bottom=0,0,1,1 {TOP.geom} {TOP.nchannels} -over -o overgradient.exr

    # trim 20 pixels from all sides
    oiiotool input.exr -cut "{TOP.width-40}x{TOP.height-40}+{TOP.x+20}+{TOP.y+20}" -o cropped.exr

    # Write the time in the lower left corner of the image
    oiiotool -v input.jpg --text:size=30:x=30:y={TOP.height-30} {TOP.DateTime} -o out.jpg

Anything enclosed in braces { } will undgo expression substitution.

Currently, expressions may be:
* number
* Simple arithmetic: expr + expr, expr - expr, expr * expr, expr / expr
* Retrieve image metadata:
     *  TOP.metadata         current image
     *  IMG[i].metadata      ith image on the stack (IMG[0] is TOP)
     *  IMG[name].metadata   named image

We may allow more complex expressions later, but this seems like a good start for now.

Feedback appreciated. And what clever things can you think to do with this feature?

(Note: this review incorporates pull request #1062 and #1063. You need not review those again here if you don't want. All the changes specific to this feature are on oiiotool.h and oiiotool.cpp, though they make use of the new features in those other two PR's.)


lgritz:
Response to this proposal internally at my facility has been very positive. I'll let it sit for a few more days to gather feedback and/or reviews by anybody looking over the code. But by early next week, if there are no objections, I will  merge it.

lgritz:
Only positive comments (offline), no objections, so I will merge this.

justinfx:
Sorry to dredge up an old merge request, but I had a question about the expression support. Am I correct that it isn't currently complex enough to handle something like expressing a "center padded" fit? Something along the lines of:

```
$ oiiotool -v in.0001.exr --create 2048x1556 3 --paste +0+202 -o out.0001.jpg
``` 

... where the input size is 2048x1152, and I would like to center pad to 2048x1556, but dynamically offset it. It would have to support something like `--paste +0+{(IMG[1] - IMG[0]) / 2}` right? Or am I just completely overlooking another way to accomplish this output, without needing expressions?

justinfx:
And I realize I could have written this like:

```
$ oiiotool -v in.0001.exr --fit 2048x1556 --crop 2048x1556+0+202 -o out.0001.jpg
```

But still wondering if expressions can be used to offset dynamically.

lgritz:
You were very close, just need to change the image references `IMG[0]` to include `.width` or whatever to specify the metadata:

    --paste +0+{(IMG[1].width-IMG[0].width)/2}+{(IMG[1].height-IMG[0].height)/2}

I think that should work.

justinfx:
Ah, thanks for clarifying. And I also just realized that this feature wasn't merged into 1.5. At my studio I think our latest available build is 1.5.16 :-/

lgritz:
Aha, yes, that's a new 1.6/master feature. IIRC, oiiotool.cpp has diverged quite a bit and backporting would be a big pain.

If it's any consolation, I am aiming to branch 1.6 relatively soon. Next few weeks, I hope.

PR #1066, opened 2015-02-18
JPEG: write density fields into header properly.

lgritz:
Didn't realize that jpeg_set_defaults would clobber those fields.


lgritz:
Reported on oiio-dev as fixing the reported problem, so I merged into master and RB-1.5.

PR #1067, opened 2015-02-19
Change 'bool supports(const std::string&)' to 'int supports(string_view)'

lgritz:
By changing to int, we allow future queries to return more information than just a bool (for example, perhaps a supports("maxchannels") would return the maximum number of channels allowed by the format). By changing the argument to a string_view, we make the call less expensive by no longer requiring a temporary string be constructed in many cases.

This should not require any changes to client-side source code, since all existing queries just return true or false (1 or 0), and the string_view can accept either a char* or a std::string (or for that matter, a ustring).

However, if you have written any custom ImageInput or ImageOutput plugins, you will need to change the signature of any supports() methods in your
subclasses.

I have bumped the OIIO_PLUGIN_VERSION, and after merging will also bump the OIIO release version.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -72,6 +72,9 @@ Public API changes:
    #1063 (1.6.1)
  * Python bindings:
     * Added previously-M.I.A. ImageSpec::erase_attribute(). #1063 (1.6.1)
+ * ImageInput and ImageOutput supports() method has been changed to accept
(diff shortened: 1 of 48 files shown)


lgritz:
Any comments or objections?

lgritz:
I will merge this tomorrow unless there are objections voiced.

PR #1070, opened 2015-02-25
Filesystem improvements

lgritz:
1. From Max Liani: improve exception safety of Filesystem::get_directory_entries. 

2. From LG: add Filesystem::rename(), and fix Filesystem::copy() which got the sense of the error code backwards.


lgritz:
LGTM.

lgritz:
backported to RB-1.5 as well

PR #1071, opened 2015-02-25
Odds and ends from Max Liani

lgritz:
Submitting on behalf of Max:

1. In formatspec.cpp, simplify format_raw_metadata and fix the string case to handle string arrays.

2. Imagecache: improve printing of IO rates when the time spent is 0 (avoid divide by zero)

3. Comment typo fixes


lgritz:
LGTM.

lgritz:
backported to RB-1.5 as well

PR #1072, opened 2015-02-25
maketx improvement: write to temp file, rename upon successful completion

lgritz:
This improves robustness by eliminating the possibility of a half-written texture file that lacks all the MIPmap levels (which might be inexplicably slow to render if nobody realizes it's a partial file) or is otherwise malformed in a way that leads to chaos. Basically, a crash or failure no longer will write to the output filename, only to a temporary file. The temp file is renamed/moved when the full file write is complete without errors.

Submitted on behalf of Max Liani. (With just a teensy bit of minor editing by LG.)

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -857,6 +857,14 @@ make_texture_impl (ImageBufAlgo::MakeTextureMode mode,
         }
     }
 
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, and I checked with our texture pipeline developer and got a thumb's up (we already employed a similar strategy externally to maketx, so will be nice to make it a feature).

lgritz:
backported to RB-1.5 as well

PR #1073, opened 2015-02-25
refactored and extended lanczos filter

Tristimulus:
Hi Larry.

I've refactored the lanczos filter code a bit to make it easier to add more filter kernel sizes. I've also added the lanczos2 filter to match the "Lanczos4" filter in Nuke. (It looks like the Foundry labels them based on the diameter of the filter kernel instead of the radius. OIIO "lanczos3" seems to match Nukes "Lanczos6".)
All the lanczos filters now share a base filter class which has a size parameter of 1.0.
I've also re-assigned the "lanczos" filter name in the FilterxD::create factory method to this new base filter. Previously it looks like it would have created a lanczos3 filter instead, but this doesn't actually seem to work in oiiotool anyway.

Further I've also fixed 2 issues that were causing compiler warnings.

Thank you and kind regards,

Matthias

### DIFF
--- a/src/libutil/filter.cpp
+++ b/src/libutil/filter.cpp
@@ -273,7 +273,7 @@ public:
     }
     string_view name (void) const { return "blackman-harris"; }
     static float bh1d (float x) {
-	if (x < -1.0f || x > 1.0f)  // Early out if outside filter range
+    if (x < -1.0f || x > 1.0f)  // Early out if outside filter range
             return 0.0f;
         // Compute BH.  Straight from classic BH paper, but the usual
         // formula assumes that the filter is centered at 0.5, so scale:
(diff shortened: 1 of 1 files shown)


lgritz:
Our naming scheme is based on the number of lobes (this is common), not on the width.

lgritz:
I'm not sure this is quite the right approach. Add "lanczos2" (2-lobed Lanczos) if you want, but it is considered by most to be a different filter, not just a wider or narrower scaled version. The "width" parameter to the constructor should be ignored, as it was before, because Lanczos filters don't scale by width (the way Gaussian or Box does). They are a fixed width by definition. The difference between the 2-lobe and 3-lobe Lanczos filters can be interpreted as a difference in "window", but only certain discrete values make sense. You wouldn't want it possible for somebody to request 2.5.

lgritz:
Never got a reply, amendment, or anybody asking me to fix or merge it. Closing.

PR #1074, opened 2015-02-26
TypeDesc - changes to better support arrays of unspecified length

lgritz:
* Allow 'equivalent' comparison to allow an unspecified-length array to
  match any array of the same type.


lgritz:
It case it wasn't clear, this is a patch that came from Max Liani (with very minor modification from me).

So LGTM!

lgritz:
backported to RB-1.5 as well

PR #1080, opened 2015-03-02
IC/TS: add create_thread_info() and destroy_thread_info() methods

lgritz:
ImageCache/TextureSystem: add create_thread_info() and destroy_thread_info() methods that let an app explicitly create an IC/TS per-thread info that the app can completely manage (and is responsible for destroying).

This augments the existing get_perthread_info(), which retrieves a unique Perthread* from a thread-specific pointer (whose storage and lifetime is controlled by the library, not the app).

The addition of the explicit create/destroy allows very savvy apps to pre-allocate PerThreadInfo that can persist even as threads come and go, and avoid the thread-specific-pointer retrieval. But beware -- it is up to the app to ensure that any created Perthread info (a) is eventually destroyed, and (b) under no circumstances may be shared by more than one thread at the same time. Bad things will happen. Don't cross the streams.

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -122,12 +122,22 @@ public:
 
     /// Define an opaque data type that allows us to have a pointer
     /// to certain per-thread information that the ImageCache maintains.
+    /// Any given one of these should NEVER be shared between running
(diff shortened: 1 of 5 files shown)


fpsunflower:
LGTM

PR #1081, opened 2015-03-03
Fix build error in float4.xyz0() with SSE version < 4

brechtvl:
Please backport to RB-1.5, since the same issue exists there.


lgritz:
Duh, yes, sorry, I will merge into both branches immediately.

Note to self: any time I mess with SSE, make sure it passes all the tests even when compiled with USE_SIMD=0.

lgritz:
Merged into both branches. Sorry again, that's one I have no good excuse for not catching.

brechtvl:
Thanks! It was just a minor inconvenience for me :)

PR #1082, opened 2015-03-04
allow e.g. %04d to match more than 4 digits, e.g. 10000

mjmvisser:
A tiny patch (1 character!) that allows %0Nd to match more than N digits. For example, if a sequence overflows 4 digits, frame 10000 should still be matched. 

This behavior is consistent with Nuke and RV.

cheers,
-Mark


lgritz:
LGTM. I will merge.

lgritz:
Do you need a backport to 1.5?

PR #1083, opened 2015-03-07
Fixes to texture handle changes

lgritz:
Ugh, I did it wrong. The find_file methods that did the name to IC handle lookups did more than just that -- that's also where the ImageInputs were opened (if not yet opened), duplicates found, 'used' flags updated, errors reported, etc.

It wasn't symptomatic until we started messing around with invalidating the files in the cache, which doesn't invalidate the handles themselves (and shouldn't), but at that point the files would be closed, and the re-opening wouldn't happen properly without the call to find_file, which of course wouldn't happen at all since the handle was already retrieved.

The solution is to split into separate "find" (just convert name to file cache handle, creating if necessary), and "verify" (open and other housekeeping and error checking). It's mostly just splitting the existing find_file in half, then cleaning up the many many consequent changes that make use of that properly. The name-based functions just call find, and then pass to the handle-based functions. The handle-based functions need to call "verify" on the handle it has, to ensure that the handle is ready for use. Once this split was done, it also became clear that a thread_info pointer needed to be passed to some routines where we didn't pass it before. There shouldn't be any performance hit from this, it's just re-apportioning the same amount of work across two functions.

Pulling on this piece of yarn unraveled much of the sweater, but I think I've got it put back together correctly now.

### DIFF
--- a/src/doc/imagecache.tex
+++ b/src/doc/imagecache.tex
@@ -389,16 +389,41 @@ settings of named options.  For example,
 \subsection{Opaque data for performance lookups}
 \label{sec:imagecache:api:opaque}
 
-\apiitem{Perthread * {\ce get_perthread_info} ()}
+\apiitem{Perthread * {\ce get_perthread_info} (Perthread *thread_info=NULL) \\
+Perthread * {\ce create_perthread_info} () \\
+void {\ce destroy_perthread_info} (Perthread *thread_info)}
 \indexapi{get_perthread_info}
+\indexapi{create_perthread_info} \indexapi{destroy_perthread_info}
 \NEW % 1.6
-Retrieves an opaque handle for per-thread info, to be used for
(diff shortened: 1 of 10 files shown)


lgritz:
Updated the review with further changes:

More IC/TS perthread fixes

App-owned Perthead pointers (the ones preallocated by create_perthread_info) had no mechanism for clearing their per-thread microcaches of tile references even after they were marked as "purge." That's because the actual purging was previously done in the get_perthread_info function that was used to retrieve the true thread-specific-pointers. This could lead to inconsistent texture results for different threads, if an invalidate() for that file had happened at just the right time.

The fix the strategy is that now get_perthread_info takes an optional Perthread* argument, which if supplied is used, and only if NULL is passed does it retrieve and/or allocate a TSP. Then in either case, it does the other housekeeping (including purging, when requested) before returning the pointer (either the supplied one passed in, or the created/retrieved one).

PR #1086, opened 2015-03-07
Disable OpenSSL by default

lgritz:
A while back, we found that using OpenSSL's SHA-1 implementation rather than our usual one resulted in a significant speedup of maketx.

I've done some retimings, now that a lot else has changed about maketx's internals (including adopting the practice of hashing blocks of scanlines, making it amenable to multithreading), and found that the speedup we get by using OpenSLL's SHA-1 is barely measurable. The minuscule speedup doesn't seem worth the trouble of imposing OpenSSL as a dependency.

In light of this, I propose this patch to change the default to using the SHA-1 implementation that's compiled in, thus removing OpenSSL as a dependency.

Now it's still an option, mainly so I can continue to compare the two methods. But it's possible that in the future, we may remove the potential for using OpenSSL altogether.


fpsunflower:
LGTM

Reducing the number of dependencies is always a good thing :+1:

PR #1090, opened 2015-03-18
strhash replacement -- improve hashing of large strings by 10x or more

lgritz:
A non-OIIO app that happens to use OIIO's ustring extensively was performing badly when dealing with an absurd number of very long strings. We experimentally found that our Strutil::strhash (and therefore all the underlying hashing for ustring) was a primary culprit.

* Add Google's farmhash (MIT license) and update our prior use of xxhash to a more modern version that includes a nice fast 64 bit hash.

* Lots of benchmarking indicates that old strhash was the worst of the bunch.

* Settled on farmhash as the best replacement for strhash's guts.  For extremely short strings, it's slightly slower than the old implementation, but the crossover point is ~6 chars, and the longer the string, the bigger the speedup.  It's well over 10x faster than the old implementation for long strings (100-500 chars).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -162,6 +162,11 @@ Developer goodies / internals:
    AxBxCxDx, blend0not (1.6.2)
  * array_view enhancements that let you initialize an array_view<const float>
    from a const std::vector<float>&.  #1084 (1.6.2/1.5.14)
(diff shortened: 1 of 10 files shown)


fpsunflower:
LGTM

PR #1091, opened 2015-03-18
Modify ustring_test to give it thread wedging benchmark abilities

lgritz:
Modify ustring_test to give it thread wedging benchmark abilities
(modeled after spinlock_test)

To benchmark its thread performance, you'd use it like this (hint for Chris):

    build/spinux1/src/libutil/ustring_test -wedge -trials 5 -iters 10000000 -threads 32


fpsunflower:
Just what I needed! LGTM

PR #1092, opened 2015-03-20
Free memory used by UstringTable on exit

Nazg-Gul:
Previously values of this map were only allocated in make_unique()
and never freed afterwards. This isn't harmful, but makes it much
more difficult to troubleshoot possible leaks in software which
uses OpenImageIO,

Added a destructor to UstringTable now which frees memory used
by map values.

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -83,11 +83,21 @@ typedef null_lock<null_mutex> ustring_write_lock_t;
 #define USE_CONCURRENT_MAP 1
 #endif
 
+class UstringTable : public
 #if USE_CONCURRENT_MAP
-typedef unordered_map_concurrent <string_view, ustring::TableRep *, Strutil::StringHash, Strutil::StringEqual, 8> UstringTable;
+    unordered_map_concurrent <string_view, ustring::TableRep *, Strutil::StringHash, Strutil::StringEqual, 8>
 #else
-typedef boost::unordered_map <string_view, ustring::TableRep *, Strutil::StringHash, Strutil::StringEqual> UstringTable;
+    boost::unordered_map <string_view, ustring::TableRep *, Strutil::StringHash, Strutil::StringEqual>
 #endif
+{
+ public:
+   ~UstringTable() {
+       for (UstringTable::iterator it = begin(); it != end(); ++it) {
+           ustring::TableRep *rep = it->second;
+           free(rep);
+       }
+   }
+};
 
 std::string ustring::empty_std_string ("");
 


fpsunflower:
Have you measured the difference in shutdown time with this added? I am actually in the process of making a new implementation of `ustring` and I found that I can make it quite a bit faster by just letting memory leak (since by definition there is no way to free a ustring).

Would you be willing to wrap the code in `#ifndef NDEBUG` so that it only gets executed on debug builds?

Nazg-Gul:
For our case of blender it's nearly unmeasurable difference of 0.00001s. But we're using jemalloc, which si a good idea on linux for us because of much less memory fragmentation.

I'll be fine if the code is wrapped with either NDEBUG check. Also just throwing an idea: could be a cmake option disabled by default?

Also, if the speed is critical, could worth using memory pool/arena?

fpsunflower:
Yes speed is critical, you would be surprised how much memory we spend on strings in our renderer :) Of course its only going to be noticeable for cases with multiple millions of strings added. For example try doing:
`ustring_test --threads 24 --trials 1 --iters 100000000`

This will create 100M strings which is a good stress test. In the current implementation, this take a long time to shutdown after the benchmark is over.

My new implementation does use memory pools, but simply throws away the old pointers when allocating new blocks. I'm not sure its worth the overhead of tracking them just so they can be released.

The other problem with releasing ustrings on shutdown is the old "destruction order fiasco", any global destructors that happen to run ''after'' the global ustring one but access a `ustring` in some way could trigger a very sneaky crash.

lgritz:
How about instead, add an explicit clear() method, whose dangerous use is only for apps that want to reclaim all memory at shutdown? Another use is for benchmarks that want to do multiple time trials and have every one do the same series of allocations.

Nazg-Gul:
That's actually a good point of destructor order fiasco. We're just lucky enough not to use ustring from any destructors.

An explicit clear() method would be like the only reliable way to deal with this. It'll totally work for us, so if it also doesn't cause troubles from your side then would consider this approach having lest degree of "suckness" :)

Nazg-Gul:
Forgot to ask, how to process with the change? I'm totally up to updating patch so it'll introduce a dedicated clear() callback. But should i do it against master, or the ustring refactor branch is published somewhere so i can create patch against it or i'd need to wait for the ustring refactor to be merged to master?

lgritz:
Maybe your existing patch is appropriate against RB-1.5 if you want the current release to have this feature, but for master, wait until Chris submits his patch because he has totally overhauled the internals and almost certainly your patch will be entirely wrong for the new implementation.

fpsunflower:
I am just polishing up the new implementation now, I will post a pull request against master.

I would recommend waiting for that patch to land in master, and then implementing `clear()` on top of that one.

lgritz:
I'm not inclined to backport Chris's total internal ustring overhaul to any release branch (at least, not without some time passing to make sure it's well exercised in a production environment). So if you want the ability to free the table in RB-1.5, you may want to keep the patch you've already written against the "old" implementation and submit a PR for it against RB-1.5.

But if you're only concerned with the future, then just re-implement it atop Chris's forthcoming patch.

fpsunflower:
New patch was just reviewed and pushed to master. It should be self-explanatory, it just drops most `malloc()` calls without worrying about freeing them. At the same time, it shouldn't be too hard to maintain a short stack of allocated pools so you can free them in a `clear()` method.

lgritz:
Closing. I think Chris' work a few years ago superseded this. I don't think there is any harm in just dropping the allocations when the app exists.

PR #1093, opened 2015-03-20
Fixes for OpenJpeg/Jpeg2000

Nazg-Gul:
There are couple of things here:

First of all, make OIIO look for j2k libraries in the multiarch folder on linux

Second of all, it seems alpha wasn't handled totally correct as we believe. This in fact was already submitted and some initial discussion started in #943 but while preparing other patches i needed to update master branch, which closed the original pull request. Not sure i can point existing pull-request to a different branch, so made a new pull-request from a dedicated branch.

Would be cool to nail down what's the proper way to deal with alpha in j2k :)

### DIFF
--- a/src/cmake/modules/FindOpenJpeg.cmake
+++ b/src/cmake/modules/FindOpenJpeg.cmake
@@ -67,6 +67,7 @@ set (OpenJpeg_include_paths
 
 set (OpenJpeg_library_paths
   /usr/lib
+  /usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}
   /usr/local/lib
   /sw/lib
   /opt/local/lib)
--- a/src/jpeg2000.imageio/jpeg2000input.cpp
+++ b/src/jpeg2000.imageio/jpeg2000input.cpp
@@ -38,8 +38,46 @@
 
 OIIO_PLUGIN_NAMESPACE_BEGIN
 
+namespace {
 
-static void openjpeg_dummy_callback(const char*, void*) {}
+void openjpeg_dummy_callback(const char*, void*) {}
+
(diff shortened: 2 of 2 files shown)


lgritz:
Seems reasonable, LGTM.

This patch assumes that the only alpha you'll ever find in a JPEG-2000 file is "unassociated alpha", i.e. the colors are NOT already "premultiplied" by the alpha value. Can you confirm that this is your understanding of the J2K spec? Do you have any documentation that this is the case?

lgritz:
Oh, just noticed that now you are premultiplying the colors upon reading, but you have not altered jpeg2000output.cpp to do the inverse when writing -- that is, divide colors by alpha (like we do for PNG and other unassociated-alpha-only formats).

Without that change, our writer would not be producing output that our reader would correctly interpret, and vice versa.

lgritz:
If you don't want to bother, let me know, and I can pretty easily do the output part myself and commit both changes at the same time.

lgritz:
Yeah, don't worry about this. I can easily fix the writer. 

Also, in looking this over, I have a number of additional fixes for JPEG-2000 beyond your patch. I will review those separately.

Nazg-Gul:
Ok, cool. Thanks for the reviews :)

lgritz:
I'm going to merge what you've done -- which is definitely a step forward -- and close the ticket. Be aware that the "jpeg2000" test will fail, because the writer and reader no longer match. But this will be temporary, I will submit a fix (and more!) hopefully later today.

Sergey, do you need the JPEG2000 improvements backported to RB-1.5, or is it sufficient to have it in master?

Nazg-Gul:
Awesome. And it should sufficient to have it in master.

lgritz:
More J2K improvements in #1097

PR #1094, opened 2015-03-20
fmath: safe_fast_pow should special-case x^2.0

lgritz:
fmath: safe_fast_pow should special-case x^2.0 to have as much precision and speed as multiply. We're assuming that ^2.0 issues will be way more noticeable than other integer powers, which we'll continue to sweep under the rug. If you want more exact, use safe_pow. But getting squaring right seems like the right balance.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -883,8 +883,8 @@ inline T safe_logb (T x) {
 #endif
 }
 
-/// Safe pow: clamp the domain so it never returns Inf or NaN has divide by
-/// zero error.
+/// Safe pow: clamp the domain so it never returns Inf or NaN or has divide
+/// by zero error.
 template <typename T>
 inline T safe_pow (T x, T y) {
     if (y == T(0)) return T(1);
@@ -893,11 +893,8 @@ inline T safe_pow (T x, T y) {
     if ((x < T(0)) && (y != floor(y))) return T(0);
     // FIXME: this does not match "fast" variant because clamping limits are different
     T r = std::pow(x, y);
-    // Clamp to avoid Inf values.  We purposely clamp at considerably
-    // less than FLOAT_MAX (but still bigger than anyone is likely to
-    // need) so that subsequent additions or multiplications are less
-    // likely to overflow and end up with an Inf right afterwards.
-    const T big = T(1e30);
+    // Clamp to avoid returning Inf.
+    const T big = std::numeric_limits<T>::max();
     return clamp (r, -big, big);
 }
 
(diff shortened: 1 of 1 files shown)


lgritz:
Updated... I did some more tests that indicated that pow(x,1) also added inaccuracy, so also special case that. I think other powers are much less common and less noticeable if they have a few ULPs of imprecision.

src/include/OpenImageIO/fmath.h
```diff
@@ -897,7 +897,7 @@ inline T safe_pow (T x, T y) {
     // less than FLOAT_MAX (but still bigger than anyone is likely to
     // need) so that subsequent additions or multiplications are less
     // likely to overflow and end up with an Inf right afterwards.
-    const T big = T(1e30);
+    const T big = std::numeric_limits<T>::max();
```
fpsunflower: Any reason to change this constant? The comment above would have to be updated ...

lgritz: It seemed more principled to use the max float value, and also the 1e30 limit imposed here in safe_pow was nowhere near what would result from fast_safe_pow (something e37).

fpsunflower: I'm ok with it, but you just might want to rewrite the comment above?

It specifically says we picked a value less than `FLOAT_MAX` to avoid subsequent overflows.

lgritz: Ah, right. Do you think those "subsequent overflows" are likely to happen? There are only a few particular ops where we try to clamp Inf overflows. We certainly don't do it for ordinary multiplication or addition. I guess returning FLOAT_MAX means we have no wiggle room at all, but I don't honestly know if it matters.

fpsunflower: I agree. `pow` with large arguments is probably not that common. We do our best to safeguard the common pitfalls where its cheap to do so, but there's no particularly good default for this one.

Nuke the comment and it LGTM.

lgritz: OK, I'll fix up the comment.

I think that any large constant we pick can have subsequent overflows (e.g. multiply the result of two pow's that got clamped to 1e30, and you still have Inf). Clamping at 1e30 could have cases that get silently clamped to something several orders of magnitude less than their perfectly valid results. It's a tradeoff I don't have a 100% confident answer for. But I think clamping at FLOAT_MAX, even with no wiggle room, is at least an *objective* value if the main goal is just to prevent safe_pow from returning Inf. Any other value is a totally arbitrary choice.

lgritz:
Updated with fixed comments.

PR #1095, opened 2015-03-20
New optimized ustring implementation of make_unique

fpsunflower:
This new implementation uses a custom data structure specific to this use case. It avoids re-hashing the string multiple times redundantly and uses memory pools to reduce the cost of allocations. It maintains the current strategy of using a binned hash table for concurrency, but uses rw spin locks to improve scalability. In my benchmarks, this new implementation is 2x to 3x faster.

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -67,7 +67,12 @@ typedef null_lock<null_mutex> ustring_read_lock_t;
 typedef null_lock<null_mutex> ustring_write_lock_t;
 #endif
 
+// 0 turns off the custom implementation
+// 1 is vanilla custom map
+// 2 is the optimized concurrent custom map
+#define USE_CUSTOM_MAP      2
 
+#if !USE_CUSTOM_MAP
 #if defined(__i386__) && !defined(__clang__) && !defined(_MSC_VER)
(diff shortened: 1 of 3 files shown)


src/libutil/ustring.cpp
```diff
@@ -89,11 +94,191 @@ typedef unordered_map_concurrent <string_view, ustring::TableRep *, Strutil::Str
...
+
+private:
+    enum { NUM_BINS = 32 }; // NOTE: must be power of 2
+
+    typedef TableRepMap<(1 << 20) / NUM_BINS, (4 << 20) / NUM_BINS> Bin;
+
+    Bin bins[NUM_BINS];
+
+    Bin& whichbin(size_t hash) {
+        size_t h = (size_t) murmur::fmix (uint64_t(hash));  // scramble again
```
lgritz: Since you're masking the 64 bit hash to use low bits only (based on table size), why should this extra call to fmix() be superior to the less-expensive technique of using higher bits from the hash to determine the bin? If it's a strong hash, those should be uncorrelated to the low bits you use for the slot within the bin, no?

lgritz: I mean, it's fine with me if the fmix approach is not detectably slower. It just surprises me.

lgritz:
Outstanding results! LGTM.

fpsunflower:
Merged.

I will double check the `fmix()` vs. high order bits thing again. But `fmix` seemed to be ever so slightly faster.

PR #1097, opened 2015-03-22
JPEG-2000 improvements

lgritz:
Reading:
* Fix horribly broken YUV->RGB conversion
* Correctly recognize sRGB encoding
* Pass along the ICC profile, if present
* Handle all bit depth precisions (previously, only 8, 10, 12, 16)
* Set the full/display window correctly
* Deal with differing per-channel data windows and sampling rates

Writing:
* De-associate alpha to match reader (and apparently the J2K spec).
* Handle all bit depth precisions
* Output ICC profile.

### DIFF
--- a/src/jpeg2000.imageio/jpeg2000input.cpp
+++ b/src/jpeg2000.imageio/jpeg2000input.cpp
@@ -33,6 +33,7 @@
 #include "OpenImageIO/filesystem.h"
 #include "OpenImageIO/fmath.h"
 #include "OpenImageIO/imageio.h"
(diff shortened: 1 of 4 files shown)


lgritz:
Any opinions on this, @Nazg-Gul?

Nazg-Gul:
Ah, for some reason thought it was merged. From reading the description and changes themselves it seems all great. Can do tests a bit later today.

lgritz:
OK, thanks, I merged. Works ok for me, let me know if you notice any trouble with it.

PR #1101, opened 2015-03-26
Deal with older version of libopenjpeg not having icc fields.

lgritz:
And having no principled way to detect the version from the header.


fpsunflower:
LGTM

Issue #1102, opened 2015-03-27
Compiling Issue: CentOS 7 (64 bit)


propersquid:
Hey, I'm trying to compile OIIO on the master branch (updated just before submitting ticket) on CentOS 7, and I'm getting an error. I compiled it through the make file in oiio, and I've tried make nuke && make to verify it isn't some left over cruft from a previous compile.

```
[ 85%] Building CXX object src/libOpenImageIO/CMakeFiles/imagespec_test.dir/imagespec_test.cpp.o
In file included from /home/scott/Repositories/oiio/src/libOpenImageIO/imagespec_test.cpp:34:0:
/home/scott/Repositories/oiio/src/libOpenImageIO/imagespec_test.cpp: In function ‘void test_get_attribute()’:
/home/scott/Repositories/oiio/src/include/OpenImageIO/unittest.h:68:60: error: passing NULL to non-pointer argument 1 of ‘std::basic_ostream<_CharT, _Traits>::__ostream_type& std::basic_ostream<_CharT, _Traits>::operator<<(long int) [with _CharT = char; _Traits = std::char_traits<char>; std::basic_ostream<_CharT, _Traits>::__ostream_type = std::basic_ostream<char>]’ [-Werror=conversion-null]
              << "\tvalues were '" << (x) << "' and '" << (y) << "'\n"), \
                                                            ^
/home/scott/Repositories/oiio/src/libOpenImageIO/imagespec_test.cpp:226:5: note: in expansion of macro ‘OIIO_CHECK_NE’
     OIIO_CHECK_NE    (spec.find_attribute("foo"), NULL);
     ^
/home/scott/Repositories/oiio/src/include/OpenImageIO/unittest.h:68:60: error: passing NULL to non-pointer argument 1 of ‘std::basic_ostream<_CharT, _Traits>::__ostream_type& std::basic_ostream<_CharT, _Traits>::operator<<(long int) [with _CharT = char; _Traits = std::char_traits<char>; std::basic_ostream<_CharT, _Traits>::__ostream_type = std::basic_ostream<char>]’ [-Werror=conversion-null]
              << "\tvalues were '" << (x) << "' and '" << (y) << "'\n"), \
                                                            ^
/home/scott/Repositories/oiio/src/libOpenImageIO/imagespec_test.cpp:227:5: note: in expansion of macro ‘OIIO_CHECK_NE’
     OIIO_CHECK_NE    (spec.find_attribute("Foo"), NULL);
     ^
[...]

PR #1104, opened 2015-03-28
Fix unit test methodology regarding NULL pointers

lgritz:
Use OIIO_CHECK_ASSERT(foo != NULL), not OIIO_CHECK_EQUAL(foo,NULL)
because the latter can try to print the NULL, which on some platforms
could be a problem depending on exactly how NULL is defined.


lgritz:
Fixes #1102

PR #1106, opened 2015-03-29
Silenced a warning about non-initialized values.

propersquid:
Compiling was failing on a few non-initialized values. Also, Atom strips excess whitespace automatically, so the whitespace commits are due to that.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -262,7 +262,7 @@ hashrand (int x, int y, int z, int c, int seed)
 
 
 // Return a hash-based normal-distributed pseudorandom value.
-// We use the Marsaglia polar method, and hashrand to 
+// We use the Marsaglia polar method, and hashrand to
 OIIO_FORCEINLINE float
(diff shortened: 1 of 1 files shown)


src/libOpenImageIO/imagebufalgo_draw.cpp
```diff
@@ -298,7 +298,7 @@ noise_uniform_ (ImageBuf &dst, float min, float max, bool mono,
     // Serial case
     for (ImageBuf::Iterator<T> p (dst, roi);  !p.done();  ++p) {
         int x = p.x(), y = p.y(), z = p.z();
-        float n;
+        float n = 0.0;
         for (int c = roi.chbegin;  c < roi.chend;  ++c) {
             if (c == 0 || !mono)
```
lgritz: Ha, I just noticed, the 

    if (c == 0 || !mono)

should actually be

    if (c == roi.chbegin || !mono)

I will fix this when I merge your changes.

lgritz:
LGTM, thanks for the fix. I'll take care of the other bug in the vicinity when I merge.

lgritz:
Merged!

PR #1108, opened 2015-03-31
Fix occasional NaNs in alpha channel with 3 channel images.

brechtvl:
We found an issue on Windows where the SIMD texture filtering code would sometimes generate NaNs in the alpha channel for 3 channel images.

What happens is that when doing an SIMD load of 4 floats, the 4th alpha channel value uses uninitialized memory. Note the pixel buffer is made big enough in `ImageCacheTile::memsize_needed()` to avoid reading past the end of the array so that's fine, the problem is uninitialized memory.

When the image size is not a multiple of the tile size, then tiles near the edge contain uninitialized values. Sometimes these are NaN. I don't know if the fact that these are uninitialized is a bug, if so then that should be fixed instead, but I'm not sure where exactly.

Here the fix is to use `blend0()` and `blend0not()` rather than multiplication with zero which doesn't filter out NaNs.

If you want to reliably reproduce this bug, you can apply this patch:
https://github.com/brechtvl/oiio/commit/51119ae23144b688d261f71303bd23dd960aa1f2

And run `testtex` with this image file:
https://github.com/brechtvl/oiio/raw/51119ae23144b688d261f71303bd23dd960aa1f2/green_exr.tex

### DIFF
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -75,12 +75,12 @@ simd::float4 uchar2float4 (const unsigned char *c) {
 }
 
 
-static const OIIO_SIMD4_ALIGN float channel_masks[5][4] = {
-    { 0.0f, 0.0f, 0.0f, 0.0f },
-    { 1.0f, 0.0f, 0.0f, 0.0f },
-    { 1.0f, 1.0f, 0.0f, 0.0f },
-    { 1.0f, 1.0f, 1.0f, 0.0f },
-    { 1.0f, 1.0f, 1.0f, 1.0f }
+static const OIIO_SIMD4_ALIGN mask4 channel_masks[5] = {
+    mask4(false, false, false, false),
+    mask4(true,  false, false, false),
+    mask4(true,  true,  false, false),
+    mask4(true,  true,  true,  false),
+    mask4(true,  true,  true,  true),
 };
 
 }  // end anonymous namespace
@@ -1678,11 +1678,11 @@ TextureSystemImpl::sample_closest (int nsamples, const float *s_,
         accum += weight * texel_simd;
     }
 
-    simd::float4 channel_mask = *(simd::float4 *)(channel_masks[actualchannels]);
-    accum *= channel_mask;
+    simd::mask4 channel_mask = channel_masks[actualchannels];
+    accum = blend0(accum, channel_mask);
     if (nonfill < 1.0f && nchannels_result > actualchannels && options.fill) {
         // Add the weighted fill color
-        accum += ((1.0f - nonfill) * options.fill) * (simd::float4::One() - channel_mask);
+        accum += blend0not(float4((1.0f - nonfill) * options.fill), channel_mask);
     }
 
     *accum_ = accum;
@@ -1953,11 +1953,11 @@ TextureSystemImpl::sample_bilinear (int nsamples, const float *s_,
         }
     }
 
-    simd::float4 channel_mask = *(simd::float4 *)(channel_masks[actualchannels]);
-    accum *= channel_mask;
+    simd::mask4 channel_mask = channel_masks[actualchannels];
+    accum = blend0(accum, channel_mask);
     if (use_fill) {
         // Add the weighted fill color
-        accum += ((1.0f - nonfill) * options.fill) * (simd::float4::One() - channel_mask);
+        accum += blend0not(float4((1.0f - nonfill) * options.fill), channel_mask);
     }
 
     *accum_ = accum;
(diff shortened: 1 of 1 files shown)


fpsunflower:
I'm probably missing something obvious, but when reading only 3 channels, don't we just overwrite alpha with 1 at the end?

brechtvl:
Yes, that's the code that this patch is modifying, where it fills in the alpha, usually with 1.

The problem is that it makes the assumption that `0.0 * everything = 0.0`, but that's not the case with NaNs.

lgritz:
There may be an even better solution. I was just the other day thinking about this (not NaNs specifically, but just about how we fetch into the padded area).

When I first SIMD-ized the texture code, I didn't think there was a performant way to read anything but 4 adjacent floats, so I added the padding and multiply by a 1.0/0.0 mask to get rid of the unused values. But you are quite right -- multiplying by 0 would not fix a NaN in uninitialized memory.

But anyway, I was thinking that I had subsequently added simd::float4::load (float*vals,int n), which is capable of reading 4, 3, 2, or 1 float (making the rest 0.0), for each case using what I think is the most efficient way to load them. I had not known about efficient ways to read 2 or 3 floats when I first wrote the texture code. Before committing this, we should test your method (on real benchmark scenes) versus the simd.load(buf,3) technique, to see which has higher performance.

lgritz:
The reason I was musing about this was that I want to SIMD-ize the ImageBufAlgo code, but had struggled with the idea that if somebody constructs the kind of ImageBuf that "wraps" app image memory, it would not necessarily have the proper padding at the end, and I would be reading off the end of the buffer. In contemplating this, I realized that the load(ptr,n) would work. And further, I wondered if load(ptr,n) would be more efficient than load(ptr) + the masking operation in the texture code, and if so, switching may provide a slight speedup to texture.

This was literally just a few days ago that it occurred to me, and I had not yet gotten a chance to try it out and run benchmarks, am in the middle of working in an entirely different area of the code and don't want to lose my momentum.

It would be very easy for you to try this alternative, Brecht, and then we can use whichever gives better performance.

Apologies if my mis-timing has caused you to do unnecessary work.

fpsunflower:
The other implementation to benchmark would be `memset` of the padded area to 0, which would only incur a cost when reading the tile but be "free" after that.

brechtvl:
I think this patch is at least as fast as the current code in any case, as `_mm_and_ps()` and `_mm_andnot_ps()` are faster than `_mm_mul_ps()` (see [here](https://software.intel.com/sites/landingpage/IntrinsicsGuide/)). But if you want to try a better fix then that's fine with me of course.

fpsunflower:
That's true -- looking at your patch more carefully, its true that it should be just as fast if not faster than what we were doing before.

lgritz:
OK, if this is at least as fast and fixes the NaN problem, I will commit it, and then later go back and compare it to the partial-load approach (which I still need to explore to solve the ImageBuf padding problem.

lgritz:
LGTM. Will merge and backport to 1.5.

lgritz:
Merged.

lgritz:
I don't yet know why, but this change has been tracked down to be responsible for NaNs in the texture derivatives! Be careful. I'm looking into it.

brechtvl:
Oh, sorry about that, I missed that case, probably just need to do this:

```diff
diff --git a/src/libtexture/texturesys.cpp b/src/libtexture/texturesys.cpp
index c306638..af75723 100644
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -1847,8 +1847,8 @@ TextureSystemImpl::sample_bilinear (int nsamples, const float *s_,

     *accum_ = accum;
     if (daccumds_) {
-        *daccumds_ = daccumds * channel_mask;
-        *daccumdt_ = daccumdt * channel_mask;
+        *daccumds_ = blend0(daccumds, channel_mask);
+        *daccumdt_ = blend0(daccumdt, channel_mask);
     }
     return true;
 }
@@ -2234,8 +2234,8 @@ TextureSystemImpl::sample_bicubic (int nsamples, const float *s_,

     *accum_ = accum;
     if (daccumds_) {
-        *daccumds_ = daccumds * channel_mask;
-        *daccumdt_ = daccumdt * channel_mask;
+        *daccumds_ = blend0(daccumds, channel_mask);
+        *daccumdt_ = blend0(daccumdt, channel_mask);
     }
     return true;
 }
```

lgritz:
oh, yeah, I think so

lgritz:
I backed out your change so we could do a build.

I will test your additional change and re-commit your work if that makes it ok.

lgritz:
And while I'm at it, I will augment the OIIO testsuite with at least one test that verifies that texture derivatives are ok. :-)

lgritz:
The testsuite DID have a test for derivs, and it did write the wrong image. But idiff said it was a match! Because it tested for filing if A-B > threshold, rather than passing if A-B < threshold. Remember that all greater and less comparisons fail when NaN is involved, so the particular case of all incorrect pixels being NaN would not be recognized as a mismatched image by idiff. Wowza!

lgritz:
Merged the fix back into master, Brecht. Works fine with those extra lines for the derivatives.
I'm backporting to 1.5 as well, just running the final tests and then will merge it.

brechtvl:
Thanks! And sorry for the trouble.

PR #1109, opened 2015-04-01
Make IBA::compare() (and therefore idiff & oiiotool -diff) robust to NaN/Inf values

lgritz:
The way the comparisons were arranged, an image that had NaNs would not
notice that the NaN values were different between two images. Yikes!


fpsunflower:
LGTM

PR #1110, opened 2015-04-02
Another tweak IBA::compare() for nonfinite values

lgritz:
Make sure mismatched non-finite values are flagged as the biggest differences in the images, so you can find them right away.


fpsunflower:
LGTM

PR #1111, opened 2015-04-02
Squash testtex-induced NaNs

lgritz:
Working on some texture speedups, and image diffs are getting me down. Don't ask how, but this led me to finally find the place in testtex where the warp pattern ends up with NaN or Inf values in the texture coordinates. Changing a clamp does the trick, but then I need to update all the ref images that had the NaN/Inf values in them.


fpsunflower:
LGTM

lgritz:
Wait, it turns out that this transformation sucks -- it makes too much of the image a special case of the filter covering the whole image, and I'm worrying that this makes any benchmark timings not very representative. I'm going to need to change something.

lgritz:
Modified the transformation to something more to my liking. No need to re-review, it's the same idea, I just prefer the new layout of texture in the image.

PR #1114, opened 2015-04-09
Allow half and uint16 pixels directly in ImageCache tiles.

lgritz:
Previously we kept only uint8 or float tiles in the IC, everything else was automatically converted to internal storage as float, for texturing speed reasons (though at 2x cache space hit for 16-bit formats). This choice was based on benchmarks done a long time ago, and we suspected that after the recent SSE-based overhaul of the TS internals that the assumptions may no longer be valid. Indeed, they weren't.

This change allows half and uint16 files to store in their native format within ImageCache tiles, thus cutting their cache space pressure by half.

Also modification of all the places in TextureSystem where it assumed that IC tile pixels would be onlu uint8 or float (added cases for half and uint16).  By using SIMD conversion of those formats to float on-the-fly during texture filtering, we achieve uint16 speed identical to uint8 speed, half speed identical to uint8 speed when "f16c" CPU capability support is selected at build time, though slightly slower when no f16c instructions are available. In practice, the no-f16c path is 10% slower in texture math benchmarks, but this is not noticeable in real-world rendering benchmarks, more than made up for by improved chances of desired tiles being in cache.

There are a variety of supporting changes mixed in here:

* platform.h namespace safety improvements, and also addition of a runtime way to check certain CPU capabilities. (In the end, I didn't keep the texture-side code I was working on that used this change, but I would like to keep the platform-side changes, they may come in handy and I've already done the work.)

* simd.h additions of constructors and load methods to initialize int4 and float4 from pointers to short, unsigned short, char, and unsigned char.

* simd.h additions of ctr/load methods to initialize float4 from half values.

* Change top-level build scripts to allow the USE_SIMD=... to be a comma-separated list of capabilities to enable, rather than just a single level. (Example: USE_SIMD=avx,f16c).


cmstein:
LGTM

PR #1118, opened 2015-04-13
FIX windows build and few minor cmake configure improvements

ejerome:
Hello,
Thanks for sharing this great project.
I managed to build your project under windows 8.1 within MSVC11 but without your binaries you provided from the wiki (it cause to failed the build).
I rebuilt most of the images projects dependencies under MSVC11 x64 and use your THIRD_PARTY_TOOLS_HOME and append some minor things to make build your project.
3rdParty folder structure like this : 
+---oiio
|   ---3rdParty
|       ---windows
|           +---glew-1.5.1
|           |   +---include
|           |   ---lib
|           +---gtest-1.3.0
|           |   +---include
|           |   ---lib
|           +---ilmbase-2.2.0
|           |   +---include
|           |   +---lib
|           +---jasper-1.900.1
|           |   +---include
|           |   ---lib
|           +---jpeg-8d
|           |   +---include
|           |   +---lib
|           +---libpng-1.7.0.b54
|           |   +---bin
|           |   +---include
|           |   +---lib
|           +---LibRaw-0.17.0-Alpha1
|           |   +---include
|           |   +---lib
|           +---openexr-2.2.0
|           |   +---include
|           |   +---lib
|           |   +---share
|           +---openjpeg-1.3
|           |   +---include
|           |   ---lib
|           +---tbb-21_20090511oss
|           |   ---lib
|           +---tiff-3.8.2
|           |   +---include
|           |   +---lib
|           ---zlib-1.2.8
|               +---bin
|               +---include
|               +---lib
|               +---share

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,7 @@ endif ()
 
 
 if (MSVC)
+    add_definitions (-DNOMINMAX)
     add_definitions (-D_CRT_SECURE_NO_DEPRECATE)
     add_definitions (-D_CRT_SECURE_NO_WARNINGS)
     add_definitions (-D_CRT_NONSTDC_NO_WARNINGS)
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -1,6 +1,10 @@
 ###########################################################################
 # Find libraries
 
+if(WIN32 AND NOT THIRD_PARTY_TOOLS_HOME)
+    message(STATUS "You can set THIRD_PARTY_TOOLS_HOME to try to auto find external dependencies.")
+endif()
+
 setup_path (THIRD_PARTY_TOOLS_HOME 
 #            "${PROJECT_SOURCE_DIR}/../external/dist/${platform}"
             "unknown"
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ endif ()
 
 
 if (MSVC)
+    add_definitions (-DNOMINMAX)
```
lgritz: I'm wary of this change. It will surely protect a build of OIIO itself, but I fear it will mask errors in our public headers that will bite users and developers who use our headers in other projects that don't provide this definition.

I think we should NOT define this project-wide, but rather hunt down and fix all of our individual header files where this needs to be set.

src/cmake/externalpackages.cmake
```diff
@@ -9,17 +13,26 @@ setup_path (THIRD_PARTY_TOOLS_HOME
...
-    file (GLOB _external_dir_lockfiles "${THIRD_PARTY_TOOLS_HOME}/*.d")
+    file (GLOB _external_dir_lockfiles "${THIRD_PARTY_TOOLS_HOME}/*")
     foreach (_dir_lockfile ${_external_dir_lockfiles})
         # Grab the tool directory_name.d
         get_filename_component (_ext_dirname ${_dir_lockfile} NAME)
         # Strip off the .d extension
         string (REGEX REPLACE "\\.d$" "" _ext_dirname ${_ext_dirname})
-        set (CMAKE_INCLUDE_PATH "${THIRD_PARTY_TOOLS_HOME}/include/${_ext_dirname}" ${CMAKE_INCLUDE_PATH})
-        set (CMAKE_LIBRARY_PATH "${THIRD_PARTY_TOOLS_HOME}/lib/${_ext_dirname}" ${CMAKE_LIBRARY_PATH})
+		if(EXISTS "${THIRD_PARTY_TOOLS_HOME}/include/${_ext_dirname}")
```
lgritz: Strange indentation?

src/cmake/modules/FindOpenColorIO.cmake
```diff
@@ -45,7 +44,18 @@ MACRO(FindOpenColorIO)
         /sw/lib
         /opt/local/lib
         DOC "The OCIO library")
-
+    option(OCIO_STATIC OFF "Use STATIC lib mean use specific config for use")
```
lgritz: I'm sure this is an ESL issue, but I'm not sure I understand the explanation string here?

lgritz:
Thanks much for the contribution! This looks pretty good to me, other than the couple of comments.

I'll need to test this on a few platforms I have access to, to be sure nothing has broken as a result.

I don't have access to Windows, so I'm hoping that somebody else on Windows can test this patch out and verify that it works for them. (Not that I don't trust that it works for you... but sometimes it's easy to make the build scripts work for one person's quirky environment and find out too late it is horribly broken for other people. It's helpful to verify that it works for a second person with an independently set up work environment.)

CMakeLists.txt
ejerome: OK. I got errors from: oiio/src/include/OpenImageIO/fmath.h
Due to all use of std::numeric_limits<float>::min() and std::numeric_limits<float>::max()
The error is preceded by : warning C4003: nombre de paramètres réels insuffisants pour la macro 'max'
It caused by the windows.h include which define min and max macro : http://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c
I saw a workaround here:  oiio/src/include/OpenImageIO/platform.h
but it seem to be not enough...

src/cmake/externalpackages.cmake
ejerome: Gloups, sorry about that, you're right.

src/cmake/modules/FindOpenColorIO.cmake
ejerome: In order to deal with dll import/export OCIO define OpenColorIO_EXPORTS which define macros for import or export but which should be empty in case OCIO is build as static lib. To avoid windows error when building OIIO with static OCIO, I defined OpenColorIO_STATIC preprocessor (pull request : https://github.com/imageworks/OpenColorIO/pull/392).
I just realized that the one for tinyxml should be fixed too... sorry.
Should I correct according to all your comments and commit again (and redo the pull request)? Or it's not necessary? Thanks.

ejerome:
Definitively agree with you :)
I'm sure it is very easy to build your project and test it with another windows machine contributor :)
I was on Windows 8.1 with MSVC11 (visual 2012) x64 mode.

m-mitchell:
Hi, I'm on Win7 x64, MSVC11. I'd be happy to test out the patch; do you think I could get a copy of the dependencies you built?

ejerome:
Hello,

Thanks.
I just send you an email with my built 3rdParty packages.
Let us know.

thanks

m-mitchell:
Hi again,

With ejerome's third-party packages & his patch, I was able to compile the OIIO core. :+1: Thanks Jerome! Some of the plugins still don't build correctly but having a working core out of the box is a much nicer jumping-off point for Windows. :smile:

ejerome:
Cool :)
Yes I didn't build all MSVC 3rdParty for all plugins (only for libjpg, libpng, libtiff, openEXR and libRaw).

CMakeLists.txt
roelandschoukens: I do most of my development on Windows, and disabling `NOMINMAX` is more or less required if you include `Windows.h` in C++. Any sane build configuration will define either this macro, or `WIN32_LEAN_AND_MEAN`. I think it is fair to expect users of the header files to ensure `min` and `max` are not defined as macros.

dnjiang:
Hi ejerome,

I am looking for the updated pre-built external libs for Windows. Do you have a link somewhere? Please see the new issue I just raised: #1532

Thanks,
Dan

lgritz:
I'm embarrassed to say that although I seem to have approved this change long ago, I never merged it (I think I was waiting on a more definitive "yes, this is the right thing" from other Windows users, and the conversation petered out before I felt like I'd heard that clearly). Now I fear that the cmake files have changed so much in the mean time that there are substantial merge conflicts, and for that matter I don't know if the basic issues have been addressed by other patches along the way.

So I'm just going to apologize, close this PR, and reiterate that we always welcome new or updated PRs that improve the Windows build experience.

PR #1120, opened 2015-04-15
oiiotool --autocc

lgritz:
This set of changes adds an --autocc (auto color convert) option to oiiotool. When specified, it makes oiiotool:

* Examine any input and output filenames, looking to see if they contain substrings which are any of the known OpenColorIO color space names.

* If so, then it will assume that an input file contains data in that color space, and will automatically convert it to a scene-linear space upon input. If there is no such hint in the filename, it will then examine any "oiio:ColorSpace" metadata found in the file and do the conversion based upon that (this metadata is set by any ImageInput that can discern the color space of the input data, for example, it knows that JPEG data is always sRGB).

* Similarly, when requesting output (-o), it will interpret any color space names in the filename to be a request to convert to that color space as part of the output process. If no hint is found in the filename, but it appears to be a JPEG file (ending in .jpg or .jpeg), it will make sure the colors are converted to sRGB, as required by the JPEG standard.

* If a color space name is found embedded in the requested output filename and OCIO has a concrete idea of a specific pixel data type that is appropriate for the color space, it will also be used as a hint for a kind of automatic "-d" selecting output data format (e.g., color space "vd16" will try to output as uint16 pixels). 

Here's an example of how this all comes together. Let's say you have a DPX file in 10 bit log format (let's say, color space "lg10"), and you want to use oiiotool to brighten it by 10% in the R,G,B channels (but of course, 10% in a linear color space), and then save it as a TIFF in "vd8" color space (by our in-house convention, this is an video display color space, akin to sRGB, assumed to be 8 bits per channel). The old, fully explicit way is, assuming we follow our naming conventions rigorously:

    oiiotool in_lg10.dpx -colorconvert lg10 lnf -mulc 1.1,1.1,1.1,1.0 -colorconvert lnf vd8 -d uint8 -o out_vd8.tif

now, assuming you have the right color space names defined in your OCIO configuration, you can do the same operation in the more compact:

    oiiotool --autocc in_lg10.dpx -mulc 1.1,1.1,1.1,1.0 -o out_vd8.tif

So, everything will be converted to a canonical linear space upon input, any math oiiotool performs will then be correctly linear, and then convert to appropriate spaces for output, all determined by file naming conventions and the help of OCIO.

For anyone who cares, this roughly emulates the behavior of an old in-house image processing tool which has at this point been replaced by oiiotool.

One last thing... all this heavy use of color conversion reminded me that the underlying ImageBufAlgo::colorconvert() is not multithreaded, as is the rest of IBA. So I fixed that while I was at it. This speeds up oiiotool color conversion on my workstation by 8x!

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -127,6 +127,8 @@ Fixes, minor enhancements, and performance improvements:
     * Bug fix: input file data format didn't always end up in the output.
       (1.6.3)
     * --channels bugs were fixed when dealing with "deep" images. (1.6.3)
+    * All the color space conversion operations run much faster now,
+      since the underlying IBA::colorconvert() has been parallelized. (1.6.3)
   * maketx & TextureSystem:
     * TextureSystem/IC now directly stores uint16 and half pixel data in
       the cache rather than converting internally to float for tile storage,
@@ -152,6 +154,8 @@ Fixes, minor enhancements, and performance improvements:
       image differences when the pixels that differed had NaN or NaN or
       Inf values! Now it is right. #1109 (1.6.3/1.5.13)
     * channels() bugs were fixed when dealing with "deep" images. (1.6.3)
+    * colorconvert() has been parallelized, and thus on most systems will
+      now run much faster. (1.6.3)
  * OpenEXR:
     * Improved handling of density and aspect ratio. #1042 (1.6.0)
     * Fix read_deep_tiles() error when not starting at the image origin.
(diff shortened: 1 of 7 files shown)


lgritz:
I updated this after discovering an OCIO call I didn't know about, that lets me ask it what data format it thinks is right for a named color space in a non-SPI-specific way. Thus, I will enable the color->bitdepth inference for everybody, not just SPI. 

(I'll edit the description for this PR to reflect the new intended behavior.)

cmstein:
LGTM

PR #1121, opened 2015-04-15
UTF-8 support for render_text algo

sopvop:
UTF8 decoder is from [here](http://bjoern.hoehrmann.de/utf-8/decoder/dfa/), It is a simple and fast decoder, and will take some invalid utf-8 input. It should not be a problem, since FreeType will most likely not find such glyph. And it does not affect program stability.

It is still up to user to supply font which supports all the needed codepoints.


lgritz:
This basically LGTM, though I wonder if maybe utf8_to_unicode() should simply be in strutil.h and the implementation of decode be put in strutil.cpp, rather than making the new file utf8_dfa.h just for that?

sopvop:
Sure, I'll do that.
I just wanted to put code copyrighted by decode author into separate file.
Also I think decode function is small enough and is called for every char so it should probably be kept as `inline` in header.

sopvop:
Since utf8_to_unicode is public API now, so I decided to make it a bit simpler. OIIO is not about text encodings :)

Maybe render_text should be overloaded to take std::u32string, if available in complier?

But that would require detection of c++11 features in current compilers and std::* libraries, which is painful and error prone, and would be too much work for such a small feature.

lgritz:
If you #include "platform.h", there will be a macro OIIO_CPLUSPLUS11, which it's ok to conditionally compile with.

I like what you've done with the code. I'm not at all worried about the function not being inlined. In the context of any image-handling app that uses OIIO, it seems very unlikely that the function call overhead of the few times it needs to convert unicode characters is ever going to be the performance bottleneck.

I'l test this out andl merge it (with perhaps some extremely minor formatting tweaks).

Do you need this backported to 1.5, or is master sufficient for you?

sopvop:
Unfortunately not all compilers claiming to be c++11 support new strings, so that macro is not enough.
And master is enough. Thank you.

lgritz:
Merged, with some minor formatting changes, and addition of various docs updates.

PR #1126, opened 2015-04-22
Make TextureSystem::create(shared=true) truly create a shared TS.

lgritz:
It had been creating a separate TS for every create call, but with a
shared ImageCache (if requested).  That's not good enough. If two spots
in an app each request a shared TS, They probably want it to actually be
the same one, including any attributes already set in it.


lgritz:
Updated with changes to avoid destruction order fiasco with respect to the ImageCache that's referred to by the TextureSystem. Now we leak a single TS, which is small anyway, and it only happens when the app exits. There is no consequence to this other than valgrind warnings about a leak that doesn't really matter.

PR #1136, opened 2015-05-05
Tweaks to TypeDesc to allow arraylen -n to mean unsized, with default length n.

lgritz:
(Submitted on behalf of Daniel Heckenberg)

Instead of arraylen == -1 meaning "unsized" and length < -1 having no
meaning, -n means "unsized with default length n".

This is helpful in supporting OSL arrays with runtime-settable lengths. It doesn't
have any immediate impact on OIIO per se, which has no internal uses for the
"array of indeterminite length" notation.

### DIFF
--- a/src/include/OpenImageIO/typedesc.h
+++ b/src/include/OpenImageIO/typedesc.h
@@ -54,6 +54,7 @@
 
 #include "export.h"
 #include "oiioversion.h"
+#include "dassert.h"
 
 
 OIIO_NAMESPACE_ENTER
@@ -162,12 +163,26 @@ struct OIIO_API TypeDesc {
     /// Return the number of elements: 1 if not an array, or the array
     /// length.
     size_t numelements () const {
+        DASSERT_MSG (arraylen >= 0, "Called numelements() on TypeDesc "
+                     "of array with unspecified length (%d)", arraylen);
         return (arraylen >= 1 ? arraylen : 1);
     }
+
+    /// Does this TypeDesc describe an array?
+    bool is_array () const { return (arraylen != 0); }
+
+    /// Does this TypeDesc describe an array, but whose length is not
+    /// specified?
+    bool is_unsized_array () const { return (arraylen < 0); }
+
+    /// Does this TypeDesc describe an array, whose length is specified?
+    bool is_sized_array () const { return (arraylen > 0); }
 
     /// Return the size, in bytes, of this type.
     ///
     size_t size () const {
+        DASSERT_MSG (arraylen >= 0, "Called size() on TypeDesc "
+                     "of array with unspecified length (%d)", arraylen);
         size_t a = (size_t) (arraylen > 0 ? arraylen : 1);
         if (sizeof(size_t) > sizeof(int)) {
             // size_t has plenty of room for this multiplication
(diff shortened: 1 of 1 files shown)


src/include/OpenImageIO/typedesc.h
```diff
@@ -241,14 +241,14 @@ struct OIIO_API TypeDesc {
     /// inequality is differing vector semantics.
     friend bool equivalent (const TypeDesc &a, const TypeDesc &b) {
         return a.basetype == b.basetype && a.aggregate == b.aggregate &&
-               (a.arraylen == b.arraylen || (a.arraylen == -1 && b.arraylen > 0)
-                                         || (a.arraylen > 0   && b.arraylen == -1));
+               (a.arraylen == b.arraylen || (a.arraylen < 0 && b.arraylen > 0)
```
fpsunflower: If I am reading this correctly it means that something like `float[3]` is equivalent to a `float[]` which happens to have size 4? However a `float[3]` is not equivalent to a `float[4]`?

Don't you want to `abs(a.arraylen) == abs(b.arraylen)` instead?

(I see the previous code also had this behavior, so maybe I'm not understanding the intent behind this method)

fpsunflower:
Actually reading the description again -- it sounds like `arraylen` for the unsized case only captures the notion of "default" length, which kind of explains the semantics above.

This has the potential to be confusing, since "default" can mean different things to different users.

Without having looked at the OSL side review yet -- the previous implemention of `TypeDesc` fit my mental model better of `float[]` just being effectively a "pointer" type of unknown length. My understanding was that you want `TypeDesc` to be a lightweight way to describe C types, and C types don't really have the notion of "default" length (unless you intend this to become some way of effectively representing "slices" - ala "`array_view`").

Are we just trying to shoe-horn a length inside this type that really belongs on the OSL side?

Or have I mis-understood the intent?

lgritz:
OK, so previously, arraylen<0 already meant "unspecified length", i.e. float[].  But in practice, we always used -1 for that value and blithely ignored the possibility that it might take on a value <=-2.

I believe the intent of this patch is that all those places we tested specifically for -1, now we test for < 0. In some sense this is safer (having it set to -2 would probably have made for some very strange behavior). This patch makes it so that no values of the arraylen field will lead to undefined behavior.

On the OIIO side, that's all it does.

But in addition, this frees the particular negative value to have some meaning to an app (though not to TypeDesc internals), and on the OSL side, it seems convenient for it to mean "the number of initializers we were given for this unspecified-length array."

I think there are two questions to ask:

1. On the OIIO side, should we take all negative values to indicate "array of unspecified length", thus leaving no values for arraylen whose meaning is undefined. I believe this is an improvement and we should accept it.

2. On the OSL side, should we use this more robust behavior surrounding the range of negative arraylen values to encode the notion of "array of unspecified length, but with THIS many initializers" in the TypeDesc itself (with OIIO per se unaware that we are using particular values for particular meanings; OIIO itself treating all negative value lengths as having equivalent meaning, and comparing as equal). Or, is it sloppy to overload this meaning on the TypeDesc::arraylen, and we should instead make a new field in the Symbol (? or the TypeSpec?) that explicitly encodes the number of initializers that were found? That one is more shaky, and I would certainly support having an explicit field in the appropriate place rather than try to shoehorn it into TypeDesc.

lgritz:
I just pushed a follow-up delta to this PR that clarifies some comments, and adds is_array(), is_sized_array(), and is_unsized_array() methods, using those in several places rather than explicitly testing the arraylen field definitely makes the intent clearer.

So I think that is an improvement.

But in doing so, I think I've uncovered some conceptual cracks. Chris, as usual, may be on to something.

Let's take operator==. Should two arrays of indefinite length be considered equal types? I think, probably, yes. When we used -1 only to signify unspecified length, the == test worked in an intuitive way. But now we could have two arrays of indefinite length (one with -2, and one with -5), and it will fail the equality test. Should it fail? I don't know.

I don't know how to describe, in the comments to operator==, why arraylen -2 and arraylen -5 (which we previously said were semantically both meaning the same thing, "unspecified length array") should not be considered equal. And to explain it in terms of numbers of initializers or default length would seem likely to provoke a "huh?" in the reader rather than a sense of understanding. Similarly, it's not immediately clear why asking for the numelements() should give the abs(arraylen) as its result, without further explanation that the number does encode something in particular. So that kind of trashes my earlier argument that this keeps prior TypeDesc semantics and merely eliminates the possibility of values leading to undefined behaviors.

One way out of this is to further explain, on the OIIO side, that the particular negative arraylen value has a meaning after all: that although the length is unspecified, if the array ever needs to be allocated (e.g., it *needs* length, and nobody else has yet come along to change the unspecified length to a fixed length), that there is a notion of default length to use in that case. Chris might say "that's not a C concept", but I don't think that's a strong argument, considering that we also have aggregate and vecsemantics fields that encode things about the type that doesn't correspond to anything in C. The "not C" angle is not a showstopper for me.

I will observe that the use of particular negative values to mean a default length of the unspecified array leaves a conceptual hole: there is NO way to say that the array has unspecified length, and also has no initializers/default length, if you catch my drift. I'm not sure if this is important -- perhaps not, since the alternative (status quo) also doesn't give you any notion at all about default lengths.

The other solution is to admit that Chris is right, abandon (or at least delay) this patch while we search for an OSL-only solution for how OSL should encode the notion that an array of unspecified length may have a number of initializers. (Similarly to how OSL encodes that a type is a closure, or designates a struct, completely on the OSL side without burdening the compact and simple TypeDesc.)

I'm on the fence. An OSL-side solution is probably cleaner, but on the other hand I don't necessarily object to the idea of having the notion of "unspecified length, but if you are forced to allocate space for it and need to fix a length, here's the size to use."

lgritz:
I think what I'll do is noodle with Daniel's OSL-side patch, see if it can be made to work cleanly and more readably without any changes to TypeDesc, see how far I get. If that pans out, then we can abandon or alter the TypeDesc PR. If not, we have a stronger argument for it.

fpsunflower:
Yep - that summarizes where I was getting confused.

There are really two bits of information:
  * a type can be an array or not (note: we currently distinguish between `float` and `float[1]`)
  * a type can be tagged as being a variable length array (encoded as negative `arraylen` in this patch)

The question is what "variable length array" really means. It does seem convenient on the OSL side, but its meaningless (I think?) outside of this context.

So maybe what we need to do is *remove* all notions of "unsized" or "varlen" arrays from OIIO's `TypeDesc` and handle the "varlen" concept purely on the OSL side where it does have a meaning. This way we can think of `TypeDesc` purely as something that describes concrete types that can exist in memory. As you said earlier, OIIO does not have any use for the "varlen" concept.

fpsunflower:
And just to make a case for the currently proposed change:

We could clarify that negative array lengths mean: "this type is a *view* onto externally allocated memory". The absolute value then tells you how many elements it is safe to access. I think this could also be useful on the OSL side for "varlen" arrays because it would reflect that although they are very much like arrays, their length is "controlled" externally (by the default value, the instance value or some other connection). I could also see it being useful in other places that use `TypeDesc` but that is more speculative.

Positive lengths would imply that this type is both an array and *owns* its own memory (like how you can declare an array of 3 floats on the stack in C).

If we took this stance, we might want to include the `arraylen==0` case to mean: "this is a view onto an empty `float` array" (kind of how you can have a `string_view` of length 0).

This would imply either additional bits, or making `arraylen==1` be used for regular types. In other words (`float` and `float[1]` would become the same thing).

lgritz:
I definitely don't think 'float' and 'float[1]' are the same thing at all. I'm firm on that. Like in C, you can't pass a float value to a function parameter whose type was declared as float[1], or vice versa. They are definitely not the same.

I understand what you're saying about views, but I never though of TypeDesc as describing that at all, it seems like a much bigger semantic change. Also note that the emerging C++ concept of views (string_view, and soon array_view) do not have any notion of unspecified lengths; they always have a definite length.

TypeDesc is meant to encode a type, not the allocation status of a variable. That is subtly different.  Consider the type of a function parameter. It may be "float[]", meaning it's a float array. But until the function is called and a variable is bound, you don't know its allocation status. But the type of the function parameter is not indeterminate, or shifting around like Shrödinger's Cat; it's an array, but of unspecified length. I want to be able to capture that concept.

One solution is to make a separate field for "arrayness": scalar, fixed length array, unspecified length array. And a separate (unsigned) length. For a scalar, the only length that makes sense is 1. For a fixed length array, the length is the length. For an array of unspecified length, a length of 0 could mean the usual float[] (no information on the length) while a nonzero length could mean that the array represents a view to a known safe length. Though, that's still not exactly what we want on the OSL side -- it's not a view, it's just the idea that the default initializer had a certain number of values, so if it's not overridden with a different instance value (an array of fixed length), that's the size that we'll eventually allocate. That does seem to argue for trying to capture it all on the OSL side.

fpsunflower:
That all makes sense.To me it suggests we should just remove the current support for "unsized" (-1) case from `TypeDesc` and make it `unsized`.

Hopefully not too many places on the OSL side rely on that case. My guess is they are mostly addressed by the parallel review. Its just a matter of moving the concept from `TypeDesc` to `TypeSpec`.

Maybe on the OSL side we want to always keep `TypeDesc` of the raw type (no array) and always measure the array view something in `TypeSpec` (at least until the concrete length is known for sure).

lgritz:
In https://github.com/imageworks/OpenShadingLanguage/pull/497 , I proposed an alternate patch on the OSL side that does not need any of these changes. I therefore recommend that we abandon this patch, it's unnecessary.

dheckenberg:
Thanks for the very detailed comments.  As observed, I took the path of greatest convenience and least code change for the patch, but I agree there are conceptual issues which would be best to solve cleanly.

The concept which the PR encodes is to describe an instance of a sized array which is allowed to change its size, which maps well to the "varlen" terminology.  The alternative being discussed is that TypeDesc doesn't so much apply to instances as definitions.

dheckenberg:
I agree: let's abandon this change and make it self-contained in OSL.

lgritz:
Rather than fully abandon, I decided that I still liked a few of the helper methods I had added, and DASSERT like in the corresponding OSL review, both to improve readability of the code and reduce the chances of subtle bugs in the use of this class. Updated the PR to have just these changes and no others.

fpsunflower:
LGTM

dheckenberg:
Looks good -- as noted in the OSL review, public access to the arraylen member leaves the possibility of bugs in the unsized case, but this is definitely an improvement.

PR #1140, opened 2015-05-08
Misc WIN32 and VS2010 fixes

darkhorse64:
CHANGES:
* imagebufalgo_util.h: export type_merge symbol
* platform.h : declare __cpuidex function
* filesystem.cpp : include `<windows.h>` after `"OpenImageIO/platform.h"`. Otherwise, NOMINMAX gets defined too late.
* py_paramvalue.cpp : make sure code uses make_tuple from boost::python and not make_tuple
from std::tr1. VS2010 gets confused with the current code.

Except for the second, all changes should be backported to RB 1.5


lgritz:
LGTM, thanks for the fixes, I will merge these into both branches shortly.

lgritz:
Backported to 1.5 as well.

PR #1143, opened 2015-05-11
Remove old implementations of ustring

fpsunflower:
Now that our custom implementation of ustring tables has been working for a while, it should be ok to remove the old one.


lgritz:
LGTM

PR #1146, opened 2015-05-14
adds a --pixelaspect %f to oiiotool that intelligently scales the image in X or Y to match the given pixel aspect ratio

mjmvisser:
As described in issue #113. I changed the name to pixelaspect to reflect more accurately what the action does.

e.g., given a 16x32 image with a PAR of 2.0, --pixelaspect 1.0 will result in a 32x32 image with a PAR of 1.0.

This is useful when converting non-square-pixel images to formats that do not support non-square pixels.

Tested with non-square pixel images from here:
http://www.schaik.com/pngsuite/pngsuite_cnk_png.html

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -2572,6 +2572,97 @@ action_fit (int argc, const char *argv[])
 
 
 
+static int
+action_pixelaspect (int argc, const char *argv[])
+{
+    if (ot.postpone_callback (1, action_fit, argc, argv))
+        return 0;
+    Timer timer (ot.enable_function_timing);
+    bool old_enable_function_timing = ot.enable_function_timing;
+    ot.enable_function_timing = false;
+    string_view command = ot.express (argv[0]);
+
+    float new_paspect = Strutil::from_string<float> (argv[1]);
+    if (new_paspect <= 0.0f) {
+        ot.error (command, Strutil::format ("Invalid pixel aspect ratio '%g'", new_paspect));
+        return 0;
+    }
+
+    // Examine the top of stack
+    ImageRecRef A = ot.top();
+    ot.read ();
+    const ImageSpec *Aspec = A->spec(0,0);
+
+    // Get the current pixel aspect ratio
+    float paspect = Aspec->get_float_attribute ("PixelAspectRatio", 1.0);
+    if (paspect <= 0.0f) {
+        ot.error (command, Strutil::format ("Invalid pixel aspect ratio '%g' in source", paspect));
+        return 0;
+    }
+
+    // Get the current (if any) XResolution/YResolution attributes
+    float xres = Aspec->get_float_attribute( "XResolution", 0.0);
+    float yres = Aspec->get_float_attribute( "YResolution", 0.0);
+
+    // Compute scaling factors and use action_resize to do the heavy lifting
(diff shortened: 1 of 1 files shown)


src/oiiotool/oiiotool.cpp
```diff
@@ -2572,6 +2572,97 @@ action_fit (int argc, const char *argv[])
 
 
 
+static int
+action_pixelaspect (int argc, const char *argv[])
+{
+    if (ot.postpone_callback (1, action_fit, argc, argv))
```
lgritz: action_pixelaspect!  (not fit)

src/oiiotool/oiiotool.cpp
```diff
@@ -2572,6 +2572,97 @@ action_fit (int argc, const char *argv[])
...
+        return 0;
+    Timer timer (ot.enable_function_timing);
+    bool old_enable_function_timing = ot.enable_function_timing;
+    ot.enable_function_timing = false;
+    string_view command = ot.express (argv[0]);
+
+    float new_paspect = Strutil::from_string<float> (argv[1]);
```
lgritz: ot.exrpress (argv[1])

That's what expands any command line expressions.

lgritz:
Very close to LGTM, nothing I can't fix on my end as I merge. I'll make a few minor fixes to it and commit shortly.

lgritz:
I'm toying with the idea of renaming it --scalepixelaspect or something like that. I'm slightly worried that --pixelaspect will be confusing: does it rescale the image, or just change the metadata? The name should make it clear that it's not merely a synonym for --attrib PixelAspectRatio ...

You have any thoughts on this, Mark?

lgritz:
I'm ok with the name, though would consider renaming if you think of something better.

I'll merge tomorrow, plus my extremely minor mods, docs, and test case.

lgritz:
Mark, before I merge this, I want to check one more thing with you. This patch appears to always enlarging one dimension to conform to the requested aspect ratio. For example a 256x256 par=1.0 file, given --pixelaspect 2.0, will result in a 256x512 image, not a 128x256. This approach is nicely conservative -- you don't lose any detail that was in the original image. But some people may be surprised to find the file quite a bit bigger than before. I think that your approach is probably the superior one (never lose detail), but I wanted to be extra sure that's the intended behavior before I merge it.

mjmvisser:
Yeah, I went through similar variations on the name before going with --pixelaspect. I decided against using --scalepixelaspect because it doesn't make sense to scale a pixel aspect ratio. For a while it was --setpixelaspect, but like you mention, that implies (at least to me, anyway) setting the metadata without modifying the image. I suppose it could be --scaletopixelaspect, but that's pretty long for a command line arg. My compromise was to try and make the --help for it as concise as I could.

You're right, I went with the conservative approach intentionally to avoid losing detail. Maybe we could make that explicit in the --help?

     "--pixelaspect %@ %g", action_pixelaspect, NULL, "Scale up the image's width or height to match the given pixel aspect ratio (options: filter=%s)",

cheers

lgritz:
I'm happy with both the name and the behavior. I'll explain in the documentation about the conservative approach.

mjmvisser:
Awesome, thanks!

lgritz:
Merged into master, along with my very small mods, docs, and tests.

PR #1150, opened 2015-05-23
Win32 fixes

darkhorse64:
`rintf` is not part of Visual Studio 2010 libraries. The first commit provides a default implementation.

The second commit is more controversial. Currently, `psd` & `jpeg` plugins do not compile on Windows for master and RB1.5 because `jpeg` headers define stuff that conflicts with `<windows.h>`. The patch attempts to prevent this in a simple manner but it is quite ugly.

The real issue here is that `<windows.h>` is included with each and every OIIO file: IMHO, it should not because this header is only really needed when using `Win32 API` (timers, threads, locks, ...).

I think a better patch would be to restore `osdep.h` usage in a few selected places and to remove `<windows.h>` from `platform.h`. I can certainly do this if you will.

### DIFF
--- a/src/include/OpenImageIO/missing_math.h
+++ b/src/include/OpenImageIO/missing_math.h
@@ -166,6 +166,11 @@ cbrtf (float val) {
 }
 
 
+inline float
+rintf (float val) {
+    return val + copysignf(0.5f, val);
+}
+
 #elif _MSC_VER >= 1800 && __cplusplus <= 201103L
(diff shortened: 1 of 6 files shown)


lgritz:
LGTM. As soon as I verify that nothing breaks on other platforms, I'll merge this into master and RB-1.5.
Thanks for the patch.

lgritz:
Merged into both branches

PR #1151, opened 2015-05-26
reimplemented expressions to handle order-of-operations, parentheses and string concatenation

mjmvisser:
Hi Larry, I ended up biting the bullet and adding support for proper expressions. This change should be backwards-compatible with the previous parser, but now also supports ~~string quoting and concatenation (e.g. {"prefix_" + TOP.filename} or {prefix_ + TOP.filename}) and~~ brackets & order-of-operations (e.g. {TOP.width+2*(TOP.x+1)}. 

I couldn't see any way to check if a string_view exactly represented a number, so I added a new template function to strutil.h: `string_is<T>(string_view s) -> bool`. `string_is<int>()` is unused, but included for completeness.

I also updated the docs, beefed up the expr* tests and added another test for string concatenation (all pass on my side).

Inspiration came from this post: http://www.strchr.com/expression_evaluator

Feedback greatly appreciated!

cheers,
-Mark


lgritz:
Just as a side note, string concatenation could have already been accomplished as

    prefix_{TOP.filename}

or

    {expressionA}{expressionB}

src/include/OpenImageIO/strutil.h
```diff
@@ -209,6 +209,26 @@ std::string OIIO_API join (const std::vector<std::string> &seq,
...
 
+// Helper template to test if a string is a generic type
+template<typename T>
+inline bool string_is (string_view s) {
+    return false; // Generic: assume there is an explicit specialization
+}
+// Special case for int
+template <> inline bool string_is<int> (string_view s) {
+    char *endptr = 0;
+    strtol (s.c_str(), &endptr, 10);
```
lgritz: I think that in this case you want to use data(), not c_str().

So... because a string_view is not necessarily null-terminated, but c_str() must return a valid null-terminated C string, the way c_str() works is tricky. If the character immediately following the string_view region is a null, then c_str just returns data(), piece of cake. But if it doesn't -- that is, if the view is an interior section of another string -- then in order to safely return a NULL string that the caller doesn't have to free, it actually creates a ustring (which can safely reside in the ustring table for the duration of the process execution) and returns the ustring's c_str(). 

So the bottom line is that if you think you'll call this on the interior of strings frequently, c_str() will be expensive and wasteful, creating lots of "permanent" ustring table entries needlessly. Using data() will avoid this. The only down-side is that it may end up reading past the end of the view. But that should be ok, because it's still guaranteed to eventually be null-terminated.

testsuite/oiiotool/run.py
```diff
@@ -256,8 +256,8 @@
 # test expression substitution
 command += oiiotool ("tahoe-small.tif --pattern fill:top=0,0,0,0:bottom=0,0,1,1 " +
                      "{TOP.geom} {TOP.nchannels} -d uint8 -o exprgradient.tif")
-command += oiiotool ("tahoe-small.tif -cut {TOP.width-40}x{TOP.height-40}+{TOP.x+20}+{TOP.y+20} -d uint8 -o exprcropped.tif")
-
+command += oiiotool ("tahoe-small.tif -cut '{TOP.width-20* 2}x{TOP.height-40+(4*2- 2 ) /6-1}+{TOP.x+100.5-80.5 }+{TOP.y+20}' -d uint8 -o exprcropped.tif")
```
lgritz: There are a couple of funny places with spaces, like "(4*2- 2 )".  Can you confirm that those spaces are not significant?

testsuite/oiiotool/run.py
```diff
@@ -256,8 +256,8 @@
...
+command += oiiotool ("tahoe-small.tif -o '{\"exprstrcat\" + TOP.compression + \".tif\"}'")
```
lgritz: I understand that you are just testing string concatenation, but I think any sane person would find this more clear:

    oiiotool tahoe-small.tif -o exprstrcat{TOP.compression}.tif

I'm not necessarily arguing against '+' for string concatenation, but I am unclear when one would prefer that notation, and whether supporting it that adds any unnecessarily complexity to the expression implementation.

lgritz:
Outstanding! LGTM, except for the note about c_str() vs data() in the new string functions.

Very cool.

I always figured we'd go back and do something like this. I just started with the simplest possible implementation, not being entirely sure if anybody would find expressions useful enough to put any more work into it. Apparently they do!

testsuite/oiiotool/run.py
mjmvisser: That's intentional: testing that whitespace is insignificant. Should I add a comment?

testsuite/oiiotool/run.py
mjmvisser: Good point, I'll just remove string concatenation from the parser.

src/include/OpenImageIO/strutil.h
mjmvisser: Fixed!

mjmvisser:
Updated as suggested:
- removed string concatenation
- use data instead of c_str in string_is functions
- updated tests and docs accordingly

Thanks!

testsuite/oiiotool/run.py
lgritz: No, it's ok. I just wanted to be sure you were merely testing accepting it as input, and not that it's significant in the "required" sense.

src/oiiotool/oiiotool.h
```diff
@@ -171,13 +171,15 @@ class Oiiotool {
...
-    //     left-to-right (no operator precedence).
+    //   * arbitrary math using operators +, -, *, / and parentheses
+    //     (order of operations is respected).
     //   * IMG[n].metadata for the metadata of an image. The 'n' may be an
     //     image name, or an integer giving stack position (for example,
     //     "IMG[0]" is the top of the stack; also "TOP" is a synonym). The
     //     metadata can be any of the usual named metadata from the image's
     //     spec, such as "width", "ImageDescription", etc.
+    //   * string constants (space-delimited or double-quoted)
+    //   * string concatenation using the + operator
```
lgritz: You removed string concatenation with +, right? Probably should remove this line from the comment.

lgritz:
LGTM.

Only problem I see is one comment, but I can fix that when I do the merge.

mjmvisser:
Whoops, missed that one. Fixed.

lgritz:
Merged.

PR #1160, opened 2015-06-03
Simplify ninja instructions, and set ninja path customizations for SPI

lgritz:


lgritz:
If Chris likes this, I'll make the analogous change to OSL.

fpsunflower:
LGTM!

PR #1163, opened 2015-06-10
Fix trilinear mipmap texture lookup giving an invalid alpha fill

brechtvl:
When using trilinear mipmap texture lookup, alpha values of 2.0 are returned when using an RGB image without an alpha channel.

This happens because the `sampler` functions are implemented in such a way that the fill code assumes the `weight` values to add up to 1. This fix makes `texture_lookup_trilinear_mipmap()` apply the level weights in the same way as `texture_lookup()`, by multiplying it afterwards instead of passing it in the `weight` argument.

### DIFF
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -1264,16 +1264,16 @@ TextureSystemImpl::texture_lookup_trilinear_mipmap (TextureFile &texturefile,
         if (! levelweight[level])  // No contribution from this level, skip it
             continue;
         float4 r, drds, drdt;
-        weight[0] = levelweight[level];
         ok &= (this->*sampler) (1, sval, tval, miplevel[level],
                                 texturefile, thread_info, options,
                                 nchannels_result, actualchannels, weight,
                                 &r, dresultds ? &drds : NULL, dresultds ? &drdt : NULL);
         ++npointson;
-        *(simd::float4 *)(result) += r;
+        float lw = levelweight[level];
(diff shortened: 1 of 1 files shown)


brechtvl:
Also, it would be good to backport this fix to RB-1.5.

src/libtexture/texturesys.cpp
```diff
@@ -1264,16 +1264,16 @@ TextureSystemImpl::texture_lookup_trilinear_mipmap (TextureFile &texturefile,
...
             continue;
         float4 r, drds, drdt;
-        weight[0] = levelweight[level];
         ok &= (this->*sampler) (1, sval, tval, miplevel[level],
                                 texturefile, thread_info, options,
                                 nchannels_result, actualchannels, weight,
                                 &r, dresultds ? &drds : NULL, dresultds ? &drdt : NULL);
         ++npointson;
-        *(simd::float4 *)(result) += r;
+        float lw = levelweight[level];
```
lgritz: I think this will be slightly faster as

    float4 lw = levelweight[level];

mixing scalar float and vector float is more expensive than vector/vector. Since lw might get used 3 times (for the deriv case), it's better to store it as a float4 once.

I can make this fix on my end, just letting you know.

brechtvl:
Ok, sure.

lgritz:
Edited and committed. Also altered to use a temp accumulator in the same way it's used in the full anisotropic case. Merged into both main and RB-1.5.

PR #1164, opened 2015-06-12
Fix multithreaded build with make -jN no longer working.

brechtvl:
It's now always building single threaded after commit 735d43a.

GNU make requires ${MAKE} to be used to detect the child process is make, see the explanation for the `jobserver unavailable: using -j1` error message here:
https://www.gnu.org/software/make/manual/html_node/Error-Messages.html

The code duplication isn't great, but I don't know how to avoid it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,9 +247,6 @@ endif
 ifeq (${USE_NINJA},1)
 MY_CMAKE_FLAGS += -G Ninja
 BUILDSENTINEL := build.ninja
-RUN_BUILD := ${NINJA} ${MY_NINJA_FLAGS}
-else
-RUN_BUILD := ${MAKE} ${MY_MAKE_FLAGS}
 endif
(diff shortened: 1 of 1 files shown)


lgritz:
OK, sure.

This is actually how I had it first, then thought I was being clever by combining the clauses.

BTW, you should try ninja, it's really speedy.

brechtvl:
I've been using ninja for years with Blender, it's great.

Just thought this would be good to fix anyway.

lgritz:
Merged. Thanks for the tip.

I'll do the same thing on the OSL side, if you haven't already submitted a PR there as well.

brechtvl:
I don't have a pull request for OSL, so go ahead.

lgritz:
https://github.com/imageworks/OpenShadingLanguage/pull/528

Issue #1166, opened 2015-06-13
Ubuntu 14.04 build error due to old ffmpeg


wrosecrans:
Building oiio master on an Ubuntu LTS  (14.04) machine, I have libavcodec v54, but oiio requires functions that were introduced in v55.  Happily, the new functions are apparently just renamed old functions as part of an ffmpeg API cleanup.  Here is the build error:
/storage/dev/repos/oiio/src/ffmpeg.imageio/ffmpeginput.cpp: In member function ‘virtual bool OpenImageIO::v1_6::FFmpegInput::open(const string&, OpenImageIO::v1_6::ImageSpec&)’:
/storage/dev/repos/oiio/src/ffmpeg.imageio/ffmpeginput.cpp:231:30: error: ‘av_frame_alloc’ was not declared in this scope
     m_frame = av_frame_alloc();

Here is a stackoverflow question where somebody ran into something similar:

http://stackoverflow.com/questions/24057248/ffmpeg-undefined-references-to-av-frame-alloc

I don't know if this is the best way, but the following after the "extern" block that includes the ffmpeg headers seems to do it.  Iw as able to build with this added to ffmpeginput.cpp

#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(55,28,1)
#define av_frame_alloc  avcodec_alloc_frame
#define av_frame_free   avcodec_free_frame
#endif

I dunno if compatibility with Ubuntu 14.04's ffmpeg is important, but it is the most recent Long Term Support release so I am hoping a few lines is worth it.  Should I make a pull request out of this, or does somebody else want to sort it out?

lgritz:
Yes, if you have a fix, please do make a pull request, and I'll happily merge it.

lgritz:
Fixed by #1168

PR #1168, opened 2015-06-15
Fix support for older ffmpeg version on Ubuntu 14.04

wrosecrans:
Mentioned in issue 1166, tested on my machine with a clean build.


lgritz:
I'll test this out on the two platforms I use, and if it works, I will merge your fix.

lgritz:
Merged this into master and RB-1.5

lgritz:
Fixed #1166

PR #1173, opened 2015-06-24
Remove unused boost dependency from public header hash.h

fpsunflower:
Had this small cleanup sitting locally for a while. These include lines are not used.


lgritz:
LGTM

PR #1175, opened 2015-06-26
Separate test for SSE3 and SSSE3

sambler:
nocona cpu's have SSE3 without SSSE3 which breaks by including tmmintrin.h

clang 3.3 - 3.7 and gcc 4.2 4.6 and 4.8 all seem to force this to be an error.

```/usr/include/clang/3.4.1/tmmintrin.h:28:2: error: "SSSE3 instruction set not enabled"```

While nocona cpu's are getting old, when a user has set nocona as the target arch the SSSE3 header comes in and breaks the build.


lgritz:
LGTM, I will merge.

lgritz:
Merged into master and RB-1.5.

PR #1176, opened 2015-06-26
ustring fixes for new gcc (5.1+) std::string ABI

### LINKED ISSUE #1174: ustring::string() returns empty string when using the new C++11 ABI
I noticed that when using GCC 5.1 with the new [libstdc++ ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) enabled, `oslc foo.osl` (from OpenShadingLanguage) would choose `.oso` as the output filename instead of `foo.oso`.

oslc uses `shader_decl()->shadername().string() + ".oso"` as the default output filename if one is not provided via the `-o` option. Further testing indicates that OpenImageIO's `ustring::string()` method is broken and returns an empty string on the new C++11 ABI.

Bellow is an example test program and GDB backtrace that show the issue.

```c
#include <OpenImageIO/ustring.h>

int main(void) {
    auto a = OpenImageIO::v1_4::ustring("123");
    std::string b = "123";

    std::cout << "a = " << a.string() << std::endl << "b = " << b << std::endl;
    return 0;
}
```

```
Breakpoint 1, main () at test.cc:8
8		std::cout << "a = " << a.string() << std::endl << "b = " << b << std::endl;

(gdb) p a
$1 = {static npos = 18446744073709551615, m_chars = 0x61d870 "123", 
  static empty_std_string = <optimized out>}

(gdb) p a.string
$2 = {const std::__cxx11::string &(const OpenImageIO::v1_4::ustring * const)} 0x4010e6 <OpenImageIO::v1_4::ustring::string[abi:cxx11]() const>

(gdb) p &a.string()
$3 = (const std::__cxx11::string *) 0x61d838

(gdb) p a.string()
$4 = ""

(gdb) c
Continuing.
a = 
b = 123
```

lgritz:
This makes it work (and adds a test).

We still need to return and do something more clever for that case. I'll need to install gcc 5.1 first to see how the internals work. But in the mean time, at least this will unbreak things for people already on gcc 5.1.

Fixes #1174

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -318,7 +318,12 @@ ustring::TableRep::TableRep (string_view strref, size_t hash)
     // the std::string to make it point to our chars!  In such a case, the
     // destructor will be careful not to allow a deallocation.
 
-#if defined(__GNUC__) && !defined(_LIBCPP_VERSION)
+#if defined(__GNUC__) && !defined(_LIBCPP_VERSION) && defined(_GLIBCXX_USE_CXX11_ABI) && _GLIBCXX_USE_CXX11_ABI
+    // NEW gcc ABI
+    // FIXME -- do something smart with this.
+
+#elif defined(__GNUC__) && !defined(_LIBCPP_VERSION)
+    // OLD gcc ABI
     // It turns out that the first field of a gcc std::string is a pointer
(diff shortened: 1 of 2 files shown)


lgritz:
Reported in #1174 that this fixes it, so I merged into both master and RB-1.5.

PR #1177, opened 2015-06-29
Clarify iv build failure message

wrosecrans:
This is fairly pedantic and tiny, but I was being dumb today before I ingested enough caffeine while building oiio.  Slightly clearer warning might save somebody a few minutes doing what I did at some point in the future.

The old error message mentioned Qt and OpenGL, so I tried fixing Qt (which wasn't broken) because I didn't realize it was actually failing because I didn't have GLEW on the machine.  New warning mentions GLEW explicitly in addition to Qt and OpenGL, which matches the actual Cmake check.   No change to build functionality, just extra idiot-proofing.


lgritz:
LGTM, I will merge.

lgritz:
Merged

PR #1180, opened 2015-07-06
Fix bugs with rangecompress and rangeexpand when using luma.

edgarv:
This fixes a bug where the (potentially uninitialized) destination buffer was used as the source for computing luma, when the destination and source buffers are different. Also adds a check against invalid scaling due to zero or negative luma (which might appear as residual error from a previous computation), and additional cases for the test suite.


lgritz:
LGTM, thanks for the fix.

lgritz:
Merged into master and RB-1.5

Issue #1089, opened 2015-03-17
Old unused-localtype-defs warning now trips up on Clang 3.6


mdriftmeyer:
From a prior email discussion:

``Re: OpenImageIO Release-1.4.15

Oh, I get it!

In the main CMakeLists, circa line 88 in RB-1.4 (and line 94 in master), it says

    add_definitions ("-Wunused-local-typedefs")

But this is backwards. I think we want

    add_definitions ("-Wno-unused-local-typedefs")

Richard, can you give that a quick try on your end and see if it suppresses the warnings? If so, I'll add it on my end.''

With the release of Clang 3.6 the same GCC 4.8 warnings suppressed are now happening with Clang.

I added:

add_definitions ("-Wno-unused-local-typedefs")

under the CMakeLists.txt line 95

if (CMAKE_COMPILER_IS_GNUCC AND (NOT CMAKE_COMPILER_IS_CLANG) AND (NOT ${GCC_VERSION} VERSION_LESS 4.8))
    # suppress a warning that Boost::Python hits in g++ 4.8
    add_definitions ("-Wno-error=unused-local-typedefs")
    add_definitions ("-Wno-unused-local-typedefs")  // Line 95

Now everything compiles cleanly, including the testing tree and all tests. Python/Python3 with boost compiles cleanly now with libboost 1.55 Debian.

//Compile with clang libc++
USE_LIBCPLUSPLUS:BOOL=OFF

Will keep lib++ use off until that's resolved.

mdriftmeyer:
Good lord. Pulling trunk and you still don't have the following added to CMakeLists.txt

add_definitions ("-Wno-unused-local-typedefs")

Without it and Clang 3.6 and newer craps the bed on BOOST PYTHON.

lgritz:
PR to fix this in #1182.

lgritz:
Committed the fix to master and RB-1.5.

PR #1182, opened 2015-07-12
Fixes for clean build with clang 3.6

lgritz:


lgritz:
Fixes #1089

lgritz:
Backported to RB-1.5 as well as master.

lgritz:
Oops, this fix will fail for certain older versions of clang.
Hastily preparing a fix...

PR #1185, opened 2015-07-17
Have TIFF output recognize special "tiff:write_exif" metadata.

lgritz:
If present and set to 0, will not write an Exif tag directory to the file, even if Exif metadata is otherwise present. This can be useful when writing a TIFF file intended to be read by an application that links to a version of libtiff that is too old to properly read the Exif data and don't want it confused.

Note in particular that PRMan may fall into that category. Therefore, if you are using 'maketx' to prepare texture files for use with PRMan, you may find it prudent to set this special metadata like this:

    maketx input.tif -prman -attrib "tiff:write_exif" 0 -o output.tx

### DIFF
--- a/src/doc/builtinplugins.tex
+++ b/src/doc/builtinplugins.tex
@@ -973,6 +973,42 @@ photometric mode (will convert to ``minisblack'').
 The TIFF plugin attempts to support all the standard Exif, IPTC, and XMP
 metadata if present.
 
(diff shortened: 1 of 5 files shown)


lgritz:
Amended the patch to have maketx automatically set this attribute when the --prman command-line option is used.

johnburnett:
Confirmed this makes prman happy again.  Tested with both 19.0.1468883 and 20.0.1505069.

PR #1187, opened 2015-07-19
added the colorconfig arg to python imageBufAlgo API methods

jephill:
I needed a way to swap the ColorConfig in the python API when using the imageBufAlgo.colorconvert, imageBufAlgo.ociolook, and imageBufAlgo.ociodiscplay methods. Rather than exposing the entire ColorConfig class, I just pass a string pointing to the desired OCIO config file and let C++ create the object.

### DIFF
--- a/src/python/py_imagebufalgo.cpp
+++ b/src/python/py_imagebufalgo.cpp
@@ -27,7 +27,7 @@
 
   (This is the Modified BSD License)
 */
-
+#include "OpenImageIO/color.h"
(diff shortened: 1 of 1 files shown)


lgritz:
Sure, LGTM. Sorry for the delay.
Do you need this in 1.5, or is master sufficient?

jephill:
Fantastic -- master should be sufficient.

lgritz:
Merged into master, and I also added the new call signatures to the docs source.

PR #1193, opened 2015-07-24
GIF write support

lgritz:
Add GIF output (previously we only had read support for GIF).

I use the handy project found here: https://github.com/ginsweater/gif-h which is public domain. (Thanks, Charlie Tangora. And thanks, Chris Kulla for alerting me to this project.)

As a fun side effect (and indeed, primary goal), it is now possible to create animated GIF images with OIIO:

    oiiotool foo*.jpg -siappendall -attrib FramesPerSecond 10.0 -o anim.gif

(Read all the named JPEG files, append them into subimages of a single "multi-image", set the FramesPerSecond attribute, and output as GIF.)

Along the way, there are some other minor changes which should be self-explanatory, including standardizing on "FramesPerSecond" as the metadata name for fps.


fpsunflower:
LGTM

Very cool! I take it the experiments with other palette generation algorithms didn't pan out? Or not worth the effort compared to what `gif.h` ships with?

lgritz:
I started using gif_lib (like we use with the reader) and wrote a couple of my own quantizers to generate palette images and was dissatisfied. I was chasing what I assume are bugs in my implementation, when you pointed me to the gif.h project. I tried that instead, it worked fine, had a vastly simpler API than gif_lib (which I don't find it at all self-explanatory and there aren't good docs), and the palette generation looked much better than mine. So I just threw out what I did and stuck to gif.h. There didn't seem to be any point chasing down the bugs in mine when gif.h did the job just fine.

I wish I had the equivalent one-header-file-just-works solution for reading GIF. This experienced really soured me on gif_lib. But the reader works, so there's no point taking time to mess with it.

Also, this gave me a newfound (extremely negative) impression of GIF. I mean, I knew it was bad, but whoa, is it BAD. If it weren't for the ubiquity of animated GIFs (and support for them in browsers) and nicety of being able to generate them with an OIIO toolchain, I would have no trouble eschewing support for GIF on principle.

lgritz:
Just because somebody is bound to ask...

My biggest complaints about the GIF format are that it's restricted to 3-channel images (with the option to have pixels transparent, so it's like having a 1-bit alpha as well), 8 bits of precision, and PALETTE as the only option -- you can only have 256 discrete colors in an image. Also, no usable metadata. It's got "80's dialup modem" written all over it. It's only because other formats like JPEG, PNG, or TIFF didn't think to include support for time series or subimages, that GIF has become ubiquitous for no-sound animated images. Now we can't get rid of it. It's a travesty.

GifLib, which I'd never used before (somebody else wrote and contributed our GIF reader), is pretty awful. The API makes no sense to me, it's not well documented in the header or anywhere else. For a format that requires palette images, I find it astounding that gif_lib doesn't provide a built-in utility function that generates a decent palette from an RGB image.

But it's neato being able to use oiiotool to make animated GIFs. Very handy for online documentation.

PR #1195, opened 2015-07-28
PTex build changes -- use external PTex if available.

lgritz:
Using FindPTex.cmake from the PTex project itself, search for an external PTex library. Use it if found.

If external PTex is not found, and as long as USE_PTEX is not disabled, fall back on the existing "embedded source" PTex implementation.

The fact that we have an embedded copy of the Ptex source is a bit of an anachronism. I think PTex is common enough in VFX pipelines now that for anybody who needs it, we can probably expect them to have a proper installation. I will eventually remove the embedded one entirely, once I get the all-clear from everybody (including my facility, which needs it for a little while longer).

I also added a simple testsuite entry for the ptex reader.


johnburnett:
Looks good to me, thanks!

PR #1198, opened 2015-08-06
imageinput: added read_native_channels, fixed bug in multipart exrs

melevine:
Added a new method read_native_channels to imageinput and python wrappers for read_native_scanlines and read_native_channels. Fixed a bug for multi-part exrs, was setting m_pixel type incorrectly


lgritz:
Closing with apologies for letting this fester for so long. At some point, `read_native_scanlines` and `read_native_tiles` were extended with channel ranges, which I think gives most of the desired functionality of this proposal (though less general than an arbitrary channel list/ordering).

PR #1199, opened 2015-08-06
fixed compiler warnings due to type mismatch (size_t vs. int)

mjmvisser:
Fixes this compiler warning/error:

`error: comparison between signed and unsigned integer expressions`

cheers


lgritz:
LGTM, I'll merge it.

PR #1200, opened 2015-08-18
Improve C++11 and C++14 readiness

lgritz:
This is all still compatible with C++03, but clean up our macros, add USE_CPP14, and lay groundwork for making declarations for constexpr and noexcept with the right compatibility requirements.


fpsunflower:
LGTM

PR #1201, opened 2015-08-19
Timer: enums, naming, print upon destruction

lgritz:
Also some cleanup of ScopedTimer -- doesn't need to be a template.

I did this a while back, never submitted, but now I want to use part of it.


fpsunflower:
LGTM

PR #1202, opened 2015-08-20
OIIO_STATIC_ASSERT and OIIO_STATIC_ASSERT_MSG

lgritz:
Use standard stuff for C++ >= 11 varieties, fall back on boost otherwise.

There are some place where I'd like to use static assertions.


fpsunflower:
LGTM

Issue #1203, opened 2015-08-20
LIBRAW_DECODER_FLATFIELD no longer supported


cwilling:
The most recent version of LibRaw (0.17.0) no longer defines LIBRAW_DECODER_FLATFIELD, therefore compilation of oiio fails with:
/tmp/SBo/oiio-Release-1.5.18/src/raw.imageio/rawinput.cpp: In member function ‘virtual bool OpenImageIO::v1_5::RawInput::open(const string&, OpenImageIO::v1_5::ImageSpec&, const OpenImageIO::v1_5::ImageSpec&)’:
/tmp/SBo/oiio-Release-1.5.18/src/raw.imageio/rawinput.cpp:207:48: error: ‘LIBRAW_DECODER_FLATFIELD’ was not declared in this scope
             if (!(decoder_info.decoder_flags & LIBRAW_DECODER_FLATFIELD)) {

Removing the relevant lines from rawinput.cpp allows the build to complete, as does hiding them behind an #ifdef LIBRAW_DECODER_FLATFIELD (enabling its use with older versions of LibRaw). Not sure if that has unintended runtime consequences though.

chris

lgritz:
Fixed in #1204

PR #1204, opened 2015-08-21
Only reference LIBRAW_DECODER_FLATFIELD within an #ifdef that ensures…

cwilling:
… that it is defined

This change addresses Issue #1203 

LibRaw 0.17.0 no longer defines LIBRAW_DECODER_FLATFIELD. Hiding its use behind an #ifdef enables continuing use of earlier, as well as the most recent, LibRaw versions.


lgritz:
LGTM, thanks for the fix. I will merge shortly.

lgritz:
Fixes #1203

PR #1205, opened 2015-08-23
Refactor array_view to be more in line with what is developing for C++17

lgritz:
This is very close to backwards-compatible with our first implementation (which we did not yet use in any public APIs, so we shouldn't feel very conservative about it anyway).

The biggest change is that it is now templated on Rank (number of dimensions), so you can have 2D, 3D, etc., array views. The default is 1D, which works much as before.

Even for the 1D case, we re-jiggered our implementation and API to be very compatible with what appears to be on its way into C++17. After C++17 is released, we hope to phase out our own, so we seek compatibility. One change that came along with this is the offset<> and bounds<> templates in coordinate.h.

Also for conformance to what appears to be heading for C++17, we changed our implementation of array_view_strided to have the strides measured by sizeof(T), instead of in bytes. I kind of regret this choice -- bytes are more flexible if using an array_view to span oddly sized structures, say -- but I fear that incompatibility with C++17 (for a same-named class) is a worse evil. If it becomes a problem, we can always make a separate array_view_byte_strided class later.

### DIFF
--- a/src/include/OpenImageIO/array_view.h
+++ b/src/include/OpenImageIO/array_view.h
@@ -1,7 +1,6 @@
 /*
-  Copyright 2014 Larry Gritz and the other authors and contributors.
+  Copyright 2015 Larry Gritz and the other authors and contributors.
   All Rights Reserved.
-  Based on BSD-licensed software Copyright 2004 NVIDIA Corp.
 
   Redistribution and use in source and binary forms, with or without
   modification, are permitted provided that the following conditions are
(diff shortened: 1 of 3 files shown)


lgritz:
After more than a week, nobody has said, "no, don't make it like C++17", so I'm going to do it. I think array_view is a really handy idiom anyplace where we previously would have passed a ptr and length as separate parameters, or where we would have had separate overloads for std::vector<T>& and T*, analogous to the kinds of situations where sring_view is useful.

PR #1207, opened 2015-08-26
TIFF support a wider range of compression and photometric modes.

lgritz:
The TIFF reader now correctly notices and reports the full range of possible compression modes supported by libtiff. The writer still only supports a subset of them, primarily because nobody needs them and I don't want to bother carefully testing their correctness.

Enable (previously-broken) reading of JPEG-compressed TIFF and YCrCb encoded images (even when the chroma is subsampled), as well as LAB and LOG encodings. For these rare edge cases we fall back on libtiff's "ReadRGBAImage". This makes some compromises (such as forcing uint8 and buffering the whole image), but I think it's well worth supporting these in a low-performance and low-precision way rather than not at all, yet not worth the immense effort to get fully performant support for these varieties that are extremely rare.

Modify the testsuite/tiff-suite test to include several more libtiff example images, for the new modes that we handle ok now.

This patch also adds some half-baked support for writing JPEG-encoded TIFF, but as of this checkin it is disabled because I still don't have it working correctly.

### DIFF
--- a/src/doc/builtinplugins.tex
+++ b/src/doc/builtinplugins.tex
@@ -1010,6 +1010,40 @@ Output attribute & Type & Meaning \\
 \end{tabular}
 
 
+\subsubsection*{TIFF compression modes}
+
+\noindent The full list of possible TIFF compression mode values are as
+follows ($ ^*$ indicates that \product can write that format, and is not
+part of the format name): \\
+    {\kw none}$ ^*$  ~
+    {\kw lzw}$ ^*$  ~
+    {\kw zip}$ ^*$  ~ \\
+    {\kw ccitt_t4}  ~
+    {\kw ccitt_t6}  ~
+    {\kw ccittfax3}  ~
+    {\kw ccittfax4}  ~
+    {\kw ccittrle2}  ~
+    {\kw ccittrle}$ ^*$  ~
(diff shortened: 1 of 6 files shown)


lgritz:
I updated this patch today -- now for tiled images that need the RGBA path, it can read each tile individually, without needing to buffer the whole image. So it's actually quite efficient for tiled images (as an example: tiled JPEG-compressed images). Scanline images still buffer the whole image because there is no TIFFReadRGBAScanline! Only whole image, tile, or "strip." I guess I could buffer one strip, but I'm inclined to wait until this pops up as a performance issue for somebody, since these modes are pretty rarely used.

lgritz:
Updated again -- I have output with JPEG compression working.

lgritz:
It's not clear why, presumably a bug in libtiff, but sprinkling calls to TIFFCheckpointDirectory seems to make the write work properly. I think that the patch now currently lets us support both read and write of JPEG compression in TIFF files, plus reading of several additional corners of TIFF including YCbCr, SGI log, and others.

lgritz:
Had a chance to try this out?

lgritz:
I'm going to merge this, as I think it is at least a 90% fix, if not a complete one. We can always return to patch it if there is another corner case we've missed.

PR #1208, opened 2015-08-27
Make maketx reject input images that have overscan.

lgritz:
It's hard to know how to handle them sensibly. Because display and data window origins and sizes are integers (in all file formats we know of), there's no way to downsize them for MIP levels and preserve that property. This makes for misalignments or bleeding at edges or other undesirable behavior.  Rather than have people perplexed at the results, just have maketx reject overscan files.

Skip texture-overscan and maketx-cropover tests, we don't expect them to work.

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -1107,6 +1107,14 @@ make_texture_impl (ImageBufAlgo::MakeTextureMode mode,
     }
 
     bool orig_was_overscan = (roi != roi_full);
+    // FIXME? For now, we don't have a strategy for making the MIPmaps in a
+    // sensible way for source image with overscan. Just reject them. Maybe
+    // someday we will return and do something smart, but for now a good way
+    // to deal with this case eludes us.
+    if (orig_was_overscan) {
+        outstream << "maketx ERROR: can't handle source images with overscan\n";
+        return false;
+    }
     if (orig_was_overscan) {
         configspec.attribute ("wrapmodes", "black,black");
     }
--- a/testsuite/texture-cropover/run.py
+++ b/testsuite/texture-cropover/run.py
@@ -2,6 +2,9 @@
 
 # Tests input image which is partial crop, partial overscan!
 
+# Skip this test. For now, we don't support overscan as texture input.
+exit ()
+
 command += oiiotool(parent + "/oiio-images/grid.tif " +
                     "--crop 500x1000+250+0 --fullsize 1000x800+0+100 -o grid-cropover.exr")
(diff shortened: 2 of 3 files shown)


malcolmhumphreys:
I would expect maketx to print a warning about having to crop the image to the display window and still provide a mipmap file.

Or maybe an error and then a flag to do this.

On 27 August 2015 06:16:04 BST, Larry Gritz <notifications@github.com> wrote:
>It's hard to know how to handle them sensibly. Because display and data
>window origins and sizes are integers (in all file formats we know of),
>there's no way to downsize them for MIP levels and preserve that
>property. This makes for misalignments or bleeding at edges or other
>undesirable behavior.  Rather than have people perplexed at the
>results, just have maketx reject overscan files.
>
>Skip texture-overscan and maketx-cropover tests, we don't expect them
>to work.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/OpenImageIO/oiio/pull/1208
>
>-- Commit Summary --
>
>  * Make maketx reject input images that have overscan.
>
>-- File Changes --
>
>    M src/libOpenImageIO/maketexture.cpp (8)
>    M testsuite/texture-cropover/run.py (3)
>    M testsuite/texture-overscan/run.py (23)
>
>-- Patch Links --
>
>https://github.com/OpenImageIO/oiio/pull/1208.patch
>https://github.com/OpenImageIO/oiio/pull/1208.diff
>
>---
>https://github.com/OpenImageIO/oiio/pull/1208

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

fpsunflower:
Would it make sense to mipmap the file as if the display window was equal to the data window and then do the offset during the texture lookup?

I can see why mipmapping the data "naively" leads to fractional pixels that don't make sense, but it seems that this "fractional" stuff could be handled fine during the texture lookup.

lgritz:
@malcolmhumphreys Yes, that's another approach -- simply crop to the display window and warn. At least the mapping would be right, even though there would be no texture data outside the 0-1 window.

@fpsunflower Chris, that's a possibility. In fact, the texture system kinda looks like it's already trying to do that. Maybe I need to take a closer look at exactly where it's failing. 

Let's put this PR on hold for the moment, I'll dig a bit more and see if I can find a more satisfying understanding of why overscan isn't working and if it can be ok. I feel like I tried to make it work at some point and now I need to retrace my steps.

lgritz:
Upon careful review of the sampling code, I now think that overscan works reasonably well and has for a long time. I made a new test, which I'll add to the testsuite, that really convinces me that it's fine.

I hereby retract this pull request, and I'm going to go back to the user who kicked this off and figure out why they think they're having problems.

Issue #1211, opened 2015-09-10
Nuke txReader crash with clip browser preview enabled


nrusch:
In Nuke, if you have the preview pane visible in the clip browser, the txReader will cause a crash if a .tx file is selected. This is simply because `TxReaderFormat::knobs` is not called in that context, as no format-specific knobs need to be displayed.

The simple fix is to test for a non-NULL `Knob*` in `::setMipLabels`. So this:

```c++
    void setMipLabels(std::vector<std::string> items) {
        mipLevelKnob_->enumerationKnob()->menu(items);
    }
```
becomes this:
```c++
    void setMipLabels(std::vector<std::string> items) {
        if (mipLevelKnob_)
            mipLevelKnob_->enumerationKnob()->menu(items);
    }
```

Source: https://github.com/OpenImageIO/oiio/blob/master/src/nuke/txReader/txReader.cpp#L56

I was considering going for a full pull request, but since this is such a minor fix, I figured it would be simple enough to just describe it. If you would prefer I go the pull request route, just let me know.

Thanks.

lgritz:
Yeah, if you don't mind? A PR is just a whole lot easier for me to just integrate in.

nrusch:
Sure. PR is https://github.com/OpenImageIO/oiio/pull/1212

lgritz:
Fixed by #1212.

PR #1212, opened 2015-09-11
nuke: Handle NULL Knob* in TxReaderFormat::setMipLabels

nrusch:
Fixes a crash in the Nuke txReader when a .tx file is selected in the Nuke clip browser while the preview pane is open.

Issue: https://github.com/OpenImageIO/oiio/issues/1211


lgritz:
Fixes #1211

lgritz:
LGTM, I will merge it. Thanks for the fix.

lgritz:
I also backported to 1.5

PR #1213, opened 2015-09-15
ociofiletransform

lgritz:
In ImageBufAlgo (C++ and Python) and via oiiotool command line, we exposed OpenColorIO color conversion, look transforms, and display transforms.

There's something called a "filetransform" in OCIO which we didn't expose and now somebody needs it. This patch does it.

Also fix a minor bug in oiiotool (oiiotool.cpp:3530) using --autocc that will crash if the color space isn't found in the filename, because it checks for metadata before the file has been read, and falls off a NULL pointer when looking for the ImageSpec. A simple call to read the file in that case fixes.


fpsunflower:
LGTM

PR #1217, opened 2015-09-17
Changes for SPI build procedures

lgritz:
I don't expect anyone not at SPI to understand or care about this. It's all in the site/spi specific parts, so won't affect anyone else.

Bump SpComp2 version and change spcomp2 build to use clang.

Major refactor of the Makefile-bits-arnold and Makefile-bits-spcomp2 to bring them in sync. It was a mess to try to understand how they differed before. Now you can xxdiff them and they are 90% identical, easy to see how they differ. I think that at some point after our upcoming OS upgrade (which will deprecate the whole 'spinux1' platform) I'll probably completely unify these two files. But for now, this is already a big cleanup accomplished, and I'm exhausted.

The overhaul isn't quite as awful as it looks. The arnold makefile had the spinux1 case first, then rhel4, whereas the spcomp2 version was the other way around. That alone made it virtually impossible to understand the diffs between them with any usual diffing tool. Moving them to have the same ordering of platform cases helped a lot, then I started reconciling differences from there. There *are* still differences between how we build for SpComp2 versus Arnold, but I've tried to make it so that they are only real differences, not merely notational differences in the makefiles that are equivalent.

### DIFF
--- a/site/spi/Makefile-bits-arnold
+++ b/site/spi/Makefile-bits-arnold
@@ -11,64 +11,139 @@ endif
 # Default namespace
 NAMESPACE ?= 'OpenImageIO_Arnold'
 MY_CMAKE_FLAGS += -DEXTRA_CPP_ARGS:STRING="-DOIIO_SPI=1"
-PYVER ?= 2.6
-PYVERNOSPACE=${shell echo ${PYVER} | sed "s/\\.//"}
 NUKE_VERSION ?= 8.0v5
+SPCOMP2_SHOTTREE_LOCATION = /shots/spi/home/lib/SpComp2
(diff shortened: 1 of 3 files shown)


fpsunflower:
Since they are mostly identical -- would it be possible to consolidate them into one big file and two smaller ones?

lgritz:
The differences are not all at the beginning or all at the end, and are in some cases inside the platform cases.

Many of the differences are spinux1 only, the rhel4 builds are much closer to identical. That's why I want to postpone the rest of the unification until spinux1 goes away.

lgritz:
It's not perfect, but at least it's no longer an inscrutable mess. I'm satisfied with that for now, knowing that I'll do a second simplifying overhaul in a few months.

PR #1221, opened 2015-09-24
Make TIFF read robust to subimge-to-subimage changes in some header fields

### LINKED ISSUE #1220: Bizzarre tiff causes OIIO segfault.  To fix or not?
Somehow, we ended up with a very bizarre .tx file that crashes OpenImageIO when reading some parts of it (as with 'oiiotool --stats', or during a render, for example).  Other than the fact that is is a perfectly valid tiff, I'll spare the details.  I still can't fathom how it was created.  Regardless, the specific cause of the crash is this optimization in tiffinput.cpp:

    // If we're emulating a MIPmap, only resolution is allowed to change
    // between MIP levels, so if we already have a valid level in m_spec,
    // we don't need to re-parse metadata, it's guaranteed to be the same.
    bool read_meta = !(m_emulate_mipmap && m_tif && m_subimage >= 0);

It turns out that in this particular case, that's not a safe assumption.  The question is--what's the best way to keep this optimization and still provide some safety from oddly composed, but otherwise perfectly valid, files?

One thought--add a 'pedantic=1' option (that could be set via env var) which forces read_meta to be always true.

Thoughts?

lgritz:
Rearrange the order in which we examine TIFF header order. When reading MIPmap levels (beyond the first one), we assume that all the ordinary metadata should be the same for all levels, so we don't laboriously read all that stuff. But in our zeal to skip this needless work, we took the early out before examining some essential fields such as planarconfig, photometric interpretation, palette processing, and extrasamples.  If a file should change this from one MIP level to the next -- which no sane file should do, but hey, it did happen in the wild -- that would bring chaos. Now we are more robust to this, while still avoiding all the detailed metadata reading for non-head MIP levels.

Fixes #1220

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -737,17 +737,13 @@ TIFFInput::readspec (bool read_meta)
         break;
     }
 
-    // If we've been instructed to skip reading metadata, because it is
-    // guaranteed to be identical to what we already have in m_spec,
-    // skip everything following.
-    if (! read_meta)
-        return;
-
     // Use the table for all the obvious things that can be mindlessly
     // shoved into the image spec.
-    for (int i = 0;  tiff_tag_table[i].name;  ++i)
-        find_tag (tiff_tag_table[i].tifftag,
-                  tiff_tag_table[i].tifftype, tiff_tag_table[i].name);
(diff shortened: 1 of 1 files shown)


lgritz:
Fixes #1220

PR #1225, opened 2015-10-02
resolves "error: comparison between signed and unsigned integer expressions"

mjmvisser:
Hi, this patch fixes a compiler warning due to signed/unsigned comparison mismatch.

Unless there's a valid reason to have a negative number of bits per pixel that I'm not aware of?

cheers


lgritz:
LGTM. I will merge this shortly.

lgritz:
merged

Issue #1123, opened 2015-04-17
ImageCache: On-demand layer reading for multi-layered files


MrKepzie:
Hi,

First off thanks for this amazing library. We use it extensively in Natron  (open-source compositing software) (https://github.com/MrKepzie/Natron/tree/workshop via OpenFX: https://github.com/MrKepzie/openfx-io ) for most of the image reading/writing.

We have received some complaints by our users that the reading of multi-layered EXR files was freakishly slow compared to other similar applications. I have narrowed it down to the ImageCache of OIIO. We make use of the ImageCache::get_pixels function. The problem is that if the image is not present into the cache, the OIIO cache internaly will create a new Tile and calls
    ImageInput::read_image
 which in turns calls   
    ImageInput::read_scanlines(ybegin,yend,z,0,m_spec.nchannels,format,data,xstride,ystride)

So basically regardless of what we requested via the *get_pixels* function with *chbegin* and *chend*, the Tile will always decode all available layers in the image. Needless to say that for EXR files rendered out with many passes it takes a while.  

Could this be improved to only read the channels that were requested by *ImageCache::get_pixels* in the first place ?

Our temporary workaround is to use the ImageInput::read_scanlines API directly, but we suffer from the lack of caching because if the user of Natron requests more than 1 layer to display, it will open as many ImageInput on our side that the the user requested, which is not efficient at all.

lgritz:
The design choice was that rather than storing each channel separately in the cache, since all channels are usually used together (at least, for RGBA files), we should store them together.

But I think you do identify an important edge case -- if there are truly a large number of channels, but the typical access pattern is to only need a few of them.

I think we could change the way the cache behaves. A tile in cache could be identified by not just (x, y, z, subimage, miplevel) but also by (firstchannel, nchannels), so that different channel sets in the same file would end up in different tiles. We could constrain it so that a file with, say, 5 or fewer channels always stays on one tile even if you ask for a subset, so asking for 2 channels in a 4-channel file would still keep all four channels in the tile, whereas if you ask for channels 17-20 in a 32 channel file would end up with a tile containing just those channels, saving the cache space and maybe the I/O as well (for some formats). I think that accommodation would preserve full performance of the 99% case of asking for 3 or 4 channels from a 3 or 4 channel file.

Let me look at the code and formulate a plan. This sounds like a very reasonable request.

MrKepzie:
[...]

PR #1226, opened 2015-10-03
Allow ImageCache/TextureSystem to store partial channel sets in cache.

lgritz:
Allow ImageCache/TextureSystem to store partial channel sets in cache.

ImageCache's original design would read all channels into a tile,    because (1) most files have only a few channels, (2) most uses (for    example, texture) tend to use most or all of the channels in the file,    in close temporal proximity if not in the same lookup, (3) most file    formats don't have a way to read a channel subset that's any more    efficient than reading all of them.

However, for the rare case that there are MANY channels, only a few of    which are needed (for example, looking up 3 channels of textures from a    40 channel file), this is very wasteful of texture space, and MAY    (depending on the format) also be wasteful of I/O.

So this patch:

* Modifies ImageCache so that a tile can store any contiguous subset      of channels. This means the TileID stores a chbegin/chend range,      the hashes consider this, and certain ImageCache public API calls      now have varieties that take chbegin/chend parameters.

* ImageCache will create cache tiles that store exactly the channel      range requested. The default is all channels in the tile. Use subsets      wisely -- if you have a 4-channel file and you separately look up      channel ranges [0,1) and [0,3), it will redundantly read and store      the overlapping channels!

* TextureSystem is modified to ask for all channels when there are only      a few channels in the file, but when there are more than      max_tile_channels (a new settable attribute of the TS) channels in      the file, it will ask the IC for only the channels it needs for the      texture lookup. The max_tile_channels is set by default to 5, which      means for the vast majority of files (R, RGB, RGBA, RGBAZ) it will      read all channels and behave as before. But if you have a mega-channel      file, it will try to read and cache only the subset of channels you      requested with the texture call. If you don't want the new behavior,      just set max_tile_channels to something bigger.

The bottom line is that if your app uses IC with the old API calls, or    if you use TS with texture files having 5 or fewer channels (which is    almost always the case), you should see no change in behavior or    performance. But for those of you who for some reason deal with files of    many many channels, accessing a small subset of them via texture or IC    should result much more efficient cache behavior because the IC won't be    caching channels you aren't accessing.

### DIFF
--- a/src/doc/imagecache.tex
+++ b/src/doc/imagecache.tex
@@ -634,13 +634,15 @@ image file will be filled with zero values.
          \bigspc\spc int xbegin, int xend, int ybegin, int yend, \\
          \bigspc\spc int zbegin, int zend, int chbegin, int chend,\\
          \bigspc\spc TypeDesc format, void *result, \\
-         \bigspc\spc stride_t xstride, stride_t ystride, stride_t zstride) \\
+         \bigspc\spc stride_t xstride, stride_t ystride, stride_t zstride,\\
+         \bitspc\spc int cache_chbegin=0, int cache_chend=-1) \\
 bool {\ce get\_pixels} (ImageHandle *file, Perthread *thread_info, \\
          \bigspc\spc int subimage, int miplevel, \\
          \bigspc\spc int xbegin, int xend, int ybegin, int yend, \\
          \bigspc\spc int zbegin, int zend, int chbegin, int chend,\\
          \bigspc\spc TypeDesc format, void *result, \\
-         \bigspc\spc stride_t xstride, stride_t ystride, stride_t zstride)}
+         \bigspc\spc stride_t xstride, stride_t ystride, stride_t zstride,\\
+         \bitspc\spc int cache_chbegin=0, int cache_chend=-1)}
 For an image specified by either name or handle,
 retrieve the rectangle of pixels and subset of channels
 of the designated {\cf subimage} and {\cf miplevel}, storing the pixel values
@@ -656,7 +658,10 @@ The rectangular region and channel set to be retrieved includes
 {\cf begin} but does not
 include {\cf end} (much like STL begin/end usage).
 Requested pixels that are not part of the valid pixel data region of the
-image file will be filled with zero values.
+image file will be filled with zero values. The optional parameters
+{\cf cache_chbegin} and {\cf cache_chend} can be used to tell the
+\ImageCache to read and cache a subset of channels (if not specified,
+all the channels of the file will be stored in the cached tile).
 \apiend
 
 
--- a/src/doc/texturesys.tex
+++ b/src/doc/texturesys.tex
@@ -551,6 +551,15 @@ The default value of zero means that all missing channels will get
 the ``fill'' color.
 \apiend
 
+\apiitem{int max_tile_channels}
+\NEW % 1.6
+Sets the maximum number of color channels in a texture file for which all
+channels will be loaded as cached tiles. Files with more than this number
+of color channels will have only the requested subset loaded, in order
+to save cache space (but at the possible wasted expense of separate tiles
+that overlap their channel ranges). The default is 5.
+\apiend
+
 \apiitem{string latlong_up}
 Sets the default ``up'' direction for latlong environment maps (only
 applies if the map itself doesn't specify a format or is in a format
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -43,6 +43,8 @@
 
 OIIO_NAMESPACE_BEGIN
 
+struct ROI;
+
 namespace pvt {
 // Forward declaration
 class ImageCacheImpl;
@@ -243,7 +245,9 @@ public:
     /// enough to accommodate the requested rectangle (taking into
     /// consideration its dimensions, number of channels, and data
     /// format).  Requested pixels outside the valid pixel data region
-    /// will be filled in with 0 values.
+    /// will be filled in with 0 values. The optional cache_chbegin and
+    /// cache_chend hint as to which range of channels should be cached
+    /// (which by default will be all channels of the file).
     ///
     /// Return true if the file is found and could be opened by an
     /// available ImageIO plugin, otherwise return false.
@@ -252,34 +256,47 @@ public:
                     int ybegin, int yend, int zbegin, int zend,
                     int chbegin, int chend, TypeDesc format, void *result,
                     stride_t xstride=AutoStride, stride_t ystride=AutoStride,
-                    stride_t zstride=AutoStride) = 0;
+                    stride_t zstride=AutoStride,
+                    int cache_chbegin = 0, int cache_chend = -1) = 0;
     virtual bool get_pixels (ImageHandle *file, Perthread *thread_info,
                     int subimage, int miplevel, int xbegin, int xend,
                     int ybegin, int yend, int zbegin, int zend,
                     int chbegin, int chend, TypeDesc format, void *result,
                     stride_t xstride=AutoStride, stride_t ystride=AutoStride,
-                    stride_t zstride=AutoStride) = 0;
+                    stride_t zstride=AutoStride,
+                    int cache_chbegin = 0, int cache_chend = -1) = 0;
 
     /// Define an opaque data type that allows us to have a pointer
     /// to a tile but without exposing any internals.
     class Tile;
 
-    /// Find a tile given by an image filename, subimage & miplevel, and
-    /// pixel coordinates.  An opaque pointer to the tile will be
-    /// returned, or NULL if no such file (or tile within the file)
-    /// exists or can be read.  The tile will not be purged from the
-    /// cache until after release_tile() is called on the tile pointer
-    /// the same number of times that get_tile() was called (refcnt).
-    /// This is thread-safe!
+    /// Find a tile given by an image filename, subimage & miplevel, channel
+    /// range, and pixel coordinates.  An opaque pointer to the tile will be
+    /// returned, or NULL if no such file (or tile within the file) exists
+    /// or can be read.  The tile will not be purged from the cache until
+    /// after release_tile() is called on the tile pointer the same number
+    /// of times that get_tile() was called (refcnt). This is thread-safe!
+    /// If chend < chbegin, it will retrieve a tile containing all channels
+    /// in the file.
     virtual Tile * get_tile (ustring filename, int subimage, int miplevel,
-                             int x, int y, int z) = 0;
+                             int x, int y, int z,
+                             int chbegin = 0, int chend = -1) = 0;
     virtual Tile * get_tile (ImageHandle *file, Perthread *thread_info,
                              int subimage, int miplevel,
-                             int x, int y, int z) = 0;
+                             int x, int y, int z,
+                             int chbegin = 0, int chend = -1) = 0;
 
     /// After finishing with a tile, release_tile will allow it to
     /// once again be purged from the tile cache if required.
     virtual void release_tile (Tile *tile) const = 0;
+
+    /// Retrieve the data type of the pixels stored in the tile, which may
+    /// be different than the type of the pixels in the disk file.
+    virtual TypeDesc tile_format (const Tile *tile) const = 0;
+
+    /// Retrieve the ROI describing the pixels and channels stored in the
+    /// tile.
+    virtual ROI tile_roi (const Tile *tile) const = 0;
 
     /// For a tile retrived by get_tile(), return a pointer to the
     /// pixel data itself, and also store in 'format' the data type that
@@ -317,14 +334,25 @@ public:
                            const ImageSpec *config=NULL) = 0;
 
     /// Preemptively add a tile corresponding to the named image, at the
-    /// given subimage and MIP level.  The tile added is the one whose
-    /// corner is (x,y,z), and buffer points to the pixels (in the given
-    /// format, with supplied strides) which will be copied and inserted
-    /// into the cache and made available for future lookups.
+    /// given subimage, MIP level, and channel range.  The tile added is the
+    /// one whose corner is (x,y,z), and buffer points to the pixels (in the
+    /// given format, with supplied strides) which will be copied and
+    /// inserted into the cache and made available for future lookups.
+    /// If chend < chbegin, it will add a tile containing the full set of
+    /// channels for the image.
+    virtual bool add_tile (ustring filename, int subimage, int miplevel,
+                     int x, int y, int z, int chbegin, int chend,
+                     TypeDesc format, const void *buffer,
+                     stride_t xstride=AutoStride, stride_t ystride=AutoStride,
+                     stride_t zstride=AutoStride) = 0;
+    // DEPRECATED(1.6) -- add a tile with all channels.
     virtual bool add_tile (ustring filename, int subimage, int miplevel,
                      int x, int y, int z, TypeDesc format, const void *buffer,
                      stride_t xstride=AutoStride, stride_t ystride=AutoStride,
-                     stride_t zstride=AutoStride) = 0;
+                     stride_t zstride=AutoStride) {
+        return add_tile (filename, subimage, miplevel, x, y, z, 0, -1,
+                         format, buffer, xstride, ystride, zstride);
+    }
 
     /// If any of the API routines returned false indicating an error,
     /// this routine will return the error string (and clear any error
--- a/src/include/OpenImageIO/texture.h
+++ b/src/include/OpenImageIO/texture.h
@@ -334,6 +334,7 @@ public:
     ///     int failure_retries : how many times to retry a read failure
     ///     int deduplicate : if nonzero, detect duplicate textures (default=1)
     ///     int gray_to_rgb : make 1-channel images fill RGB lookups
+    ///     int max_tile_channels : max channels to store all chans in a tile
     ///     string latlong_up : default "up" direction for latlong ("y")
     ///
     virtual bool attribute (string_view name, TypeDesc type, const void *val) = 0;
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -1543,6 +1543,28 @@ ImageCacheImpl::getstats (int level) const
                 out << (void *)this;
         }
         out << ") ver " << OIIO_VERSION_STRING << "\n";
+
+        std::string opt;
+#define BOOLOPT(name) if (m_##name) opt += #name " "
+#define INTOPT(name) opt += Strutil::format(#name "=%d ", m_##name)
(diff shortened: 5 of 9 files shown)


lgritz:
Fixes #1123

lgritz:
Any objections to this, or feedback? I will merge in the next couple days if nobody objects.

johnhaddon:
What are the implications for an application that requests a single channel at a time via `ImageCache::get_pixels()`, but will typically do so in turn for each of the RGBA channels? Presumably in the past a single tile would have serviced each of those requests, so there would only be one read at the ImageInput level? Does this change mean that potentially each channel access via `get_pixels()` will now pay for its own read, to fill its own tile, and that for many formats each of those reads will also read all the other channels and throw them away?

OpenImageIO:
Currently, yes, this patch makes IC behave as "I cache what you asked for", so if you call IC::get_pixels with each channel separately, it will end up doing multiple reads and caching the channels as separate tiles.

The IC is dumb and literal, but the TextureSystem tries to be smart about common idioms. In particular, when the TS functions (including TS::get_texels) encounter files with a reasonable number of channels (by default, 5, which would include RGB and RGBA, the 99% common cases), it will ask the IC for a tile containing all channels, even if the call itself is requesting only a subset thereof. Only for "many channel" files does it narrow the range in the cached tiles. It's a settable option of the TextureSystem, so you can just set "max_tile_channels" to 1000 or something to turn off this behavior entirely (from the TextureSystem's point of view).

So I can see a few ways to proceed:

1. Rather than IC::get_pixels(), you could call TS::get_texels(), and it will handle the issue of coalescing the channels into one tile for <= 5 channel files (and/or you can set the max_tile_channels option to be larger and avoid the issue). If you notice a performance difference between IC::get_pixels and TS::get_texels (they are somewhat different implementations and the latter may be clunkier), let's address that separately, it's fixable if it's a legit issue.

2. If you must operate at the level of IC rather than TS, you could ask IC::get_pixels for all channels (so it will cache all channels) and then copy out the one channel you need. Or change your code to loop over all channels for the tile at once, rather than separate the requests for different channels.

3. Or if you don't want to change your code, don't want to bother with an extra buffer copy, but must use IC, then I could add the same kind of max_tile_channels logic to the IC::get_pixels, so that it would automatically cache all channels for moderately-channeled images.



> On Oct 14, 2015, at 6:25 AM, John Haddon <notifications@github.com> wrote:
> 
> What are the implications for an application that requests a single channel at a time via ImageCache::get_pixels(), but will typically do so in turn for each of the RGBA channels? Presumably in the past a single tile would have serviced each of those requests, so there would only be one read at the ImageInput level? Does this change mean that potentially each channel access via get_pixels() will now pay for its own read, to fill its own tile, and that for many formats each of those reads will also read all the other channels and throw them away?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/OpenImageIO/oiio/pull/1226#issuecomment-148048248>.
> 

--
Larry Gritz
lg@larrygritz.com

johnhaddon:
Thanks for the info. It sounds like our current use pattern (this is in Gaffer's ImageReader node) is pretty much worst-case for this new setup, but I don't want to be the fly in the ointment if this works well for everyone else. Gaffer actually has its own caching mechanism for all nodes, but we currently turn it off for the ImageReader on the grounds that we're getting caching from OIIO::ImageCache anyway. It might be that longer term we need to drop down a level and make direct use of OIIO::ImageInput coupled with our own cache. We're in the process of adding support for deep data, and since ImageCache doesn't appear to support that we may end up going that route sooner rather than later anyway.

I wonder if it might be beneficial to add a query to ImageInput to help drive the decision making at the cache level, rather than use the current heuristics. Perhaps a query like "if I were to read this particular channel, what is the full set of channels you would be forced to read by the underlying format?". Or "if I were to read this channel, which other channels would you recommend also reading from a performance standpoint?". It seems that deferring to the formats themselves has the potential for allowing better performance while removing the need for the user to guess an appropriate value for "max_tile_channels", which may well vary from file to file and format to format anyway. It feels like this is the kind of query I'd be looking for if we were to bolt ImageInput directly to our own cache.

Anyway, assuming this is getting merged to master only, I think we have time to formulate a plan either with our own cache or using one of your suggestions, so don't let my comments stand in the way of anything.

HughMacdonald:
In addition to John's points, do you think it might be worth-while to be able to request a selection of channels (possibly only 1 channel) and, at the same time, give ImageCache a hint as to which channels to load alongside the requested one. That way, get_pixels() would cause more channels to be cached than it is actually returning.

As John mentioned, though, we're in the process of implementing Deep support in Gaffer, so will most likely have to shift over to using ImageInput and our own caching, so all this may be moot.

lgritz:
@johnhaddon  and @HughMacdonald  : It sounds like the easiest solution is to make a variety of ImageCache::get_pixels that let you hint at how much you want cached, in addition to how much you want to retrieve at the moment. I'll make that part of this PR. This should be really easy and I may have time to do it today.

Every bit of this affair is too big a change (and an API break) to go anywhere but master, so if you are on an RB, there is no need to deal with this right away.

But I think that it's silly for you to reinvent the caching wheel here. I've been asked for it a couple times already, it's on my list. I'd rather give the IC the ability to handle deep data, than have people abandoning IC for their own separately-developed equivalents. The whole idea of OIIO is to avoid pointless duplication.

lgritz:
@johnhaddon and @HughMacdonald : for get_pixels, do you think it would be better to add `int cache_chbegin/cache_chend` parameters to give the precise range of what you want cached? Or is it sufficient (or simply better for you) if it's a simple bool indicating whether to cache all channels together, or exactly the channels you're retrieving with that call to get_pixels? Are those the only two useful choices, or is there merit to caching subsets (example: you're asking for channel 2 only, but you want it to make a cache tile of channels 0-4, but the file itself has 16 channels).

Another alternative is to just make get_pixels do what the TextureSystem functions to: automatically cache all channels together for the normal case of <= 5 channel files, and cache what you ask for for many-channel files. Is that (with no API change) a good solution for you? Or will you routinely have >5 channel files that you still need to get_pixels one channel at a time and don't want the data split across separate cached tiles?

HughMacdonald:
I think that having the ability to pass in a list of channels to cache, but to only return a subset of that would be ideal. I think it would be good to keep it flexible. If cache_chbegin/cache_chend are not specified, then ImageCache would do its own logic about what it thinks it might want to cache, but if they are specified, then we know that the requesting code has made its own decisions, so just go with that.

If chbegin/chend aren't within the cache_chbegin/cache_chend range, then I'm not sure whether you'd want to fail, or just extend the cache channel range to encompass the requested channels.


If you're open to adding deep data support to ImageCache, then that would be great for us - I can't speak for @johnhaddon, but I'd definitely be happy to leave that work at the OIIO end of things.


Thanks

lgritz:
I'm fine with specifying a caching range to IC::get_pixels -- it keeps with the strategy of it retrieving and caching exactly what you ask for and not trying to be too smart under the covers.

No worries on the deep stuff, it certainly needs to be done anyway, and knowing that somebody will use it helps to bump it up in priority in my mind. We're using deep more and more here as well, so it's an investment of time that is directly beneficial on my end. I'll open a separate issue for that so we can discuss how it should work.

HughMacdonald:
Great - we'll definitely be using it - I'm sure I'll get very involved with the discussion! I've spent the last couple of months implementing Deep for Gaffer. Mostly working at the moment, with the image I/O the last major part to do.

Thanks, and I'll keep an eye out for the new issue.

johnhaddon:
Sounds like you two are hashing out something entirely sensible that will definitely work for us, but I'm not 100% convinced we won't end up using bare ImageInputs and our own cache in the long run, for a couple of reasons :

- The natural unit of computation in gaffer is a single tile (of a prescribed size - not necessarily matching the ImageCache) of a single channel, and no matter what happens with the ImageCache API, we can't avoid the buffer copy from the ImageCache into a tile in Gaffer. That copy is expensive enough to have shown up as a fairly large cost in profiling before, so eliminating it is worthwhile.
- All other computed data in Gaffer (including the 3d side) lives in one cache, and I reckon having a single control over the amount of memory used is mostly preferable to having multiple controls.

lgritz:
OK, do whatever you need to, but I still feel like I need to extend IC to handle deep data either way.

Maybe IC will still serve a purpose for you. If you are maintaining a separate cache on your own with a custom tile size that does not match the tiles of the file itself (or the fact that some files are scanline), and caching channels separately, you're still going to run into an efficiency problem even with ImageInput (or for that matter, even if directly using the underlying format reader libraries). It's going to be hard to avoid redundant reads, or else reading more than you need to fill the tiles, if you know what I mean. It may, in the end, still be worth using IC to buffer the parts of the image you need (in file-specified tile sizes) and then doling them out in app-specified-tile sized chunks.

lgritz:
Updated the patch with the new get_pixels parameters.

MrKepzie:
cc me

MrKepzie:
So if I'm not mistaken, reading 4 channels out of a 32 channel file and setting them to cache_chbegin and cache_chend will only read and cache them, but if reading only 4 channels and setting *cache_chbegin* to 0 and *cache_chend* to *spec.nchannels* will actually read all channels and cache all channels.

This can be quite misleading since there seem to be 2 different usages here:

1) call get_pixels on all channels to cache them (cache_chbegin = 0, cache_chend =-1), but only retrieve a sub-portion of them at a time
2) call get_pixels to read and cache only what is requested (the point of the pull request in the first place), i.e: cache_chbegin = chbegin and cache_chend = chend

Also there could be a case where the developer could call ImageCache::get_pixels but some of the channels would not get cached at all! (even if read) which is quite the opposite of what is expected of a cache in the first place.

Shouldn't the first use case be covered by a new function which would actually just cache pixels? It seems that the get_pixels function is trying to do much more than what it is expected to do, that is: read and cache what's requested. 

An alternative which would make it easier to understand would be 2 separate functions:

- *cache_pixels* (..., chbegin, chend) : which explicitly reads and caches the requested channels (but does not fetch them) and would do nothing if they are already cached

- *get_pixels*: which reads the requested channels (and cache them if they did not exist in the cache)

This way, use-case 1) would be:

    cache_pixels(0,-1)

then  for each requested channel

   get_pixels(chbegin,chend)

The second use case would just call get_pixels(chbegin,chend).

I know this idea breaks compatibility with existing code, but since we are already changing the API (though not the behaviour) with cache_chbegin,cache_chend, why not make it clear, or make a second version of it and mark the first one deprecated.

HughMacdonald:
> Also there could be a case where the developer could call ImageCache::get_pixels but some of
> the channels would not get cached at all! (even if read) which is quite the opposite of what is
> expected of a cache in the first place.

This is where I suggested that it would internally expand the cached channels to include the requested channels. So if a developer called get_pixels( chbegin = 0, chend = 4, cache_chbegin = 7, cache_chend = 10 ), then it would actually cache channels 0..10

I'm certainly not averse to having two calls - one to cache extra channels and another to actually get the data, but it does mean that you'd be doubling up on a lot of parameters, rather than having just two extra parameters to the existing get_pixels().

lgritz:
It can't work as two calls. A tile containing chans 0-3 and a tile with the same coordinates containing chans 0-31 are considered different tiles, hash differently, and searching for one will not magically turn up the other. So if you ask it to cache chans 0-31 and then get_pixels of 0-3, it will not know that you previously cached all channels and find that tile. The get_pixels retrieval needs to know what channel range to look for in the cache.

I don't think it's confusing. There are three cases:

1. `get_pixels (..., chbegin, chend, /*default cache_chbegin/chend*/)` : this has behavior identical to how it's always been (convenient for an unchanged API call appearance), which is to look for a tile containing all channels of the file (reading and caching all channels, if not already done), then fill your buffer with just the channels you asked to retrieve. Do this if you don't care, or if your files tend to be a small number of channels, or if you have many channels but know that you'll get around to needing them all eventually.

2. `get_pixels (..., chbegin, chend, ... chbegin, chend)` : look for a tile containing all just the channels you are retrieving now (reading and caching just those channels, if the tile is not yet resident), then fill your buffer with the channels you asked to retrieve. Do this your files tend to be a large number of channels and you only need this subset.

3. `get_pixels (..., chbegin, chend, ... cache_chbegin, cache_chend)` : you may wish to look for (and read/cache if not found) a tile containing a superset of the channels you need at the moment. Do this your files tend to be a large number of channels, you will only need the subset [cache_chbegin,cache_chend], but you are asking for an even more narrow subset at this very moment (but know you'll use the rest of the cached subset eventually).

(1) is the common case, you need all channels. (2) is more memory efficient if you're using a few channels from a many-channel file, for example you will only need channels 12-15 of a 32 channel image.

(3) is a weird case (but apparently one that John and Hugh need) where, for example, you will eventually need channels 12-15 of a 32-channel image, but your code is structured so that you are calling get_pixels for just one channel at a time (chan 12, then later chan 13, then later chan 14, then later chan 15, but no others).

I think the current proposed API is simple and has the nice property that the old parameter sequence (leaving the new params at their default) gives the old behavior, which is also the behavior you want for the common case.

As I currently rigged it, if you give an invalid cache_begin/end range (for example, if the cache begin/end doesn't contain all the channels you are asking to retrieve), it will fall back on case (1) and just read and cache all channels. If you don't want all channels read and cached, it's up to you to make a caching request that makes sense given your expected usage pattern. It will *always* work, but it may not be memory or I/O efficient if you don't know what you're doing.

MrKepzie:
Fair enough, as my edit says, as long as the documentation is clear on that, there might be other developers reaching unexpected behaviour.

johnhaddon:
> (3) is a weird case (but apparently one that John and Hugh need) where, for example, you will eventually need channels 12-15 of a 32-channel image, but your code is structured so that you are calling get_pixels for just one channel at a time (chan 12, then later chan 13, then later chan 14, then later chan 15, but no others).

I was agreeing with everything until this was labelled weird :) If you give users arbitrary control to load and composite images, then you really don't know what channels you need until you execute, and you won't know the full set until you finish. Just as if you give users arbitrary control to sample textures from multiple shaders, you don't know what was actually needed till the render finishes. Ideally in these scenarios you don't want to pay for any more than you ask for, but all you can do right now is hedge your bets by making two guesses : 

1) If the user wants channel N now, what others might they want in the future?
2) If I ask for channel N from the cache, what other channels will the underlying ImageInput be forced to read due to the restrictions of the file format?

Hopefully for 1) it's reasonable to guess "the other channels in the same layer", or "a shader will do all its lookups at once and other shaders will be using different textures, so punt".

For 2), I still believe a method to query ImageInput directly might be valuable. It seems like the ideal cache granularity is the intersection of the channels from 1) and 2), so I guess this query is only useful if there are formats for which the set from 2) is smaller than the set from 1)?

lgritz:
Yes, I was using "weird" to mean "so far, John's request is the only one I'd heard." As you explain it, it is entirely sensible.

How about an ImageInput::supports() query that reveals whether the format supports true channel subset reads?

johnhaddon:
> How about an ImageInput::supports() query that reveals whether the format supports true channel subset reads?

I think the ideal is less a true/false thing and more a "which particular subsets of channels are mutually dependent when loading" thing, which I guess can't be squeezed through the `supports()` API. But I'll have to reveal my ignorance - I don't know enough about the underlying formats to know if this is a useful distinction in practice. A quick read of the EXR docs suggested that for EXR you have to get all channels for an image (or image part) at once anyway? Would I be right in thinking that a multipart file would present itself as multiple sub images via OIIO, and that therefore EXR can't actually read channel subsets independently at all as far as OIIO would be concerned?

lgritz:
I wasn't sure, so I looked more carefully at the OpenEXR spec, and it appears that although the various channels are separated (RRRRGGGGBBBBAAAA), they are compressed as one data block and thus not separately addressable prior to reading and decompression. So it's not possible to read just "G" alone; it has to read the whole scanline (of that "part"), thought the APIs make it easy to only copy certain channels to your app-side buffer.

So there is no I/O advantage to reading a channel subset (or "layer") of an OpenEXR file. The only advantage is in cache storage -- if you are using a tiny subset of the channels in the "part" of the file, you will get better cache performance if you don't cache the channels you don't expect to use.

Note, though, that for OpenEXR 2.0 and beyond, you can have "parts", and each part can be separately read and addressed. So rather than have 32 channels, with the first 4 being RGBA, and the next 4 being the diffuse AOV, and the next 4 being the specular AOV, or whatever, it may be much more efficient (if you intend to need to read only one of those "layers") to store each layer as a different "part": one part for RGBA, the next separate part in the same file for diffuse, etc. That kind of layout (which demands OpenEXR 2.x) WILL have I/O efficiency if you only need one of these AOVs.

For TIFF, there is a choice of "contiguous" planar configuration (RGBARGBARGBA) or "separate" configuration (RRRGGGBBBAAA). If a particular file is stored as separate, then there is indeed I/O savings to be had by requesting only some channels. But if you're going to use all the channels, contig is more efficient, so we always use contig.

I'm not aware of any other formats that we support which allow truly separate reading of channels.

Bottom line is, there is only an I/O advantage to partial channel reads for TIFF, and only if the file uses the packing order that we ordinarily advise against.

I'm still happy to try to implement this check, but I don't think it will really tell you anything useful.

I don't think it's reasonable for OIIO to tell you "which particular subsets of channels are mutually dependent when loading." That's a semantic distinction that depends a lot on what the app intends to do, I don't see how OIIO could be helpful in giving any hints there.

lgritz:
Is everybody ok with this change? Should I merge it into master?

The memory issues being described in #1123 don't appear to be related to this patch, so I see no reason to hold it on account of that. (And, in fact, I also claim that there is no leak, but merely an example of a use case that isn't aligned with the intended design of ImageCache.)

MrKepzie:
Looks good. I have fixed the ImageCache usage on our end, taking your advice.

HughMacdonald:
Looking at the code, it appears that, if the chbegin..chend range crosses outside of cache_chbegin..cache_chend range, it will just cache the full range. Is that right?

I know that that isn't a situation that a user should ever find themselves in, but do you think that it's better to cache the full range rather than just expanding the cache range to encompass chbegin..chend?

lgritz:
The behavior I made was that if the cache range is invalid, it will cache all channels. The alternative is that if the range is invalid, use the requested channels as the cache range (or expand the cache range to include the requested channels?).

I'm not sure there's a strong argument one way or the other, but here's a weak one for my way: If the range is invalid, somebody has made a mistake, and maybe they'll make a mistake on the next call as well, so I think it's better to cache all channels and reduce I/O rather than have each query need to read the tile from disk again. Another weak argument is that this is the case for default arguments (0,-1), and by asking for all channels we are simply preserving the old behavior exactly.

Is there a good argument for why -- again, for a nonsensical cache range that we hope we'll never encounter from an app -- you would expect "expand the range to cover just these channels" is better than "expand the range to cover all channels"?

HughMacdonald:
I think your argument makes perfect sense. As you said, it's an error on the programmer's part if it happens, so best to just go back to the old behaviour.

PR #1227, opened 2015-10-04
oiiotool --fit bug fix when the image doesn't need to be resized.

### LINKED ISSUE #1224: oiiotool: Inconsistent behavior with --fit (bug or feature?)
i've encountered some unexpected behavior with oiiotool's `--fit` command, under a specific circumstance, and I am looking for clarification on whether this is intended behavior or a bug.

When attempting to fit a source image to a target resolution where one dimension is unchanged and the other is larger than the source, padding does not occur and you are given the original source res as output.

Tested against release 1.5.11 and master.

**Steps to reproduce**

```
# basic 1.78 aspect source image
oiiotool --create 2048x1152 3 -o /tmp/width.png

# Fit image to different sizes
oiiotool /tmp/width.jpg --fit:pad=1 2048x871 -o /tmp/out_235.jpg
oiiotool /tmp/width.jpg --fit:pad=1 1920x1080 -o /tmp/out_hd1080.jpg
oiiotool /tmp/width.jpg --fit:pad=1 2048x1536 -o /tmp/out_133.jpg
```

**What I expect**

```
      /tmp/out_235.jpg:  2048x871
/tmp/out_hd1080.jpg:  1920x1080
      /tmp/out_133.jpg:  2048x1536
```

**What I get**

```
      /tmp/out_235.jpg:  2048x871
/tmp/out_hd1080.jpg:  1920x1080
      /tmp/out_133.jpg:  2048x1152
```

An interesting variation on the case that doesn't work as expected:

```
oiiotool /tmp/width.jpg --fit:pad=1 2047x1536 -o /tmp/out_133.jpg
# 2047x1536
oiiotool /tmp/width.jpg --fit:pad=1 2049x1536 -o /tmp/out_133.jpg
# 2049x1536
```

So, if I make sure that both dimensions change, even by one pixel, then it will do the padding to the larger dimensions. The way that I need to work around this, involves using a `--cut` to really pad the image out as expected. This requires needing to provide a proper offset value, and I have found I can only use this approach conditionally when I know that only one target dimension is larger and the other is the same:

```
oiiotool /tmp/width.jpg --fit:pad=1 2048x1536 --cut 2048x1536+0-192 -o /tmp/out_133.jpg
```

lgritz:
Some --fit tasks only need to reposition and pad the image, not actually do a filtered resize. But we had the "reposition" adjustments inside the the "resize" case.

Fixes #1224


lgritz:
Tweaked and added unit test.

My production tests check out ok, so I'll merge this today if nobody objects.

lgritz:
Merged and also backported to RB-1.5.

justinfx:
Thanks for doing this!

lgritz:
No prob! I've integrated into RB-1.5, but I haven't tagged a release. Do you need that right away?

justinfx:
We have locally implemented an edge-case workaround for this bug, so I wouldn't say we are hinging on its release. When it does eventually get tagged, we can then get a build of it and remove this workaround from our project.  Nutshell: No Rush! :-)

PR #1230, opened 2015-10-06
PFM file read: scanline order of pfm is bottom-to-top

### LINKED ISSUE #1170: PFM (part of pnminput) rows should be bottom to top
PFM is an exception in the PNM families: rows are bottom to top, see http://netpbm.sourceforge.net/doc/pfm.html and http://www.pauldebevec.com/Research/HDR/PFM/
PPM and friends are top to bottom, see http://netpbm.sourceforge.net/doc/ppm.html
I don't know how to do this cleanly in pnminput.cpp, do you want me to try submitting a patch?

lgritz:
Since PFM is uncompressed, we merely remember the file position right after the header ends, and when we need a scanline, seek to the right spot.

Fixes #1170

### DIFF
--- a/src/pnm.imageio/pnminput.cpp
+++ b/src/pnm.imageio/pnminput.cpp
@@ -54,13 +54,14 @@ private:
     };
 
     std::ifstream m_file;
+    std::streampos m_header_end_pos; // file position after the header
     std::string m_current_line; ///< Buffer the image pixels
     const char * m_pos;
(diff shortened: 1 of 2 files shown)


lgritz:
Any comments or feedback?

lgritz:
No objections so I'm going to commit this. If there's a problem with it, we can re-open.

PR #1233, opened 2015-10-13
TIFF CMYK support

lgritz:
Needed at my place so that our usual oiiotool-based color correction pipeline can deliver certain marketing assets that need to be CMYK for some reason. 

The refactor of the TIFF code to make all this work was a little more involved than I'd hoped. It's not conceptually hard, but some code got shuffled around.


lgritz:
@danwexler helpfully points me to this important documentation: https://plus.google.com/+IrreverentMonk/posts/eeWT6Dao2sh

PR #1234, opened 2015-10-14
Fix bug in IFF output of 16 bit uncompressed data.

### LINKED ISSUE #933: output of 16-bit uncompressed iff is broken
I found this one when attempting to convert uncompressed 16-bit dpx to iff, but it works with any input:

$ oiiotool colorbars.png -d uint16 --compression none -o bad.iff
$ iv bad.iff

Forcing rle compression or uint8 data (or both) results in good output.

It's definitely something in the output code path as Maya reads bad.iff identically to iv.

Sample image to test with.

colorbars

Definitely low priority since there's an easy workaround (just use --compression rle when generating iff), just posting in case anyone else runs into the same issue.

cheers,
-Mark

lgritz:
Weird construct when copying the 2 bytes of the uint16 value.

Fixes #933

### DIFF
--- a/src/iff.imageio/iffoutput.cpp
+++ b/src/iff.imageio/iffoutput.cpp
@@ -120,7 +120,7 @@ IffOutput::open (const std::string &name, const ImageSpec &spec,
     // currently only RGB RLE compression is supported, we default to RLE
     // as Maya does not handle non-compressed IFF's very well.
     m_iff_header.compression =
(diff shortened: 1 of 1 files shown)


lgritz:
Note: all the interesting discussion of this issue was in #933.

I wish that when an issue and a PR were linked, it would actually merge them. 
Is there a way to do that? Can I submit a PR that addresses an existing issue, without needing to open a second identifying number for the code PR?

PR #1235, opened 2015-10-14
test issue

lgritz:
ignore, I'm trying something


lgritz:
Aha, I successfully turned an "issue" into a "pull request!" But when I did so, I got a message saying this feature is deprecated and might not work in the future. Hmmm.

Issue #1237, opened 2015-10-16
OpenEXR does not claim to support deepdata


HughMacdonald:
I've just realised that OpenEXR doesn't return true for supports( "deepdata" ).

Happy to put in a PR for it myself - it's a quick fix!

Thanks

HughMacdonald:
I thought I'd do a PR for this myself. Found in https://github.com/OpenImageIO/oiio/pull/1238

PR #1238, opened 2015-10-16
OpenExrOutput : Add true return for supports("deepdata")

HughMacdonald:
Fix issue raised in https://github.com/OpenImageIO/oiio/issues/1237


lgritz:
LGTM, thanks for catching this.

lgritz:
Fixes #1237

lgritz:
I also backported to RB-1.5

PR #1240, opened 2015-10-21
Change tile & file limit assertions from ASSERT to DASSERT.

### LINKED ISSUE #1239: Assert failure in ImageCache
I've just found myself getting an assertion failure in ImageCache.

ImageCache should be thread-safe, right? I'm only getting this error when I run the processing in multiple threads in parallel.

Here's the error:

    /Users/john/dev/gafferDependencies.git/oiio-Release-1.5.17/src/libtexture/imagecache.cpp:2097:
        failed assertion 'sweep != m_tilecache.end() && "no way m_tilecache can be empty and use
        too much memory"'

I'm doing this through Gaffer, but what I'm doing is:

* Parallel loading 64x64 tiles from a 4096x3112 image
* Clear the ImageCache
* Repeat another 19 times

It seems as though it mainly happens when I have more than 10 repeats, but I have seen it happen with fewer. I'm guessing it's just chance. It's also more common the more threads I have processing.

Any ideas?

lgritz:
We realize how a particular combination of multithread, low cache memory
limits, and big tiles / untiled can inadvertently trigger these
assertions. Let's turn it into a DEBUG-only assertion for now, so it
won't happen for release builds.

As a more long-term plan, we should probably automatically bump up the
cache limits any time we encounter a tile that is itself bigger than
the memory limit, to avoid hitting this impossible situation.

For detailed discussion, see https://github.com/OpenImageIO/oiio/issues/1239

Fixes #1239

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -1217,7 +1217,7 @@ ImageCacheImpl::check_max_files (ImageCachePerThreadInfo *thread_info)
     // looking up the filename of the first entry in the file cache.
     if (! m_file_sweep_name) {
         FilenameMap::iterator sweep = m_files.begin();
-        ASSERT (sweep != m_files.end() &&
+        DASSERT (sweep != m_files.end() &&
                 "no way m_files can be empty and have too many files open");
         m_file_sweep_name = sweep->first;
     }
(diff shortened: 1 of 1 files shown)


lgritz:
I will backport this short-term fix to 1.5 as well.

fpsunflower:
LGTM

fpsunflower:
I would also be fine with commenting it out -- as mentioned in the discussion, the assert itself is not always valid.

PR #1244, opened 2015-10-27
Separate global "threads" and "exr_threads" attributes.

lgritz:
The global OIIO "threads" attribute controls the default "fan out" of
spawned threads when an OIIO call has work that can be parallelized
(like for ImageBufAlgo functions). Apps with essentially single logic
thread leave this at 0, which means to spawn #cores threads, whereas
apps with lots of app-side threads set it to 1 to prevent OIIO from
internally spawning many threads for each app calling thread.

Unfortunately, we were using the same attribute value to set the size of
OpenEXR's thread pool size. In an app where we set "threads" to 1 to
prevent OIIO fan-out, this was unfortunately also limiting OpenEXR's
worker thread pool to 1, which means the various threads would block
against each other.

It seems like the best solution is to break this apart into separate
attributes: "threads" to control the OIIO fan-out, and "exr_threads"
to control the size of OpenEXR's thread pool. Most apps need not mess
with either of these.

### DIFF
--- a/src/doc/imageioapi.tex
+++ b/src/doc/imageioapi.tex
@@ -805,11 +805,31 @@ attribute is a string), the attribute will not be modified, and {\cf
 \vspace{10pt}
 \index{threads} \label{sec:attribute:threads}
 Some \product operations can be accelerated if allowed to spawn multiple
-threads to parallelize the task.  (Examples: decompressing multiple
-simultaneously-read OpenEXR scanlines, or many \ImageBuf operations.)
-This attribute sets the maximum number of threads that will be spawned.
-The default is 1.  If set to 0, it means that it should use as many
+threads to parallelize the task.  (Examples: simultaneous format conversions
+of multiple scanlines read together, or many \ImageBufAlgo operations.)
+This attribute sets the default number of threads that will be spawned
+for these operations (the ``fan out'').
+The default is 0, which means that it should spawn as many
 threads as there are hardware cores present on the system.
+
+Situations where the main application logic is essentially single threaded
+(i.e., one top-level call into OIIO at a time) should leave this at the
+default value, or some reasonable number of cores, thus allowing lots of
+threads to fill the cores when OIIO has big tasks to complete. But
+situations where you have many threads at the application level, each of
+which is expected to be making separate OIIO calls simultaneously, should
+set this to 1, thus having each calling thread do its own work inside of
+OIIO rather than spawning new threads with a high overall ``fan out.''
+\apiend
+
+\apiitem{int exr_threads}
+\vspace{10pt}
+\index{exr_threads}
+\NEW % 1.6
+Sets the internal OpenEXR thread pool size. The default is to use as many
+threads as the amount of hardware concurrency detected.
+Note that this is separate from the OIIO
+\qkw{threads} attribute.
 \apiend
 
 \apiitem{string plugin_searchpath}
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -316,12 +316,12 @@ void set_exr_threads ()
     static spin_mutex exr_threads_mutex;  
 
     int oiio_threads = 1;
-    OIIO::getattribute ("threads", oiio_threads);
+    OIIO::getattribute ("exr_threads", oiio_threads);
 
     spin_lock lock (exr_threads_mutex);
     if (exr_threads != oiio_threads) {
         exr_threads = oiio_threads;
-        Imf::setGlobalThreadCount (exr_threads == 1 ? 0 : exr_threads);
+        Imf::setGlobalThreadCount (exr_threads);
     }
 }
 
--- a/src/libOpenImageIO/imageio.cpp
+++ b/src/libOpenImageIO/imageio.cpp
@@ -51,6 +51,7 @@ OIIO_NAMESPACE_BEGIN
 namespace pvt {
 recursive_mutex imageio_mutex;
 atomic_int oiio_threads (boost::thread::hardware_concurrency());
+atomic_int oiio_exr_threads (boost::thread::hardware_concurrency());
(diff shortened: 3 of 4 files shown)


src/openexr.imageio/exrinput.cpp
```diff
@@ -316,12 +316,12 @@ void set_exr_threads ()
...
 
     int oiio_threads = 1;
-    OIIO::getattribute ("threads", oiio_threads);
+    OIIO::getattribute ("exr_threads", oiio_threads);
 
     spin_lock lock (exr_threads_mutex);
     if (exr_threads != oiio_threads) {
         exr_threads = oiio_threads;
-        Imf::setGlobalThreadCount (exr_threads == 1 ? 0 : exr_threads);
+        Imf::setGlobalThreadCount (exr_threads);
```
fpsunflower: Wasn't the previous code trying to say that if requesting only one thread -- disable the EXR thread pool instead of launching a single thread?

src/doc/imageioapi.tex
```diff
@@ -805,11 +805,31 @@ \subsection{Miscellaneous Utilities}
...
+which is expected to be making separate OIIO calls simultaneously, should
+set this to 1, thus having each calling thread do its own work inside of
+OIIO rather than spawning new threads with a high overall ``fan out.''
+\apiend
+
+\apiitem{int exr_threads}
+\vspace{10pt}
+\index{exr_threads}
+\NEW % 1.6
+Sets the internal OpenEXR thread pool size. The default is 0, which means
```
fpsunflower: Based on my reading of the code on the EXR side -- 0 is used to disable threading, it doesn't automatically imply the number of cores.

fpsunflower:
Based on what I am reading in the openexr code (so far) if the thread pool is passed 0 threads, the pool is disabled and the decoding should happen on the parent thread, and two threads reading different files should not block each other.

Have I missed something?

src/openexr.imageio/exrinput.cpp
lgritz: Yes, it was trying to say that. But in practice that's not how OpenEXR was behaving.

lgritz:
You have missed the fact that OpenEXR doesn't seem to do what it says. Bug? Mis-design? Mis-understanding? In any case, when we Imf::setGlobalThreadCount(0), one thread ends up making progress while the others block.

It seems that the most flexible and expedient solution (and the only one that would work with pre-fixed OpenEXR, if we were inclined to fix OpenEXR ourselves) is to decouple the "OIIO thread fan-out" from the "IlmImf thread pool size". The defaults are sensible for both, and individual apps can tune each to their heart's content.

fpsunflower:
Ok - I see the problem now. The global thread pool that actually does the work will execute the task immediately when `nthread=0` but will still lock itself (thus preventing any other thread for doing anything).

https://github.com/openexr/openexr/blob/91015147e5a6a1914bcb16b12886aede9e1ed065/IlmBase/IlmThread/IlmThreadPool.cpp#L429

This does seems like a mis-design to me. Seems like something worth bringing up with the EXR team.

fpsunflower:
So I think the only thing to fix in this patch is the documentation for the meaning of `exr_threads=0`.

fpsunflower:
Actually on looking at it again -- it seems that two files can never be read simultaneously (because the whole thread pool is locked as soon as one file is being read). If you have exr_threads>0 you are just speeding up the reading of a single file.

Am I reading that right?

lgritz:
Oy, I don't know. That would sure be unfortunate. The idea was supposed to be that if you set exr threads to 0, it would use the calling thread and not the pool. 

I'm not sure your assessment is true, though I'll look more closely at the exr implementation code to check.

The history is that it all used to work fine and have good performance in itview with our old defaults (setting both OIIO fan-out, and implicitly the exr thread pool size, to the hardware_concurrency). Robert noticed that since he was spawning many threads on the app side, it was producing too many threads, especially for DPX reading which for format conversion ended up calling some parallel OIIO code. I advised that he use OIIO::attribute("threads",1) to limit the fan-out since he himself was managing the number of threads (different threads reading different frames). That solved his problem at the time, but later he noticed that for EXR files specifically, performance took a terrible hit, and he traced it to the fact that threads were blocking against each other.

Since this was not a problem before I told him to set oiio threads to 1, I'm presuming that when we re-decouple the oiio thread fan-out from the exr pool size, we'll have a good compromise.

We may be back here again later (or, inside the exr code) if this patch still doesn't work for him.

fpsunflower:
No sorry - when multiple threads are active, the process of adding tasks locks the pool but the actual execution is "detached" so multiple files should be ok to read concurrently.

But the "0" case does feel like an oversight to me. It also doesn't look like it would be terribly hard to patch.

lgritz:
Why, then, didn't we block when multiple exr files were being read at once, before Robert was using threads=1?

Maybe we DID block, but by using lots of threads, at least the decode went quickly so we didn't block for long?

We can (should?) fix the exr side of things. But still, it will take a while for that patch to be accepted and proliferate to new versions and get installed everyplace people use it. So I think this change is still necessary on our side as a workaround.

lgritz:
Yeah, right? If there are 8 cores, it doesn't really matter if 8 files are being decoded by one thread each, or if one file is being decoded by 8 cores while 7 others take their turns -- same total amount of time, and in both cases approximately 100% core utilization.

The problem is when 7 cores are waiting for 1 core to read a file by itself.

lgritz:
https://github.com/openexr/openexr/blob/91015147e5a6a1914bcb16b12886aede9e1ed065/IlmBase/IlmThread/IlmThreadPool.cpp#L429

I totally agree with your assessment of this code, Chris. Looks like a bug/mis-design in OpenEXR. I have reported it to the openexr-devel list.

lgritz:
I updated the patch with the change in docs, as suggested.

fpsunflower:
LGTM

PR #1246, opened 2015-10-30
Fix data copy error in IC::get_pixels, add unit test for ImageCache

lgritz:
Reported in #1239


lgritz:
@MrKepzie, have you been able to test if this patch fixes the problem you reported in #1239?

MrKepzie:
Will check tomorrow morning (Paris time), was doing refactoring of our code and did not have time to check properly that issue

MrKepzie:
This fixes the issue.

PR #1249, opened 2015-10-31
OpenEXR refactor of translating OIIO standard metadata to EXR names.

### LINKED ISSUE #1242: Errors should always be printed out
Some errors are only printed out when NDEBUG is undefined.  This bit us when we spent a bit of time trying to figure out why the following applied on an EXR was still resulting in a ratio of 1.

    spec.attribute("PixelAspectRatio", 2);

It wasn't until we recompiled OIIO in debug mode that we finally saw an error explaining that:
```
Caught OpenEXR exception: Cannot assign a value of type "int" to image attribute "pixelAspectRatio" of type "float".
Don't know what to do with int pixelAspectRatio
```
One simple solution is to get rid of the NDEBUG checks in these places.  A quick look shows exif.cpp, imageioplugin.cpp, and exroutput.cpp as possible places where fails are likely silently handled (I only confirmed exroutput.cpp, the others might be non-issues).  Another option is to allow these warnings to be printed if we defined some flag at compile time such as DISABLE_SILENT_FAILS

lgritz:
Refactor is an improvement but the real motivation was to ensure that
the OpenEXR standard metadata knew what types they were expecting, and
could more deftly handle type mismatches including conversions when the
intent is obvious (example: passing an int rather than a float for
PixelAspectRatio).

Fixes #1242

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -790,38 +790,67 @@ OpenEXROutput::figure_mip (const ImageSpec &spec, int &nmiplevels,
 
 
 
+struct ExrMeta {
+    const char *oiioname, *exrname;
+    TypeDesc exrtype;
+
+    ExrMeta (const char *oiioname=NULL, const char *exrname=NULL,
+             TypeDesc exrtype=TypeDesc::UNKNOWN)
+        : oiioname(oiioname), exrname(exrname), exrtype(exrtype) {}
+};
+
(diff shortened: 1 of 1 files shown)


lgritz:
@ThiagoIze, does this approach look ok to you -- automatically performing the type conversion when the "standard" OpenEXR metadata is float but the user passes int, and vice versa?

The whole "do we print debug information" is coming shortly as a second patch.

ThiagoIze:
Looks great!

PR #1251, opened 2015-11-04
OIIO::debugmsg() preferred over raw debugging output to stderr or stdout.

lgritz:
It's echoed to stderr if global attribute("debug") is nonzero.
It defaults to 0 for ordinary runs, 1 for DEBUG, and the default is
overridden by he value of environment variable OPENIMAGEIO_DEBUG, if
present.

Suggested in #1242 

It is preferred that OIIO-internal code that's tempted to echo anything to stderr or stdout for debugging purposes (for example, selectively only when not NDEBUG) use this instead.

I have done this in a couple spots, but have not comprehensively scoured the code looking for examples. We can do it lazily, as we are visiting other parts of the code base.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -1281,8 +1281,11 @@ OIIO_API std::string geterror ();
 /// Documented attributes:
 ///     int threads
 ///             How many threads to use for operations that can be sped
-///             by spawning threads (default=1; note that 0 means "as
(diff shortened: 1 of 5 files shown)


lgritz:
Updated the PR with a name change -- debugmsg() instead of debug() will create fewer name clashes with apps that have 'debug' as a symbol and also might be inclined to have 'using namespace OIIO'.

PR #1252, opened 2015-11-05
Bug fix in special pole handling for env map lookups + tile channel subsets

lgritz:
Bug fix in special pole handling for env map lookups + tile channel subsets.

Oy, remember that recent ImageCache change where tiles may hold channel subsets for many-chaneled files? It doesn't play well with the particular way that pole_color walked through the tile's pixel memory to compute the pole color, and in fact to get the pole color right we need all the channels (because we want to do it just once per MIP level).

The simplest way of handling this is just to NOT do the tile channel constriction when dealing with environment maps, at the one-tile MIP levels (the only time that the pole color facing happens).

The only way this would come up is when doing environment map lookups right smack at the poles, only when blurred enough that the MIP level fits on one tile, and only for environment maps with more than 5 channels.  So of course, we had one of those. The incorrect pointer arithmetic from mistaking the number of channels in the file for the number stored in the cached tile caused us to read some uninitialized memory, in turn occasionally generating a NaN, which, wouldn't you know, affected some probability sampling for IBL. Sheesh!

### DIFF
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -1620,24 +1620,26 @@ TextureSystemImpl::pole_color (TextureFile &texturefile,
     if (! levelinfo.onetile)
         return NULL;   // Only compute color for one-tile MIP levels
     const ImageSpec &spec (levelinfo.spec);
-    size_t pixelsize = texturefile.pixelsize(subimage);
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

PR #1253, opened 2015-11-06
Fix typo in Tiff plugin documentation

HughMacdonald:
I spotted a typo in the documentation for the Tiff plugin, so this fixes it...


lgritz:
Thanks for the fix

PR #1256, opened 2015-11-09
Fix recent exr metadata break

lgritz:
A few checkins ago, we refactored some EXR metadata output, making a
table.  Well, turns out that this can be subject to the "initialization
order fiasco", because the static table in exroutput.cpp referenced the
static global variables TypeDesc::TypeMatrix (and several others) that
are defined in typedesc.cpp, and thus may not yet have been initialized.

Why, exactly, the fairly trivial constructors don't automatically bake
down into POD that can be statically initialized in the DSO, I don't
understand.

But anyway, slight modification of the table to not depend on any prior
initializations fixes it.

Also put in automatic float[16]-to-matix44 translation, though it turns
out that wasn't the problem after all, and beefed up a debug message.

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -799,22 +799,28 @@ struct ExrMeta {
         : oiioname(oiioname), exrname(exrname), exrtype(exrtype) {}
 };
 
+// Make our own type here because TypeDesc::TypeMatrix and friends may
+// not have been constructed yet. Initialization order fiasco, ick!
+static TypeDesc TypeMatrix (TypeDesc::FLOAT,TypeDesc::MATRIX44,0);
+static TypeDesc TypeTimeCode (TypeDesc::UINT, TypeDesc::SCALAR, TypeDesc::TIMECODE, 2);
+static TypeDesc TypeKeyCode (TypeDesc::INT, TypeDesc::SCALAR, TypeDesc::KEYCODE, 7);
+
 static ExrMeta exr_meta_translation[] = {
     // Translate OIIO standard metadata names to OpenEXR standard names
-    ExrMeta ("worldtocamera", "worldToCamera", TypeDesc::TypeMatrix),
-    ExrMeta ("worldtoscreen", "worldToNDC", TypeDesc::TypeMatrix),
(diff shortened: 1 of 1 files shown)


src/openexr.imageio/exroutput.cpp
```diff
@@ -799,22 +799,28 @@ struct ExrMeta {
...
-    ExrMeta ("XResolution", "xDensity", TypeDesc::TypeFloat),
-    ExrMeta ("ExposureTime", "expTime", TypeDesc::TypeFloat),
-    ExrMeta ("FNumber", "aperture", TypeDesc::TypeFloat),
-    ExrMeta ("oiio:subimagename", "name", TypeDesc::TypeString),
-    ExrMeta ("openexr:dwaCompressionLevel", "dwaCompressionLevel", TypeDesc::TypeFloat),
-    ExrMeta ("smpte:TimeCode", "timeCode", TypeDesc::TypeTimeCode),
-    ExrMeta ("smpte:KeyCode", "keyCode", TypeDesc::TypeKeyCode),
+    ExrMeta ("worldtocamera", "worldToCamera", TypeMatrix),
+    ExrMeta ("worldtoscreen", "worldToNDC", TypeMatrix),
+    ExrMeta ("DateTime", "capDate", TypeDesc::STRING),
```
fpsunflower: What makes the `STRING` case safe but not the `TypeMatrix` case?

Its a shame we don't have `constexpr` available yet ...

lgritz: TypeDesc::STRING is just an enum (i.e. an int) for the "base type".

TypeDesc(TypeDesc::STRING) works just fine, assembles the full TypeDesc on the fly (base type STRING, scalar, no array).

TypeDesc::TypeString is a full TypeDesc, constructed in typedesc.cpp. Or, in this case, not yet constructed.

Yes, I think in a C++11 world, TypeString (etc) will be constexpr fully defined in typedesc.h, and not globals initialized in typedesc.cpp.

lgritz: I cannot wait for the switch to C++11 as minimum baseline.

fpsunflower: Oh I see now, I was confused by the silent cast.

LGTM!

PR #1259, opened 2015-11-12
ImageInput/ImageOutput API change to allow thread policy per file.

lgritz:
Please see also #1258, this is an alternate proposal for the same basic issue. I think this one is far less intrusive to the APIs, and am leaning toward this if nobody strongly prefers the other.

----

This allows control over thread spawning for any ImageInput or
ImageOutput internals that may have the opportunity to parallelize work,
similar to the parameter on the various ImageBufAlgo functions. A value
of 0 means to use the global OIIO::attribute("nthreads") value (which in
turn defaults to have a thread fan-out equal to the hardware
concurrency). For a sole thread of execution, you want a big fan-out for
tasks that can be parallelized, whereas when the calling function is one
of many concurrent threads already underway, you just want the work done
directly by the calling thread, without spawning additional threads that
will oversubscribe the hardware.

The reason that we want per-ImageInput (or ImageOutput) overrides rather
than relying solely on the global "nthreads" attribute is that OIIO may
be used by multiple parts of an application (or multiple dependent
libraries or plugins, each unaware of the others), and so it is not
particularly safe under those circumstances to set the global value and
assume that it will stay that way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ project (OpenImageIO)
 # Release version of the library
 set (OIIO_VERSION_MAJOR 1)
 set (OIIO_VERSION_MINOR 6)
-set (OIIO_VERSION_PATCH 5)
+set (OIIO_VERSION_PATCH 6)
 set (OIIO_VERSION_RELEASE_TYPE "dev")   # "dev", "betaX", "RCY", ""
 
 cmake_minimum_required (VERSION 2.6)
--- a/src/doc/imagebuf.tex
+++ b/src/doc/imagebuf.tex
@@ -437,6 +437,21 @@ have correct knowledge of its pixel memory layout.  USE WITH EXTREME
 CAUTION.
 \apiend
 
+\apiitem{void {\ce threads} (int n) \\
+int {\ce threads} () const}
+\index{threads}
+\NEW % 1.6
+Get or set the threading policy for this \ImageBuf, controlling the
+maximum amount of parallelizing thread ``fan-out'' that might occur during
+expensive operations. The default of 0 means that the global
+{\cf attribute("threads")} value should be used (which itself defaults to
+using as many threads as cores; see Section~\ref{sec:attribute:threads}).
+
+The main reason to change this value is to set it to 1 to indicate that the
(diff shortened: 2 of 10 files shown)


lgritz:
OK, I've abandoned #1258 and decided this approach is better (leave all the API calls as they were except for adding a threads() method to set or get the value).

I have amended this PR to be more complete, change the name from nthreads to threads (to match the global attribute, which was called "threads" all along), and also add a similar facility to ImageBuf (which passes correctly to any underlying ImageInput or ImageOutput).

erictabellion:
Arriving after the battle with only a few things to add to the conversation:

- The OpenExr library already has its own thread-pool to deal with threading .exr loading and de-compression. OpenExr's thread-pool is currently configured by the OIIO exr plugin and is being set according to the global OIIO "threads" attribute (search for ``Imf::setGlobalThreadCount`` in the OIIO coebase). This is probably sub-optimal when both aforementioned main use cases are happening within the same app: single or lightly threaded image input client with a desired wide thread fan-out (i.e. loading whole environment-maps to compute CDFs for importance sampling), followed by heavily threaded texture system client with no thread fan-out desired (i.e. rendering). I would expect the exr plugin to also change to configure the exr thread pool according to the context in which the exr image is being opened.

- Personally I wouldn't mind seeing OIIO rely on TBB on the long run for its thread-pooling needs. As it has been discussed here and in #1258, it's probably a heavy hammer for addressing the particular problem of dealing well with the 2 main use cases so this PR looks good to me!

Thanks for requesting feedback!

lgritz:
The fact that TBB is among those listed in http://vfxplatform.com/ gives me some confidence that it's getting mainstream and reasonable to expect it to be present for installs on desktop systems (at least in the VFX world). The biggest problem with TBB is that it's Intel only, so every place we use TBB needs some kind of reasonable fallback on non-Intel platforms (mobile).

Good point about the EXR threading -- the patch I merged doesn't try to pass on the per-file thread instructions to the IlmImf internals. Perhaps it should, but based on the recent bugs I found in their thread pool, I somewhat dreaded the task of verifying that their per-file thread counts were was working properly. So for this moment, I punted, but it's probably something to come back to.

PR #1262, opened 2015-11-15
C++11 cleanup

lgritz:
Some C++11 related cleanup.

The most important item is that when C++11 is detected, std::unordered_map will always be used rather than boost::unordered_map.

The rest is mostly changes to comments or minor rearrangements of headers.

### DIFF
--- a/src/include/OpenImageIO/array_view.h
+++ b/src/include/OpenImageIO/array_view.h
@@ -35,24 +35,24 @@
 #include <stdexcept>
 #include <iostream>
 
+#if OIIO_CPLUSPLUS_VERSION >= 11
+# include <initializer_list>
+# include <type_traits>
+#else /* FIXME(C++11): this case can go away when C++11 is our minimum */
+# include <boost/type_traits.hpp>
+#endif
+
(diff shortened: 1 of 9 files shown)


fpsunflower:
`std::unordered_map` is also available pre-C++11 on a bunch of platforms, do you want to include that case too?

lgritz:
Chris: Nah, we're so close to C++11 as a baseline, I'm not inclined to change the "not C++11" case, other than to make sure it continues compiling as it always has.

But for the C++11 case, I am trying to make it as "clean" as possible, i.e. use std over boost every place where it's not hard to do, with the goal of having no boost referenced in *public* APIs or by client apps. We still have a number of facilities that need it internally, and that's ok, because it's much less of a headache for users when it's part of the OIIO library build, than when it's part of their app's build.

lgritz:
This is the first in a series of changes to accomplish this "no boost in public headers" goal. I have a bunch more teed up but am testing and reviewing one at a time to keep the mess to a minimum.

fpsunflower:
Makes sense. LGTM

PR #1265, opened 2015-11-17
Added MATRIX33 aggregate to TypeDesc. Used in OpenEXR input/output

Shootfast:
Added a new `MATRIX33` to `TypeDesc`, and made the existing `TypeMatrix` an alias for `TypeMatrix44`.

Also changed the c_str representation of matrix to always return the explicit `matrix33` or `matrix44`, though left the fromstring behaviour the same.

### DIFF
--- a/src/libutil/typedesc.cpp
+++ b/src/libutil/typedesc.cpp
@@ -175,8 +175,6 @@ TypeDesc::c_str () const
     std::string result;
     if (aggregate == SCALAR)
         result = basetype_name[basetype];
-    else if (aggregate == MATRIX44 && basetype == FLOAT)
-        result = "matrix";
     else if (aggregate == VEC4 && basetype == FLOAT && vecsemantics == NOXFORM)
         result = "float4";
     else if (vecsemantics == NOXFORM) {
@@ -185,6 +183,7 @@ TypeDesc::c_str () const
         case VEC2 : agg = "vec2"; break;
         case VEC3 : agg = "vec3"; break;
         case VEC4 : agg = "vec4"; break;
+        case MATRIX33 : agg = "matrix33"; break;
         case MATRIX44 : agg = "matrix44"; break;
         }
         result = std::string (agg) + basetype_code[basetype];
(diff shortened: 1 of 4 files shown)


src/libutil/typedesc.cpp
```diff
@@ -175,8 +175,6 @@ TypeDesc::c_str () const
     std::string result;
     if (aggregate == SCALAR)
         result = basetype_name[basetype];
-    else if (aggregate == MATRIX44 && basetype == FLOAT)
-        result = "matrix";
```
lgritz: I'm certain that this change is going to break OSL.

lgritz:
This all LGTM except for my reservations about apps that assume that TypeMatrix.c_str() returns "matrix". I need to see how much needs to be fixed on the OSL side, and consider if anybody else would be affected.

src/libutil/typedesc.cpp
Shootfast: Yeah, I left the `fromstring` method so that "matrix" still converts to `matrix44`, but felt that it was better to be specific for the `c_str` method. There was also the special case for "float4" which, although I left alone, I was curious as to why we have the special cases at all (though I assume it's also an OSL thing).

lgritz: OIIO is a dependency for OSL, not the other way around, so a few classes that are needed for both live in OIIO but may have some particular design quirks that came directly from imagining how it would be used in a shading system or renderer.

For c_str(), I would like to make a pitch for keeping "matrix" to mean the 'usual' case of 4x4 transformation matrices used in computer graphics, and only need the numbers for other cases (3x3 is the only other one we have planned at the moment). I think that's similar in spirit to how "vector" is understood to be the usual case of vector3, but we use "vector2" or "vector4"

lgritz: You know, now that I'm looking at TypeDesc::from_string versus TypeDesc::c_str(), I'm seeing some inconsistencies. Let me take a closer look at this issue and see what can be changed without breaking much in OSL. Give me a little time and I'll report back here. We may be ripe for some cleanup, so I'm taking a hard line on anything quite yet.

lgritz:
See #1267 for my alternate proposal, which is mostly @Shootfast's work, with just a couple minor additions and reverting the "matrix" vs "matrix44" behavior change.

lgritz:
Mark, do you have opinions on #1267? Is that ok with you?

lgritz:
@Shootfast

lgritz:
Subsumed by #1267

PR #1266, opened 2015-11-18
C++11 use boost alternatives (mainly std)

lgritz:
A series of changes with the aim of using only std, no boost, in public headers. For the most part, these only come into play with compiling with USE_CPP11=1. C++03 compiles still fall back to the boost versions.

There were three main sets of changes:

* thread.h: use std::mutex, recursive_mutex, lock_guard, thread, and make a simple thread_group, all to take the place of boost components of the same names. Also remove mention of thread_specific_ptr from thread.h, use the boost one directly from the couple of specific places in internals where we need it.

* refcnt.h handles all the reference-counted pointers -- use std::shared_ptr instead of boost, and make our own intrusive_ptr.

* imagebufalgo_util.h now uses the OIIO-namespaced thread_group, bind, ref, cref (which are std or our own for C++11, boost for C++03).

I believe that this gets us to a state where, when using C++11, there are NO references to boost headers or types in OIIO's public headers, and thus no need to expose them in client application code that uses OIIO. We still use some boost internally at OIIO build time, and that's fine. Yet more of that will go away once we eventually move to C++11 as a baseline (after our next major release branch).

None of this is resulting from any antipathy toward boost. Like I said, there are uses of boost in our internals that we have no particular intent to remove (and no obvious replacements). But there is a benefit to keeping it out of the public APIs, primarily to avoid versionitis issues in apps that may require multiple dependent libraries that each use or expect different Boost versions. We strive for OIIO to not add to those headaches, unless there is no alternative. With C++11, a whole lot of alternatives crop up, as a sizeable portion of the bits of boost we use have migrated to be part of the standard library.

### DIFF
--- a/src/include/OpenImageIO/refcnt.h
+++ b/src/include/OpenImageIO/refcnt.h
@@ -39,18 +39,105 @@
 #ifndef OPENIMAGEIO_REFCNT_H
 #define OPENIMAGEIO_REFCNT_H
 
-#include "thread.h"
-#include "oiioversion.h"
+#include <OpenImageIO/thread.h>
 
-// Use Boost for shared pointers
-#include <boost/shared_ptr.hpp>
-#include <boost/intrusive_ptr.hpp>
+#if OIIO_CPLUSPLUS_VERSION >= 11
+# include <memory>
+#else
+# include <boost/shared_ptr.hpp>
+#endif
+
 
 
 OIIO_NAMESPACE_BEGIN
 
+#if OIIO_CPLUSPLUS_VERSION < 11
 using boost::shared_ptr;
-using boost::intrusive_ptr;
+#else
+using std::shared_ptr;
+#endif
+
+
+
+/// A simple intrusive pointer, modeled after std::shared_ptr and
+/// boost::intrusive_ptr.
+template<class T>
+class intrusive_ptr
+{
+public:
+    typedef T element_type;
+
+    /// Default ctr
+    intrusive_ptr () OIIO_NOEXCEPT : m_ptr(NULL) { }
+
+    /// Construct from a raw pointer (presumed to be just now allocated,
+    /// and now owned by us).
+    intrusive_ptr (T *ptr) : m_ptr(ptr) {
+        if (m_ptr) intrusive_ptr_add_ref(m_ptr);
+    }
+
+    /// Construct from another intrusive_ptr.
+    intrusive_ptr (const intrusive_ptr &r) : m_ptr(r.get()) {
+        if (m_ptr) intrusive_ptr_add_ref (m_ptr);
+    }
+
+#if OIIO_CPLUSPLUS_VERSION >= 11
+    /// Move construct from another intrusive_ptr.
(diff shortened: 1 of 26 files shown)


src/include/OpenImageIO/refcnt.h
```diff
@@ -39,18 +39,105 @@
...
+    /// Assign from intrusive_ptr
+    intrusive_ptr & operator= (const intrusive_ptr &r) {
+        intrusive_ptr(r).swap (*this);
+        return *this;
+    }
+
+#if OIIO_CPLUSPLUS_VERSION >= 11
+    /// Move assignment from intrusive_ptr
+    intrusive_ptr & operator= (intrusive_ptr&& r) OIIO_NOEXCEPT {
+        intrusive_ptr (static_cast<intrusive_ptr&&>(r)).swap(*this);
```
fpsunflower: Why is this static cast needed? It looks like you are casting to the same type it already is ...

src/libOpenImageIO/imagebufalgo.cpp
```diff
@@ -927,7 +926,7 @@ ImageBufAlgo::fillholes_pushpull (ImageBuf &dst, const ImageBuf &src,
     topspec.set_format (TypeDesc::FLOAT);
     ImageBuf *top = new ImageBuf (topspec);
     paste (*top, topspec.x, topspec.y, topspec.z, 0, src);
-    pyramid.push_back (boost::shared_ptr<ImageBuf>(top));
+    pyramid.push_back (OIIO::shared_ptr<ImageBuf>(top));
```
fpsunflower: Unrelated to this particular change -- but maybe we should use `make_shared` instead of `new`?

fpsunflower:
LGTM - I think this will help a lot (though it would be nice if the last remaining internal uses could go away too ....)

src/include/OpenImageIO/refcnt.h
lgritz: To be honest, being shaky on move semantics myself, I modeled that after what boost did for their intrusive_ptr's. I don't know if it's strictly necessary, but I wasn't sure how to determine whether it was superfluous or not.

src/libOpenImageIO/imagebufalgo.cpp
lgritz: Yes, eventually. Can't be done yet, without either a proliferation of "#if" or making a solid C++03 implementation of make_shared.

lgritz:
For now, the goal is merely getting boost out of the public headers. That could be achieved to help those who want it now (mostly to keep the fumbling with boost out of their apps that use OIIO) with a minimum of changes and not a lot of extra #if's.

Trying to use every C++11 idiom now doesn't seem wise, it would just clutter the code with extra '#if' cases. When we branch and move master to C++11 minimum, there will certainly be another big refactor to update all our idioms (like using make_shared and unique_ptr rather than raw new and delete) and further reduce internal boost use (for example, boost::regex can become std::regex).

PR #1267, opened 2015-11-19
Alternative to #1265 openexr 3x3 matrix

lgritz:
Please see #1265. GitHub doesn't let me propose changes on the same ticket, grr. This is 99% the same as @shootfast's proposal, but I have

1. Added the Python binding changes he forgot.

2. Modified the docs appropriately.

3. Reverted a tiny bit so that the string representation of 4x4 matrices remains "matrix" rather than changing it to "matrix44". I'm proposing that we use "matrix33" is for 3x3, but stay with plain "matrix" for the usual common CG case of 4x4, for the same reason that "point" is the common CG case without needing to be called "point3"; and this way, we don't introduce any compatibility breaks for apps that expect the current behavior (especially for OSL).

If everybody is ok with this, before merging I will squash it down to just two commits: (a) all the changes to TypeDesc, together; and (b) the changes to OpenEXR. But for review purposes, it may be clearer to be able to see exactly what changes I made atop Mark's original proposal.

### DIFF
--- a/src/include/OpenImageIO/typedesc.h
+++ b/src/include/OpenImageIO/typedesc.h
@@ -85,7 +85,7 @@ struct OIIO_API TypeDesc {
                     HALF, FLOAT, DOUBLE, STRING, PTR, LASTBASE };
     /// AGGREGATE describes whether our type is a simple scalar of
     /// one of the BASETYPE's, or one of several simple aggregates.
-    enum AGGREGATE { SCALAR=1, VEC2=2, VEC3=3, VEC4=4, MATRIX44=16 };
+    enum AGGREGATE { SCALAR=1, VEC2=2, VEC3=3, VEC4=4, MATRIX33=9, MATRIX44=16 };
     /// VECSEMANTICS gives hints about what the data represent (for
     /// example, if a spatial vector, whether it should transform as
     /// a point, direction vector, or surface normal).
@@ -295,6 +295,8 @@ struct OIIO_API TypeDesc {
     static const TypeDesc TypeVector;
     static const TypeDesc TypeNormal;
     static const TypeDesc TypeMatrix;
+    static const TypeDesc TypeMatrix33;
+    static const TypeDesc TypeMatrix44;
     static const TypeDesc TypeTimeCode;
     static const TypeDesc TypeKeyCode;
     static const TypeDesc TypeFloat4;
--- a/src/doc/imageioapi.tex
+++ b/src/doc/imageioapi.tex
@@ -89,6 +89,7 @@ by the enumerated type {\cf AGGREGATE}:
 {\cf VEC2} & two values representing a 2D vector. \\
 {\cf VEC3} & three values representing a 3D vector. \\
 {\cf VEC4} & four values representing a 4D vector. \\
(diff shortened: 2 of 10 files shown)


src/include/OpenImageIO/typedesc.h
```diff
@@ -85,7 +85,7 @@ struct OIIO_API TypeDesc {
                     HALF, FLOAT, DOUBLE, STRING, PTR, LASTBASE };
     /// AGGREGATE describes whether our type is a simple scalar of
     /// one of the BASETYPE's, or one of several simple aggregates.
-    enum AGGREGATE { SCALAR=1, VEC2=2, VEC3=3, VEC4=4, MATRIX44=16 };
+    enum AGGREGATE { SCALAR=1, VEC2=2, VEC3=3, VEC4=4, MATRIX33=9, MATRIX44=16 };
```
fpsunflower: Hopefully we'll never need `MATRIX22` ...

src/include/OpenImageIO/typedesc.h
```diff
@@ -295,6 +295,8 @@ struct OIIO_API TypeDesc {
     static const TypeDesc TypeVector;
     static const TypeDesc TypeNormal;
     static const TypeDesc TypeMatrix;
+    static const TypeDesc TypeMatrix33;
+    static const TypeDesc TypeMatrix44;
```
fpsunflower: Is it worth adding `TypeMatrix44` as an alias of `TypeMatrix`? (There is no `TypePoint3` for example).

src/include/OpenImageIO/typedesc.h
lgritz: Hopefully. :-)

I mean, we could add another tag for MATRIX22, but it would break the current scheme of having the AGGREGATE enum's numeric representation be the number of scalars, it makes the computation of the total size of the type really simple without extra conditionals or table lookups.

There were many possible ways to organize TypeDesc's internal layout. At some point it was an arbitrary choice. I'm digressing, but another one I sometimes wish I'd done would be to separate the base type from one enum into orthogonal "int or float", "signed or unsigned", and "number of bits".

src/include/OpenImageIO/typedesc.h
lgritz: I don't have strong feeling one way or the other. I can take it out and remove the redundancy. I agree that it doesn't add anything.

After the final switch to C++11, these will get changed from static const to constexpr anyway.

fpsunflower:
Other than nitpicking the addition of `TypeMatrix44` (which seems confusing if we are calling them `matrix` and `matrix33` in the serialized representation), LGTM.

lgritz:
@Shootfast do you have any opinion on this change?

lgritz:
@Shootfast last chance to comment -- I will merge this if I don't hear from you today, because I am eager to branch a RB-1.6 to put it into beta, and want core classes like TypeDesc to freeze afterwards, if possible. If you want to plead the case for changing TypeMatrix.c_str() from "matrix" to "matrix44", speak now.

Shootfast:
Sorry Larry, forgot to reply to this. Yes, all looks good from my end! Thanks!

PR #1268, opened 2015-11-21
Travis setup

lgritz:
This is an initial setup. At this moment, it only builds for OSX using clang.  Just getting my feet wet with it, but eventually we'll flesh it out to a more extensive build matrix.

I've still got lots of bells and whistles to add, and there's some debug scaffolding. 

Just a first stab, but it's already becoming kind of useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+language: cpp
+sudo: false
+compiler: clang
+#    - clang
+#    - gcc
+
+os: osx
+#    - linux
+#    - osx
+
+cache:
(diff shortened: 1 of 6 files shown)


lgritz:
If there are no objections, I'm going to merge this today, and then continue to refine as we move forward. Obvious next steps are to add Linux, clang vs gcc, C++03 vs C++11.

But even now with just one build platform, it's super cool to go to GH to look at a pull request and have it displayed right there that the PR has already been built cleanly and passes basic tests.

PR #1269, opened 2015-11-23
More Travis work -- now builds Linux+gcc and also OSX+clang

lgritz:
Also a variety of other fixes and refactors of how I organize the Travis.


lgritz:
Added more: now build default (C++03, SSE2), also build with SSE4.2, also build C++11.
I think that this mostly wraps up the Travis CI work for now.

.travis.yml
```diff
@@ -1,33 +1,66 @@
...
 # before_script:
 
-script: make VERBOSE=1 cmakesetup && make -j 2 && make test
+# Strategy: make a full default build + test (default to sse2 and C++03)
+# Then just check that the build works fine for sse4.2.
+# And finally, do a build and test with C++11.
+script:
+    - make VERBOSE=1 cmakesetup && make -j 2 && make test
+    - make nuke && make -j 2 USE_SIMD=sse4.2 && make test
+    - make nuke && make -j 2 USE_CPP11=1 && make test
```
danwexler: For each platform you run all three test sets?

.travis.yml
```diff
@@ -1,33 +1,66 @@
...
+      - gcc-4.8
+      - g++-4.8
+      - libboost1.55-all-dev
+      - libilmbase-dev
+      - libopenexr-dev
+      - libgif-dev
+      - libtiff4-dev
+      - libopenjpeg-dev
+      - libwebp-dev
+      - ffmpeg
```
danwexler: What determines if you list something in sources, packages, or in the install_homebrew_deps.bash?

danwexler:
LGTM, assuming the checks pass.

.travis.yml
lgritz: I couldn't figure out how to add new rows to the matrix for a category they don't already have (their docs are not clear on how to use entirely different sets of compiler flags).

But most importantly, each build job is completely independent, and involves installation of all the dependent packages (including building IlmBase and OpenEXR from scratch!), so this way I am amortizing that cost over several builds that differ only in compiler flags on the OIIO side. This makes the whole set build faster, and also keeps me well within the compute resources they allocate to free accounts for open source.

The only down side is, if it fails, you have to go to the site and look at the logs carefully enough to figure out which of the three builds in the job is the one that failed, e.g. does it only fail for C++11, or what?

.travis.yml
lgritz: The install_homebrew_deps.bash is OSX only.

This bit here here is Linux only.

The "apt:" category is for a Linux-side plugin that runs apt-get on the packages you list. You have to do it that way if you use "containers", which is encouraged, because containers don't let you 'sudo' which is ordinarily required for apt-get. Sources means servers to apt-get from. Packages is which packages to install.

lgritz: Don't think for a minute that the docs were clear about any of this!

It took a lot of hunting through docs (which are in many cases outdated), reading their blog posts that explain things not spelled out in the docs (especially for recent changes), and hunting around the net for other C++-based projects that use Travis and seeing what their .travis.yml files looked like.

And I've been through 98 builds so far, figuring this out by trial and error!

.travis.yml
Nava2: Pardon the ignorance, but what about using [ENV variables](https://docs.travis-ci.com/user/speeding-up-the-build/) for the flags? 

For example, 
```
env:
  - BUILD_FLAGS=""
  - BUILD_FLAGS="USE_SIMD=sse4.2"
  - BUILD_FLAGS="USE_CPP11=1"

script:
  - make VERBOSE=1 cmakesetup
  - make -j2 $BUILD_FLAGS 
  - make test
```

This would both grow the build matrix by an order of 3, and it would also run the containers in parallel. It would need to be tested by the joys of trial and error, but should work.

lgritz: Yeah, good idea, I think that would do it.

The only down side is, each one needs to install/build the dependencies independently. Maybe I shouldn't worry about that until the whole thing reaches some limit.

Thanks for the tip!

Nava2: You're correct about the dependencies. Unfortunately, [Amdahl's Law](https://en.wikipedia.org/wiki/Amdahl's_law) springs to mind here. If the cost of dependencies far outweighs the test/build over-head, this wouldn't improve anything. 

However, if you're running `make nuke` everytime, isn't the build rebuilding all of the internal dependencies anyway? For things like OpenEXR 2.*, here is a suggestion:

```yml
env:
  - BUILD_TYPE="default"
  - BUILD_TYPE="cpp11"
  - BUILD_TYPE="simd"
```

Within the build OpenEXR script, instead of installing it into `${HOME}/openexr-install`, installing it into `.ccache/${BUILD_TYPE}/openexr-install`. When the script runs, run `make install` against the `.ccache/` directory instead of building it from scratch. The `make install` should override whichever version the container had. 

I'm just spit-balling, I'd suggest going with what you have and perhaps keeping these ideas as a future improvement! If I'm not being helpful, please let me know.

lgritz: That worked great! Amending the ticket to include this method.

PR #1277, opened 2015-12-04
Deep OpenEXR: recognize newer AR/AG/AB channel name convention.

brechtvl:
In the latest [OpenEXR documentation](http://openexr.com/TechnicalIntroduction.pdf), it is recommended to use AR/AG/AB channel names for RGB alpha, instead of RA/GA/BA. The latter was recommended in an earlier version of the documentation, and still recognized in this patch for compatibility.


lgritz:
LGTM, thanks for the fix.
Are you on master, or do you need a backport?

brechtvl:
For me it doesn't need to be backported, there's nothing depending on this fix. It's just something I noticed when using `oiiotool --flatten`.

PR #1278, opened 2015-12-04
TIFF read_scanlines of images with unassociated alpha didn't honor stride

lgritz:
This could run off the end of a buffer when reading images with
unassociated alpha, if the calling app passed an unusual ystride (or,
in the case I found, using a negative stride to flip the process
in the process of reading it).


fpsunflower:
LGTM

PR #1279, opened 2015-12-04
Minor platform.h related fixes

lgritz:
(Note: I just found this lurking in my tree, something I did many weeks ago and never submitted.)

* Make sure LIKELY/UNLIKELY are defined for clang and Intel.

* More straightforward way to generate bools for LIKELY/UNLIKELY.

* Use OIIO_UNLIKELY in the ASSERT macros, since by definition we
  aren't expecting the condition to usually be true.

* Feature-testing macro definitions for compilers that lack them
  (and use them).

* More robust test for including windows.h.

* Remove redundant definitions from thread.h and hashes.cpp.

### DIFF
--- a/src/include/OpenImageIO/dassert.h
+++ b/src/include/OpenImageIO/dassert.h
@@ -35,6 +35,8 @@
 #include <cstdio>
 #include <cstdlib>
 
+#include "platform.h"
+
 
 /// \file
 ///
@@ -68,7 +70,7 @@
 
 #ifndef ASSERT
 # define ASSERT(x)                                              \
-    ((x) ? ((void)0)                                            \
+    (OIIO_LIKELY(x) ? ((void)0)                                 \
          : (fprintf (stderr, "%s:%u: failed assertion '%s'\n",  \
                      __FILE__, __LINE__, #x), abort()))
 #endif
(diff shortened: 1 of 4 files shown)


src/include/OpenImageIO/dassert.h
```diff
@@ -68,7 +70,7 @@
 
 #ifndef ASSERT
 # define ASSERT(x)                                              \
-    ((x) ? ((void)0)                                            \
+    (OIIO_UNLIKELY(x) ? ((void)0)                               \
```
fpsunflower: Shouldn't asserts be "likely"?
We expect them to always be true?

lgritz: Oh yeah, I got it backwards. 



On December 4, 2015 12:41:01 PM PST, Christopher Kulla <notifications@github.com> wrote:
>> @@ -68,7 +70,7 @@
>>  
>>  #ifndef ASSERT
>>  # define ASSERT(x)                                              \
>> -    ((x) ? ((void)0)                                            \
>> +    (OIIO_UNLIKELY(x) ? ((void)0)                               \
>
>Shouldn't asserts be "likely"?
>We expect them to always be true?
>
>---
>https://github.com/OpenImageIO/oiio/pull/1279/files#r46729716

-- 
Larry Gritz
lg@larrygritz.com

fpsunflower:
LGTM otherwise!

lgritz:
Updated

PR #1281, opened 2015-12-06
maketx -u : remake the file even with matching dates, for different command lnes

lgritz:
Compare the command line from the last build (in the metadata) with the
current one, to help decide whether the texture needs to be rebuilt.


lgritz:
Merged into master. I'm hesitant to merge into RB-1.6, which we are trying to freeze all but bug fixes to turn into a stable release. I think what I'll do is keep it master only until we are out of beta on 1.6, and then if nobody has complained about this behavior change in master, I'll backport for the next 1.6 update release.

PR #1283, opened 2015-12-07
Enable TIFF write of 'half' pixels, if "tiff:half" attribute is nonzero.

lgritz:
Enable TIFF write of 'half' pixels, if "tiff:half" attribute is nonzero.

Set up a global OIIO attribute "tiff:half" that sets the default globally,    and also have the ImageOutput for TIFF recognize a per-file "tiff:half"    override. The default in both cases is 0, meaning that requests for    half pixel float files in the absence this attribute nonzero will be    automatically be upgraded to float, which is more widely read.

The reason for this is that nearly every non-OIIO-based reader fails to    correctly recognize half-valued pixels in TIFF files, including    PhotoShop and Nuke 9. So you are urged not to write half tiff, unless    you are sure that the only consumer of those files uses OIIO (which can    read them properly), or have verified that other consumers in your    pipeline get it right.

Here is the Adobe tech report describing this TIFF extension:     http://chriscox.org/TIFFTN3d1.pdf

A typical OIIO use of this would be to create a half TIFF texture file:

        maketx in.tif -attrib tiff:half 1 -d half -o half.tx

Why would you do this? Because in my benchmarks, texture access of    a half TIFF texture file is nearly twice the raw read performance of    the equivalent OpenEXR half texture.

Along the way, I added the missing get_int_attribute, get_float_attribute, and get_string_attribute equivalents to the global OIIO::getattribute, to match how we have done it for the versions that are methods of ImageSpec, TextureSystem, and a number of other classes.

### DIFF
--- a/src/doc/builtinplugins.tex
+++ b/src/doc/builtinplugins.tex
@@ -982,10 +982,10 @@ options are supported:
 
 \vspace{.125in}
 
-\noindent\begin{tabular}{p{2.0in}|p{0.5in}|p{2.75in}}
+\noindent\begin{tabular}{p{1.8in}|p{0.5in}|p{2.95in}}
 Configuration attribute & Type & Meaning \\
 \hline
-\qkw{oiio:UnassociatedAlpha} & int & If nonzero, will leave alpha unassociated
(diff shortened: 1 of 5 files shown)


fpsunflower:
LGTM

PR #1287, opened 2015-12-10
oiiotool --subimage now takes subimage name as well as numeric index.

### LINKED ISSUE #1271: EXR-2.0 part/subimage handling in OpenImageIO / oiiotool
Processing EXR-2.0 image sequences with oiiotool becomes complicated when extracting channels from multipart image sequences with varying structure.

EXR-2.0 supports distributing channels within an image throughout multiple 'parts'.
E.g, part 1 may contain "R,G,B", part 2 contain "A", part 3 contain "plate.R","plate.G","plate.B"
Splitting into parts can rarely read channels slowing down access, allow mixing of tiles and scanlines, or allow different compression schemes to be used on different channels.

Parts are handled using subimages in OpenImageIO, but they are conceptually different: EXR parts are still one image, it's just the storage of different channels can be separated. subimages are independent, but related, images.

We have a situation where an extra layer (say "CG.R","CG.G","CG.B","CG.A") appears within the sequence, and gets assigned as part 3. The plate layer is then moved to part 4.
To process just "plate" for all images in a sequence, we need to check the part structure for every file and call oiiotool with the correct -subimage for each frame For some frames, we need to pass "-subimage 3" and for others "-subimage 4".

Possible improvements could be:
 * an option to force an EXR-2.0 to open as a single 'flat' image (parts are merged at load time)
 * support in OpenImageIO to collapse subimages into a single image, combining the channels from each into a single image. This should be the same as doing --sisplit followed by one --chappend for each subimage in the stack (but wouldn't require a priori knowledge of how many parts the file has)
 * support for accessing subimages by name instead of number. EXR-2.0 parts must have a unique name, so doing -subimage "plate" will likely always pull out the right thing regardless of which subimage that is.

lgritz:
Fixes #1271


lgritz:
@peterhillman, what do you think of this?

lgritz:
I'm going to merge into master. We can always revise if you don't like what I've done.

Issue #1003, opened 2014-11-27
FFMPEPInput subimages


sobotka:
iv shows the correct number of frames with the FFMPEG extensions, but when flipping through the subimages, only one updated frame for approximately every twelve subimages appears to be displayed.

lgritz:
Was this ever resolved?

sobotka:
No. For posterity sake, I'll outline the general problem for those folks that are having more time than myself at the current moment.

All codecs can end up being a mish-mash of I (whole), P (predictive based on previous), and B (bi-directional predictive forward and back) frames. Many H264 encoders rely heavily on B frames to optimize for size.

If we assume a trivial example, the problem manifests itself quickly. Let's assume a series:

Frame Number | 0 | 1 | 2 | 3 | 4
-------------|---|---|---|---|---
Frame Type   | I | P | I | B | I

If OIIO seeks to frame 0, no problem as we have a full frame in the decoded buffer. Frame 1, despite being a predictive frame, we have already sifted through the decode and should have enough information to decode the frame<sup>1</sup>. Frame 2 is again a fresh start and not a problem as it is a full frame. Frame 3 however, is where our issue crops up.

While we can seek to frame 3, it being a bi-directional predictive frame means we rely on both the previous frame and the next I frame, which in this case is frame 4. We do not yet have enough information to pull a full frame off of the buffer.

In FFMPEG terms, the proper technique would be to find our desired frame. From there, we need to seek to the first I frame previous, which is I believe typical FFMPEG behavior. So in our above example, I believe the seek will return frame 2. Then we must decode past the I frame using avcodec_decode_video2 until we pop a full frame, tracking the decoded frames as we go. Once we hit the next I frame, we can figure out what frame the artist / technician desired and present it accordingly.

Phew. Apologies for the spam.

lgritz:
Thanks for clarifying.

lgritz:
I think we long ago merged a PR that was supposed to fix this (#1288). Closing.

PR #1288, opened 2015-12-10
ffmpeginput

fafik23:
FIX ffmpeginput to read all frame. Tested with ffmpeg-2.8.1

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -44,6 +44,7 @@ extern "C" { // ffmpeg is a C api
 #include <boost/thread/once.hpp>
 
 #include "OpenImageIO/imageio.h"
+#include  <iostream>
 
 OIIO_PLUGIN_NAMESPACE_BEGIN
 
@@ -84,7 +85,10 @@ private:
     int m_last_search_pos;
     int m_last_decoded_pos;
     bool m_offset_time;
+    bool m_codec_cap_delay;
     bool m_read_frame;
+
+    int64_t m_start_time;
     // init to initialize state
     void init (void) {
         m_filename.clear ();
@@ -102,6 +106,9 @@ private:
         m_last_decoded_pos = 0;
         m_offset_time = true;
         m_read_frame = false;
+        m_codec_cap_delay=false;
+    	m_subimage = 0;
+        m_start_time=0;
     }
 };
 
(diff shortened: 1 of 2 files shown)


src/ffmpeg.imageio/ffmpeginput.cpp
```diff
@@ -206,29 +213,31 @@ FFmpegInput::open (const std::string &name, ImageSpec &spec)
         !strcmp (m_codec_context->codec->name, "dvvideo")) {
         m_offset_time = false;
     }
+
+   m_codec_cap_delay=m_codec_context->codec->capabilities&AV_CODEC_CAP_DELAY ;
```
lgritz: In the version of ffmpeg on one of my systems, AV_CODEC_CAP_DELAY is not found.

But there is a CODEC_CAP_DELAY. Are they the same, and I'm using a slightly older version, after which the AV_ was prepended?

do you think it's sufficient if I just add something like:

    #ifndef AV_CODEC_CAP_DELAY
    #define AV_CODEC_CAP_DELAY CODEC_CAP_DELAY
    #endif

fafik23: We can swich AV_CODEC_CAP_DELAY to CODEC_CAP_DELAY.
https://ffmpeg.org/doxygen/2.8/libavcodec_2avcodec_8h_source.html#l01064




2015-12-10 19:08 GMT+01:00 Larry Gritz <notifications@github.com>:

lgritz: OK, I'll make that change as part of the merge.

lgritz:
Can you explain briefly what problem this addresses? Is this related to issue #1003? Or is it something else?

fafik23:
first it resolved problem with codec AV_CODEC_CAP_DELAY
http://www.ffmpeg.org/doxygen/2.8/group__lavc__decoding.html#ga99ee61b6dcffb7817a275d39da58cc74

second i remove condition to run avcodec_decode_video2 this make that we
put to codec only frame with pts >= pts we looking thats why we see only
key frames.
if (current_pos >= m_subimage) {
avcodec_decode_video2 (m_codec_context, m_frame, &finished, &pkt)
}

Last I add if( current_frame == frame && finished) to be sure we display
exactly this frame we need (current_frame is take from m_frame->pkt_pts
http://www.ffmpeg.org/doxygen/2.8/structAVFrame.html#a47f555732f9c18928de727a163cd7cb6
). this can make infinity loop when something go wrong.

I have not  tested this on codec with bi-directional frame.


2015-12-10 19:40 GMT+01:00 Larry Gritz <notifications@github.com>:

lgritz:
Thanks, I'll add that to the commit message before I merge.

lgritz:
Merged into master, with the fix described above and also a few very small formatting changes.

I'll eventually merge this into RB-1.6, but we're trying to lock it down for a release right now, so I don't want to merge it there until after the beta is over and a release is tagged. That will also give people a little while to test this further in master.

PR #1289, opened 2015-12-11
DeepData refactor

lgritz:
DeepData is ok for simple passing of data in and out of the ImageInput/ImageOutput functions, but its internal layout is actually horrible for doing any actual work. (Exercise for the reader: try to imagine how to add or delete individual samples in a particular pixel of a DeepData.)

I have longer-term goals of overhauling its internals to make these operations natural and efficient. But as a first step, I need to properly hide the current internals so that I'm free to change them later.

This patch has hides the internals of DeepData -- everything is private now, and the nontrivial parts are hidden behind a PIMPL idiom. So "client" code is now forced to use the public methods. This is going to be slightly painful for apps that are used to reaching into the guts and manipulating the underlying vectors directly. Sorry about that. But once you switch to using the API, it will allow us to iterate on the internals to improve them, without any danger of further breakage of code that relies on using DeepData.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -63,6 +63,9 @@
 
 OIIO_NAMESPACE_BEGIN
 
+class DeepData;
+
+
 /// Type we use for stride lengths.  This is only used to designate
 /// pixel, scanline, tile, or image plane sizes in user-allocated memory,
 /// so it doesn't need to represent sizes larger than can be malloced,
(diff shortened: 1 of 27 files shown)


src/include/OpenImageIO/imageio.h
```diff
@@ -459,6 +460,16 @@ struct OIIO_API DeepData {
     /// Use with care.
     void *channel_ptr (int pixel, int channel);
     const void *channel_ptr (int pixel, int channel) const;
+
+    array_view<const unsigned int> all_nsamples () const;
+    array_view<const char> all_data () const;
+    void * const * all_pointers () const;  // Caution: expect depcrecation
```
fpsunflower: typo: `deprecation`

fpsunflower:
LGTM

src/include/OpenImageIO/imageio.h
lgritz: thanks, fixed

lgritz:
I won't commit this until after I get a review on the Arnold-side changes.

lgritz:
Updated slightly -- I moved the channeltypes vector into the Impl.

lgritz:
One more change: split the DeepData class into its own header, deepdata.h. No code or behavior changed with this last adjustment, just moved around.

lgritz:
Un-LGTM'ing this, requesting second look.

This weekend I got obsessed, could not help myself from essentially completing the next round of work so I could do it as one single change. I have amended this PR with the complete package. Here is the commit message from the second round:

Overhaul of DeepData internals and minor change to API

Internally, change the data storage order from [pix][chan][samp] to    [pix][samp][chan], so that now all the channels of one sample are    contiguous (as we store them internally for flat images).

We should have done it this way all along. I thought the other order was    required for OpeneEXR (based on a misreading of the example code), but    it turns out that proper use of strides lets us store the channels    interleaved in memory if we want.

No longer store pointers internally at all, just need to keep the prefix    sum of the number of samples for each pixel, in order to quickly compute    the data offset of a particular sample or channel.  When we need    pointers "externally" (currently, only for read/write of OpenEXR), use a    new get_pointers() method that calculates them on the fly.

This lets us add insert_samples() and erase_samples(), which let us    (simply, if not cheaply) alter the samples in an existing DeepData, which    will be the basis of much more extensive support for deep files in    ImageBufAlgo and oiiotool (which support deep files now in only an    extremely minimal set of operations).

Get rid of DD::channel_ptr(), which no longer makes a lot of sense, and    instead have data_ptr(). Use with caution, because the pointers will no    longer be valid if anything causes the internal data buffer to     reallocate.

Get rid of the DD::alloc() method. We no longer require client code to    trigger the allocation of the data array. Instead, it will automatically    happen the first time you read or write a pixel value if it's not    already allocated.

PR #1290, opened 2015-12-15
Fix missing Python bindings for OIIO::getattribute

lgritz:
This is embarrassing: somehow we never added the global getattribute() calls to the Python bindings.

Also notice that the Python bindings for ImageSpec::getattribute were incorrectly bound to "get_attribute". So add "getattribute" and consider "get_attribute" a deprecated synonym.

I will backport this to 1.6, despite being at RC stage, since this is an obvious bug.

### DIFF
--- a/src/doc/imageioapi.tex
+++ b/src/doc/imageioapi.tex
@@ -785,6 +785,7 @@ other's global error messages.
 \apiitem{bool {\ce attribute} (string_view name, TypeDesc type,
   const void *val)}
 \index{globalattribute}
+\label{sec:globalattribute}
 \indexapi{attribute}
 
 Sets an global attribute (i.e., a property or option) of \product.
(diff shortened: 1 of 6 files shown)


iwootten:
Thanks for the swift patch for this @lgritz!

lgritz:
@iwootten Do you work from master or a release branch? If the latter, which one?

iwootten:
I'm using the 1.5.21 release. It looks like some stuff has gone into RB-1.5 since then, so I guess that means I'm a couple of commits back on RB-1.5.

lgritz:
We are a few days from making 1.6 be the official release. I don't suppose you are in a position to switch to 1.6? Or do I need to backport to 1.5 as well?

lgritz:
Merged into master, RB-1.6, RB-1.5.

iwootten:
Thanks for this! Successfully using it on the 1.5.22 release.

Issue #1292, opened 2015-12-17
master fails to compile on CentOS 6.6


sebastianelsner:
I think the issue was introduced in the deep refactoring pull request:

$ make
[  9%] Built target OpenImageIO_Util
[  9%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/deepdata.cpp.o
/workspace/build/oiio/src/libOpenImageIO/deepdata.cpp: In member function ‘size_t OpenImageIO::v1_6::DeepData::Impl::data_offset(int, int, int)’:
/workspace/build/oiio/src/libOpenImageIO/deepdata.cpp:84: warning: comparison between signed and unsigned integer expressions
In file included from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/x86_64-redhat-linux/bits/c++allocator.h:35,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/allocator.h:49,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/string:44,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/locale_classes.h:43,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/bits/ios_base.h:44,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/ios:44,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/ostream:41,
                 from /usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/iostream:41,
                 from /usr/local/include/OpenEXR/half.h:90,
                 from /workspace/build/oiio/src/libOpenImageIO/deepdata.cpp:36:
/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/ext/new_allocator.h: At global scope:
/usr/lib/gcc/x86_64-redhat-linux/4.4.7/../../../../include/c++/4.4.7/ext/new_allocator.h: In instantiation of ‘__gnu_cxx::new_allocator<const OpenImageIO::v1_6::TypeDesc>’:
[...]

PR #1294, opened 2015-12-17
Fix compiler warnings

lgritz:
The trickiest issue involved DeepData::get_pointers (std::vector<const void*>&). Some compilers won't eat a vector of a const type. Others seem fine.    Sidestep the problem by removing this variety of DeepData::get_pointers,    it's not necessary with the right casting inside the non-const version.


lgritz:
Fixes #1292, I hope.

PR #1296, opened 2015-12-19
DeepData: need to make internal alloc() thread-safe

lgritz:
The old DeepData required clients to call an explicit deep_alloc, which
they could do serially, after which parallel data accesses were
thread-safe. But the new DeepData doesn't have deep_alloc; instead, the
allocation happens implicitly whenever data needs to be accessed (i.e.,
in the potentially parallel section). So the internals themselves need
to ensure thread safety of the allocation, whenever it happens.


fpsunflower:
The change looks reasonable - do I understand correctly that this is purely for thread safe *reading* ?

lgritz:
This allows for thread-safe reading or writing of existing samples.

Anything that changes the number of samples in a pixel is not thread-safe and should have appropriate locks on the app side.

PR #1305, opened 2015-12-31
Speed up common data format conversion

lgritz:
The vast majority of image files store their data as uint8, uint16, or half. But we tend to use float buffers internally (IBA functions, oiio internals, etc.). So we are constantly converting from uint8/uint16/half to and from float. The convert_type template (in fmath.h) carries a lot of water -- basically any time we copy from buffer to buffer with format conversion. 

This patch attempts to analyze and improve performance for convert_type by specializing the template for the common paths, and for each one use the best SIMD-based code I could construct.

Summary of results (in round numbers):
* uint16 or uint8 -> float with SSE improve by > 2x.
* float -> uint16 or uint8 with SSE improve by > 7x.
* float -> half improves by 5x when F16C CPU feature is used.
* half -> float improves by 10% when F16C CPU feature is used.

Example benchmarks (on my laptop, built with USE_SIMD=sse4.2,f16c):

Before:

     Benchmark conversion of  uint8 ->  float :  1495.3 Mvals/sec
     Benchmark conversion of  float ->  uint8 :   182.2 Mvals/sec
     Benchmark conversion of uint16 ->  float :  1482.2 Mvals/sec
     Benchmark conversion of  float -> uint16 :   181.1 Mvals/sec
     Benchmark conversion of   half ->  float :  2408.3 Mvals/sec
     Benchmark conversion of  float ->   half :   677.8 Mvals/sec

After:

     Benchmark conversion of  uint8 ->  float :  3300.5 Mvals/sec
     Benchmark conversion of  float ->  uint8 :  1312.7 Mvals/sec
     Benchmark conversion of uint16 ->  float :  2934.7 Mvals/sec
     Benchmark conversion of  float -> uint16 :  1518.8 Mvals/sec
     Benchmark conversion of   half ->  float :  2609.6 Mvals/sec
     Benchmark conversion of  float ->   half :  3160.6 Mvals/sec

The numbers are millions of conversions per second, higher numbers are better.

Note that these benchmarks are isolating just the buffer conversion utility convert_types. You will not see speedups this large in an application or even an oiiotool command, because buffer copies and conversions are just one part of the work (this doesn't speed up the actual I/O from disk!). For large files, I am seeing timeable improvements in oiiotool operations, however.

I don't expect this to improve TextureSystem performance, since the ImageCache keeps tiles in uint16, uint8, or half format in the cache. This is more aimed at oiiotool and apps that use ImageBufAlgo functions.

Along the way, some odds and ends:
* fmath_test adds conversion benchmarks.
* timer.h added utilities to help with micro-benchmarks by thwarting the optimizer's ability to notice certain unused results.
* Some new and useful simd methods.


lgritz:
Bug found. Updated posted.

fpsunflower:
LGTM

PR #1307, opened 2016-01-07
DeepData::get_pointers - Fix pointer botch for 0-sample pixels

lgritz:
When there are no samples for a pixel, the pointer to a particular sample (which should not be used) will compute something meaningless, which triggered the assertion.


fpsunflower:
LGTM

PR #1308, opened 2016-01-07
ImageBuf::Iterator performance improvement -- reduce "iterator penalty" by 50%

lgritz:
Another review from winter break. :-)

This small change -- guaranteeing inlining of two methods and slight
rearrangement of conditionals -- cuts the time it takes an iterator to
traverse an image by half, thus making it only a little bit slower than
looping over an array of pixels.


fpsunflower:
LGTM! Nice speedup for such a small change

PR #1312, opened 2016-01-08
Including <climits> in SHA1.h fixes some typing/linking issues.

### LINKED ISSUE #1298: undefined reference to `OpenImageIO::v1_6::CSHA1::Update(unsigned char const*, unsigned long)
OIIO 1.6.8 fails to link to one of its own utils only on i686. x86_64 works fine in this regard. This is on Arch Linux against boost 1.60 with gcc 5.3. Building with clang fails likewise which makes me think it's a wrongly configured build system.

x86_64 build log (works): http://b.1339.cf/srradra.log
i686 build log (fails): http://b.1339.cf/tsealju.log

### LINKED ISSUE #1311: SHA1.h misses #include <climits> for the definition of ULONG_MAX
resulting in SHA1.cpp usint UINT_32=unsigned int, and other sources using UINT_32=unsigned long

lgritz:
Fixes #1311
Fixes #1298

### DIFF
--- a/src/include/OpenImageIO/SHA1.h
+++ b/src/include/OpenImageIO/SHA1.h
@@ -74,6 +74,8 @@
 #ifndef ___SHA1_HDR___
 #define ___SHA1_HDR___
 
+#include <climits>
+
 #include "export.h"
 #include "oiioversion.h"
 


lgritz:
Fixes #1311
Fixes #1298

devernay:
looks good!
#1311 was a duplicate of #1298, indeed

lgritz:
Merged into master and also the head of RB-1.6.
Thanks for figuring this out!

devernay:
I figured this out when compiling oiio on 32-bits PPC (OS X 10.5), using gcc 4.9.

Note that some atomics are not available on PPC32 and require linking with libatomic.
-latomic is necessary on ppc for __atomic_fetch_add_8 and __atomic_exchange_8

I had to configure OIIO with the flags
 -DCMAKE_SHARED_LINKER_FLAGS=-latomic -DCMAKE_EXE_LINKER_FLAGS=-latomic
to make it compile.

This should probably be tested in cmake: if the compiler is GCC-compatible and __atomic_fetch_add_8 is not available, look for it in -latomic
It is done in LLVM:
https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/CheckAtomic.cmake

i still haven't tested the PPC OIIO, though

PR #1314, opened 2016-01-11
simd.h include file logic cleanup.

lgritz:
immintrin.h seems to always include a bunch of other files that we included explicitly, so those extra includes can go away.

Also, the guard around x86intrin.h needs to ensure that gcc version is at least 4.4 and it's also unnecessary unless f16c is requested.


devernay:
works fine for me!

PR #1315, opened 2016-01-12
Fix pesky type problems on some 32 bit platforms

lgritz:


lgritz:
Merged. Now sure why it didn't auto-close the ticket.

PR #1319, opened 2016-01-18
Draw lines, boxes

lgritz:
Weekend airplane project. Must be useful for something, right?

`IBA::render_point()` draws a single pixel.

`IBA::render_line()` draws a line segment between two designated pixels (it    can easly be used for one segment of a poly-line, with an option to avoid    rasterizing the first point, so polyline vertices are not double-drawn).

`IBA::render_box` draws a filled or unfilled box between the two diagonal corner pixels.

`oiiotool --line x1,y1,x2,y2,...` draws a polyline

`oiiotool --box x1,y1,x2,y2` draws a box


lgritz:
After sending the first PR last night, I decided we really did need IBA::render_box(), so I added that.

PR #1321, opened 2016-01-19
Expression fix: allow oiiotool command line expression metadata names to contain ':'

lgritz:


sebastianelsner:
fixes, issue, thanks!

PR #1324, opened 2016-01-22
Fix rla output when writing files with nonzero image origin

lgritz:


lgritz:
Unfortunately, my original diagnosis was wrong, as revealed when I went to add an actual test to the testsuite. Remember, "untested code should be presumed to be wrong." So it's a more extensive fix involving reading and writing, updated the PR.

PR #1336, opened 2016-02-02
IC: upon read failure, pause *between* tries, not *after* tries.

lgritz:
On the last try (or the only try, if failure_retries was 0), we were
needlessly delaying.


fpsunflower:
LGTM!

PR #1344, opened 2016-02-04
Change default to USE_CPP11=1

lgritz:
This does NOT make the codebase C++11-only. Although that will happen
sooner or later, almost certainly during the course of 1.7 development.
But for now, this just means that it defaults to C++11, and you need to
proactively use 'make USE_CPP11=0' (or 'cmake -DOIIO_BUILD_CPP11=0') to
stay on C++03 compatibility. This is to ensure that nobody is surprised
when the C++11 switch eventually comes, by requiring them to have
continued awareness that they are living on borrowed time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,17 +11,12 @@ os:
 #osx_image: xcode7
 dist: trusty
 
-# Do a default build (C++03, sse2), an sse4.2 build, and a C++11 build.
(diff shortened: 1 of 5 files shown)


lgritz:
Posted for a week with no urgent objections. 

Going once... going twice...

PR #1350, opened 2016-02-10
Fix precision discrepancy

lgritz:
A while back, we changed convert_type (the kind that operates on arrays)
to use a float as intermediate when appropriate, and double for certain
bigger types, instead of always using double no matter what. This sped up
certain buffer conversions substantially.

But at that time, we neglected to adjust the 1-at-a-time variety of
convert_type to use the same logic. This led to occasional discrepancies
between the math of type converting a whole buffer versus doing it one
element at a time. Only for certain values, and generally just one
LSB. But that's enough to have some strange inconsistencies.

Unfortunately, one place that this can crop up is in computing SHA-1
hashes, including maketx -- for certain textures where the source is
uint8 and must be converted to float internally for the MIPmap resizing,
a single bit flip in an LSB can actually change the SHA-1 hash that is
placed in the header in order to detect duplicate textures. This can
happen even if the final pixel values (once converted back to uint8 for
output) are still 100% correct, because for better or for worse, the
hashing is done on the intermediate float buffer of the first
level. This only happens for certain files, containing certain values,
and thus I think it's worth the occasional failing to detect a duplicate
texture (only if you mix old-maketx and new-maketx textures in a single
ImageCache), in order to have more stable consistency with the floating
point conversion.


fpsunflower:
LGTM

PR #1352, opened 2016-02-11
Fix PNM endian issue.

### LINKED ISSUE #1343: 16-bits PNM are MSB-first, pnminput.c's raw_to_raw should swap bytes if sizeof(T)==2
reference: http://netpbm.sourceforge.net/doc/pamendian.html
note that this is already taken care of in pnmoutput.c

lgritz:
Sent in by Frédéric Devernay.
Fixes #1343


devernay:
works great!

PR #1353, opened 2016-02-11
Fix UTF-8 support by fstream on MINGW

MrKepzie:
Added patch corresponding to https://github.com/OpenImageIO/oiio/issues/1349

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -39,11 +39,12 @@
 #include <boost/filesystem.hpp>
 #include <boost/foreach.hpp>
 #include <boost/regex.hpp>
-#include <boost/shared_ptr.hpp>
+
 
 #include "OpenImageIO/dassert.h"
 #include "OpenImageIO/ustring.h"
 #include "OpenImageIO/filesystem.h"
+#include "OpenImageIO/refcnt.h"
 
 #ifdef _WIN32
 #include <windows.h>
@@ -611,13 +612,11 @@ open_ifstream_impl(string_view path,
         return 0;
     }
     try {
-        ret->open(
 #ifdef _WIN32
-                  wpath.c_str(),
+        ret->open(wpath.c_str(),mode);
 #else
-                  path.c_str(),
+        ret->open(path.c_str(),mode);
 #endif
-                  mode);
     } catch (const std::exception & e) {
         delete ret;
         return 0;
@@ -643,13 +642,11 @@ open_ofstream_impl(string_view path,
         return 0;
     }
     try {
-        ret->open(
 #ifdef _WIN32
-                  wpath.c_str(),
+        ret->open(wpath.c_str(),mode);
 #else
-                  path.c_str(),
+        ret->open(path.c_str(),mode);
 #endif
-                  mode);
     } catch (const std::exception & e) {
         delete ret;
         return 0;
@@ -674,6 +671,14 @@ Filesystem::open (std::istream** stream,
         return;
     }
     *stream = open_ifstream_impl(path, mode | std::ios_base::in);
+    if (!*stream) {
+        return;
+    }
+    if ((*stream)->fail()) {
+        delete *stream;
+        *stream = 0;
+        return;
+    }
     if (mode & std::ios_base::ate) {
         (*stream)->seekg (0, std::ios_base::end);
     } else {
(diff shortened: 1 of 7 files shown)


MrKepzie:
I made fixes, will remake a new pull request.

src/libutil/filesystem.cpp
```diff
@@ -39,6 +39,7 @@
 #include <boost/filesystem.hpp>
 #include <boost/foreach.hpp>
 #include <boost/regex.hpp>
+#include <boost/shared_ptr.hpp>
```
lgritz: Instead of boost/shared_ptr.hpp, let's include OpenImageIO/refcnt.h, which defines an OIIO::shared_ptr<> that underneath is std::shared_ptr on C++11 and uses boost::shared_ptr otherwise (and for each case, includes the right headers).

src/libutil/filesystem.cpp
```diff
@@ -514,6 +523,182 @@ Filesystem::open (std::ofstream &stream, string_view path,
...
+void
+Filesystem::open (std::istream** stream,
+                  string_view path,
+                  std::ios_base::openmode mode)
+{
+    if (!stream) {
+        return;
+    }
+    *stream = open_ifstream_impl(path, mode | std::ios_base::in);
+    if (mode & std::ios_base::ate) {
```
lgritz: I think you need to handle open_ifstream_impl failing to open the file and setting *stream to NULL.

src/libutil/filesystem.cpp
```diff
@@ -523,13 +708,19 @@ Filesystem::read_text_file (string_view filename, std::string &str)
...
-    Filesystem::open (in, filename);
+    boost::shared_ptr<std::istream> in;
+    {
+        std::istream* inraw;
+        Filesystem::open (&inraw, filename);
+        if (inraw) {
+            in.reset(inraw);
+        }
+    }
+
```
lgritz: Would this section be slightly shorter/clearer as:

    std::istream* inraw;
    Filesystem::open (&inraw, filename);
    boost::shared_ptr<std::istream> in (inraw);

???

src/libutil/filesystem.cpp
```diff
@@ -514,6 +523,182 @@ Filesystem::open (std::ofstream &stream, string_view path,
...
+                  string_view path,
+                  std::ios_base::openmode mode)
+{
+    
+    if (!stream) {
+        return;
+    }
+    *stream = open_ofstream_impl(path, mode | std::ios_base::out);
+    if ((mode & std::ios_base::app) == 0) {
+        (*stream)->seekp (0, std::ios_base::end);  // force seek, otherwise broken
```
lgritz: Again, here I am worried about a file open failure within open_ofstream_impl leading to *stream == NULL.

src/libutil/filesystem.cpp
```diff
@@ -514,6 +523,182 @@ Filesystem::open (std::ofstream &stream, string_view path,
...
+    }
+    try {
+        ret->open(
+#ifdef _WIN32
+                  wpath.c_str(),
+#else
+                  path.c_str(),
+#endif
+                  mode);
+    } catch (const std::exception & e) {
```
lgritz: Just IMHO, I believe that this would be clearer as

```
#ifdef _WIN32
    ret->open (wpath.c_str(), mode);
#else 
    ret->open (path.c_str(), mode);
#endif 
```

rather than having the #ifdef split up within the argument list of the function call, particularly when the whole function call can fit on one line.

src/libutil/filesystem.cpp
lgritz: In fact, it turns out that this is why it failed the Travis-CI builds -- we have testsuite tests that purposely try to open nonexistant files, and if that happens the app will crash if you dereference *stream like this.

MrKepzie:
Will make those fixes later this week

MrKepzie:
Just added those fixes

lgritz:
LGTM.

I've squashed and rebased, and as soon as I double check that it passes all the Travis tests, will merge and also backport to 1.6.

lgritz:
I don't seem to have a CLA from you. I won't hold up the merge for that because for a bug fix like this, it's hard to imagine any IP issues, but it would be nice to have on file, especially to streamline any future contributions that may be large enough to be an issue.

lgritz:
Merged into master and RB-1.6.
Thanks for the fixes!

MrKepzie:
I will sign the CLA and send it to you on monday. Btw @lgritz do you have an idea on the last comment I made on the closed ticket here https://github.com/OpenImageIO/oiio/issues/1123 ?

lgritz:
Sorry for the delay. Answered on the other ticket.

MrKepzie:
@lgritz There is still a bug remaining in this implementation: the stdio_filebuf is never deleted by the destructor of ostream/istream according to the documentation. We must explicitly delete the stdio_filebuf ourselves. 
Basically we can not have a shared_ptr<istream> lying around because we'd need to maintain separatly a shared_ptr<stdio_filebuf> (plus it'd need to be included everywhere). 

I will add the RAII struct in the Filesystem API to handle the destructors properly and so that we include stdio_filebuf.h only in Filesystem.cpp

PR #1354, opened 2016-02-12
Fix error reading tiles for mixed-format exr files.

### LINKED ISSUE #1335: 1.6.x fails to read correctly mixed bit depth EXR file
1.5.23 reads correctly the following file https://github.com/MrKepzie/build-oiio/blob/master/chopper.exr 
with oiiotool:

`oiiotool chopper.exr -o chopper_o.exr `

But seems to be broken in 1.6.9: the output file for the RGBA pass only has the red channel set and other passes are blanked

lgritz:
Fixes #1335

### DIFF
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -481,8 +481,7 @@ ImageInput::read_tiles (int xbegin, int xend, int ybegin, int yend,
                         return false;
                 } else {
                     buf.resize (full_tilebytes);
-                    ok &= read_tile (x, y, z, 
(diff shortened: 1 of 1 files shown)


MrKepzie:
This correctly fixes the issue

lgritz:
I also backported to RB-1.6.

MrKepzie:
We'll probably move soon to RB-1.6 then. We have several patch also to validate against 1.6 (mostly related to MingW support)

lgritz:
Great. As they come through one by one, I will merge them into 1.6. Let me know when we have tackled them all, and then I'll tag a release at that point.

PR #1359, opened 2016-02-18
Impove ffmpeg backwards compatibility

sopvop:
Allows building with ancient ffmpeg on Centos6 and up to ffmpeg 3.0

Also api documentation state that av_free_packet should be called before
each av_read_frame call, unless packet is duped. Not doing that
can sometimes lead to segfaults. Better be safe.

Calling avcodec_close with null pointer caused segfaults on ancient ffmpeg.

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -37,8 +37,33 @@ extern "C" { // ffmpeg is a C api
 
 
 #if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(55,28,1)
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM. Builds fine on all the Travis images as well as my home machine. I just need to double check that it builds properly with the systems I use at work, if so I will merge it on Monday.

lgritz:
Merged into master.
Will merge into RB-1.6 in a couple days if nobody complains about trouble in master.

lgritz:
Or do you even need it on RB-1.6?

sopvop:
No, I'm on master

PR #1360, opened 2016-02-19
Faster unit tests for debug, code coverage, and continuous integration.

lgritz:
Some of our unit tests do benchmarking which is helpful for interactive,
optimized builds, but for DEBUG, code-coverage, or CI builds, the times
mean nothing. So in those cases, reduce the number of iterations and don't
do multiple trials. This speeds up running testsuites for those builds.

### DIFF
--- a/src/libOpenImageIO/compute_test.cpp
+++ b/src/libOpenImageIO/compute_test.cpp
@@ -279,9 +279,9 @@ getargs (int argc, char *argv[])
 {
     bool help = false;
     ArgParse ap;
-    ap.options ("atomic_test\n"
+    ap.options ("compute_test\n"
                 OIIO_INTRO_STRING "\n"
-                "Usage:  atomic_test [options]",
+                "Usage:  compute_test [options]",
                 // "%*", parse_files, "",
                 "--help", &help, "Print help message",
                 "-v", &verbose, "Verbose mode",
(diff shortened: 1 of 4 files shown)


src/libOpenImageIO/compute_test.cpp
```diff
@@ -308,6 +308,14 @@ getargs (int argc, char *argv[])
 
 int main (int argc, char *argv[])
 {
+#if !defined(NDEBUG) || defined(OIIO_TRAVIS) || defined(OIIO_CODECOV)
+    // For the sake of test time, reduce the default iterations for DEBUG,
+    // CI, and code coverage builds. Explicit use of --iters or --trials
+    // will override this, since it comes before the getargs() call.
+    iterations /= 10;
```
fpsunflower: Why not just `iterations=1` ?

lgritz: Many of the tests are multithreaded, so I want the job size big enough for parts to be done in parallel, and for long enough to raise the probability of race conditions changing outcomes. But for debug or CI renders, they sure don't need to be long enough to get accurate timing information. So I figured that speeding them up by 90% was good enough.

fpsunflower: Makes sense. LGTM!

PR #1362, opened 2016-02-20
Improve Python image input behavior for native data

lgritz:
Python bindings of ImageInout.read_image, read_scanline, read_scanlines,
read_tile, read_tiles:

* If you ask for HALF, since there is no half in Python, return the half
  values packed into an 'unsigned short' array. (Old behavior was buggy
  and confusing: it packed the half values into a float array of half
  the expected size.)

* If you ask for UNKNOWN, like in C++ it means to give the raw/native
  data, packed into an 'unsigned char' array. (Old behavior was that
  UNKNOWN signified to return data in the type specified by spec.format.
  Now, if you want spec.format, pass spec.format.)

* Let the default be type=FLOAT, generate a float array. (Old default was
  UNKNOWN spec.format.)

### DIFF
--- a/src/doc/pythonbindings.tex
+++ b/src/doc/pythonbindings.tex
@@ -727,15 +727,15 @@ specified subimage or MIP level).
 \end{code}
 \apiend
 
-\apiitem{array ImageInput.{\ce read_image} (type=OpenImageIO.UNKNOWN) \\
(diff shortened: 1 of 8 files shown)


lgritz:
I merged into master. I'm going to let it sit in master for a few days, and if nobody complains, then I'll backport it to RB-1.6.

PR #1366, opened 2016-02-22
[Mingw] Bug fix in Filesystem

MrKepzie:
Fix recent changes to Filesystem::open where internal buffers of istream/ostream would never get deleted on MINGW.

### DIFF
--- a/src/include/OpenImageIO/filesystem.h
+++ b/src/include/OpenImageIO/filesystem.h
@@ -45,6 +45,7 @@
 #include <cstdio>
 #include <ctime>
 #include <fstream>
+#include <cassert>
 #include <string>
 #include <vector>
 
@@ -52,8 +53,18 @@
 #include "oiioversion.h"
 #include "string_view.h"
 
+#if defined(_WIN32) && defined(__GLIBCXX__)
+#define FILESYSTEM_USE_STDIO_FILEBUF 1
+#include <ext/stdio_filebuf.h> // __gnu_cxx::stdio_filebuf
+#endif
 
 OIIO_NAMESPACE_BEGIN
+
+#if FILESYSTEM_USE_STDIO_FILEBUF
+typedef __gnu__cxx::stdio_filebuf<char> stdio_filebuf;
+#else
+typedef std::basic_filebuf<char> stdio_filebuf;
+#endif
 
 /// @namespace Filesystem
 ///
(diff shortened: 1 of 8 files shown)


MrKepzie:
Note that while this correctly fixes memory leaks, there are still issues with UTF-8 files. I'm trying to find the root cause of that.

lgritz:
Wow, this is getting complicated.

Travis tests are failing on this PR: https://travis-ci.org/OpenImageIO/oiio/builds/111006447

How is this issue addressed with other software? 

Is there anything in C++11 (or 17?) or Boost that does a proper combination of UTF-8 and streams and resource management in a portable way?

lgritz:
I should probably have looked into this more deeply when the first code review came around.

OK, back to first principles. 

The problem is that (on Windows? or everywhere?) the std ifstream/ofstream constructors don't take UTF-8 filenames. So we created a Filesystem::open() wrapper that on Unix/Linux will just open the stream as usual but on Windows converts the filename from UTF-8 to UTF-16 and then uses a version of stream.open() that accepts wstring.

Right so far?

But then it turns out that if you're on Windows but using MinGW rather than MSVS, that doesn't exist? So another strategy was necessary?

But now it's morphed into new classes and resource management and totally different calling sequences on the other side (the clients of Filesystem). This is making me suspect that maybe this is the wrong approach.

Briefly looking at http://www.cplusplus.com/reference/fstream/ifstream/, I'm noticing that ifstream is actually typedef to a template instantiation, std::basic_ifstream<charT>.

What if, instead of all this, we used a simpler approach:

1. In filesystem.h, define OIIO::ifstream, which for most platforms simply typedefs to std::ifstream, but on MinGW is std::basic_ifstream<wchar_t>. 

2. Thus, the new versions of open() can be reverted, and just the old version can have minor modifications with `#if` for the MinGW case. On MinGW, will a std::basic_ifstream<wchar_t> properly open a UTF-16 filename?

3. The PNM and PSD cases directly manage their own streams, so for those the conversion is trivial -- just switch to OIIO::ifstream.

4. OpenEXR is the only hard case, because we have to pass a stream to the IlmImf API. But wait, we have our own OpenEXRInputStream wrapper. So maybe it's easy there, too.

(Repeat the same logic for output streams.)

I'm just sketching this out, not sure if it will work (and have no way to test myself on MinGW). But it sounds like if we can make something like this work, it might be less intrusive overall.

Am I even remotely understanding this all?

MrKepzie:
Your understanding of the problem is right, the matter is only related to compilation under Windows with MingW/GCC.

According to http://stackoverflow.com/questions/6524821/opening-stream-via-function the only way to fix UTF8 filenames on Windows when compiling with GCC is to use **__gnu_cxx::stdio_filebuf**.
The problem is, according to [this link](http://stackoverflow.com/questions/6524821/opening-stream-via-function) the destructor of **istream** and **ostream** *does not destroy nor performs any operations on the associated stream buffer object.*

So to be able to use this solution we have to keep a istream around and a stdio_filebuf around. 
The usage of the *Filesystem::open* should stay the same across different platforms, hence passing another argument is not really wise. 

Even if we were to do the following:

```
#ifdef FILESYSTEM_USE_STDIO_FILEBUF
typedef std::istream if stream
#else
typedef std::if stream if stream
#endif

```;;
It would not work because we would still have to keep the stdio_filebuf around, hence the Wrapper. 

Another solution come to my mind:

We could do keep the typedef on anything else than MingW, but on MingW, we inherit std::istream so that the *__gnu_cxx::stdio_filebuf* is actually a member of the class, hence it would be deleted when the istream would be deleted. We would still need to add a *close()* function because it is used at some places across OpenImageIO and does not exist in the API of std::istream.

Do you prefer this solution ? (less intrusive) but more hackish to understand

-------

Anyway, the reason why the problem is not *fully* solved yet, is because the *_wsopen_s* function used in *open_fstream_impl* is failing, returning errno = ERRNOENT (meaning *No such file or directory*) whereas  the file actually exists. It only fails when using filepaths with weird characters, such as "€ é è", eg.:
`C:\Users\MysuperUsér\Vidéos\Mes vidéos de vacances\t€t€.mov"`

**Comment by MrKepzie** [comment]
I found the bug: it was in the StrUtil::utf8_to_utf16 function. An extra 0 character was appended to the wide string, making the _wsopen_s function actually fail. 

I have found this function extremely useful to check that a string is correctly UTF-8 encoded:


```
// We check that strings contained in StringParam are UTF-8 encoded
// See http://openfx.sourceforge.net/Documentation/1.4/ofxProgrammingReference.html#kOfxParamTypeString
//
// From http://stackoverflow.com/questions/1031645/how-to-detect-utf-8-in-plain-c
static bool is_utf8(const char * string)
{
  if(!string)
    return 0;
  
  const unsigned char * bytes = (const unsigned char *)string;
  while(*bytes)
  {
    if( (// ASCII
         // use bytes[0] <= 0x7F to allow ASCII control characters
         bytes[0] == 0x09 ||
         bytes[0] == 0x0A ||
         bytes[0] == 0x0D ||
         (0x20 <= bytes[0] && bytes[0] <= 0x7E)
         )
       ) {
      bytes += 1;
      continue;
    }
    
    if( (// non-overlong 2-byte
         (0xC2 <= bytes[0] && bytes[0] <= 0xDF) &&
         (0x80 <= bytes[1] && bytes[1] <= 0xBF)
         )
       ) {
      bytes += 2;
      continue;
    }
    
    if( (// excluding overlongs
         bytes[0] == 0xE0 &&
         (0xA0 <= bytes[1] && bytes[1] <= 0xBF) &&
         (0x80 <= bytes[2] && bytes[2] <= 0xBF)
         ) ||
       (// straight 3-byte
        ((0xE1 <= bytes[0] && bytes[0] <= 0xEC) ||
         bytes[0] == 0xEE ||
         bytes[0] == 0xEF) &&
        (0x80 <= bytes[1] && bytes[1] <= 0xBF) &&
        (0x80 <= bytes[2] && bytes[2] <= 0xBF)
        ) ||
       (// excluding surrogates
        bytes[0] == 0xED &&
        (0x80 <= bytes[1] && bytes[1] <= 0x9F) &&
        (0x80 <= bytes[2] && bytes[2] <= 0xBF)
        )
       ) {
      bytes += 3;
      continue;
    }
    
    if( (// planes 1-3
         bytes[0] == 0xF0 &&
         (0x90 <= bytes[1] && bytes[1] <= 0xBF) &&
         (0x80 <= bytes[2] && bytes[2] <= 0xBF) &&
         (0x80 <= bytes[3] && bytes[3] <= 0xBF)
         ) ||
       (// planes 4-15
        (0xF1 <= bytes[0] && bytes[0] <= 0xF3) &&
        (0x80 <= bytes[1] && bytes[1] <= 0xBF) &&
        (0x80 <= bytes[2] && bytes[2] <= 0xBF) &&
        (0x80 <= bytes[3] && bytes[3] <= 0xBF)
        ) ||
       (// plane 16
        bytes[0] == 0xF4 &&
        (0x80 <= bytes[1] && bytes[1] <= 0x8F) &&
        (0x80 <= bytes[2] && bytes[2] <= 0xBF) &&
        (0x80 <= bytes[3] && bytes[3] <= 0xBF)
        )
       ) {
      bytes += 4;
      continue;
    }
    
    return 0;
  }
  
  return 1;
}
```



I recommend using it in debug mode whenever calling StrUtil::utf8_to_utf16

**Comment by MrKepzie** [comment]
I made a attempt in the branch inherit-fstream in my [fork](https://github.com/MrKepzie/oiio/tree/inherit-fstream)  to just typedef and inherit basic_istream/basic_ostream on Mingw. This had to introduce a new header to separate the MingW stuff (fstream_mingw.h)

I'm going to run a few tests and modifications on it before it is ready

**Comment by lgritz** [comment]
I took a quick look at your fork, and I liked what I saw. I think this is a much more straightforward approach, revert back almost completely to what we had before, with some extra stuff just for that one platform, emulating all the old simple call signatures.

I urge you to close/abandon this PR and instead start a new one with the inherit-fstream work (when you feel it's ready).

**Comment by MrKepzie** [comment]
Will do. Note that I stripped a few functions off the API of ifstream/ofstream because some are C++11.
Anyway everything that is used withing OIIO is implemented in those classes.

**Comment by MrKepzie** [comment]
This lack of UTF8 handling for MingW also goes beyond OpenImageIO and is present in OpenEXR too, in the function Imf::isOpenExrFile because they use standard ifstream. 

I'll provide a patch to the OpenEXR guys.

I've run into issues also with every boost::filesystem functions if we give them a std::string instead of a std::wstring. I'm trying to figure out where it is failing in boost right now.

**Comment by MrKepzie** [comment]
Aight: taken from boost::filesystem::path.hpp:
```

    //  For Windows, wchar_t strings do not undergo conversion. char strings
    //  are converted using the "ANSI" or "OEM" code pages, as determined by
    //  the AreFileApisANSI() function, or, if a conversion argument is given,
    //  using a conversion object modeled on std::wstring_convert.
```

Meaning there is no actual conversion from UTF-8 to UTF-16 done when they convert char* to wchar_t* internally. 

They do not use *MultiByteToWideChar (CP_UTF8,...* to convert strings as done in StrUtils::utf8_to_utf16. 

So every call to boost::filesystem on Windows has to be prepended with a call to StrUtils::utf8_to_utf16

In the end it's not really a *bug* in libstdc++ it's just that it's an undefined *standard* that MSVC imposed, so no patching of libstdc++ would be *right*. Only at the library level - say OpenImageIO - we can know whether we want to convert from UTF8 to UTF16 or any other encoding.

I'll fix this in filesystem.cpp and also make the same fixes to OpenEXR...

MrKepzie:
Quick update: after applying the same fixes to OpenEXR it now works like a charm. I'll re-make a new pull-request and also make one to OpenEXR

PR #1368, opened 2016-02-24
Mingw: fix fstream using inheritance of basic_stream 

MrKepzie:
See original pull-request: https://github.com/OpenImageIO/oiio/pull/1366

Less intrusive implementation which only requires user to use OIIO_NAMESPACE::ifstream / OIIO_NAMESPACE::ofstream instead of the original fix proposed in the pull request #1366.


MrKepzie:
I can actually make a new branch which would embed all these commits in a single 1 if you prefer

lgritz:
That would be slightly easier but don't worry if it's lots of trouble. After it's all fixed up, I will squash it to one commit before I merge it.

src/cmake/modules/FindOpenJpeg.cmake
```diff
@@ -80,7 +80,9 @@ if (OPENJPEG_HOME)
        ${OPENJPEG_HOME}/lib64)
   set (OpenJpeg_include_paths
        ${OpenJpeg_include_paths}
-       ${OPENJPEG_HOME}/include)
+       ${OPENJPEG_HOME}/include
+       ${OPENJPEG_HOME}/include/openjpeg
+       ${OPENJPEG_HOME}/include/openjpeg-1.5)
```
lgritz: There seems to be a mixture of the stream changes with other things?

src/include/OpenImageIO/filesystem.h
```diff
@@ -193,52 +210,16 @@ OIIO_API FILE *fopen (string_view path, string_view mode);
 ///
 OIIO_API std::string current_path ();
 
-/// Deprecated: Will not work correctly on GCC with MingW, prefer
-/// the version below taking a ofstream**
 /// Version of std::ifstream.open that can handle UTF-8 paths
 ///
-OIIO_API void open (std::ifstream &stream, string_view path,
+OIIO_API void open (OIIO_NAMESPACE::ifstream &stream, string_view path,
```
lgritz: Just OIIO::ifstream.

(They are equivalent, but let's stick to the namespace alias we use everywhere else.)

src/include/OpenImageIO/SHA1.h
```diff
@@ -91,6 +91,9 @@
 #endif
 #endif
 
+#ifdef WIN32
```
lgritz: WIN32 or _WIN32?

lgritz: Also, why is this change necessary?

src/libutil/sysutil.cpp
```diff
@@ -212,7 +212,7 @@ Sysutil::get_local_time (const time_t *time, struct tm *converted_time)
 #ifdef _MSC_VER
     localtime_s (converted_time, time);
 #else
-    localtime_r (time, converted_time);
+    *converted_time = *localtime(time);
```
lgritz: This seems like a bad idea. The localtime() uses some kind of internal buffer, no? And localtime_r is specifically to be reentrant and thread-safe.

src/cmake/modules/FindOpenJpeg.cmake
MrKepzie: Had trouble compiling on Mingw, these other fixes slipped through. I can remove them and commit them separately

src/libutil/sysutil.cpp
MrKepzie: Will enclose with a specific case for MINGW related. They were part of the changes needed to make it compile. Will strip it off the commits

src/cmake/modules/FindOpenJpeg.cmake
lgritz: I think it's easier to break this into two commits (could be one PR, or two) where one is all the stream changes, and the other are all the other miscellaneous changes for MinGW build.

If it's just too much trouble for you to do that, I bet I can tease it apart when I rebase and squash. It's not make or break, but the reviews are easier when orthogonal changes are separated.

lgritz:
I'm hoping to release 1.6.11 on Tuesday. Do you think you'll have this working shortly?
If not, I think I might revert the prior MinGW changes from 1.6 before I do that release -- since we still don't have a perfect MinGW, I don't think that's really breaking anything, and I'd rather not publish the "new" API calls that we now intend to eliminate.

MrKepzie:
Yes I will take time before tuesday

PR #1369, opened 2016-02-26
Texture: Fix formula for MIP level blending amount

lgritz:
Luca Fascione points out that we used a strange nonlinear construction
for the blending between adjacent MIP levels, which is computed as a 0-1
value and is based on comparing the ideal filter size for the sample
with the resolutions of the MIP levels. The more straightforward linear
mix that Luca suggests instead is a teensy bit sharper (but very close,
you need to look carefully to see the difference).

The line in question has not been modified since 2008, quite early in
OIIO's development. I have no recollection of doing it this way on
purpose. I can only guess that without adequate coffee, I just botched
coding the formula, and since it still mapped to 0-1 and was exact at
the endpoints, never noticed that it was a little weird in the
middle. The effect is a subtle bias toward the lower-res MIP level,
which has the effect of adding a slight blur.

This causes a lot of testsuite reference images to update.


fpsunflower:
LGTM! I'm anxious see the net impact on texture sharpness in practice.

PR #1371, opened 2016-02-26
Support separate ilmbase installation and other OpenEXR fixes

sopvop:
Restore usage of separate ILMBASE_ROOT, since some package managers and
OSes install ilbase and OpenEXR into different paths.

On some versions of OpenEXR ImfInt64.h includes ImathInt64.h directly,

libHalf is often installed without suffix, cmake should try to find it.


lgritz:
LGTM. I'll double check on my machines, and if it builds ok I'll merge this. Thanks for the fixes.

lgritz:
Merged.

PR #1373, opened 2016-02-29
Fix support for UTF-8 filenames on MingW

MrKepzie:
See original pull-request: #1366

Less intrusive implementation which only requires user to use OIIO_NAMESPACE::ifstream / OIIO_NAMESPACE::ofstream instead of the original fix proposed in the pull request #1366.

This is cleaned-up in a single commit with other compilation fixes stripped-off

### DIFF
--- a/src/include/OpenImageIO/filesystem.h
+++ b/src/include/OpenImageIO/filesystem.h
@@ -45,6 +45,7 @@
 #include <cstdio>
 #include <ctime>
 #include <fstream>
+#include <cassert>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


lgritz:
Looks good!

I'll do a few tests and then merge. Thanks for the extra work, I think this is a much better solution.

MrKepzie:
ok! Still have to print the CLA... they switched from xerox printers to something else last month, it's now hard to print anything here :>

lgritz:
I never print anything out just to sign it anymore, just directly paste a scan of my signature on PDFs. I like PDFPen (https://smilesoftware.com/pdfpenpro) for my PDF editing needs, but I'm sure there are also many other programs that will do this. (I'm not saying this to make you spend $50 just to get me a signature, just letting people know about a program I get a lot of good use of myself, since you mentioned the topic.)

Issue #1375, opened 2016-03-04
[bug] exr_threads attribute wrongly set


MrKepzie:
The exr_threads attribute is by default set to 0, as commented in the header:

```
///     int exr_threads
///             The size of the internal OpenEXR thread pool. The default
///             is to use the full available hardware concurrency detected.

```

Unfortunately, 0 in OpenEXR means *single-threaded* and not full available hardware concurrency, as seen in ImfThreading.h:
```

//	Threading in the Imf library is controllable at two granularities:
//
//	* The functions in this file [...] threads to perform all work.  The default number of
//	  global worker threads is zero (i.e. single-threaded operation;
//	  everything happens in the thread that calls the library).
//

```
Also, taken from *IlmThreadPool.cpp* , the global-thread pool used in OpenEXR by default is initialised with 0 as numThreads:

```
ThreadPool&
ThreadPool::globalThreadPool ()
{
    //
    // The global thread pool
    //
    
    static ThreadPool gThreadPool (0);

    return gThreadPool;
}
```

And in *ThreadPool::addTask*, you can see that if the **numThread** is 0, then it will execute things in the current thread:

```
    if (_data->numThreads == 0)
    {
        task->execute ();
        delete task;
    }
```

If you want I can make a pull request, the fix is trivial

MrKepzie:
@lgritz Actually we need to tweek this parameter, because one needs to be able to set the number of threads of OpenEXR thread pool to 0 to actually *clean* the thread-pool or just to disable the thread-pool.

*ThreadPool::setNumThreads(0)* is equivalent to the desctructor of the *ThreadPool::Data* which calls *ThreadPool::Data::finish()*. 

We have this particular need because we encountered a bug where if OpenEXR was used from a DLL 
loaded as a plug-in inside the main-application, the static thread-pool would actually get killed in such order that *pthread_join(...)*  would fail. 

We have not been able to reproduce this bug when making a small unit test directly using OpenEXR, but when we would actually call *LoadLibrary* on a DLL doing that exact same unit test it would always fail.

The work-around we found is to kill the threads of the thread-pool *before* the static variable destruction. The only way to do this is to call *setNumThreads(0)*.

Anyway, I think the **exr_threads** attribute should behave as such:

- -1: Disable thread-pool, everything runs in the caller thread (as it was prior to my pull request)
- 0: Use all cores to remain consistent with the definition of *OIIO_THREADS*
- *X*: Use *X*  threads

[...]

PR #1376, opened 2016-03-04
Fix exr_threads attribute

MrKepzie:
This PR fixes the following issue https://github.com/OpenImageIO/oiio/issues/1375

### DIFF
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -85,6 +85,7 @@
 #include "OpenImageIO/filesystem.h"
 #include "OpenImageIO/imagebufalgo_util.h"
 #include "OpenImageIO/deepdata.h"
+#include "OpenImageIO/sysutil.h"
 
 #include <boost/scoped_array.hpp>
 
@@ -319,7 +320,9 @@ void set_exr_threads ()
 
     int oiio_threads = 1;
     OIIO::getattribute ("exr_threads", oiio_threads);
-
+    if (oiio_threads == 0) {
+        oiio_threads = Sysutil::hardware_concurrency();
+    }
     spin_lock lock (exr_threads_mutex);
     if (exr_threads != oiio_threads) {
         exr_threads = oiio_threads;


src/openexr.imageio/exrinput.cpp
```diff
@@ -319,7 +320,9 @@ void set_exr_threads ()
 
     int oiio_threads = 1;
     OIIO::getattribute ("exr_threads", oiio_threads);
-
+    if (oiio_threads == 0) {
+        oiio_threads = Sysutil::hardware_concurrency();
```
lgritz: The default value (if they never set the attribute) is physical_concurrency(), which is the number of true cores. The hardware_concurrency, on some processors may be a higher number of virtual (hyperthreaded) cores.

It's hard to know which is "best" as the default. But I think they should be the same whether it's the default or explicitly set to 0.

Do you have strong feelings about hardware_concurrency versus physical_concurrency? My rationale was that using h_c rarely had a big gain in performance, but took the chance of "over-threading", so I erred on the side of conservatively picking p_c. But I could be talked into a different position.

lgritz:
LGTM, other than my comment on making sure we do either hardware_concurrency or physical_concurrency consistently in both places.

src/openexr.imageio/exrinput.cpp
MrKepzie: I never actually profiled testing *virtual* vs. *physical*, but for instance here I have 2 Xeon with 6 hyper threaded core each, which means overall 24 threads. Using physical would only ever use half of the processor, and would never exceed 1200% in **htop**. 

I'm not sure you need to worry about *over-threading*, basically as long as the processor is capped, this is because it is doing something useful: everything that OpenEXR is doing with these threads is uncompressing data.

 I think it's then the responsibility of the application using OIIO to actually handle this correctly, OIIO should just allow full access to all threads.

lgritz: OK, makes sense to me. I'll merge this.

Do you think I should change the default (if the user never sets the attribute) to also be hardware_concurrency rather than the current physical_concurrency?

Have you actually benchmarked 12 vs 24 threads on your system?

lgritz:
Fixes #1375

src/openexr.imageio/exrinput.cpp
MrKepzie: I think we should use hardware_concurrency everywhere indeed, as said this is the responsibility of the client to use it correctly.
I'll benchmark 12vs24 next week if I get time and get back on the topic, but I'm pretty sure OpenEXR handles this well because there's only ever 1 thread doing system I/O in the library

fpsunflower:
+1 from me for using `hardware_concurrency` instead of `physical_concurrency` everywhere

PR #1377, opened 2016-03-07
SIMD enhancements

lgritz:
Several miscellaneous simd.h enhancements

* addnot wraps the _mm_andnot_ps/_mm_andnot_si128 (with fallback when SSE is not available)

* Slight tweaking of madd to be extra sure we get fma instructions with clang (requires direct use of the SSE registers and a compilation flag; gcc seems smarter).

* Template SimdSize<T>::size reveals the SIMD vector size of a type, or 1 for any type that isn't one of our simd types. Handy for certain kinds of template metaprogramming where the type might be a simd type or might be a scalar.


CMakeLists.txt
```diff
@@ -292,6 +292,10 @@ if (NOT USE_SIMD STREQUAL "")
             else ()
                 add_definitions ("-m${feature}")
             endif ()
+            if (feature STREQUAL "fma" AND CMAKE_COMPILER_IS_CLANG)
+                # Clang needs an explicit flag to make fma ops
+                add_definitions ("-ffp-contract=fast")
```
fpsunflower: What else does this let the compiler do? This can break other code that relies on IEEE semantics I think.

src/include/OpenImageIO/fmath.h
```diff
@@ -228,13 +228,20 @@ clamp (simd::float4 a, simd::float4 low, simd::float4 high)
...
+    // explicitly asked not to, clang will do so if -ffp-contract=fast.
     return a * b + c;
 }
 
+#if OIIO_SIMD_SSE
+template <>
+inline simd::float4 madd (const simd::float4& a, const simd::float4& b,
+                          const simd::float4& c) {
+   // Directly access the underlying __m128 to ensure a true fma
+   return a.simd() * b.simd() + c.simd();
```
fpsunflower: Since this is SIMD code anyway, why not just invoke the FMA intrinsic?

lgritz: As far as I can tell, there is no preprocessor symbol that lets me know when it has been enabled.
So at the source code level, I don't know when it's safe to use the intrinsic.

CMakeLists.txt
lgritz: No, fma does not break IEEE, there is no loss of precision. This directive only controls the conversion of a*b+c to fma, it's not like -ffast-math. It's also only necessary on clang, which for some reason (that many would consider a bug) does not automatically use fma even with -mfma is used, whereas gcc does. I expect this will be fixed in some future clang.

fpsunflower: FMA on its own improves precision, but not when taken as a part of a larger expression.

I don't think its that simple. Reading online quickly -- fma _can_ reduce precision, for example when you multiply a complex number by its conjugate (because you end up with something like `a*b-b*a`. You would get 0 normally (and expect to mathematically), but switching to FMA will break the expected cancellation.

Another good example is the way you have to write a LERP using fma's to get the desired properties:
https://fgiesen.wordpress.com/2012/08/15/linear-interpolation-past-present-and-future/

I don't think a compiler can figure it out on its own, and the "old" lerp would start breaking if an FMA was randomly inserted for half the expression.

I would much rather opt-in to FMA explicitly than have it be inserted by the compiler.

lgritz: None of this takes effect unless you build with the -mfma (triggered by fma being one of the options in USE_SIMD=...).

Lerp -- oy, isn't there ever any good news where new hardware makes things easier for programmers?

fpsunflower: Since these functions are all in public headers, whatever decision we make here could be propagated to all users (or they would get worse performance).

Thats why the opt-in behavior seems safer to me. It lets this code take advantage of FMA when enabled, but also lets client programs build without `fp-contract=fast`.

Do you have any good pointers that document the difference between `fp-contract=fast` and `fp-contract=on` ?

I would also love to re-measure the performance/accuracy of our fast math approximations with FMA enabled.

lgritz: It *IS* opt-in, it only is triggered by building with USE_SIMD=fma, which does not happen by default.

Note that it's only clang that needs this fp-contract business. For gcc, once -mfma is specified on the command line, the compiler will automatically turn a*b+c into fma instructions. So presumably, smarter people than us have already grappled with this and decided it's safe.

As a matter of fact, I was fooling around with some time trials this weekend in the presence of -mfma, which is how this patch came about. Most of our workstations at work don't have the hardware to support it, but on my laptop it showed appreciable differences for certain benchmarks. Enough to make me wish we had fma everywhere so we could enable it for production builds.

fpsunflower: But you either opt-in for the entire code-base, or nothing at all.

There is no way to get the benefit of FMA only for the fast math functions for example.

lgritz: I can kinda get you there, but imperfectly.

The place in the CMakeLists.txt where we parse the USE_SIMD options, in addition to setting -mfma, I could also -DFMA_ENABLED=1 -- that is, make my own preprocessor symbol for the one that is lacking.

Then, we could create a madd_aggressive() that would use fma instructions any time FMA_ENABLED is true, even if -ffp-contract=fast is not turned on, whereas if you further turned on -ffp-contract=fast you would get it potentially everywhere because the compiler would look for places to do it automatically.

Now, there are two problems with this approach:

1. While this would work for our own codebase, non-OIIO apps that liked to use fmath.h or simd.h for convenience would not end up with fma from ordinary madd_aggressive unless they also took the effort to set the special symbol and the correct compiler flags.

2. gcc currently has -ffp-contract=fast turned on by default any time -mfma is enabled, so you'd need to explicitly use -ffp-contract=off if you wanted to ensure that fma's are ONLY used for madd_aggressive and not anywhere else.

You asked earlier about -ffp-contract=on versus fast. The answer is that "on" is not currently implemented, and is thus the same as "off": https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

fpsunflower: I would be fine with upgrading all uses of `madd` (i.e.: I don't see a need to make a seperate `madd_aggressive`), because we already use it only in specific parts of OIIO.

It just comes down to the risk vs. reward of turning on FMA blindly in users's code (for example what it does to `lerp`).

From what I saw on the `clang` mailing list, it seemed like there was hesitation to follow GCC's default. To be fair, the standard makes little guarantees with any of this - so anything is fair game from the compiler side.

My main worry is that once `fma` is on, if the compiler is allowed to swap it in anywhere it likes (or not) you could end up having subtle behavioral differences between opt and debug, or even between opt builds based on code movement around the expressions. It some ways it ends up bringing back the undesirable behavior of the old x87 FPU that had mixed precision depending upon what was written back to memory.

All that being said -- I still very much want to take advantage of FMA where possible ... I just wish there was a way to ''guarantee'' where it gets applied.

fpsunflower: I played a little in compiler explorer, and gcc seems fairly good about consistently using fma where possible (even with `-O0`).

But my concerns about cases like `lerp` remain. We would be loosing the guarantee that:

`lerp(x, a, b) == lerp(1 - x, b, a)`

Which I think some code implicitly assumes.

lgritz: Do you think we should try to detect use of -mfma, and when detected, substitute an implementation of lerp that uses the alternate construct discussed in that page you cited earlier?

fpsunflower: On my machine, `gcc -mfma` add the pre-processor symbol `__FMA__=1`. Maybe we could use that to detect when `-mfma` was specified on the command line? (It also has defines for AVX, SSE4.1, POPCNT and others that are implied by `-mfma`).

If we can detect it this way -- maybe we can go the intrinsics route and worry about changing the default of `fp-contract` another day?

As an example of how much leeway the compilers have, you can compare this snippet [1] with clang and gcc under various versions and under various flavors of fp-contract.

[1] https://goo.gl/NOvwhR

lgritz:
Updated after learning that there is a __FMA__ preprocessor symbol after all! And in discussion with Chris, consensus was to be safe for now by telling the compiler not to emit fmadd ops unless explicitly asked to. We can revisit in the future. Also, for symmetry, I added nmsub.

src/include/OpenImageIO/simd.h
```diff
@@ -2347,6 +2370,61 @@ OIIO_FORCEINLINE float4 max (const float4& a, const float4& b)
...
+#endif
+}
+
+
+
+/// Fused multiply and add: (a*b + c)
+OIIO_FORCEINLINE float4 madd (const simd::float4& a, const simd::float4& b,
+                              const simd::float4& c)
+{
+#if OIIO_SIMD_SSE && __FMA__
```
fpsunflower: What about creating a special `OIIO_FMA_ENABLED` macro that digests the compiler specific `__FMA__` into something portable? Just thinking about MSVC ...

src/include/OpenImageIO/fmath.h
```diff
@@ -225,16 +225,38 @@ clamp (simd::float4 a, simd::float4 low, simd::float4 high)
...
+/// Fused multiply and add: (a*b + c)
 template <typename T>
 inline T madd (const T& a, const T& b, const T& c) {
-    // NOTE:  in the future we may want to explicitly ask for a fused
-    // multiply-add in a specialized version for float.
-    // NOTE2: GCC/ICC will turn this (for float) into a FMA unless
-    // explicitly asked not to, clang seems to leave the code alone.
+    // NOTE: GCC/ICC will turn this (for float) into a FMA unless
+    // explicitly asked not to, clang will do so if -ffp-contract=fast.
     return a * b + c;
```
fpsunflower: What about using instrinsics to turn this into FMA as well? (When FMA is on)

lgritz:
New update:

* use OIIO_FMA_ENABLED (triggered by either __FMA__ or __AVX2__)
* try to use actual std::fma(), which is presumably the correct intrinsic, when C++11 is available (I can't find a reliable intrinsic to use without C++11)
* more symmetry: add msub and nmadd, so now we have the full complement of madd, nmadd, msub, nmsub.

fpsunflower:
Without C++11 can't you just use `_mm_fmadd_ss` ?

lgritz:
I'll try that

lgritz:
Update: use intrinsics directly when available.

lgritz:
OK with this as it ended up?

fpsunflower:
Yes - sorry, must have forgotten to hit send.

LGTM!

PR #1379, opened 2016-03-07
Improve filters

lgritz:
Sometimes you pull on one little thread, and the sweater starts to unravel. Remember to assume that: (1) untested code is buggy code, and (2) untimed code is slow code.

* Add a unit test for filters that creates a graph of the shapes of all   our filters, and benchmarks them. Guess what -- some are much slower than others, and some don't look quite like I thought!

* Lanczos3 was found to look awful, with a weird divot near 0 due to the fast_cos  approximation.  One call to fast_cos apparently is ok, but multiply the two fast_cos's  against each other and they had a very strange shape near the f(0)  area. Compromised by going back to a full precision cosine, but using  trig identities to cut down on the number of cosine calls.

* Speed up Blackman-Harris by using trig identities.

* Scale down catmull-rom, it was the only one we had that didn't have a  maximum value of 1. Technically, that's fine, since all conceivable uses  of the filter would normalize a set of samples. But when graphing it, it  was not helpful for it to be scaled differently than all the others.


fpsunflower:
How about using `fast_sin` and `fast_cos`? Those should have much better precision.

src/libutil/filter.cpp
```diff
@@ -230,9 +230,10 @@ class FilterCatmullRom1D : public Filter1D {
         x = fabsf(x);
         float x2 = x * x;
         float x3 = x * x2;
-        return (x >= 2.0f) ? 0.0f :  ((x < 1.0f) ?
+        float y = (x >= 2.0f) ? 0.0f :  ((x < 1.0f) ?
```
fpsunflower: Might as well fold the overall scale into the coefficients?

src/libutil/filter.cpp
```diff
@@ -289,8 +281,20 @@ class FilterBlackmanHarris1D : public Filter1D {
...
+#if 0
+        // original -- three cos calls!
         return A0 + A1 * cosf(2.f * m_pi * x) 
              + A2 * cosf(4.f * m_pi * x) + A3 * cosf(6.f * m_pi * x);
+#else
+        // Use trig identintities to reduce to just one cos.
+        // https://en.wikipedia.org/wiki/List_of_trigonometric_identities
+        // cos(2x) = 2 cos^2(x) - 1
+        // cos(3x) = 4 cos^3(x) − 3 cos(x)
+        float cos2pix = cosf(2.f * m_pi * x);
```
fpsunflower: I believe `fast_cospi` works fine for this one.

lgritz:
fast_sin is more accurate than fast_sinpi?
I guess I thought they were the same other than the factor of pi. No?

src/libutil/filter.cpp
```diff
@@ -289,8 +281,20 @@ class FilterBlackmanHarris1D : public Filter1D {
         const float A2 =  0.14128f;
         const float A3 = -0.01168f;
         const float m_pi = float (M_PI);
+#if 0
```
fpsunflower: If you nuke the line that does `x=(x+1)/2`, you can flip all the constants to be positive.

fpsunflower: Yeah the functions are maybe poorly named `fast_sinpi` shoould really be `veryfast_sinpi`. But `fast_sin` and `fast_cos` are close to full precision.

lgritz: I'll add a note in the comments to clarify this.
I benchmarked on both Linux and Mac and found fast_sin to be no faster than std::sin. FWIW.

src/libutil/filter.cpp
lgritz: You know, I think I'm going to try another approach with catrom. I'm going to revert that to the original (widely published) formula, and merely change my graphing approach to rescale it

lgritz:
Updated:
* Revert catmull-rom, normalize my graph instead.
* Fix the filter list, "catrom" -> "catmull-rom". The list should have names that match filter->name()!
* Comment in fmath about fast_sinpi being faster and less accurate than fast_sin.

fpsunflower:
LGTM

PR #1381, opened 2016-03-08
exr_threads: 0 means hardware_concurrency, -1 means no thread-pool

MrKepzie:
**exr_threads**: use -1 to disable usage of OpenEXR thread-pool (and to clean its threads)

See my comment in https://github.com/OpenImageIO/oiio/issues/1375


lgritz:
LGTM, I will merge.

By the way, it's ok to just skip straight to the PR for something like this, there is no reason to file a separate issue and then immediately a PR.

lgritz:
Sorry, forget the last thing I said. I didn't realize you were appending to an existing issue.

PR #1384, opened 2016-03-17
simd exp, log, pow (exact and approximate/fast versions), and speed up sRGB<->linear

lgritz:
simd exp, log, pow (exact and approximate/fast versions)

Also add code to simd_test to not only verify that they work, but also
benchmark them.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -2288,6 +2288,17 @@ OIIO_FORCEINLINE float4 abs (const float4& a)
 #endif
 }
 
+
+
+/// Return 1.0 for each component >= 0, -1.0 for each negative component.
+OIIO_FORCEINLINE float4 sign (const float4& a)
+{
+    float4 one(1.0f);
+    return blend (one, -one, a < float4::Zero());
+}
+
+
+
 /// Per-element ceil.
 OIIO_FORCEINLINE float4 ceil (const float4& a)
 {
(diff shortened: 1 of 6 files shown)


src/include/OpenImageIO/simd.h
```diff
@@ -2469,6 +2480,132 @@ OIIO_FORCEINLINE float4 nmsub (const simd::float4& a, const simd::float4& b,
...
+#else
+    return float4 (std::log(a[0]), std::log(a[1]), std::log(a[2]), std::log(a[3]));
+#endif
+}
+
+
+
+// Full-precision pow(x,y) of all lanes of a SIMD vector.
+OIIO_FORCEINLINE float4 pow (const float4& x, const float4& y) {
+    return exp(log(abs(x)) * y);
```
fpsunflower: Inserting the abs is a fairly big approximation, for instance `pow(-3.0, 3.0)` would return a positive value which may not be expected. It all depends on the use case though ...

Maybe it would better to use this formulation "directly" in the spots you want to optimize?

fpsunflower:
Looks good otherwise, are you able to hit a 4x speedup in microbenchmarks?

src/include/OpenImageIO/simd.h
lgritz: That might be an oversight. I see that for fast_pow (in fmath.h), I multiplied by sign(x). I think I meant to do that for the full-precision pow as well.

lgritz:
It's a complex landscape, easier to just report results than try to characterize it simply.

Here's what I'm getting on my laptop. This is without fma enabled, which really speeds up the madds.
The ones without "simd" in the name are scalar.


 expf:   171.2 Mvals/sec
 fast_exp:   135.3 Mvals/sec
 simd::exp:   397.6 Mvals/sec
 simd::fast_exp:   635.7 Mvals/sec
 logf:   169.0 Mvals/sec
 fast_log:   167.6 Mvals/sec
 simd::log:   358.3 Mvals/sec
 simd::fast_log:   627.4 Mvals/sec
 powf:    64.1 Mvals/sec
 simd::fast_pow:   252.3 Mvals/sec

fpsunflower:
Wow - that seems to be super-linear in some cases! (Though I'm sure the way is benchmarked might have something to do with that).

`fast_safe_pow` is doing more than just looking at `sign(x)` -- its also checking if `y` is an odd or even integer. Otherwise `pow(-3.0, 2.0)` would be wrong. For fractional powers it returns 0 for negative `x` (so that `pow(-3.0, 0.5)` works as expected).

But this is all getting closer to more of a `safe_pow` type of behavior. The question is what the use case will be. If `x` is always going to be positive, you might as well just use `exp(log(x)*y)` directly and not even bother making a wrapper?

fpsunflower:
It looks like some platforms are having trouble the inline asm incantation in the benchmark script.

lgritz:
inline asm -- Yes, I have a fix for that in progress. :-)

lgritz:
Crap, do I still have pow wrong? 

The use case I had in mind for short-term use was actually linear<->sRGB conversion. So I'm really only concerned with positive x inputs (or at least, clamping x to a minimum of 0 is fine).

I could get rid of the full-precision simd::pow for now, return to it if we need it.

I could rename fast_pow to fast_pow_positive or something, and document it as working only for positive inputs?

Or leave the exp and log, and leave it to the individual use cases how careful they want to be. Ugh.

lgritz:
Or should I try to fully duplicate fast_safe_pow in simd?
I can try that and see how bad the slowdown is with all the special case handling. But I wanted something really fast.

fpsunflower:
Thats what I suspected the next use-case would be. My vote would be to just implement `exp` and `log` and use those directly for the sRGB math.

fpsunflower:
I think the corner cases in `pow` really preclude a super fast implementation. So you would either end up with lots of variants or making assumptions that don't fit with some use cases.

So I think leaving it out is acceptable.

For OSL (`fast_safe_pow`), we already have those somewhat "loose" math semantics, so it was easier to bake in some assumptions (although there is still definitely a cost to them). But its definitely just wasted cycles for something like gamma conversion.

lgritz:
Yes, I think you are correct. So my plan is:

* remove simd pow and safe_pow, since they are not correct.

* use the clamp/exp/log directly in the srgb code.

* if it comes up in other contexts that make us think we should have something in simd.h or fmath.h, we can cross that bridge when we come to it.

fpsunflower:
Sounds good!

lgritz:
Amended as follows:

* Removed full-precision pow, which was incorrect.

* Altered and renamed approximate pow that only works for positive inputs: fast_pow_pos(). This can be handy a number of places.

* Amended this review with the changes that were the original use case, which is SIMD evaluation of sRGB <-> linear conversion.

fpsunflower:
LGTM

PR #1392, opened 2016-03-28
WIP Support random access writes to subimages/MIPmaps.

lgritz:
DESIGN REVIEW

This adds to ImageOutput a new version of write_scanlines() and
write_tiles() that takes subimage and miplevel designations for formats
that support true random access to subimages/miplevels without the need
to write a subimage or miplevel to completion before moving on to another.

We use supports("random_subimage_access") to designate ImageOutput
plugins that support this.

The presumed use case is for a renderer that is generating many AOVs
simultaneously, each one to a separate subimage in the same file (in
OpenEXR parlance, to different "parts") and as it renders scanlines or
buckets (tiles), wants to write all the AOVs for the current bucket to
their respective subimage, without needing to buffer all the AOVs for
the entire image.

Most ImageOutput implementations are not expected to support this
feature.  In fact, the only format I'm aware of that is capable of
supporting it is OpenEXR.

As of this checkin, the OpenEXR support as not yet been added. This is
just a conceptual prototype for review of the API.

### DIFF
--- a/src/doc/imageoutput.tex
+++ b/src/doc/imageoutput.tex
@@ -811,6 +811,18 @@ queried using:
             ...
 \end{code}
 
+Generally, a subimage and/or MIP-map level is written to completion before
+the next {\cf open()} call designates the start of the next subimage or MIP
+level (assuming that the file format supports those concepts in the first
+place). For some file formats, their \ImageOutput may support truly random
+write access across subimages or MIP levels via a variants of {\cf
+write_scanlines()} and {\cf write_tiles()} that take scanline and MIP
+level indices. Support for these calls may be detected by:
+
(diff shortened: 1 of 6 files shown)


lgritz:
Nobody commented, positively or negatively. I don't think anybody needed this badly. We sure don't. Only one file format I know of could directly support it. And in recent years, the leading renderers have mostly switched away from (for each bucket { for each pixel in the bucket ; for each AA sample in the pixel; }) which naturally wants to write one bucket to all AOVs, and instead turn the loops inside out (for each AA sample, for each pixel in the image...) that tends to want to write whole images at once. 

That's a long winded way of saying, abandoning this idea for now, closing.

PR #1394, opened 2016-03-30
Fixed oiio:ColorSpace typo

Shootfast:
The `oiio:ColorSpace` metadata attribute is misnamed, not sure what implications fixing this has.


lgritz:
LGTM, good catch.

lgritz:
I updated the testsuite output that changed as a result, and pushed the fix to master and RB-1.6.

PR #1395, opened 2016-04-01
Work towards getting LINKSTATIC=1 to truly work for all dependencies.

lgritz:
This is work started by Dan Wexler, and then handed off to me for a bit more generalization / refactoring.

A bunch of the cmake stuff is being refactored as we get this functional and in the process learn a bit more about cmake.

As of this commit, to build a static-only library, we still need to disable a number of specific features (at least on my laptop):
USE_FFMPEG=0 USE_FREETYPE=0 USE_OPENCV=0 USE_FIELD3D=0 USE_OCIO=0

This work isn't done, I want to eventually get it completely working for all dependencies and all formats. But this is a solid step in the right direction, want to checkpoint and integrate this much before moving on.


lgritz:
Plugged away at this over the weekend and it has expanded to be a bigger cmake script refactor that will make continuing work on LINKSTATIC=1 and EMBEDPLUGINS=0 better.

PR #1397, opened 2016-04-04
oiiotool --fixnan for deep files

lgritz:
Make IBA::fixNonFinite (and oiiotool -fixnan) work with deep images.

Also speed them up (for the non-deep case) by fixing a bug where, even
though it split the image for threads, each thread examined the whole
image instead of just its part. Oops.

Also modify oiiotool --stats for deep files to print locations of nonfinites.


fpsunflower:
LGTM

PR #1398, opened 2016-04-08
ColorConfig bug fix -- failed to find ColorProcessor for NoOp transformation

lgritz:
Just a bad variable declaration, inner scope masked outer scope. Led to maketx refusing to do a certain color transform.


fpsunflower:
LGTM

How did this ever work ?

lgritz:
Works fine if the color transformation you're asking for is not a "no op" transformation!

We had a show with two color space names, one was effectively an alias of the other, and somebody tried to transform between them. Should be harmless, but was getting an error message.

lgritz:
It's recently-changed code in that area, I think I introduced the bug with a semi-recent alteration to the logic in that function. Probably did not get noticed until I upgraded the facility default this morning.

PR #1399, opened 2016-04-09
Appveyor -- first stab

lgritz:
First stab at using Appveyor (like TravisCI, but for Windows). It *kinda* works, in the sense that I git it far enough that it installs several dependencies, and compiles most of OIIO. But it has link errors, and I've about hit the limit of what I know about Windows development. Since this can't hurt anything, I think it's best to have it reviewed, merge it, and hope somebody with more Windows + Appveyor experience can suggest the changes (assumed to be minor) to let it build fully. At which point, it will be an important component of making sure we keep changes to the software from breaking the Windows build.

This PR includes a bunch of minor changes to fix warnings that showed up with the MS compiler.

You can see the results of the run here: (I think that link should be publicly readable)

https://ci.appveyor.com/project/lgritz/oiio/build/lg-appveyor-59/job/phg4w2jt2qamlnbi

I'm open to suggestions! I truly have no no idea how to fix it.


brechtvl:
This looks like a mismatch between compiler flags, generally you must use the same /MT, /MTd, /MD or /MDd flag for compiling all libraries. The Blender libraries used are built with /MT, so to link with them you must also use /MT for OIIO instead of the default /MD.

The Blender OIIO build script uses /MT:
https://svn.blender.org/svnroot/bf-blender/trunk//lib/win64_vc14/OpenImageIO/

.appveyor.yml
```diff
@@ -0,0 +1,133 @@
...
+    #   PYTHON_DIR: "C:\\Python27"
+    - CMAKE_PLATFORM: "Visual Studio 14 2015 Win64"
+      PYTHON_DIR: "C:\\Python27-x64"
+      tbs_arch: "x64"
+      tbs_tools: "msvc14"
+      tbs_static_runtime: 0
+
+install:
+  - cinstall: python
+  - cinst ninja
```
c42f: What's the distinction between `cinstall` and `cinst` here?

.appveyor.yml
```diff
@@ -0,0 +1,133 @@
...
+  - ps: if($env:build_platform -eq 'x64') {
+          $env:vcvar_arg = 'x86_amd64';
+        }
+        else {
+          $env:vcvar_arg = 'x86';
+        }
+  - call "C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/vcvarsall.bat"
```
c42f: I avoided having to mess with vcvarsall by using the cmake visual studio generator - in that case cmake seems finds the compiler automagically.  Maybe you need it for the ninja generator though?

.appveyor.yml
```diff
@@ -0,0 +1,133 @@
...
+  - mkdir include
+
+  # ZLIB
+  # - svn co https://svn.blender.org/svnroot/bf-blender/trunk//lib/win64_vc14/zlib
+  # - set CMAKE_PREFIX_PATH=%CMAKE_PREFIX_PATH%;%APPVEYOR_BUILD_FOLDER%/deps/zlib
+  # FYI, alternate way to get zlib:
+  - appveyor DownloadFile https://github.com/madler/zlib/archive/v1.2.8.tar.gz
+  - 7z x v1.2.8.tar.gz -so | 7z x -si -ttar
+  - cd zlib-1.2.8
+  - cmake -G "Visual Studio 14 2015 Win64" .  -DCMAKE_INSTALL_PREFIX=C:/Sys -DBUILD_STATIC=%BUILD_STATIC% -G Ninja
```
c42f: Seems a little strange that there's both `-G "Visual Studio 14 2015 Win64"` and `-G Ninja` here.

.appveyor.yml
lgritz: The distinction is that I probably cut one line from one project's appveyor.yml, and the other line from a different project's appveyor.yml.

This is a truly Frankenstein's monster of build scripts. I literally don't know what 90% does.

I'm flying blind here, googling other projects' appveyor files and cutting and pasting the bits that look like they do things that my project needs. :-)

.appveyor.yml
lgritz: Ah, so this black magic line is not needed at all?

.appveyor.yml
lgritz: It does, doesn't it? I copied those lines from elsewhere. I saw some other project that needed zlib, cut the lines and pasted them into mine, pushed the tree, and went to appveyor to see how the build went -- if it appeared to install the package, I moved on to the next dependency without trying to understand how or why the previous one worked.

lgritz:
@brechtvl : so it's just a matter of having an add_definitions("/MT") in my cmake?

.appveyor.yml
c42f: Haha :-)

AFAICT from a search, `cinst` is a batch command alias of `choco install`, to install ninja via the chocolatey package manager.
When lines inside the `install` block are prefixed with `<interpreter>:` that's the script interpreter used to run them.  Presumably `cinstall` is the appveyor alias of chocolatey install... not sure though.

.appveyor.yml
c42f: Probably not.  It basically depends on whether any tools need to find the compiler `cl` in the path and platform (x86 or x64) environment vars in the environment.  With the visual studio generator cmake seems to find everything automatically (using the registry? or install path searching?  I'm not sure.)

c42f: BTW, the same probably goes for the powershell expression on the line above as well.

lgritz:
For the purposes of appveyor, should I prefer to do a static library build or a dynamic library build?
Seems like a lot of the issues just come down to doing it consistently.

brechtvl:
Yes, it's a matter of doing it consistently, besides that it doesn't matter here. In CMakeLists.txt you could do something like this, but it would need to be an option since not everyone want this.

```
set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MT")
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /MTd")
```

In the Blender OIIO build script we use -DCMAKE_CXX_FLAGS_RELEASE="..." and over the flags entirely instead of appending /MT.

The Boost and zlib libraries would need to be built with the same flags as well, it's not clear to me where those Boost libraries in the script are from.

Or alternatively you could build everything with /MD, but then you can't use the Blender libraries.

c42f:
> In the Blender OIIO build script we use -DCMAKE_CXX_FLAGS_RELEASE="..." and over the flags entirely instead of appending /MT.

Yeah, that's a good way to get everything consistent.  The same ABI issues exist with gcc, but to a much lesser extent (-std=c++11 vs c++98 is the only one which I've had to worry about myself).

lgritz:
OK, so I grabbed the Blender libraries just for convenience. Maybe I should jettison those in favor of building those dependencies myself, now that I have other examples of how to do so.

So, in a world where the Blender deps were not an issue, would it be better to build everything with /MT or /MD?

c42f:
Personally I'm inclined to build third party libs by hand, especially if they have portable build systems of their own and aren't too huge.  Large deps can lead to long build times, but the appveyor build cache could help you there if necessary.

No idea about /MT vs /MD sorry!

lgritz:
Got it!

With some help from the openexr mail list, among others, and a little bit of messing with this in my spare time, I was able to get all the important dependencies to build, fix the handful of compiler errors in OIIO itself, and get to a clean build.

It's a more extensive patch than I originally thought, but it works. At this point, it build the whole project cleanly, so therefore is already going to be a big help by preventing us from breaking the Windows build with new checkins. I don't have it running the testsuite as part of the appveyor process yet; I wanted to get the build checkpointed before moving on to running tests. Will do that after this is committed.

appveyor.yml
```diff
@@ -0,0 +1,160 @@
...
+
+install:
+  - cinstall: python
+  - ps: if($env:build_platform -eq 'x64') {
+          $env:vcvar_arg = 'x86_amd64';
+        }
+        else {
+          $env:vcvar_arg = 'x86';
+        }
+  - call "C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/vcvarsall.bat"
```
c42f: I'm not quite sure, but can you get rid of the ps and call commands above?

appveyor.yml
```diff
@@ -0,0 +1,160 @@
...
+  - mkdir deps
+  - cd deps
+  - mkdir lib
+  - mkdir include
+
+  # ZLIB
+  - appveyor DownloadFile https://github.com/madler/zlib/archive/v1.2.8.tar.gz
+  - 7z x v1.2.8.tar.gz -so | 7z x -si -ttar
+  - cd zlib-1.2.8
+  - cmake -G "Visual Studio 14 2015 Win64" .  -DCMAKE_INSTALL_PREFIX=C:/Sys
```
c42f: could use `-G "%CMAKE_PLATFORM%"`

appveyor.yml
lgritz: I'm not quite sure either. A lot of this whole process is black magic.

I'll comment these lines out and try one more build, just to see if it works. But if that breaks things, I'm going to commit what I have for now rather than try to understand why it didn't work.

I need to make a second pass over this all (including making it run the testsuite in addition to performing the build), but I'm really itching to commit what I have so far. I've been working on this in tiny bits of free time here and there for weeks.

c42f:
Sure, LGTM really.

appveyor.yml
lgritz: Fixed

appveyor.yml
lgritz: Yes, turned out that eliminating those lines entirely had no apparent ill effect.

PR #1400, opened 2016-04-12
Handle newer Ptex versions.

### LINKED ISSUE #1396: Find Ptex recipe out of date
### Problem

The latest Ptex has a file called PtexVersion.h. The existing Cmake recipe searches for versions in Ptexture.h which is no longer valid.

Describe the problem or suggestion here.

**Expected behavior:**

I expect that with an old Ptex, Ptexture.h will be searched, but a new Ptex will allow the search in PtexVersion.h

This patch below works for top of tree Ptex. My Cmake knowledge isn't yet good enough to do the right thing for new or old Ptex, but I need to be able to use new Ptex, not the ptex embedded in the OIIO sources. 

diff --git a/src/cmake/modules/FindPTex.cmake b/src/cmake/modules/FindPTex.cmake
index f8fdfc2..9294fcb 100644
--- a/src/cmake/modules/FindPTex.cmake
+++ b/src/cmake/modules/FindPTex.cmake
@@ -83,15 +83,15 @@ else ()
             DOC "The Ptex library")
 endif ()
 
-if (PTEX_INCLUDE_DIR AND EXISTS "${PTEX_INCLUDE_DIR}/Ptexture.h" )
+if (PTEX_INCLUDE_DIR AND EXISTS "${PTEX_INCLUDE_DIR}/PtexVersion.h" )
 
-    file(STRINGS "${PTEX_INCLUDE_DIR}/Ptexture.h" TMP REGEX "^#define PtexAPIVersion.*$")
+    file(STRINGS "${PTEX_INCLUDE_DIR}/PtexVersion.h" TMP REGEX "^#define PtexAPIVersion.*$")
     string(REGEX MATCHALL "[0-9]+" API ${TMP})
     
-    file(STRINGS "${PTEX_INCLUDE_DIR}/Ptexture.h" TMP REGEX "^#define PtexFileMajorVersion.*$")
+    file(STRINGS "${PTEX_INCLUDE_DIR}/PtexVersion.h" TMP REGEX "^#define PtexFileMajorVersion.*$")
     string(REGEX MATCHALL "[0-9]+" MAJOR ${TMP})
 
-    file(STRINGS "${PTEX_INCLUDE_DIR}/Ptexture.h" TMP REGEX "^#define PtexFileMinorVersion.*$")
+    file(STRINGS "${PTEX_INCLUDE_DIR}/PtexVersion.h" TMP REGEX "^#define PtexFileMinorVersion.*$")
     string(REGEX MATCHALL "[0-9]+" MINOR ${TMP})
 
     set(PTEX_VERSION ${API}.${MAJOR}.${MINOR})

lgritz:
Fixes #1396

Update FindPtex.cmake to the newer one from Disney's ptex-utils project, which accounts for a change in which header file contains the version number.

Also, remove the "bundled" ptex source files. This is more trouble than it's worth. Let's assume that if somebody building OIIO wants Ptex support, they should install Ptex on their system.

Can anybody think of a good reason why we would need to continue bundling ptex source?


fpsunflower:
LGTM

I think this is a good idea, we already have enough dependencies that trying to save one by bundling source is not really helping that much (and causes trouble for people who have to use the same version consistently everywhere).

And since we (SPI) don't use Ptex at the moment, it just helps streamline our builds too.

lgritz:
It was a badly outdated Ptex anyway. I'm not sure why we did it that way originally.

PR #1412, opened 2016-05-11
Fix jpegoutput aspect ratio

aoblet:
X_density and Y_density were swapped.

- PAR = xResolution/yResolution.
- So if we set yResolution to a constant C, xResolution becomes PAR*C
- We add + 0.5 for round.

### DIFF
--- a/src/jpeg.imageio/jpegoutput.cpp
+++ b/src/jpeg.imageio/jpegoutput.cpp
@@ -174,17 +174,20 @@ JpgOutput::open (const std::string &name, const ImageSpec &newspec,
         m_cinfo.density_unit = 2;
     else
         m_cinfo.density_unit = 0;
+
     m_cinfo.X_density = int (m_spec.get_float_attribute ("XResolution"));
     m_cinfo.Y_density = int (m_spec.get_float_attribute ("YResolution"));
-    float aspect = m_spec.get_float_attribute ("PixelAspectRatio", 1.0f);
-    if (m_cinfo.X_density <= 1 && m_cinfo.Y_density <= 1 && aspect != 1.0f) {
+    const float aspect = m_spec.get_float_attribute ("PixelAspectRatio", 1.0f);
+    if (aspect != 1.0f && m_cinfo.X_density <= 1 && m_cinfo.Y_density <= 1) {
         // No useful [XY]Resolution, but there is an aspect ratio requested.
         // Arbitrarily pick 72 dots per undefined unit, and jigger it to
         // honor it as best as we can.
-        m_cinfo.X_density = 72;
-        m_cinfo.Y_density = int (m_cinfo.X_density * aspect);
-        m_spec.attribute ("XResolution", 72.0f);
-        m_spec.attribute ("YResolution", 72.0f*aspect);
+        // XResolution / YResolution == PixelAspectRatio
+
+        m_cinfo.Y_density = 72;
+        m_cinfo.X_density = int (m_cinfo.Y_density * aspect + 0.5);
+        m_spec.attribute ("XResolution", (float) m_cinfo.X_density);
+        m_spec.attribute ("YResolution", (float) m_cinfo.Y_density);
     }
 
     m_cinfo.write_JFIF_header = TRUE;


aoblet:
#1413

lgritz:
Sorry for the delay. This particular bit of logic always makes my head hurt, so I procrastinated a bit in looking at this.

I'm not convinced that it was wrong before. 

If you have a pixel aspect ratio < 1, say 0.5, that means that pixels are tall and skinny, whereas a pixel aspect of 2.0, say, would have pixels wide and short, if you know what I mean.

Tall skinny pixels will have HIGHER xdensity than ydensity, because obviously there are more pixels per inch (or cm, whatever) horizontally than vertically, right?

That means that ydensity = xdensity * pixelaspect, as I had it before. Because aspect<1 should mean that ydensity < xdensity, and aspect > 1 implies that xdensity < ydensity.

If you think I'm wrong, please explain your logic, provide an example file for which our current software fails, or point me to a place in the JPEG/JFIF spec or libjpeg docs that says I've got this backwards. (Maybe I do.)

aoblet:
From [JPEG_File_Interchange_Format](https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format):
if units is 0: `pixel aspect ratio = Xdensity:Ydensity`

Currently in oiio, the code is:
`Y_density = X_density * aspect`
So:
`aspect = Y_density / X_density`

I agree with you that the field name suggests display density. But it seems that the parameter is considered as an inch density or cm density if `units == 1` or `units == 2` but it becomes a ratio if `units == 0`.

The jpeg writer code from Nuke is:
```
  cinfo.Y_density = 1200;
  cinfo.X_density = (UINT16)(iop->format().pixel_aspect() * cinfo.Y_density + .5);
```
The images generated from Nuke are correctly interpreted in Nuke and RV, but not the images converted with oiiotool.

lgritz:
This is exactly the head-hurting part I was talking about. :-)

Aspect ratios are always width/height. Like how a wide screen frame might a have an aspect ratio of 2.35. So a pixel aspect ratio of 2 means that the pixels are wide and short.

Density is the inverse of width or height, it's measured PER inch or cm. So pixel aspect ratio can be computed as ydensity/xdensity, which is the same ratio as xsize/ysize.

Here is the official JFIF spec: https://www.w3.org/Graphics/JPEG/jfif3.pdf

Note that the Wikipedia page is obviously copying from here, but has changed the wording slightly. In the JFIF spec (p. 6), it says:

    units        Units for the X and Y densities
                        units = 0: no units, X and Y specify the pixel aspect ratio
                        units = 1: X and Y are dots per inch
                        units = 2: X and Y are dots per cm
    Xdensity     Horizontal pixel density
    Ydensity     Vertical pixel density

It does NOT say that "pixel aspect ratio = Xdensity:Ydensity" as found in the Wikipedia article. In fact, reading through the history of that page, this specific edit was made on 15 June 2015. No explanation is given. I really, really think this is a typo or misunderstanding by the Wikipedia editor on that day. (I know, it's shocking that there can be a mistake on Wikipedia.)

You are right that the Nuke reader and writer (for those playing along at home with Nuke, see the `Documentation/NDK/examples/jpeg{Reader,Writer}.cpp` files) define 'aspect' as xdensity/ydensity.

I think that the mismatch is that this variable (or concept) in Nuke is misnamed. It's not the aspect ratio of the pixels, it's the CORRECTION FACTOR to convert to square pixels for all the internal compositing to work properly.

* When talking to OIIO, the metadata "PixelAspectRatio" always means pixel width size / pixel height size. And density means density!

* Internal to Nuke, this 'aspect' means how to scale to make square pixels. It's the inverse of what OIIO (and the rest of the world) mean by pixel aspect ratio.

Both Nuke and OIIO are each internally self-consistent (though I would argue that OIIO's terminology is more precise). The mistake is made by humans giving Nuke's correction factor to OIIO as the aspect ratio, or vice versa, without realizing that they are inverses of each other, just with confusing names.

I think. It would be great for somebody on the Nuke team to comment on this. 

I really want to avoid a situation where OIIO needs to contradict the clear wording of the JFIF specification just to be compatible with Nuke -- which is obviously a high priority for us given the VFX focus of most OIIO users.

Do you have a reference image you can send me? An anamorphic JPEG that was not generated by  either Nuke or OIIO, that we can look at in both and see if they agree?

lgritz:
Well, dammit, I just tried the test of generating a file

    oiiotool -pattern checker 512x512 3 -attrib PixelAspectRatio 0.5 -o check.jpg

and loading it in Nuke, and it is indeed showing up as WIDE, not NARROW.

Ugh.

lgritz:
When I do the same thing, outputting to an exr file, I get the correct appearance in Nuke: tall skinny checkers.

So it really does look like one of Nuke or OIIO is just plain backwards when it comes to interpreting the density fields in JPEG files.

lgritz:
Polling other apps...

On my Mac, Preview.app doesn't seem to honor the pixel aspect for either JPEG or OpenEXR -- when I create a file as described above, Preview just displays it as if it were square.

Gimp is just as oblivious.

Does anybody have PhotoShop and can try the experiment?

Other apps?

mjmvisser:
Hi Larry, I just had a colleague open the jpeg generated from this in Photoshop (CC 2015):

    oiiotool -pattern checker 512x512 3 -attrib PixelAspectRatio 0.5 -o check.jpg

It was stretched horizontally (WIDE), same as RV/Nuke. Oddly it also complained that the image was "damaged", then popped up a second dialog warning that aspect ratios were for viewing purposes only. But the opened image seemed fine.

lgritz:
So, thus far we have three widely-used VFX packages that appear to compute (for JPEGs) aspect = xdensity/ydensity, despite the JFIF spec and common sense about what "density" means, and every other app we've tried ignores the aspect entirely.

I guess we need to do whatever Nuke/rv/PhotoShop do, but it just kills me to have to make the code be so clearly wrong. I would feel so much better if somebody with deep knowledge of JFIF would tell me why my thinking about this is incorrect. :-)

If x pixels are spaced 100 per inch and y pixels are 200 per inch, then the pixels are wider than they are tall, giving an aspect ratio of 2.0, ydensity/xdensity. Aaagh!

lgritz:
OK, I'm fairly convinced: Nuke, PhotoShop, ffmped, and rv all agree, so I think we have little choice but to conform, regardless of how this seems to contradict the wording of the JFIF text.
I'll give this until tomorrow for more comment, but then I'll commit your fix if there are no objections.

aoblet:
They all agree indeed. But as we said perhaps because of a misunderstanding with "density" word/concept.
I've contacted The Foundry's support in order to get their feedback. The ticket is in progress.

lgritz:
Do you think I should merge now, or wait a few days for you to get an answer back?

I would be very surprised if they were willing to change, right or wrong, especially considering the array of other software that seems to be wrong in the same way.

aoblet:
Waiting a few days could be a good option. I gave to The Foundry the the url of this PR. If they answer me by mail, I'll let you know.

aoblet:
Some news from the Foundry :

(The Foundry)
May 24, 22:21

Hi Alexis,

Thank you for contacting The Foundry Support.

I have created a bug regarding this in our database:
TP 214991 Fix Jpeg output pixel aspect ratio

This will send a notification to engineering and will open and investigation.

lgritz:
Thanks, Alexis, you're being very accommodating of my initial objections.

I'm guessing that between the fact that PhotoShop also gets it backwards, plus the whole issue of how changing now would cause any previously-written-by-Nuke JPEG files to be interpreted wrong, it's very unlikely that they would want to change (and I wouldn't blame them). But we can wait until we get a preliminary response from the engineers, just in case they see it differently.

In fact, I think we have yet to identify a single (non-OIIO) app that behaves in accord with the JFIF spec. Unless we can find apps at least as popular/important as Nuke & PhotoShop that get it right, I think we're going to have to grit our teeth and all agree to be wrong in the same way. Following the spec to the letter, when all apps have been getting it backwards up to now, is just inviting more headaches, I'm sorry to say.

aoblet:
One month has passed without news from The Foundry... 
I'm sorry for that.

I think you could merge.

lgritz:
Yes, merged. (With some very minor edits, mostly to add comments that explain the situation.)

aoblet:
It seems that Solid Angle follows JFIF Spec: [Solid Angle doc](https://support.solidangle.com/display/NodeRef/options#options-aspect_ratio)

lgritz:
Solid Angle uses OIIO underneath.

The terminology here is confusing. OIIO adheres to the usual convention of aspect = horizontal_size / vertical_size. All apps that talk to OIIO use that convention when expressing aspect ratio, blissfully unaware of the fact that just in the case of JPEG files, OIIO is filling out the fields backwards in order to have the same interpretation of the density fields as Nuke and PhotoShop.

lgritz:
The SolidAngle docs are wrong about OpenEXR, incidentally. The EXR header ImfStandardAttributes.h is very clear:

    //
    // xDensity -- horizontal output density, in pixels per inch.
    // The image's vertical output density is xDensity * pixelAspectRatio.
    //

This is the usual definition. Aspect = ydensity/xdensity = xsize/ysize. That's what OIIO does, and always has done, on the app side of the interface. That's what the JFIF spec says, too. But it's not what Nuke and PhotoShop do when filling out the header fields of a JFIF file, so now we conform to the wrong way for JFIF, also.

But let's be very clear, apps can use OIIO's APIs without any knowledge of what file format is being used underneath. That's the whole point of OIIO. To any app using OIIO, aspect = xsize/ysize.

PR #1414, opened 2016-05-14
TIFF: Tighten up logic for non-zero origin of "full" size (aka display window)

lgritz:
TIFF is incapable of representing non-(0,0) origin "full" (or display) window. So there are some edge cases we need to handle more deftly:

* When writing to TIFF, if the ImageSpec says the full/display window is   nonzero origin, set it to 0 origin and expand the width&height   accordingly.  That's the best we can do, it's all TIFF can represent.

* When reading TIFF that does not say anything about the full window   at all, assume it should be interpreted as equal to the data window.

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -651,12 +651,6 @@ TIFFInput::readspec (bool read_meta)
         m_spec.width = (int)width;
         m_spec.height = (int)height;
         m_spec.depth = (int)depth;
-        m_spec.full_x = 0;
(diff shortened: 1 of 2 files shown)


rmv:
LGTM.

I think this might change the way maketx deals with data origins on tiff files. Currently it extends the mips to include (0,0). However that is really a borderline case.

The new behavior seems to match the "consensus" when loading those tiffs and the change will make things consistent across OIIO tools.

lgritz:
Merged into master, also into RB-1.6 (but not yet tagged as a release).

PR #1416, opened 2016-05-22
to_native_rectangle -- handle case with native data but non-contiguous strides

lgritz:
This function is used as a helper by various ImageOutput implementations.

If you passed data to write_scanline or write_tile with UNKNOWN format (indicating "native" data) but the data were noncontiguous (the strides didn't match fully packed sizes), you would previously hit an assertion in some circumstances.


rmv:
LGTM, I checked that it works for our use cases

PR #1417, opened 2016-05-25
ImageCache/TextureSystem: collect stats on redundant reads

lgritz:
A "redundant" read is when we read a tile that has been read before, but    had been evicted from cache and needs to be read again. We can use this    to measure effectiveness of the cache, detect thrashing, and narrow down    which files in particular are experiencing the kind of incoherent access    patterns that leads to redundant reads.

The statistics output now includes: the total number of redundant tiles    and bytes are printed, the per-file stats and totals say the number of    redundant tile reads and the total bytes (of uncompressed I/O) that it    represents, and at the end the "top 3" files are printed that    experienced the most redundant I/O.

This way this works basically involves keeping track of one bit per tile    that indicates if the tile was already read from disk. The extra storage    for a 4k x 3k MIP-mapped image with 64x64 tiles is 512 bytes, so even    with thousands of files, we're talking about a very modest memory    hit. The bit is set to 1 when a tile is read, and if it was already 1,    we increment a per-file counter of the number of redundant tiles and    bytes. This is done with atomics to guard against the the occasional    thread clash.

I needed an "atomic fetch and OR" operation, which did not previously    exist in my thread.h, so I added that, and also for the sake of    completeness, added the same for AND and XOR.

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -257,6 +257,144 @@ atomic_exchange_and_add (volatile long long *at, long long x,
 
 
 
+/// Atomic version of:  r = *at, *at &= x, return r
+/// For each of several architectures.
+inline int
+atomic_exchange_and_and (volatile int *at, int x,
+                        memory_order order = memory_order_seq_cst)
+{
+#ifdef NOTHREADS
+    int r = *at;  *at &= x;  return r;
+#elif defined(OIIO_USE_GCC_NEW_ATOMICS)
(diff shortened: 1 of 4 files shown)


fpsunflower:
LGTM, this doesn't look like its too much extra work to track.

lgritz:
No, it's a few dozen to a few hundred extra bytes per file (depending on the resolution), and only when a tile is actually read from there is a single atomic 'or', and two more atomic adds if it actually was redundant. The disk read will so totally dominate that case, it's really negligible cost.

PR #1418, opened 2016-05-26
oiiotool: --fullsize followed by --attrib could lose the fullsize

lgritz:
The --fullsize wasn't updating the spec properly in both places, and the
subsequent --attrib that re-copies some items from the underlying
ImageBuf's spec into the ImageRec's copy of the spec would overwrite the
previous changes of --fullsize. The --origin code path got it
right. Easy to adjust.


fpsunflower:
LGTM

PR #1420, opened 2016-06-01
Optimizations and UTF-8 Improvements

ElaraFX:
## Description

- Added debug_error to prevent texture system from printing too many messages, which slows down rendering significantly. Users' input data are usually imperfect.

- Removed is_regular check from Field3D input to align with other plugins.

- Removed exception throws from OpenEXR input, which were constantly causing crashes when file format is unknown to OIIO.

- Improved UTF-8 to Unicode handling in file system for Windows.

## Tests

- Manually tested.

------


lgritz:
Can you rebase this on top of the current master so we can see your diffs more clearly?

lgritz:
Never mind. The diffs are clear, it's just the history that's weird. I can rebase when I merge if it's trouble for you.

lgritz:
I'm worried that going from too many errors printed to none at all is going to be just as problematic, but in the other direction!

Maybe a good compromise is to have a TS/IC option that limit the number of error messages printed for each file. Add an error count to the ImageCacheFile, go back to the old error() calls, but for a few carefully-chosen places only issue the error message if the count++ doesn't exceed the limit?

But first, can we take a step back and make sure I understand the issue. What error message, under what conditions, were you getting too many of?

ElaraFX:
@lgritz 

The main problem we met are missing textures. Our users' scene usually have a few missing textures, while users don't really have time to fix those scenes manually, and it's unnecessary because those objects are usually invisible in the final images.

Current OIIO implementation generate a lot of error messages for missing textures, even with missing color supplied, the error message formating and geterror still affect the rendering performance. And many places in both texture system and image cache keeps on reporting the same missing texture issues. I think it's better to only report such issues once for each specific texture file, but keeping track of this can be performance bottlenecks too I guess.

So now although I know these are aggressive and dirty changes, I don't really have better idea for this. Maybe you can give me some insights here. Thanks.

lgritz:
Quite coincidentally, I just submitted this PR: #1423 

Can you check that out and see if it helps to eliminate some of the many error messages? In the particular case that I was addressing, it changes from a (not very helpful "(unknown error)" on every texture lookup) to a single helpful error ("Could not open foo.exr") just the first time each file tries to open.

That may not cover all cases, but let's see how much it takes care of. I could also easily make it simply put a cap on the number of error messages it issues for each file. The theory being, if you haven't gotten the point after 5 error messages about a file, it doesn't do anybody any good to see more, especially if they don't give any additional information not present in the initial error messages.

I do still like your other changes here -- fixing the thrown exception in the EXR code, and the UTF-8 stuff. I'll double check those and then merge them in isolation.

lgritz:
It's interesting that you say that missingcolor is not working. I'll look into that.

lgritz:
By the way, there is a reason I don't notice this -- where I work, we have it set up so that the renderer aborts on the first error (on the theory that it's better to error early and hard, rather than find out only after a 10 hour render that a texture was missing and therefore the shot is incorrect).

So I can tend not to notice if there's a situation where multiple errors are overly verbose or not helpful.

Please do point out such situations, and I'll try to fix them. They are not intentional.

ElaraFX:
@lgritz 

I just notice that the `geterror` function does clear error messages when it's called for the first time, so the error messages appear to print only once. So there seems to be no issue in this area.

However, I still have concerns with performance impacts of `geterror` and `error` routines, there seems to be string copies, formatting, and returning class objects... I don't want to have these calls for every texture lookup... We tested and compared the rendering performance with or without these calls about two years ago, so I cannot remember the exact data, but the impact is significant.

In our renderer, we never abort on the first error, because errors like missing texture is too common for our users... But I understand for a very processional team, it's better to abort on error as earlier as possible... We just don't have such teams here...

For the missing color feature, I suggest not print any error messages when `missingcolor` is supplied, because missing color should have already provided a **visual** warning for that case. But in `TextureSystemImpl::texture`, you can see that `verify_texturefile` is called **before** missing_texture call, and in `verify_texturefile`, error message is already printed, which seems to be not so good for rendering performance.

For other fixes not related to performance, I can send you a separate pull request. Do you want me to do that, or you think it's easier to merge them on your own?

lgritz:
Well, the original idea was that you shouldn't be getting errors for missing textures on every lookup. If every lookup is failing something is going very wrong and bad performance is probably the least of your problems.

But that said, if I implement the "print a maximum number of errors per file" as I was suggesting, then it seems fairly obvious that we check the count before the call to error(), and therefore the expense of the string formatting and whatnot should be avoided for all but the first few calls.

I agree that the intent of missingcolor is for the warning to be visual, and not have any errors issued. It sounds like there is an error in this case, I will look into that.

Let me see if I can merge your other changes (that I agree with) from what you gave me. If I can do that without trouble, there's no need to submit a new PR for it.

lgritz:
I updated #1423 with some additional approaches to suppress unhelpful error messages, including the addition of a "max_errors_per_file" attribute that you can set for the TextureSystem.

lgritz:
I looked into the missingcolor issue, and although you are correct that the missingcolor check is done after verify_texturefile, you will notice that in the implementation of TextureSystemImpl::missing_texture explicitly throws out any error messages when the 'missingcolor' is set:

```
bool
TextureSystemImpl::missing_texture (TextureOpt &options, int nchannels,
                                    float *result, float *dresultds,
                                    float *dresultdt, float *dresultdr)
{
    for (int c = 0;  c < nchannels;  ++c) {
        if (options.missingcolor)
            result[c] = options.missingcolor[c];
        else
            result[c] = options.fill;
        if (dresultds) dresultds[c] = 0;
        if (dresultdt) dresultdt[c] = 0;
        if (dresultdr) dresultdr[c] = 0;
    }
    if (options.missingcolor) {
        // don't treat it as an error if missingcolor was supplied
        (void) geterror ();   // eat the error
        return true;
    } else {
        return false;
    }
}
```

So I'm not convinced that missingtexture is broken.

Can you try out patch #1423 (and set the "max_errors_per_file" somewhat low, like 5) and see if you are still experiencing either too many errors printed, or any performance problems? I think that patch should probably solve most of the problem. If not, let's try to identify exactly what the case is that is generating the error.

lgritz:
OK, I looked more closely at the other commits in your patch.

Correct me if I'm wrong, but it looks to me that because commit ceaad58 ("All directory iterators should use the consistent code to handle UTF-8 strings") fixes the Filesystem::is_regular() function to handle UTF-8 names, the earlier commit (6b9cbc4f) where you remove a few scattered calls to is_regular elsewhere, is no longer necessary?

Also, commit c6d7ff0090 ("Fixed a build break on Windows caused by _MSC_VER") seems to be obsolete -- the code it fixes (the incorrect macro symbol _MSC_VER_) appears to no longer be in the current master or the head of RB-1.6. So I think that one can be abandoned.

Commit de21f19af2b ("Don't throw error...") looks good, I think.

So do you agree that if my proposed patch addressing errors (#1423) works for you, then the ones that are worth merging are de21f19af2b and ceaad58 only?

lgritz:
Ugh, sorry, one more look at de21f19 and I'm confused again. I'm worried about the consequences of removing those exceptions, if OpenEXR expects them and perhaps is expecting a thrown exception rather than returning false. 

If those exceptions are a problem, then perhaps it's because of a missing try/catch elsewhere? Perhaps the most robust solution is to get that right.

Do you have a stack trace under the conditions where it fails, so we can try to make a guess about which call we should be surrounding with try/catch?

lgritz:
I've merged the directory iterator fix. (About half of your patch related to UTF-8 filenames had already been implemented by someone else and merged a couple months ago! But you sill caught some remaining cases that they had missed.)

For the one about OpenEXR exceptions, I think that eliminating the exceptions is the wrong approach, but rather the correct strategy is to find out in which call the uncaught exception is occurring, and simply add the missing try/catch in the right place.

The others all appear to be things that have been fixed already in the master by other people, in the time since you forked, or which I think that my own patch #1423 addresses (the error stuff).

lgritz:
Have you had a chance to try out #1423 to see if that helps with your too-many-errors situation?

lgritz:
I've merged parts of this, and also my own #1423 that I think solves a lot of the error problems.
It's my belief that between the two of these, the original intent of this PR is fulfilled, so I'm going to close it. If there are further changes necessary, please file a new PR (against a fully up to date master).

PR #1423, opened 2016-06-02
Avoid needless verbosity for broken texture error messages

lgritz:


lgritz:
Pushed additional changes:

* New IC/TS attribute: "max_errors_per_file" puts a limit on how many
      error messages will be printed for each file. More than the first few
      almost certainly will not provide any useful additional information,
      but will clutter your logs and harm performance.
* Suppress printing of empty error messages.
* In testtex, make sure error printouts lock against each other to keep
      from making unintelligible console output when many errors are issued
      simultaneously from different threads.

lgritz:
Even though the person who inspired this has not replied with test results, nobody else has objected so I'm going to merge it.

PR #1424, opened 2016-06-02
Save space in ImageCacheFile by eliminating unused shadow matrix fields.

lgritz:
I put these fields here in the very early days of IC development,
anticipating that I would fully elaborate the shadow() lookup functions,
for which these matrices are needed. But it was never a priority, so
they are effectively unused, yet take up a lot of space (64 bytes * 4,
per ImageCacheFile).

I'm not quite brave enough to remove them entirely, out of superstition
that either somebody else is using a modified OIIO that implements
shadow maps, or that the very minute I remove them, some situation will
come up where they are important. So for now I will #ifdef them out.


fpsunflower:
LGTM!

PR #1426, opened 2016-06-03
UDIM support for textures.

lgritz:
Background: "UDIM" refers to an emergent convention in VFX for texture
atlases, where separate painted textures are applied to different parts
(faces or face sets) of a complex object. Instead of needing an
elaborate mapping of face IDs to arbitrary filenames, you simply use a
uv range extending beyond [0,1], with each unit square of texture space
corresponding to a different physical texture and a simple mapping of
tile indices to filenames based on a pattern.  In practice, there are a
few competing and subtly different conventions, and we try to support
the common ones.

This patch allows TextureSystem::texture() calls to reference "virtual"
filenames that get expanded to concrete files based on the specific
lookup coordinates.  These virtual filenames are identified on the basis
of (a) there not being a real file that exists in the filesystem with
that name, and (b) the filename containing one or more of the following
literal substrings: "<UDIM>", "<U>", "<V>", "<u>", "<v>".
(The angle brackets are part of the substring being searched, but the
double quotes are not.)

The coordinate transformation is as follows:

    utile = floor(u)
    vtile = floor(v)
    u = fmod (u, 1)
    v = fmod (v, 1)

And then the concrete filename is constructed using the following
substitutions of strings with decimal numbers:

    <U> -->  utile+1
    <V> -->  vtile+1
    <u> -->  utile
    <v> -->  vtile
    <UDIM> --> 1001 + utile + vtile+10

As an example, a texture lookup to "paint.<UDIM>.tif" at coordinates
(1.4, 3.8) will, under the covers, be translated into a lookup from
"paint.1032.tif" at coordinates (0.4, 0.8).  (The UDIM convention
is 1001+utile+vtile*10 == 1032 in this case.)

If this seems strange, suffice it to say that it's derived from widely
used conventions in texture painting systems used in VFX production.

This is done with minimal expense, the substitutions and decoding do not
need to be performed for each texture lookup. The basic strategy is:

* When the file is first opened, the name is checked for inclusion of
  any of the special tokens, and if found (and the filename doesn't
  really exist on the filesystem), then it's marked as one of these UDIM
  virtual textures.

* Internally, the virtual texture's ImageCacheFile record holds an
  associative array mapping tile ids (specifically: (vtile<<32) + utile)
  to the ImageCacheFile* of the concrete file for that section.

* The texture() call, if the file is a virtual UDIM texture, will look
  up the concrete texture handle. Only if this atlas tile has not yet
  been accessed (thus, establishing the mapping for the first time),
  does it do the string searching and formatting to construct the
  concrete filename, thus this expense is negligible in a render, rather
  than being part of the per-texture-lookup expense.

* This id->handle associative map for each virtual texture is
  implemented as a boost::flat_map, which is a sorted vector whose
  performance for small records (such as this) usually outperforms that
  of an unordered_map because of superior memory locality.

* The associative maps are kept thread-safe by a pool of reader/writer
  locks. Since it only needs to write the first time an atlas tile is
  used, it's almost always in "read" mode, and thus threads tend not to
  lock against each other. Even if multiple threads are trying to write
  simultaneously, the fact that it's a mutex_pool ensures that they are
  usually not accessing the same lock and therefore rarely block each
  other.


est77:
In our renderer, we flip the v coordinate (1.0 - v) when doing texture lookups using OIIO, for compatibility with our previous texture system.
Is there a way to do it when using UDIMs?

Thanks for the PR!

lgritz:
Yeah, I can do that.

If it's renderer-wide, I'm thinking that a TextureSystem attribute can control it. Do you agree? It doesn't need to be only for some textures but not others, right?

And would you prefer that the v flip just happen for UDIM textures, or would you prefer a TextureSystem-wide reversal of all lookups? (non-UDIM also)

est77:
> if it's renderer-wide, I'm thinking that a TextureSystem attribute can control it. Do you agree? It doesn't need to be only for some textures but not others, right?

Yes, I agree. An attribute that v flips all lookups would be perfect.
I think that the only case in which you would not want to do the v flip would be ptex textures.

Thanks!

lgritz:
I have submitted a separate PR for the v flip: https://github.com/OpenImageIO/oiio/pull/1428

The changes are orthogonal, you need both in order to test UDIM + v flip. It seemed cleaner to make them separate PRs so they can be merged into release branches separately, if need be.

est77:
Thanks for both pull requests. I'll try them as soon as possible.

lgritz:
Note: On #1428, @est77 reports that the UDIM patch seems to pass his basic tests.
I'll merge these when I get a chance and we can always return if there's further refinement needed to the UDIM work.

PR #1428, opened 2016-06-06
TextureSystem option "flip_t" will flip vertical direction of texture lookups

lgritz:
This is helpful for renderers that adhere to the convention that the t=0 texture coordinate is the visible "bottom" of the texture when displayed.


fpsunflower:
In the udim case, where does the flip occur? Per tile?

lgritz:
Yeah, udim resolution happens before the coordinate flip. So effectively, it's flipping the orientation of the texture within each file. but the tile IDs remain the same (that's what pairs the texture file with the geometry parts).

lgritz:
That is, when both this change and the udim one are merged. They are actually orthogonal, the merges don't conflict on any line!

est77:
I could finally test the PR (sorry for the delay). It works.
Thanks!

lgritz:
Hi, just to clarify, are you referring to just the flip_t stuff, or did you also try out the UDIM patch?

est77:
> Hi, just to clarify, are you referring to just the flip_t stuff, or did you also try out the UDIM patch?
I tried both together, I merged the PRs locally. The flip_t works. Our test suite passes.
I also tested UDIM and it seems correct, but I don't have any complex scene to test with, 
only simple things and only using the UDIM tag.

lgritz:
Awesome!

I'll merge both, then. If there are situations where the UDIM work is less than ideal, we can always adjust it later.

oliverbjames:
Shouldn't the results returned in dresultdt also be inverted if flip_t is true?

lgritz:
We flip the input dtdx and dtdy (see lines 1010). Does that not achieve this properly? No, I suppose not, since the computation of daccumdt (in the sample functions) doesn't appear to consider the sign of the input derivatives.

I believe you are correct. Stay tuned, I'll submit a fix today.

lgritz:
Fix for the bug @oliverbjames pointed out in #1462

oliverbjames:
That was quick!
Thanks, Larry

Issue #1429, opened 2016-06-09
Add support for 10-bit and 12-bit movie files


skycaptain:
### Problem

When reading 10-bit DPX files via `ImageInput` they get automatically upsampled to UINT16 type so that no information is lost. However, when reading a 10-bit Quicktime ProRes file via the included ffmpeg API it is interpreted as a UINT8 type. Therefore, information is lost.
I know that handling movie files is not of first priority, but since the ffmpeg plugin is already checking for pixel formats, adding 10-bit and 12-bit formats and a switch might be sufficient. 

### Versions

* OIIO branch/version: 1.6.13
* OS: OSX 10.11.5

lgritz:
Fixed by #1430, closing.

PR #1430, opened 2016-06-10
Added support for 10-bit and 12-bit movies

skycaptain:
## Description

This PR adds support for 10-bit and 12-bit movies (e.g. ProRes Quicktimes) to be read as 16-bit images (see #1429). This PR also includes some cosmetic fixes (e.g. strip trailing spaces) for `ffmpeginput.cpp`.

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -53,6 +53,7 @@ extern "C" { // ffmpeg is a C api
 #if LIBAVUTIL_VERSION_INT < AV_VERSION_INT(51,74,100)
 #  define AVPixelFormat PixelFormat
 #  define AV_PIX_FMT_RGB24 PIX_FMT_RGB24
+#  define AV_PIX_FMT_RGB48 PIX_FMT_RGB48
 #  define AV_PIX_FMT_YUVJ420P PIX_FMT_YUVJ420P
 #  define AV_PIX_FMT_YUVJ422P PIX_FMT_YUVJ422P
 #  define AV_PIX_FMT_YUVJ440P PIX_FMT_YUVJ440P
@@ -119,6 +120,8 @@ private:
     AVCodec *m_codec;
     AVFrame *m_frame;
     AVFrame *m_rgb_frame;
+    int m_stride;
+    AVPixelFormat m_dst_pix_format;
     SwsContext *m_sws_rgb_context;
     AVRational m_frame_rate;
     std::vector<uint8_t> m_rgb_buffer;
@@ -141,6 +144,7 @@ private:
         m_frame = 0;
         m_rgb_frame = 0;
         m_sws_rgb_context = 0;
+        m_stride = 0;
         m_rgb_buffer.clear();
         m_video_indexes.clear();
         m_video_stream = -1;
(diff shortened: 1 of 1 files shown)


lgritz:
Fixes #1429

lgritz:
This looks like the right approach to me.

I'll do some testing. 

Do you have any links to 10 or 12 bit movie files, so I can give it a spin?

skycaptain:
ARRI has some ProRes sample footage of their AMIRA camera which is in ProRes 422 (10-bit) and ProRes 4444 (12-bit): https://www.arri.com/camera/amira/learn/amira_sample_footage/.
Please note, ffmpeg currently has a bug, as it detects and interprets a 12-bit ProRes file as 10-bit. This has already been reported to upstream.

lgritz:
I have one request:

When a "non-whole" number of bits is promoted to the next word size (for example, 10 bit input changed to 16 bits for presentation to the app side of OIIO's APIs), it is customary to add metadata that communicates the original depth in the file:

    m_spec.attribute ("oiio:BitsPerSample", original_bit_depth);

(Examples include dpxinput.cpp, rlainput.cpp, tiffinput.cpp, and others.)

The reason it's helpful is that if you read input, then write it again to another file, you want to preserve the original bit depth, which otherwise you would lose track of. (The relevant output plugins will notice the metadata and, for example, revert the 16 bits it receives from the app back to 10 or 12 bits for final output.)

With the addition of this change, I am prepared to accept and merge this patch.

skycaptain:
Your're right. I will implement that. I'm also considering implementing a proper metadata interface for ffmpeg, since especially Quicktime files contain a lot of metadata, which may be forwarded to OIIO.

lgritz:
Yes, please! More thorough metadata reading from the movie files would be an extremely welcome addition. (But let's do that as a separate PR; there's no need to hold this part up for the more extensive metadata changes.)

Thanks.

skycaptain:
Of course. That was exactly my intention.

skycaptain:
This seams to be quit a bit more difficult than at first sight. avlib just provides the number of channels and the bits per pixel. The later describes the bits used for a decompressed, but not RGB-transformed image. E.g. a `yuv420p10le` image has 15 bits per pixel, while a `yuv444p10le` image has 30 bits per pixel. Both would use 10 bits per pixel transformed into RGB. This would also be the value I would except for `oiio:BitsPerSample`.

lgritz:
I agree, we want oiio:BitsPerSample to mean the number of significant bits per decompressed RGB channel. That may be different than the number of bits used in the color-transformed, compressed representation.

skycaptain:
I'm sorry that it took so long; I was quite busy the last weeks. I've added the BitsPerSample to the attributes. Apparently I was wrong, since libav actually exposes [bits_per_raw_sample](https://www.ffmpeg.org/doxygen/3.1/structAVCodecParameters.html#a796f5860107695d36c3a7fb30453da79), which should be equal to our BitsPerSample. But as said above, ffmpeg reads 12-bit ProRes as 10-bit. Therefore, I was only able to verify this for 8-bit and 10-bit files.

lgritz:
OK, LGTM. I'll merge into master.

lgritz:
merged

PR #1432, opened 2016-06-13
Fix to recent warning patch

lgritz:
Amends #1419

Slight error failed to mark files as "written" in some output cases where
a temporary image has to be made during output (the temp was marked as
written, rather than the original). This resulted in an incorrect warning
about "modified images without outputting them. Did you forget -o" when
clearly there was a -o.


rmv:
LGTM and checked that it works

PR #1434, opened 2016-06-15
Rethink default channel naming, just use R, G, B, A.

lgritz:
In particular, 1-channel images (from formats without channel naming) will now get R, 2-channel images get R, G.

It's not perfect, because it's probably not really "red" and "green." But before, we were setting the default names to "A" and "I","A".  Which is probably much much worse, because "A" means alpha, and lots of operations treat alpha as special (such as PNG output dividing all the non-alpha channels by the alpha value!).

This is just for formats that don't actually store channel names in the files. Which is most formats.

### DIFF
--- a/src/libOpenImageIO/formatspec.cpp
+++ b/src/libOpenImageIO/formatspec.cpp
@@ -155,37 +155,26 @@ void
 ImageSpec::default_channel_names ()
 {
     channelnames.clear();
+    channelnames.reserve (nchannels);
     alpha_channel = -1;
     z_channel = -1;
-    switch (nchannels) {
-    case 1:
-        channelnames.push_back ("A");
-        break;
-    case 2:
-        channelnames.push_back ("I");
-        channelnames.push_back ("A");
-        alpha_channel = 1;
-        break;
-    case 3:
-        channelnames.push_back ("R");
-        channelnames.push_back ("G");
-        channelnames.push_back ("B");
-        break;
-    default:
-        if (nchannels >= 1)
-            channelnames.push_back ("R");
-        if (nchannels >= 2)
-            channelnames.push_back ("G");
-        if (nchannels >= 3)
-            channelnames.push_back ("B");
-        if (nchannels >= 4) {
-            channelnames.push_back ("A");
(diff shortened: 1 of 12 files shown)


fpsunflower:
How about `L` and `LA` ? I think that would match the typical use cases better (unless there is a specific reason to try and be a subset of `RGBA`).

fpsunflower:
In other words -- I think the main case to fix is 1 channel images (alpha alone doesn't make sense for sure). I don't think I've ever seen a 2 channel image in the wild - so maybe `R,G` is just as valid as `L,A`.

lgritz:
As somebody pointed out on the oiio-dev side of this conversation, OpenEXR's DWA compressions actually care about the channel names and only apply the lossy compression to certain named channels. Y would work, but I think L would not.

I agree that 2 channel images are extremely rare, if not virtually extinct in the wild, so maybe it's not worth overthinking that case or forcing it to be consistent with the 1-channel case, punt and go with R, G.

So the leading contenders for one-channel are R and Y.

My best argument for R is that it makes the logic for n channels totally consistent, no special cases for 1-channel.

My best argument for Y is that a one-channel file is almost certainly intended to be grayscale intensity, not "red" per se, and "Y" is already commonly a channel name for grayscale intensity (HSY, YIQ, YCrCb, etc.).

fpsunflower:
`Y` and `R,G` works for me. I like it better than calling the 1 channel case `R` since expanding to `R,0,0` is rarely what you want. You almost always want `Y,Y,Y`.

lgritz:
OK, I'll revise this PR. After sleeping on it, I also like Y better than R for single channel. I'll keep R,G for the rare 2-channel case, because I just don't think there's a better choice.

This PR is just about choosing the default channel names for formats that lack channel names in the file. There are some other channel name-related issues that I'll tackle in separate PRs.

lgritz:
Updated patch to use name "Y" for the special case of a single channel in the file.

rmv:
Second version LGTM

fpsunflower:
LGTM2

PR #1436, opened 2016-06-17
gcc6 missleading indentation

michelmno:
gcc6 sets by default -Werror=misleading-indentation
That detects bad indentation in
src/dpx.imageio/libdpx/Writer.cpp


lgritz:
Thanks, merging.

michelmno:
may be my patch was incorrect, and may be we should use signed char as suggested in https://github.com/KhronosGroup/OpenCOLLADA/issues/439

PR #1443, opened 2016-06-28
Build / Compilation fixes. Windows, OSL, OpenEXR

micler:
CMake modifications: 
OpenEXR dll flag is no longer linked to the LINKSTATIC flag. I've never been able to link against OpenEXR statically (many symbol errors). If it's really needed we could add a dedicated flag in CMake, but using the LINKSTATIC flag is definitely not practical.

Some includes were missing in windows, so we had to modify several headers.
Added a fix for boost string initialization.

thread.h file has been split into two files: thread.h and atomic.h, to remove a link issue when linking against OpenSL: including thread.h pulled some headers that we definitely did not want to include.


lgritz:
This looks fine to me. I'll make sure it compiles and then merge if so.

Will you need this backported to the release branch, or is master ok?

micler:
Master is ok, i'm not building from a release branch.
I'm wondering if it would be interesting to backpatch part of the fixes: the OPENEXR flag and windows fixes. These should be easy to port, unlike the split of the header.

lgritz:
I've merged it into master (split into two commits -- one for the atomic business, a second for all the little compiler warnings and windows stuff).

Tomorrow I'll see if either or both can be easily backported to the release branch, and will do so if it's not difficult.

lgritz:
I merged the build stuff into RB-1.6, not yet the atomic split. I'm not sure if I want to backport that, need to think about it.

PR #1454, opened 2016-07-22
Fix build errors on Windows when making Unicode builds

roelandschoukens:
## Description

Fixes build errors on windows when using the _Unicode Character Set_ (which any serious project should be using these days). Practically this means that `TCHAR` is defined as a `WCHAR`, which is a 16-bit integral type. A `TCHAR[]` is encoded as UTF-16. There are a few places in the source code which assume a `TCHAR` is equal to a char (or alternatively, a `TCHAR*` can be converted to a `std::string`, this assumption is false.

Since the rest of the OpenImageIO project assumes strings are encoded using UTF-8 (?), I have removed the dependency on the `<tchar.h>` and any calls to win32 API functions explicitly use the wide versions (with the `W` suffix). The input / result is converted immediately before / after the API call (an approach inspired by http://utf8everywhere.org/)

Also `simd.h` depends on `missing_math.h` for the `roundf` function.

**Note:** on MSVC 2013 and older, `snprintf` is named `_snprintf`, I'm not sure what your preference is to deal with that.
## Tests

I think these changes are covered by existing tests, if any.

### DIFF
--- a/src/include/OpenImageIO/SHA1.h
+++ b/src/include/OpenImageIO/SHA1.h
@@ -121,27 +121,10 @@
 #define SHA1_WIPE_VARIABLES
 #endif
 
-#if defined(SHA1_HAS_TCHAR)
-#include <tchar.h>
-#else
 #ifdef _MSC_VER
-#include <tchar.h>
+#define _sntprintf _snprintf
 #else
-#ifndef TCHAR
-#define TCHAR char
-#endif
-#ifndef _T
-#define _T(__x) (__x)
-#define _tmain main
-#define _tprintf printf
-#define _getts gets
-#define _tcslen strlen
-#define _tfopen fopen
-#define _tcscpy strcpy
-#define _tcscat strcat
 #define _sntprintf snprintf
-#endif
-#endif
 #endif
 
 // Fallback, if no 64-bit support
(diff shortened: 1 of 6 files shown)


lgritz:
I'm not sure why this didn't get properly reviewed long ago. Apologies. It actually looks like all positive changes to me. I will rebase myself and test, and if it all seems to work and passes CI, I'll merge it.

lgritz:
Looks good, passes tests, merging.

I'm going to merge it directly from my tree, since I had to make some minor edits when I rebased. I wish that GitHub had more choices for how to close a ticket: "Close ticket, this has been merged separately" would be helpful here, so it doesn't look the same as a rejected or abandoned PR.

Apologies again for letting this sit around for so long.

roelandschoukens:
Hi, no problem.

I noticed `_snprintf` is not a direct replacement for `snprintf` — it will not null-terminate the string if truncation occurs. This is something to watch out for if it it ever used for arbitrary length strings.

lgritz:
Can you think of any reason why we should still use _snprintf at all, instead of std::snprintf which by being part of C++11 (our minimum requirement these days) is expected to work on all platforms?

roelandschoukens:
I think Visual Studio 2013 doesn’t have `snprintf`.

lgritz:
Ah, but now I'm confused about our minimum requirements. The [INSTALL.md](https://github.com/OpenImageIO/oiio/blob/master/INSTALL.md) file says we support VS 2013, but [platform.h](https://github.com/OpenImageIO/oiio/blob/f930dba5b7f8be3dabd9226c4bc8b5ba77531d36/src/include/OpenImageIO/platform.h#L102) sure looks like we would not currently build with anything less than VS2015. 

Also [this MSDN page](https://msdn.microsoft.com/en-us/library/hh567368.aspx) seems to indicate that VS2013 doesn't understand constexpr, among other C++ features we use liberally through the code nowadays.

So I think maybe INSTALL.md is wrong, and we're actually VS2015 minimum and have been for a while.

What do you think?

roelandschoukens:
I didn't look in detail, just checked INSTALL.md for system requirements. It looks like VS2015 is required so you can leave it out and just use `snprintf`.

PR #1457, opened 2016-07-23
Clamp ImageCache max_open_files below the system limit

lgritz:
There is trouble if the user tries to set ImageCache's "max_open_files" below the limit of open file handles. Clamp it somewhat below that number.


fpsunflower:
Would it be possible to emit a warning at the same time?

lgritz:
Unfortunately, ImageCache doesn't have a warning mechanism -- the calls either succeed or they don't (if they don't, one can extract an error message).

lgritz:
Not an easy way to emit warnings. You ok with this anyway?

fpsunflower:
Yes! LGTM

PR #1461, opened 2016-08-02
Added support for converting ImageBuf to OpenCV IplImage

scott-wilson:
Hey,

This PR is to add support for converting from ImageBuf to IplImage. I noticed there was the code to go one way, but not the other. If there's any bad code smell, let me know. I'm primarily a Python programmer, so there's a few things that I'm use to being able to do in Python that I can't seem to do in C++. It could be a limitation of the programmer (most likely), or the language.

If this successfully goes through, then I'll work on implementing OpenCV's Mat next.

I have tested this on a simple image that I grabbed from the net, so it works, but may break under more testing. I want to clean up the code more before I try throwing more image types at this.

One thing that isn't supported right now is the Half type because I'm not sure how is the best way to handle that. Suggestions welcome!


src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -71,16 +71,28 @@ endif ()
...
+
+if (${OPENEXR_HOME})
+  set (GENERIC_LIBRARY_PATHS ${OPENEXR_HOME})
+endif()
+
+if (${ILMBASE_HOME})
+  set (GENERIC_LIBRARY_PATHS ${ILMBASE_HOME})
+endif()
+
+set (GENERIC_LIBRARY_PATHS
```
lgritz: Won't this lose the GENERIC_LIBRARY_PATHS value that you just set?
Maybe these should be successively appending rather than assigning?

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,161 @@ IplImage *
...
+    // If the image has 4+ channels, then reduce to 4 channels, and use BGRA
+    // order (OpenCV ordering).
+    if (spec.nchannels >=3) {
+        int channelorder[spec.nchannels];
+
+        // Set the channel order
+        for (int i = 0; i < spec.nchannels; ++i) {
+            channelorder[i] = i;
+        }
+
```
lgritz: May I suggest, in general, all this vertical whitespace is not needed?

lgritz:
I think the openexr cmake stuff should be separated into a separate PR, if possible, and for that matter I think there are bugs in it.

The OpenCV part looks good, thanks for that. 

How should we proceed? Do you want me to try cleaning it all up myself and merging? Or did you want to do the separation, fixing, etc., on your end?

src/cmake/modules/FindOpenEXR.cmake
scott-wilson: You're probably right. Will fix.

src/libOpenImageIO/imagebufalgo_opencv.cpp
scott-wilson: I find the whitespace helps me chunk the code into blocks of logic. I can trim this if you prefer, though.

scott-wilson:
I'll break this into two PRs, and submit the changes tonight. Thanks for the review!

src/libOpenImageIO/imagebufalgo_opencv.cpp
lgritz: Yeah, I do that, too, it's just a tradeoff about chunk size. My view is that the occasional blank line to separate quite different tasks within the same function, each of which may be a dozen or more lines, can be very helpful helpful, whereas if the whitespace is coming every 2-3 lines, it's too granular and the space isn't really adding to understanding, and in fact may hinder understanding by spreading out the code so that less is viewable in a single screenful, if you know what I mean.

This is purely a personal stylistic preference. I'm not claiming that your way is invalid or illogical. I don't even feel strongly enough about it to insist that you do it. Just pointing how to nudge it slightly toward the prevailing style of the codebase, if you are so inclined.

lgritz:
Also... when you break it into 2 PRs... would you mind rebasing them on top of the current master, to clean up the history so that each one looks like a single linear commit rather than a tangle of merges?

If not, I can do that by hand when I merge it, it's no big deal. But if it's already a single commit right on top of the current master in the PR, I can integrate it with a single button and that reduces friction on my end.

scott-wilson:
Works for me. I was planning on rebasing as the last step after I get the lgtm.

scott-wilson:
Alright, quick heads up that I've split the PRs into two, fixed the cmake, and will tidy up the PRs tonight, or tomorrow.

scott-wilson:
Alright, PR cleaned up. I've added in half support, since that's supported on the OIIO side. I convert it to float, since OpenCV doesn't support half.

One thing I noticed when testing is that I get segfaults on bit depths higher than 8 int. As far as I understand from gdb, it isn't on my end, but I couldn't tell you for sure where it is.

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,182 @@ IplImage *
 ImageBufAlgo::to_IplImage (const ImageBuf &src)
 {
 #ifdef USE_OPENCV
-    return NULL;
+    ImageSpec spec = src.spec();
+    ImageBuf tmp = src;
```
lgritz: It seems like the "usual" case (3 or 4 channel image) is going to perform this copy (more on that in a moment) unnecessarily, since just below (if spec.channels >= 3) it will call ImageBufAlgo::channels that also will copy the image into tmp.

But now that I'm looking at the code, I'm concerned that ImageBuf=ImageBuf is not well behaved. I've intentionally tried to define `ImageBuf::operator=` to be unimplemented so that this is disallowed. Yet somehow it compiles. It's not supposed to.

OK, so the way I think this should really work is like this:

```
    const ImageBuf *tmp = &src;
    ImageBuf localcopy;
    if (spec.channels >= 3) {
        ...
        ImageBufAlgo::channels(localcopy, src, ...);
        tmp = &localcopy;
        ...
    }

```

then some uses of `tmp` will need to be replaced by `*tmp` below, but you can see that there are never unnecessary copies made.

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,182 @@ IplImage *
 ImageBufAlgo::to_IplImage (const ImageBuf &src)
 {
 #ifdef USE_OPENCV
-    return NULL;
+    ImageSpec spec = src.spec();
```
lgritz: I think this should be

```
    const ImageSpec &spec (spec.spec());
```

There's no reason to make a full copy of the spec, right? I don't think you intend to modify a local copy, you are just making a shorthand to refer to src.spec() in a less verbose way, and that's exactly what a `const&` is good for.

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,182 @@ IplImage *
...
+
+            if (!dst) {
+                DASSERT (0 && "Could not create dst IplImage.");
+                return NULL;
+            }
+
+            std::vector<unsigned char> data;
+            data.resize(spec.width * spec.height * spec.depth * spec.nchannels);
+
+            if (!tmp.get_pixels(ROI::All(), TypeDesc::UINT8, &data[0])) {
```
lgritz: I think that in many cases, there is an extra copy happening here as well. If spec.format.basetype == UINT8, then the ImageBuf's data is already unsigned char, so why are we calling get_pixels to copy it? That's only necessary if the data are non-local somehow, e.g. the image is backed by ImageCache, which can only happen for <=2 channel images since for >=3 channels, we made a copy (which will always be 'local'). So it seems to me that we can almost always eliminate this copy by directly testing the important case:

```
    if (spec.format.basetype == TypeDesc::UINT8 && tmp.storage == ImageBuf::IMAGECACHE) {
        // Type is right and the IB is not backed by an ImageCache, so
        // directly read from the local data of the IB.
        cvSetData (dst, tmp.localpixels(), dst->widthStep);
    else {
        // Either we are backed by an ImageCache, or we somehow still need a data
        // format conversion, so make a local copy with get_pixels.
        std::vector<unsigned char> data;
        data.resize (spec.width * spec.height * spec.depth * spec.nchannels);
        tmp.get_pixels (ROI::All(), TypeDesc::INT8, &data[0]);
        cvSetData(dst, &data[0], dst->widthStep);
    }
```

Voila, now we have no unnecessary pixel copy!

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,182 @@ IplImage *
...
+                DASSERT (0 && "Could not get pixels from ImageBuf.");
+                return NULL;
+            }
+
+            cvSetData(dst, &data[0], dst->widthStep);
+
+            break;
+        }
+        // OpenCV doesn't have a half type, so promote to float.
+        case TypeDesc::HALF: {
```
lgritz: I think you can just more the `case TypeDesc::HALF:` to share the code block with FLOAT, no need to duplicate it like this, right?

src/libOpenImageIO/imagebufalgo_opencv.cpp
```diff
@@ -136,7 +137,182 @@ IplImage *
...
+            if (!tmp.get_pixels(ROI::All(), TypeDesc::DOUBLE, &data[0])) {
+                DASSERT (0 && "Could not get pixels from ImageBuf.");
+                return NULL;
+            }
+
+            cvSetData(dst, &data[0], dst->widthStep);
+
+            break;
+        }
+        default:
```
lgritz: Maybe for the `default` case, rather than making an error, you should just have `default:` also direct to the UINT8 case (or the FLOAT case?), meaning that if it's not a supported type, just convert to float and keep chugging?

lgritz:
A whole bunch of comments, sorry. I think I starting figuring that anybody doing interoperability between ImageBuf and OpenCV is probably concerned with performance, so making this conversion have as few unnecessary data copies as possible is going to be important.

lgritz:
Also, Scott, it would be great to get a CLA from you. I won't hold up these PRs for it, but when you get a chance, it would be good for all concerned to have in place as you contribute more code.

scott-wilson:
Thanks for the review!

I'll take a look at the fixes tonight (if I have any free time after work), and update the PR. Also, what's the process for getting the CLA to you?

lgritz:
CLA -- you can find it in the doc directory. Print, sign, scan/photo, email back to me privately. There is a CLA for personal use, and another "corporate" one that you should have your employers sign if your contributions are part of your job, or if there is any chance that they might think that they own the results of your coding even in off-hours (which is quite common, if you don't have a separate agreement with them stating that things you work on with your own time are yours alone).

lgritz:
There was nothing "wrong" with your code per se... but I think we'll be happy that we made this code as efficient as possible from the start.

lgritz:
@scott-wilson , did you ever push an update for this with the fixes?

scott-wilson:
Sorry, not yet. Had a lot of overtime at work, and now getting ready to
move. I can probably look into this again around mid October.

lgritz:
No prob, take your time, I just wanted to make sure I wasn't accidentally dropping it on the floor.

scott-wilson:
Alright, back in the development seat (yay!)

I tried the following chunk of code:
```cpp
if (spec.format.basetype == TypeDesc::UINT8 && tmp->storage() == ImageBuf::IMAGECACHE) {
    // Type is right and the IB is not backed by an ImageCache, so
    // directly read from the local data of the IB.
    cvSetData(dst, tmp->localpixels(), dst->widthStep);
    std::vector<unsigned char> data;
} else {
    // Either we are backed by an ImageCache, or we somehow still need a data
    // format conversion, so make a local copy with get_pixels.
    std::vector<unsigned char> data;
    data.resize (spec.width * spec.height * spec.depth * channels);
    tmp->get_pixels (ROI::All(), TypeDesc::INT8, &data[0]);
    cvSetData(dst, &data[0], dst->widthStep);
}
```

And I get the following compile error:
```
/home/scott/Repositories/oiio/src/libOpenImageIO/imagebufalgo_opencv.cpp: In function ‘IplImage* OpenImageIO::v1_7::ImageBufAlgo::to_IplImage(const OpenImageIO::v1_7::ImageBuf&)’:
/home/scott/Repositories/oiio/src/libOpenImageIO/imagebufalgo_opencv.cpp:185:49: error: invalid conversion from ‘const void*’ to ‘void*’ [-fpermissive]
                 cvSetData(dst, tmp->localpixels(), dst->widthStep);
                                                 ^
In file included from /home/scott/Repositories/oiio/src/libOpenImageIO/imagebufalgo_opencv.cpp:37:0:
/usr/include/opencv2/core/core_c.h:403:14: error:   initializing argument 2 of ‘void cvSetData(CvArr*, void*, int)’ [-fpermissive]
 CVAPI(void)  cvSetData( CvArr* arr, void* data, int step );
              ^
make[2]: *** [src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_opencv.cpp.o] Error 1
make[1]: *** [src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/all] Error 2
make: *** [all] Error 2
```

I looked at the imagebuf.h/cpp files and it looks like it should support casting to a void or const void. I'm probably missing something, but I'm currently stuck at this point.

lgritz:
if (spec.format.basetype == TypeDesc::UINT8 && tmp->storage() == ImageBuf::IMAGECACHE) {
       // Type is right and the IB is not backed by an ImageCache, so

You sure you don't maybe have that backwards? The comment says "not backed by ImageCache", but the test is saying "if it is backed by an ImageCache".

How is `tmp` declared?

It looks like cvSetData is being stupid, wanting a `void*` when it clearly is just copying data out of that buffer (so should be `const void*`). You could just try a cast:

    cvSetData(dst, (void *)tmp->localpixels(), dst->widthStep);

scott-wilson:
So, running into a strange issue with the output image. I have a feeling it may be due to a padding issue, or something along those lines. This appears in my earlier draft of the code, so I think it is skewing my results a little bit.

Code
----
```cpp
IplImage *
ImageBufAlgo::to_IplImage (const ImageBuf &src)
{
#ifdef USE_OPENCV
    const ImageSpec &spec = src.spec();
    const ImageBuf *tmp = &src;
    ImageBuf localcopy;
    int channels = std::min(spec.nchannels, 4);

    // If the image has 4+ channels, then reduce to 4 channels, and use BGRA
    // order (OpenCV ordering).
    if (channels >=3) {
        // OpenCV images support up to 4 channels (BGRA)
        int channelorder[channels];

        // Set the channel order
        for (int i = 0; i < channels; ++i) {
            channelorder[i] = i;
        }

        channelorder[0] = 2;  // B
        channelorder[1] = 1;  // G
        channelorder[2] = 0;  // R

        if (!ImageBufAlgo::channels(localcopy, src, channels, &channelorder[0])) {
            DASSERT (0 && "Could not swap channels.");
            return NULL;
        }
        tmp = &localcopy;
    }

    // Create an IplImage to write to.
    CvSize size = cvSize(spec.width, spec.height);
    IplImage *dst;

    // Get the pixel data from the ImageBuf and send it to the IplImage.
    switch (spec.format.basetype) {
        case TypeDesc::UINT8: {
            dst = cvCreateImageHeader(size, IPL_DEPTH_8U, channels);

            if (!dst) {
                DASSERT (0 && "Could not create dst IplImage.");
                return NULL;
            }

            if (tmp->storage() != ImageBuf::IMAGECACHE) {
                // Type is right and the IB is not backed by an ImageCache, so
                // directly read from the local data of the IB.
                cvSetData(dst, (void *)tmp->localpixels(), dst->widthStep);
            } else {
                // Either we are backed by an ImageCache, or we somehow still
                // need a data format conversion, so make a local copy with
                // get_pixels.
                std::vector<unsigned char> data;
                data.resize (spec.width * spec.height * spec.depth * channels);
                tmp->get_pixels (ROI::All(), TypeDesc::INT8, &data[0]);
                cvSetData(dst, &data[0], dst->widthStep);
            }

            break;
        default:
            DASSERT (0 && "unknown TypeDesc");
            return NULL;
    }

    return dst;
#else
    return NULL;
#endif
}
```

Test code
---------
```cpp
#include <ostream>

#include <opencv2/core/core.hpp>
#include <opencv2/highgui/highgui.hpp>
#include <opencv2/core/core_c.h>
#include <opencv2/highgui/highgui_c.h>
#include <OpenImageIO/imagebufalgo.h>

namespace oiio = OpenImageIO;

int main(int argc, char const *argv[]) {
  oiio::ImageBuf src("/home/scott/Projects/oiio-test/images/input.jpg");
  IplImage *dst = oiio::ImageBufAlgo::to_IplImage(src);
  cvSaveImage("/home/scott/Projects/oiio-test/images/output.jpg", dst);

  return 0;
}
```

Input images
------------
input
input2

Output images
-------------
output
output2

My current theories
-------------------
- There's a difference in how the data is laid out. As far as I understand, OpenCV expects that a 3 channel image is laid out BGRBGRBGR... Is OpenImageIO doing any sort of padding?
- I am passing the wrong step into the cvSetData (step calculation: row * type size * channels). I tried manually setting some values but I either get worse results, or OpenCV raises an exception and segfaults.

Is there something else that I could be doing wrong?

scott-wilson:
Thanks @cdstouter. That does the trick. The only things that I see that needs some figuring out are:

- [ ] Support to convert from half (probably just convert to float).
- [ ] Load the pixels in the function, rather than rely on them being loaded before hand.
- [ ] I'm getting a slight colour difference between the input and output. I need to see what could be causing this.

scott-wilson:
One thing I noticed while doing this was that OpenCV doesn't seem to have any way we can store metadata in the IplImage. So, there's an ICCProfile tag in my input png with the colour curve, but that doesn't get transferred over to the IplImage, causing the difference in colour. I don't know how big of a problem this will be.

scott-wilson:
Hey @lgritz, not sure if you don't have any free time for review right now, or this fell to the side, but wanted to let you know that the code is ready for a round of review when you are able.

lgritz:
Sorry, I hadn't realized this got resolved.

In general, it LGTM and I would be ready to merge it, but when I try it on my end, I get a link error because cvCvtColor is not found. Maybe that's an OpenCV 2.x thing but I have 3.x installed? I think it needs to be updated to something that would work in opencv2 and 3, or appropriate #if's to switch based on versions.

Also, I'm no OpenCV expert, but is it also the case that we should be switching from IplImage to cv::Mat?

lgritz:
I think you want cv::cvtColor instead of cvCvtColor.

scott-wilson:
Alright, I'll install 3 and see if I can get the same problem you are. Also, I was planning on adding support for converting to IplImage (since we have converting from right now), then move onto converting to/from cv::Mat.

lgritz:
You don't need to install 3.  If you check on your end (with 2) that cv::cvtColor works, and I'll check on my end (opencv3), if it works both places, let's declare victory and merge!

And it's fine to do the cv::Mat in a second PR. Let's try to quickly get in what  you already have done.

lgritz:
Wait, it's worse than that. We

    #include <opencv2/...>

It seems that we're really only set up for OpenCV2 in the first place.
Why does it work on my end at all?

lgritz:
I think I have a fix, hang on.

lgritz:
OK, forget all of that. The source should be fine. On my end, the problem seems to be the libs.

Would you be so kind as to try this patch on top of yours, and see if it works ok on your system?

    --- a/src/cmake/modules/FindOpenCV.cmake
    +++ b/src/cmake/modules/FindOpenCV.cmake
    @@ -51,7 +51,7 @@ set (libdirs "${PROJECT_SOURCE_DIR}/lib"
                  )
    
    
    -set (opencv_components opencv_highgui opencv_core)
    +set (opencv_components opencv_highgui opencv_imgproc opencv_core)
     if (NOT ${OpenCV_VERSION} VERSION_LESS 3.0.0)
         set (opencv_components opencv_videoio ${opencv_components})
     endif ()

lgritz:
OK, I have a full fix, verified by removing opencv3 and installing opencv2 on my end. Tonight I will merge your patch, and then my link fix right on top of it.

lgritz:
Merged into master.

lgritz:
Also backported to RB-1.7

PR #1462, opened 2016-08-04
Texture bug fix: flip_t didn't flip the result t derivative.

lgritz:
Fixes #1428

Added a FIXME about an unrelated issue, I will investigate soon but did not want to garble this quick fix review.


fpsunflower:
LGTM. I agree this is necessary - can you think of a test case (maybe in OSL?) to show that it works?

lgritz:
Yes, it will be easy to add a test.

lgritz:
Changed the existing testsuite/texture-flipt to also check the derivatives.

PR #1464, opened 2016-08-07
Updated FindOpenEXR.cmake

scott-wilson:
There were a couple of issues with the FindOpenEXR.cmake that caused me issues while compiling.

1. The /usr/local paths were behind the /usr paths. I think it makes sense that if you compiled a new version of a library, the dependent code should point to that one.
2. Not having OPENEXR_HOME or ILMBASE_HOME set would cause the GENERIC_LIBRARY_PATHS to point to /lib (/lib64 in my case) before the paths that I expect them to be.

This should also fix the bug that I introduced in my previous pull request with the OpenCV support.


src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -71,16 +71,28 @@ endif ()
...
+if (${OPENEXR_HOME})
+  set (GENERIC_LIBRARY_PATHS ${GENERIC_LIBRARY_PATHS} ${OPENEXR_HOME})
+endif()
+
+if (${ILMBASE_HOME})
+  set (GENERIC_LIBRARY_PATHS ${GENERIC_LIBRARY_PATHS} ${ILMBASE_HOME})
+endif()
+
+set (GENERIC_LIBRARY_PATHS
     ${OPENEXR_INCLUDE_PATH}/../lib
```
lgritz: Doesn't ${GENERIC_LIBRARY_PATHS} need to be mentioned here, or else you're clobbering the ones you just set?

scott-wilson: Can't believe I missed that. Will fix.

lgritz:
LGTM now.

I'm away for a couple days, but as soon as I get back and am able to test this patch on my work machine to make sure nothing breaks there, I'll commit it.

src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -71,16 +71,29 @@ endif ()
...
 # List of likely places to find the libraries -- note priority override of
 # OPENEXR_CUSTOM_LIB_DIR and ${OPENEXR_HOME}/lib.
-set (GENERIC_LIBRARY_PATHS 
-    ${OPENEXR_CUSTOM_LIB_DIR}
-    ${OPENEXR_HOME}/lib
-    ${ILMBASE_HOME}/lib
+
+# If there's no OPENEXR_HOME or ILMBASE_HOME, then the path will point to
+# "/lib", which may not always be wanted/expected.
+if (${OPENEXR_CUSTOM_LIB_DIR})
```
lgritz: Good thing I tested at work, where we do set OPENEXR_CUSTOM_LIB_DIR!

This is the wrong syntax. It should be

    if (OPENEXR_CUSTOM_LIB_DIR)

And similar below.

I'll fix it on my end and merge.

lgritz:
Merged into master.

Issue #1463, opened 2016-08-05
Cineon gets wrong colorspace metadata


devernay:
### Problem

Cineon files, either standard ones like the Kodak DLAD, or Nuke-generated ones present in the Nuke assets, get wrong metadata.

Note that I am not a Cineon specialist, but there is definitely something wrong there.
- the Kodal DLAD is said to be gamma-corrected with a gamma of 0 (see my comment in #154).
- on Nuke-generated files, the Gamma is `4.6006e-41` (which, by the way is `1.0` read with the wrong endianness), and values for cineon:HighData cineon:HighQuantity cineon:WhitePoint cineon:RedPrimary cineon:GreenPrimary cineon:BluePrimary are totally off. Note that it may be a Nuke bug.
- I do not know what "gamma" means in Cineon terms, but I thought Cineon files were always log-encoded (in which case OIIO should never return "GammaEncoded" as colorspace). Shouldn't that gamma conversion (if any) be applied _after_ log-decoding?

See the output of iinfo and ImageMagick's identify in the gist below for more details.
### Steps to Reproduce
1. download Kodak DLAD files from http://motion.kodak.com/motion/support/technical_information/lab_tools_and_techniques/digital_lad/default.htm
2. download Nuke assets from http://thefoundry.s3.amazonaws.com/downloads/Nuke_BasicWorkflows_Media.tgz
3. command:

```
find . -name '*.cin' | while read f; do echo "\$ iinfo -v $f"; iinfo -v "$f"; echo "\$ identify -verbose $f"; identify -verbose "$f"; echo ; done > cineon.log
```

output: https://gist.github.com/devernay/32bfe5fdf2501344e4075014a68260be
### Versions
- OIIO branch/version: 1.6.15
- OS: OS X 10.11.6
- C++ compiler: Apple LLVM version 7.3.0 (clang-703.0.31)

devernay:
it seems like the header data is bigendian, and thus all 32-bit values have to be swabbed on x86, but the libcineon source code seems to only swabs the image data, not the header data (thus the crazy values read from nuke-encoded cineons). Did I miss something?

GraphicsMagick swabs the header members too:
http://hg.code.sf.net/p/graphicsmagick/code/file/7fa54cb64ff8/coders/cineon.c#l512

ImageMagick also swabs header members (it sets image->endian, and then uses ReadBlob*() which swabs):
https://github.com/ImageMagick/ImageMagick/blob/master/coders/cin.c#L450

[...]

Issue #154, opened 2011-08-30
Cineon cleanup


jeremyselan:
The cineon and dpx codepaths share many common chunks of code. However, the dpx has had many bugfixes that were not ported to cineon.

For example, if we compare cineon.imageio/libcineon/BaseTypeConverter.h dpx.imageio/libdpx/BaseTypeConverter.h  we see the dpx has code to preserve white point scaling.  However, cineon does not have that code (and it should!).

Additionally, cineon has a compile warning that should be addressed:

`In file included from /net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/Writer.cpp:41:                                                                   
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/WriterInternal.h: In function ‘void cineon::WritePackedMethod(IB*, IB*, int, bool, cineon::BufferAccess&) [with IB = unsigned int, int BITDEPTH = 32]’:                                                                                                                      
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/WriterInternal.h:273:   instantiated from ‘int cineon::WriteBuffer(cineon::OutStream*, cineon::DataSize, void*, cineon::U32, cineon::U32, int, cineon::Packing, bool, int, char*, bool&) [with IB = unsigned int, int BITDEPTH = 32, bool SAMEBUFTYPE = true]’               
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/Writer.cpp:286:   instantiated from here                                                               
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/WriterInternal.h:145: warning: right shift count is negative                                           
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/WriterInternal.h: In function ‘void cineon::WritePackedMethod(IB*, IB*, int, bool, cineon::BufferAccess&) [with IB = long unsigned int, int BITDEPTH = 64]’:                                                                                                                 
/net/homedirs/jeremys/git/oiio.js/src/cineon.imageio/libcineon/WriterInternal.h:273:   instantiated from ‘int cineon::WriteBuffer(cineon::OutStream*, cineon::DataSize, void*, cineon::U32, cineon::U32, int, cineon::Packing, bool, int, char*, bool&) [with IB = long unsigned int, int BITDEPTH = 64, bool SAMEBUFTYPE = true]’          
[...]

PR #1466, opened 2016-08-09
More robust test for gamma value in cineon files.

lgritz:
Pointed out in #154 by Frédéric Devernay.

### DIFF
--- a/src/cineon.imageio/cineoninput.cpp
+++ b/src/cineon.imageio/cineoninput.cpp
@@ -208,6 +208,12 @@ CineonInput::open (const std::string &name, ImageSpec &newspec)
     }
     m_spec.attribute ("Orientation", orientation);
 
+#if 1
+    // This is not very smart, but it seems that as a practical matter,
+    // all Cineon files are log. So ignore the gamma field and just set
+    // the color space to KodakLog.
+    m_spec.attribute ("oiio:ColorSpace", "KodakLog");
+#else
     // image linearity
     // FIXME: making this more robust would require the per-channel transfer
     // function functionality which isn't yet in OIIO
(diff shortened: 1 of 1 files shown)


lgritz:
@devernay

devernay:
ok, except I still don't understand if a "reasonable" gamma value (not like oiio extracts from Nuke-produced cineons, see #1463) means that the cineon response curve is a gamma curve, or if it's the Cineon (log) curve followed by gamma...

lgritz:
Updated this PR with a new strategy: just label the color space "KodakLog" and forget about the gamma. This seems to be fine and corresponds to the real-world state of Cineon files, as far as we know. We can always revisit if people are encountering Cineon files that are not log encoded.

lgritz:
Fixes #1463

lgritz:
Does this seem like a reasonable approach for now?

devernay:
yes, great!
I would also change CINEON_SET_ATTRIB_COORDS to check for 0.,0 so that the Kodak Digital LAD reads correctly.
neither the white point nor chromaticities can have (0,0) coordinates anyway
```
#define CINEON_SET_ATTRIB_COORDS(x)     m_cin.header.x (floats);            \
                                        if (!isinf (floats[0])              \
                                            && !isinf (floats[1]) && !(floats[0] == 0. && floats[1] == 0.))          \
                                            m_spec.attribute ("cineon:" #x, \
                                                TypeDesc (TypeDesc::FLOAT,  \
                                                2), &floats[0])
```

lgritz:
I made this change, too. As soon as the CI tests are done, I'll merge this if it's ok.

Is master good for you on this, or do you want it backported to the release branch right away?

devernay:
master is OK, as I'm hacking it on my side anyway: https://github.com/MrKepzie/openfx-io/blob/master/OIIO/ReadOIIO.cpp#L1715

PR #1467, opened 2016-08-09
CMake: fixed how to find ocio header with OCIO_INCLUDE_PATH

cchampet:
## Description
Hello there! This change fixes an issue when we try to find the ocio header with __OCIO_INCLUDE_PATH__ defined (and not __OCIO_PATH__).

Example (with boost, python-2.7 and cmake installed on your system):
```bash
export OCIO_INCLUDE_PATH=/path/to/ocio/include
export OCIO_LIBRARY_PATH=/path/to/ocio/lib
cmake . -DVERBOSE=1
# should see:
# -- Found OCIO library...
# -- Found OCIO includes...
```
## Tests
Tested with CMake 3.4.3.

### DIFF
--- a/src/cmake/modules/FindOpenColorIO.cmake
+++ b/src/cmake/modules/FindOpenColorIO.cmake
@@ -24,13 +24,16 @@ include (FindPackageMessage)
             message(STATUS "OCIO LIBRARY_PATH explicitly specified: ${OCIO_LIBRARY_PATH}")
         endif()
     endif ()
-    FIND_PATH( OCIO_INCLUDES OpenColorIO/OpenColorIO.h
+    FIND_PATH(OCIO_INCLUDES
+        OpenColorIO.h
+        PATHS
         ${OCIO_INCLUDE_PATH}
         ${OCIO_PATH}/include/
         /usr/include
         /usr/local/include
         /sw/include
         /opt/local/include
+        PATH_SUFFIXES OpenColorIO
         DOC "The directory where OpenColorIO/OpenColorIO.h resides")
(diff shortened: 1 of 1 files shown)


src/cmake/modules/FindOpenColorIO.cmake
```diff
@@ -24,13 +24,15 @@ include (FindPackageMessage)
             message(STATUS "OCIO LIBRARY_PATH explicitly specified: ${OCIO_LIBRARY_PATH}")
         endif()
     endif ()
-    FIND_PATH( OCIO_INCLUDES OpenColorIO/OpenColorIO.h
+    FIND_PATH(OCIO_INCLUDES
+        OpenColorIO.h
         ${OCIO_INCLUDE_PATH}
```
lgritz: I'm not sure this is correct. The shorthand is

    FIND_PATH (OCIO_INCLUDES name path path path)

But if you use the longhand version, such as with PATH_SUFFIXES, you seem to also need the token PATHS here.

At least, it fails for me if I don't and works for me if I do, and that appears to be what is described by the docs here: https://cmake.org/cmake/help/v3.0/command/find_path.html

cchampet: Indeed the doc seems to be clear on that. I've overwrote the commit and added the token __PATHS__. It produces the correct result on my configuration (cmake 3.4.3).

lgritz:
LGTM, thanks for the fix.

Do you need me to also merge into a release branch, or is master sufficient?

cchampet:
It's sufficient to put this code to the master branch for me. Thanks!

PR #1468, opened 2016-08-10
Fix roundf: identifier not found

NixonK:
## Description

Got this error while compiling with "Visual Studio 11 2012 Win64":
```
~\src\include\OpenImageIO/simd.h(2318): error C3861: 'roundf': identifier not found [~\build\src\libutil\simd_test.vcxproj]
```

This fixes the build for me but I'm not sure if this is the correct way to fix it.

------

Fixes `~\src\include\OpenImageIO/simd.h(2318): error C3861: 'roundf': identifier not found [~\build\src\libutil\simd_test.vcxproj]`


lgritz:
LGTM, thanks.

PR #1469, opened 2016-08-11
Fix Filesystem::searchpath_find on Windows

edgarv:
Both `Strutil::utf8_to_utf16` and `Strutil::utf16_to_utf8` had a bug in which the converted string would include the trailing `'\0'`, e.g., `"foo"` (len 3) would become `L"foo\0"` (len 4). This caused an infinite recursion on `Filesystem::searchpath_find` during recursive searches.

Also reduce the implicit conversions in `Filesystem::searchpath_find` happening when calling the helper oiio functions. Instead, keep explicit `boost::filesystem::path` objects for as long as possible and call the `boost::filesystem` methods directly.

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -158,42 +158,56 @@ Filesystem::searchpath_split (const std::string &searchpath,
 
 
 std::string
-Filesystem::searchpath_find (const std::string &filename,
+Filesystem::searchpath_find (const std::string &filename_utf8,
                              const std::vector<std::string> &dirs,
                              bool testcwd, bool recursive)
 {
-    bool abs = Filesystem::path_is_absolute (filename);
+#ifdef _WIN32
+    const boost::filesystem::path filename (Strutil::utf8_to_utf16 (filename_utf8));
+#else
+    const boost::filesystem::path filename (filename_utf8);
+#endif
+    bool abs = filename.is_absolute();
 
     // If it's an absolute filename, or if we want to check "." first,
     // then start by checking filename outright.
     if (testcwd || abs) {
-        if (Filesystem::is_regular (filename))
-            return filename;
+        if (boost::filesystem::is_regular_file (filename))
+            return filename_utf8;
     }
 
     // Relative filename, not yet found -- try each directory in turn
-    BOOST_FOREACH (const std::string &d, dirs) {
+    BOOST_FOREACH (const std::string &d_utf8, dirs) {
         // std::cerr << "\tPath = '" << d << "'\n";
-        boost::filesystem::path f = d;
-        f /= filename;
-        // std::cerr << "\tTesting '" << f << "'\n";
-        if (Filesystem::is_regular (f.string())) {
-            // std::cerr << "Found '" << f << "'\n";
+#ifdef _WIN32
+        const boost::filesystem::path d(Strutil::utf8_to_utf16 (d_utf8));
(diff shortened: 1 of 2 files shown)


src/libutil/filesystem.cpp
```diff
@@ -158,42 +158,56 @@ Filesystem::searchpath_split (const std::string &searchpath,
...
+        if (boost::filesystem::is_regular_file (filename))
+            return filename_utf8;
     }
 
     // Relative filename, not yet found -- try each directory in turn
-    BOOST_FOREACH (const std::string &d, dirs) {
+    BOOST_FOREACH (const std::string &d_utf8, dirs) {
         // std::cerr << "\tPath = '" << d << "'\n";
-        boost::filesystem::path f = d;
-        f /= filename;
```
edgarv: Assigning a string to a `path` was triggering a conversion to wide characters using the boost default conversion facet.

src/libutil/filesystem.cpp
```diff
@@ -158,42 +158,56 @@ Filesystem::searchpath_split (const std::string &searchpath,
...
-        // std::cerr << "\tTesting '" << f << "'\n";
-        if (Filesystem::is_regular (f.string())) {
```
edgarv: `f.string()` triggered a narrowing conversion using boost.

src/libutil/strutil.cpp
```diff
@@ -524,8 +524,8 @@ Strutil::utf8_to_utf16 (string_view str)
 {
     std::wstring native;
     
-    native.resize(MultiByteToWideChar (CP_UTF8, 0, str.c_str(), -1, NULL, 0));
-    MultiByteToWideChar (CP_UTF8, 0, str.c_str(), -1, &native[0], (int)native.size());
+    native.resize(MultiByteToWideChar (CP_UTF8, 0, str.c_str(), str.length(), NULL, 0));
```
lgritz: Actually, I think that if you're passing the length explicitly, we should also pass `str.data()` rather than `str.c_str()` since if str is not already null-terminated, my `.c_str()` method will create a ustring in order to get back a null-terminated string with persistent memory.

src/libutil/strutil.cpp
```diff
@@ -537,8 +537,8 @@ Strutil::utf16_to_utf8 (const std::wstring& str)
 {
     std::string utf8;
 
-    utf8.resize(WideCharToMultiByte (CP_UTF8, 0, str.c_str(), -1, NULL, 0, NULL, NULL));
-    WideCharToMultiByte (CP_UTF8, 0, str.c_str(), -1, &utf8[0], (int)utf8.size(), NULL, NULL);
+    utf8.resize(WideCharToMultiByte (CP_UTF8, 0, str.c_str(), str.length(), NULL, 0, NULL, NULL));
```
lgritz: same thing here -- `str.data()` if we're passing the length explicitly.

lgritz:
Good to see you, Edgar, it's been a while.

Good catch on this.

I made a note about a slight efficiency you can add, but pending that, LGTM. I'll wait for you to patch that, then I'll merge and backport.

edgarv:
I should have said hi during SIGGRAPH :)

Thank you for noticing the string_view issue. I amended the commit as per your suggestion. Hopefully I will submit a few other pending fixes soon.

lgritz:
LGTM! I will merge now.

PR #1471, opened 2016-08-17
deal with vs 2015 being pedantic about alignment issues.

bdeluca:
It still breaks compiling  tryConvert(...) as it wasnt to call tryConvert(T) rathern than tryConvert(const T&)
```
s:\storage\git\oiio.git\src\include\openimageio\tinyformat.h(178): error C2718: 'const OpenImageIO::v1_7::simd::int4': actual parameter with requested alignment of 16 won't be aligned
s:\storage\git\oiio.git\src\include\openimageio\tinyformat.h(600): note: see reference to class template instantiation 'tinyformat::detail::is_convertible<const T &,int>' being compiled
        with
        [
            T=OpenImageIO::v1_7::simd::int4
        ]
s:\storage\git\oiio.git\src\include\openimageio\tinyformat.h(883): note: see reference to function template instantiation 'void tinyformat::detail::FormatIterator::accept<T1>(const T &)' being compiled
        with
        [
            T1=OpenImageIO::v1_7::simd::int4,
            T=OpenImageIO::v1_7::simd::int4
        ]
s:\storage\git\oiio.git\src\include\openimageio\tinyformat.h(944): note: see reference to function template instantiation 'void tinyformat::detail::format<T1,T2>(tinyformat::detail::FormatIterator &,const T1 &,const T2 &)' being compiled
        with
        [
            T1=OpenImageIO::v1_7::simd::int4,
            T2=OpenImageIO::v1_7::simd::int4
        ]
S:\storage\git\oiio.git\src\include\OpenImageIO/strutil.h(91): note: see reference to function template instantiation 'void tinyformat::format<T1,T2>(std::ostream &,const char *,const T1 &,const T2 &)' being compiled
        with
        [
            T1=OpenImageIO::v1_7::simd::int4,
            T2=OpenImageIO::v1_7::simd::int4
        ]
S:\storage\git\oiio.git\src\libutil\simd_test.cpp(420): note: see reference to function template instantiation 'std::string OpenImageIO::v1_7::Strutil::format<OpenImageIO::v1_7::simd::int4,OpenImageIO::v1_7::simd::int4>(const char *,const T1 &,const T2 &)' being compiled
        with
        [
            T1=OpenImageIO::v1_7::simd::int4,
            T2=OpenImageIO::v1_7::simd::int4
        ]
```

### DIFF
--- a/src/include/OpenImageIO/color.h
+++ b/src/include/OpenImageIO/color.h
@@ -245,7 +245,7 @@ inline float sRGB_to_linear (float x)
                            : powf ((x + 0.055f) * (1.0f / 1.055f), 2.4f);
 }
 
-inline simd::float4 sRGB_to_linear (simd::float4 x)
+inline simd::float4 sRGB_to_linear (const simd::float4& x)
 {
     return simd::select (x <= 0.04045f, x * (1.0f/12.92f),
                          fast_pow_pos (madd (x, (1.0f / 1.055f), 0.055f*(1.0f/1.055f)), 2.4f));
@@ -260,7 +260,7 @@ inline float linear_to_sRGB (float x)
 
 
 /// Utility -- convert linear value to sRGB
-inline simd::float4 linear_to_sRGB (simd::float4 x)
+inline simd::float4 linear_to_sRGB (const simd::float4& x)
 {
     // x = simd::max (x, simd::float4::Zero());
     return simd::select (x <= 0.0031308f, 12.92f * x,
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -216,7 +216,7 @@ OIIO_FORCEINLINE uint64_t rotl64 (uint64_t x, int k) {
 /// clamp a to bounds [low,high].
 template <class T>
 inline T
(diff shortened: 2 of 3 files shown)


bdeluca:
This builds on windows in appveyor here and fails in linux/osx
on my systems 
This fails on windows and builds in linux/osx

lgritz:
I don't quite understand why this should be an error. Are aligned types not allowed to be passed as value (stack) parameters?

bdeluca:
its a little bit over my head, but I think there is no guarantee that the copied value is aligned. 
ms says this https://msdn.microsoft.com/en-us/library/sxe76d9e.aspx

it seems to have been triggered by me trying to use boost 1.60 there is some more about it here https://datastax-oss.atlassian.net/browse/CPP-341

lgritz:
Are you compiling for a 32 bit target?

Why do my appveyor builds succeed?

bdeluca:
I don't think It was a 32 bit target but with nmake It was hard to be 100% sure. 
I think it came down you appveyor running boost 1.59 vs the boost 1.60 trying. 

Next week I can come back to trying it with 1.60 again and confirm it was this. All my cpu is compiling other stuff today.

PR #1481, opened 2016-09-08
ImageCache/TextureSystem: collect stats on total size on disk.

lgritz:
The size statistics reported by IC/TS are the *data* size -- that is,
the size of all pixels of the image if they were read into memory, as
represented in the cache (uncompressed, and in one of the data types
supported by the cache).

But this doesn't accurately reflect the size of the images as they exist
on disk. The three main ways they differ are: (a) the image data doesn't
count the size of the headers, metadata, etc; (b) the pixel data on disk
is often compressed, making it much smaller stored on disk than the
in-cache size; (c) the data on disk may be in a format not directly
supported by the cache, and is "expanded" upon being read (for example,
10 bit per channel data would be expanded to 16 bits per channel in
the cache, palette images might be represented by one byte per pixel
on disk but 3 bytes of RGB per pixel in the cache, and so on).

This patch records the actual file size on disk for each file referenced
by the ImageCache.

Now the stats try to more clearly describe these two separate values:

    Total pixel data size of all images referenced : 120.0 MB
    Total actual file size of all images referenced : 6.6 MB

The stats also can be accessed through the IC or TS APIs, as totals for
all files via getattribute queries for "stat:image_size" and
"stat:file_size", and for individual files using get_image_info queries
of the same name (all are int64 values).

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -148,6 +148,7 @@ ImageCacheStatistics::init ()
 //    tiles_current = 0;
 //    tiles_peak = 0;
     files_totalsize = 0;
+    files_totalsize_ondisk = 0;
     bytes_read = 0;
 //    open_files_created = 0;
 //    open_files_current = 0;
@@ -192,6 +193,7 @@ ImageCacheStatistics::merge (const ImageCacheStatistics &s)
 //    tiles_current += s.tiles_current;
 //    tiles_peak += s.tiles_peak;
     files_totalsize += s.files_totalsize;
+    files_totalsize_ondisk += s.files_totalsize_ondisk;
     bytes_read += s.bytes_read;
 //    open_files_created += s.open_files_created;
 //    open_files_current += s.open_files_current;
@@ -289,6 +291,7 @@ ImageCacheFile::ImageCacheFile (ImageCacheImpl &imagecache,
       m_mipused(false), m_validspec(false), m_errors_issued(0),
       m_imagecache(imagecache), m_duplicate(NULL),
       m_total_imagesize(0),
+      m_total_imagesize_ondisk(0),
       m_inputcreator(creator),
       m_configspec(config ? new ImageSpec(*config) : NULL)
 {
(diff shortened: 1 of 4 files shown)


src/libtexture/imagecache.cpp
```diff
@@ -467,6 +470,7 @@ ImageCacheFile::open (ImageCachePerThreadInfo *thread_info)
     // Since each subimage can potentially have its own mipmap levels,
     // keep track of the highest level discovered
```
fpsunflower: I'm not sure I follow this part.

We only count the size of the largest subimage? Why not count the size of the subimage we actually read?

Does the same logic make sense for the filesize?

lgritz: No, we're looping over subimages and MIP levels and SUMMING to get the total data size.

For file size, the calculation is outside the loop (see line 586) and just asks for Filesystem::file_size(). Just like there is not a reliable way to associate a size on disk with one particular tile, neither is there for one MIP level or one subimage. It's all or nothing for that stat.

lgritz:
Comments?

lgritz:
You ok with this, Chris?

Soren and Dan Kripak, does this look helpful to you?

fpsunflower:
LGTM!

src/libtexture/imagecache.cpp
fpsunflower: So subtracting the old and adding the new is to account for the case where you might be re-reading the same file?

lgritz: Yeah, and in particular, the case where the file itself has changed! (Like, between interactive re-renders, they update the texture.)

dankripac:
Yes looks good to me (by eye). thanks!

Will get a chance to run it in our build soon.

lgritz:
Wait, I just realized, I should not have changed the old name of the attribute to retrieve the stat. I'm going to make sure that the old name aliases to the new one, to avoid breaking software that tries to retrieve the stat with the old name.

I'll make that change before I merge the fix into master.

PR #1482, opened 2016-09-09
Make sure the sign bit of expm1 is correct for small arguments

fpsunflower:
The sign bit of expm1 could be wrong (due to the "denormals crush" technique we use to speed things up).

Since expm1 is a strictly increasing function passing through 0, we can just force in the sign bit we want from the argument.


lgritz:
OK, LGTM.

How did this come up as an issue?

fpsunflower:
Reviewing some volume sampling code to make it more numerically robust -- and the `1-exp(...)` pattern comes up a lot. Obviously hilarity ensues if you get the sign bit wrong in a pdf :)

PR #1483, opened 2016-09-09
Add an optimized (and sign accurate) version of log1p

fpsunflower:
`log1p` is the natural companion to `expm1` and we were missing a fast version of it.

Just like in `expm1`, care is taken to get the sign bit right for small arguments (which simply calling `fast_log(1+x)` gets wrong for roughly 50% of inputs between -1 and 1).

After more careful measurement, I also found I could relax the threshold for small angles in `expm1` without changing the maximum error. This makes it slightly faster.


lgritz:
LGTM. I'll merge as soon as the CI tests finish.

PR #1485, opened 2016-09-09
oiiotool: Add support for writing out mixed channel bitdepths

### LINKED ISSUE #1445: Keeping original bit types with oiiotool
Hi,

I'm not sure if this is intentional but when I use "oiiotool" on a multi-different-bit-layer exr, "oiiotool" seems to change the original bit types.

### Problem

**Expected behavior:** 
I would like to keep the original bit types when merging images if the images have same bit types, and only when specifying "-d" option to change the bit types, I would like it to convert the bit types to the specified bit types.
Basically, what I'm trying to do is to merge multiple tiling images that have same AOVs and bit types.

**Actual behavior:**
"oiiotool" seems to convert all channels to 32 bit even when changing only medatadata.
(Help says "Options that change current image metadata (but not pixel values)" but it changes bit types to "float" from "half", and it may change the pixel values)

### Steps to Reproduce

I'm putting 2 things, one is merging 2 images and one is simply to add metadata only.

## merging images with "--over" option

1) make 2 multi-different-bit-layer exr files like this:
(Basically, I have 2 images that have exact same AOVs and bit types but different dataWindow for tailing renders)
$ exrheader tile_1_1.exr
...
channels (type chlist):
    A, 16-bit floating-point, sampling 1 1
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
    depth.Z, 32-bit floating-point, sampling 1 1
    diffuse.blue, 16-bit floating-point, sampling 1 1
    diffuse.green, 16-bit floating-point, sampling 1 1
    diffuse.red, 16-bit floating-point, sampling 1 1
...
dataWindow (type box2i): (0 0) - (1027 1555)
...

$ exrheader tile_2_1.exr
...
channels (type chlist):
    A, 16-bit floating-point, sampling 1 1
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
    depth.Z, 32-bit floating-point, sampling 1 1
    diffuse.blue, 16-bit floating-point, sampling 1 1
    diffuse.green, 16-bit floating-point, sampling 1 1
    diffuse.red, 16-bit floating-point, sampling 1 1
...
dataWindow (type box2i): (1028, 0) - (2047 1555)
...

2) merge the 2 images with "--over" option
$ oiiotool tile_1_1.exr --over tile_2_1.exr -o output.exr

3) chech the bit types and the resolution
$ exrheader output.exr
...
channels (type chlist):
    A, 32-bit floating-point, sampling 1 1
    B, 32-bit floating-point, sampling 1 1
    G, 32-bit floating-point, sampling 1 1
    R, 32-bit floating-point, sampling 1 1
    depth.Z, 32-bit floating-point, sampling 1 1
    diffuse.blue, 32-bit floating-point, sampling 1 1
    diffuse.green, 32-bit floating-point, sampling 1 1
    diffuse.red, 32-bit floating-point, sampling 1 1
...
dataWindow (type box2i): (0 0) - (2047 1555)
...

## just adding a metadata
1) make an multi-different-bit-layer exr like this:
$ exrheader image.0000.exr
...
channels (type chlist):
    A, 16-bit floating-point, sampling 1 1
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
    depth.Z, 32-bit floating-point, sampling 1 1
    diffuse.blue, 16-bit floating-point, sampling 1 1
    diffuse.green, 16-bit floating-point, sampling 1 1
    diffuse.red, 16-bit floating-point, sampling 1 1
...
2) add a metadata
$ oiiotool image.0000.exr  --sattrib "test" "test value" -o output.exr

3) chech the bit types and the metadata
$ exrheader output.exr
...
channels (type chlist):
    A, 32-bit floating-point, sampling 1 1
    B, 32-bit floating-point, sampling 1 1
    G, 32-bit floating-point, sampling 1 1
    R, 32-bit floating-point, sampling 1 1
    depth.Z, 32-bit floating-point, sampling 1 1
    diffuse.blue, 32-bit floating-point, sampling 1 1
    diffuse.green, 32-bit floating-point, sampling 1 1
    diffuse.red, 32-bit floating-point, sampling 1 1
...
test (type string): "test value"
...

### Versions

* OIIO branch/version: "tags/Release-1.6.14" and "master"
(I've checked v1.6.14 and master but they gave the same results)
* OS: cent 6
* C++ compiler: gcc 4.8.3

Best Regards,
Kazuki Takahashi

AWhetter:
## Description

Currently when oiio reads in an image with channels that have different bitdepths, all channels get converted to the bitdepth of the image. This fixes that by retaining the format of each channel when reading in the image.

oiiotool would also ignore mixed channel bitdepth information, so now it retains the information and writes out the channels at whatever their bitdepth is in the image buffer.

Fixes #1445 

## Tests

I can't provide the test data that I have and I can't generate some test data either. If someone is able to supply some test data then I'd be happy to write the tests.

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -839,6 +839,7 @@ ImageBufImpl::read (int subimage, int miplevel, bool force, TypeDesc convert,
         m_spec.format = convert;
     else
         m_spec.format = m_nativespec.format;
+    m_spec.channelformats = m_nativespec.channelformats;
     m_pixelaspect = m_spec.get_float_attribute ("pixelaspectratio", 1.0f);
     realloc ();
 
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -4282,6 +4282,8 @@ output_file (int argc, const char *argv[])
             // For deep files, must copy the native deep channelformats
             if (spec.deep)
                 spec.channelformats = (*ir)(s,0).nativespec().channelformats;
+            else
+                spec.channelformats = (*ir)(s,0).spec().channelformats;
             // If it's not tiled and MIP-mapped, remove any "textureformat"
             if (! spec.tile_pixels() || ir->miplevels(s) <= 1)
                 spec.erase_attribute ("textureformat");


lgritz:
I'm not sure this patch is correct.

ImageBuf is intended to provide a simplified interface to reading and storing an image, and one of the simplifications it makes is that it internally stores all channels in a single data format. If the file has differing data formats per channel (which, BTW, only a few of the formats support at all), it "rounds up" to the format with the highest precision and range. So a file with 'half' RGBA and 'float' z will end up as a 'float' ImageBuf.

ImageBuf.spec() describes what's in the BUFFER.

ImageBuf.nativespec() describes what was in the FILE.

So I think that given this design choice, the current code is correct and it would be a mistake to copy the per-channel format information from nativespec to spec.

Now, if you want to argue that ImageBuf should be capable of storing data with differing per-channel formats, that's another issue entirely. The main counter-argument is to consider what it would do to the complexity and performance of basically anything that uses ImageBuf (like every ImageBufAlgo function).

AWhetter:
I think it should be able to handle per-channel bitdepths. If a supported image format can then oiio should be able to in my opinion. It would be weird to do `oiiotool input -o output.exr` and have the output be different from the input for example.

lgritz:
> I think it should be able to handle per-channel bitdepths.

Depends on what "it" is. The ImageBuf class is designed not to, and the defense of this design decision is just to imagine the added complexity and performance penalty this would drop on the doorstep of every single ImageBufAlgo function. They are already complicated by having to consider several data types, but at least they can consider it once per image and branch appropriately. If there were decisions for each channel for each pixel, it would be very bad for performance and also much harder to correctly write an image processing operation that worked with ImageBuf (by "correctly" I mean, "not accidentally failing to consider every possible permutation of per-channel types.").

>  If a supported image format can then oiio should be able to in my opinion.

*OIIO* certainly can, just not via the ImageBuf class. You can use a raw ImageInput and ImageOutput to read and write the data in its native per-channel data types. You just pass the various read and write methods TypeDesc::UNKNOWN as the buffer type, and that tells it to use the native in-file types.

> It would be weird to do oiiotool input.exr -o output.exr and have the output be different from the input for example.

Indeed it would! I would consider it a bug.

oiiotool may locally store in-memory as a single data type for all channels (in fact, it will convert it all to float, unless you tell it not to -- this is for speed), but it remembers the format it was on disk and when you write, tries hard to duplicate that format, including any per-channel differences. So my intent is for `oiiotool input.exr -o output.exr` to preserve the original formats by the time the output gets back to disk. But not necessarily while it's in memory.

AWhetter:
> So my intent is for oiiotool input.exr -o output.exr to preserve the original formats by the time the output gets back to disk. But not necessarily while it's in memory.

Ah this makes sense. Do you know how I would access this information in `output_file` of `oiiotool.cpp` currently? I'm not overly familiar with oiio code yet (in case you hadn't guessed :wink:).

lgritz:
I see that this is broken -- the oiiotool input.exr -o out.exr doesn't quite work. I'll prepare a patch.

I'm not sure what you mean by "access this information in output_file". Can you clarify what you're trying to do?

AWhetter:
I meant access the information about the output file in the `output_file` function in `oiiotool.cpp` so that I can fix `oiiotool input.exr -o out.exr` :smiley: I believe it's the root cause of #1445. But you'll be able to fix it quicker than I can, so go ahead :wink:

lgritz:
See #1491

PR #1487, opened 2016-09-13
Fix OpenEXR input and output of chromaticities

lgritz:
"Chromaticites" has a custom attribute type to OpenEXR, so need a bit of special handling to get it right.


lgritz:
Updated for correct capitalization of "chromaticities"

PR #1490, opened 2016-09-15
RAW reader: turn off some auto-adjustments

lgritz:
Per discussion on oiio-dev, the consensus seems to be:

* Turn off auto-bright adjustment
* turn on camera whie balance
* set maximum thresholding to 0

Add "config" overrides for all of these, for anybody who doesn't like
the default.

Also, document these, and the other existing-but-undocumented configuration
settings.


lgritz:
Any comments on this? HP?

lgritz:
Discussion was on oiio-dev, LGTHP.

PR #1493, opened 2016-09-22
OpenEXR: allow dwaaCompressionLevel to be appended to compression name.

lgritz:
For example, oiiotool in.exr -compression dwaa:200 -o out.exr

This is more convenient than separately needing to set the
"openexr:dwaCompressionLevel" metadata.

This came from discussion on a thread on oiio-dev mail list.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -221,6 +221,9 @@ Fixes, minor enhancements, and performance improvements:
       PixelAspectRatio based on the "XResolution" and "YResolution"
       attributes. #1453 (Fixes #1214) (1.7.4/1.6.16)
     * Fix setting "chromaticity" metadata in EXR files. #1487 (1.7.5)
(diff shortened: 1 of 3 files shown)


lgritz:
Anybody give this a try? Did you like?

lgritz:
If there are no objections, I'm going to commit this, because I really want to push out a release candidate today.

lgritz:
OK, merged. We can always return to this if something is found to be wrong.

maxnbk:
I found this feature super useful and worked as-needed, but found this was not clearly expressed in the oiiotool -h docs and thus wound up here to see if this was actually implemented via mail group searching.
It may be advisable to change the help string to something along the lines of:
"--compression %d(:%d)       Set the compression method (and optionally method-specific quality)"

PR #1494, opened 2016-09-22
oiiotool: new -o option: -o:all=n <pattern>

lgritz:
This causes all images on the stack to be output (not just the top), using
the filename pattern (assumed to have an integer substitution '%d'). The
number substituted will be index of the image (starting with the bottom
of the stack, beginning with number n).

Example of how this is useful:

    oiiotool multiimage.tif -sisplit -o:all=1 subimage%04d.tif

Splits multiimage.tif into individual subimages and saves them all
as separate images with names subimage0001.tif, subimage0002.tif, etc.


danwexler:
Testing...

lgritz:
Did it work?

danwexler:
Works! LGTM. Thanks!

PR #1495, opened 2016-09-23
oiiotool: slight refactoring of attribute setting internals

lgritz:
--attrib, --iscolorspace, --keyword, --caption, --orientation, and maybe
some other attribute-setting commands, confusingly applied only to the
first subimage of a multi-subimage file. Now by default they will apply
to all the subimages, which almost certainly is what people probably
expect.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -701,7 +701,8 @@ set_string_attribute (int argc, const char *argv[])
         ot.warning (argv[0], "no current image available to modify");
         return 0;
     }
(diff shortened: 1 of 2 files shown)


lgritz:
I'm abandoning the bulk of this PR, since the person who had requested this change has since discovered that the existing "-a" flag already accomplishes what he needed.

So now there is NO behavior change compared to prior releases.

But I did keep some of the internal refactoring: instead of the set_attribute function using the "global" setting ot.allsubimages, it takes a parameter, and the several places that it's called pass ot.allsubimages as the argument. This merely gives us some extra future flexibility for how set_attribute is used.

PR #1496, opened 2016-09-26
SPI build fixes

lgritz:
Of concern only to SPI peeps, won't affect anybody else.

* Link against static libtiff to solve .so hell.
* Make SPI-specific visibility map file
* Fix Nuke location


fpsunflower:
LGTM

lgritz:
After they install the .a file somewhere on the system for me, I'll change it again to it isn't pointing to the custom one I built in my area. But since it's linked statically, that won't hurt anything for now, once a build is done it no longer refers back to my home.

PR #1499, opened 2016-09-28
RLA fix: bad channel logic made RGBZ file think channel 3 was alpha

lgritz:
Bad channel logic made RGBZ file think channel 3 was alpha as well as Z. Confusing!


lgritz:
I've pushed updates. After fixing this and doing more testing, I identified a few additional minor errors that were related to reading and writing RLA files with mixed channel types. I think it's all cleared up now.

PR #1501, opened 2016-09-29
Fix oiiotool --mosaic MxN when the image stack has < M*N images on it.

lgritz:
Fill in with black spots, rather than fail mysteriously.


fpsunflower:
LGTM

PR #1504, opened 2016-09-29
Overhaul of metadata printing.

lgritz:
Airplane project from last weekend. This is for you, @danwexler 

* ImageSpec::to_xml() improvement: output contents of extra_attrib.
* ImageSpec::serialize() serializes the spec, as text or xml. Leaves
  room for other formats in the future. (JSON? YAML?)
* 'oiiotool -info:format=xml' prints metadata as xml

A bunch of refactoring to support this.

### DIFF
--- a/src/doc/imageioapi.tex
+++ b/src/doc/imageioapi.tex
@@ -662,6 +662,19 @@ values) for certain known metadata.
 \apiend
 
 
+\apiitem{std::string {\ce serialize} (SerialFormat format, SerialVerbose verbose) const}
+\NEW % 1.8
+Returns, as a string, a serialized version of the \ImageSpec.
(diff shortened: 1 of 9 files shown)


lgritz:
I should point out, this doesn't change the formatting of the text output that oiiotool has always done, although it significantly refactors the code that does it. So nobody's output they are depending on should change. But if you want xml instead, this makes it possible.

danwexler:
Compiled cleanly. Looks like oiiotool changes caused some unit test failures in my code. Debugging.

lgritz:
How's this looking now, Dan? And please note I updated the patch since you wrote the "compiled cleanly" comment.

danwexler:
LGTM! Passing all my unit tests. Thanks!

PR #1513, opened 2016-10-02
C++11 as new minimum language standard. No turning back.

lgritz:
C++11 means that our new minimum compiler requirements are:
* clang >= 3.4, and using -std=c++11
* gcc >= 4.8, and using -std=c++11
* MSVS 2015 or newer

No more USE_CPP11 option.

We also allow C++14 compilation (make USE_CPP=14) and test it in Travis. (Note that we changed the name in order to be better positioned for new standards.)

Needless to say, this is master (future 1.8), and will not be backported to RB-1.7. The 1.7 family of releases will continue to be buildable as C++03.


fpsunflower:
LGTM!

PR #1518, opened 2016-10-05
Fix dither bug -- was mixing up buffer copy regions, garbling pixels.

lgritz:


lgritz:
Updated the patch -- I didn't quite solve all the problems the first time. There was an additional buffer size mistake.

Actually, I think the second commit of this patch is the REAL solution, and seems to solve the problem completely even without the earlier commit. But I'm keeping the first anyway, because it still unclutters and clarifies the surrounding code.

PR #1519, opened 2016-10-05
Fix bug in recently modified UDIM code

lgritz:
Forgot to recheck for broken file after changing UDIM name pattern into
a concrete file. A better solution is to fully recurse (just one level)
to perform all the initial checks on the concrete file.


fpsunflower:
LGTM

PR #1523, opened 2016-10-08
A number of fixes for Windows platform

darkhorse64:
## Description

The first three commits are needed to compile OIIO on Windows. I have checked the unavailability of the intrinsics in `<immintrin.h>`
The fourth commit fixes a bug on Windows where OIIO is unable to load its plugins. The reason is that `directory_iterator.path().string()` does not return a suitable path but must be first converted back to utf8. I haven't checked but the `regex `part looks also suspicious in this respect; Using a `wregex` looks more consistent.

## Tests

Tested by loading various images with iv

It would be nice to have those changes backported to 1.7

### DIFF
--- a/src/ffmpeg.imageio/CMakeLists.txt
+++ b/src/ffmpeg.imageio/CMakeLists.txt
@@ -2,8 +2,10 @@
 # marked m_format_context->streams[i]->codec as deprecated.
 # FIXME -- at some point, come back and figure out how to fix for real
 # before the field disappears entirely.
-set_source_files_properties (ffmpeginput.cpp
+if (NOT MSVC)
+    set_source_files_properties (ffmpeginput.cpp
                              PROPERTIES COMPILE_FLAGS "-Wno-deprecated-declarations")
+endif()
 
 if (USE_FFMPEG AND FFMPEG_FOUND)
     add_oiio_plugin (ffmpeginput.cpp ffmpegoutput.cpp
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -2467,7 +2467,7 @@ template<int i> OIIO_FORCEINLINE bool8 shuffle (const bool8& a) {
 
 template<int i>
 OIIO_FORCEINLINE bool extract (const bool8& a) {
-#if OIIO_SIMD_AVX
+#if OIIO_SIMD_AVX && !_WIN32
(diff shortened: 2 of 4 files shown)


lgritz:
LGTM. 

I have some concerns about the extract/insert intrinsics: they are spread over a few places and I fear that we don't have a clear idea of what conditions lead to those intrinsics being missing or know when we can go back to using those intrinsics. But I have an idea for another way to handle this that I think I prefer.

So my plan is to merge your other three fixes immediately (and backport them to 1.7, and then I want to mess a bit with the simd.h changes and see what you think of my proposal, before merging that.

Sound reasonable?

lgritz:
Do you have access to AVX2? I'm wondering if on Windows, these intrinsics are defined at the AVX2 level but not AVX. But I don't have a good way to test that theory.

lgritz:
I'm wondering if the expedient change is just to alter these:

    #if OIIO_SIMD_AVX && !_WIN32

to

    #if OIIO_SIMD_AVX2 || (OIIO_SIMD_AVX && !_WIN32)

this is based on the hypothesis that these are defined for AVX2 even on Windows. Can you give that a try on your end and see if it works? But of course, only if you have AVX2, or you won't know.

lgritz:
I've 3 of the commits into both master and RB-1.7.

I'm holding off on the simd.h one until I hear back from you about your opinion on the AVX2 issue. I'm ok merging it as-is, but would prefer to use the intrinsic for AVX2 if it is available.

darkhorse64:
Tough question being not an expert in that area

I have tried both options: compiling with AVX2 (simd_test crashes which likely means my processor is too old) and compiling with AVX (simd_test is fine). In either cases, these intrinsics are missing. I have also checked VisualStudio 2015 include files (I have an up to date version). Nothing.

I have skimmed through Embree source code but it does not make use of these intrinsics and there is nothing similar to the problematic OIIO functions. 

I have googled for these intrinsics: it's not completely clear but it seems that these intrinsics are only known in the GCC and LLVM worlds.

Therefore, I am inclined to conclude that these intrinsics are now unsupported on VisualStudio 2015 whatever is your target but that it may change in the future.

lgritz:
I think you answered the question adequately -- it doesn't appear that those intrinsics are available with VS 2015 even with AVX2 enabled.

I'll stick to your original patch, then.

lgritz:
All merged into master and RB-1.7. Thanks for the fixes!

PR #1525, opened 2016-10-10
Make temporary texture filename unique to avoid classes on shared drives

lgritz:
maketx (and IBA::make_texture underneath) writes its results to a temporary file to avoid overwriting an existing file and leaving an invalid file there if the command should be terminated before it's done. After it's done being written, the temp file is "moved" to the final destination, replacing it atomically.

But as Ramon points out, if multiple processes are running on the same shared disk partition and might try to maketx the same texture at the same time, they could clobber each other by both writing to identically named temp files. So the solution is to ask the OS to make a unique name for that temporary file, using Filesystem::unique_path().

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -828,8 +828,11 @@ make_texture_impl (ImageBufAlgo::MakeTextureMode mode,
     // chance a crash during texture conversion will leave behind a
     // partially formed tx with incomplete mipmaps levels which happesn to
     // be extremely slow to use in a raytracer.
+    // We also force a unique filename to protect against multiple maketx
+    // processes running at the same time on the same file.
     std::string extension = Filesystem::extension(outputfilename);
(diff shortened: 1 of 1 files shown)


rmv:
Thanks Larry! LGTM

The move/rename from a file in the same folder is atomic on OSX/unix/Windows so this should ensure that the final .tx is valid even if concurrent processes try to do it at the same time.

That workflow is not desirable. But with this change at least there will be no corrupted .tx left behind.

lgritz:
I'll merge this now, I think it's a good temp solution.

For longer term, what are you thinking? Some kind of file locking to prevent two processes from maketx'ing the same texture at once? So the second one will then have to wait for the first one to finish? I can think of all sorts of subtle bugs -- can the first one grab a lock and then crash or be terminated in such a way that it dies without releasing the lock, and then the second one doesn't progress either? It might be better to just tolerate the occasional redundant work in order to assure that whole new classes of bugs or race conditions don't need to be considered.

rmv:
I completely agree with your previous comment about tolerating some redundant work to avoid the thornier issue of reliable synchronization, that was part of the motivation for this.

PR #1527, opened 2016-10-11
Add ANSI terminal support to Windows platform

darkhorse64:
## Description

Windows 10 supports VT100 escape sequences since the Anniversary Edition. Except for the define, this patch uses APIs that have been present for quite some time, so it should fail gracefully on older versions of Windows
## Tests

Test  with simd_test. It looks much better !

------

### DIFF
--- a/src/libutil/sysutil.cpp
+++ b/src/libutil/sysutil.cpp
@@ -63,6 +63,7 @@
 
 #ifdef _WIN32
 # define WIN32_LEAN_AND_MEAN
+# define DEFINE_CONSOLEV2_PROPERTIES
 # include <windows.h>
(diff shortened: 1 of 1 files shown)


lgritz:
Nice!  I assume this is harmless on older versions of Windows? Does this impose any limitation on OS or compiler versions?

darkhorse64:
Sorry for the bad commit. Continuous integration is a nice thing.

Regarding compiler version, none. Regarding OS version, Windows 2000 is the minimal requirement but, of course, you need Windows 10 for a meaningful result.

lgritz:
LGTM, I will merge.

PR #1528, opened 2016-10-12
FindOpenColorIO.cmake: Fix for location of OCIO headers

nrusch:
## Description

Fixes the search for the OpenColorIO headers by removing the trailing `OpenColorIO` subdirectory from the `OCIO_INCLUDES` result, since the includes in `color_ocio.cpp` are already qualified.

Confirmed working on RB-1.6.


lgritz:
LGTM, merging.

andrewhwood:
thank you!

PR #1530, opened 2016-10-13
Fix for UDIM channels query

lgritz:
Logic was botched, it was still getting errors if the first file was not
present. Needed to verify the file and check that it wasn't broken,
because only in verify is it actually opened. Just checking for a
non-NULL result from resolve_udim() isn't enough.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -2501,20 +2501,17 @@ ImageCacheImpl::get_image_info (ImageCacheFile *file,
         // if different panels of the same UDIM scheme have different
         // numbers of channels! Search the 10x100 files for a match, give
         // up if not found.
-        ImageCacheFile *concretefile = NULL;
-        for (int j = 0; j < 100 && !concretefile; ++j) {
+        for (int j = 0; j < 100; ++j) {
             for (int i = 0; i < 10; ++i) {
                 float s = i + 0.5f, t = j + 0.5f;
-                if (ImageCacheFile *c = resolve_udim (file, s, t)) {
-                    concretefile = c;
-                    break;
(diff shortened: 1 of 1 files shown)


caron:
Sweet! We have been getting UDIM tile 1001 missing in our Arnold renders for ages now. Gonna ask the devs if they can make sure to get this fix.

lgritz:
This fix is only for the gettextureinfo side of things.

There's a secondary issue that I'm tackling tomorrow, about what should happen to the texture() lookup itself if the u,v it's given happen to correspond to a nonexistent udim file. (That is, if the udim set is *sparse*.) Currently it issues an error, and I think that's not what we want. We're still haggling over the best solution.

Stay tuned, I'm going to try to get something together tomorrow.

lgritz:
Oh, I *think* that you can work around that problem now if you use the "missingtexture" option -- this by design tells the texture lookup that in the event of a nonexistant or broken texture, to substitute the color for the result and not consider it an error. Perhaps the expedient solution to your errors is just to pass a reasonable missingcolor value for your UDIM textures.

caron:
I talked with Thiago, it seems this isn't related to what we are experiencing. So, sorry about that... But! UDIM best practices are still a bit loose and any improvement in that area is appreciated. The missingcolor value can be useful, but in this case the 1001 is never actual rendered. I am 100% sure there are NO uv values in that tile. I will talk with SolidAngle directly. Again, sorry for the noise.

lgritz:
We're learning as we go, too, which is why these fixes are kind of dribbling out bit by bit. 

UDIM is the gift that keeps giving -- every time we think we've got it, we discover another subtle assumption we'd had about ordinary texture that is different for UDIM.

Overall it seems like a good idea, but I'm trying to get all the loose ends tied up.

Issue #1455, opened 2016-07-22
Nuke txReader doesn't restore tx_mip_level properly


dbr:
With the txReader plugin (at least in Nuke 10.0v3, didn't check in older versions yet):

1) load a .tx file into Read node
2) select a different "mip level" option
2) do something like cut and paste the node, or save and reopen the script

This produces an error such as `Read1.tx_mip_level: Bad value for tx_mip_level : 1`

I assume this is because Nuke is restoring the knob value earlier than the `txFmt_->setMipLabels(mipLabels);` call occurs, so there is no options in the dropdown.

As a workaround I just changed the dropdown into a basic integer knob:
```c++
        mipLevelKnob_ = Enumeration_knob(cb, &mipLevel_, EMPTY,
                                         "tx_mip_level", "mip level");
```
..to:
```c++
        mipLevelKnob_ = Int_knob(cb, &mipLevel_, "tx_mip_level", "mip level");
```

..and commented the content of the `setMipLabels` method.

Not the best solution as it loses the nice descriptive labels on the parameter which say the size of each level.. but as a short-term fix it's preferable to not saving properly (and is backwards/forwards compatible as the knob is serialised the same way)

nrusch:
Interesting. I wonder if the order of operations in Reader initialization and loading changed in 10.0. I'll try to take a quick look at this tomorrow to figure out more.

nrusch:
OK, I think I have a decent solution for this. Basically, it involves using a hidden `Int_knob` to store the actual mip level index, but still exposes the pulldown of nice labels to the user.

I was toying with the idea of just throwing the `SAVE_MENU` flag on the enumeration knob, but that feels a little too brute-force, and would create a much larger footprint for the knob's serialized data.

I've tested the two-knob solution, and it appears to be working properly, so I'll put a pull request together. It would be great if you could test it as an outside party so I can be sure I'm not overlooking anything obvious. :)

nrusch:
@dbr Pull request is up if you get the chance to give it a spin any time soon.

lgritz:
Merged the PR, I think this is fixed now.

PR #1531, opened 2016-10-13
nuke: Fix txReader to properly restore saved mip level knob value

nrusch:
## Description

Proposed fix for https://github.com/OpenImageIO/oiio/issues/1455.

The issue is caused by the fact that the user-facing `Enumeration_knob` is populated from within the Reader's constructor, which is too late when the Read is created with a non-default mip level (e.g. when loading a saved script).

To work around this without resorting to saving all of the knob's menu items, the menu index is no longer stored by the Read or used in the op hash. Instead, its value is copied to a hidden `Int_knob`, which drives the actual Reader and gets saved when the Read is serialized.

Also, the txReader now sets the bit-depth metadata key properly when the input file contains `half` data, which was an earlier oversight.

### DIFF
--- a/src/nuke/txReader/txReader.cpp
+++ b/src/nuke/txReader/txReader.cpp
@@ -35,17 +35,38 @@ static const char* const EMPTY[] = {NULL};
 
 class TxReaderFormat : public ReaderFormat {
     int mipLevel_;
+    int mipEnumIndex_;
     Knob* mipLevelKnob_;
+    Knob* mipLevelEnumKnob_;
 
 public:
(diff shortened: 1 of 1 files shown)


dbr:
Seems good to me - thanks!

lgritz:
fixes #1455 
(I think you have to say it like that)

lgritz:
LGTM. I'll test it on my end and if it builds properly, I'll merge it into master and RB-1.7.

lgritz:
Builds for me. Merging.

PR #1534, opened 2016-10-16
DICOM reader. DICOM is the main format used for medical imaging.

lgritz:
A little rough around the edges, but seems to work. We'll probably come
back and revise as shortcomings are discovered, but this is enough to
get started.


lgritz:
Updated the patch slightly, to turn off the dcmtk library's icky logging of error messages to the console.

lgritz:
What do you think, @danwexler ?

PR #1535, opened 2016-10-18
Replace BOOST_FOREACH with C++11 'range for'

lgritz:
For master only, where C++ is the minimum.

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -38,8 +38,6 @@
 #endif
 #include <vector>
 
-#include <boost/foreach.hpp>
-
 #include <QtCore/QSettings>
 #include <QtCore/QTimer>
 #include <QtGui/QApplication>
@@ -158,7 +156,7 @@ ImageViewer::ImageViewer ()
 
 ImageViewer::~ImageViewer ()
 {
-    BOOST_FOREACH (IvImage *i, m_images)
+    for (auto i : m_images)
         delete i;
 }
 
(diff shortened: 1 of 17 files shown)


src/iv/imageviewer.cpp
```diff
@@ -595,7 +593,7 @@ ImageViewer::readSettings (bool ui_is_set_up)
     linearInterpolationBox->setChecked (settings.value ("linearInterpolation").toBool());
     darkPaletteBox->setChecked (settings.value ("darkPalette").toBool());
     QStringList recent = settings.value ("RecentFiles").toStringList();
-    BOOST_FOREACH (const QString &s, recent)
+    for (auto&& s : recent)
```
fpsunflower: Did you find a good set of guidelines on when to use `auto` vs `auto&` vs `auto&&` vs any of the `const` prefixed variants? Its one aspect of C++11 I'm not 100% clear on yet.

lgritz:
Here's my current understanding:

It should be obvious that `auto` makes a copy by value every time (fine for something like an `int` array), `const auto&` is a reference that's fine if the structure you're iterating over is const, and `auto&` is a reference that lets you modify the objects.

`auto&&` **if I understand correctly** means it's an a magic rvalue reference that will probably retain either the const or non-const of the items. So in theory, this is the only one you actually need.

Early C++11 guides seemed to say that `auto&&` would be the preferred form, but pointed out that nobody knows what it means so maybe you want regular `auto` or `const auto` just to be clear. I can understand that in 2011, but in 2016 I'm leaning toward favoring `auto&&` everywhere, people should know what that means.

But, I'm new at this and am a little wishy-washy. Sometimes I look at the loop and think "that would be slightly more clear if I spelled out that I think it's const" and occasionally I even think "in this context, it would be more clear if I put the full type rather than use `auto`", so you'll see a mixture; I admit that I haven't yet converged on a 100% consistent style. I may get over this hesitation and use `auto&&` everywhere soon.

fpsunflower:
I did some brief googling and `auto&&` does seem to be the recommended way of doing things.

LGTM!

PR #1538, opened 2016-10-19
Qt migration WIP

lgritz:
* strip directories from #includes of Qt headers.
* Reorder OpenGL-related headers.

This is preliminary work towards eventual migration to Qt5.

The original header locations are incompatible with Qt5, but if we
eliminate the directories, they will mostly work for both Qt4 and Qt5.

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -28,9 +28,6 @@
   (This is the Modified BSD License)
 */
 
-#include "imageviewer.h"
-#include "ivgl.h"
-
 #include <iostream>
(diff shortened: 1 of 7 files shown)


ilovezfs:
Did you get it to build with Qt 5?

lgritz:
@ilovezfs : No, this is not the full Qt5 solution, but it is a necessary step along the way.

ilovezfs:
Ah ok. I was wondering if you'd already made it happen locally :)

lgritz:
I'm getting close...

PR #1539, opened 2016-10-19
Fix warnings on some 32 bit platforms

lgritz:
For some reason, certain 32 bit platforms have gcc errors about this spot leading to uninitialized variables (they don't! but gcc can't seem to figure that out). The 64 bit platforms figure it out just fine, so just eat the cost of a couple useless assigns on the platforms where it's a problem.


hobbes1069:
Applies cleanly to 1.7.7 and seems to work! LGTM.

lgritz:

PR #1540, opened 2016-10-19
oiiotool should exit with non-zero status when command-line args fail to parse properly

mike1158:
oiiotool should exit with non-zero status when command-line args fail to parse properly.


lgritz:
yes, LGTM

lgritz:
Merged into master and RB-1.7

Issue #1537, opened 2016-10-18
Poor/regressive performance using oiiotool to strip channels out of an exr


mike1158:
This command:

```
oiiotool input.exr --ch R,G,B,A -d half -o output.exr
```

seems to take an inordinate amount of time in 1.6.14 or 1.7.7.  It is faster, but still, I think, underperforming in 1.5.14.  Note the very high system time for 1.6.14 and 1.7.7:

From OpenImageIO-1.5.14:
$ time oiiotool input.exr --ch R,G,B,A -d half -o output.exr
real    0m34.560s
user    0m42.027s
sys     0m0.398s

From OpenImageIO-1.6.14:
$ time oiiotool input.exr --ch R,G,B,A -d half -o output.exr
real    1m56.511s
user    2m45.863s
sys     6m20.403s

From OpenImageIO-1.7.7:
$ time oiiotool input.exr --ch R,G,B,A -d half -o output.exr
real    1m15.012s
user    1m57.619s
sys     6m43.690s

OS is CentOS6.6.  All three versions of OIIO were built against OpenEXR 2.2.0.

Original input.exr file is 20M, output.exr is 279K.  I can supply the input.exr on request.  Output of "oiiotool --info input.exr":

input.exr            : 5760 x 2880, 50 channel, half/half/half/half/float/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half/half openexr

Sorry, I haven't had time to dig into possible causes yet, but the performance regression seemed bad enough to log an issue.

Cheers,
-miker

lgritz:
Whoa, that's crazy! I would have expected a few seconds, not a minute or more. 

I'm working on a theory. Would you mind running it again (with 1.7) using the "--runstats" option and paste the output?

Also, perhaps I can get a peek at the file itself? Is there a version I can download?

lgritz:
I tried to reproduce on my end (just using `oiiotool -create 5760x2880 50 -d half -o in.exr`), and then the subsequent `oiiotool in.exr -ch R,G,B,A -o out.exr` runs in about 18 seconds for me, is about the same time regardless of whether I use 1.5/1.6/1.7/master, and in no case does it have a huge 'sys' time. This is all on CentOS 7, and happening over a network disk partition. (Perhaps you can also try that on your end -- does it have similar runtime as your own image?)

[...]

PR #1541, opened 2016-10-20
oiiotool -i extensions that make channel extraction efficient

lgritz:
This is spawned by the observation in issue #1537 that extracting a small number of channels from a many-channeled file is very expensive. This is exacerbated when the image is large, has a huge number of channels, not all channels having the same data type, and the file is untiled.

In #1537, the situation was a 50 channel scanline EXR that was a mixture of half and float, resolution 5760x2880, and a simple extraction of the R,G,B,A channels took 1:34. 

This set of patches accomplishes the following:
- Improvements to the logic that handles per-channel format conversion when reading scanline files: (a) use threads; (b) convert batches of adjacent channels that have the same data type, instead of handling each channel individually.
- ImageBuf::read() is extended with a new variety that lets you populate the IB with a (contiguous) subset of channels existing in the file being read. This always saves the buffer memory of the unneeded channels and the computation required to do any data format conversions of those channels, and for some image file formats might also save disk I/O (it does for OpenEXR, but not for most file formats).
- oiiotool option `-i:type=...` explicitly overrides the internal datatype, rather than converting internally to float (which is good for speeding up math computation on pixels, but not helpful when just shuffling channels around).
- oiiotool option `-i:ch=...` reads only the listed channels of the input. I works just like having a normal input and then a separate `--ch` command, except that by knowing at input time which channels you want, it has the potential to save all the buffer memory, type conversion, and data copying of the unneeded channels, and for some formats (EXR - yes! but not most others) can also save I/O for the unneeded channels.

| time | peak mem | command |
| --- | --- | --- |
| 1:34 | 3.4 GB | **original**: `oiiotool input.exr -ch R,G,B,A -d half -o out.exr` |
|  |  | _after per-channel logic improvement:_ |
| 0:32 | same | `oiiotool input.exr -ch R,G,B,A -d half -o out.exr` |
| 0:10 | 0.2 GB | `oiiotool -i:type=half input.exr -ch R,G,B,A -d half -o out.exr` |
| 0:02 | 0.4 GB | `oiiotool -i:ch=R,G,B,A input.exr -d half -o out.exr` |
| 0:01 | 0.2 GB | `oiiotool -i:type=half:ch=R,G,B,A input.exr -d half -o out.exr` |

Bottom line: on my machine, this improves time by at least 60x and memory by 15x. Pretty good for one (very very late) evening's work.

Let that be a lesson to you: giving me a specific example file + command line that performs slower than you'd like can sometimes result in really big improvements!


lgritz:
@mike1158 Ping?

lgritz:
@mike1158 ?

mike1158:
Sorry for taking so long to get back to you.  I applied the patch and get effectively the same results you do.  Thanks again!
-miker

PR #1542, opened 2016-10-24
Allow multiple key/value pairs in OCIO wrappers.

lgritz:
Extend ColorConfig::createLookTransform & createDisplayTransform to
accept mutiple key/value pair, by allowing their context_key and
context_value paramters to each be comma-separated lists (which
obviously must both contain the same number of items).


lgritz:
Merge

PR #1547, opened 2016-10-28
Fix to TIFF handling of certain unusual tags, which also affected raw files

lgritz:
Many RAW files are really structured like TIFF underneath, and we
forcefully open them as a TIFF file to read their metadata, which is not
read as well by libraw. OK, but they also contain some of these oddball
tags, and that could lead to crashes. This patch makes it more robust to
some of these odd cases.


fpsunflower:
LGTM

PR #1548, opened 2016-10-28
templatize round_to_multiple so it works with other types (like size_t)

lgritz:
Caused me to also modify the template to clamp to prevent a cascade of type errors.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -137,12 +137,12 @@ pow2rounddown (int x)
 
 
 
-/// Round up to the next whole multiple of m.
-///
-inline int
-round_to_multiple (int x, int m)
+/// Round value up to the next whole multiple.
+/// For example, round_to_multiple(7,10) returns 10.
+template <typename V, typename M>
+inline V round_to_multiple (V value, M multiple)
 {
-    return ((x + m - 1) / m) * m;
+    return V (((value + V(multiple) - 1) / V(multiple)) * V(multiple));
 }
 
 
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -994,7 +994,7 @@ ImageBuf::write (ImageOutput *out,
         } else {
             // Big scanline image: break up into scanline strips
             imagesize_t slsize = bufspec.scanline_bytes();
-            int chunk = clamp (round_to_multiple(budget/slsize, 64), 1, 1024);
+            int chunk = clamp (round_to_multiple(int(budget/slsize), 64), 1, 1024);
             std::unique_ptr<char[]> tmp (new char [chunk*slsize]);
             for (int z = 0;  z < outspec.depth;  ++z) {
                 for (int y = 0;  y < outspec.height && ok;  y += chunk) {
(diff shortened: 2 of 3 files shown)


src/include/OpenImageIO/fmath.h
```diff
@@ -137,12 +137,12 @@ pow2rounddown (int x)
...
 
 
-/// Round up to the next whole multiple of m.
-///
-inline int
-round_to_multiple (int x, int m)
+/// Round value up to the next whole multiple.
+/// For example, round_to_multiple(7,10) returns 10.
+template <typename V, typename M>
+inline V round_to_multiple (V value, M multiple)
```
fpsunflower: Why does `multiple` need its own type? Shouldn't it be `V` as well?

Do we have unit tests for these functions? (in particular to exercise to mixed argument or overflow cases).

src/include/OpenImageIO/fmath.h
```diff
@@ -214,9 +214,9 @@ OIIO_FORCEINLINE uint64_t rotl64 (uint64_t x, int k) {
 
 
 /// clamp a to bounds [low,high].
-template <class T>
+template <class T, class S>
 inline T
-clamp (T a, T low, T high)
+clamp (T a, S low, S high)
```
fpsunflower: I'm not sure I follow what situation would lead to more type errors with all arguments being of the same type. It seems its mixing types that would lead to trouble, for example:

`size_t clamp(size_t, int ,int)` would complain about comparing signed and unsigned?

fpsunflower:
As a rule of thumb, I find the template argument rules of C++ so confusing, that I find it makes me feel safer if there is only one argument type that needs to be deduced. Particularly when combining with the funny rules for float vs int vs unsigned automatic casts...

It does mean more warnings sometimes, but usually helps clarify the intent.

src/include/OpenImageIO/fmath.h
lgritz: I wanted it to be the same type, but then you have:

    uint32_t x = ...;
    uint32_t y = round_to_multiple (x, 4);

4 is an int, so you get an error.

We DO have unit tests for this, though they studiously use the same type. I'll set up additional unit test cases that would have failed before and now succeed.

lgritz:
So, if I make these have only one type, then you have to say things like:

    x = clamp (foo.size(), size_t(0), size_t(16));

instead of the much more readable

    x = clamp (foo.size(), 0, 16);

lgritz:
We can abandon this. The other thing I was doing that led me down this path has another way to express it.

src/include/OpenImageIO/fmath.h
fpsunflower: I guess `round_to_multiple (x, 4u)` is a bit less nice to read.

I would at least change the implementation to:
```c
return V (((value + V(multiple) - 1) / V(multiple)) * V(multiple));
```
Just to avoid the rules about mixed argument arithmetic.

src/include/OpenImageIO/fmath.h
lgritz: Those suggestions sound totally reasonable. Let me look at it, but in any case I no longer need to do this before the other thing I was really trying to submit.

lgritz:
Pushed a revision -- abandoned the "clamp" part, used your suggestion for the round_to_multiple template, and added unit test.

lgritz:
I've cut back on this and what I kept, I did exactly what Chris recommended. So I'm going to merge.

Issue #1550, opened 2016-10-31
FTBFS on rebuilding the documentation


mfvescovi:
### Versions ###
* OIIO version: 1.7.7
* OS: Debian unstable/sid
* C++ compiler: gcc 6.2.0

Hi Larry.

As reported early this month via IM, compiling OIIO v1.7.7 in Debian unstable/sid for testing purpose before the official upload to Debian archives fails while building documentation stuff, namely the figures set prepared and bundled in the PDF file.

A snippet of the buildlog follows:
--------------- >8 ---------------
make[2]: Entering directory '/<<PKGBUILDDIR>>/src/doc'
/usr/bin/make figures
make[3]: Entering directory '/<<PKGBUILDDIR>>/src/doc'
(cmake -E make_directory figures ; \
 cd figures ; \
 OIIOTOOL=""../../../build/src/oiiotool/oiiotool"" bash ../makefigures.bash ; \
)
Using OIIOTOOL=../../../build/src/oiiotool/oiiotool
oiiotool ERROR: colorconvert : Could not construct the color transform
oiiotool ERROR: colorconvert : Could not construct the color transform
oiiotool WARNING: -o : Can't save 4 channels to jpeg... saving only R,G,B
oiiotool WARNING: -o : Can't save 4 channels to jpeg... saving only R,G,B
make[3]: Leaving directory '/<<PKGBUILDDIR>>/src/doc'
--------------- >8 ---------------

Later on, PDF creation fails with:
--------------- >8 ---------------
Overfull \hbox (34.445pt too wide) in paragraph at lines 2029--2032
[]\OT1/ptm/m/n/10.95 If \OT1/pcr/m/n/10.95 oiiotool\OT1/ptm/m/n/10.95 's global
 \OT1/pcr/m/n/10.95 -a \OT1/ptm/m/n/10.95 flag is used (\OT1/ptm/b/n/10.95 a\OT
1/ptm/m/n/10.95 ll subim-ages)), or if the op-tional \OT1/pcr/m/n/10.95 --crop:
allsubimages=1
<use figures/tahoe-small.jpg>

LaTeX Warning: File `figures/crop.jpg' not found on input line 2043.

! Package pdftex.def Error: File `figures/crop.jpg' not found.

See the pdftex.def package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.2043 ...egraphics[width=1.5in]{figures/crop.jpg}
                                                  
[...]

PR #1551, opened 2016-11-01
Fix errors when doing identify transform when no OCIO is present.

lgritz:
Well, sometimes you (and by "you" I mean some script you wrote) ask
oiiotool to do a 'colortransform sRGB sRGB' or some other color
transform that is just identity. Fine. But if $OCIO environment variable
is not defined so no OCIO config is found, that was an error!

Simple fix -- detect that case and return a do-nothing color processor.


lgritz:
Fixes #1550

PR #1552, opened 2016-11-04
colormap

lgritz:
* ImageBufAlgo::color_map (C++ and Python)
* oiiotool --colormap

Both take the intensity of each pixel in the input image (either one channel, or the luminance of its RGB channels) and run it through a linearly interpolated color map. Built in color maps that can be specified by name include "blue-red", "spectrum", and "heat". Or raw values can be used.

Example:

coloromap-spectrum

Along the way:

* added a matrix44 constructor from 16 floats
* minor fixes to array_view_strided
* fmath: add interpolate_linear utility


fpsunflower:
LGTM!

PR #1553, opened 2016-11-04
oiiotool: New expression substitution: MINCOLOR, MAXCOLOR, AVGCOLOR

lgritz:
In the expression substitution context, these "pseudo-metadata" names
will evaluate to be comma-separated lists of the per-channel pixel value
minimum, maximum, and average, respectively.

As an example, the following oiiotool command line will do contrast
stretching, remapping intensity values to fill the full range of [0-1]:

    oiiotool in.tif -subc {TOP.MINCOLOR} -divc {TOP.MAXCOLOR} -o out.tif

It performs this by first subtracting from every pixel the minimum pixel
value of the input (at that point, the TOP image on the stack), and then
dividing every pixel by the new maximum value, resulting in an output
image whose minimum pixel value is (0,0,0) and maximum is (1.0,1.0,1.0).


fpsunflower:
LGTM!

Very cool! This seems like a very flexible way to do a bunch of things ...

In your example - it should be:
`oiiotool in.tif -subc {TOP.MINCOLOR} -divc {TOP.MAXCOLOR-TOP.MINCOLOR} -o out.tif`

If you actually want the output to be `[0,1]` I think.

lgritz:
Nope! By the time it is evaluating MAXCOLOR, the new "TOP" is the one that's already had MINCOLOR subtracted form it!

fpsunflower:
Oh I see! That's a bit tricky -- but it makes sense.

PR #1556, opened 2016-11-12
thread.h: modernize, thread_pool, parallel_for

lgritz:
* Bring thread_group into modern C++11 era: use std components, no more
  boost dependency, use variadic templates for a better create_thread
  signature. Also remove the last vestigages of using boost::thread_group
  instead of ours.

* Modernize spin_mutex to use C++11 std::atomic_flag instead of doing
  everything with bespoke code.

* thread_pool : persistent threads feeding from a task queue. This is
  generally a better way forward than explicitly spawning new threads
  with thread_group every time you want to do things in parallel.

* task_set : Because the thread_pool is asynchronous (fire and forget,
  or fire and get future's back), a simple task_set object lets you collect
  a set of related asynchronous tasks and wait for them to all have been
  completed before proceeding.

* parallel_for : several easy constructs for a variety of parallel
  loops, using the default thread pool. Makes parallelizing "1-D" loops
  very similar to writing an ordinary 'for' loop, or calling
  std::for_each.

* OIIO attribute("nthreads") sets the size of the thread pool.

The implementation of the thread pool is based on the CTPL project
(Apache license). It's hidden completely in thread.cpp, totally
independent of the thread.h exposed APIs that use the PIMPL idiom, so we
are free to replace the underlying pool implementation if we realize
with more experience that CTPL was not the best choice.

Note also that we have bumped the minimum version of Boost to 1.53 to
accommodate the thread_pool's internal use of boost::lockfree::queue.
That seems reasonable, considering that VFX recommended Boost 1.53 as
the recommended version in 2014, 1.55 in 2015 & 2016. Boost lockfree
queue is header-only, so this doesn't introduce any new linkage or
version mismatch issues.

Read the comments in thread.h for a full explanation, but just for
illustration in this review, here is what it looks like to use it:

Serial tasks:

    for (int i = 0; i < N; ++i)
        job(i);

Parallel tasks using thread_pool, and waiting for them to finish before
proceeding:

    thread_pool *p = thread_pool::get_thread_pool();
    work_set w;
    for (int i = 0; i < N; ++i)
        w.push (p->push (job, i));
    w.wait();

Serial loop:

    for (int i = 0; i < N; ++i) {
        a[i] += b[i];
    }

Parallel loop, will divide the range among threads:

    parallel_for (0, N, [&](int64_t i){
        a[i] += b[i];
    });

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -39,24 +39,18 @@
 #ifndef OPENIMAGEIO_THREAD_H
 #define OPENIMAGEIO_THREAD_H
 
+#include <algorithm>
+#include <atomic>
+#include <future>
+#include <memory>
+#include <mutex>
+#include <thread>
 #include <vector>
 
 #include "oiioversion.h"
+#include "export.h"
 #include "platform.h"
 #include "atomic.h"
-
-
-
-#if OIIO_CPLUSPLUS_VERSION >= 11
-# include <thread>
-# include <mutex>
-# include <atomic>
-#else   /* prior to C++11... */
-  // Use Boost mutexes & guards when C++11 is not available
-# include <boost/version.hpp>
-# include <boost/thread.hpp>
-#endif
-
 
 
 
@@ -108,35 +102,11 @@ public:
 
 
 
-#ifdef NOTHREADS
-
-// Definitions that we use for debugging to turn off all mutexes, locks,
-// and atomics in order to test the performance hit of our thread safety.
-
-typedef null_mutex mutex;
-typedef null_mutex recursive_mutex;
-typedef null_lock<mutex> lock_guard;
-typedef null_lock<recursive_mutex> recursive_lock_guard;
-
-#elif OIIO_CPLUSPLUS_VERSION >= 11
-
-typedef std::mutex mutex;
-typedef std::recursive_mutex recursive_mutex;
+using std::mutex;
+using std::thread;
+using std::recursive_mutex;
 typedef std::lock_guard< mutex > lock_guard;
 typedef std::lock_guard< recursive_mutex > recursive_lock_guard;
-typedef std::thread thread;
-
-#else
-
-// Fairly modern Boost has all the mutex and lock types we need.
-
-typedef boost::mutex mutex;
-typedef boost::recursive_mutex recursive_mutex;
-typedef boost::lock_guard< mutex > lock_guard;
-typedef boost::lock_guard< recursive_mutex > recursive_lock_guard;
-typedef boost::thread thread;
-
-#endif
 
 
 
@@ -208,21 +178,9 @@ private:
 
 
 
-
-
-#ifdef NOTHREADS
-
-typedef null_mutex spin_mutex;
-typedef null_lock<spin_mutex> spin_lock;
-
-#else
-
-// Define our own spin locks.
-
-
 /// A spin_mutex is semantically equivalent to a regular mutex, except
 /// for the following:
-///  - A spin_mutex is just 4 bytes, whereas a regular mutex is quite
+///  - A spin_mutex is just 1 byte, whereas a regular mutex is quite
 ///    large (44 bytes for pthread).
 ///  - A spin_mutex is extremely fast to lock and unlock, whereas a regular
 ///    mutex is surprisingly expensive just to acquire a lock.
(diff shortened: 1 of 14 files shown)


src/include/OpenImageIO/thread.h
```diff
@@ -505,45 +430,288 @@ class mutex_pool
...
     }
-    template<typename FUNC>
-    thread *create_thread (FUNC func) {
-        lock_guard lock (m_mutex);
-        thread *t = new thread (func);
-        m_threads.push_back (t);
+
+    template<typename FUNC, typename... Args>
+    thread *create_thread (FUNC func, Args&&... args) {
+        thread *t = new thread (func, args...);
```
fpsunflower: I think you want to use `std::forward()` here?

src/include/OpenImageIO/thread.h
```diff
@@ -505,45 +430,288 @@ class mutex_pool
...
+///
+///    void myfunc (int id) { ... do something ... }
+///
+///    thread_pool& pool (default_thread_pool());
+///    {
+///        task_set<decltype(myfunc())> tasks;
+///        // Launch a bunch of tasks into the thread pool
+///        for (int i = 0; i < ntasks; ++i)
+///            tasks.push_back (pool->push (myfunc));
+///    } // <--- MAGIC!
```
lgritz: sure

src/include/OpenImageIO/thread.h
lgritz: Where, precisely? 

Can you articulate why forward is needed here and what purpose it serves? And under what circumstances I'm *not* supposed to use forward? I admit that I'm still working on getting the hang of it.

lgritz:
I have a possible amendment to this... I'm now thinking that I'd like to keep the thread_pool here in thread.h, but move the parallel_for family of helpers into a separate parallel.h (which will grow to include other things, like 2D loops).

Should I (a) amend this PR with that move; (b) keep this PR as-is, but everybody knows I'll follow it up with a move and other things; (c) no, that's a bad idea, keep it all in thread.h even as more parallel functions are added?

src/include/OpenImageIO/thread.h
lgritz: Like this?

    template<typename FUNC, typename... Args>
    thread *create_thread (FUNC func, Args&&... args) {
        thread *t = new thread (func, std::forward<Args>(args)...);
        add_thread (t);
        return t;
    }

lgritz: So is the rule that any time a templated function takes a T&& parameter and passes it to another templated function expecting T&&, it needs a std::forward<T>()?

What's the failure circumstance when I neglect to do that?

lgritz:
My fooling around tree is a batch of changes ahead of the preliminary work shown here. I'm really having fun using those parallel_for loops in various places, and I have 2D versions that are quite slick for image-like things.

src/include/OpenImageIO/thread.h
lgritz: The example I just gave above, is that the way you would recommend writing this function?

fpsunflower: Yes.

lgritz:
Updated the PR to address Chris and Nick's comments.

lgritz:
I updated the PR with a commit that moves the parallel_for functions to a separate parallel.h.
If you don't like this, I'm happy to back it out. But take a look and weigh in. I like it better this way.

No code changed in this update, it just moved from one header to another (and split part of thread_test.cpp into parallel_test.cpp as well).

meshula:
I'm all for factoring it that way.

meshula:
I'm pretty excited to see this work, I learned a bit from it :)

lgritz:
Haha, I'm learning, too -- I discovered a horrible flaw in my logic that I'm patching up right now!

Turns out the things go horribly awry if there is indirect recursion -- for example, if you submit more tasks to the pool than there are threads, and then the tasks themselves call something that wants to add to the thread pool and expects to wait until it's done. Deadlock!

I have a fix that I really like. I'm literally minutes from submitting the PR.

And then the next batch of changes, which is how I discovered this flaw, is also very exciting -- I have "2D" parallel_for constructs, and a new thread_pool-based version of parallel_image that greatly simplifies the internals of basically every ImageBufAlgo function we have. It's really cool, I was extremely full of myself for doing it so cleverly, right up until I found the deadlock problem. :-(

lgritz:
#1560 is the crucial fix

fpsunflower:
That sounds exciting. Do you have any numbers on how this new construct compares to what we were doing before?

lgritz:
It did seem slightly faster, but I need to rerun all my tests in light of the new "work-stealing" patch.

lgritz:
For those following along, on my work Linux box, the old thread_group could spawn threads in groups of 12 at a rate of around 75,000/second (for a trivial task -- just returns), whereas with the new thread_pool I'm timing it as being able to submit around 240,000 tasks/second to the pool. So the overhead of submitting a parallel task is about 1/3 what it was before, plus it's just dead simple to write a parallel loop with the new constructs.

PR #1560, opened 2016-11-17
Fix threadpool

lgritz:
I discovered that the thread_pool could deadlock if the pool queue
filled and then the subtasks themselves added further subtasks to
the queue and waited for their results (since those subtasks would
never themselves get around to executing, with all the pool threads
waiting).

The solution is primarily to modify task_set::wait() so that instead
of unconditionally blocking for the subtasks to complete, it polls
them without blocking, and so when it finds itself waiting for queue
results to still complete, it steals some work from the queue to do
itself. This unclogs the system nicely.

I also modified parallel_for_chunked (which is the basis for the
unchunked parallel_for as well) so that the calling thread itself
executes the last (or only) subtask rather than send it to the pool.
This is yet another way to reduce pressure on the queue, and especially
for cases where you call parallel_for but only one task is necessary
(due to chunking size), it avoids any overhead of using the pool.

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -520,8 +520,11 @@ private:
 ///
 class OIIO_API thread_pool {
 public:
-    /// Initialize the pool.  If the number of threads is 0, it will be
-    /// set to the number of cores on the machine.
+    /// Initialize the pool.  If nthreads is 0, the pool will be set to one
+    /// less than the number of hardware cores on the machine (this
+    /// preventing over-threading, on the assumption that the thread that
+    /// submits to the pool will itself either have other tasks or will
+    /// steal work from the pool while it waits.
     thread_pool (int nthreads = 0, int queue_size = 128);
     ~thread_pool ();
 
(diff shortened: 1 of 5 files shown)


src/include/OpenImageIO/thread.h
```diff
@@ -601,14 +616,33 @@ OIIO_API thread_pool* default_thread_pool ();
...
+                auto status = f.wait_for (wait_time);
+                if (status != std::future_status::ready)
+                    all_finished = false;
+            }
+            if (all_finished)   // All futures are ready? We're done.
+                break;
+            // We're still waiting on some tasks to complete. What next?
+            if (++tries < 4)    // First few times,
+                continue;       //   just busy-wait, check status again
+            // Since we're waiting, try to run a task ourselves to help
```
fpsunflower: Wouldn't this lead to "over-threading"? Or are we assuming the pool is using `(num_cores-1)` at most?

lgritz: Perhaps we should make the default (threads=0) be cores-1, and just assume that the calling thread will participate?

fpsunflower: That sounds reasonable to me.

lgritz: I updated the PR with this change. The default (or if you explicitly pass nthreads=0 to the constructor or to resize) now means "number of cores - 1".  If you pass a nonzero number of cores, it takes your request literally and it's up to you to size the pool (or adjust your use of it) to avoid over-threading. I also added an optional parameter to task_set::wait() that lets you choose to do a true wait and NOT pick up any of the work yourself, but the default is to contribute to the work rather than blocking.

lgritz:
PR updated (not sure if it CC'd you when if it thinks my comment was about an outdated push)

PR #1564, opened 2016-11-18
fmath: Avoid edge case bug in interpolate_linear

lgritz:
The lerp(y[seg], y[seg+1], x), if on the very last segment and x ==
1.0f, was possible to read past the end of y into uninitialized memory.
Even though it would weight that value by 0, if it happened to contain a
NaN, NaN*0 == NaN, and it was also possible to read past the end of
usable memory. So it is safer to use an appropriate clamp to be super
sure we never read past the end of y.


fpsunflower:
LGTM

PR #1566, opened 2016-11-22
IC: be robust to changes in autotile after opening a file.

lgritz:
Changing the IC's "autotile" value triggers a call to invalidate_all(),
but even there, it was not recognizing that as a condition to invalidate
a file. For untiled files in the cache, the file should be invalidated
if the "virtual tile size" doesn't match the autotile size.

(This PR is a patch against RB-1.7, but obviously I will apply it to master as well.)


lgritz:
@ThiagoIze Confirms that this fixed his crash.

PR #1573, opened 2016-11-29
Fix seg fault at initalization in debug mode

lgritz:
(Fix discovered by J Robert Ray)

When using gcc 4.8 on centos7, with debug compile, in a certain app, we get core dumps involving SIGSEGV in __memcmp_sse4_1 deep inside operator[](std::string&&).

We don't understand it, and in fact suspect a compiler bug. Can't reproduce on other platforms or compiler releases. Sweep it under the rug with this patch, that seems to work around it, though we admit we don't know why.

### DIFF
--- a/src/libOpenImageIO/exif.cpp
+++ b/src/libOpenImageIO/exif.cpp
@@ -274,14 +274,14 @@ static const EXIF_tag_info gps_tag_table[] = {
 
 class TagMap {
     typedef std::map<int, const EXIF_tag_info *> tagmap_t;
-    typedef std::map<std::string, const EXIF_tag_info *> namemap_t;
+    typedef std::map<string_view, const EXIF_tag_info *> namemap_t;
 public:
     TagMap (const EXIF_tag_info *tag_table) {
         for (int i = 0;  tag_table[i].tifftag >= 0;  ++i) {
             const EXIF_tag_info *eti = &tag_table[i];
             m_tagmap[eti->tifftag] = eti;
             if (eti->name)
-                m_namemap[eti->name] = eti;
+                 m_namemap[string_view(eti->name)] = eti;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/libOpenImageIO/exif.cpp
```diff
@@ -280,8 +280,14 @@ class TagMap {
         for (int i = 0;  tag_table[i].tifftag >= 0;  ++i) {
             const EXIF_tag_info *eti = &tag_table[i];
             m_tagmap[eti->tifftag] = eti;
-            if (eti->name)
-                m_namemap[eti->name] = eti;
```
fpsunflower: I'm not sure I like this solution. It feels like something fishy is going on here. What is the value of `eti->name` when it crashes?

fpsunflower: Is this one of those "static initialization order" fiasco type bugs? What if you wrap both tables in a function so they can be initialized on first use instead of on program startup?

fpsunflower: Finally -- I don't think `std::string` is needed at all here. Can't you swap for `string_view` instead?

fpsunflower: Finally^2 -- these table seem small enough that I wonder if `std::map` is needed at all. I bet a good old linear search would run just as fast or faster.

lgritz:
OK, let me take another stab at it. I also suspected initialization order fiasco.

lgritz:
Updated the patch. Moved the static init of the tables into functions that return their reference. Seems to work. I also switched the std::string map keys to string_view.

fpsunflower:
LGTM!

I still think a linear search would probably be fine too (and maybe faster?) But I don't have a good sense of how this is used, or if these tables could get bigger in the future.

PR #1574, opened 2016-11-30
Fix imagecache handling of "displaywindow" and "datawindow"

danieldresser:
Hey, I don't have CLA, but hopefully it's not necessary for a trivial bugfix ( Maybe I should have just made an issue, but I wasn't sure what the issue was until tracking down the code. )

It looks like these return statements were just missed for displaywindow and datawindow - the if statements for every other dataname return true if there is a match.

I noticed this while using OSL, where the symptom is that a call like "r = gettextureinfo (filename, "datawindow", datawin);" will fill out the datawindow correctly, but will incorrectly set r to 0.

The corresponding test case in OSL checks the return values for other gettextureinfo calls, but does not check the return value for displaywindow and datawindow:
https://github.com/imageworks/OpenShadingLanguage/blob/master/testsuite/gettextureinfo/test.osl

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -2588,6 +2588,7 @@ ImageCacheImpl::get_image_info (ImageCacheFile *file,
             d[4] = spec.y + spec.height - 1;
             d[5] = spec.z + spec.depth - 1;
         }
+        return true;
     }
     if (dataname == s_displaywindow && datatype.basetype == TypeDesc::INT &&
         (datatype == TypeDesc(TypeDesc::INT,4) ||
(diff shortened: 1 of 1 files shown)


lgritz:
Oops!  You certainly are correct. I'll merge this immediately, and backport to 1.7 as well.

Interesting that the OSL test doesn't check the return value. I assure you that wasn't on purpose. I guess that's why I didn't notice that the return value was incorrect. Well, like I enjoy saying (but apparently not always taking to heart): "untested code is broken code."

CLA -- right, for simple one-line bug fixes, I'm totally lax about it. It seems totally implausible that anybody can claim that the contents of "return true" is their intellectual property. :-)  But as the patch size increases, I get a little more insistent (less for my protection than yours; at the heart of it, the CLAs are to protect contributors from their employers), and certainly by the time it becomes a "feature" and actual new code that somebody might consider protectable intellectual property, it's crucial.

Issue -- If you have a proposed solution, I prefer just submitting a PR directly. Near simultaneous submission of both an issue and a pull request seems redundant to me, I can't think of what benefit there is to doing twice the paperwork. I think where an "issue" is useful is a situation where you know that something is legit broken, but you don't know the solution, so the issue is kind of a "to do" reminder for the project to eventually fix it and a repository for the information necessary to reproduce the problem.

lgritz:
I will update the test you mention on the OSL side as well.

PR #1577, opened 2016-12-07
New OIIO::getattribute() queries: "input_format_list", "output_format_list"

lgritz:
Just like the existing "format_list" returns a comma-separated list of
all formats supported by any built-in or found plugins, the new queries
return the lists of input and output formats, respectively. An example
of how they may differ is if a particular format is available for
reading but not writing.

At the time of this writing, ffmpeg, psd, ptex, raw, and softimage are
input formats but not output formats. Presently, there are no
output-only formats.

In organizing this, we removed entirely the *output.cpp files for
several formats that did not support outputs, with those now-unneeded
files containing only nonfunctional stubs.


lgritz:
This has languished for a while, but I think it's still a good idea. If nobody has objections, I will merge it tomorrow.

lgritz:
No objections, so I'm merging.

PR #1579, opened 2016-12-08
Add Strutil::printf and fprintf.

lgritz:
They are like std printf/fprintf, except with all the typesafe goodness of Strutil::format (a.k.a. Tinyformat), and also I rigged them to internally use a mutex to prevent simultaneous threads from clobbering each other's console or file output. The full contents of printf's from different threads may end up interleaved non-deterministically, but each print individually will be "atomic" and not get mixed up character-by-character with those of other threads.

These are so much easier now that we are C++11 and can rely on variadic templates.

### DIFF
--- a/src/include/OpenImageIO/strutil.h
+++ b/src/include/OpenImageIO/strutil.h
@@ -44,6 +44,7 @@
 #include <string>
 #include <cstring>
 #include <cstdlib>
+#include <cstdio>
 #include <vector>
 #include <map>
 
@@ -79,16 +80,54 @@ OIIO_NAMESPACE_BEGIN
 /// @brief     String-related utilities.
 namespace Strutil {
 
+/// Output the string to the FILE* stream in a synchronized fashion, so that
+/// buffers are flushed and internal mutex is used to prevent threads from
+/// clobbering each other -- output strings coming from concurrent threads
+/// may be interleaved, but each string is "atomic" and will never splice
+/// each other character-by-character.
+void OIIO_API sync_output (FILE *file, string_view str);
+
+
 /// Construct a std::string in a printf-like fashion.  In other words,
 /// something like:
 ///    std::string s = Strutil::format ("blah %d %g", (int)foo, (float)bar);
 ///
-/// The printf argument list is fully typesafe via tinyformat; format
-/// conceptually has the signature
-///
-/// std::string Strutil::format (const char *fmt, ...);
-TINYFORMAT_WRAP_FORMAT (std::string, format, /**/,
-    std::ostringstream msg;, msg, return msg.str();)
+/// Uses the tinyformat library underneath, so it's fully type-safe, and
+/// works with any types that understand stream output via '<<'.
+template<typename T1, typename... Args>
+std::string format (string_view fmt, const T1& v1, const Args&... args)
+{
+    return tinyformat::format (fmt.c_str(), v1, args...);
+}
+
+inline string_view format (string_view str) { return str; }  // base case
+
+
+/// Output formatted string to stdout, type-safe, and threads can't clobber
+/// one another.
+template<typename T1, typename... Args>
+void printf (string_view fmt, const T1& v1, const Args&... args)
+{
+    sync_output (stdout, format(fmt.c_str(), v1, args...));
+}
+
+inline void printf (string_view str) {  // base case
+    sync_output (stdout, str);
+}
+
+/// Output formatted string to an open FILE*, type-safe, and threads can't
+/// clobber one another.
+template<typename T1, typename... Args>
+void fprintf (FILE *file, string_view fmt, const T1& v1, const Args&... args)
+{
+    sync_output (file, format(fmt.c_str(), v1, args...));
+}
+
+inline void fprintf (FILE *file, string_view str) {  // base case
+    sync_output (file, str);
+}
+
+
 
 /// Return a std::string formatted from printf-like arguments.  Like the
 /// real sprintf, this is not guaranteed type-safe and is not extensible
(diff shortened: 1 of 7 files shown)


src/include/OpenImageIO/strutil.h
```diff
@@ -79,16 +80,54 @@ OIIO_NAMESPACE_BEGIN
...
-/// The printf argument list is fully typesafe via tinyformat; format
-/// conceptually has the signature
-///
-/// std::string Strutil::format (const char *fmt, ...);
-TINYFORMAT_WRAP_FORMAT (std::string, format, /**/,
-    std::ostringstream msg;, msg, return msg.str();)
+/// Uses the tinyformat library underneath, so it's fully type-safe, and
+/// works with any types that understand stream output via '<<'.
+template<typename T1, typename... Args>
+std::string format (string_view fmt, const T1& v1, const Args&... args)
```
fpsunflower: I think you want `Args&&` and then `std::forward(args)...` to pass them along to tiny format to avoid extra copies in some cases?

(it may not really matter here since all the things you'd want to `printf` are likely to be POD's though)

src/include/OpenImageIO/strutil.h
```diff
@@ -79,16 +80,54 @@ OIIO_NAMESPACE_BEGIN
...
+}
+
+inline void printf (string_view str) {  // base case
+    sync_output (stdout, str);
+}
+
+/// Output formatted string to an open FILE*, type-safe, and threads can't
+/// clobber one another.
+template<typename T1, typename... Args>
+void fprintf (FILE *file, const char* fmt, const T1& v1, const Args&... args)
```
fpsunflower: Why `string_view fmt` in `printf` vs `const char* fmt` in `fprintf` ?

I actually think `const char* fmt` is probably better since its probably not a good idea to use dynamic strings as the format string (the real C `printf` has compiler warnings when you try to do that).

src/include/OpenImageIO/strutil.h
lgritz: I just copied the basic methodology from tinyformat.h. But you're probably right about Args&&, I'll give that a try.

We actually do use non-PODs all the time, one of the nice things about these tinyformat-wrapped functions is that they will print any data type that has a << stream output operator.

src/include/OpenImageIO/strutil.h
lgritz: Wait, I think I take this back. I'm trying to use string_view wherever we can. The mistake here is that fprintf didn't use it (but format and printf did). Using string_view just lets us accept a wider variety of string types as input, and expresses the general principle that it's better to pass a pointer and length (that's what a string_view is) rather than a raw pointer only.

I guess I could go either way on this one, but I don't see any penalty to using string_view. That kind of usage is where things are going in C++ land in general. But I don't feel super strongly about it. Is there a principled reason why a `const char*`, but no length, is better here?

src/include/OpenImageIO/strutil.h
lgritz: Wait, remind me in what circumstances Args& would produce an extra copy but Args&& would not?

I understand why you would prefer && over pass-by-value, but pass-by-reference should never incur a copy.

src/include/OpenImageIO/strutil.h
fpsunflower: Yeah no strong feelings about it. The issue that `fmt` shouldn't be dynamically driven can't really be guarded against in C++ (even with the `const char*` API).

I'm ok with both being `string_view`.

src/include/OpenImageIO/strutil.h
fpsunflower: Actually I don't think this matters. I believe it means that if one of the arguments was a temporary `std::string` - `Arg&&` allows it to be "moved" in. It may not really matter here since we don't expect `format` to take ownership of anything it looks at. On the other hand if this function was going to retain one of its arguments (like adding them to a data structure), the move semantics are an important optimization.

Also -- I could be wrong about this too, but do you need the base case and the explicit `T1` ? I think it might be fine to just have: `format(fmt, const Args&... args)`  since you're not actually doing anything with `v1` other than pass it along like the others.

lgritz: Oh, I see what you were getting at with the base case... do I need it at all if I take away the T1? I'll try that and see what happens.

lgritz: Turns out the base case is needed after all, because the underlying Tinyformat template does not provide one. Their rationale is that you don't need the formatting if there are no arguments, but I kind of like the symmetry, and since our printf/fprintf also ensure that the contents are output atomically in a threaded environment, I think it's nice to be able to use them without arguments.

lgritz:
Updated. No change to rvalue refs or base case. But fixed to use string_view consistently, and also updated ustring::format() to match.

src/include/OpenImageIO/strutil.h
fpsunflower: `std::forward<Args>(args)...` should expand to `std::forward(a0),std::forward(a1),std::forward(a2),...`
vs.
`std::forward<Args>(args...)` which would expand to `std::forward(a0,a1,a2)` which is not what you want.

fpsunflower:
But as you said, if you need the base case anyway then its moot. LGTM!

src/include/OpenImageIO/strutil.h
lgritz: Oh, wow, I didn't get the subtlety of the placement of `...`

I'm still not sure it makes any difference in this case, since we aren't providing a "sink" for the arguments, we are just looking at the references.

PR #1581, opened 2016-12-11
Improvements to thread_pool and 2D parallelism

lgritz:
It took me a long time of working on this as a background task, but it's finally in good shape. This is a continuation of the thread pool & parallel loop work I merged a few weeks ago.

------

2D parallelism loops:

* Add parallel_for_2D varieties for easy thread_pool parallelization over 2D domains.
* Refactor ImageBufAlgo::parallel_image:
  - Use thread pool rather than spawning new threads.
  - Options struct includes max number of threads to use, domain subdivision hints, minimum number of pixels per thread task, and custom pool. Users should almost never need to adjust any of these.
* Refactor all the places using parallel_image to use the new style, and to use lambdas in a pleasing way rather than the prior idiom of recursing and handling the 1-thread and many-thread cases separately.
* More miscellaneous refactor of places where we used thread_group launch into use of parallel_for with pool.
* The resulting refactor of the IBA functions is pleasing to me, and the whole thing is much more idiomatic C++11.

thread pool improvements:

* New thread_pool method, this_thread_is_in_pool(), lets a caller find out if it (the calling thread) is a pool thread, so it may avoid submitting more jobs to the pool and clogging it up. A pool thread (which presumably is a task in a larger parallel work group) is probably better off doing any further subtasks itself rather than enqueuing them.  That also helps naive users from writing tasks that accidentally deadlock. This is implemented with a thread-specific pointer that lets the pool keep track of which threads participate. The various parallel_for functions employ this to avoid recursively splitting parallel tasks into more enqueued parallel tasks.
* Change the thread_pool to be able to switch (via #if) between a traditional mutexed queue and the boost::lockfree::queue. Interestingly, in my benchmarks on both OSX and Linux, the mutexted queue is slightly faster! So it remains the default for now.
* Allow environment variable OPENIMAGEIO_THREADS, if set, to override the hardware_concurrency (number of cores) as the default pool size and default OIIO threading count. This is primarily for debugging, so it's easy to force an artificially high or low thread count.

### DIFF
--- a/src/libutil/thread.cpp
+++ b/src/libutil/thread.cpp
@@ -103,8 +103,6 @@ private:
 
 OIIO_NAMESPACE_BEGIN
 
-bool debug_threads = false; // FIXME -- remove this after debugging
-
 static int
 threads_default ()
 {
--- a/src/libutil/thread.cpp
+++ b/src/libutil/thread.cpp
@@ -49,14 +49,71 @@
 #include <future>
 #include <memory>
 
-// Use boost::lockfree::queue for the task queue
-#include <boost/lockfree/queue.hpp>
-
 #include <OpenImageIO/sysutil.h>
 #include <OpenImageIO/thread.h>
+#include <OpenImageIO/sysutil.h>
+#include <OpenImageIO/strutil.h>
+
+#include <boost/thread/tss.hpp>
+
+#if 0
+
+// Use boost::lockfree::queue for the task queue
+#include <boost/lockfree/queue.hpp>
+template<typename T> using Queue = boost::lockfree::queue<T>;
+
+#else
+
+#include <queue>
+
+namespace {
+
+template <typename T>
+class Queue {
+public:
+    Queue (int size) { }
+    bool push(T const & value) {
+        std::unique_lock<Mutex> lock(this->mutex);
+        this->q.push(value);
+        return true;
+    }
+    // deletes the retrieved element, do not use for non integral types
+    bool pop(T & v) {
+        std::unique_lock<Mutex> lock(this->mutex);
+        if (this->q.empty())
+            return false;
+        v = this->q.front();
+        this->q.pop();
+        return true;
+    }
+    bool empty() {
+        std::unique_lock<Mutex> lock(this->mutex);
+        return this->q.empty();
+    }
+private:
+    typedef OIIO::spin_mutex Mutex;
+    std::queue<T> q;
+    Mutex mutex;
+};
+
+}
+
+#endif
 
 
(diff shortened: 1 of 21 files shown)


lgritz:
Anybody have comments on this one?

lgritz:
I amended this PR with another thread pool improvement:

thread_pool: improved logic for wanting no extra worker threads.

The thread pool size is the number of extra worker threads, not counting    the calling thread (which also participates in the work).

But if you truly want to run single-threaded, with no extra worker    threads, but still use the therad_pool/parallel_for interface, there    wasn't a correct way to do that, since resize(1) would give you one    worker thread + the caller, and resize(0) would use the default number    of cores.

So the change is that resize(-1) means "use hardware_concurrency-1"    and resize(0) means "no worker threads".

Now, when tasks are pushed onto the thread_pool, if there are no    workers, it will just run the task immediately on the calling    thread. This isn't quite as efficient as bypassing the pool entirely,    because it still needs to package up a lambda and make a future. But the    single-threaded case is not really the case we're optimizing for, and at    least this lets you have a uniform calling sequence (or use of    parallel_for) that doesn't break if you have some reason to not want any    extra worker therads at all. For any really expensive task, I don't think     you'll notice the overhead.

src/libutil/thread.cpp
```diff
@@ -49,15 +49,72 @@
...
 
 OIIO_NAMESPACE_BEGIN
 
+bool debug_threads = false; // FIXME -- remove this after debugging
+
+static int
+threads_default ()
+{
+    int n = Strutil::from_string<int>(Sysutil::getenv("OPENIMAGEIO_THREADS"));
+    if (n < 1)
```
cmstein: should this be (n<0) for maximum concurrency?

lgritz: There are two different concepts: How many threads we want in the app, and how many workers in the thread pool.

If you don't want to "overthread", you probably want the pool to be cores-1, because you already have one thread putting things into the work queue and performing some of the work itself.

Number of default threads is equal to number of cores. It doesn't make sense to have 0 threads, because something needs to run. So requesting anything < 1 means "use the default", which is number of cores (unless overridden by OPENIMAGEIO_THREADS, which I mainly use for debugging).

The *pool* size, however, is a different matter. There, it's reasonable to request 0, which means that the calling thread should do all the work, and there are no extra threads just watching the pool's work queue. The pool will default to threads_default-1.

Make sense?

cmstein: I think so -- just getting a little confused over which was which in which context. :-)

PR #1582, opened 2016-12-11
Fix exr_threads value -1 to disable thread pools not working.

brechtvl:
29766348b27a2030820147561ef47cf44835c595 never actually worked, because the -1 value was getting clamped.


lgritz:
Makes sense, LGTM.

lgritz:
Backported to RB-1.7

PR #1590, opened 2016-12-26
Add correct link for Windows build instructions

matekm:
This commit fix the link to Windows build instruction in INSTALL.md The correct link is https://sites.google.com/site/openimageio/building-oiio-on-windows


lgritz:
LGTM.

But really, I think it probably needs some updating. I get the impression that building on Windows is still tricky. Let me start a new thread on oiio-dev about this when I collect my thoughts.

PR #1591, opened 2016-12-28
Address sanitizer support, and resulting bug fixes

lgritz:
Add build option `SANITIZE=...` to trigger builds with various sanitizers.

Work through address sanitizer first -- fixes for four different latent bugs that it found. (See the individual commit message for explanations if interested, most are simple buffer overruns.)


lgritz:
Updated this PR with a few more fixes found by asan/lsan. (When run on Linux, address sanitizer also includes leak checks.) After fixing all of these little buglets, I get a clean enough testsuite to add a sanitizer run to the Travis test matrix. (Trick: I had to disable the Python tests when sanitizer is used, because the Python interpreter is not linked against libasan, so that generates errors when loading the PyOpenImageIO module that's compiled in sanitized mode.)

PR #1593, opened 2016-12-30
[WIP] Use Conan package manager to build OpenImageIO on Windows

robert-matusewicz:
Hi,

I was wondering how to "fix" OIIO Windows build and one of the option that came to my mind is to use Conan C/C++ package manager (https://conan.io/).

I gave it a go and managed to build static-linked libOpenImageIO with few basic plugins (ico, png, exr, jped and probably other).

To make it work I needed to install Conan on my system. After that I added a conanfile.txt that defines required dependencies and generator.

Then, in the OIIO_ROOT/build directory I run:

`conan install .. -s compiler="Visual Studio" -s compiler.version=14 -s arch=x86 -s build_type=Debug -s compiler.runtime=MDd`

and after that I run cmake:

`cmake -D USE_SIMD=0 -D LINKSTATIC=1 -D USE_JPEGTURBO=1 ../`

it build cleanly, but I didn't manage to test it yet (need some additional  work on it).

I would say that for the Windows build it would be very good option as it easy to integrate with CMake and (hopefully) will allow to build OIIO also with clang on windows.

Obviously, there are some problems, like not all libraries being discovered by cmake due to non-standard library namings, but that's should be easy to solve either by updating conan package or having our own find* function for affected libraries.

Other question is how to maintain packages, as not all of them are found on conan.io (like boost 1.53 as example).

I'm willing to invest some time to make Windows build painless, but first I wanted to discuss possible options with you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,11 @@ if (NOT CMAKE_VERSION VERSION_LESS 3.2.2)
 endif ()
 set (CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS TRUE)
 
+if(USE_CONAN)
+    include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+    conan_basic_setup()
+endif()
+
 # N.B./FIXME - after CMake 3.0 is our minimum, we should change many of the
 # add_definitions to add_compile_options. In newer cmake versions, the
 # former is designated for definitions (-Dblah) and the latter is for
@@ -218,6 +223,7 @@ set (EXTRA_DSO_LINK_ARGS "" CACHE STRING "Extra command line definitions when bu
 set (USE_SIMD "" CACHE STRING "Use SIMD directives (0, sse2, sse3, ssse3, sse4.1, sse4.2, avx, avx2, avx512f, f16c)")
 set (USE_CCACHE ON CACHE BOOL "Use ccache if found")
 set (CODECOV OFF CACHE BOOL "Build code coverage tests")
+set (USE_CONAN OFF CACHE BOOL "Use Conan package manager to download build dependencies")
 
 # Use ccache if found
 find_program (CCACHE_FOUND ccache)
@@ -325,6 +331,15 @@ if (BUILDSTATIC)
     endif ()
 endif()
 
+if(WIN32)
+    ADD_DEFINITIONS(/bigobj)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+    SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /bigobj")
+    SET(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} /bigobj")
+    SET(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /bigobj")
+    SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} /bigobj")
+endif()
+
 # Use .a files if LINKSTATIC is enabled
 if (LINKSTATIC)
     set (_orig_link_suffixes ${CMAKE_FIND_LIBRARY_SUFFIXES})
@@ -340,8 +355,10 @@ if (LINKSTATIC)
 endif ()
 
 set (CMAKE_MODULE_PATH
+     "${CMAKE_MODULE_PATH}"
      "${PROJECT_SOURCE_DIR}/src/cmake/modules"
-     "${PROJECT_SOURCE_DIR}/src/cmake")
+     "${PROJECT_SOURCE_DIR}/src/cmake"
+     )
 
 include (util_macros)
 include (oiio_macros)
--- a/conanfile.txt
+++ b/conanfile.txt
@@ -1,0 +1,15 @@
+[requires]
+zlib/1.2.8@lasote/stable
+libtiff/4.0.6@bilke/stable
+libpng/1.6.23@lasote/stable
+OpenEXR/2.2.0@Mikayex/stable
+Boost/1.60.0@lasote/stable
+libjpeg-turbo/1.5.1@lasote/ci
+OpenSSL/1.0.2j@lasote/stable
+giflib/5.1.3@lasote/stable
+
+[generators]
+cmake
+
+[options]
+libpng:shared=False
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -54,7 +54,6 @@ if (THIRD_PARTY_TOOLS_HOME AND EXISTS "${THIRD_PARTY_TOOLS_HOME}")
     endforeach ()
 endif ()
 
-
 setup_string (SPECIAL_COMPILE_FLAGS ""
                "Custom compilation flags")
 if (SPECIAL_COMPILE_FLAGS)
@@ -82,7 +81,6 @@ include_directories (${ZLIB_INCLUDE_DIR})
 ###########################################################################
 # PNG
 find_package (PNG REQUIRED)
-
 
 ###########################################################################
 # IlmBase & OpenEXR setup
@@ -410,8 +408,11 @@ endif ()
 
 ###########################################################################
 # JPEG
-
 if (USE_JPEGTURBO)
+    if(USE_CONAN)
+        set(JPEG_INCLUDE_DIR ${CONAN_INCLUDE_DIRS_LIBJPEG})
+        set(JPEG_LIBRARY ${CONAN_LIB_DIRS_LIBJPEG-TURBO}/turbojpeg-static.lib)
+    endif()
     find_package (JPEGTurbo)
 endif ()
 if (JPEG_FOUND)


lgritz:
I like this, but before deciding on this route, let's discuss the alternatives. I'm going to start a thread on oiio-dev because of the wider readership than we will get for comments on a PR.

KindDragon:
JFYI You can try install it from [vcpkg](https://github.com/Microsoft/vcpkg)
`vcpkg install openimageio`

lgritz:
Awesome!

@KindDragon, are you the author of this?

I was glancing at the portfile https://github.com/Microsoft/vcpkg/blob/master/ports/openimageio/portfile.cmake and had a couple questions:

1. Is there a reason that it defines BUILD_TOOLS=OFF ? I would think that people who want to install it very likely will want the command line tools (in particular, oiiotool, maketx, etc.) as well as the libraries. And also setting USE_PYTHON=0 means that the python bindings will not be available, which are widely used.

2. It seems that almost all of the optional dependent libraries have their formats disabled. But looking over the available ports (https://github.com/Microsoft/vcpkg/tree/master/ports), many of them are already available as vcpkg ports -- giflib, ffmpeg, freetype, libraw, python3, openjpeg? Turning them all off disables a lot of formats and functionality. But presumably all the hard work has been done and you could just add those as dependencies are rely on the vcpkg ports, right?

Also, I'm sorry to bother you for this, but I don't even have a Windows machine, so it would be hard for me to test this myself, but would it be possible to give an example of the minimal series of commands by which a Windows user who needs OIIO could install vcpkg and then use it to get OIIO (and its dependencies) built? I'd love to get that in our install instructions, and know that it's right, and finally have one simple, canonical answer for people who want to build it on Windows.

lgritz:
@matekm, how do you feel about the vcpkg approach? Is there still a reason to try to revive this PR (I think there are merge conflicts) and complete it? Or is vcpkg the way to go here for our canonical simple Windows build approach?

lgritz:
@KindDragon, one more question: I noticed the vcpkg portfile specifies a particular version of OIIO. How does this get updated? I mean, by whom? Are you the "owner" of the port and you'll take the initiative to keep up with it? Or should I do this, meaning that with each release of OIIO, I should also submit a PR to vcpkg that updates the version it references?

KindDragon:
> @KindDragon, are you the author of this?

Yes

> Is there a reason that it defines BUILD_TOOLS=OFF ?

I just not need them for my usage, we can easily add them like that https://github.com/Microsoft/vcpkg/blob/17df9be71f23f2e5ac1c792f7220065ff940fc42/ports/grpc/portfile.cmake#L59-L65

> But presumably all the hard work has been done and you could just add those as dependencies are rely on the vcpkg ports, right?

Yes, we can do that but I don't want to enable by default big dependencies like Qt. We can enable them after vcpkg support optional dependencies https://github.com/Microsoft/vcpkg/issues/249
It's very easy to add them generally, but it just increases compilation time.

> would it be possible to give an example of the minimal series of commands by which a Windows user who needs OIIO could install vcpkg and then use it to get OIIO (and its dependencies) built?

* Clone vcpkg repository https://github.com/Microsoft/vcpkg
* Run
```
C:\src\vcpkg> .\bootstrap-vcpkg.bat
```
* Run in console: `vcpkg install openimageio`

> I noticed the vcpkg portfile specifies a particular version of OIIO. How does this get updated? I mean, by whom? Are you the "owner" of the port and you'll take the initiative to keep up with it? Or should I do this, meaning that with each release of OIIO, I should also submit a PR to vcpkg that updates the version it references?

I do not directly use the OIIO library. I create port because it required by TheiaSfm library. I will update it if Theia requires a newer version. But it easy to anyone update (just to [2 files](https://github.com/Microsoft/vcpkg/tree/master/ports/openimageio))

lgritz:
> > Is there a reason that it defines BUILD_TOOLS=OFF ?

> I just not need them for my usage, we can easily add them like that https://github.com/Microsoft/vcpkg/blob/17df9be71f23f2e5ac1c792f7220065ff940fc42/ports/grpc/portfile.cmake#L59-L65

Since I don't have an easy way to test this, would you kindly do us the favor of making this change, so that the tools are built and installed? And also enabling the dependencies by default of giflib, ffmpeg, freetype, libraw, python3, openjpeg? I think that would get the default build of OIIO to have all the format and binding support that people expect, while still excluding the ones like Qt that have a very poor build-time-to-usefulness ratio?

> Clone vcpkg repository https://github.com/Microsoft/vcpkg
> Run
> ...

Where does this result in the libraries and tools being installed? I mean, what paths would I then need to add to see the tools or build against the libraries?

> I do not directly use the OIIO library. I create port because it required by TheiaSfm library. I will update it if Theia requires a newer version. But it easy to anyone update (just to 2 files)

I am perfectly comfortable taking on the task of updating the version number when we cut new releases. I can just make it part of our release process.

Thanks for all the work on this. It's a big step forward for our project.

ras0219-msft:
>Where does this result in the libraries and tools being installed? I mean, what paths would I then need to add to see the tools or build against the libraries?

@lgritz For the full layout story, see https://vcpkg.readthedocs.io/en/latest/users/integration/.

If you use MSBuild (the default Visual Studio buildsystem), we suggest running one more command (`vcpkg integrate install`), after which you're able to directly `#include` the headers and use the library. The library files will be automatically linked, so you can directly copy-paste examples into VS, hit F5, and they should run :). This will also automatically handle DLL deployment if you're using shared libraries.

If you use CMake, then you just need to add one extra flag during your configure (`-DCMAKE_TOOLCHAIN_FILE`), then `find_package()/find_library()/find_path()` work exactly as you'd expect.

> I am perfectly comfortable taking on the task of updating the version number when we cut new releases. I can just make it part of our release process.

This would be great! Vcpkg doesn't currently have a formal maintainer role -- we accept contributions from anyone willing to provide them. As a project maintainer we'd be happy for any observations or changes you'd be willing to share :).

> > >Is there a reason that it defines BUILD_TOOLS=OFF ?
> >
> >I just not need them for my usage, we can easily add them like that https://github.com/Microsoft/vcpkg/blob/17df9be71f23f2e5ac1c792f7220065ff940fc42/ports/grpc/portfile.cmake#L59-L65
>
>Since I don't have an easy way to test this, would you kindly do us the favor of making this change, so that the tools are built and installed?

One thing to keep in mind with Vcpkg is that we're focused on delivering development libraries, not applications. We do support build-time tools that are coupled to a library runtime (e.g. protobuf's compiler and flex/lex/bison/yacc), but for more general purpose applications like curl, ssh, or imagemagick we strongly recommend against using vcpkg as a delivery vehicle. Services like Chocolatey[1] or Centennial[2] are much better for those purposes.

I don't know the circumstances of OIIO, so please take the above as context and let us know how we should proceed :).

[1] https://chocolatey.org/
[2] https://docs.microsoft.com/en-us/windows/uwp/porting/desktop-to-uwp-run-desktop-app-converter

lgritz:
@ras0219-msft :

>> I am perfectly comfortable taking on the task of updating the version number when we cut new releases. I can just make it part of our release process.

> This would be great! Vcpkg doesn't currently have a formal maintainer role -- we accept contributions from anyone willing to provide them. As a project maintainer we'd be happy for any observations or changes you'd be willing to share :).

Yes, I'm very happy to take on that task, I'll submit a vcpkg PR every time OIIO has a major release, to update the version it uses.

> One thing to keep in mind with Vcpkg is that we're focused on delivering development libraries, not applications. We do support build-time tools that are coupled to a library runtime (e.g. protobuf's compiler and flex/lex/bison/yacc), but for more general purpose applications like curl, ssh, or imagemagick we strongly recommend against using vcpkg as a delivery vehicle. Services like Chocolatey[1] or Centennial[2] are much better for those purposes.
> 
> I don't know the circumstances of OIIO, so please take the above as context and let us know how we should proceed :).

I don't want to step on vcpkg's toes or violate its conventions, but 95% of OIIO's build time is making the libraries (including the Python bindings) that developers and other packages need. So if we also just let it build the tiny sliver that adds a couple of command line tools, it could be our one-stop simple instructions for Windows users that we have long desired. It would be an ongoing agony to have to constantly explain to users who have gone through the vcpkg setup why they didn't get all the usual components and that they have to start again with a second package manager in the mix. (Also so much harder for me, not having a Windows box or expertise to do this right maintaining multiple package managers.) 

If it helps, I think I can honestly say that there are other packages that may now, or someday, be in vcpkg, that expect the oiio command line tools available, either to use directly in the process of their own builds, or to be around to execute by whatever library or thing they are building.

Another possible compromise -- is it possible/easy for the `vcpkg install` command itself to change any of the underlying options used when vcpkg does the build? For example, I'd be ok with the default vcpkg build of oiio to only build the libraries, provided there were simple instructions we could give that would tell people how to use vcpkg to trigger building the entire package including the command line utilities.

And, sorry, one more question if you would be so kind: does vcpkg have, or plan on having, any kind of build caching (for example, like Homebrew has) so that the thousands of people who need a package don't each have to build redundantly? For an individual user it hardly matters, but I'm thinking of cases like projects using Appveyor, where it could save a tremendous amount of time and compute power on a limited-resources service, if each build (and dependencies!) didn't need to start from scratch.

ras0219-msft:
The avoidance towards applications isn't really due to build time, it's to avoid
1. Windows locks all DLLs currently in use by applications. This means that if your text editor was built inside Vcpkg, you wouldn't be able to upgrade anything while it's running.
2. Unneeded dependencies and breakages. If there is an API change in a dependency that prevents some unnecessary utility from building (but the library itself is fine), it would be terrible to prevent your project from using the core library.
Another example of this would be if your project needs an _older_ version of some library that isn't supported by the newest version of your text editor. There's no need for those versions to be coupled -- the editor could use the newer version and your project could use the older version -- which means it would ideally be distributed outside that Vcpkg instance.

I do appreciate that this is more difficult from your perspective as a maintainer without consistent Windows access, but I want to express how our user experience will break down in these cases (which you probably don't want for your users either :)).

_See below for a proposed workflow that might work for you_

> ... either to use directly in the process of their own builds ...

This is the exact case where we do support utilities. Any executables that a consumer would use during their build (that are tied to the library version used in your final project's exe) need to be packaged. If your utilities fall into this bucket, then we should build them as part of the default package.

>is it possible/easy for the vcpkg install command itself to change any of the underlying options used when vcpkg does the build?

We are actively working on enabling this via "Feature Packages" (https://github.com/Microsoft/vcpkg/pull/1205). This would enable users to enable or disable dependencies or components at a very fine-grained level with a correct corresponding change to the overall dependency tree.

> And, sorry, one more question if you would be so kind: does vcpkg have, or plan on having, any kind of build caching (for example, like Homebrew has) so that the thousands of people who need a package don't each have to build redundantly? For an individual user it hardly matters, but I'm thinking of cases like projects using Appveyor, where it could save a tremendous amount of time and compute power on a limited-resources service, if each build (and dependencies!) didn't need to start from scratch.

Our current solution to this problem is the `export` command [1]. Essentially, this allows users to build a shrink-wrapped dependency archive from Vcpkg which serves as a completely transparent replacement for a full Vcpkg enlistment. This means that if your project builds on your local computer using Vcpkg, you can just run
```
vcpkg export --7zip boost libjpeg-turbo tiff libpng openexr
```
and you'll get a minimal 7zipped dependency bundle that you could host on GitHub and download onto AppVeyor. It contains the exact same CMake toolchain as a full enlistment.

We _are_ looking into what it would take to support a happy-path of prebuilt binaries (i.e. only for certain triplets with certain features and certain versions), but there are many challenges that make us think the above `export` command is generally a better choice!

_Proposed workflow_
For every major release, use the latest Vcpkg to perform one full clean build of all your dependencies and the project itself. Then, upload to GitHub attached to the release:
- Prebuilt win32-x86 tools (static linked? Should be as easy as `-DVCPKG_TARGET_TRIPLET=x86-windows-static`) in a zip file for users to quickly download and use if they just want the converter tools/image viewer/etc (If the cost/benefit looks worthwhile to you, you could upload these to Chocolatey).
- The results of `vcpkg export --7zip boost libjpeg-turbo tiff libpng openexr`, which will be pulled by your appveyor CI and save tons of time on your per-commit builds.

This does mean that you update at your own pace -- as slowly or quickly as you want. If there's a build failure over in our CI system where a dependency update breaks your latest vcpkg release, we'll open an issue over here to let you know.

[1] http://vcpkg.readthedocs.io/en/latest/users/integration/#export-command

ras0219-msft:
As an update, we now have feature packages in Vcpkg which means we can easily handle additional installation options like:
```
>vcpkg install openimageio[tools] --featurepackages
```
(note: these are examples and aren't currently implemented)

lgritz:
That's awesome, thanks! I had this on my list to follow up on. This will be a big help for people, once we make it easy to get the tools as well.

lgritz:
I'm going to close this poor old forgotten PR. But don't worry, I'm actually fooling around a bit with Conan myself and if it looks useful, I will submit an updated PR to enable easy Conan builds.

PR #1594, opened 2016-12-30
[Windows] Fix std::max error in imagebufalgo_util.h

matekm:
Visual Studio Professional 2015 can't properly deduce template type for a std::max. This patch directly specify the type in the std::max call.


lgritz:
LGTM, thanks for the fix.

PR #1595, opened 2017-01-03
Bug fix in ImageInput::read_tiles

MrKepzie:
ImageInput::read_tiles: prefix_bytes must use the user format and not the spec format since the temporary buffer is also using pixels in the user format.

Here is a sample image to test it. 

[test_tiled.exr.zip](https://github.com/OpenImageIO/oiio/files/682319/test_tiled.exr.zip)

I don't have code to share, but can guarantee it has been tested and fixes the issue.

### DIFF
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -456,7 +456,7 @@ ImageInput::read_tiles (int xbegin, int xend, int ybegin, int yend,
     stride_t full_tilewidthbytes = full_pixelsize * m_spec.tile_width;
     stride_t full_tilewhbytes = full_tilewidthbytes * m_spec.tile_height;
     stride_t full_tilebytes = full_tilewhbytes * m_spec.tile_depth;
-    size_t prefix_bytes = m_spec.pixel_bytes (0,chbegin,true);
+    size_t prefix_bytes = chbegin * format.size();
     std::vector<char> buf;
     for (int z = zbegin;  z < zend;  z += std::max(1,m_spec.tile_depth)) {
         int zd = std::min (zend-z, m_spec.tile_depth);
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -456,7 +456,8 @@ ImageInput::read_tiles (int xbegin, int xend, int ybegin, int yend,
     stride_t full_tilewidthbytes = full_pixelsize * m_spec.tile_width;
     stride_t full_tilewhbytes = full_tilewidthbytes * m_spec.tile_height;
     stride_t full_tilebytes = full_tilewhbytes * m_spec.tile_depth;
-    size_t prefix_bytes = chbegin * format.size();
+    size_t prefix_bytes = native_data ? m_spec.pixel_bytes (0,chbegin,true)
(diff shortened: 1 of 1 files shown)


lgritz:
I think maybe that's not quite right, either.

I think that the original one computed native size only (when `format` is passed as `UNKNOWN`), and your solution computes a fixed buffer type only (a concrete `format`). But really, we want it to work properly either way.

How about:

    size_t prefix_bytes = m_spec.pixel_bytes (0, chbegin, native_data);

??

Also, a few lines earlier,

    stride_t pixelsize = native_data ? native_pixel_bytes 
                                     : (format.size() * nchans);
    stride_t full_pixelsize = native_data ? m_spec.pixel_bytes(true)
                                          : (format.size() * m_spec.nchannels);

is correct, but might perhaps be more succinct as:

    stride_t pixelsize = m_spec.pixel_bytes (chbegin, chend, native_data);
    stride_t full_pixelsize = m_spec.pixel_bytes (native_data);

MrKepzie:
Actually the correct solution is 
    size_t prefix_bytes = native_data ? m_spec.pixel_bytes (0,chbegin,true)
                                     : chbegin * format.size();

The 2 other suggestions (pixelSize and full_pixelsize) do not work, even if native_data is false the wrong format size is used leading to buffer overflow and crash later on. I left them as it was.

MrKepzie:
Btw any plan on making read_tile accepting _chbegin_ and _chend_ ? It is currently impossible to read a subset of channels using the read_tile/read_tiles API, the only way is to call read_native_tiles, but this forces the external application to reproduce almost all the code of read_tiles and read_tile to handle all pixel formats conversion etc...

lgritz:
Oh, yes, I see! I was incorrect before, mixing up the m_spec.format with the `format` that describes the buffer (what to convert TO).

lgritz:
@MrKepzie , I'm not sure what you're talking about, `read_tiles()` certainly does allow a tile range. Is it possible you are thinking of a very old version?

You should never call the `read_native_*` functions from an app -- those are just the underlying application.

MrKepzie:
Yes, I know it takes a chbegin and chend in parameter but it only honours it in the case of native and contiguous data... for all other cases it allocates a temporary buffer that can contain all possible channels and calls read_tile (without chbegin, chend) resulting in the internal openexr input having to decode all channels of the tile. 

For a file with e.g 32 channels it will allocate a temporary buffer bigger than necessary and ask OpenEXR to process data that will then be just discarded. The worse is that these channels are even converted to the requested format in (read_tile)[https://github.com/OpenImageIO/oiio/blob/master/src/libOpenImageIO/imageinput.cpp#L491] even if they are going to be thrown away in the (copy_image)[https://github.com/OpenImageIO/oiio/blob/master/src/libOpenImageIO/imageinput.cpp#L495] call below.

lgritz:
Sorry for the delay. This looks right to me, I will merge it.

lgritz:

MrKepzie:
Thank you!
Should I open a feature request for the thing I was mentioning in the last message ?

lgritz:
Yeah, probably, so we don't forget it.

I see your point about that. It's a case I didn't make as smart as it should be, just falls to the general case.

PR #1599, opened 2017-01-06
maketx did not incorporate "highlight compensation" into the hash!

lgritz:
The results when comparing "maketx --hicomp" versus without that flag
were especially hilarious when using 'iv' to view the images -- for high
resolution images, it will fall back on ImageCache, which thinks two
images with the same hash are identical and don't need to be redundantly
read from disk. So, the hicomp and nohicomp images looked identical in
iv, leading to even more confusion.

So, also turn off the ImageCache de-duplication for iv so we don't
get bitten by this again. I don't think there is a legit case where we
rely on deduplication for iv, so this seems like the safe choice.

### DIFF
--- a/src/doc/oiiotool.tex
+++ b/src/doc/oiiotool.tex
@@ -1139,7 +1139,7 @@ appended arguments for {\cf -otex} and {\cf -oenv} also include:
     color conversions, and re-premultiply after. (default: 0) \\[0.5ex]
 & {\cf\small incolorspace=}\emph{string} & Specify color space conversion. \\
 & {\cf\small outcolorspace=}\emph{string} & ... \\[0.5ex]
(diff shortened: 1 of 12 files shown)


fpsunflower:
I'm not sure I follow -- does the code only hash the mip 0 pixels on the assumption that its sufficient to predict if the downstream mips will be the same?

lgritz:
Right. It needs to write the header information, including the hash, as it writes out the top level. So at that point it hasn't actually computed the later levels.

lgritz:
Yeah, when I make a new image file format (as if), remind me that a key design feature will be to be able to amend header fields of an existing file after the pixels are written, without needing to rewrite the whole thing.

lgritz:
Added a test to testsuite to verify this.

PR #1604, opened 2017-01-16
Cmake refactor and modernize cmake usage

lgritz:
Break our massive main CMakeLists.txt into a few included files, move
there all the "generic" stuff that we might want to use to populate
any new projects.

There's not much new logic here, it's just splitting an existing file, and
a bunch of renames to make some of the logic less OIIO-specific. (For
example, many places have changed from directly referencing "OpenImageIO"
to "${CMAKE_PROJECT_NAME}", which makes it directly cut-and-pastable
into other projects.)

Inspired by a weekend project (not work related) where I wanted to reuse a lot
of my cmake infrastructure.


lgritz:
Any comments? Last chance. 
It all seems to work for me on all platforms I've tried. You shouldn't notice any changes to the software itself, it's just a refactor of the build scripts.

lgritz:
OK, here we go. Let me know if anything breaks.

Issue #1603, opened 2017-01-14
Does OpenImageIO support nChannel images


StefanWoe:
Hi, just a question
can images (tiff, PSD and others) with more than 4 Channels be read into a memory buffer and written back using OpenImageIO? In the printing industry images with CMYK plus a number of spot color channels are quite common. Photoshop files must only be read, but TIFF must be written.

Is also BigTIFF supported?

Regards

Stefan

lgritz:
Yes and yes.

StefanWoe:
Thanks for your answer. In the meanwhile i found in the Documentation the following statement:

> Color space is grayscale or RGB. Non-spectral data, such as XYZ, CMYK, or YUV, are
converted to RGB upon reading.
(1.2 Simplifying Assumptions)

Doesnt that say that that even CMYK can not be handled?
Or is the Documentation incorrect?

lgritz:
Ah, interesting point I hadn't considered. OIIO is fine with an arbitrary number of channels generally, but by default will try convert a CMYK representation to RGB.

Looking at the code, it seems that the PSD reader has a way to signal it to skip the CMYK->RGB conversion (by setting a special 'configuration' attribute called "psd:RawData" to be nonzero). But there does not appear to be a corresponding way to cause the TIFF reader to not do the transformation.

That seems like an oversight. Let me see if I can quickly propose a patch to address this. Stay tuned.

StefanWoe:
Hi Igritz. That would be great help. Thanks!

lgritz:
Proposed solution in #1605, when that gets approved, I'll merge it into master as well as the RB-1.7 release branch.

StefanWoe:
Will this also work for multichannel (6, 8 or more channels?
Do you have a releaseplan (rough date) for RB_1.7?

lgritz:
It has always worked fine for an arbitrary number of channels; the issue here was only that when the file was marked as PHOTOMETRIC_CMYK, it was auto-converting to RGB (and forcing 3 colors) without a way to say that you didn't want that conversion.

RB-1.7 is the *branch* that is the tip of 1.7 release development. You can always check out that branch tag and get the latest (well, once this patch is approved and I merge it). Generally, on the first of each month, I make an official release by adding a tag to the then-tip of RB-1.7, so that would be called "Release-1.7.11", for example.

[...]

PR #1605, opened 2017-01-18
Read TIFF CMYK files as raw pixel values

lgritz:
As pointed out in #1603, the general principle that non-spectral color spaces (like CMYK) get converted to RGB upon being read by OIIO makes it quite difficult to directly read the CMYK values.

The default is (per OIIO conventions) to convert CMYK files to RGB upon input. But we now establish a new 'configuration' setting, wherein the input configuration attribute "oiio:RawColor" set to nonzero will be a     request to return raw pixel values, regardless of their apparent color     space, with no default conversion to RGB.

This should be honored by all format readers that allow non-spectral-like color     spaces. The other obvious one that comes to mind is PSD, which already     honored an undocumented "psd:RawColor", which this patch also alters     to accept the more generic "oiio:RawColor" and documents its behavior     for both PSD and TIFF.

For those of you using oiiotool, the way you force input of raw CMYK     values is like this:

        oiiotool cmyk_input.tif -o rgb.exr
        oiiotool --iconfig oiio:RawColor 1 cmyk_input.tif -o cmyk.exr

You'll see that the default behavior (first example) will result in     a 3-channel RGB image output to rgb.exr, whereas the second example     where we set the input configuration will output a 4-channel image    with channel names "C", "M", "Y", and "K". You're on your own for how    to display that properly, but it will be the original raw CMYK values.

---

In the process of chasing this down, I found a minor bug with `oiiotool --iconfig` wherein It worked for images read from the ImageCache, but for  images read directly, bypassing the cache, the configuration was  not properly propagated to the open call.


lgritz:
Fixes #1603

lgritz:
Merged into master and RB-1.7. The next tagged release of 1.7 will be on 1 Feb.

PR #1609, opened 2017-01-20
Fix declare_imageio_format declaration.

marsupial:
Declaration of declare_imageio_format currently doesn't match it's definition.


lgritz:
Well that's embarrassing!

LGTM, thanks for the spotting this.

PR #1610, opened 2017-01-23
Add xxhash::Hash class for hashing non contiguous data.

marsupial:
Exposes a nicer interface for xxhash that clients can use.

### DIFF
--- a/src/include/OpenImageIO/hash.h
+++ b/src/include/OpenImageIO/hash.h
@@ -45,6 +45,7 @@
 #include <string.h>   // for memcpy and memset
 #include <utility>
 #include <unordered_map>
+#include <type_traits>
 
 #include "export.h"
 #include "oiioversion.h"
@@ -56,6 +57,8 @@
 
 OIIO_NAMESPACE_BEGIN
 
+class ustring;
+class string_view;
 using std::unordered_map;
 using std::hash;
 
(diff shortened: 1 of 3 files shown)


lgritz:
This is not passing the CI builds.

marsupial:
I dun fixed it.

src/include/OpenImageIO/export.h
```diff
@@ -79,6 +79,7 @@
     #define OIIO_EXPORT
   #endif
   #define OIIO_LOCAL
+  #define OIIO_INLINE  __forceinline
```
lgritz: In platform.h, there's already a macro for OIIO_FORCEINLINE.

If you don't absolutely need the strict "force", then it's ok to just use the standard "inline" keyword. I don't think there are any C++11 compliant compilers that don't accept it.

lgritz:
I think the second commit is unnecessary. Is the usual "inline" keyword not sufficient? If there's truly a need to force inline behavior, then our platform.h already has an OIIO_FORCEINLINE macro.

Other than that, this LGTM. Seems like a reasonable convenience addition for those using xxhash, at no cost for those who don't use it. But can you say how you're using this or what inspired you to want to add it here?

marsupial:
Oops, didn't see the OIIO_FORCEINLINE. Assumed it would be in exports.h.

Sadly it's required for Windows otherwise it wants to link to the functions.
Possibly not needed on the template methods, but didn't want to take any chances.

Working on a caching feature for OSL and am using it to generate hashes for the shaders.
_Was_ might be more accurate, seems to be a dead end, but the work was already done.

marsupial:
Should be better now.

src/libutil/xxhash.cpp
```diff
@@ -944,5 +945,49 @@ unsigned long long XXH64_digest (const XXH64_state_t* state_in)
...
+      XXH ## SZ ## _freeState(HSTATE(SZ)); \
+  } \
+  template <> void Hash<TYPE>::append (const void *val, size_t N) { \
+      m_failed = (XXH ## SZ ## _update(HSTATE(SZ), val, N) != XXH_OK) || m_failed; \
+  } \
+  template <>  TYPE Hash<TYPE>::value () const { \
+      return XXH ## SZ ## _digest(HSTATE(SZ)); \
+  } \
+  template <> void Hash<TYPE>::append (const ustring &str) { \
+      append(str.hash()); \
```
lgritz: I'm thinking that maybe you should drop the `ustring` specialization and just let it use string_view to catch ustrings. I think it could be helpful to ensure that adding a sequence of characters will yield the same hash behavior regardless of whether it's a std::string, C++ string, or ustring. If ustring gives a different hash than the others (because you're directly using the ustring table's internal hash, rather than the xxhash as would be used for other character sequences), it would be different and I can imagine this leading to subtle bugs.

src/include/OpenImageIO/hash.h
```diff
@@ -81,6 +84,82 @@ inline size_t xxhash (const Str& s, size_t seed=1771) {
...
+    Hash (T seed = kSeed);
+    ~Hash ();
+
+    void reset (T seed);
+    void append (const void* val, size_t N);
+    T value () const;
+
+    /// Hash known string objects
+    void append (const ustring &str);
+    void append (const string_view &str);
```
lgritz: I think append(string_view) is the only string one you need. Both std::string and char* will properly convert to string_view, that's string_view's whole point.

Also note that you want to pass string_view by value, not by reference. There's no advantage to passing string_view by reference (`string_view&`), and it will just incur the extra instructions of a double indirection.

src/include/OpenImageIO/hash.h
```diff
@@ -81,6 +84,82 @@ inline size_t xxhash (const Str& s, size_t seed=1771) {
     return xxhash (s.data(), s.length(), seed);
 }
 
+template <typename T>
+class OIIO_API Hash {
```
lgritz: It would be great to have a comment here explaining the purpose and use of this class.

lgritz:
Sorry, I took another look and had some additional comments.

I like the idea of this, but if your one and only intended use for it has fallen through, I'm also ok if you want to suspend or abandon this PR for now (especially if you don't care to fix it up), and we can always pick up where we left off if another use for it comes up later.

Also... if we're going to go down the road of having a templated stateful hasher, it may be worth considering whether it can be rewritten in such a way that the hash scheme itself is part of the template specification, so that you could adapt it to any hashing scheme rather than have it be only specific to xxhash. After all, from time to time, our preferred hashing function changes. Would be nice to have a single interface.

marsupial:
I'm fine cleaning it up as it may still be of use. I agree adapting it to be more general in regards to the hash function is good, but its a start in that direction.

The only thought I differ on is to keep current ustring behavior. Its quite an optimization if you are hashing a container of them. I'm fine adding a comment stating it's behavior though! There's already a test that ustring does not hash to the same as a string.

lgritz:
Yes, I see your point about ustring. It's a tradeoff: speed if you are hashing a whole bunch of big strings, but at the cost of the possibly confusing behavior of a ustring and std::string with the same characters having different hashes.

I can see two solutions:

1. As you describe, keep the ustring specialization and just comment it with this caveat.

2. Leave `append(string_view)` ONLY, this avoiding the subtle mistake that people can make, and add a separate `append_ustring(ustring)` for savvy users to directly append a ustring explicitly with the speed gain and knowing that it's a different hash than any other string they may pass.

I'm leaning toward the second choice myself, but am happy to hear counter-arguments.

Also, I'm ok with the idea of leaving this xxhash-specific for now, if it's a pain to generalize. Then at a later time (and/or when it's needed) we can always return to more completely generalize it to handle any hash, as part of the template parameter set.

lgritz:
Amendment: I really prefer solution 2 above. I just think that it's very common to convert from ustring to char* or std::string, and if two code paths run the same characters through the hasher, it should return the same results. The bugs from this misunderstanding will be subtle and frustrating. I think we can retain all the speed advantages and none of the potential errors if we have an explicit ustring-specific method for appending.

marsupial:
I can try 2 but it seems a fair bit more code will be necessary to get a the STL interfaces working with two schemes.

lgritz:
Oh, I see your point about the STL interfaces. Hmmm. 

I get why you want a stateful incremental hasher that works on lots of complex data types.
Was there a particular reason you chose to base it on xxhash? Because if you use farmhash, it will match what ustring does.

So one possibility it to move it out of the xxhash namespace, and not promise what hash it uses (the header and public API don't say anyway). Then in the implementation of Hasher inside hash.cpp (that's probably where it should have gone anyway, since it's a utility rather than part of the original open source xxhash code), and also in ustring.cpp, put a comment that says that if either one is changed, so should the other, so that the Hasher and ustring hashes continue to match.

Or, you could just shelve the whole thing until it's really needed.

lgritz:
I'm... going to close this. Obviously the need has never materialized for this, and there seemed to still be some loose ends about how it should be designed. So let's just punt and abandon it until there's a real need.

PR #1613, opened 2017-01-27
oiiotool.cpp: fix action_crop to actually take into account the allsubimages option

e2lka:
### Problem
action_crop function in oiiotool wasn't working working with multiple subimages as ot.allsubimages wasn't taken into account.

### Solution
To fix that I just convert ot.allsubimages into a string (so the boolean will give "0" or "1") and options["allsubimages"] becomes valid to use.

### Suggestion
I could also have use directly ot.allsubimages directly, as it is into the other functions, but it feels that this change is the one that less touch the logic of the function.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -2874,7 +2874,7 @@ action_crop (int argc, const char *argv[])
     string_view size    = ot.express (argv[1]);
 
     std::map<std::string,std::string> options;
-    options["allsubimages"] = ot.allsubimages;
+    options["allsubimages"] = std::to_string(ot.allsubimages);
     ot.extract_options (options, command);
     int crop_all_subimages = Strutil::from_string<int>(options["allsubimages"]);
 


lgritz:
I think this is the correct solution. 

The idea was that when you have multiple subimages, anything that refers to specific sizes or coordinates is suspect (since the subimages can have different sizes). So we allowed a "local override" here rather than always using the global ot.allsubimages. 

But clearly I botched it, accidentally forgetting to save it as a string, and it appears to have been syntactically correct so I didn't notice the error.

Thanks for the fix! LGTM, merging into master immediately.

lgritz:
Correction: I will merge it as soon as the CI tests complete.

lgritz:
Merged into master. I'll backport to 1.7 as well.

lgritz:
Merged into RB-1.7 (we expect a tagged release next week on the 1st of the month).

e2lka:
Awesome! Thank you

PR #1614, opened 2017-02-10
For error conditions, print the full command line.

lgritz:
It's common to call oiiotool from a script, which may assemble the
command line programmatically. If it botches this and ends up with a
mangled command line, it really helps in debugging the script if its
author can see exactly what oiiotool got as a command (essentially,
as part of the original error message).

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -301,6 +301,9 @@ Oiiotool::error (string_view command, string_view explanation)
     if (explanation.length())
         std::cerr << " : " << explanation;
     std::cerr << "\n";
+    // Repeat the command line, so if oiiotool is being called from a
+    // script, it's easy to debug how the command was mangled.
+    std::cerr << "Full command line was:\n> " << full_command_line << "\n";
     exit (-1);
 }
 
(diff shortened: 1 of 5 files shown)


fpsunflower:
This sounds useful. How about listing the arguments with more of an array style? (one arg per line?)

It might help debug those funky quoting / escape symbol cases (particularly when assembling command lines through multiple levels of interpretation).

lgritz:
Well, I'm not sure how much to try to tackle here. I just wanted a quick (and not very verbose) sanity check wherein when somebody files a helpdesk ticket that quotes an error from oiiotool, I can immediately spot obvious forms of mangling and say "the argument you passed to --resize is not of the proper WxH format".

I think a detailed recounting of exactly how everything was quoted and passed might be better to put on the script side. Since on the oiiotool side, I'm only able to access the arguments after they've passed through whatever the shell does, it probably won't capture a lot of those subtleties, and it might even be misleading to give too high a level of detail, if you know what I mean.

fpsunflower:
That would be the point, to see how the arguments ended up inside oiiotool (which the user can only guess at) vs. what they print to the shell (which can look fine but still be subtly wrong - due to lack of quotes or similar issues). But its probably needlessly verbose and has other drawbacks like you can't easily copy/paste the command.

LGTM regardless - I don't think there is a solution that will be ideal for all possible scenarios and this is a simple improvement.

lgritz:
Let me try this as it is. It wasn't an external request, just something I thought of myself when I had one of those HD reports last week. If this is seen as surprisingly useful and would be even more so by being quite a bit more thorough, we can always add more detail later (including a more verbose option).

PR #1615, opened 2017-02-13
Fix to oiiotool --fit:exact=1 to avoid artifacts

lgritz:
Change the default wrap mode for the reize involved in "exact" aspect ratio fitting to "black" from the old "clamp" behavior which was found to produce artifacts in the "letterbox" region. Allow an override with an optional :wrap=name addition to -fit.

Along the way, it was helpful to add a utility method to ImageBuf that converts from string names of wrap modes to WrapMode enum.


fpsunflower:
LGTM

PR #1616, opened 2017-02-14
Add support for version 5 of the bitmap information header

Simononon:
## Description

Added support for version 5 of the bitmap information header format as per: [BITMAPV5HEADER structure](https://msdn.microsoft.com/en-us/library/windows/desktop/dd183381(v=vs.85).aspx)

Also updated version 4 to match: [BITMAPV4HEADER structure](https://msdn.microsoft.com/en-us/library/windows/desktop/dd183380(v=vs.85).aspx)

This was to fix loading the following image:  https://dl.dropboxusercontent.com/u/1960912/wicked4.bmp

Only support for the bitmap information header was added and not for ICC color profiles to keep it simple.

## Tests

Verified by loading the image in iv after applying the change and ensuring existing bmpsuite tests still passed.

------

### DIFF
--- a/src/bmp.imageio/bmp_pvt.cpp
+++ b/src/bmp.imageio/bmp_pvt.cpp
@@ -116,7 +116,7 @@ DibInformationHeader::read_header (FILE *fd)
     if (!fread (fd, &size))
         return false;
 
-    if (size == WINDOWS_V3 || size == WINDOWS_V4) {
+    if (size == WINDOWS_V3 || size == WINDOWS_V4 || size == WINDOWS_V5) {
         if (!fread(fd, &width) ||
             !fread(fd, &height) ||
             !fread(fd, &cplanes) ||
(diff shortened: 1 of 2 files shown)


lgritz:
LGTM, and it works on my end when I tried it (as you say, enabling proper read for that image, whereas it did not before). I'll backport to the release branch as well.

What's the origin of that image you gave the link for? Is it licensed in such a way that I can add it to the testsuite?

lgritz:
Also, it would be great if you could send me a CLA (though for this small patch, I will not hold up the merge while waiting for it).

Simononon:
The image was from an old [Microsoft Systems Journal article ](https://www.microsoft.com/msj/archive/S3F1.aspx) that I must have converted at some point.

It looks like any version 4 BMP converted using ImageMagick's convert command line tool will be saved as a version 5 bitmap. Here is one from the bmpsuite created using `convert g01bg.bmp g01bg2.bmp`: https://dl.dropboxusercontent.com/u/1960912/g01bg2.bmp

lgritz:
OK, great, I'll use that new one, thanks. As soon as I add that to the unit test, I'll merge your patch.

lgritz:
Merged into master already. Will merge into RB-1.7 when tests complete.

PR #1617, opened 2017-02-16
fmath: slight reformulation of clamp() to ensure NaN safety

lgritz:
Remember that floating point comparisons involving NaN always return
false. This rearrangement guarantees that clamp(x,lo,hi) always returns
a value in [lo,hi], even if x is NaN. (Specifically, it returns lo.)


fpsunflower:
LGTM

PR #1618, opened 2017-02-21
Cleanup of Strutil::format and similar functions.

lgritz:
* Found a bug in Srutil::format that was introduced recently when we
  switched it to use variadic templates -- when passed just a single bare
  string (with no arguments to format), it wasn't doing correct conversion
  of %% to %, as it should to match printf-like behavior.

* Convert all the other places that used TINYFORMAT_WRAP_FORAMT into
  much more straightforward use of variadic templates, now that we are
  C++11 minimum.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ project (OpenImageIO)
 # Release version of the library
 set (OIIO_VERSION_MAJOR 1)
 set (OIIO_VERSION_MINOR 8)
-set (OIIO_VERSION_PATCH 2)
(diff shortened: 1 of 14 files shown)


lgritz:
Amended the PR in three ways:

1. The same bug was also in ustring::format, so fixed it.

2. Updated the very old snapshot we had of tinyformat to the current TOT.

3. With the latest tinyformat and forcing use of variadic templates, format-string-only calls are allowed, I further simplified to have a single template and no "base case" to handle format string only.

PR #1620, opened 2017-02-22
Use std::regex any time it is available.

lgritz:
On the currently supported compilers, we believe the only time we need to fall back to boost::regex is for g++ 4.8.x. The rest of the time, we can rely on C++11 std::regex.


lgritz:
Turns out my logic did not work well in the situation where clang was being used to compile the source, but using the system std lib, if the system lib was from gcc 4.8 (which lacks proper regex support).

So I've updated the patch with a new strategy: use cmake's function check_cxx_source_runs() to directly test whether the std::regex facilities are found and run properly, setting -DUSE_STD_REGEX or -DUSE_BOOST_REGEX appropriately. This is cleaner and more correct than trying to guess based on just the compiler version.

PR #1625, opened 2017-02-28
Fix typo that prevented reading of all possible Exif tags in TIFF files.

### LINKED ISSUE #1624: TIFF: only a few EXIF tags are read
There is an obvious bug at:
https://github.com/OpenImageIO/oiio/blob/master/src/tiff.imageio/tiffinput.cpp#L788
```
        for (int i = 0;  tiff_tag_table[i].name;  ++i)
            find_tag (exif_tag_table[i].tifftag,
                      exif_tag_table[i].tifftype, exif_tag_table[i].name);
```

should be
```
        for (int i = 0;  exif_tag_table[i].name;  ++i)
            find_tag (exif_tag_table[i].tifftag,
                      exif_tag_table[i].tifftype, exif_tag_table[i].name);
```

lgritz:
Obvious cut and paste error. Spotted by Frédéric Devernay.

Fixes #1624

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -785,7 +785,7 @@ TIFFInput::readspec (bool read_meta)
         for (int i = 0;  tiff_tag_table[i].name;  ++i)
             find_tag (tiff_tag_table[i].tifftag,
                       tiff_tag_table[i].tifftype, tiff_tag_table[i].name);
-        for (int i = 0;  tiff_tag_table[i].name;  ++i)
+        for (int i = 0;  exif_tag_table[i].name;  ++i)
             find_tag (exif_tag_table[i].tifftag,
                       exif_tag_table[i].tifftype, exif_tag_table[i].name);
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

devernay:
looks right.
please apply on RB-1.7 too (or even RB-1.6)

lgritz:

Are you still using RB-1.6? I haven't tagged a 1.6 release since November and nobody has complained.

devernay:
We switched to 1.7 only a few weeks ago, so we don't need it, but I wondered whether there is still software building against 1.6.
The upgrade from 1.6 to 1.7 was easy (C++11 being not required), but 1.7 to 1.8 will probably be harder.

lgritz:
If anybody is still using 1.6, they have not asked me for anything to be backported since we branched 1.7. So at this point, I'm only patching 1.6 upon request.

Support and backporting of 1.7 is more active, and probably will be for longer, since it's the last C++03 branch.

PR #1626, opened 2017-02-28
RAW: Support RAW input to ACES color space.

lgritz:
The "raw" reader by default will convert its input to sRGB color space, but allows you to request that it be converted upon read to other color space (or not alter the values) by setting the "configuration" option "raw:ColorSpace" to the name of the color space you want.

This patch adds "ACES" to the list of color spaces it understands,  at least for the versions of libraw that support it (version 0.18 and newer, it seems). In cases where it's not supported, you'll get an error message like this:

    $ oiiotool -iconfig raw:ColorSpace "ACES" ~/Downloads/L1004220.DNG  -o out.jpg
    oiiotool ERROR: read : raw:ColorSpace value of "ACES" is not supported by libRaw 0.17.2

While I was in there, I changed the color space names it recognizes to be case-insensitive.

Also, in order to test that the error was reported properly, I also ran across and fixed a buglet wherein for the particular cases of "open with config" (triggered by the --iconfig option), oiiotool was not properly propagating any errors that were encountered upon opening the file.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -4089,7 +4089,12 @@ input_file (int argc, const char *argv[])
             // that information.
             ustring fn (filename);
             ot.imagecache->invalidate (fn);
-            ot.imagecache->add_file (fn, NULL, &ot.input_config);
+            bool ok = ot.imagecache->add_file (fn, NULL, &ot.input_config);
+            if (!ok) {
+                std::string err = ot.imagecache->geterror();
+                ot.error ("read", err.size() ? err : "(unknown error)");
+                exit (1);
+            }
         }
         if (! ot.imagecache->get_image_info (ustring(filename), 0, 0,
                             ustring("exists"), TypeDesc::TypeInt, &exists)
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -165,16 +165,27 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
                                              "Adobe",
                                              "Wide",
                                              "ProPhoto",
-                                             "XYZ", NULL
+                                             "XYZ",
+#if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0,18,0)
+                                             "ACES",
+#endif
+                                             NULL
                                              };
 
         size_t c;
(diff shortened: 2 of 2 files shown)


hpd:
The Python and oiiotool command-line paths both worked as expected with this update.

lgritz:
To clarify, I tried this:

    oiiotool -iconfig raw:ColorSpace "sRGB" ~/Downloads/L1004220.DNG  -o out1.jpg
    oiiotool -iconfig raw:ColorSpace "XYZ" ~/Downloads/L1004220.DNG  -o out2.jpg

As I understand it, the way the raw:ColorSpace configuration behaves is to cause libRaw to give you back values in that color space.

So I figured out1.jpg and out2.jpg should be quite different (and the one that returns XYZ, in particular, ought to look super weird in my viewer that doesn't understand XYZ).

But these files match.

Am I misusing of misunderstanding this functionality?

hpd:
The commands to exercise oiiotool are:
oiiotool -iconfig raw:ColorSpace "sRGB" test.CR2 -o test.oiio.srgb.exr
oiiotool -iconfig raw:ColorSpace "ACES" test.CR2 -o test.oiio.aces.exr

The reference libraw commands are:
dcraw_emu -w -o 1 -4 -T -W test.CR2
mv test.tiff test.dcraw_emu.srgb.tiff
dcraw_emu -w -o 6 -4 -T -W test.CR2
mv test.tiff test.dcraw_emu.aces.tiff

Exercising the OIIO Python bindings can be done using these commands:
python mkhdr.py -i test.CR2 -g 1 -o test.mkhdr.srgb.exr
python mkhdr.py -i test.CR2 -g 6 -o test.mkhdr.aces.exr

with mkhdr.py coming from
https://github.com/hpd/general/blob/master/hdr/python/mkhdr.py

The output the three tools should match very closely, for the set of 'srgb' or 'aces' images. 

To compare the validity of the matrices, you can use this Nuke. Make sure 'raw input' is checked for the Tiff images, as they contain linear values even though they are 16 bit integer images. 

The node to convert the ACES images to linear - sRGB gamut is:
set cut_paste_input [stack 0]
version 10.0 v1
push $cut_paste_input
Colorspace {
 illuminant_in ACES
 primary_in ACES
 name Colorspace1
 selected true
 xpos -237
 ypos -41
}

hpd:
I wouldn't store the results in a JPG, as the default settings for libRaw in OIIO produce linear data. Best to use EXR if possible.

PR #1629, opened 2017-03-02
Minor bug fixes in raw file input.

### LINKED ISSUE #1628: RAW: wrong default value for use_camera_matrix?
in rawinput.cpp, oiio sets use_camera_matrix to 0 by default, but the libraw default should be 1 (which means "use the camera matrix for DNG only").
ref: http://www.libraw.org/docs/API-datastruct-eng.html#libraw_output_params_t

lgritz:
Fixes #1628

1. The default value for missing "raw:use_camera_matrix" attribute should have been 1, not 0, in order to match the default behavior of libraw/dcraw.

2. The demosicing control logic wanted to erase the "raw:Exposure" attribute, but incorrectly specified the search to be constrainted to a string attribute (thereby failing to erase the attribute which is actually a float). The simple solution is that those erase_attribute calls really don't need to constrain the type at all.

3. Add support for newer (as of LibRaw 0.16) demosaic algorithms "DMT" (mode 11) and "AAHD" (mode 12). Also allow the demosaic algorithm names to be case insensitive.

4. Rename "Modified AHD" to "AHD-Mod", to simplify the name and match libraw terminology.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -30,9 +30,11 @@
 
 #include "OpenImageIO/imageio.h"
 #include "OpenImageIO/fmath.h"
+#include "OpenImageIO/strutil.h"
 #include <iostream>
 #include <time.h>       /* time_t, struct tm, gmtime */
 #include <libraw/libraw.h>
+#include <libraw/libraw_version.h>
 
 
 // This plugin utilises LibRaw:
@@ -145,9 +147,14 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
     m_processor.imgdata.params.adjust_maximum_thr =
         config.get_float_attribute("raw:adjust_maximum_thr", 0.0f);
 
-    // Use camera matrix (if config "raw:use_camera_matrix" is not 0)
+    // Use embedded color profile. Values mean:
+    // 0: do not use embedded color profile
+    // 1 (default): use embedded color profile (if present) for DNG files
+    //    (always), for other files only if use_camera_wb is set.
+    // 3: use embedded color data (if present) regardless of white
+    //    balance setting.
     m_processor.imgdata.params.use_camera_matrix =
-        config.get_int_attribute("raw:use_camera_matrix", 0);
+        config.get_int_attribute("raw:use_camera_matrix", 1);
 
 
     // Check to see if the user has explicitly set the output colorspace primaries
@@ -193,9 +200,10 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
     }
 
     // Interpolation quality
-    // note: LibRaw must be compiled with demosaic pack GPL2 to use
-    // demosaic algorithms 5-9. It must be compiled with demosaic pack GPL3 for 
-    // algorithm 10. If either of these packs are not includeded, it will silently use option 3 - AHD
+    // note: LibRaw must be compiled with demosaic pack GPL2 to use demosaic
+    // algorithms 5-9. It must be compiled with demosaic pack GPL3 for
+    // algorithm 10 (AMAzE). If either of these packs are not included, it
+    // will silently use option 3 - AHD.
     std::string demosaic = config.get_string_attribute ("raw:Demosaic");
     if (demosaic.size()) {
         static const char *demosaic_algs[] = { "linear",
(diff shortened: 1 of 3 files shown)


devernay:
Thanks!
One more thing: there are two more demosaic methods in recent libraws: DHT and a second Modified AHD, could these be added to the list?

lgritz:
Yeah, let me see if I can figure that out. Do you know how recent they are? What version libraw do I need?

lgritz:
Amended:

3. Add support for newer (as of LibRaw 0.16) demosaic algorithms "DMT"   (mode 11) and "Modified AHD 2" (mode 12). Also allow the demosaic algorithm names to be case insensitive.

devernay:
super! thanks!

devernay:
just noticed that Modified AHD 2 is called AAHD in the libraw code. That may make a nicer name (and Modified AHD is called AHD-Mod, see the top of internal/demosaic_packs.cpp in LibRaw source)

lgritz:
Amended as suggested:

3. Add support for newer (as of LibRaw 0.16) demosaic algorithms "DMT" (mode 11) and **"AAHD"** (mode 12). Also allow the demosaic algorithm names to be case insensitive.

4. Rename "Modified AHD" to "AHD-Mod", to simplify the name and match libraw terminology.

src/raw.imageio/rawinput.cpp
```diff
@@ -193,30 +200,35 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
...
                                                "AMaZE",
+#if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0,16,0)
+                                               "DHT",
+                                               "AAHD",
+#endif
                                                // Future demosaicing algorithms should go here
                                                NULL
                                                };
         size_t d;
         for (d=0; d < sizeof(demosaic_algs) / sizeof(demosaic_algs[0]); d++)
```
devernay: this should be
```
for (d=0; demosaic_algs[d] != NULL; d++)
```

src/raw.imageio/rawinput.cpp
```diff
@@ -193,30 +200,35 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
...
-            if (demosaic == demosaic_algs[d])
+            if (Strutil::iequals (demosaic, demosaic_algs[d]))
                 break;
-        if (demosaic == demosaic_algs[d])
+        if (Strutil::iequals (demosaic, demosaic_algs[d]))
```
devernay: this should be
```
if (demosaic_algs[d] && Strutil::iequals (demosaic, demosaic_algs[d]))
```

src/raw.imageio/rawinput.cpp
```diff
@@ -193,30 +200,35 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
...
             m_processor.imgdata.params.user_qual = d;
         else if (demosaic == "none") {
```
devernay: could use iequals too here

devernay:
There's another bug, the loop goes one step too far, it should be
```
for (d=0; demosaic_algs[d] != NULL; d++)
```
instead of
```
        for (d=0; d < sizeof(demosaic_algs) / sizeof(demosaic_algs[0]); d++)
```
I don't know about iequals, but == does a segfault if you compare a string to NULL. I came accross this when setting the algorithm to "none"

Similarly, the test that follows should be
```
if (demosaic_algs[d] && Strutil::iequals (demosaic, demosaic_algs[d]))
```

and the comparison to "none" could also use iequals

src/raw.imageio/rawinput.cpp
```diff
@@ -193,30 +200,35 @@ RawInput::open (const std::string &name, ImageSpec &newspec,
     }
 
     // Interpolation quality
-    // note: LibRaw must be compiled with demosaic pack GPL2 to use
-    // demosaic algorithms 5-9. It must be compiled with demosaic pack GPL3 for 
-    // algorithm 10. If either of these packs are not includeded, it will silently use option 3 - AHD
+    // note: LibRaw must be compiled with demosaic pack GPL2 to use demosaic
+    // algorithms 5-9. It must be compiled with demosaic pack GPL3 for
+    // algorithm 10 (AMAzE). If either of these packs are not includeded, it
```
devernay: includeded -> included

devernay:
line 205 includeded->included

src/raw.imageio/rawinput.cpp
lgritz: yes, thanks

src/raw.imageio/rawinput.cpp
lgritz: yes, thanks

lgritz: Actually, I think it can just be

    if (demosaic_algs[d])

since it will only be non-NULL if it matched something.

src/raw.imageio/rawinput.cpp
lgritz: yeah

src/raw.imageio/rawinput.cpp
lgritz: yes

lgritz:
Thanks for the comments! I've incorporated all the suggested fixes and pushed an update.

hpd:
This update works well in my limited testing.

devernay:
thanks for incorporating these!

PR #1631, opened 2017-03-04
TIFF: Change interpretation of XPOSITION,YPOSITION.

### LINKED ISSUE #1623: TIFF: TIFFTAG_XPOSITION is in Resolution units
### Problem

oiio interprets TIFFTAG_XPOSITION/TIFFTAG_YPOSITION as pixel units:

- https://github.com/OpenImageIO/oiio/blob/master/src/tiff.imageio/tiffinput.cpp#L689
- https://github.com/OpenImageIO/oiio/blob/master/src/tiff.imageio/tiffoutput.cpp#L298

According to the TIFF 6 spec, they are in resolution units. they should thus be multiplied by X/YResolution and rounded to the nearest int.

TIFF files with valid TIFFTAG_X/YPOSITION can be produced by nona, part of Hugin http://wiki.panotools.org/Nona

see http://hugin.sourceforge.net/docs/manual/Panorama_scripting_in_a_nutshell.html for an example of using panotools and nona to produce these tiff files from a set of photographs forming a panorama

### Versions

* OIIO branch/version: 1.7.11 + master
* OS: all
* C++ compiler:  all

lgritz:
The TIFF spec says these (which we use to generate the ImageSpec x,y fields giving the data window origin of the image) should be relative to the "resolution units" described by the XRESOLUTION and YRESOLUTION.

But OIIO has always just stored the data origin directly in this field, without regard to any scaling factor of the [XY]RESOLUTION fields.

Now, let's realize that it's not very common to have files with nonzero data origin (crop windows) in TIFF files, also not common to have any RESOLUTION tags in the files. So the problematic combination, which would be interpreted differently by versions on either side of this patch, is rare.

But one way that we can get into trouble is the following sequence: (a) a JPEG from a camera (or any number of other cases) has resolution or density data (common for cameras); (b) that JPEG is converted to TIFF, thereby copying the fields; (c) the TIFF file is cropped, and written by a version of OIIO prior to this patch. Oops.

Mitigation strategy: OIIO (at least insofar as oiiotool and maketx) likes to store its name and version in the "Software" TIFF field. So upon input, we check this, and if the file seems to have been written by OIIO with a version prior to this patch, use the old interpretation.

Fixes #1623

(Also changed some std::string& parameters used internally only to string_view, to reduce allocations from creating and destroying temporary strings as we read tag fields.) 

**I'm really nervous about this. Any objections should be voiced. If you might possibly deal with TIFF files with nonzero origin, please try this patch out before we commit it.**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -312,6 +312,10 @@ Fixes, minor enhancements, and performance improvements:
      #1642 (1.8.3/1.7.13)
    * Images with fewer than 4 channels, but one of those channels was alpha,
      were not correctly marking spec.alpha_channel. #1718 (1.8.5/1.7.16)
+   * The XPOSITION and YPOSITION tags are now interpreted as relative to
+     the RESOLUTIONUNIT, whereas before it was assumed to be measured in
+     pixels. We are confident that the new way is more in line with the
+     intent of the TIFF spec. #1631 (1.8.5)
 * webp:
    * Several new sanity checks prevent the webp reader from spending too
      much I/O time and memory reading bogus files (malformed, corrupted,
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -235,8 +235,9 @@ private:
         return ok;
     }
 
-    // Get a string tiff tag field and put it into extra_params
-    void get_string_attribute (const std::string &name, int tag) {
+    // Get a string tiff tag field and save it it as a string_view. The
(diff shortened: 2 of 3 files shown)


lgritz:
Last chance for objections. Will merge this one of these days soon.

lgritz:
This has been languishing for an exceptionally long time. I was nervous about it, but nobody has voiced objections. I'm going to run one more set of tests and then merge it unless somebody says it's a bad idea.

PR #1638, opened 2017-03-14
oiiotool --eraseattrib [regex]

lgritz:
Extend ImageSpec::erase_attribute so that the name of the attribute to
erase is a regex, and any matching attributes will be erased.

Add oiiotool --eraseattrib, taking an argument with regex pattern,
and it will strip any matching metadata from the top image.

Example uses:

    oiiooool in.jpg --eraseattrib smpte:TimeCode -o no_timecode.jpg
    oiiotool in.jpg --eraseattrib "GPS:.*" -o no_gps_metadata.jpg
    oiiotool in.jpg --eraseattrib ".*" -o no_metadata_at_all.jpg


lgritz:
User who requested this feature reports LGTM and works. I'll merge after one more set of CI tests.

Issue #1635, opened 2017-03-08
Crash on Jpeg with corrupted EXIF 


Nazg-Gul:
### Problem

OIIO will crash with specific Jpeg file with corrupted EXIF header.

**Expected behavior:** OIIO should just ignore invalid pars of EXIF.

**Actual behavior:** Any OIIO tool will crash.

This happens with 1.7.8 and latest master.

### Steps to Reproduce

Simply run any tool like iinfo on the attached file.

![wood](https://cloud.githubusercontent.com/assets/23765/23712921/96535c6c-0424-11e7-8ef4-91acff50a05b.jpg)

Here is ASAN output, just in case:  [asan.txt](https://github.com/OpenImageIO/oiio/files/828249/asan.txt)

### Versions

* OIIO branch/version:  1.7.8. and master
* OS: Linux
* C++ compiler: GCC-6

lgritz:
Fixed by #1639

lgritz:
Aside: I was unable to reproduce with this image. Maybe your original upload was broken, but by the time GitHub lets me download it, I think the offending Exif has been stripped on their end. I'm not sure GH attachments is a fruitful way to send images if the exact metadata matters.

Does GH allow zip file attachments? Maybe if you zip it up and attach the zip file, people can then download that to get the exact bit-for-bit original image.

lgritz:
Hmmm, I'm not even sure if it allows arbitrary non-image uploads. So the images it lets you attach are probably just intended to be viewed inline as documentation/explanation of the issue (which allows sanitizing of the image), and not DATA to reproduce the issue (which would require perfect preservation, including crasher corruptions).

Nazg-Gul:
Hi Larry.,

Unfortunately, i'm not sure how to reliably trigger crash on Linux. Thing is: the crash happens really easy when build with MSVC (AFAIK, it does some extra protection checks, especially in debug builds). On Linux the only reliable way to spot the issue was to compile with ASAN (-fsanitize=address). There might be some tweak to the stack protector somewhere in /sys/ to tighten check around pointer dereference, but those magic i'm not aware of. Valgrind will also show the error, but man, it's quite really slow.

As for the picture, it seems to still have the same buggy EXIF as the original file. But just in case, here is a link to original .zip archive in our bug tracker: https://developer.blender.org/F501679

Wood.jpg is the file which demonstrates the issue.

Hope that helps.

[...]

PR #1639, opened 2017-03-15
Add some extra checks to exif parser to avoid memory access issues

Nazg-Gul:
This is an attempt to solve #1635. This avoids the crash by ignoring
any directory entry which is outside of the parsing buffer (that would
be really weird if there is a legit memory pointer to an outside memory
block).

There are much more places where there is potential crash is burried
in here, but those i don't feel like going into.

For such things it's really good idea to trust no input data and verify
it, otherwise it's really easy to run into crashes and even possibly
security issues.

Maybe it's also not a bad idea to add an option tyo ignore exif headers
reading for the applications which doesn't need exif.

Unfortunately, can't run testsuite, all the tests are failing on some
`os.symlink` stuff. Why would tests mock around with file system anyway?

### DIFF
--- a/src/libOpenImageIO/exif.cpp
+++ b/src/libOpenImageIO/exif.cpp
@@ -90,7 +90,13 @@ static size_t tiff_data_sizes[] = {
 static int
 tiff_data_size (const TIFFDirEntry &dir)
 {
-    return tiff_data_sizes[(int)dir.tdir_type] * dir.tdir_count;
+    const int num_data_sizes = sizeof(tiff_data_sizes) / sizeof(*tiff_data_sizes);
+    int dir_index = (int)dir.tdir_type;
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM! These look like good extra checks on the validity of the data. I will test this a bit and then merge.

lgritz:
This seems reasonable, I will merge, but I'd love to add a test case.

The image attached to #1635 doesn't fail for me without your patch -- in fact, if I download it (I assume via right click "save link as"?) it doesn't have any Exif info or much metadata at all. Maybe GitHub strips metadata to make sure that images it displays aren't corrupt or dangerous. Can you maybe email me the original broken image? Or better yet, do you have a broken one that's fairly low-res, small enough that I can add to the testsuite to verify sane behavior?

PR #1640, opened 2017-03-17
Fix over-allocation when trying to open directory as a file

Nazg-Gul:
There are two main things changed here:

- First we check whether requested filename points to a regular file.
  This solves the issue of ftell() returning really huge value for
  file stream opened for a directory.

- Second we check ftell() output and ocmpare it to -1 to catch possible
  acess issues.

In theory it'll be nice to use Filesystem::get_size() which does all
the checking already, but leaving it as a TODO. Also the thing to note
here is that Filesystem::get_size() returns 0 for directory, which
wouldn't let us to report more meaningful error to the user about
what's exactly wrong with file one tries to open.

### DIFF
--- a/src/webp.imageio/webpinput.cpp
+++ b/src/webp.imageio/webpinput.cpp
@@ -70,8 +70,22 @@ WebpInput::open (const std::string &name, ImageSpec &spec)
 {
     m_filename = name;
 
+    // Perform preliminary test on file type.
     if (!Filesystem::is_regular(m_filename)) {
         error ("Not a regular file \"%s\"", m_filename.c_str());
+        return false;
+    }
+
+    // Get file size and check we've got enough data to decode WebP.
+    m_image_size = Filesystem::file_size(name);
+    if (m_image_size == uint64_t(-1))
+    {
+        error ("Failed to get size for \"%s\"", m_filename);
+        return false;
+    }
+    if (m_image_size < 12)
+    {
+        error ("File size is less than WebP header for file \"%s\"", m_filename);
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


Nazg-Gul:
Hi Larry,

I'll update the patch with `Filesystem::file_size()` on Monday (don't have build environment handy atm). I'm also fine and all up implementing file header check.

Ideally this is something to cover with unit tests, but i'm still struggling getting full testsuite to pass with latest OIIO master branch. So that part i'd leave for you guys to help later :)

lgritz:
I'm also happy to merge what you have so far, and you can submit a separate PR with the other fixes. Whichever way you want. Each part of this is certainly a step forward.

Nazg-Gul:
Hi again,

Did my best attempt to solve TODO and implement header test. There are still some possible improvements, but not sure how important they are at this exact PR.

lgritz:
Yes, that looks great, a really big improvement for non-webp files.

I'll test a bit and then commit (and backport).

Thanks for going the extra mile on this one.

fpsunflower:
If I understood correctly - the original bug was because the user passed a *directory* to `ImageInput::open()`, and the webp plugin happened to do something bad in that case?

Would it make sense to add logic at a higher level to make sure that invalid files aren't passed on to every single plugin? Or do we imagine that one day we may want an ImageInput plugin that treats a directory as an image?

lgritz:
Good question. I don't know. But since the "mistake a directory for a file" is not something we've seen before, it seems to be a rare edge case.

But the real design flaw was that the webp reader would read an ENTIRE file, no matter how big, before testing whether it was a legit webp! This did not depend on it being a directory.

Nor does it depend on it being named *.webp, because remember that if the first guess (based on the extension) fails to open the file -- or if any extension is not one that we recognize -- it will try every kind of ImageInput in turn until it finds one where open() succeeds. Therefore, every ImageInput variety ought to try to exclude non-compliant files with a minimum of disk read.

lgritz:
This passes CI and my hand tests on Linux and OSX, so I will merge it into master (after squashing to a single commit) and cherry-pick into RB-1.7.

fpsunflower:
Yeah that all makes sense. I was just wondering out loud if we should try to further minimize plugin churn in other cases that are obviously invalid (like: directory, missing file, 0 byte file, etc ..). But maybe those don't really deserve a dedicated shortcut since they don't need to be high performance anyway.

PR #1641, opened 2017-03-19
PSD: add support for 'cmyk', 'multichannel', and 'grayscale' color modes

### LINKED ISSUE #1603: Does OpenImageIO support nChannel images

lgritz:
Fixes #1603 

With this patch to the PSD (Photoshop) reader, I believe we now correctly read grayscale, CMYK, and "multichannel" color modes (in addition to RGB, indexed, and bitmap, as we did before). We still don't support Lab, or duotone color modes (does anybody care?).

Note that we don't correctly read 32 bit images in any of these modes (even the ones we THOUGHT we did correctly before -- they were wrong). I'll tackle that as a separate PR, or perhaps amend this one if I really find the problem immediately, we'll see. It seems to be a byte swapping issue.


lgritz:
Amended to fix 32 bit mode, mostly. Turns out 32 bit pixel channels are not uint32, they are actually float values (after endian swapping)!

I say "mostly" because there is some color space stuff to contemplate cleaning up. Seems that PhotoShop saves uint8 and uint16 images in sRGB (-ish?), whereas float images appeared to be saved as linear. I'm still looking into this matter, though. Nevertheless, what we do now (with this patch) is a whole lot less wrong than before.

lgritz:
Fixes #1603

lgritz:
Another user waiting for this gave it a thorough testing and gives it a :+1:

PR #1648, opened 2017-03-27
oiiotool: expressions recognize FRAME_NUMBER and FRAME_NUMBER_PAD

lgritz:
On oiiotool command line {expressions}, FRAME_NUMBER is replaced based
on the numeric wildcard iteration, and FRAME_NUMBER_PAD is the same
value but as a string that appears zero padded to the number of digits
specificed by --framepadding.

So, for example, to burn in frame numbers to every image in a sequence:

    oiiotool -framepadding 4 -frames 1-100 in.#.exr \
        -text:x=50:y={TOP.height-50} "Frame {FRAME_NUMBER_PAD}" \
        -o out.#.exr


fpsunflower:
LGTM

PR #1649, opened 2017-03-29
First stab at using clang-tidy

lgritz:
clang-tidy is part of the llvm/clang suite of tools. It a fancy linter, with deep ties into the clang AST, providing has a whole bunch of static code analysis tools and stylistic checks.

Turns out that cmake is pretty clang-tidy aware, so just by setting a cmake variable (CMAKE_CXX_CLANG_TIDY) to the clang-tidy command line args you want, cmake will run clang-tidy on every .cpp file that it would compile in the course of the build.

This PR adds new build flags to control this: CLANG_TIDY, CLANG_TIDY_CHECKS, CLANG_TIDY_ARGS, CLANG_TIDY_FIX

All off by default, but allows really easy use of clang-tidy while building.
For example,
    
    make CLANG_TIDY=1 CLANG_TIDY_CHECKS="-*,modernize-*"
    
or if you are really adventurous
    
    make CLANG_TIDY=1 CLANG_TIDY_CHECKS="-*,modernize-*" CLANG_TIDY_FIX=1

And then we run just one check, modernize-deprecated-headers. The second commit in the PR is the results, changing all the stray includes of old C style headers into their proper C++ names.

I did not make these changes by hand, I just used `clang-tide -fix`!

Moving forward, I'll tackle other checks one by one (and, if I agree with them, submit the changes).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,6 +281,19 @@ ifneq (${SANITIZE},)
 MY_CMAKE_FLAGS += -DSANITIZE=${SANITIZE}
 endif
 
+ifneq (${CLANG_TIDY},)
+  MY_CMAKE_FLAGS += -DCLANG_TIDY:BOOL=1
+endif
+ifneq (${CLANG_TIDY_CHECKS},)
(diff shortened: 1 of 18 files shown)


fpsunflower:
LGTM!

PR #1650, opened 2017-03-30
clang-tidy fixes: include headers in analysys

lgritz:
The first review of clang-tidy support neglected to use the --header-filter=
command, thus excluding headers from analysys.

Also, when tidying headers, if you use -fix, it's very important to
ensure that only one module is compiled at a time, or two processes
might try to modify a header at the same time, garbling it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,6 +292,8 @@ ifneq (${CLANG_TIDY_ARGS},)
 endif
 ifneq (${CLANG_TIDY_FIX},)
   MY_CMAKE_FLAGS += -DCLANG_TIDY_FIX:BOOL=${CLANG_TIDY_FIX}
+  MY_NINJA_FLAGS += -j 1
+  # N.B. when fixing, you don't want parallel jobs!
 endif
 
 ifneq (${USE_FREETYPE},)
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM

lgritz:
Yeah, it took me an extra look to realize that none of the previous things it found were in the headers, and realize that to avoid false negatives from system or dependency headers, by default it only does the tidy checks on the main module for each compile, not the material in the headers. You need to explicitly give a header filter to include them.

fpsunflower:
Makes sense. You wouldn't want it to start munging up system headers ...

lgritz:
Oh, and the filter regex is on the *included name*, not the full path.

So if the filter is "OpenImageIO/*.h", everything had better be in the form <OpenImageIO/blah.h>, because "blah.h" that just relies on the build-time include path will not match.

So coming soon is a big stupid delta where I change the form of a whole bunch of includes.

fpsunflower:
That's good. I'm a fan of referring to the public API in internal code the same way external code should refer to it. You may want to nudge the search paths you feed the compiler so that only the recommended style works.

lgritz:
Agreed -- stylistically, I have come to prefer having the includes look the same regardless of whether it's internals or an external client of the library. This gave me the excuse to do it.

In this PR, I **did** alter the search paths so that they require the OpenImageIO/blah part.  It is now like that. There remain some places that are inside quotes rather than angle brackets, I'll fix those when I spot them. Or maybe now.

PR #1651, opened 2017-03-30
Fix includes -- change lots to full <OpenImageIO/blah.h>

lgritz:
Previously we used a mix of quotes, and sometimes not having the
OpenImageIO/ qualifier.

This is helpful in clang-tidy use, because we're including headers
based on matching OpenImageIO/*.h, as a way to have it analyze all our
own headers, but no system headers.


fpsunflower:
LGTM!

lgritz:
Updated with more. Now there are also no more internal instances of

    #include "OpenImageIO/blah.h"

all have been changed to

    #include <OpenImageIO/blah.h>

no need to read through them, they are not at all exciting, it's just a global replace.

PR #1657, opened 2017-04-02
Use C++11 vector::emplace_back()

lgritz:
Using vector::push_back when constructing a new object can cause
extra copies or allocations/deallocations. The new (C++11) emplace_back
grows the vector first and then constructs the object in place, avoiding
any unnecessary moves, copies, or allocations.

Thanks, clang-tidy, for finding all of these for us with no hassle.


fpsunflower:
This is great! LGTM

lgritz:
clang-tidy is so much fun. Every day or two I pick something from the list of checks that looks interesting, kick off a build with --fix, go do something else for a while, and when I come back look over what it did. Sometimes I don't care for it, so I just revert it. But sometimes, it's something I've wanted for a long time but never had the patience to go fix item-by-item by hand. Makes me feel good after all those years of being forced to avoid C++11 idioms.

PR #1658, opened 2017-04-03
Improve TIFF I/O of color separation images

lgritz:
I had been assuming that for TIFF photometric mode "separate", it would always be 4 channels and they would be CMYK. Not so!

Turns out that there's a tag called InkSet which says if it's really CMYK or not. If it's not, check the InkNames to provide channel names. Also, the auto-conversion of CMYK to RGB should only happen if it's truly CMYK, because the conversion wouldn't be correct for arbitrary inks.

On output, also, if the request is made to output color separated mode (by setting the attribute "tiff:ColorSpace" to "CMYK" or "color separated"), set those attribute as correctly as we can -- to true CMYK if it appears to be so (4 channels, named C,M,Y,K or Cyan, Magenta, Yellow, Black), but otherwise say it's custom inks and output the channel names as ink names.

Aside: It has always bothered us that TIFF doesn't support channel names (like OpenEXR does). But it does support the InkNames tag. What if we always output the channel names as InkNames, even for non-separated photometric mode. Would there be any negative consequences? Would apps ever misbehave in the presence of an InkNames tag for an RGB photometric image, or would it just be ignored?


lgritz:
Note: the commentary on this is largely happening on #1603

lgritz:
@StephanWoe, does this look good to you?

lgritz:
@StefanWoe, does this look ok to you?

lgritz:
@StefanWoe said in #1603,
I just commented out the following line:
src\tiff.imageio\tiffoutput.cpp 449:
// if (m_photometric == PHOTOMETRIC_RGB)

Then the stored channel names seem to be correct. Without that inknames in the tiff file are not set. The only single call to
TIFFSetField (m_tif, TIFFTAG_INKNAMES,
is inside that case.

Before saving i set:
spec.attribute("tiff:ColorSpace", "color separated");
wich seems correct to me?

The problem has its root maybe in Line 396:
m_photometric = (m_spec.nchannels > 1 ? PHOTOMETRIC_RGB : PHOTOMETRIC_MINISBLACK);

lgritz:
@StefanWoe, 

> I just commented out the following line:
> src\tiff.imageio\tiffoutput.cpp 449:
> // if (m_photometric == PHOTOMETRIC_RGB)

Why does that help? Tracing through the prior code in that function, I can see only two conditions in which m_photometric would not be PHOTOMETRIC_RGB:

1. If there was only one channel in the image (in which case it would be PHOTOMETRIC_MINISBLACK).

2. If "jpeg" compression is requested (in which case it will be PHOTOMETRIC_YCBCR)

And neither of those things will happen for the multi-ink case.

I guess it can't be harmful, either, but I don't see how it helps or changes anything at all.

lgritz:
@StefanWoe, can you supply a command-line (oiiotool) or short Python (or self-contained C++) example that shows (without the changes you suggest in the earlier comment) this wrong behavior occurring?

By the way, to answer your question,

> Before saving i set:
> spec.attribute("tiff:ColorSpace", "color separated");
> which seems correct to me?

Yes, you need the color space to be "color separated" or "CMYK" in order to output the inks and set the mode to color separated. Otherwise, it will think it's just an "RGB" style image with a lot of extra channels.

lgritz:
Opinions, @StefanWoe?
Should I merge this?

StefanWoe:
Yes, sure!

PR #1660, opened 2017-04-05
timer.h: Add handy timed_thread_wedge()

lgritz:
It's for helping with benchmarks where you want to see how something scales
with number of threads.

Example output:

    threads    time   speedup  efficient  its/thread   range (best of 1)
       1        2.3     1.00x    100.00%   20000000     0.00
       2        1.2     1.97x     98.50%   10000000     0.00
       4        0.7     3.59x     89.86%    5000000     0.00

### DIFF
--- a/src/include/OpenImageIO/function_view.h
+++ b/src/include/OpenImageIO/function_view.h
@@ -1,0 +1,163 @@
+/*
+  Copyright 2017 Larry Gritz and the other authors and contributors.
+  All Rights Reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are
+  met:
+  * Redistributions of source code must retain the above copyright
+    notice, this list of conditions and the following disclaimer.
+  * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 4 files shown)


lgritz:
Update after @fpsunflower warned me against the overhead of `std::function<>`, which I was not previously aware of. Some google searching turned up a discussion and implementation of function_view, which is a much lighter-weight way to encapsulate a callable function-like object to pass around (but with the limitation of being non-owning, so just like a string_view or array_view, it is critical to make sure the view's lifetime does not exceed that of the functor it wraps).

So the review is now amended to include function_view.h (which I suspect we'll find more uses for), and then the timed_thread_wedge is changed to pass function_view's around rather than std::function.

fpsunflower:
LGTM

Did you measure any performance difference over the `std::function` implementation?

lgritz:
I didn't, specifically. I took it for granted that, by being the thinnest possible wrapper around a function pointer, it would have the same perf as func ptr, and that std::function<> couldn't possibly be faster and might be slower, as is widely documented. It also aligns well with my current tendency toward using string_view and array_view more, feels like the analogous thing for function objects.

PR #1661, opened 2017-04-06
atomic.h improvements

lgritz:
1. New atomic_fetch_add for float. It works by using an atomic_int for the heavy lifting, but using the int bit patterns of the floats we intend.

2. atomic_min/max: Some web reading makes me think that our     implementation of atomic_min and atomic_max ought to use    compare_exchange_weak rather than compare_exchange_strong. As I    understand it, the stricter requirements of the strong versions    (basically, guaranteeing to never have a "false negative" where it    thinks it needs to try again but it's actually unnecessary) can on some    architectures lead to much worse performance.
    
3. Get rid of all the pre-C++11 fallbacks. This cuts down the file considerably, leaving only the parts that aren't in C++11.


src/include/OpenImageIO/atomic.h
```diff
@@ -42,548 +42,72 @@
...
+    } while (true);
+}
+
+
+
+// Add atomically to a float and return the original value.
+OIIO_FORCEINLINE float
+atomic_fetch_add (float *af, float f)
+{
+    // C++11 doesn't support an atomic<float> with math ops.
```
fpsunflower: It doesn't have `std::atomic<float>`, but you can still use `std::atomic<float>` as a type, right?

I think that would avoid having to juggle the types back and forth.

lgritz: I believe you can have a std::atomic<float>, but it doesn't support the math (certainly not in a lock-free way). I'm providing atomic_fetch_add interfaces for both float* and std::atomic<float>.

lgritz: You have to be careful to put those template things in `backticks` or else you lose the `<float>` part of what you're talking about. :-)

fpsunflower: That's what i mean. You don't need to cast to `atomic<int>`. You can use `atomic<float>` and implement your own math ops via `compare_and_swap()` and `load()` which do work (or should).

Why do we need the `float*` overload?

lgritz: Updated. Like this?

I had done the `float*` mostly because I wasn't sure whether to trust compare_and_swap for `std::atomic<float>`, but I've confirmed that it works and has identical performance to my int trick.

fpsunflower: LGTM! This feels like a cleaner solution.

lgritz:
Once more, with feeling: I couldn't help but add the equivalent for `atomic<double>` (since my rationale the whole time was to use this for timings). Seems to work fine!

fpsunflower:
LGTM^2

PR #1662, opened 2017-04-12
Namespace overhaul

lgritz:
Mimics recent fix in OSL:

* Explicitly and automatically build the vMAJOR_MINOR versioning into
  the namespace.

* Simplify double nested namespace to just one level.

* Change OIIO_NAMESPACE_USING with `using namespace OIIO` (the macro
was obfuscating without being short or helpful).


fpsunflower:
LGTM

PR #1663, opened 2017-04-13
Fix ustring issue flagged by ubsan

fpsunflower:
Pad the length of pool allocations for ustring::TableRep so that the next one allocated will also be aligned.

This was noticed by the undefined behavior sanitizer.

I don't think this was causing any issues, but its an easy fix that doesn't cost too much.


lgritz:
LGTM

PR #1668, opened 2017-04-15
IBA::resample() and oiiotool --resample extended to work for deep images

lgritz:
Remember that "resample" is a resize with no filtering or interpolation:
for each pixel in the destination, is merely copies the source pixel whose
center is closest to the corresponding spot in the image.

This patch extends the functionality of ImageBufAlgo::resample() and
oiiotool --resample, to happily do the obvious thing for "deep" images.

This lets you quickly (but inaccurately) resize or create different-sized
proxies for deep images as you would for regular images.

Some day, it would be nice to extend a true resize() to work for deep
images, so we can have high-quality deep resizes. But that is going to
take some head-scratching about how to properly interpolate or filter
deep images.


lgritz:
Backported to 1.7

PR #1674, opened 2017-04-20
intrusive_ptr bug fix

lgritz:
I realized that if `intrusive_ptr<T>::reset(T*)` is called with a pointer
to the same object it already refers to, it could be erroneously freed
because it was decrementing the existing refcnt before incrementing the
refcnt of the new value. So rewrote reset() to be safe under that
circumstance.


fpsunflower:
LGTM

PR #1676, opened 2017-04-21
oiiotool: make --fit:exact=1 work for all subimages

lgritz:
It was just fitting the first subimage and dropping the others on the floor.


fpsunflower:
LGTM

PR #1677, opened 2017-04-23
Refactor ParamValueList

lgritz:
We can remove most of the complexity of this class by making
ParamValueList inherit from `std::vector<ParamValue>` and just add the few
extra methods we want (like a version of find that is case-insensitive
on name, but excludes potential matches that are the wrong data type).
In other words, a change of "has-a" to "is-a" makes most of the code go
away.

### DIFF
--- a/src/include/OpenImageIO/paramlist.h
+++ b/src/include/OpenImageIO/paramlist.h
@@ -39,12 +39,6 @@
 #ifndef OPENIMAGEIO_PARAMLIST_H
 #define OPENIMAGEIO_PARAMLIST_H
 
-#if defined(_MSC_VER)
-// Ignore warnings about DLL exported classes with member variables that are template classes.
-// This happens with the Rep m_vals member variable of ParamValueList below, which is a std::vector<T>.
-#  pragma warning (disable : 4251)
-#endif
-
 #include <vector>
 
 #include <OpenImageIO/export.h>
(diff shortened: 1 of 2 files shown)


src/include/OpenImageIO/paramlist.h
```diff
@@ -163,52 +157,19 @@ class OIIO_API ParamValue {
...
-    const_reference operator[] (size_t i) const { return m_vals[i]; }
-
-    void resize (size_t newsize) { m_vals.resize (newsize); }
-    size_t size () const { return m_vals.size(); }
-    bool empty () const { return size() == 0; }
-
     /// Add space for one more ParamValue to the list, and return a
     /// reference to its slot.
     reference grow () {
         resize (size()+1);
```
fpsunflower: `emplace_back()` might perform better. I'm not sure that the `resize` function is guaranteed to be amortized for adding elements one a time - the implementation might just take the resize request literally each time.

lgritz: I believe that was the goal of grow -- when you know you want one more, and are happy to do a few reallocations in order to guarantee you aren't wasting any slots.

But when you do want to amortize the cost are are ok with over-allocating, you can use emplace_back, which is now included since this class formally inherits from (rather than contains) std::vector.

PR #1680, opened 2017-05-03
Exception safety improvements for Filesystem::searchpath_find.

lgritz:
searchpath_find had a couple calls to the underlying is_directory
and is_regular_file without catching any exceptions they might throw.
Those should be changed to Filesystem is_directory and is_regular
which are our existing wrappers that already catch any exceptions.


lgritz:
Thiago reports that this works. Thanks for alerting me to the problem.
Will merge into master and RB-1.7.

PR #1681, opened 2017-05-06
Make oiiotool -autocc gracefully handle unknown color spaces.

lgritz:
When using --autocc, previously having the deduced color space be one
whose transformation was unknown was a hard error that would exit
oiiotool.  This patch changes an unknown color space be a warning, with
the rest of the oiiotool commands proceeding despite not transforming
the colors (this is just for the automatic transformations triggered by
--autocc).


lgritz:
Did you get a chance to test this, @danwexler ?

lgritz:
@danwexler reports that this works

PR #1685, opened 2017-05-11
Remove empty destructors to resolve MSVC warning C4714 

LazyDodo:
(C4714  : function marked as __forceinline not inlined)

## Description

Curious bug, MSVC refuses to inline the algorithmic operators in simd.h , but only tells you about it when you turn up the warning level to 4 or higher.  the [msdn docs](https://msdn.microsoft.com/en-us/library/a98sb923.aspx) list a number of reasons why it chooses not to inline a function, but the one that applies in this case is

> Functions returning an unwindable object by value when -GX/EHs/EHa is on.

By having a destructor (even an empty one) msvc deems the object unwindable and will refuse to inline it causing some rather severe performance issues in oiio. (blenders classroom benchmark scene rendered with OSL, reference : 09:15.15, with this fix: 06:49.26 , identical output) 

## Tests

No tests changed, besides a performance difference there's no functional changes.   Only way to really tell the change did anything is enabling assembly output from the compiler and validating it's no longer doing function calls to OpenImageIO::v1_7::simd::operator *

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -346,9 +346,6 @@ public:
     /// Default constructor (contents undefined)
     bool4 () { }
 
-    /// Destructor
-    ~bool4 () { }
-
     /// Construct from a single value (store it in all slots)
     bool4 (bool a) { load(a); }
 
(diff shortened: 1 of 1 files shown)


lgritz:
Looks good to me. Thanks for the fix, I'll merge into master and RB-1.7.

PR #1687, opened 2017-05-13
Fix cpack rename issue on OSX

### LINKED ISSUE #1682: case sensitivity issue in cmakelists.txt
This line in Cmakelists.txt -
````
file (RENAME "${CMAKE_BINARY_DIR}/cpack/README.md" "${CMAKE_BINARY_DIR}/cpack/Readme.md")
````
causes problems on recent macOS versions because the RENAME command complains that it is invalid to rename a file such that only case differs.

I thought maybe the rename could be done on the COPY in the previous line, but no.

lgritz:
Fixes #1682

A little bit of extra cleanup in that area, too.


meshula:
lgtm :)

PR #1690, opened 2017-05-22
Stop using typedefs ImageIOParameter & ImageIOParameterList

lgritz:
Just use ParamValue and ParamValueList. Those typedefs never served any purpose, just deprecate them.


fpsunflower:
LGTM

Maybe add the deprecated attribute (for gcc/clang at least?) to flag any external code that might be using them?

lgritz:
Since many people by default compile in the mode where deprecation attributes break their build, I usually phase in the deprecation over time. Doing it right in master in the same checkin where I made the change, without any real warning, seems too quick. Usually around the time that master branches for the next release (and becomes a beta), I add the deprecation attributes.

PR #1691, opened 2017-05-22
deep holdout operation

lgritz:
The deep holdout function is one that culls the deep samples in one deep    image (A) that are behind the opaque depth of a second deep image (B).    No samples of B are included in the output, it is only used to mask    parts of A. Samples of A that straddle the opaque frontier are split.

This is implemented as ImageBufAlgo::deep_holdout(), and also in  oiiotool --deepholeout.

This was suggested/requested by users. I believe the use case is that you have separately rendered volumetric elements and hard surfaces, and the user wants an oiiotool-based workflow that can inexpensively cull large sections of the volume that will get hidden by the opaque hard surfaces anyway, to make it much lighter and faster in Nuke or other processing that needs to happen to it.

Along the way, I made several other changes in the various utilities and APIs related to processing deep images (partitioned into several independent commits):

* New `ImageSpec::channelindex()` retrieves index of named channel.

* Have `DeepData::impl` find, store, and retrieve important channel indices.

* Alter `DeepData::split()` to return a bool indicating whether or not any splits occurred.

* New `DeepData::opaque_z()` method returns the z at which a deep pixel is fully opaque.

* New `DeepData::initialized()` and `allocated()` methods let you find out if the DD is initialized or allocated. As it turns out, in the end I didn't use these for the deep holdout work, but I want to keep it anyway.

This all looks complicated in toto, but if you are curious to inspect the individual commits, each one is a fairly isolated change that is easy to follow.

### DIFF
--- a/src/doc/imagebufalgo.tex
+++ b/src/doc/imagebufalgo.tex
@@ -2614,6 +2614,26 @@ the first opaque sample will be discarded, otherwise they will be kept.
 \end{code}
 \apiend
 
+\apiitem{bool {\ce deep_holdout} (ImageBuf \&dst, const ImageBuf \&src, \\
+\bigspc\spc  const ImageBuf \&holdout, ROI roi=ROI::All(), int nthreads=0)}
+\index{ImageBufAlgo!deep_holdout} \indexapi{deep_holdout} \index{deep images}
+\NEW % 1.8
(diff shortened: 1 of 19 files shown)


ben-andersen:
This is working for our test cases.
Very excited to use this.  Thanks for adding it in!

PR #1696, opened 2017-05-30
ImageCache :  Fix crash caused by dereferencing end iterator

johnhaddon:
In issue #1239, @lgritz described a situation in which the assertion in `check_max_mem()` could be triggered because `m_tilecache` could be empty, but `m_mem_used` could still be over the limit due to "dead man walking" tiles still being held in per-thread microcaches. A temporary fix was created in PR #1240, simply replacing the ASSERT with a DASSERT so that release builds wouldn't trigger the assertions. But this just allowed release builds to execute exactly the code path that the assertions were intended to protect - leading to the dereferencing of an invalid TileCache iterator. Here we simply replace the DASSERT with an early return if we detect this unexpected (but still entirely possible) situation.

## Checklist:

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -2336,8 +2336,10 @@ ImageCacheImpl::check_max_mem (ImageCachePerThreadInfo *thread_info)
     // looking up the first entry in the tile cache.
     if (m_tile_sweep_id.empty()) {
         TileCache::iterator sweep = m_tilecache.begin();
-        DASSERT (sweep != m_tilecache.end() &&
-                "no way m_tilecache can be empty and use too much memory");
+        if (sweep == m_tilecache.end()) {
+            m_tile_sweep_mutex.unlock();
+            return;
+        }
         m_tile_sweep_id = (*sweep).first;
     }
 


lgritz:
Yes, LGTM. I'll merge as soon as the CI tests finish running.

lgritz:
Merged, thanks for the fix!
I will also backport to 1.7.

johnhaddon:
Great - thanks Larry! I wasn't sure if the equivalent fix might also be necessary in `check_max_files()`, but since I don't know the internals too well I thought I'd leave that alone...

lgritz:
Ugh, it certainly does look like the same fix should be in check_max_files.
I'll do that myself and submit separately.

MrKepzie:
@lgritz See my comment in the patch that I made for our local build of OIIO:

diff --git a/src/libtexture/imagecache.cpp b/src/libtexture/imagecache.cpp
index 3474cef..599851d 100644
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -2145,6 +2145,13 @@ ImageCacheImpl::check_max_mem (ImageCachePerThreadInfo *thread_info)
     // looking up the first entry in the tile cache.
     if (m_tile_sweep_id.empty()) {
         TileCache::iterator sweep = m_tilecache.begin();
+        if (sweep == m_tilecache.end()) {
+            // Maybe we do not own on the cache because
+            // add_tile_to_cache did not find the cache entry, and
+            // holds no lock (see comment in add_tile_to_cache).
+            // In this case just return.
+            return;
+        }
         DASSERT (sweep != m_tilecache.end() &&
                 "no way m_tilecache can be empty and use too much memory");
         m_tile_sweep_id = (*sweep).first;


The issue is related to https://github.com/OpenImageIO/oiio/issues/1248, see comment N°1

lgritz:
@MrKepzie , but just a few lines earlier:

    if (! m_tile_sweep_mutex.try_lock())
        return;

So it seems to me that at this spot (where you added the lines) it MUST hold the lock; if it was unable to acquire the lock, it would have already returned. So if (sweep == m_tilecache.end()), we do indeed need to return, but I don't see how it could be right that you have not unlocked the m_tile_sweep_mutex.

Can you clarify?

MrKepzie:
It's been a while I did dive in the code but just by looking:

```
// Early out if the cache is empty
    if (m_tilecache.empty())
        return;
    // Early out if we aren't exceeding the tile memory limit
    if (m_mem_used < (long long)m_max_memory_bytes)
        return;

    // Try to grab the tile_sweep_mutex lock. If somebody else holds it,
    // just return -- leave the memory limit enforcement to whomever is
    // already in this function, no need for two threads to do it at
    // once.  If this means we may ephemerally be over the memory limit
    // (because another thread adds a tile before we have freed enough
    // here), so be it.
    if (! m_tile_sweep_mutex.try_lock())
        return;
```

Nothing is locked when if (m_tilecache.empty()) is checked, thus when getting the lock successfully, you should attempt the same condition again because another thread could have called erase() in between the 2 lines.

lgritz:
At issue is the following code (I've removed comments, and added line numbers, simply for clarity in the discussion below):

    1   if (m_tilecache.empty())
    2       return;
    3   if (m_mem_used < (long long)m_max_memory_bytes)
    4       return;
    5   if (! m_tile_sweep_mutex.try_lock())
    6       return;
    7   if (m_tile_sweep_id.empty()) {
    8       TileCache::iterator sweep = m_tilecache.begin();
    9       if (sweep == m_tilecache.end()) {
    10          m_tile_sweep_mutex.unlock();
    11          return;
    12      }
    13      m_tile_sweep_id = (*sweep).first;
    14  }
    ...

Ignoring for the moment any inclination to do a major rewrite, let's just concentrate right now on making this section of code correct.

You are saying that the check at line 1 for the tilecache being empty is before the lock that happens at line 5, and therefore  it could have become empty by the time you're at line 7 and we should check again. So I think you are suggestion to fix by adding another check between lines 6 and 7:

    if (m_tilecache.empty()) {
            m_tile_sweep_mutex.unlock();
            return;
    }

Is this the core of your original suggestion?

I *think* that in any case, the m_tilecache.empty() check is not for "correctness", but just for efficiency -- to take an early return in a case where we know (empty cache) that there are no tiles we can evict to make more room.

johnhaddon:
My understanding is that the check on line 1 is definitely not valid by the time we arrive at line 5 - `m_tileCache` can become empty immediately after we check, due to action on another thread. That's what leads to the original bug being possible in the first place.

But I didn't think that holding `m_tile_sweep_mutex` protected against removal of items from `m_tileCache` anyway - I thought it was to protect only `m_tile_sweep_id` and to make sure only one thread bothers to sweep at a time. I say this because `m_tileCache.erase()` appears to be called elsewhere without the lock - in `invalidate()` for instance. So I don't think the insertion of a second check between 6 and 7 helps does it? That's why I used the `sweep == m_tilecache.end()` instead...

MrKepzie:
Adding 

```
if (m_tilecache.empty()) {
        m_tile_sweep_mutex.unlock();
        return;
}
```

Between  6 and 7 is OK, but not necessarily needed since you added lines 9-11

This does not remove the race condition but in its own it is just not too harmful,
except for applications putting a lot of pressure on the ImageCache with tons of threads (like we used to do in our OpenFX plug-in). 

Still you should consider having a global lock on m_tilecache and perform some benchmarking on the memory management and performance in multi-threaded environments. 

I agree that this is a terrible solution for a very demanded cache, but at least you don't end up in situations where the memory limit is going wild because of race conditions.

PR #1701, opened 2017-06-06
Fix Ilmbase linker warning with LINKSTATIC on Windows.

brechtvl:
After recent refactoring OPENEXR_DLL is defined when LINKSTATIC=ON which I'm guessing was not an intentional change. This restores the behavior before the refactor, fixing a bunch of LNK4217 warnings when building with static IlmBase libraries.

This matches the OSL fix in imageworks/OpenShadingLanguage#768


lgritz:
LGTM, thanks

PR #1702, opened 2017-06-06
Fix Windows build to not force enable /MT, is not need for BUILDSTATIC

brechtvl:
As part of Appveyor support, /MT was always enabled when BUILDSTATIC=ON. The only way to build with /MD now seems to be to patch the OSL source to remove this, whereas previously it was possible to build with custom `CMAKE_CXX_FLAGS_RELEASE` or similar flags.

Note that /MT and /MD control if the MSVCRT.lib system library is statically or dynamically linked, which is orthogonal to the choice of creating a static or shared OSL library.

This commit only enables /MT for Appveyor builds, restoring the previous behavior for regular builds.

This matches the OSL fix in imageworks/OpenShadingLanguage#769


lgritz:
LGTM

lgritz:
Unfortunately, this broke our Appveyor build!
I probably should have waited for the CI tests to come back before merging.

brechtvl:
```
git fetch -q origin +refs/pull/1702/merge:
fatal: Couldn't find remote ref refs/pull/1702/merge
```

I guess Appveyor can't build closed pull requests? Or is there an error due to the changes in the commit itself?

lgritz:
Never mind, looks fine. I think I probably deleted the branch while it was trying to check it out.

PR #1705, opened 2017-06-09
IBA::deep_holdout -- fix bug when channels didn't match

lgritz:
deep_holdout() was crashing when the holdout image had more channels
than the input image. It was improper to rely on IBAPrep's default behavior
of setting the result channels to the maximum of the channels in any of
the input images. In this case, we want to force it to match the "shape"
of just the first (source) input, not the second input that only serves
to act as the depth holdout.

While I was there, though, I tightened up the logic of how IBAPrep deals
with the force_spec parameter, and also made that parameter a const ptr,
as it always should have been.


lgritz:
This PR is on hold -- we realized that what I implemented doesn't actually do the right thing.

I'm going to rename what I implemented `deep_cull()` since what it does is simply drop samples in the source images that are beyond the opaque threshold of the holdout image, without modifying any other samples. Holdout images that have no fully opaque samples have no effect on the result.

Instead, the new thing that will be the real `deep_holdout()` will be a true "holdout" operation, where for example even a partially-transparent sample in the holdout image will partially block source samples behind it. The holdout operation is one in which the holdout's alpha attenuates the source, but does not itself contribute to color or opacity. It only blocks. The deep analogy to a 2D flat holdout operation where partial holdout is allowed.

lgritz:
I pushed an update to this PR. In addition to fixing the minor issues for which I first proposed this patch, I have amended it with the following major change:

* The prior deep_holdout has been renamed deep_cull -- it just drops samples behind the opaque frontier of the holdout image.

* A new deep_holdout implementation truly does the deep implementation of the classic "holdout" compositing operation.

* A bunch of small changes to DeepData have been made to facilitate the new operation.

PR #1711, opened 2017-06-28
Port iv to Qt5

pinotree:
- switch the cmake search from Qt4 and Qt5, making use of the components for the Qt5 libraries
- adopt the automoc handling provided by cmake, so there is no more need to manually run & add moc sources
- make use of the cmake imported targets for the Qt5 libraries, so there is no need to manually care about variables for libraries and include directories

## Description

This is the swtch of the `iv` tool from Qt4 to Qt5, as Qt4 is EOL for a couple of years already.
Luckly Qt5 is generally compatible with Qt4, so there are no code changes required.

TBH I am not sure whether this is considered "small change" or not, although almost all the changes were:
- replace Qt4 with Qt5 in the build system
- use the Qt5 cmake components
- enable the cmake automoc, and drop the own code
- use the `Qt5::` imported targets
which would be done almost in the same way by anybody else skilled in Qt5/cmake...

## Tests

1. build OIIO with Qt5 headers installed, Qt5 is found, and iv is built
2. ran the newly built `iv`, and do some basic operations

## Checklist:


lgritz:
Wow. This actually works? It turned out only require build script changes, and no edits to the iv code itself was necessary to port from Qt4 to Qt5?

pinotree:
> Wow. This actually works? It turned out only require build script changes, and no edits to the iv code itself was necessary to port from Qt4 to Qt5?

Commit a0d42b4a39670a0d4f25e81b19a5112d916508ce already fixed the Qt includes, so there is nothing more to just build.  Of course `iv` needs thorough testing, but builds and runs fine, so it can be done step by step.

src/cmake/externalpackages.cmake
```diff
@@ -243,19 +243,18 @@ endif ()
 # Qt setup
 
 if (USE_QT)
+    set (qt5_modules Core Gui Widgets)
     if (USE_OPENGL)
-        set (QT_USE_QTOPENGL true)
+        list (APPEND qt5_modules OpenGL)
     endif ()
-    find_package (Qt4)
+    find_package (Qt5 COMPONENTS ${qt5_modules})
```
lgritz: ...but forgot to ADD... is what I meant here

lgritz: Wait, maybe never mind. Builds and runs fine on Linux. Just not on my OSX laptop. Must be a problem on my end.

pinotree:
> FindQt5.cmake doesn't appear to be a standard part of the CMake installation. Is this a file you intended to include but forgot to find? Where do I find the .cmake module to allow a findpackage(Qt5)?

Qt5 ships cmake configuration modules (`Qt5CoreConfig.cmake`, etc -- see the various development packages of qtbase/etc), so `find_package(Qt5 COMPONENTS Core)`:
- gives a cmake warning if `Qt5CoreConfig.cmake` is not available
- works if it installed

lgritz:
Yes, works for me! Merging. Thanks for this fix!

We are now upgraded to Qt 5.6 minimum in master (per 2017, and even 2016 VFXPlatform guidelines). 

No more Qt4.

PR #1714, opened 2017-06-30
PSD: Fix load_layer_channel issue for the layer mask channel

yamasakia:
Some PSD files failed to open. I fixed the problem of the PSDInput::load_layer_channel.

## Description

The layer mask channel has independent rectangle position from the regular channels. Mostly, they are same position, but sometimes not. PSDInput::load_layer_channel did not consider about it. Therefore, PSDInput::open failed by reading wrong data as ChannelInfo::compression when the pixel size of the layer mask channel is different from the image pixel size.

## Tests

I added new test source which caused the issue to the psd testsuite.

## Checklist:


lgritz:
Sorry for the delay. This looks good to me! Will merge right away.

Is master ok for you (it will be a release in the next several weeks), or do you really need this backported to an older release branch?

yamasakia:
I don't need a backport. Thank you!

PR #1718, opened 2017-07-07
TIFF: clean up alpha channel handling

lgritz:
For 2-channel images where the first channel is intensity and the second is
alpha, TIFF output was neglecting to set the TIFF "EXTRASAMPLES" tag to mark
the alpha channel. In particular, it just didn't consider extrasamples for images
that had fewer than 4 channels.

This was in turn causing the file, when re-input, to not realize that
the second channel was "A" or that there was an alpha channel at all
(since TIFF does not store channel names, we infer it all from the
EXTRASAMPLES tag).

One consequence of this mistake was that a grey-alpha image (say, coming
from OpenEXR or PNG or some other format that allows it) and run through
'maketx' (converting to TIFF if not otherwise specified) would lose this
knowledge, which in turn would affect the ability of the TextureSystem's
"gray_to_rgb" mode, which only applies to certain channel combinations.

Also noticed that for TIFF input, 2-channel images where the second is
alpha were still being named "R","A", but we actually changed the convention
some time ago for a single intensity channel to be named "Y", no "R", since
it's almost certainly representing luminance and not "red" per se.

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -867,8 +867,13 @@ TIFFInput::readspec (bool read_meta)
                 }
             }
         }
-        if (m_spec.alpha_channel >= 0)
+        if (m_spec.alpha_channel >= 0) {
             m_spec.channelnames[m_spec.alpha_channel] = "A";
+            // Special case: "R","A" should really be named "Y","A", since
+            // the first channel is luminance, not red.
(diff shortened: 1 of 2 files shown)


skwerner:
Looks good to me and solves the problem I saw with Y,A textures and TextureSys.

lgritz:
Thanks, merging.

PR #1719, opened 2017-07-07
simd.h overhaul for 16-wide and AVX-512 support

lgritz:
* Add float16, int16, bool16 classes.
    
* When 16x HW support (AVX512) is selected, it is used, but otherwise just fall back to 8x[2] style      implementation. The aim is to make it convenient to code AS IF we were on AVX-512, but have pretty decent performance for AVX. The 16-wide classes should also run for SSE class hardware (or no SIMD at all), but I'm not working hard to optimize its performance in that case. Let's just accept that using 16-wide classes if you don't have at least 8-wide HW available is not going to get you a performance gain over scalar code.

* Preliminary support for masked loads & stores (including propagating those to 8x and 4x, which admittedly will not be very efficient for older HW).
   
* Lots of miscellaneous fixes to the 4x and 8x classes and cleanup of      this big header file (lessons learned, and conform to more clear      semantics apparent when expanding to 16x). This includes, when AVX512 is enabled, using some of the new intrinsics to speed up 8x and occasionally 4x tasks as well, in cases where some nice new intrinsics have been added.

* Big refactor in simd_test.cpp (unit tests) to better accommodate the additional classes without too much additional replicated code for each class.

* New OIIO getattribute() tokens to query hardware and build-time SIMD capabilities: "oiio:simd" retrieves the comma-separated list of mostly-SIMD-related capabilities that were enabled at build time (e.g., "sse2,sse3,sse41,sse42,avx"), and "hw:simd" is a similar list of the capabilities detected  on the actual hardware at runtime.

* Add Travis tests that build for AVX2 and AVX-512F. The Travis VMs can't run these, but we can at least ensure that they build cleanly, thus reducing the chances of accidentally breaking those cases (so many `#if` clauses).

### DIFF
--- a/src/include/OpenImageIO/color.h
+++ b/src/include/OpenImageIO/color.h
@@ -251,7 +251,7 @@ inline float sRGB_to_linear (float x)
                            : powf ((x + 0.055f) * (1.0f / 1.055f), 2.4f);
 }
 
-inline simd::float4 sRGB_to_linear (const simd::float4& x)
+inline simd::vfloat4 sRGB_to_linear (const simd::vfloat4& x)
 {
     return simd::select (x <= 0.04045f, x * (1.0f/12.92f),
                          fast_pow_pos (madd (x, (1.0f / 1.055f), 0.055f*(1.0f/1.055f)), 2.4f));
@@ -266,9 +266,9 @@ inline float linear_to_sRGB (float x)
 
 
 /// Utility -- convert linear value to sRGB
-inline simd::float4 linear_to_sRGB (const simd::float4& x)
+inline simd::vfloat4 linear_to_sRGB (const simd::vfloat4& x)
 {
-    // x = simd::max (x, simd::float4::Zero());
+    // x = simd::max (x, simd::vfloat4::Zero());
     return simd::select (x <= 0.0031308f, 12.92f * x,
                          madd (1.055f, fast_pow_pos (x, 1.f/2.4f),  -0.055f));
 }
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -223,15 +223,15 @@ clamp (const T& a, const T& low, const T& high)
 }
 
 
-// Specialization of clamp for float4
-template<> inline simd::float4
-clamp (const simd::float4& a, const simd::float4& low, const simd::float4& high)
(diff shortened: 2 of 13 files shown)


wrosecrans:
Something called float16 that isn't a 16 bit "half" float will definitely confuse folks the first time they run into it.  Not that I have a better idea.  Just don't be shocked if some folks find it less than obvious in the future.

lgritz:
That's not a bad point, and I did get momentarily confused myself a couple times upon seeing "int16", which also has a confusing aspect to it.

On the other hand, this is all inside the "simd" namespace. So generally you'd refer to it as `simd:float16`.

If people think this is confusing, I can (ugh) rename them all like vfloat16, vint16, etc., to emphasize that they are simd vectors, as opposed to 16 bit floats. That would be a PITA, but maybe it would be more clear.

lgritz:
Offline conversation: Chris also agreed that 'vfloat', etc., probably makes sense to avoid confusion.

What I'd like to do, though, is get this PR integrated first, then have a separate one that is strictly about the great simd rename of 2017.

lgritz:
I pushed an update to this PR that additionally does the big rename of floatN,intN,boolN to vfloatN,vintN,vboolN.

The renaming is in an isolated commit.

Typedefs keep the old names working, for any 3rd party projects that use OIIO's simd.h. We'll get rid of the compatibility typedefs eventually, after we are confident that any downstream packages have transitioned.

KindDragon:
JFYI VS2017 will support it too https://blogs.msdn.microsoft.com/vcblog/2017/07/11/microsoft-visual-studio-2017-supports-intel-avx-512/

lgritz:
Any objections to this? (Notwithstanding any improvements that can be made in future follow-up patches.)

lgritz:
@KindDragon as far as the _vectorcall goes -- I've taken to making all of these functions forced inline and passing input params as `const floatN&`. I think that means that `_vectorcall` wouldn't add anything, except for true (non-inlined) functions. Do you agree?

KindDragon:
@lgritz Agree, maybe for big functions like madd, exp, transpose ...

lgritz:
Yes. Let me find a graceful way to do that as a separate review, though.

fpsunflower:
LGTM

PR #1720, opened 2017-07-07
Remove C++11 guards that are no longer needed

lgritz:
Just removing some pointless `#if` guards that are no longer needed.


fpsunflower:
LGTM

PR #1721, opened 2017-07-10
Gcc 7 1 0 / Qt 5.8 fixes

Shootfast:
## Description

The following are small fixes caught by the GCC 7.1.0 compiler when used in combination with Qt 5.8.
As it stands there are still a number of warnings being emitted, but their scope was much greater than my drive-by edit had time to fix.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -42,6 +42,7 @@
 #include <algorithm>
 #include <atomic>
 #include <future>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <thread>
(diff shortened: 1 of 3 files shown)


lgritz:
LGTM. As soon as I try it at work to make sure nothing breaks there, I will merge it.

Shootfast:
There are actually a whole bunch more compiler warnings from GCC 7.1.0, though I'm not able to address them all at the moment.

lgritz:
Do you mean you don't have time to address them, or you don't know HOW to address them? If the latter, feel free to just post the error messages, and I may be able to make a good guess about how to fix the warnings.

In the mean time, I'll see if I can set up Travis to add a gcc7 compile to the mix, that'll get things cleared up fast.

lgritz:
These changes seem fine both on my laptop and work machine, so I will merge now. Thanks for the fixes!

PR #1722, opened 2017-07-10
Fix python site-packges path for install

### LINKED ISSUE #926: OpenImageIO python module installed by default into wrong location?
$PREFIX/lib/python/site-packages

Doesn't appear to be correct. It should be /lib/python$MAJ.$MIN/site-packages to be picked up by default.

Thanks

lgritz:
Floating this test... maybe it fixes #926 ?


lgritz:
Anybody think this is helpful?

johnhaddon:
Yep, seems helpful to me. Everything else I know of installs into a versioned python directory, and sooner or later we'll be moving to python 3...

PR #1723, opened 2017-07-11
Fix invalid string_view return from channel_name()

c42f:
In this case implicit conversion caused the allocation of a temporary
std::string, which is deallocated before the string_view can be used.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -402,7 +402,7 @@ public:
     /// channelnames is not filled out.
     string_view channel_name (int chan) const {
         return chan >= 0 && chan < (int)channelnames.size()
-            ? channelnames[chan] : "";
+            ? string_view(channelnames[chan]) : "";
     }
 
     /// Fill in an array of channel formats describing all channels in


lgritz:
Wow, that's some catch! How did you find that?

LGTM, merging as soon as the CI tests finish.

Oh, and good to see you here, Chris!

c42f:
Hah, thanks!  It sure was a head scratcher for an hour or two. I found this while working on julia bindings - https://github.com/c42f/OpenImageIO.jl (very rough, be warned). After implementing type marshalling for `string_view` I was getting randomly corrupted results in the julia REPL when calling `channel_name`.

By the way, julia is a super awesome language and I think the VFX industry would love it. It's got the speed of C and the productivity of python (or more so, for heavy numerical work). The main julia image library currently relies on ImageMagic but it's pretty crusty and people really need something more capable. I told them oiio was the production quality tool they were looking for :-)

lgritz:
Neato, thanks for the plug in Julia land.

So sorry for the bug, it's all my fault. (Though in my defense, it's damn subtle.)

lgritz:
CI tests pass, merging.

c42f:
No worries at all, bugs are a fact of life.  I'm not yet sure I even need `channel_name`, I'm still at the exploratory stage with understanding how the wrapper library (CxxWrap.jl) works.

PR #1733, opened 2017-07-24
SIMD batched texture APIs

lgritz:
There was previously a set of TextureSystem API calls (with multi-point TextureOptions struct and a fancy VaryingRef helper class) that was a multi-point interface, but we never did much with it and in retrospect, it does not use a data organization that is good for future-looking architectures. We are therefore depcrecating it, it will remain for 1.8 in case anybody is using it, but will thereafter be removed. 

But we are adding a new set of batched routines that are organized differently, we think in a way that will be well-suited for better parallelism.

The new API assumes that the inputs and outputs have a SOA rather than AOS layout, and the arrays are a fixed width, defined by Tex::BatchWidth, which is a constant and set to 16. Build-time overrides let us change this for experimentation and future expansion. You can texture fewer points than this using the RunMask, which is a bit field saying which "lanes" should be used. Inputs will not be read, or outputs written, for the lanes where the run mask is 0.

The goal is for this to really fly on chips with ISA that includes AVX-512 support (naturally 16 wide), and to also perform very well for AVX2 class machines (emulating 16x operations with pairs of 8x SIMD instructions). I'm not very hopeful that old SSE hardware (4x wide) will have a high payoff using the batched functionality rather than the single-point texturing, but we'll see what happens. We'll also experiment with 32 wide (there's no true 32 wide hardware, but pairs of 16-wide instructions may dual-issue in nice ways that provides an advantage, we'll see).

I'm not planning to ever remove the single-point scalar API calls. There will always be a use for them and they'll certainly be more efficient for applications that don't have a good way to batch the calls.

This pull request only implements the batched APIs -- the internals are still scalar, they just loop over the enabled lanes and call the scalar functionality.

I'm actually working on the true batched reversions right now, but they aren't ready to review yet. Expect a second PR for that as soon as I have good results. But in the mean time, this allows applications to start experimenting with the batched interface.

The testtex program now has modes that exercise the batched interfaces, and the testsuite has been doctored so that every test whose name contains the substring "texture" will run twice: once using the old scalar mode, and once using the batched API calls. This means that starting now, we will immediately notice any breakages or regressions in the batched versions, as well as have a convenient way to benchmark the scalar versus batched calls.

If you have any issues with this organization of the batched APIs -- for example, if you can see that I'm making some kind of stupid mistake that I'll regret later -- please speak up NOW. I'd prefer that the APIs have a good chance of being stable, since we are getting close to branching a release. It's much easier to address any deficiencies now, rather than after I'm finished fully implementing the internals.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ set (USE_EXTERNAL_PUGIXML OFF CACHE BOOL
      "Use an externally built shared library version of the pugixml library")
 set (PUGIXML_HOME "" CACHE STRING "Hint about where to find external PugiXML library")
 option (USE_DICOM "Use DICOM if DCMTK is found" ON)
-
+set (TEX_BATCH_SIZE "" CACHE STRING "Force TextureSystem SIMD batch size (e.g. 16)")
 set (SOVERSION ${OIIO_VERSION_MAJOR}.${OIIO_VERSION_MINOR}
      CACHE STRING "Set the SO version in the SO name of the output library")
 option (BUILD_OIIOUTIL_ONLY "If ON, will build *only* libOpenImageIO_Util")
@@ -106,6 +106,9 @@ if (NOT OIIO_THREAD_ALLOW_DCLP)
     add_definitions ("-DOIIO_THREAD_ALLOW_DCLP=0")
 endif ()
 
+if (TEX_BATCH_SIZE)
+    add_definitions ("-DOIIO_TEXTURE_SIMD_BATCH_WIDTH=${TEX_BATCH_SIZE}")
+endif ()
 
 # Set the default namespace
 set (PROJ_NAMESPACE "${${PROJ_NAME}_NAMESPACE}")
--- a/Makefile
+++ b/Makefile
@@ -266,6 +266,10 @@ ifneq (${USE_SIMD},)
 MY_CMAKE_FLAGS += -DUSE_SIMD:STRING="${USE_SIMD}"
 endif
 
+ifneq (${TEX_BATCH_SIZE},)
+MY_CMAKE_FLAGS += -DTEX_BATCH_SIZE:STRING="${TEX_BATCH_SIZE}"
+endif
+
 ifneq (${TEST},)
 TEST_FLAGS += -R ${TEST}
 endif
@@ -499,9 +503,10 @@ help:
 	@echo "      OIIO_BUILD_TOOLS=0       Skip building the command-line tools"
 	@echo "      OIIO_BUILD_TESTS=0       Skip building the unit tests"
 	@echo "      BUILD_OIIOUTIL_ONLY=1    Build *only* libOpenImageIO_Util"
-	@echo "      USE_SIMD=arch            Build with SIMD support (choices: 0, sse2, sse3,"
-	@echo "                                  ssse3, sse4.1, sse4.2, f16c, avx, avx2"
-	@echo "                                  comma-separated ok)"
+	@echo "      USE_SIMD=arch            Build with SIMD support (comma-separated choices:"
+	@echo "                                  0, sse2, sse3, ssse3, sse4.1, sse4.2, f16c,"
+	@echo "                                  avx, avx2, avx512f)"
+	@echo "      TEX_BATCH_SIZE=16        Override TextureSystem SIMD batch size"
 	@echo "  make test, extra options:"
 	@echo "      TEST=regex               Run only tests matching the regex"
 	@echo ""
--- a/src/cmake/oiio_macros.cmake
+++ b/src/cmake/oiio_macros.cmake
@@ -96,42 +96,37 @@ macro (oiio_add_tests)
         message (STATUS "  -> You can find it at ${_ats_URL}\n")
     else ()
         # Add the tests if all is well.
-        if (DEFINED CMAKE_VERSION AND NOT CMAKE_VERSION VERSION_LESS 2.8)
-            set (_has_generator_expr TRUE)
-        endif ()
+        set (_has_generator_expr TRUE)
         foreach (_testname ${_ats_UNPARSED_ARGUMENTS})
             set (_testsrcdir "${CMAKE_SOURCE_DIR}/testsuite/${_testname}")
             set (_testdir "${CMAKE_BINARY_DIR}/testsuite/${_testname}")
-            if (_ats_LABEL MATCHES "broken")
-                set (_testname "${_testname}-broken")
-            endif ()
             if (_ats_TESTNAME)
                 set (_testname "${_ats_TESTNAME}")
             endif ()
-            if (_has_generator_expr)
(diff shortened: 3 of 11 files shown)


lgritz:
Amended with updated comments in the header file and the new methods documented in the main OIIO documentation.

lgritz:
A design question, for those who want to play along:

I implemented this by having the TextureOptBatch actually hold the full arrays of the "varying" fields (which really are only width and blur).

Design alternative #1: Instead of float arrays, it could have been an actual SIMD type, Tex::FloatWide (which is a typedef of the simd.h type of length BatchWidth).

Design alternative #2: instead of storing either an array or a SIMD type, it could just be a `float *`, pointing to some area of app memory containing the array, and thus making the TextureOptBatch structure much smaller (and trading an extra level of pointer indirection for, maybe, avoiding an extra copy)?

lgritz:
I decided that for now, I like the way I first proposed the API, not the alternatives. 

If there are no objections, I'll merge this. I have also been working on the implementation, but there's still a lot to do before it's ready for a review.

AlexMWells:
Sorry for the delay.  

On the TextureOptBatch, having embedded data vs. pointers to arrays
 float sblur[Tex::BatchWidth];   
vs.
float *sblur;   

As you currently have it, there is an extra copy.  From OSL's point of view we have a Wide symbol in memory that could be directly passed vs. copy.  However, that would mean a set of pointers instead of struct members that can be directly accessed.  I'm wondering what it would look like if those where just passed in as part of the texture call itself?
virtual bool texture (ustring filename, TextureOptBatch &options,.... const float * sblur, ...);

AlexMWells:
I am also trying to decide if the TextureOpt, TextureOptBatch is more like a context if it is not constant.   I'm also concerned about overhead of parameter passing to the texture() calls and any subsequent calls.  Do compiler's end up repushing all parameters to nested calls?  I'm wondering if TextureOptBatch could be a base class with specific derived versions for texture|texture3d|environment that have the const float *s, const float *t,
   const float *dsdx, const float *dtdx,
   const float *dsdy, const float *dtdy
as data members, then the calling side can populate it once and all nested calls can just access that single structure.

lgritz:
Oy, this had to come the day AFTER I merged the patch?  :-)

From OSL's point of view (correct me if I'm wrong), there's a TextureOpt in the ShadingContext, and that's where the memory lives for the value of the blur, etc. If the TextureOpt contains just a pointer, then you still need to store the value somewhere. So that's why I left it in the TextureOpt(Batch).

Right now, the ONLY varying things in the TextureOptBatch are the blur and width fields. If we passed blur and width pointers as separate parameters to the texture API calls, we would (a) still need to store them somewhere else, and (b) have an awful lot more parameters to those calls.

I don't understand your second comment. Nested calls?

AlexMWells:
I'll backup first and say the interface looks fine and will work.  I think the main decision made in the interface is that the caller is responsible for binning and providing a mask, and since OSL knows exactly which options are varying it is the best position to do that work efficiently. 

I'm just discussing other options (6 of one/half dozen of another).  

By "Nested calls", I just mean as parameters get passed down to calls that actually do the work, each subsequent function call needs its parameters pushed onto the stack.  So if there are 6 pointers being passed, they got pushed onto the stack, function called, inside another function is called taking maybe 4 of those 6 parameters, so those 4 are read from the stack, then 4 pushed back onto the stack in the order expected by the nested function call.

I'm just hypothesizing that instead of pushing the parameters onto the stack, a structure on the stack could be populated and a single pointer to that structure be pushed on the stack through any calls.  It is arguable TextureOptBatch does this, but I was looking at all those other parameters to the 'texture' call and wondering if sending over an extended "TextureBatchContext : public TextureOptBatch" that has all of the parameters present.  Savings would be in not needing to repush the parameters.  At a minimum I suspect we will be dispatching to a helper function based on the values in TextureOptBatch and need to repass all those parameters.

AlexMWells:
for blur and width, if passed as a texture param, yes caller would be responsible for having a float[Tex::BatchWidth] to back the pointer it passed.  If the blur or width is truly varying, then OSL already has a float[Tex::BatchWidth] backing the symbol being passed.  The extra work would be for uniform value, in which case caller would have to broadcast value out to a float[Tex::BatchWidth] on the stack and pass that.  Its a fair trade off if you are planning on just having the texture code assume blur and width are always varying (is there any saving's possible from detecting they are uniform or having the caller indicate so?)

AlexMWells:
A totally different approach could be to have all of the varying parameters to texture be a single structure with no pointers, but actually backed by float[Tex::BatchWidth].  So instead of pushing pointers on the stack, the actual wide values would be stored in the structure.  Then subsequent code can use a single pointer to the structure + offsets known at compile time to reduce register pressure.  Trade off is higher cost to setup the structure, but maybe cost to copy wide data is similar copying a pointer.

AlexMWells:
And all of my discussion could fall under the heading of premature optimization and might be best getting any batched system up and running so we have a baseline to try some of these permutations from.

lgritz:
High-order bit: I agree that this is all premature optimization. I will almost certainly leave it as I currently have it until I have enough of the work done to get reliable benchmarks, and then I can try some of the permutations we have discussed and let the empirical results guide us.

I realized after I posted, but unfortunately did not get to replying until after you had, that my argument about where to store it is silly. If the TextureOptBatch structure just had a pointer, that would be fine, because it's coming from a variable anyway. Though, still, one or two extra copies of a float16 is, I bet, not even going to be detectable in the context of everything else a texture lookups does.

I don't think the "nesting" is as bad as you hypothesize -- mostly a few intermediate calculations are done and it's those that have to be passed along, not the original values. For example, the blur and width are used to rescale the derivatives, and in turn the derivs are boiled down to MIPmap levels and the weighting between them. 

I don't think there's any advantage to considering blur and width as uniform. The way they are used is primarily to rescale the derivatives (roughly speaking, scaled_derivs = derivs * width + blur), and since we can count on the input derivatives as being varying, even a uniform width and blur would need to be internally promoted to a vector for the resulting simd math ("broadcast", as you say in Intel intrinsic land). I doubt that there's a perf difference whether the broadcast happens on the app side of the API or in the TextureSystem, but it certainly simplifies the API if it assumes the varying case and doesn't need to handle both varieties.

AlexMWells:
Sounds good.

That perhaps brings it back to the original question of which parameters should be forced to be uniform.
My question is, given a set of uniform values that determines an algorithm code path (interpmode , wraps, mipmode) is there any reason why varying subimages couldn't be pushed through the same algorithm?  For that matter any reason multiple texturehandles couldn't be pushed through the same algorithm?

Corollary question, what are chances the  (interpmode , wraps, mipmode) are the same for different subimages?  And what are chances the  (interpmode , wraps, mipmode) are the same for different texture handles?

Unrelated question, why offer an interface taking 'ustring filename' versus just making user call get_texture_handle themselves?

PR #1734, opened 2017-07-25
Simplify array_view and array_view_strided to 1D only.

lgritz:
These were templated on dimensionality of the array, and that made the implementation much more complex and quite hard to read and understand. But we never used anything but the 1D case. The reason it was templated on rank before was to emulate the array_view that was proposed for C++17, but that has since been scrapped, so keeping compatibility with it is no longer a priority.

When we remove "Rank" from the template and take away all the machinery that isn't needed for the simple 1D case, a lot of the complexity goes away, and these templates will be a lot easier to understand and modify in the future.

If anybody is using the 2D (or higher) use cases for array_view, or think it's important to keep that functionality (and accompanying implementation complexity), please speak up.


lgritz:
Anybody have pro/con arguments?

lgritz:
If I don't hear any objections by the end of the day, I will merge this simplification.

lgritz:
Last chance to object to simplifying the array_view template to handle the 1D case only.

PR #1735, opened 2017-07-26
Fix important performance issues with IBA::deep_merge()

lgritz:
When merging, high number of splits was crippling performance due to
constant realocation.

New strategy more carefully pre-computes the NUMBER of splits expected
in the merged pixel, and allocates enough storage for the result pixel
to hold the number of samples in both of the input images, and also the
expected number of new segments that will result from splitting.

This speeds things up for very complicated frames by a couple orders of
magnitude.

### DIFF
--- a/src/libOpenImageIO/deepdata.cpp
+++ b/src/libOpenImageIO/deepdata.cpp
@@ -547,7 +547,8 @@ void
 DeepData::insert_samples (int pixel, int samplepos, int n)
 {
     int oldsamps = samples(pixel);
-    set_capacity (pixel, oldsamps + n);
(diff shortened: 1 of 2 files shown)


lgritz:
Updated with a more thorough accounting for splits that even considers the possibility that either of the source images might *start out* with overlapping samples, before it even tries to merge. Which, it turns out, was the case in the problematic test cases.

This reduces by another 1-2 order of magnitude.

So now this really complicated test does the deep merge on my laptop in about 10 seconds.

ben-andersen:
This works extremely well now.  Thank you Larry.

PR #1740, opened 2017-08-08
oiio: fixed move constructor of ParamValue class.

edubois:
## Description

ParamValue move constructor was not copying the internal ParamValue type.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/paramlist.h
+++ b/src/include/OpenImageIO/paramlist.h
@@ -117,6 +117,7 @@ public:
         : m_name(p.m_name), m_nvalues(p.m_nvalues), m_interp(p.m_interp),
           m_copy(p.m_copy), m_nonlocal(p.m_nonlocal)
     {
+        m_type = p.m_type;
         m_data.ptr = p.m_data.ptr;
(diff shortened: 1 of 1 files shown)


edubois:
Please note that this bugfix is an important one. I don't know how ParamValue could have worked before.

src/include/OpenImageIO/paramlist.h
```diff
@@ -117,6 +117,7 @@ class OIIO_API ParamValue {
         : m_name(p.m_name), m_nvalues(p.m_nvalues), m_interp(p.m_interp),
```
lgritz: Let's put it here, where the other initializers are (except for the ones that are part of unions, which is harder to do in the initializer list).

lgritz:
LGTM. Good catch!

I suspect that it went unnoticed because it's the "move" constructor, and the reliance on this C++11 feature is relatively new for us? But yeah, surprised it hasn't made noticeable bugs before.

src/include/OpenImageIO/paramlist.h
lgritz: n/m, I merged as is.

PR #1741, opened 2017-08-08
raw: unpack the pixels when you need them, not when you open

lgritz:
I'm concerned about this breaking raw reads. Looks ok to me and works on my cursory tests, but other stakeholders that rely on the raw format reader should please try this patch on their end and let me know if it's ok!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -296,6 +296,10 @@ Fixes, minor enhancements, and performance improvements:
      images (supported in libraw 0.18 or newer). #1626 (1.8.3/1.7.13)
    * Add "raw:user_sat" configuration attribute to the reader.
      #1666 (1.7.15/1.8.4)
+   * The pixels are now decoded (expensive) only when they are read, not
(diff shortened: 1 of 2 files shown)


lgritz:
I got an email confirmation from the user who suggested this fix, indicating that it works.

I'd love to also hear from @danwexler that this for sure doesn't break his stuff.

lgritz:
Merged

akihiroy:
The vertical raw picture is broken after this commit.

lgritz:
@akihiroy Can you point me to (or send me) an email that exhibits this problem?

akihiroy:
@lgritz The picture width and height looks horizontal orientation, but the actual pixel data looks vertical orientation.
You can reproduce by the last raw file (sony_a3000_20.arw) in http://www.photographyblog.com/reviews/sony_a3000_review/sample_images/

lgritz:
Sorry for the breakage. I think that #1863 will address the problem.

PR #1743, opened 2017-08-15
A variety of warning/buglet fixes revealed by clang-tidy static analysis

lgritz:
Some notable/interesting fixes:

* Marking of ImageInput/ImageOutput subclasses as `final` fixes a no-no that I had overlooked, which is that we usually called close() from within the class destructors. But it's virtual, and it's generally considered unsafe to call a virtual function within a destructor (unless it's a final class) because it might not end up being the version for that level of the class hierarchy.

* alloca(0) has undefined behavior, so we change the OIIO_ALLOCA macros to detect that case and return NULL rather than blindly calling alloca(0).

* Calling defined() within a macro definition itself is apparently not guaranteed to work with all compilers (though in practice it seems ok). 

* I'm starting to favor the idiom of specifying the default initializers of non-static struct/class data members in their declaration, rather than in long lists of initializers in the constructors themselves. It's a compact notation, removes redundancy (or inadvertent mismatches) when you have multiple constructors, and if you rearrange structure member ordering, you don't have to chase down the warnings about initializer lists being in the wrong order.

### DIFF
--- a/src/bmp.imageio/bmp_pvt.h
+++ b/src/bmp.imageio/bmp_pvt.h
@@ -141,7 +141,7 @@ namespace bmp_pvt {
 
 
 
-class BmpInput : public ImageInput {
+class BmpInput final : public ImageInput {
  public:
     BmpInput () { init (); }
     virtual ~BmpInput () { close (); }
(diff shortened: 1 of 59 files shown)


fpsunflower:
LGTM!

PR #1744, opened 2017-08-16
Convert to range-for loops

lgritz:
Suggestions courtesy of clang-tidy.

### DIFF
--- a/src/field3d.imageio/field3doutput.cpp
+++ b/src/field3d.imageio/field3doutput.cpp
@@ -278,8 +278,8 @@ Field3DOutput::put_parameter (const std::string &name, TypeDesc type,
                 Strutil::split (format_list, format_prefixes, ",");
                 format_prefixes_initialized = true;
             }
-            for (size_t i = 0, e = format_prefixes.size();  i < e;  ++i)
-                if (Strutil::iequals (prefix, format_prefixes[i]))
+            for (const auto& f : format_prefixes)
+                if (Strutil::iequals (prefix, f))
                     return false;
         }
     }
--- a/src/fits.imageio/fitsoutput.cpp
+++ b/src/fits.imageio/fitsoutput.cpp
@@ -217,8 +217,8 @@ FitsOutput::create_fits_header (void)
         std::vector<std::string> values;
         if (keyname == "Comment" || keyname == "History" || keyname == "Hierarch") {
             Strutil::split (value, values, m_sep);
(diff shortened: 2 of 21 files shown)


src/field3d.imageio/field3doutput.cpp
```diff
@@ -278,8 +278,8 @@ Field3DOutput::put_parameter (const std::string &name, TypeDesc type,
                 Strutil::split (format_list, format_prefixes, ",");
                 format_prefixes_initialized = true;
             }
-            for (size_t i = 0, e = format_prefixes.size();  i < e;  ++i)
-                if (Strutil::iequals (prefix, format_prefixes[i]))
+            for (const auto & f : format_prefixes)
```
fpsunflower: I prefer `const auto&` - did `clang-tidy` suggest the spacing?

fpsunflower:
Other than the whitespace nit pick -- LGTM!

Very cool to see how many cases it can recognize.

src/field3d.imageio/field3doutput.cpp
lgritz: Yes. In most cases, I just let it be. Though in cases where I chose to go and correct its choice of variable names, I also "fixed" the spacing.

If you feel strongly about it, I can fix them all. But I was prepared to just accept the free work that clang-tidy did for me and not nit pick a slightly different, but equally readable, spacing than I would have used if I'd made the changes by hand.

lgritz:
clang-tidy did a pretty good job. Much of the time, it picks good variable names for you. For example, if you have a loop like

    for (size_t i = 0; i < values.size(); ++i)
        sum += values[i];

it is smart enough to see that the array you are iterating, "values", ends in "s", so it makes it singular like this:

    for (auto & value : values)
        sum += value;

Neat!

But there are some cases where it screws it up:

    for (size_t i = 0; i < meshes.size(); ++i)

will turn into

    for (auto & meshe : meshes)

Oops.  So there were a few cases like that where I went in and edited by hand to pick better names, and in those cases I also fixed any discrepancies with spacing around the `&`.

src/field3d.imageio/field3doutput.cpp
fpsunflower: I think its worth making them all consistent.

lgritz:
Updated with whitespace changes.

fpsunflower:
LGTM! Thanks :)

PR #1745, opened 2017-08-16
Fix regression in environment()

lgritz:
My stupid bug, introduced during a minor refactoring related to the
static analysis from two commits ago. The static analysis was right,
by my fix to an apparent possible null ptr dereference was flawed. I
avoided the null dereference, alright, but broke some other logic in
the process.


fpsunflower:
LGTM!

PR #1746, opened 2017-08-18
Support (and minor fixes) for building with clang 5.0

lgritz:
The most important bit is to disable warnings about deprecated auto_ptr in the Nuke 10 headers.


lgritz:
Updated with the SPI portions reverting a change in directory for where to find llvm 4.0.

fpsunflower:
LGTM

PR #1747, opened 2017-08-19
oiio/master: Use GNUInstallDirs for installation paths

dracwyrm:
Change all hard coded paths to GNUInstallDirs varibles.

Remove variable setting in main CMakeFiles.txt file since
GNUInstallDirs does all this for you.

Signed-off by: Jonathan Scruggs <j.scruggs@gmail.com>

As discussed in the similar PR from OpenShadingLanuage. I can sign the CLA as needed.

The new Python install directory code is just my attempt at a single variable to hold the install path. The idea was to show how GNUInstallDirs can be used for the path. You may be able to think of a better way. None of the prefixing needs to be done that the old way has since those prefixes are set by CMake, even the exec ones. Let me know your thoughts and I can update this accordingly.

I can't test the python module as the boost-python libs are not being found. There was this issue with another program that we fixed, but I can't remember how. I'll look at the detection code later. This has nothing to do with the new install path code I have, but the detection routines. This was always an issue with OpenImageIO.

### DIFF
--- a/src/cmake/install.cmake
+++ b/src/cmake/install.cmake
@@ -1,56 +1,30 @@
 ###########################################################################
-# Paths for install tree customization.  Note that relative paths are relative
-# to CMAKE_INSTALL_PREFIX.
-set (DEFAULT_BIN_INSTALL_DIR   "bin")
-set (DEFAULT_LIB_INSTALL_DIR   "lib")
-set (DEFAULT_INCLUDE_INSTALL_DIR "include/${PROJECT_NAME}")
+# Set install paths for the python modules
+# TODO: Figure out how to get the correct python directory
+
 if (UNIX AND NOT SELF_CONTAINED_INSTALL_TREE)
-    # Try to be well-behaved and install into reasonable places according to
-    # the "standard" unix directory heirarchy
     # TODO: Figure out how to get the correct python directory
-    set (DEFAULT_PYLIB_INSTALL_DIR "lib/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/site-packages")
-    set (DEFAULT_PYLIB3_INSTALL_DIR "lib/python3/site-packages")
-    set (DEFAULT_DOC_INSTALL_DIR "share/doc/${PROJECT_NAME}")
-    set (DEFAULT_MAN_INSTALL_DIR "share/man/man1")
-    set (DEFAULT_FONTS_INSTALL_DIR "share/fonts/${PROJECT_NAME}")
+    set (DEFAULT_PYLIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/python/site-packages")
+    set (DEFAULT_PYLIB3_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/python3/site-packages")
 else ()
     # Here is the "self-contained install tree" case: the expectation here
     # is that everything related to this project will go into its own
     # directory, not into some standard system heirarchy.
     set (DEFAULT_PYLIB_INSTALL_DIR "python")
     set (DEFAULT_PYLIB3_INSTALL_DIR "python3")
-    set (DEFAULT_DOC_INSTALL_DIR "doc")
-    set (DEFAULT_MAN_INSTALL_DIR "doc/man")
-    set (DEFAULT_FONTS_INSTALL_DIR "fonts/${PROJECT_NAME}")
 endif ()
 if (EXEC_INSTALL_PREFIX)
     # Tack on an extra prefix to support multi-arch builds.
-    set (DEFAULT_BIN_INSTALL_DIR   "${EXEC_INSTALL_PREFIX}/${DEFAULT_BIN_INSTALL_DIR}")
-    set (DEFAULT_LIB_INSTALL_DIR   "${EXEC_INSTALL_PREFIX}/${DEFAULT_LIB_INSTALL_DIR}")
     set (DEFAULT_PYLIB_INSTALL_DIR "${EXEC_INSTALL_PREFIX}/${DEFAULT_PYLIB_INSTALL_DIR}")
     set (DEFAULT_PYLIB3_INSTALL_DIR "${EXEC_INSTALL_PREFIX}/${DEFAULT_PYLIB3_INSTALL_DIR}")
-    set (DEFAULT_FONTS_INSTALL_DIR "${EXEC_INSTALL_PREFIX}/${DEFAULT_FONTS_INSTALL_DIR}")
 endif ()
 # Set up cmake cache variables corresponding to the defaults deduced above, so
 # that the user can override them as desired:
-set (BIN_INSTALL_DIR ${DEFAULT_BIN_INSTALL_DIR} CACHE STRING
-     "Install location for binaries (relative to CMAKE_INSTALL_PREFIX or absolute)")
-set (LIB_INSTALL_DIR ${DEFAULT_LIB_INSTALL_DIR} CACHE STRING
-     "Install location for libraries (relative to CMAKE_INSTALL_PREFIX or absolute)")
 set (PYLIB_INSTALL_DIR ${DEFAULT_PYLIB_INSTALL_DIR} CACHE STRING
      "Install location for python libraries (relative to CMAKE_INSTALL_PREFIX or absolute)")
 set (PYLIB3_INSTALL_DIR ${DEFAULT_PYLIB3_INSTALL_DIR} CACHE STRING
      "Install location for python3 libraries (relative to CMAKE_INSTALL_PREFIX or absolute)")
-set (INCLUDE_INSTALL_DIR ${DEFAULT_INCLUDE_INSTALL_DIR} CACHE STRING
-     "Install location of header files (relative to CMAKE_INSTALL_PREFIX or absolute)")
-set (DOC_INSTALL_DIR ${DEFAULT_DOC_INSTALL_DIR} CACHE STRING
-     "Install location for documentation (relative to CMAKE_INSTALL_PREFIX or absolute)")
-set (FONTS_INSTALL_DIR ${DEFAULT_FONTS_INSTALL_DIR} CACHE STRING
-     "Install location for fonts (relative to CMAKE_INSTALL_PREFIX or absolute)")
-if (UNIX)
-    set (MAN_INSTALL_DIR ${DEFAULT_MAN_INSTALL_DIR} CACHE STRING
-         "Install location for manual pages (relative to CMAKE_INSTALL_PREFIX or absolute)")
-endif()
+
 set (PLUGIN_SEARCH_PATH "" CACHE STRING "Default plugin search path")
 
 set (INSTALL_DOCS ON CACHE BOOL "Install documentation")
@@ -69,10 +43,7 @@ if (CMAKE_SKIP_RPATH)
     set (CMAKE_SKIP_RPATH FALSE)
     unset (CMAKE_INSTALL_RPATH)
 else ()
-    set (CMAKE_INSTALL_RPATH "${LIB_INSTALL_DIR}")
-    if (NOT IS_ABSOLUTE ${CMAKE_INSTALL_RPATH})
-        set (CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${LIB_INSTALL_DIR}")
-    endif ()
+    set (CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}")
     # add the automatically determined parts of the RPATH that
     # point to directories outside the build tree to the install RPATH
     set (CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
@@ -80,20 +51,3 @@ else ()
         message (STATUS "CMAKE_INSTALL_RPATH = ${CMAKE_INSTALL_RPATH}")
     endif ()
 endif ()
-
-
-
-# Macro to install targets to the appropriate locations.  Use this instead of
-# the install(TARGETS ...) signature.
-#
-# Usage:
-#
-#    install_targets (target1 [target2 ...])
-#
-macro (install_targets)
-    install (TARGETS ${ARGN}
-             RUNTIME DESTINATION "${BIN_INSTALL_DIR}" COMPONENT user
-             LIBRARY DESTINATION "${LIB_INSTALL_DIR}" COMPONENT user
-             ARCHIVE DESTINATION "${LIB_INSTALL_DIR}" COMPONENT developer)
-endmacro ()
-
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -644,20 +644,27 @@ resolve_font (int fontsize, string_view font_, std::string &result)
         if (systemRoot.size())
             font_search_dirs.push_back (std::string(systemRoot) + "/Fonts");
         font_search_dirs.emplace_back ("/usr/share/fonts");
+        font_search_dirs.emplace_back ("/usr/share/fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/Library/Fonts");
+        font_search_dirs.emplace_back ("/Library/Fonts/OpenImageIO");
         font_search_dirs.emplace_back ("C:/Windows/Fonts");
+        font_search_dirs.emplace_back ("C:/Windows/Fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/usr/local/share/fonts");
+        font_search_dirs.emplace_back ("/usr/local/share/fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/opt/local/share/fonts");
+        font_search_dirs.emplace_back ("/opt/local/share/fonts/OpenImageIO");
         // Try $OPENIMAGEIOHOME/fonts
         string_view oiiohomedir = Sysutil::getenv ("OPENIMAGEIOHOME");
         if (oiiohomedir.size())
             font_search_dirs.push_back (std::string(oiiohomedir) + "/fonts");
+            font_search_dirs.push_back (std::string(oiiohomedir) + "/share/fonts/OpenImageIO");
         // Try ../fonts relative to where this executing binary came from
         std::string this_program = OIIO::Sysutil::this_program_path ();
         if (this_program.size()) {
             std::string path = Filesystem::parent_path (this_program);
             path = Filesystem::parent_path (path);
             font_search_dirs.push_back (path+"/fonts");
+            font_search_dirs.push_back (path+"/shared/fonts/OpenImageIO");
         }
     }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,8 @@ if (CMAKE_USE_FOLDERS)
     set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 endif ()
 
+include (GNUInstallDirs)
+
 set (CMAKE_MODULE_PATH
      "${PROJECT_SOURCE_DIR}/src/cmake/modules"
      "${PROJECT_SOURCE_DIR}/src/cmake")
@@ -193,7 +195,9 @@ if (USE_PYTHON3 AND boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
 endif ()
 
 add_subdirectory (src/include)
-add_subdirectory (src/doc)
+if (INSTALL_DOCS)
+    add_subdirectory (src/doc)
+endif ()
 add_subdirectory (src/fonts)
 add_subdirectory (src/nuke)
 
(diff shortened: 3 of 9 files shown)


lgritz:
Getting back to this now, sorry for the delay.

First, I propose the following minor changes to fix problems finding the fonts for some tests:

```
diff --git a/src/fonts/CMakeLists.txt b/src/fonts/CMakeLists.txt
index 02afec1b..1ea7a843 100644
--- a/src/fonts/CMakeLists.txt
+++ b/src/fonts/CMakeLists.txt
@@ -2,7 +2,7 @@ file (GLOB public_fonts "*/*.ttf")

 if (INSTALL_FONTS AND USE_FREETYPE)
     install (FILES ${public_fonts}
-             DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/fonts/oiio
+             DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/fonts/OpenImageIO
              COMPONENT fonts
 )
 endif ()
diff --git a/src/libOpenImageIO/imagebufalgo_draw.cpp b/src/libOpenImageIO/imagebufalgo_draw.cpp
index 4b8e48ec..24669bbe 100644
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -644,14 +644,17 @@ resolve_font (int fontsize, string_view font_, std::string &result)
         if (systemRoot.size())
             font_search_dirs.push_back (std::string(systemRoot) + "/Fonts");
         font_search_dirs.emplace_back ("/usr/share/fonts");
+        font_search_dirs.emplace_back ("/usr/share/fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/Library/Fonts");
         font_search_dirs.emplace_back ("C:/Windows/Fonts");
         font_search_dirs.emplace_back ("/usr/local/share/fonts");
+        font_search_dirs.emplace_back ("/usr/local/share/fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/opt/local/share/fonts");
+        font_search_dirs.emplace_back ("/opt/local/share/fonts/OpenImageIO");
         // Try $OPENIMAGEIOHOME/fonts
         string_view oiiohomedir = Sysutil::getenv ("OPENIMAGEIOHOME");
         if (oiiohomedir.size())
-            font_search_dirs.push_back (std::string(oiiohomedir) + "/fonts");
+            font_search_dirs.push_back (std::string(oiiohomedir) + "/share/fonts/OpenImageIO");
         // Try ../fonts relative to where this executing binary came from
         std::string this_program = OIIO::Sysutil::this_program_path ();
         if (this_program.size()) {
```

You need not update the patch if you don't want; I'm happy to just apply these changes on my end as I push the commits.

I am a little concerned about the movement of the python libraries. Not that there's anything wrong with where you've put them, but I'm worried about breaking people's current installations.

I will open a discussion about this on oiio-dev and see what people think. But if everyone is ok with it, I think this change is fine.

dracwyrm:
@lgritz I need to think about the python thing. There may be a solution because one program finds python and is very clean and elegant code. I was thinking of updating this patch with that, so I'll add in your modifications as well. :) Python is the tricky thing and I wish there was a nice standard, like with GNUInstallDirs as it would save a lot of headaches.

lgritz:
Also, I think I'm not comfortable changing anything about the installation layout for 1.7.

But that's ok, because I'm about to branch for 1.8! Changes are fair game there until we're officially out of beta. But I'd like to resolve this quickly so we can get it in and do the branch.

dracwyrm:
@lgritz Just pushed a new patch that changed the python and boost logic. It's more sane than my previous proposal. :) Diffs really don't handle change in whitespace very well. You will need to look at the file carefully for errors, shouldn't be any, though as I think I have everything ported over from 1.7 work I did.

I added another library detection to boost_python in the form of boost_python-${PYTHON_VERSION} as this fixes detection on Gentoo and installs the library to the correct location: usr/lib64/python3.5/site-packages/OpenImageIO.so

I left the patch separate so it's easier to see the changes and to make changes. This fully compiles on my end.

Understandable about not changing the 1.7 base. I'll close that PR.

src/libOpenImageIO/imagebufalgo_draw.cpp
```diff
@@ -644,14 +644,18 @@ resolve_font (int fontsize, string_view font_, std::string &result)
...
         font_search_dirs.emplace_back ("C:/Windows/Fonts");
         font_search_dirs.emplace_back ("/usr/local/share/fonts");
+        font_search_dirs.emplace_back ("/usr/local/share/fonts/OpenImageIO");
         font_search_dirs.emplace_back ("/opt/local/share/fonts");
+        font_search_dirs.emplace_back ("/opt/local/share/fonts/OpenImageIO");
         // Try $OPENIMAGEIOHOME/fonts
         string_view oiiohomedir = Sysutil::getenv ("OPENIMAGEIOHOME");
         if (oiiohomedir.size())
             font_search_dirs.push_back (std::string(oiiohomedir) + "/fonts");
+            font_search_dirs.push_back (std::string(oiiohomedir) + "/share/fonts/OpenImageIO");
```
lgritz: I think we both missed one more... a few lines below this, 

    font_search_dirs.push_back (path+"/fonts");

should be

    font_search_dirs.push_back (path+"/shared/fonts/OpenImageIO");

right?

src/cmake/externalpackages.cmake
```diff
@@ -107,14 +121,14 @@ else ()
     if (NOT USE_STD_REGEX)
         list (APPEND Boost_COMPONENTS regex)
     endif ()
-    find_package (Boost 1.53 REQUIRED
-                  COMPONENTS ${Boost_COMPONENTS})
+    find_package (Boost REQUIRED COMPONENTS ${Boost_COMPONENTS})
```
lgritz: What's up with this change? Don't we need that "1.53" to enforce the minimum Boost version we intend?

lgritz:
I'm afraid I don't understand the purpose or workings of the additional changes you just pushed. What did you do with the Python stuff, exactly? A bunch of the Python 2 vs 3 logic changed. With your new code, how does it know whether to build Python2 or 3 support?

lgritz:
At this point, I'm kind of wondering -- though, tell me if this doesn't even make logical sense -- whether maybe we should split this up into multiple PRs? In my mind, there are several logically separate things you're trying to do: (1) Use GNUInstallDirs to simplify the installation parts of the scripts and make the install directory structure more standard (including putting docs and fonts in $prefix/share); (2) Possibly move the installation of the python libs from $prefix/python to $prefix/lib/python/site-packages; (3) Fix some problems related to boost python? (4) Change the Python 2/3 build logic (in ways I don't fully understand).

Which of these can be teased apart and reviewed/merged separately? Some seem like easier changes to achieve consensus on and test than others.

lgritz:
So here is how I see this:

First, we need to make a decision about what changes we would like to make in the directory layout of the install area (I think there are three potential issues: docs, fonts, and python module).

It's very important that we do this right away, because my original goal was to have September 1 as the date to branch and declare beta. Ideally, any change like this would be before the branch, but I can accept it being within the following several days for the "next beta." But there is some point fairly soon -- a later beta or certainly when we start calling it a "release candidate" -- in which I'm no longer going to accept any breaking changes (except in the new master, which will be the *next* major release).

So it's important to make the directory layout decision immediately and implement it as expediently as possible, in isolation of any other non-breaking changes that we could do later.

THEN at our leisure, we can figure out how to best structure and simplify the CMake scripts or do any other fine tuning that doesn't actually change the location of the installed files.

dracwyrm:
@lgritz All valid points. I'll make this a strict change to GNUInstallDirs and leave all python/boost changes out. I think it's too big to fully see all the changes properly, anyways. GNUInstallDirs is easily tested and doesn't change logic as it's just changing the directory name from a hard-coded name to a variable.

I'll create another PR to add in the small fix of making boost_python detectable on some other systems. That's the part in adding python detection to externalpackages.cmake so I can populate the python version variable and call find_library with this:
NAMES boost_python boost_python-${PYTHON_VERSION}
Which will search for boost_python, or boost_python-3.5 (for example), as Gentoo has this structure:
```
/usr/lib64/libboost_python-2.7-mt.so
/usr/lib64/libboost_python-2.7.so
/usr/lib64/libboost_python-2.7.so.1.63.0
/usr/lib64/libboost_python-3.5-mt.so
/usr/lib64/libboost_python-3.5.so
/usr/lib64/libboost_python-3.5.so.1.63.0
```
Looking at that, that reminds me, multi-threaded boost is turned on, so shouldn't we be using the -mt versions of boost_python?

The python will be the hardest and will require many refinements. :) The logic I used in this patch is based on what Partio did. It detects what python you have via:
```
EXECUTE_PROCESS(COMMAND python -c "import sys;print('%s.%s'%sys.version_info[0:2])"
 +                OUTPUT_VARIABLE PYTHON_VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
```
Then it uses python boost and python of that version to compile, by using the ${PYTHON_VERSION} variable.

When it comes to python, do you want to create both a 2.7 and 3.x library if both are there? Some systems could have multiple 3.x versions installed, and with CMake it is possible to use all versions of python installed, like with boost above, it installs 2.7 and 3.5, I could have 3.6 installed and it will install to all three versions on my system.

Cheers.

dracwyrm:
I left python the way it was save for the adding ${CMAKE_INSTALL_LIBDIR}. The original variables are there along with the original note. This patch should be more easily tested with not so many drastic changes.

lgritz:
I appreciate your patience with this.

OK, so what we have now in this PR unclutters the scripts a bit, but the only externally-visible effect is that it moves the installation location of the docs to from $prefix/doc to $prefix/share/doc/OpenImageIO, and the fonts from $prefix/fonts to $prefix/share/fonts/OpenImageIO. Correct?

OK, that LGTM pending my double checking it on the systems I have access to, then I will merge it.

So, next item on the immediate agenda: finalizing any changes to the installation location of the python modules. We can take that work up in a second PR.

All the other changes, including fine points of Python2/3 (do we want the option to build both, for example?) we can tackle at our leisure, I don't think they are things I would be bothered by changing after the beta branching has occurred.

lgritz:
On my work machine (CentOS7), these changes end up making the installation for the libraries be $prefix/lib64 instead of $prefix/lib.  Presumably this is GNU conventions?

Is anybody concerned about this? 

At work, we have some build customizations, so I can easily switch it back to `lib` just for us. But I'm worried about others.

Is there anything wrong with keeping things in /lib in all cases? That sure does make it easier to document! Does anybody really still have installations with a mix of 32 bit and 64 bit libraries that needs separate lib directories?

dracwyrm:
With the original variables, the values were set at:
set (DEFAULT_DOC_INSTALL_DIR "share/doc/${PROJECT_NAME}")
set (DEFAULT_FONTS_INSTALL_DIR "share/fonts/${PROJECT_NAME}")

CMAKE_INSTALL_DOCDIR is the same as what use to be there, but set by the system, so if you were to change the project name, then DOCDIR would change. Linux distros usually have the predefined list of how these variables are set, so where Redhat likes its docdir may be different than debian, so this will install to the systems preferred place with the name of the project subdir.

With fonts, I just put in DATAROOTDIR for share, since that can change on systems. I overlooked/forgot the ${PROJECT_NAME} variable when I first did this, but could change OpenImageIO to the variable if you want.

When it comes to libdir, sometimes lib is symlinked to lib64 or lib32. However, if you have a 64 bit system and compiling it as 64 bit, it is correct to install it to lib64, since lib symlinks to it, then nothing has changed. On your system, you should be able to ls /usr/lib/libOpenImageIO* and see all the libs, even though it was installed to lib64 (if you have the symlink).

However, the issues are that if you compile it as 32 bit on a 64 bit system and the paths are hardcoded to lib, then it will install to lib, which is really lib64. This means that a 32 bit library was installed to a 64 bit directory. This is hugely bad. Also, if you try to do a multilib install, as in install both the 64 and 32 bit versions, then this would be impossible with a hard coded lib destination. GNUInstallDirs will automatically switch to the correct lib install directory. This is the reason it is now forbidden on Gentoo to install to just lib, it must be lib64 or lib32.

Hope this helps.

SoapGentoo:
>Is anybody concerned about this?
Yes, and the way it is (i.e. with this PR), **is a good thing**. Distros need to be able to change the libdir. **Installing into `lib/` unconditionally is wrong and breaks a ton of systems** and requires us to hack build systems. By using `GNUInstallDirs`, you make the destinations customisable, and everyone wins.

hobbes1069:
Just leaving my $0.02...

Yes, /usr/lib64 is correct for x86_64 builds for pretty much all RPM based linux distros... Fedora/RHEL/CentOS/SL and as far as I know Suse....

src/cmake/install.cmake
```diff
@@ -1,56 +1,13 @@
...
-endif ()
-# Set up cmake cache variables corresponding to the defaults deduced above, so
-# that the user can override them as desired:
-set (BIN_INSTALL_DIR ${DEFAULT_BIN_INSTALL_DIR} CACHE STRING
-     "Install location for binaries (relative to CMAKE_INSTALL_PREFIX or absolute)")
-set (LIB_INSTALL_DIR ${DEFAULT_LIB_INSTALL_DIR} CACHE STRING
-     "Install location for libraries (relative to CMAKE_INSTALL_PREFIX or absolute)")
+# Set install paths for the python modules
+# TODO: Figure out how to get the correct python directory
+set (DEFAULT_PYLIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/site-packages")
```
lgritz: I think this line is problematic because PYTHONLIBS_VERSION_MAJOR (and MINOR) are not defined until it reads src/python/CMakeLists.txt, which is after this. So this ends up installing the python modules in a directory named `$prefix/lib/python.` (yes, with the `.` in the path), whereas it used to go in `$prefix/python`. I know that's "wrong" and we want to fix it, but let's confine this review to only using the GNUInstallDirs and moving the locations of fonts and docs (which can't really "break" anything important), and not change anything about the Python location at all, which people need to be prepared for as a breaking change.

lgritz: Maybe, just for this checkin, we change this to

    set (DEFAULT_PYLIB_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/python")

so that will keep things exactly as they were for this checkin, and then in the other review we can fix all the python stuff once and for all?

dracwyrm: This was the original setting in the original code, just changed lib to CMAKE_INSTALL_LIBDIR:
https://github.com/OpenImageIO/oiio/blob/master/src/cmake/install.cmake#L11
Did this work before? I didn't even notice this when I made the change, so this is a really good catch. Was this a bug then?

Would it be good to use set (DEFAULT_PYLIB_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/python${PYTHON_VERSION}/site-packages") since it's set at compile time? Or not?

In the new PR, I have a way around this. :)

dracwyrm: I looked at the 1.7 source and it's this:
set (DEFAULT_PYLIB_INSTALL_DIR "lib/python/site-packages")
So, I'll change it to that since that's the old default.

dracwyrm:
I updated the line so it resembles the way it was in 1.7:
set (DEFAULT_PYLIB_INSTALL_DIR "lib/python/site-packages")

lgritz:
SO close. And I see now why some things happened the way they did. There was actually a bit of logic in the old install.cmake that did this:

```
if (UNIX AND NOT SELF_CONTAINED_INSTALL_TREE)
    # Try to be well-behaved and install into reasonable places according to
    # the "standard" unix directory heirarchy
    # TODO: Figure out how to get the correct python directory
    set (DEFAULT_PYLIB_INSTALL_DIR "lib/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/site-packages")
    set (DEFAULT_PYLIB3_INSTALL_DIR "lib/python3/site-packages")
    set (DEFAULT_DOC_INSTALL_DIR "share/doc/${PROJECT_NAME}")
    set (DEFAULT_MAN_INSTALL_DIR "share/man/man1")
    set (DEFAULT_FONTS_INSTALL_DIR "share/fonts/${PROJECT_NAME}")
else ()
    # Here is the "self-contained install tree" case: the expectation here
    # is that everything related to this project will go into its own
    # directory, not into some standard system heirarchy.
    set (DEFAULT_PYLIB_INSTALL_DIR "python")
    set (DEFAULT_PYLIB3_INSTALL_DIR "python3")
    set (DEFAULT_DOC_INSTALL_DIR "doc")
    set (DEFAULT_MAN_INSTALL_DIR "doc/man")
    set (DEFAULT_FONTS_INSTALL_DIR "fonts/${PROJECT_NAME}")
endif ()
```

So you see, there was this magic variable `SELF_CONTAINED_INSTALL_TREE` that when true, would put the python module in in `$prefix/python` rather than the usual `$prefix/lib/python2.7/site-packages`.  And similarly, that put docs in `$prefix/doc` and fonts in `$prefix/fonts` instead of `share/...`.

And guess what? The `make` wrapper that I like to use happens to set SELF_CONTAINED_INSTALL_TREE. So that explains why I was seeing and used to a certain directory layout. And somebody who directly used `cmake` instead (let me guess -- you did?) was always seeing things in share and site-packages.

So here's what I propose, and I'm happy to do it on my end when I merge, you don't have to make the change if you won't want to bother.

For this checkin only, let's restore the logic (just for python, don't bother with fonts and doc) that puts things right in `$prefix/python` if `SELF_CONTAINED_INSTALL_TREE` is true. So that way, for this checkin, the python module will not change for anybody.

Then, we'll clean it all up, including giving up on the SELF_CONTAINED nonsense, for the big python PR.

OK?

dracwyrm:
I can add that back in, but with the minor bug fix of removing the MAJOR and MINOR variables. 

Gentoo used the self contained section even without the makefile you mentioned. Originally, when I changed everything, I changed this section only:
```if (UNIX AND NOT SELF_CONTAINED_INSTALL_TREE)
    # Try to be well-behaved and install into reasonable places according to
    # the "standard" unix directory heirarchy
    # TODO: Figure out how to get the correct python directory
    set (DEFAULT_PYLIB_INSTALL_DIR "lib/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/site-packages")
    set (DEFAULT_PYLIB3_INSTALL_DIR "lib/python3/site-packages")
    set (DEFAULT_DOC_INSTALL_DIR "share/doc/${PROJECT_NAME}")
    set (DEFAULT_MAN_INSTALL_DIR "share/man/man1")
    set (DEFAULT_FONTS_INSTALL_DIR "share/fonts/${PROJECT_NAME}")
else ()
```
Changing all the lib's to LIBDIR and share's to DATAROOTDIR, however, it was still installing to lib instead of lib64, it was only until I changed the "else" part of it did it actually install right, but then the "if" part and "else" part were exactly the same.

That's why I ended up removing the if/else, and since Gentoo uses self contained build which installed fine, I declared it working. I was overriding the DEFAULT_PYLIB_INSTALL_DIR variable on my end, so I never caught the "python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}" bug.

dracwyrm:
I think I added it back in right:
https://github.com/OpenImageIO/oiio/pull/1747/files#diff-fd579dba3beb2af8589601fa07c59010R5

Also, you technically don't need to give up on self contained, that's what GNUInstallDirs variables are for. :) David (SoapGentoo) knows more about it than me, he's my mentor on CMake and knows a lot about this stuff, so I run my patches by him before posting here or other upstreams.

lgritz:
Sorry for all the trouble. I'll check this all out and merge if it seems to work ok, then we can move on.

It's just this one checkin that I didn't want anything to change where the python module went. For the separate PR that's about changing Python, I'm totally wiling to accept that we change the install location (assuming we get consensus).

I don't recall writing the self-contained stuff, that must have been somebody else, and I honestly didn't think about it much. So I am not sure I have much invested in keeping it, as long as everybody knows the change is coming and can pinpoint which PR it is where things change.

lgritz:
YES, LGTM!

I will merge tomorrow -- I want to be at work to triple check that nothing breaks there (I have a patch I need to make on top of it, just for SPI, to avoid the change in lib directory name because it will break a bunch of our scripts).

dracwyrm:
@lgritz Is there no symlink from lib to lib64? Though, a lot of systems are removing the symlink now and just going with lib64 or lib32 or just lib on a pure system. If your work system has a symlink, then nothing should break.

My system:
```
# ls -ld /usr/lib*
lrwxrwxrwx   1 root root      5 Sep 17  2015 /usr/lib -> lib64
drwxr-xr-x   6 root root  12288 Aug 21 12:42 /usr/lib32
drwxr-xr-x 188 root root 229376 Sep  2 20:58 /usr/lib64
drwxr-xr-x  21 root root   4096 Sep  2 20:29 /usr/libexec
```

lgritz:
OK, checks out. Merging!

PR #1751, opened 2017-08-24
IC/TS: Improve error messages for broken files.

lgritz:
If a file is "broken", remember the error message that caused us to
consider it broken, so it can be properly reported later.

This is particularly helpful for files that are rejected for being
untiled or not mip-mapped (if "accept_untiled" or "accept_unmipped" are
false).  It's confusing to see "invalid file" as if the image file
itself was corrupted, much more helpful to see a specific error message
about the reason why it was rejected.


fpsunflower:
LGTM!

PR #1755, opened 2017-08-28
Benchmark utilities

lgritz:
Move "benchmarking" utilities -- time_trial, timed_thread_wedge,  DoNotOptimize, clobber_all_memory -- to a separate benchmark.h  and benchmark.cpp. The stuff in timer.h is JUST the Timer and  directly related material.

New Benchmarker class tries to do micro-benchmarking more cleverly.  Wrap up some stuff we had done elsewhere (often redundantly).  Benchmark multiple trials, report the statistics. Select number of  iterations automatically (!), also try to correctly compute and  subtract all the loop overhead.

Read the comments in benchmark.h for  explanation, but to sum up with an example:

      Benchmarker bench;
      float val = 0.5f;   clobber (val);  // Scrub compiler's knowledge of the value
      bench ("acos", [&](){ DoNotOptimize(std::acos(val)); });
      bench ("fast_acos", [&](){  // alternate indentation style
          DoNotOptimize(OIIO::fast_acos(val));
      });

Which produces output like this:

    acos      :    4.3 ns,  230.5 M/s (10x2097152, sdev=0.4ns rng=31.2%, med=4.6)
    fast_acos :    3.4 ns,  291.2 M/s (10x2097152, sdev=0.4ns rng=33.0%, med=3.4)


fpsunflower:
Very neat - how about putting the stddev in ns closer to the timing?

For example: `4,3ns +/- 0.4ns`

Also - do you have an example of how one would write a test that measures a range of values?

lgritz:
Sure. Do you think people will be confused about whether it's stddev or range or what?

What do you mean by "test a range of values"?

lgritz:
PS. This was my "road trip project" last week.

lgritz:
For anybody who uses this -- when timing fairly simple code sequences (especially when it boils down to just one or two instructions), it is really tricky to ensure that the compiler doesn't optimize it away completely. You have to watch out for the "invalid" printouts that it gives when the time is too small, that usually is an indication that something optimized away. Sometimes you need to fool around with DoNotOptimize, clobber, or other tricks.

I don't think this is an inherent flaw in Benchmarker -- I did a spot check of Google Benchmark and Hayai, and get similar results (both in the timing results, as well as in how finicky they are to ensure that you're timing what you think you're timing).

fpsunflower:
You could draw a column header maybe to help clarify? And then maybe the other entries wouldn't have to be rendered as `name=val`.

Let's say I want to benchmark `fast_acos` for all inputs from -1 to 1 (or random values in that range), how do I write that? If I'm reading correctly, the code you posted just repeatedly tests `fast_acos(0.5)`.

lgritz:
Header -- I guess? The problem is that if the test names get really long, they won't all line up any more.

You can always set verbose=0 and do all your own output of the results.

To benchmark fast_acos for a bunch of inputs, for example, you could do something like this. I'm assuming that you don't want millions of output lines, so in this example I am just gathering the fastest and slowest times, and also setting a fixed (and fairly small) number of iterations, because the auto iteration count will be too long for "all the values".

    Benchmarker bench;
    bench.verbose (0);   // suppress console output
    bench.iterations (1000);  // less accurate but much faster than auto iterations
    double slowest = 0, slowest = 1e30;
    for (float f = minval; f <= maxval; f += step) {
        bench (Strutil::format("fast_acos(%f)", f),
                    [&](){ DoNotOptimize(fast_acos(f)); });
        slowest = std::min (slowest, bench.avg());
        fastest = std::max (fastest, bench.avg());
    }
    std::cout << "fast_acos ranged from " << slowest << " to " << fastest << " depending on input.\n";

lgritz:
Or, did you mean that you wanted ONE benchmark, which cycles through input values rather than using the same value for all the iterations? For that, I might do something like:

    float f = -1.0;
    bench ("fast_acos", [&](){ DoNotOptimize(fast_acos(f)); f += step; });

But then, you are timing the increment as well as the acos.

lgritz:
Oops, that doesn't quite work, does it? You would want `f` to be reset to -1 for each "trial", and also for `step` to be set specifically to 2.0/bench.iterations().

I think to do this right, we'd want an option to have a "setup" procedure for each trial, what would not be part of the timing. I thought of this kind of thing, but figured it was one of many things we may want to add to this class later. Small steps.

lgritz:
I pushed an amendment that refines the output.  New default format for default verbosity looks like:

    cos             :    6.1 ns (+/- 0.5ns),  163.4 M/s

where the `+/-` number is the standard deviation, as Chris suggested.

I also changed it so that the trials/iterations, range, and median are printed only if verbose >= 2.

fpsunflower:
LGTM

PR #1757, opened 2017-08-30
Fix statics contributing to static initialization order fiasco

### LINKED ISSUE #1750: maketx relies on initialization order of static variables when OCIO support is enabled
### Problem

OIIO relies on initialization order of static variables when OCIO support is enabled.

First static variable (`maketx.cpp` line 73):

    static ColorConfig colorconfig;

This leads to a call to `ColorConfig::Impl::inventory()`:

    void
    ColorConfig::Impl::inventory ()
    {
    #ifdef USE_OCIO
        if (config_) {
            bool nonraw = false;
            for (int i = 0, e = config_->getNumColorSpaces();  i < e;  ++i)
                nonraw |= ! Strutil::iequals(config_->getColorSpaceNameByIndex(i), "raw");
            // ...
        }
        // ...
    #endif
        // ...
    }

`Strutil::iequals()` calls `boost::algorithm::iequals()` with `loc` as 3rd argument. `loc` is the second static variable (`strutil.cpp` lines 303-305):

    namespace {
    static std::locale loc = std::locale::classic();
    }

Minimal standalone repro case that exhibits the problem:

    #include <locale>

    extern std::locale loc;

    inline bool iequals(const std::locale& Loc = std::locale()) {
        return true;
    }

    struct C {
        C()     {
            iequals(loc);
        }
    };

    static C c;

    static std::locale loc = std::locale::classic();

    int main() {
        return 0;
    }

### Versions

* OIIO branch/version: 1.7.10
* OS: Windows 10 64-bit
* C++ compiler: Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24213.1 for x64

lgritz:
Fixes #1750

These two spots were implicated in the initialization order fiasco. (See: https://isocpp.org/wiki/faq/ctors#static-init-order )

The solution for maketx.cpp is that since we only are using the ColorConfig for the sake of generating a help message, there's no reason it can't be an ordinary local variable in the one (rarely called) function where it's used.

For the std::locale involved in strutil.cpp, use the old trick where instead of a file static, we have a function returning a reference, which itself has a function-static pointer that's allocated the first time it's needed. This leaks (a single std::locale), but who cares?

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -338,6 +338,12 @@ static std::locale& loc ()
     // inside the function that is allocated the first time it's needed. It
     // will "leak" in the sense that this one locale is never deleted, but
     // so what?
+#if OIIO_MSVS_BEFORE_2015
+    // MSVS prior to 2015 could not be trusted to make in-function static
+    // initialization thread-safe, as required by C++11.
+    static spin_mutex mutex;
+    spin_lock lock (mutex);
+#endif
     static std::locale *loc = new std::locale (std::locale::classic());
     return *loc;
 }
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -330,15 +330,24 @@ Strutil::wordwrap (string_view src, int columns, int prefix)
 
 
 
-namespace {
-static std::locale loc = std::locale::classic();
+
+static std::locale& loc ()
+{
+    // Rather than have a file-level static locale which could suffer from
+    // the static initialization order fiasco, make it a static pointer
+    // inside the function that is allocated the first time it's needed. It
+    // will "leak" in the sense that this one locale is never deleted, but
+    // so what?
+    static std::locale *loc = new std::locale (std::locale::classic());
+    return *loc;
 }
+
 
 
 bool
 Strutil::iequals (string_view a, string_view b)
 {
-    return boost::algorithm::iequals (a, b, loc);
+    return boost::algorithm::iequals (a, b, loc());
 }
 
 
@@ -352,7 +361,7 @@ Strutil::starts_with (string_view a, string_view b)
 bool
 Strutil::istarts_with (string_view a, string_view b)
 {
-    return boost::algorithm::istarts_with (a, b, loc);
+    return boost::algorithm::istarts_with (a, b, loc());
 }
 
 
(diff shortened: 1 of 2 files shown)


src/libutil/strutil.cpp
```diff
@@ -330,15 +330,24 @@ Strutil::wordwrap (string_view src, int columns, int prefix)
...
-static std::locale loc = std::locale::classic();
+
+static std::locale& loc ()
+{
+    // Rather than have a file-level static locale which could suffer from
+    // the static initialization order fiasco, make it a static pointer
+    // inside the function that is allocated the first time it's needed. It
+    // will "leak" in the sense that this one locale is never deleted, but
+    // so what?
+    static std::locale *loc = new std::locale (std::locale::classic());
```
dictoon: Why not simply

    static std::locale loc = std::locale::classic();

?

Moreover, one needs to be aware that this is not thread safe.

lgritz: If it's static rather than a pointer, you can have the same order fiasco upon destruction. This one "leaks" (never frees) the loc, but it will definitely not free it before some other destructor might need it (say, if a destructor for some reason used any of the strutil functions).

Initialization of static variables within functions *is* thread-safe, that's guaranteed by C++11. 

You can see the discussion of this solution here: https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use

dictoon: > If it's static rather than a pointer, you can have the same order fiasco upon destruction.

Ok, hadn't thought about destruction.

> Initialization of static variables within functions is thread-safe, that's guaranteed by C++11

Ok, that's new with C++11 then, I didn't know.

dictoon: Or maybe it was thread-safe all along? https://stackoverflow.com/a/8102145/393756

dictoon: Sorry for the spam, but there are more details here: https://stackoverflow.com/a/1661564/393756

In particular, it looks like it wasn't thread-safe before Visual Studio 2015...

dictoon: So all in all, while I agree with the fix on principle, I just want to reiterate that it probably won't be thread-safe with VS 2012 and VS 2013 (both used quite a lot because they are the mandatory compilers for 3ds Max 2015/2016, and probably some versions of Maya as well).

Note that it's necessary a big deal...

dictoon: Note that it's _not_ necessary a big deal...

lgritz: I'll put in a fix for that...

lgritz:
Amended for thread safety fix on older MSVS.

dictoon:
LGTM!

PR #1760, opened 2017-09-02
Make boost_python and python detection more generic

dracwyrm:
## Description
Currently, there is redundant code when it comes to boost_python detection. The code in externalpackages.cmake never worked with Gentoo, however, the code in python's CMakeLists.txt did. I made the code more variable, so it doesn't need to be changed for newer versions of python and package maintainers should be able to change anything they want, or add newer version detection before OpenImageIO is updated. Also, there is redundant code when it comes to python 2 and 3 detection and compiling. CMake provides python version detection, so we can remove the separate code dealing with different versions of python and just use variables.

The boost_python detection was solely moved to src/python/CMakeLists.txt so it is run only once and only if USE_PYTHON is selected.

Since python site-packages directory changes from distro to distro, in fact, it's not even called site-packages on some, I run a python command to get the directory. Then strip the CMake compile directory structure from the path leaving only the correct destination. The final directory can be overridden.

## Tests
Gentoo uses different formatting and this was able to find the boost_python module and install to the correct site-packages directory.

## Checklist:


src/python/CMakeLists.txt
```diff
@@ -58,36 +69,51 @@ if (APPLE)
 #    set (PYTHON_LIBRARIES /opt/local/lib)
 endif ()
 
+# Use python executable to find the site packages dir if not defined
```
dracwyrm: I've been debating this code or just having  default value of:
${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_FOUND}/site-packages

But I think some distros use python3/site-packages? Then some don't even call it site-packages. I was trying to be generic here, but I just don't know how I feel about this bit of code. What do you all think?

dracwyrm: Apparently this is what automake does.

lgritz:
Now that #1747 has been merged, you should rebase this on top of the new master, fix it up, and let's get this all over with including the python overhaul!

CMakeLists.txt
```diff
@@ -72,9 +72,7 @@ option (USE_OPENGL "Include OpenGL support" ON)
 option (USE_QT "Include Qt support" ON)
 option (FORCE_OPENGL_1 "Force iv to use OpenGL's fixed pipeline")
 option (USE_PYTHON "Build the Python bindings" ON)
-option (USE_PYTHON3 "Build the Python3 bindings" OFF)
-set (PYTHON_VERSION 2.6)
-set (PYTHON3_VERSION 3.2)
+set (PYTHON_VERSION "2.7" CACHE STRING "Target version of python to find")
```
dracwyrm: Is min version still 2.6 or now 2.7? But can be changed with a configure flag, anyways.

src/python/CMakeLists.txt
```diff
@@ -1,54 +1,29 @@
...
-string (REGEX MATCH "python3\$" _py3_subdir ${CMAKE_CURRENT_BINARY_DIR})
-
-if (_py3_subdir)
-    set (BUILD_PY3 ON)
-else ()
-    set (BUILD_PY3 OFF)
+# Narrow the search for python to only those supported
+if (NOT DEFINED Python_ADDITIONAL_VERSIONS)
+    # Support newer versions first. Can be overridden.
+    set (Python_ADDITIONAL_VERSIONS "3.7" "3.6" "3.5" "3.4" "2.7")
```
dracwyrm: I'm not 100% positive that this is needed. These values are default in the find package anyways. Also, I think I may need to add CACHE STRING to it so that it shows up when listing all user settable options, then it is know that the python version range can be set manually.

src/python/CMakeLists.txt
```diff
@@ -1,54 +1,29 @@
...
-    unset(PYTHON_LIBRARY)
-    unset(PYTHON_LIBRARY CACHE)
-    unset(PYTHON_INCLUDE_DIR)
-    unset(PYTHON_INCLUDE_DIR CACHE)
-    unset(PYTHON_INCLUDE_PATH)
-    unset(PYTHON_INCLUDE_PATH CACHE)
+# The version that was found may not be the default or user
+# defined one.
+set (PYTHON_VERSION_FOUND ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR})
+find_package (PythonLibs ${PYTHON_VERSION_FOUND} REQUIRED)
```
dracwyrm: Should there be a warning message about the target version and the actual found version being different?

src/python/CMakeLists.txt
```diff
@@ -1,54 +1,29 @@
...
-        if (Boost_${boost_py3_lib_name}_FOUND)
-            #Not the most beautiful thing to do, but that gets them included in
-            #the target_link_libraries(…) call farther down
-            set (Boost_PYTHON_LIBRARIES ${Boost_${boost_py3_lib_name}_LIBRARIES})
+if (NOT BOOST_CUSTOM)
+    # Finding the python component for boost is a little tricky, since it has
+    # different names on different systems. Try the most common ones.
+    foreach (_boost_py_lib python-${PYTHON_VERSION_FOUND} python
+             python${PYTHON_VERSION_MAJOR}
+             python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
```
dracwyrm: Not sure all these are needed, so I left them in there. It might be a good idea to have people with completely different systems to try with a smaller set. I have python-${PYTHON_VERSION_FOUND} for systems with multiple versions of boost python installed, so the same version as python is located.

src/python/CMakeLists.txt
```diff
@@ -58,36 +33,47 @@ if (APPLE)
...
+    set (PYTHON_SITE_DIR "${CMAKE_INSTALL_LIBDIR}/${_py_site_dir}")
+endif ()
+
 # Disable some warnings for Clang, it's a little too picky with boost
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     add_definitions ("-Wno-array-bounds")
 endif ()
 
-if (BUILD_PY3)
+if (${PYTHON_VERSION_MAJOR} EQUAL "3")
```
dracwyrm: Not sure this is needed as the output name of the module is not called this?
usr/lib64/python3.5/site-packages/OpenImageIO.so
So we might just need set (target_name PyOpenImageIO) and that's it?

src/python/CMakeLists.txt
```diff
@@ -58,36 +33,47 @@ if (APPLE)
...
     set (target_name Py3OpenImageIO)
 else ()
     set (target_name PyOpenImageIO)
 endif ()
 
-if (BOOST_CUSTOM OR Boost_FOUND AND PYTHONLIBS_FOUND)
+if (BOOST_CUSTOM OR boost_PYTHON_FOUND)
```
dracwyrm: Boost and Python detection will cause configure to fail if not found. However, boost_python will not fail since it's in quiet mode.

src/python/CMakeLists.txt
```diff
@@ -128,21 +114,14 @@ if (BOOST_CUSTOM OR Boost_FOUND AND PYTHONLIBS_FOUND)
...
+    install (TARGETS ${target_name}
+             RUNTIME DESTINATION ${PYTHON_SITE_DIR} COMPONENT user
+             LIBRARY DESTINATION ${PYTHON_SITE_DIR} COMPONENT user)
+else ()
+    # If Boost python components were not found, turn off all python support.
+    message (STATUS "Boost python support not found!")
+    if (APPLE AND USE_PYTHON)
+        message (STATUS "   If your Boost is from Macports, you need the +python26 variant to get Python support.")
+     endif ()
+     message (FATAL_ERROR "Python module cannot be built. Either disable python support or check your boost installation.")
```
dracwyrm: The user specifically said that they wanted the python module, so not detecting it could mean an issue with the system, so fail the configure phase so they can fix it or disable the python module support.

dracwyrm:
I pushed the latest update. I was able to successfully detect python and boost_python for versions 2.7, 3.4, and 3.5 while all three were installed at the same time. Running configure while changing the desired version works. Boost_python installed on my system in the following way:
```
/usr/lib64/libboost_python-2.7-mt.so
/usr/lib64/libboost_python-2.7.so
/usr/lib64/libboost_python-2.7.so.1.63.0
/usr/lib64/libboost_python-3.4-mt.so
/usr/lib64/libboost_python-3.4.so
/usr/lib64/libboost_python-3.4.so.1.63.0
/usr/lib64/libboost_python-3.5-mt.so
/usr/lib64/libboost_python-3.5.so
/usr/lib64/libboost_python-3.5.so.1.63.0
```
I added a bunch of comments, once those answered and I figure out why appveyor failed, then I can remove the WIP status for a more complete review.

Sorry it took so long. I had a migraine that made looking at a computer screen difficult.

CMakeLists.txt
lgritz: I think probably 2.7 is the proper default minimum, since that has been the case for VFXPlatform since as early as 2014 (this year, it's specifically 2.7.5 minimum). Also note that VFXPlatform is planning a transition to Python 3 in CY 2019.

src/python/CMakeLists.txt
lgritz: Let's take it out, then, and see what happens.

src/python/CMakeLists.txt
lgritz: Probably

src/python/CMakeLists.txt
lgritz: It's *probably* the case that each one of these was from an actual instance of somebody fixing it for some distro or weird case. So unless there's harm in it, I wouldn't change it.

src/python/CMakeLists.txt
lgritz: You mean that automake uses the python executable to find the site packages dir? This strikes me as the most foolproof thing, no?

src/python/CMakeLists.txt
lgritz: Yes, I agree about this being an error.

src/python/CMakeLists.txt
dracwyrm: That's what I thought and left it in. :)

CMakeLists.txt
dracwyrm: I'll leave the minimum as 2.7 then.

src/python/CMakeLists.txt
dracwyrm: Yeah. I was told that automake does something very similar to this to find the site packages dir. I think there is a way to make it more robust, but I need to as Soap about the idea.

src/python/CMakeLists.txt
```diff
@@ -1,54 +1,29 @@
...
-            #Not the most beautiful thing to do, but that gets them included in
-            #the target_link_libraries(…) call farther down
-            set (Boost_PYTHON_LIBRARIES ${Boost_${boost_py3_lib_name}_LIBRARIES})
+if (NOT BOOST_CUSTOM)
+    # Finding the python component for boost is a little tricky, since it has
+    # different names on different systems. Try the most common ones.
+    foreach (_boost_py_lib python-${PYTHON_VERSION_FOUND} python
+             python${PYTHON_VERSION_MAJOR}
+             python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
+        find_package (Boost ${Boost_VERSION} QUIET COMPONENTS ${_boost_py_lib})
```
lgritz: The CI builds are failing on this line.

lgritz:
This is failing the CI tests.

dracwyrm:
Odd. I'm just using the found boost version. I guess it doesn't matter about version since there can only be one boost version installed. The version installed should be the correct one because externalpackages.cmake verified it. I think the version can be removed from this line.

dracwyrm:
Finally it passes the tests! Now I just need to see about a final code cleanup.

dracwyrm:
@lgritz Why is this happening with job #2030.4? The python detection is:
```
-- Python include dirs /usr/include/python2.7
-- Python libraries    /usr/lib/x86_64-linux-gnu/libpython2.7.so
-- Python site packages dir /usr/lib/python2.7/dist-packages
-- Python to include 'lib' prefix: OFF
-- Python to include SO version: OFF
-- Python version 2.7.6
-- Python version major: 2 minor: 7
-- Boost python libraries /usr/lib/x86_64-linux-gnu/libboost_python.so
```

However, this happens:
```
Install the project...
CMake Error at src/python/cmake_install.cmake:50 (file):
  file INSTALL cannot copy file
  "/home/travis/build/OpenImageIO/oiio/build/linux64/src/python/OpenImageIO.so"
  to "/usr/lib/python2.7/dist-packages/OpenImageIO.so".
```
Is it supposed to install to a different directory?

dracwyrm:
@lgritz On those tests, both appveyor and travis, is there a way to see where the files were installed to for verification things are in the right directory?

lgritz:
Something is going wrong here.

When I do a build on my laptop (I use the 'make' wrapper, so it should be "installing" to ./dist/macosx), I end up with the python module being put here:

```
 dist/macosx/local/Cellar/python/2.7.13_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/OpenImageIO.so
```

Yowza! Presumably it's somehow being confused by the fact that it found python itself somewhere in my Homebrew area (/usr/local/Cellar/...), but this should have nothing to do with where we install the OpenImageIO components!

I think your original idea was correct about installing in `${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_FOUND}/site-packages`

lgritz:
Yes, I confirm that replacing that whole clause involving PYTHON_SITE_DIR with the following:

    set (PYTHON_SITE_DIR "${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_FOUND}/site-packages")

works fine for me.

Shall I just merge from my end, with this correction?

dracwyrm:
@lgritz The issue comes in that some distros have dist-packages instead of site-packages, like the travis checks have shown. The other detection that I had might work better. It depends on the output of this command when called from cmake:
`python -m site --user-site`
It should be two directories separated with a colon, for example on my system when called from CMake:
`Full User Site: /var/tmp/portage/media-libs/openimageio-1.7.16/homedir/.local/lib64/python3.5/site-packages`
The regex that I have now cuts off everything before python and then I add ${CMAKE_INSTALL_LIBDIR}. This will, in hope, get the correct path whether or not dist- or site-packages is used, or just python{major_version} is used, or python-py.... so on.

I added in a message that will need to be removed before committing if this works. If you could just configure and look at the message output beginning with "Full User Site:...." and paste that here then I can see if this works. However, if there is more than one "python" in the path, then it might not work and may need to use --user-base as well. But, the path should be the full path that CMake is using to compile, and not the python path.

dracwyrm:
I already noticed that appveyor uses
`C:\Users\appveyor\AppData\Roaming\Python\Python27\site-packages`
hmm. I need to look for a better regex. I need to think.

devurandom:
As a random idea: Would it be possible to filter `sys.path` for those paths ending in `"packages"` and beginning with `sys.prefix`? There is also `site.getsitepackages()`, which might help? How does PIP figure out the install directory? Can it maybe used directly to figure out the correct path?

dracwyrm:
@devurandom That's an idea I need to investigate. :)

@lgritz I pushed through an update which makes the change to the original, original way.
```
if (NOT DEFINED PYTHON_SITE_DIR)
    set (PYTHON_SITE_DIR "${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_FOUND}/site-packages")
endif ()
```
We can always push another update later that makes site/dist dir finding more robust on all systems. I was trying to find a regex that would take the last occurrence of Python and select everything to the end because on the appveyor system, they have Python\Python27\site-packages, so it's a double Python and the 27 doesn't have a '.' character. This is why I'm trying to find it dynamically.

There is one way I thought involves using calls to both python -m site --user-site and python -m site --user-base. I tried it on the appveyor system and it worked, but it's a double python call. This is the test code I used (I know the variables can be reused, but I had messages outputting them):
```
execute_process (COMMAND ${PYTHON_EXECUTABLE} -m site --user-base
                 OUTPUT_VARIABLE _py_full_site_base
                 OUTPUT_STRIP_TRAILING_WHITESPACE)

execute_process (COMMAND ${PYTHON_EXECUTABLE} -m site --user-site
                 OUTPUT_VARIABLE _py_full_site_dir
                 OUTPUT_STRIP_TRAILING_WHITESPACE)

# Remove the base path
string (REPLACE "${_py_full_site_base}" "" _py_partial_site_dir "${_py_full_site_dir}"

# Cut off the temporary compile dir
string (REGEX MATCH "([pP]ython.*)" _py_site_dir ${_py_partial_site_dir})

# Construct proper install path 
set (PYTHON_SITE_DIR "${CMAKE_INSTALL_LIBDIR}/${_py_site_dir}")
```
That is a lot and a bit messy, so I'm hoping to find a better way.

lgritz:
LGTM!

I'm all for a second round of fixes later that can have more robust logic that will handle some of the odd system install locations correctly. 

But for now, this is a solid step forward. People on systems that have a trickier final system-wide installation location can always customize it by using -DPYTHON_SITE_DIR=/custom/dir

(Aside: I think I would like to change that to PYTHON_INSTALL_DIR, but we can do that later.)

Thanks for your patience on this. We've gone round and round quite a few steps more than is usual for a PR.

I will merge as soon as I see the latest round of CI tests finish.

dracwyrm:
I prefer round and round and a lot of steps because it ensures good code upon entering the main tree. :) Thanks for your patience as well. Now, I'll backport these final fixes to 1.7 to put that in Gentoo's Main Repo. I'll put these changes to my fork if anyone else is interested in them for 1.7. I have only GNUInstallDir fixes for 1.6, as I'm putting in the last version of series in case of compatibility issues with other programs, but will last rite it in a year if nothing comes up.

Thanks again.

devurandom:
@dracwyrm Thanks for the Gentoo love!

rtlanceroad:
I kinda like the old way for supporting python3 by adding a separate subdirectory. I have 2.7.14(maintenance release) as dominant and 3.6.2 as alternative, now problem is I can't build oiio for python3 because python is linked to python2.7 by default. how to build oiio for those two versions simultaneously like old school?

dracwyrm:
@rtlanceroad It worked for me by changing the PYTHON_VERSION to the version I wanted. Like -DPYTHON_VERSION="3.6". I have python linked to 3.5.4, but able to install for 2.7 and 3.4.

rtlanceroad:
@dracwyrm I think that's not the answer I want. I mean simultaneously, that is, there will be a `/usr/lib/python2.7/site-packages/libPyOpenImageIO.so` and `/usr/lib/python3.6/site-packages/libPyOpenImageIO.so` if I get 2.7 and 3.6 co-installed. I have built boost with python2(library soname is libboost_python.so.1.65.1) and python3(library soname is libboost_python3.so.1.65.1), then after I build oiio with the following:
```
cmake .. \
		-DCMAKE_BUILD_TYPE=Release \
		-DCMAKE_INSTALL_PREFIX=/usr \
		-DCMAKE_INSTALL_LIBDIR=/usr/lib/ \
		-DUSE_PYTHON=ON \
		-DPYLIB_LIB_PREFIX=/usr/lib \
                -DUSE_OPENSSL=ON
```
I only get oiio with python2's library(/usr/lib/python2.7/site-packages/libPyOpenImageIO.so) and miss python3's. What's wrong there?

Also I check [your PR for gentoo](https://github.com/gentoo/gentoo/pull/5815), I see there is `DUSE_PYTHON3` which is not defined in CMakeLists.txt from oiio top src directory, does that flag work? and could it be auto detected and built for each python3 libraries if multiple python3 libraries(3.2 3.3 3.4 3.5 3.6) and their compatible boost libraries are installed ?

rtlanceroad:
At the time now, I am trying to build oiio 1.8.5.

dracwyrm:
@rtlanceroad Am I reading it correctly that you want both 2.7 and 3.6 versions of the OIIO module installed at the same time?

To install the 3.6 module, did you try:
```
cmake .. \
		-DCMAKE_BUILD_TYPE=Release \
		-DCMAKE_INSTALL_PREFIX=/usr \  ### don't think this line is needed?
		-DCMAKE_INSTALL_LIBDIR=lib/ \
		-DUSE_PYTHON=ON \
                -DUSE_OPENSSL=ON
		-DPYTHON_VERSION="3.6"
```
That's should give you only the 3.6 version. I can make a patch to do both 2.7 and 3.x versions at the same time. I've been wanting to find a way to install for all versions of python installed, because in the old way, you could have 2.7 or one 3.x version. However, in my case I have 3.4 and 3.5, and soon 3.6 will be installed. Then there are testing versions of python. I want to detect all that are on the system to install the module for all of them. I have a theory on how to do this, just need a clean way of doing it.

The code on my repo hasn't been updated. Been working on finalizing the package for Gentoo and work.

rtlanceroad:
> @rtlanceroad Am I reading it correctly that you want both 2.7 and 3.6 versions of the OIIO module installed at the same time?

Exactly. If possible, oiio module for all of them(2.7, 3.2, 3.3, 3.4, 3.5, 3.6) at the same time.

>  I want to detect all that are on the system to install the module for all of them

Your patch is really appreciated if you do so, that's exactly what I want to get.

PR #1765, opened 2017-09-10
CMake: Fix Python detection on Gentoo Linux

devurandom:
## Description

This fixes Python detection on Gentoo Linux

## Tests

Tested by building on Gentoo Linux using [my modified ebuild](https://github.com/devurandom/gentoo-overlay/tree/master/media-libs/openimageio).

## Checklist:

- [ ] N/A: If this is more extensive than a small change to existing code, I
- [ ] N/A: I have updated the documentation, if applicable.
- [ ] N/A: I have ensured that the change is tested somewhere in the testsuite
- [ ] N/A: My code follows the prevailing code style of this project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,10 +185,10 @@ if (NOT EMBEDPLUGINS AND NOT BUILD_OIIOUTIL_ONLY)
     endforeach ()
 endif ()
 
-if (USE_PYTHON AND boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
+if (USE_PYTHON AND Boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
     add_subdirectory (src/python)
 endif ()
-if (USE_PYTHON3 AND boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
+if (USE_PYTHON3 AND Boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
     #build the python3 module in a different binary directory since it will
     #have the same name as the python2 module (e.g. OpenImageIO.so)
     add_subdirectory (src/python src/python3)
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -107,6 +107,10 @@ else ()
     find_package (Boost 1.53 REQUIRED
                   COMPONENTS ${Boost_COMPONENTS})
 
+    if (CMAKE_GENTOO_BUILD OR CMAKE_BUILD_TYPE STREQUAL Gentoo)
+        # Gentoo's patches to CMake make sure that we will always find exactly the version of Python required by Gentoo
+        find_package (Boost 1.53 COMPONENTS ${Boost_COMPONENTS} python)
+    else ()
     # Try to figure out if this boost distro has Boost::python.  If we
     # include python in the component list above, cmake will abort if
     # it's not found.  So we resort to checking for the boost_python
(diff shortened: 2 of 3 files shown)


src/cmake/externalpackages.cmake
```diff
@@ -107,6 +107,10 @@ else ()
     find_package (Boost 1.53 REQUIRED
                   COMPONENTS ${Boost_COMPONENTS})
 
+    if (CMAKE_GENTOO_BUILD OR CMAKE_BUILD_TYPE STREQUAL Gentoo)
+        # Gentoo's patches to CMake make sure that we will always find exactly the version of Python required by Gentoo
+        find_package (Boost 1.42 COMPONENTS ${Boost_COMPONENTS} python)
```
lgritz: I'm a little worried about the designation of boost 1.42 here when we've advertised the current minimum as 1.53. Was this on purpose, or just a cut-and-paste from someplace else where we have a version mismatch?

CMakeLists.txt
```diff
@@ -185,10 +185,10 @@ if (NOT EMBEDPLUGINS AND NOT BUILD_OIIOUTIL_ONLY)
     endforeach ()
 endif ()
 
-if (USE_PYTHON AND boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
+if (USE_PYTHON AND Boost_PYTHON_FOUND AND NOT BUILD_OIIOUTIL_ONLY)
```
lgritz: What's up with the capitalization change? Did we have this wrong all along? Why did it work before on the other platforms? Is it possible that this is something that changed at some point and might be different for other Boost versions?

lgritz:
We're going to need to be really careful about interactions between this and #1760.

src/cmake/externalpackages.cmake
devurandom: This patch was originally against version 1.7.13, but it applied cleanly against current master. Thus I did not notice that the requirement on Boost changed in the meantime. I fixed my patch to also require Boost 1.53.

CMakeLists.txt
devurandom: I believe since you're using `find_package(Boost ...)`, the general pattern is that this module should set `Boost_FOUND` and similar. That is also what the CMake source-code says: https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindBoost.cmake#L27

lgritz: Agreed.

devurandom:
> We're going to need to be really careful about interactions between this and #1760.

I would suggest to apply this PR first, since it is smaller in scope, sticks very closely to the current code (hence it will produce only a small amount of conflicts) and gets the build-system to comply with the way the upstream CMake Boost module works. #1760 would then be rebased onto this.

lgritz:
OK, I will try this out tomorrow and if it doesn't break anything at work, I will merge it. (It already passes the CI tests on Travis.)

dracwyrm:
@lgritz and @devurandom The PR that I have for Boost/Python completely fixes detection on Gentoo (I've tried it out with python versions 2.7, 2.4, 2.5) without having the separate if statement:
```
 if (CMAKE_GENTOO_BUILD OR CMAKE_BUILD_TYPE STREQUAL Gentoo)
      # Gentoo's patches to CMake make sure that we will always find exactly the version of Python required by Gentoo
      find_package (Boost 1.53 COMPONENTS ${Boost_COMPONENTS} python)
else ()
```
The PR I have would actually remove most of what's in this PR as some of those sections do not exist anymore. If you could try PR https://github.com/OpenImageIO/oiio/pull/1760 I would be really grateful, as the more people that test it, the more robust it would be.

Once finalized, this patch will be back ported to 1.7 for the Gentoo ebuilds I'm submitting. :) PR https://github.com/gentoo/gentoo/pull/5384 Will be updated with the latest. I haven't updated it as I'm working on these upstream patches. Also, I will be making the ebuilds multi-lib.

lgritz:
OK, agreed with @dracwyrm, let's tackle #1760 first. If it also works for Gentoo, then this PR can be abandoned.

lgritz:
We merged #1760. Please give that a try; if it fixes things, then you should just close this PR. If it still needs touch-ups for Gentoo, then please rebase on top of the new master and see what changes are needed after Jonathan's changes.

devurandom:
I trust @dracwyrm will take care of Gentoo as he promised.

PR #1768, opened 2017-09-15
Fix IBA::fillholes_pushpull when display window != data window

lgritz:
We pasted the result into the dest to be aligned with the dst's data
origin, instead of the correct source origin. So for cropped, overscanned,
or shrink-wrapped images, it would botch the origin.

While I was there, I slightly simplified the error checking by using
existing IBAPrep functionality to check for the same edge cases.


src/libOpenImageIO/imagebufalgo.cpp
```diff
@@ -1098,7 +1084,7 @@ ImageBufAlgo::fillholes_pushpull (ImageBuf &dst, const ImageBuf &src,
 
     // Now copy the completed base layer of the pyramid back to the
     // original requested output.
-    paste (dst, dstspec.x, dstspec.y, dstspec.z, 0, *pyramid[0]);
+    paste (dst, src.spec().x, src.spec().y, src.spec().z, 0, *pyramid[0]);
```
lgritz: This line is the actual bug fix. The rest is just cleanup, testing, and docs clarification.

fpsunflower:
LGTM!

PR #1772, opened 2017-09-28
Refactor of metadata printing and Exif read/write

lgritz:
"Maker notes" are extra camera manufacturer-specific metadata stored as
opaque blobs inside Exif records. The manufacturers don't seem to
publish the specs for these blobs, but they have been reverse engineered
and with some hunting you can find tinkerer-prouced documentation (at a
range of levels of completeness or correctness).

With this patch, we do a substantial refactor of exif.cpp to incorporate
full awareness of maker notes into our metadata processing. At this
stage, we only read Canon makernotes, and quite incompletely. But we get
a lot more of the metadata than we used to (previously we just ignored
any makernotes we found), and the structure is now sound for adding more
complete support for this and other cameras. It involved a substantial
refactor of exif.cpp, as well as some other assorted tweaks.

This will be an ongoing project to support more cameras, but the foundation
is now set for this work.

### DIFF
--- a/src/libOpenImageIO/CMakeLists.txt
+++ b/src/libOpenImageIO/CMakeLists.txt
@@ -29,7 +29,8 @@ if (NOT MSVC)
 endif()
 
 list (APPEND libOpenImageIO_srcs
-                          deepdata.cpp exif.cpp formatspec.cpp imagebuf.cpp
+                          deepdata.cpp exif.cpp exif-canon.cpp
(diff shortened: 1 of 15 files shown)


lgritz:
Closing, there's more work to be done here. Will re-submit later.

lgritz:
Re-opening with new try, I think it is much more complete now.

PR #1775, opened 2017-10-02
string_view -- document a subtle but rare bug, turn off address sanitizer

lgritz:
See the comments in the patch for details. It's been like this for a
long time. I think in practice, we'll never hit the bug. But we'll
probably phase this method out anyway, since it's not part of the C++17
std::string_view.


fpsunflower:
LGTM

PR #1777, opened 2017-10-06
PSD: Fix parse issue of layer mask data

yamasakia:
Some PSD files failed to open. I fixed the problem of the PSDInput::load_layer.

## Description
The length of the layer mask data was assumed to be several fixed sizes, but it actually varies depending on the mask parameter flags.

We can create this kind of files by using the layer mask options. https://helpx.adobe.com/in/photoshop/using/masking-layers.html#adjust_mask_opacity

## Tests
I added new test source which caused the issue to the psd testsuite.

## Checklist:


lgritz:
I'm really sorry about letting this languish for so long. 

LGTM.

PR #1780, opened 2017-10-06
ImageCache: improve thread contention

lgritz:
Increase the number of shards in the unordered_map_concurrent used by the ImageCache for file and tile caches. As we get machines with larger core counts and thread them more heavily, the previous limits were causing bottlenecks.

This solution is not perfect nor complete. But it does push the bottleneck from ~16 to more like ~32 threads. I'm separately working on some other strategies to reduce the amount of locking, with an eye toward every-growing core counts. But this is a solid step forward for our newer machines.

I also made some of the timings printed by testtex be just a bit more explanatory.

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -69,6 +69,8 @@ namespace pvt {
 // shadow maps!
 #define USE_SHADOW_MATRICES 0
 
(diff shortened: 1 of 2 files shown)


fpsunflower:
LGTM - this seems like a reasonable improvement for the short term.

PR #1785, opened 2017-10-12
More principled handling of float<->string

lgritz:
This refactor comes after a belated realization that our numeric conversion routines (principally `from_string<>`) are "locale-dependent" via their use of strtod, as well as atof() scattered through the code.

In particular, this means that when software using this code is running with a locale in which the decimal mark is something other than '.' (e.g., many European locales that use ',' for decimal mark), these functions will incorrectly parse text floating point numbers that use the standard period as decimal mark.  For example, if the locale is "fr_FR", atof("123.45") will return 123.0, rather than the correct 123.45.

At the core of our fix is the introduction of Strutil::strtof/strtod which by default is "locale-independent" (i.e. always uses '.') and also has varieties that take a std::locale& explicitly. The implementation uses strtod_l/strtof_l (nonstandard, but available on Linux, OSX, BSD, and a differently-named but equivalent function on Windows), with a gross but simple fallback for any straggler platforms (MINGW?) that copies the string and replaces the '.' with the native locale's mark.

This is pretty inexpensive: on Linux, the locale-independent version of our new Strutil::strtof has the same cost as the system locale-dependent atof/strtof, and the explicit locale version is only about 20% slower. On OSX, atof is remarkably fast, so our version that uses strtof_l is about 2x slower, but that seems acceptable given that it gives us all the control we want, and in comparison, another strategy involving istringstream was 10x slower. Unless this is shown to be an important bottleneck, I'm rejecting the alternative of embedding a full custom strtod implementation as being unnecessarily inviting maintenance costs and potential bugs. The approach we chose is minimally invasive and relies on system libraries for the heavy lifting.

Additionally:

* Added a new stof(), stoi(), stoul() to replace the clunkier template syntax of from_string<>. The stof(), like our new strtof(), has a locale-independent as well as an explicit-local versions.

* A handful of places that still had old calls to std atof and strtod were changed to Strutil::stof.

* Strutil::format and ustring::format (and our copy of tinyformat underneath) were touched up so that the versions of format() that return strings will use the "C" locale, versions added that return strings and use an explicitly-passed locale, and the versions that append their results to existing streams continue to operate as before by honoring the existing locale conventions of the streams they are using.

* Several places where we assembled strings using std::ostringstream, I forced the stream to use classic "C" locale in cases where it was likely to have any floating-point data output.

**TIP** going forward: For any I/O that must be persistent (files saved or read), or always be formatted identically regardless of the computer being used, always take care to use locale-independent (i.e. classic "C" locale) formatting functions and also making sure to initialize any std::ostringstream with `stream.imbue(std::locale::classic())`. For ephemeral I/O or UI elements that you want to display correctly internationalized for the user's country, then use the versions with explicit locale `std::locale()` and use the default initialization of output streams.

As an aside, I do wish that C/C++ had adopted the convention of default using the classic (globally uniform) locale for all functions that don't take an explicit locale, and leave the explicit locale functions just for programs that are intentionally trying to do country-by-country localization. This is all just a huge mess, and I hate that our library can have subtle I/O bugs when used by someone in another country because of a particular global locale setting that we have no control over and can't modify simply without potentially breaking the surrounding app.

Also, for anybody who has bothered to read this far, it's not unreasonable for you all to tell me that I've totally over-engineered this, and that as far as you're concerned, anything OIIO reads and writes should always and only use '.' decimal point and omit any locale flexibility, and that I should simplify all of this accordingly. But I think what I've done is made the default modes of everything behave this way at nominal cost, and only additionally (and at no extra expense if you don't use it) provided flexibility to use our functions to internationalize for nice UI and such for those who really want it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ addons:
       - libavformat-dev
       - libswscale-dev
       - libavutil-dev
+      - locales
       # - qtbase5-dev   # FIXME: enable Qt5 on Linux
       # - bzip2
       # - libtinyxml-dev
--- a/src/fits.imageio/fitsinput.cpp
+++ b/src/fits.imageio/fitsinput.cpp
@@ -348,7 +348,7 @@ FitsInput::add_to_spec (const std::string &keyname, const std::string &value)
     // converting string to float or integer
     bool isNumSign = (value[0] == '+' || value[0] == '-' || value[0] == '.');
     if (isdigit (value[0]) || isNumSign) {
-        float val = atof (value.c_str ());
+        float val = Strutil::stof (value);
         if (val == (int)val)
             m_spec.attribute (keyname, (int)val);
         else
--- a/src/include/OpenImageIO/optparser.h
+++ b/src/include/OpenImageIO/optparser.h
@@ -40,6 +40,7 @@
 #define OPENIMAGEIO_OPTPARSER_H
 
 #include <string>
+#include <OpenImageIO/strutil.h>
 
 OIIO_NAMESPACE_BEGIN
 
@@ -67,9 +68,9 @@ optparse1 (C &system, const std::string &opt)
     char v = value.size() ? value[0] : ' ';
     if ((v >= '0' && v <= '9') || v == '+' || v == '-') {  // numeric
         if (strchr (value.c_str(), '.'))  // float
-            return system.attribute (name.c_str(), (float)atof(value.c_str()));
+            return system.attribute (name, Strutil::stof(value));
         else  // int
-            return system.attribute (name.c_str(), (int)atoi(value.c_str()));
+            return system.attribute (name, Strutil::stoi(value));
     }
     // otherwise treat it as a string
 
--- a/src/include/OpenImageIO/strutil.h
+++ b/src/include/OpenImageIO/strutil.h
@@ -97,23 +97,41 @@ void OIIO_API sync_output (std::ostream &file, string_view str);
 ///
 /// Uses the tinyformat library underneath, so it's fully type-safe, and
 /// works with any types that understand stream output via '<<'.
+/// The formatting of the string will always use the classic "C" locale
+/// conventions (in particular, '.' as decimal separator for float values).
 template<typename... Args>
 inline std::string format (string_view fmt, const Args&... args)
 {
     return tinyformat::format (fmt.c_str(), args...);
 }
 
+/// A version of Strutil::format() that uses an explicit locale.
+template<typename... Args>
+inline std::string format (const std::locale& loc, string_view fmt,
+                           const Args&... args)
+{
+    return tinyformat::format (loc, fmt.c_str(), args...);
+}
+
 
 /// Output formatted string to stdout, type-safe, and threads can't clobber
-/// one another.
+/// one another. This will force the classic "C" locale.
 template<typename... Args>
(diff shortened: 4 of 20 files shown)


lgritz:
Note: I've also submitted the couple of changes I made internal to tinyformat.h to the tinyformat project:  https://github.com/c42f/tinyformat/pull/44  We're negotiating over the right way to do that, so some minor things may change along the way and we'll propagate it back here.

lgritz:
There is one decision I made in this patch that I'm not 100% sure of:

As submitted, `Strutil::fprintf()` (appends formatted output to a stream) honors the existing locale associated with the stream.

An alternate design would be for `Strutil::fprintf()` to ALWAYS print "C" locale style even to a stream, and have a second version that takes an explicit locale, to which could could pass mystream.getloc() if you wanted to preserve the stream locale.

I don't know which is better. The alternative has the advantage of being 100% consistent across all the Strutil functions -- default behavior always is C locale and you have to override to get something localized. But the stream overall will appear to have inconsistent behavior if you output to it with a mix of direct output (`mystream << myfloat`) and Strutil-formatted string (`mystream << Strutil::format("%g",myfloat)`.

I'm leaning toward preferring the "alternate", but seeking advice.

lgritz:
Pushed more fixes and refactoring after it was pointed out that there were some additional latent bugs related to places where we implicitly assumed that a string_view was null terminated or within a null-terminated string. Also fixed important bugs in Strutil::parse_int and parse_float where it was definitely not safe to assume that the string_view passed was the null-terminated edge of the string, and so, for example if a string_view pointed to the characters "12" but the very next characters were "345" parse_int would return 12345 instead of the correct 12.

fpsunflower:
The cases you are fixing here sound good to me, but I'm still worried that this is more complexity than if we just grabbed a known good atoi/atof and used that all the time. OIIO is not a localization library. I think it would be fine to document its parsing/format functions to be intended for predictable serialization and forget locales exist altogether.

Chances are if you really are designing some international UI application you are using Qt which has much more comprehensive solutions for all this stuff (AFAIK).

lgritz:
I think that's an extremely valid position -- just declaring that because at its heart, OIIO is fundamentally about persistent and predictable I/O (duh), even its generic utility components ought to use "C" locale only and be done with it, and if you want something else you should seek another library.

The counter-argument is simply the observation that it's only 10% more code to also have (optional) entry points that would let you conform to C++'s behavior. It's actually not a lot of extra complexity (I would consider it a larger addition of unnecessary complexity to have our own string conversion functions -- at least the way I've done it, we rely on std for all the actual character handling).

So you think I should just drop the loc-aware calls entirely, leave France to puzzle over its commas without us, and I guess we could always add them back later if a need comes up?

lgritz:
On the other hand, if we took an existing high-quality atoi/atof, we could adapt it to work on a string_view directly -- without it needing null termination and without the need to make a temporary std::string.

Hmmm...

fpsunflower:
That's what I was thinking yes - a directly `string_view` enabled parser.

lgritz:
ok, ok, let me take a look

lgritz:
Counter-proposal in #1796

lgritz:
We decided to go with the alternate solution of #1796, so I'm going to close this PR. We can always revisit if something comes up that necessitates OIIO exposing locale-aware formatting or parsing.

PR #1788, opened 2017-10-15
ColorConfig refactor

lgritz:
* Change ColorConfig to return shared ptrs to ColorProcessors. This frees
  the caller from having to own the resource and remember to delete it.
  It also makes leaks impossible.

* Establish an internal cache of ColorProcessors, so that repeated requests
  for the same processor (i.e., same color space transformation) doesn't
  need to do any allocations or trigger anything expensive on the OCIO
  side.

* Added new create...() methods that take ustring arguments, in addition
  to the existing ones that take string_view. That version is faster
  if you already have ustrings on the app side -- especially the part of
  looking it up in the cache. The string_view ones are still fine to use,
  they just build ustrings on their own.

* A little bit of C++ modernization.

NOTE that this is a breaking change, although a minor one. Any applications
that had explicitly asked for a ColorConfig and used it to request a raw
ColorProcessor will need to make some minor changes:
  - The return type of the ColorConfig::create... has changed from
    ColorProcessor* to ColorProcessorHandle.
  - You don't need to free the handle, it's a shared pointer now and will
    clean itself up.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -33,6 +33,8 @@
 #include <string>
 #include <vector>
 
+#include <boost/container/flat_map.hpp>
+
 #include <OpenEXR/half.h>
 
 #include <OpenImageIO/strutil.h>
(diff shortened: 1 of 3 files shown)


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -49,6 +51,64 @@ namespace OCIO = OCIO_NAMESPACE;
...
+    ColorProcCacheKey (ustring in, ustring out,
+                       ustring key=ustring(), ustring val=ustring(),
+                       ustring looks=ustring(), ustring display=ustring(),
+                       ustring view=ustring(), ustring file=ustring(),
+                       bool inverse=false)
+        : inputColorSpace(in), outputColorSpace(out),
+          context_key(key), context_value(val), looks(looks), file(file),
+          inverse(inverse)
+    {
+        hash = (inputColorSpace.hash() ^ outputColorSpace.hash() ^
```
fpsunflower: I don't think xoring the hash values together is a great way to combine them (not that we have a convenient way around this right now).

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -49,6 +51,64 @@ namespace OCIO = OCIO_NAMESPACE;
...
+                looks.hash() ^ display.hash() ^ view.hash() ^ file.hash())
+             + (inverse?6421:0);
+    }
+
+    friend bool operator< (const ColorProcCacheKey& a, const ColorProcCacheKey& b) {
+        if (a.hash < b.hash) return true;
+        if (b.hash < a.hash) return false;
+        // They hash the same, so now compare for real
+        if (a.inputColorSpace < b.inputColorSpace) return true;
+        else if (b.inputColorSpace < a.inputColorSpace) return false;
```
fpsunflower: `else` is not needed here and make the conditions a bit harder to read.

Ideally the string comparisons would be only done once instead of twice each too.

src/libOpenImageIO/color_ocio.cpp
lgritz: This is not going to be a heavily contended hash table -- any run is going to have a fairly limited number of transformations (because there are a small number of color space names).

lgritz: Oh, though I see the a particular case is problematic: it will hash from-to and to-from identically. I will add some multipliers in there to prevent that.

src/libOpenImageIO/color_ocio.cpp
lgritz: Oh, I used the wrong operator. Changing all these to foo.c_str() < bar.c_str() because there's no reason it needs to be alphabetical.

PR #1791, opened 2017-10-23
unittest.h: Make the reference to Strutil fully qualified with OIIO::

lgritz:
It previously worked for the OIIO tests that all seemed to have a
'using namespace OIIO', but failed for external projects that just wanted
to use unittest.h and weren't inside namespace OIIO.


fpsunflower:
LGTM

PR #1794, opened 2017-10-24
Fix CMake error with MSVC compiler

KindDragon:
This is simple change that fix MSVC error:
```
CMake Error at src/libOpenImageIO/CMakeLists.txt:154 (add_dependencies):
  The dependency target "" of target "OpenImageIO" does not exist.
```


lgritz:
LGTM! Thanks for the fix.

PR #1801, opened 2017-11-05
Rewrite Python bindings using pybind11

lgritz:
pybind is a great project by Wenzel Jakob, very much in the spirit of
Boost.Python (and with substantially similar syntax), but is
C++11-based, light-weight, header-only, and compiles in a fraction of
the time as code that uses Boost.Python.

So the primary change here is to switch underlying implementation of our
Python bindings. But in the process, I made a number of improvements and
changes to the Python bindings, some of which are not strictly backwards
compatible, so this is a BREAKING CHANGE that may require some (hopefully
minor) changes to your OIIO-using Python code.

* All of the functions that read and write blocks of pixels (such as
  ImageInput.read_image(), ImageOutput.write_tile(), or
  ImageBuf.get_pixels()) no longer use old-style Python arrays, but now
  use NumPy ndarray objects. I'm guessing that NumPy is such a common
  Python package that it's not an onerous dependency to add, the n-D
  array syntax is really handy for images (e.g., pixel regions are
  naturally structured as [y][x][channel] multidimensional arrays and
  very easy to use that way), and pybind11 makes it very easy to
  manipulate NumPy ndarray objects but comparitively difficult to use
  old Python arrays.

* We now transparently accept strings anyplace that a TypeDesc is used
  as a function parameter. For example,
  `ImageBuf.set_write_format(TypeFloat)` can alternatively be called as
  `ImageBuf.set_write_format("float")`.

* Some specialized enum types (such as ImageInput.OpenMode) have been
  replaced by strings: `ImageInput.open (filename, ImageInput.Create)`
  should now be called as `ImageInput.open (filename, "create")`

It should be encouraging that of all the Python test files in our test
suite, `testsuite/python-*/src/test-*.py`, a very small number of source
lines needed to be changed.

Also note that we include pybind11 as a git submodule (since it's far
from being a common installed dependency) This is the first time we've
used a submodule, and I'm not 100% convinced this is the way to go...
we may revise this and instead make the build download it if not found.


lgritz:
After some discussion with Chris, I decided to just git clone pybind11 if we need a local copy, rather than using git submodules. (I can't discern any penalty for doing it this way, and it gives us a little more flexibility as well as not scaring people who don't like submodules.)

Also changed the download logic just a bit. It is now:

* If BUILD_PYBIND11_FORCE is true, download a local copy (regardless of      any system install).

* Otherwise, look for a system-installed copy, and use it if found.

* Otherwise, if BUILD_MISSING_PYBIND11 is true (default), download      a local copy.

So, the default behavior is to use an installed copy, or auto-download if that's not found. But it's possible to force a download (ignoring system) or to disallow a download (making missing pybind11 simply an error).

Also, none of this happens if the build is USE_PYTHON=0.

fpsunflower:
I can't speak to the backwards compatibility issues - but the binding code itself looks much nicer!

lgritz:
I've seen no objections here or on the mail list, so I'm going to merge this today.

PR #1803, opened 2017-11-10
TextureSys: bump max_tile_channels limit to 6

lgritz:
A use case has come up that makes me think 6 channel textures are going to be important for us, so I want the threshold for splitting tiles to include that situation.


fpsunflower:
LGTM

PR #1810, opened 2017-11-24
New "--bumpslopes" exporter option for maketx command

lecocqp:
The main purpose of this option is to export additional gradient informations along bump maps to ease the implementation of bump to roughness techniques.

The resulting output is a single 6 channels bump map containing the height, the first and second slopes moment values stored in subsequent mip-map levels.

Channel data layout:

channel[0] = h
channel[1] = dh/ds
channel[2] = dh/dt
channel[3] = (dh/ds)^2
channel[4] = (dh/dt)^2
channel[5] = (dh/ds) * (dh/dt)

Note that if the input texture is a normal map, then channel[0] = -1.0f and the corresponding normal can be simply retreived as follow:

n = normalize( -(dh/ds), -(dh/dt), 1.0 )

The normals are assumed to be defined in the tangent space.

## Checklist:

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -62,6 +62,7 @@
   using boost::regex_replace;
 #else
 # include <regex>
+#include <vector>
   using std::regex;
   using std::regex_replace;
 #endif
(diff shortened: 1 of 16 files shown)


lgritz:
First, let me say: (a) this is very cool, and the code is really good except for some very nit-picky comments I'll have; (b) for the benefit of everybody else, the point of this feature is that the extra information can allow a renderer to properly antialias bump/displacement maps and turn them into roughness variations in your BSDFs at the proper scale; (c) we will also submit an OSL-side PR that will fully explain how to use these maps in a renderer.

Pascal, I think you're on the right track with the new enum that controls texture creation (MakeTxBumpWithSlopes).

But I am less certain about the pros and cons of designating the resulting file as a different textureformat ("Bumpslope" vs "Plain Texture"). 

In no other case do we have different texture formats for different channel semantics. For example, a 3-channel color diffuse albedo map is "Plain Texture", but so is a specularity map, or a roughness map, or an "isolation mask", or for that matter a bump map! So the use case or number of channels does not seem to have a 1-to-1 correspondence with textureformat. But note that there are separate textureformats for "Latlong Environment" and "CubeFace Environment", even though their pixels represent the same conceptual quantity.

So I'm not sure I've ever articulated this before, but looking at how we have used it, the distinction between textureformats appears to be what kind of *lookup coordinates* they use. And these bumpslope maps are indexed by normal 2D surface parameters just like any other plain texture map (but differently from the other types -- it's not a direction vector, or a 3D shadow lookup, etc.).

I'm also concerned about back compatibility: where are all the places -- both elsewhere in OIIO, as well as OSL or any other apps or pipelines using maketx'ed files -- that might have subtle errors with the introduction of a new textureformat? I worry that it could introduce more troubles than it solves.

I'm not 100% convinced of my arguments above. I do like the idea that there is a clear marker in the metadata that indicates that this is a bump slope texture. That allows error checking on the other end, and automatic detection (from a shader, say) of whether it should treat the texture as an ordinary height-only bump or the bump/slope kind of texture. But perhaps 6 channel textures are so unusual that the channel count itself is a pretty reliable indicator that it's a bumpslope texture instead of a bump height texture?

So I'm very willing to entertain efforts to talk me out of this position, but I think I would recommend making the textureformat still be "Plain Texture" for the bumpslope maps.

lecocqp:
Thanks Larry! 

After reading your comments, I'm thinking too that it's safer to set the textureformat as "Plain Texture" for the bumslope maps. As you suggest, the number of channel is the good way to check whether we are dealing with a regular or an extended bump-map. So, I will follow your recommendation.

lgritz:
OK, good. I was primarily afraid that there is code like this scattered through the world:

    if (texturetype == "Plain Texture")
         do something for 2D coordinate lookup;
    else if (texturetype == "LatLong Environment")
         do something else for directional lookup;

and introducing a new texture type will -- SILENTLY -- introduce subtle bugs or unhandled cases into people's software. 

I think just continuing to use "Plain Texture" for these 2D surface parameter maps will be the least surprising thing for downstream software.

src/libOpenImageIO/maketexture.cpp
```diff
@@ -931,7 +1034,20 @@ make_texture_impl (ImageBufAlgo::MakeTextureMode mode,
...
     }
-
+    
+    if (mode == ImageBufAlgo::MakeTxBumpWithSlopes ){
+        ImageSpec newspec = src->spec();
+        newspec.width = newspec.full_width = src->spec().width;
+        newspec.height = newspec.full_height = src->spec().height;
+        newspec.tile_width = newspec.tile_height = 0;
+        newspec.format = TypeDesc::FLOAT;
+        newspec.nchannels = 6;
```
lgritz: I think in this spot you should adjust newspec.channelnames, so that at least for exr files, the channels will have meaningful names.

src/libOpenImageIO/maketexture.cpp
```diff
@@ -428,7 +428,110 @@ lightprobe_to_envlatl (ImageBuf &dst, const ImageBuf &src, bool y_is_up,
     return true;
 }
 
+// compute image gradient in s,t space using a Sobel filter
+static void 
+sobel_gradient(const ImageBuf &src, int x, int y, int width, int height, float *h, float* dh_ds, float *dh_dt ){
+            
+    ASSERT (src.spec().nchannels==1);
```
lgritz: Instead of asserting (and checking on every pixel), I think it would be better to detect the wrong number of input channels farther upstream and give a helpful error, and/or just use channel [0] regardless of how many are in the file.

src/libOpenImageIO/maketexture.cpp
```diff
@@ -931,7 +1034,20 @@ make_texture_impl (ImageBufAlgo::MakeTextureMode mode,
         mode = ImageBufAlgo::MakeTxEnvLatl;
         src = latlong;
     }
-
+    
+    if (mode == ImageBufAlgo::MakeTxBumpWithSlopes ){
+        ImageSpec newspec = src->spec();
```
lgritz: This is probably the right place to check and issue errors for any invalid inputs, like unexpected number of channels in the bump image, etc.

src/libOpenImageIO/maketexture.cpp
```diff
@@ -428,7 +428,110 @@ lightprobe_to_envlatl (ImageBuf &dst, const ImageBuf &src, bool y_is_up,
...
+// compute image gradient in s,t space using a Sobel filter
+static void 
+sobel_gradient(const ImageBuf &src, int x, int y, int width, int height, float *h, float* dh_ds, float *dh_dt ){
+            
+    ASSERT (src.spec().nchannels==1);
 
+    // 3x3 pixels block
+    const float* pix[3];
+
+    pix[1] = static_cast<const float*>(src.pixeladdr(x, y));
```
lgritz: Aha, so the problem here is that pixeladdr will return nullptr if the image pixels aren't in memory. It's a little bit of a dangerous method.

In the calling function, you are already using ImageBuf::Iterator to loop over the destination (output) image. I think that here, you should similarly be using ImageBuf::ConstIterator to access the source image. Not only will that be robust to use of the underlying ImageCache, but also you can make the iterators handle the border conditions for you, eliminating all the special cases you have below.

lgritz: So what I'd recommend is that you pass a reference to the dst iterator here, let's call it `ImageBuf::Iterator<float> &dstpix`, and then iterate over the 3x3 region of source like this:

    ImageBuf::ConstIterator<float> srcpix (src, dstpix.x()-1, dstpix.x()+2, dstpix.y()-1, dstpix.y()+2, 0, 1, WrapClamp);
    static const float sobelweight[9] = { ... };
    for (int i = 0; !srcpix.done(); ++srcpix, ++i) {
        // accumulate to dh_ds, etc., using a table weight[] with the 3x3 coefficients
        *dh_ds += sobelweight[i] * src[0];  // handles wrap!
        ...
    }

lecocqp:
Many thanks Larry for your precious comments! 

I just posted a revised version using iterators. So much easier and it simplifies the code a lot!

src/libOpenImageIO/maketexture.cpp
```diff
@@ -62,6 +62,7 @@
   using boost::regex_replace;
 #else
 # include <regex>
+#include <vector>
```
lgritz: Are you sure you want this line here? This is a close that is specific to using regex in certain compiler combinations.

src/libutil/thread.cpp
```diff
@@ -139,6 +139,7 @@ class thread_pool::Impl {
     void resize(int nThreads) {
         if (nThreads < 0)
             nThreads = std::max (1, int(threads_default()) - 1);
+        nThreads = 0;
```
lgritz: er, this is the line I tried to eliminate. Can you remove that?

lgritz:
Except for those two lines I pointed out (the one fixing my error with the `threads=0`, and the other with the odd `#include<vector>`), it LGTM!

src/libOpenImageIO/maketexture.cpp
lecocqp: No idea from where it comes. It just appeared with the last merge.
I will remove it!

lgritz:
I went to merge this and do a few minor touch-ups (a little formatting conforming and so on) and noticed a problem: This code only worked if the source image was 32 bit float pixels!  I went ahead and made the minor changes on my end to allow it to work for any pixel type (turned one of your functions into a template and called with the OIIO_DISPATCH_COMMON_TYPES macro). 

I think I did everything right, but after I merge, you should also double check on your production examples just to be sure I didn't break anything along the way.

Also (and not unrelated -- this is how I discovered the problem), I made the testsuite example read and write 'half' data, because for full float, I was getting LSB differences between the platforms that made it hard to make a single test result reference output. For 'half', the reduced precision makes them all match in the CI tests.

I'll merge the updated version today.

lecocqp:
Thanks Larry, looks good to me. I will check on my production tests when merged and updated.

lgritz:
Merged.

It's a limitation of GitHub that there's no easy way for me to mark this as "merged" if I didn't merge it through the interface. It will look "closed", but just know that it's not closed-and-rejected, it's closed-and-merged-with-minor-changes.

PR #1812, opened 2017-11-25
Increase "local" space of ParamValue

lgritz:
ParamValue holds "small" data locally (in the struct), and only
allocates when above that threshold. Previously, the threshold was 8
bytes, the size that the allcoated pointer would be anyway. But it seems
to me that the way we us ParamValue frequently must contain a color (3
floats), so I bumped the local space from 8 to 16 bytes (meaning it can
hold up to 4 floats or ints before a malloc). This increases the total
size of a ParamValue from 32 to 40 bytes, but I think it will make it
fairly rare to allocate, so we'll come out ahead.

This should all be inconsequential from an API perspective.

### DIFF
--- a/src/include/OpenImageIO/paramlist.h
+++ b/src/include/OpenImageIO/paramlist.h
@@ -113,13 +113,9 @@ public:
     }
 
     // Rvalue ref -- "move constructor"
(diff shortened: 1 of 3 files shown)


fpsunflower:
LGTM

PR #1819, opened 2017-12-03
Fixes to maketx image hashing

lgritz:
I'm worried there are some edge cases where the hashes aren't quite as
deterministic as we want, across all platforms and use cases.

I believe I found a bug in ImageBufAlgo::computePixelHashSHA1. Images
smaller than the hashing block size would not have their hash include
the "extrainfo". So I restructured slightly to be extra sure that all
images go through the same parallel_for loop that computes the block
hashes. Small images that only need one block still take an early out,
but only if they also have no extrainfo to add to the hash.

I decided to also get rid of the option to use OpenSSL for its hash, the
different for the two code paths, we could have different hashes for
people who build with OpenSSL versus those who build without it.

This required updating the testsuite reference images for some maketx
test cases involving small images (where the extrainfo wasn't being
used.

Although this may change the embedded maketx hashes for some images,
therefore making it possible that an old-maketx-compiled image will not
correctly be identified as being pixel-identical to a
new-maketx-compiled image, I don't think this should be a concern,
because it was only a problem with images having fewer than 256
scanlines, so those aren't the ones where it's critical to detect
duplicates (the main way that the maketx embedded hashes are used).

I also beefed up the code in ImageCacheImpl::verify_file where we check
for duplicates using the hash fingerprints. I became worried about what
could happen if you had two textures that were different "shapes" but
the same hashes -- such as a 64x64 4 channel texture and a 128x128
single channel texture, where the "unwrapped pixel sequence" (if you
know what I mean) was the same, they would end up with the same hash.
So I modified the check to only count identical hashes as a match if the
width, height, depth, channels, number of subimages, and miplevels also
were the same for the two images.


lgritz:
I see now that I was a little too harsh in my assessment of the single-block case: it did take the extrainfo into account after all. So I'm backing out of that small part. But I'm still keeping the removal of the openssl path and the fix related to making sure that images of different shapes aren't fooled by having the same pixel hashes.

PR #1822, opened 2017-12-07
Performance tweaks

fpsunflower:
A small collection a minor performance tweaks. The most important is probably correctly aligning the `ImageCache` struct because this isn't guaranteed by the C++ standard.

These aren't moving the needle too much just yet, but show a small decrease in the relevant functions in vtune.

### DIFF
--- a/src/include/OpenImageIO/platform.h
+++ b/src/include/OpenImageIO/platform.h
@@ -38,6 +38,8 @@
 
 #pragma once
 
+#include <utility> // std::forward
+
 // Make sure all platforms have the explicit sized integer types
 #if defined(_MSC_VER) && _MSC_VER < 1600
    typedef __int8  int8_t;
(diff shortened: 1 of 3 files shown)


src/libtexture/imagecache.cpp
```diff
@@ -1287,15 +1287,8 @@ ImageCacheFile *
...
-    if (found == m_fingerprints.end()) {
-        // Not already in the fingerprint list -- add it
-        m_fingerprints[finger] = file;
-    } else {
-        // In the list -- return its mapping
-        file = found->second.get();
-    }
-    return file;
+    // Insert if missing, otherwise return old value
+    return m_fingerprints.emplace(finger, file).first->second.get();
```
lgritz: Ooh, I didn't know about emplace!

src/libtexture/imagecache.cpp
```diff
@@ -3533,7 +3524,7 @@ ImageCache::create (bool shared)
     }
 
     // Doesn't need a shared cache
-    ImageCacheImpl *ic = new ImageCacheImpl;
+    ImageCacheImpl* ic = aligned_new<ImageCacheImpl>();
```
lgritz: Why does it matter if ImageCacheImpl uses aligned_new?

src/include/OpenImageIO/platform.h
```diff
@@ -396,6 +398,26 @@ inline bool cpu_has_avx512cd() {int i[4]; cpuid(i,7,0); return (i[1] & (1<<28))
 inline bool cpu_has_avx512bw() {int i[4]; cpuid(i,7,0); return (i[1] & (1<<30)) != 0; }
 inline bool cpu_has_avx512vl() {int i[4]; cpuid(i,7,0); return (i[1] & (0x80000000 /*1<<31*/)) != 0; }
 
+// portable alligned malloc
```
lgritz: s/alligned/aligned/

src/libtexture/imagecache.cpp
fpsunflower: Because it contains an `unordered_map_concurrent` which itself contains cache aligned bins.

If the structure doesn't get aligned properly, those bins could end up straddling cache lines, leading to false sharing.

Its somewhat unfortunate that `operator new` doesn't do this correctly by default, but there is no overload of `new` (before c++17) that accepts the alignment as an argument. So anytime you want to `new` a type with alignment larger than the default platform alignment, you need to use such a wrapper.

fpsunflower:
Fixed typo.

lgritz:
LGTM

PR #1823, opened 2017-12-07
unordered_map_concurrent performance improvement

fpsunflower:
Use the upper bits of the hash for bin selection instead of hashing the hash.

This is a small but measurable improvement on the benchmark I am looking at now.

### DIFF
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -410,11 +410,28 @@ private:
     atomic_int m_size;   // total entries in all bins
     Bin m_bins[BINS];    // the bins
 
+    static constexpr int log2(unsigned n) {
+        return n < 2 ? 0 : 1 + log2(n / 2);
+    }
+
     // Which bin will this key always appear in?
     size_t whichbin (const KEY &key) {
-        size_t h = m_hash(key);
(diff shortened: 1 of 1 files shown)


src/include/OpenImageIO/unordered_map_concurrent.h
```diff
@@ -74,13 +74,16 @@ OIIO_NAMESPACE_BEGIN
 ///
 
 template<class KEY, class VALUE, class HASH=std::hash<KEY>,
-         class PRED=std::equal_to<KEY>, size_t BINS=16,
+         class PRED=std::equal_to<KEY>, int LOG2_BINS=4,
```
lgritz: I think this is a confusing change. Can we keep it BINS, use static_assert to ensure it's a power of 2 (if we must), and figure out the bin shift via constexpr and bit twiddling tricks?

fpsunflower: I can give it a try ...

fpsunflower:
Pushed update. I agree its cleaner this way.

fpsunflower:
Tweaked a bit more to protect against potential mistakes in the future.

lgritz:
LGTM

lgritz:
Be sure the CI passes before you merge. Don't want this to fail for some platforms.

PR #1824, opened 2017-12-08
Remove read_before_insert parameter

fpsunflower:
It has been set to false by default for a while now and setting it to true seems to be always slightly slower


lgritz:
LGTM

PR #1825, opened 2017-12-09
Replace std::unordered_map in the concurrent hash map

fpsunflower:
This improves performance of the texture cache by a bit (proper benchmark numbers running over the weekend).

This is a very conservative first step, lots more can be done to further speed this up, and there are more uses of `std::unordered_map` we should benchmark, but I want to do those one at a time.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -541,7 +541,6 @@ macro (find_or_download_pybind11)
         message (STATUS "Building local Pybind11")
         set (PYBIND11_INSTALL_DIR "${PROJECT_SOURCE_DIR}/ext/pybind11")
         set (PYBIND11_GIT_REPOSITORY "https://github.com/pybind/pybind11")
-        set (PYBIND11_GIT_TAG "https://github.com/pybind/pybind11")
         if (NOT IS_DIRECTORY ${PYBIND11_INSTALL_DIR}/include)
             find_package (Git REQUIRED)
             execute_process(COMMAND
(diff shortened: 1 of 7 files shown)


src/cmake/externalpackages.cmake
```diff
@@ -569,5 +568,37 @@ macro (find_or_download_pybind11)
     endif ()
 endmacro()
 
+###########################################################################
+# Tessil/robin-map
+
+set (BUILD_ROBINMAP_VERSION "1e59f1993c7b6eace15032f38b5acb0bc34f530b" CACHE STRING "Preferred Tessil/robin-map version, of downloading/building our own")
```
lgritz: I think it's fine to abbreviate this -- just enough digits to make it very unlikely that it will collide. It's customary to do 6-8 hex digits.

src/include/OpenImageIO/unordered_map_concurrent.h
```diff
@@ -127,13 +127,13 @@ class unordered_map_concurrent {
 
         // Dereferencing returns a reference to the hash table entry the
         // iterator refers to.
-        typename BinMap_t::value_type & operator* () {
+        const typename BinMap_t::value_type & operator* () const {
```
lgritz: Are you sure about this? Versus adding a second, `const` dereference? Might not some use of unordered_map_concurrent, current or future, desire a dereference that returns a reference to a mutable object?

fpsunflower: The API for the non-const flavor is a bit different than the STL with this particular map, so I figured I would punt on this until we needed it.

lgritz: ok, fine then

lgritz:
LGTM

fpsunflower:
I have to retract this PR. It turns out the cache flushing logic (which sweeps through the table) makes use of iterator invalidation properties that most alternative hash implementations don't follow.

So a deeper rethink is required here ...

lgritz:
Not so hasty... as discussed offline, I think I might have a solution. Instead of closing, let's leave this open and just hold it until I try my idea, and if it works out, you can just rebase on top of it and the rest of your scheme can work just fine.

lgritz:
Proposed fix in #1826.

fpsunflower:
Loop in Larry's proposed fix from #1826. Benchmarks are cooking ...

fpsunflower:
Benchmarks on production frames showed up to a 5% net speed improvement. We can do better, but its a start.

The picture in vtune looks better as well, with `find_tile_in_cache` no longer being the top hotspot.

PR #1826, opened 2017-12-12
IC Cache "clock" algorithm -- fix iterator invalidation problem

lgritz:
It seems that the clock sweeping algorithm for finding cache pages to
evict (i.e., the uts of ImageCacheImpl::check_max_mem) relied on the
fact that it could erase entries from the cache and continue sweeping
with the iterator.

This worked fine for the original unordered_map, and as well for the
unordered_map_concurrent as long as the individual shards were also
implemented by unordered_map -- but only because unordered_map used
chaining and thus an iterator would still be valid after an erase
(assuming the iterator wasn't pointing to the erased item itself).

But when Chris tried using a different underlying hash table which didn't
have this property, things broke.

Some refactoring of check_max_mem fixes the problem. Basically, I just
save the ID of the advanced iterator, allow the iterator to be
invalidated by the erase operation and then re-establish a new iterator
by using the ID we saved in the previous step. Some other cleanup of the
loop also helps make the logic of the whole affair a little more clear.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -1331,40 +1331,41 @@ ImageCacheImpl::check_max_files (ImageCachePerThreadInfo *thread_info)
     // we used it), we just remember the filename of the next file to
     // check, then look it up fresh.  That is m_file_sweep_name.
 
-    // If we don't have a valid file_sweep_name, establish it by just
-    // looking up the filename of the first entry in the file cache.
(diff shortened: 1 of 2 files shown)


fpsunflower:
Looks good, let me add this to my branch and benchmark the two changes together.

lgritz:
If it passes tests on your end (including eliminating the crashes you had because of the underlying problems), feel free to merge this so you can rebase your other change on top of it.

PR #1830, opened 2017-12-16
New hash functions in hash.h and revamped hash_test

fpsunflower:
Just add a new hash (fasthash64) and one called falkhash that is disabled by default, but quite fast for large buffers if you have the AES instruction set

I tried a whole bunch more, but on my work machine - farmhash (our current default) is still the best.

fasthash64 is interesting because it is quite short (inline-able) and easy for the compiler to specialize for short lists of 64 bit ints. In these contexts it can do better than farmhash, and is quite flexible.

There's also an unrelated minor optimization for udims in here I had sitting around.

### DIFF
--- a/src/include/OpenImageIO/hash.h
+++ b/src/include/OpenImageIO/hash.h
@@ -59,6 +59,84 @@ OIIO_NAMESPACE_BEGIN
 using std::unordered_map;
 using std::hash;
 
+namespace fasthash {
+
+/* The MIT License
+   Copyright (C) 2012 Zilong Tan (eric.zltan@gmail.com)
+   Permission is hereby granted, free of charge, to any person
+   obtaining a copy of this software and associated documentation
+   files (the "Software"), to deal in the Software without
+   restriction, including without limitation the rights to use, copy,
+   modify, merge, publish, distribute, sublicense, and/or sell copies
+   of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+   The above copyright notice and this permission notice shall be
+   included in all copies or substantial portions of the Software.
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+   EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+   NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+   BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+   ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+   CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE.
+*/
+
+// Compression function for Merkle-Damgard construction.
+// This function is generated using the framework provided.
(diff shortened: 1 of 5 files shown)


lgritz:
If farmhash is the fastest, what are the conditions in which the new one is preferred?

I think the __AES__ symbol will only be defined if we add "aes" to the USE_SIMD flags. Did you benchmark that way to see how much faster it is? All of our work machines support AES, so if it is helpful, you can add it in site/spi/Makefile-bits-arnold.

src/include/OpenImageIO/hash.h
```diff
@@ -59,6 +59,172 @@ OIIO_NAMESPACE_BEGIN
...
+	for (const uint64_t v : buf) {
+		h ^= mix(v);
+		h *= m;
+	}
+	return mix(h);
+}
+
+} // namespace fasthash
+
+#ifdef __AES__
```
lgritz: The way you're done this, the function itself is only defined when compiled for __AES__. So anyplace you call it will also need to be in `#ifdef __AES__` and have some non-AES alternative.

src/libutil/hash_test.cpp
```diff
@@ -191,7 +130,41 @@ getargs (int argc, char *argv[])
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For additional information about the PCG random number generation scheme,
+ * including its license and other licensing options, visit
+ *
+ *     http://www.pcg-random.org
+ */
+
+struct pcg32_random_t {                     // Internals are *Private*.
```
lgritz: Am I right in understanding that there is no way to seed this? So what's the point? If only used in this unit test, which is obviously not performance-critical (not in the generating of the strings, anyway), why bother with this versus just using std random numbers?

lgritz:
LGTM, modulo the questions/comments I had.

src/libutil/hash_test.cpp
fpsunflower: Yeah I simplified it since its just used for the test. It would be easy to add a seeding constructor if we wanted it elsewhere.

I can switch to `std::random` here. The actual values don't matter.

src/include/OpenImageIO/hash.h
fpsunflower: That's right - this particular hash only makes sense when the processor supports those AES instructions. My guess was it would be reserved for people who know they are on recent enough x86 platforms.

I'm kind of on the fence about including it to be honest. Its faster, but only if you have a fairly large buffer. Given that OIIO has to be portable, its unlikely we'd use it internally.

fpsunflower:
Let me post the benchmarks, just to show the performance of the AES based hash.

src/include/OpenImageIO/hash.h
lgritz: I'd be very comfortable with it as long as the `#if` is *inside* the function -- in other words, if it has an AES-less fallback.

lgritz:
Keep fashhash64 for sure.

Maybe for the AES one, move into hash_test.cpp until we find an actual use for it?

Should we make the hash_test also try to compute some metric of how good the hashes are? Like, how evenly they would partition the data into buckets without clumping? (Is there an easy test for hash goodness that isn't a project unto itself?)

lgritz:
Semi-related Q's: Are there any hashes or RNGs that are designed with the goal of being progressively accelerated by SSE, AVX, AVX512 (but that still give the same answer in any case)?

fpsunflower:
Testing a hash function thoroughly is best left to projects like https://github.com/rurban/smhasher

I'll amend the patch.

fpsunflower:
Address comments.

lgritz:
LGTM

fpsunflower:
Merged.

Issue #1804, opened 2017-11-11
OIIO 1.8.6 fails to build from source (FTBFS) on arm{el,hf} in Debian


mfvescovi:
Hi Larry!

While building brand new 1.8.6 on Debian buildd infrastructure, armel and armhf builds fail due to some Qt5 conflicting declaration.

A snip of the first error in buildlog reports:
> [ 91%] Building CXX object src/iv/CMakeFiles/iv.dir/imageviewer.cpp.o
cd "/<<PKGBUILDDIR>>/build/src/iv" && /usr/bin/c++  -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NO_DEBUG -DQT_OPENGL_LIB -DQT_WIDGETS_LIB -DUSE_FREETYPE -DUSE_OCIO=1 -DUSE_OPENCV -DUSE_STD_REGEX -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I"/<<PKGBUILDDIR>>/build/src/iv/iv_autogen/include" -I"/<<PKGBUILDDIR>>/build/include" -I"/<<PKGBUILDDIR>>/src/include" -I/usr/include/arm-linux-gnueabi -I/usr/include/OpenEXR -I/usr/include/GL -isystem /usr/include/arm-linux-gnueabi/qt5 -isystem /usr/include/arm-linux-gnueabi/qt5/QtCore -isystem /usr/lib/arm-linux-gnueabi/qt5/mkspecs/linux-g++ -isystem /usr/include/arm-linux-gnueabi/qt5/QtGui -isystem /usr/include/arm-linux-gnueabi/qt5/QtWidgets -isystem /usr/include/arm-linux-gnueabi/qt5/QtOpenGL  -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIE   -Wall -fno-math-errno -Wno-error=unused-local-typedefs -Wno-unused-local-typedefs -Wno-unused-result -Wno-error=misleading-indentation -Wno-aligned-new -Wno-error=noexcept-type -std=c++11 -UUSE_FIELD3D -fPIC -std=gnu++11 -o CMakeFiles/iv.dir/imageviewer.cpp.o -c "/<<PKGBUILDDIR>>/src/iv/imageviewer.cpp"
[...]

PR #1834, opened 2017-12-18
Use QOpenGLFunctions for GL extensions, drop GLEW and obsolete GL stuff

### LINKED ISSUE #1831: Imageviewer can not be built on GLES platforms
### Problem

GLEW is only available for Desktop GL platforms, thus building the imageviewer is not possible on
GLES platform (more specifically, where Qt5 is built with the config option --opengl es2).

**Expected behavior:**  

Qt5 provides its own wrapper for Desktop GL/GLES compatibility (http://doc.qt.io/qt-5/qopenglfunctions.html#details), so there should be no need for GLEW at all.

**Actual behavior:**  
Build aborts due to incompatible GL function signatures from GLEW and GLES.

### Steps to Reproduce

1. Build Qt5 with config option --opengl es2
http://doc.qt.io/qt-5/embedded-linux.html#configuring-a-specific-device
2. Try to build openimageio

### Versions

* OIIO branch/version: 1.8.6
* OS: Linux, openSUSE TW
* C++ compiler: GCC 7.2

StefanBruens:
Instead of relying on GLEW for resolution of GL extension function entry
points use QOpenGLFunctions, which provides the same, but also allows to
run on both OpenGL desktop and GLES platforms.

Also drop extension based shader support. Shaders are supported since
OpenGL 2.0, so ARB shaders are likely no longer used and make the code
harder to read. Assume PBOs are available, which is true since OpenGL 1.5.

Text rendering in the pixelview overlay was apparently broken (no text
drawn at all), switch to QPainter based text drawing, as
QGLWidget::renderText() is not supported by QOpenGLWidget.

Signed-off-by: Stefan Brüns <stefan.bruens@rwth-aachen.de>

## Description

This fixes #1831 and #1804

## Tests

Tested on Linux with Intel graphics (Haswell generation, OGL 4.5), Mesa software rasterizer (llvmpipe)

## Checklist:


lgritz:
Wow, this is really great! The code LGTM. 

The pixelview is a little odd -- it has the values, but not the "close-up" view, and those white values might be hard to read if the image is vey light. But I'm not tempted to quibble about this, because it's certainly better than the current state, which is that it doesn't work well at all. We can always return to fix it.

It works for me (OSX). But I feel like I really need to double check that it works properly on my work Linux machines before I merge it, and I'm out of the office for the next two weeks. So, would it be really bad to wait until then so I can be sure? Unless one of my colleagues at SPI wants to try pulling the patch, building, and verifying that iv seems to build and work ok, in which case I'll merge upon hearing that message. (But if not, and it's up to me to test after I get back to the office, please just be patient... the delay does not mean I'm not very happy to receive and incorporate your patch. I just can't break things while I'm away.)

lgritz:
I should warn you that I recently did some image display for another program, with straight Qt and no explicit OpenGL, and the performance was adequate for that task, and I strongly suspect it would be for `iv` as well. (I think Qt will still use OpenGL underneath, to accelerate appropriate things, when available).

So on that basis, I had more or less decided was probably smart to remove the explicit OpenGL dependency entirely, and implement all of `iv` with conventional Qt widgets (not a GL window). I even started (just barely) to work on it today and hoped to maybe have something ready to show by the time I was back from the holiday.

So I would appreciate feedback on any or all of the following possibilities: (Note: When I say "you" here, I mean me, LG; I am saying this in your imagined voice.)

* Keep OpenGL -- the present patch (and obvious follow-ups) fixes everything about OpenGL that was worrisome, and various future design directions of `iv` (that somebody is interested in contributing to) might depend on OpenGL's features or speed in ways that can't be easily replicated without it.

* Abandon the present patch if you can remove OpenGL and have an adequately functioning iv. No point merging this if it's about to be ripped out soon.

* What the heck, merge this patch now, because even if a no-OpenGL patch is coming soon, (a) this is the safe thing to backport to 1.8, even if master removes OpenGL, and (b) it is a step forward that will hold us until you finish the no-OpenGL patch and/or act as insurance in case you can't get it working fully or in a timely fashion.

* While removing OpenGL is a worthy long-term goal, you have better and more important things to do to improve the code base right now (oh so many things), so let this patch stand as the short-term GL fix and maybe we'll come back to the GL-removal task later.

hobbes1069:
Well I know it doesn't directly help you test this PR but I'm tempted to go
ahead and try it for Fedora Rawhide since I haven't been able to build
1.8.x yet.

Thanks,
Richard

fpsunflower:
I haven't looked at this patch in detail, but I would be in favor of keeping GL support so we can apply OCIO conversions "live" via pixel shaders instead of having to bake them in memory.

Using Qt to mask cross-platform OpenGL complexities seems like the right call.

lgritz:
@fpsunflower Possible counter-arguments: (a) we don't currently have OCIO support *at all* in iv itself, so this also supposes that somebody wants to do the work for that (It would be immediately useful to many people, including for us at work, so I think it would be wise to do, but somebody needs to make time to do it). (b) iv currently doesn't support playing movies or frame sequences, only individual images (hitting a key to move from one image to another if multiple images are loaded), so it's not clear that using OCIO's OpenGL facilities is a meaningful performance advantage over using OCIO's CPU-side functions in memory on the image data before passing it to Qt as the image to display. (c) Just like with oiiotool, for iv we need to accommodate builds without OCIO available, sites that don't use OCIO or can't find an OCIO config at runtime, and working OCIO facilities that encounter color spaces that are not in the OCIO configuration (common example: studio configs that lack an "sRGB" space, so OCIO is helpless, even though we damn well know exactly what needs to be done to convert a JPEG to linear values so we can do math on it or convert to a linear format like OpenEXR).

I'm not particularly trying to talk us out of OpenGL. In fact, it's less work for me to just abandon the Qt-only route I barely started, so if anything I should be biased in that direction of the OpenGL status quo. But I do think we should be realistic about what we *actually* get out of direct interface with OpenGL and whether it's worth any continued development burden.

wrosecrans:
On Tue, Dec 19, 2017 at 12:04 AM, Larry Gritz <notifications@github.com>
wrote:

> I should warn you that I recently did some image display for another
> program, with straight Qt and no explicit OpenGL, and the performance was
> adequate for that task, and I strongly suspect it would be for iv as
> well. (I think Qt will still use OpenGL underneath, to accelerate
> appropriate things, when available).
>
> So on that basis, I had more or less decided was probably smart to remove
> the explicit OpenGL dependency entirely, and implement all of iv with
> conventional Qt widgets (not a GL window). I even started (just barely) to
> work on it today and hoped to maybe have something ready to show by the
> time I was back from the holiday.
>

Is there any value in the existence of iv as de-facto sample code on how
other people can use iv in OpenGL apps that don't depend on Qt?  Would it
make sense to maintain it as that, or possibly consider having two code
paths in iv -- one pure OpenGL and one Qt native so people can copy it as
they need?

I'm inclined to agree with you that "just throw it at Qt and hope for the
best" is probably fine from a performance perspective.  I once did a video
player that basically wired ffmpeg output to a QImage on a QLabel and was
bewildered that it didn't seem to be a problem and worked just fine.

lgritz:
> Is there any value in the existence of iv as de-facto sample code on how
> other people can use iv in OpenGL apps that don't depend on Qt?

I'm not sure what that means. 

>  Would it make sense to maintain it as that, or possibly consider having two code
> paths in iv -- one pure OpenGL and one Qt native so people can copy it as
> they need?

My feeling is that iv (with just one code path) has been under-maintained over the years and deserves a lot more love than it's been getting. I'm not sure how either a second code path, or double duty as canonical example code (versus merely doing its task as a running program), would make the situation better. Unless those came along with a champion who's willing to put continued work into iv to help it evolve.


> I once did a video
> player that basically wired ffmpeg output to a QImage on a QLabel and was
> bewildered that it didn't seem to be a problem and worked just fine.

I know, right? The sole reason I did the initial iv implementation using an OpenGL widget within Qt was that I assumed there would be a huge performance penalty over using QImage/QLabel, and never bothered to empirically test it. But now that I have done so in another app, my experience agrees with yours that it's not necessary to meet the performance goals of iv.

fpsunflower:
Larry: makes sense about OCIO + frame playback -- I'm all in favor of transitioning towards QImage then. 

It sounds like this patch is still a nice cleanup in the interim though.

lgritz:
So I'm currently leaning toward accepting this patch as-is (pending test at work), *eventually* plan to transition to QImage/QLabel just to simplify the code/build/dependencies further, but no rush on that because this patch fixes just enough to make it no longer a pressing issue.

lgritz:
Looks good and checks out at work, too. Merging.

lgritz:
I'll also backport to 1.8.

PR #1842, opened 2018-01-14
Linking with OpenCV highgui component is not needed for 3.X and newer.

hobbes1069:
## Description

It is not necessary to link against the OpenCV highgui component in version 3.X and above as it has been split into 3 components.

## Checklist:

- [x ] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x ] I have updated the documentation, if applicable.
- [x ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/cmake/modules/FindOpenCV.cmake
+++ b/src/cmake/modules/FindOpenCV.cmake
@@ -51,8 +51,10 @@ set (libdirs "${PROJECT_SOURCE_DIR}/lib"
              )
 
 
-set (opencv_components opencv_highgui opencv_imgproc opencv_core)
+set (opencv_components opencv_imgproc opencv_core)
 if (NOT ${OpenCV_VERSION} VERSION_LESS 3.0.0)
+    set (opencv_components opencv_videoio ${opencv_components})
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM

sambler:
This isn't completely right - libutil/paramlist_test needs to link with highgui.

```
FAILED: src/libutil/paramlist_test
...
src/libOpenImageIO/libOpenImageIO.so.1.9.3: undefined reference to 'cvCreateCameraCapture'
src/libOpenImageIO/libOpenImageIO.so.1.9.3: undefined reference to 'cvReleaseCapture'
src/libOpenImageIO/libOpenImageIO.so.1.9.3: undefined reference to 'cvQueryFrame'

```
So at least when building tests we still need to include highgui. That is if paramlist_test should be testing with camera capture.

hobbes1069:
I'll wait for Larry to comment on whether camera capture is necessary. If so I'll relook and this. I know it's standard practice but I hate globbing libraries together and blindly linking them all or nothing with a target. -Wl,--as-needed can help with this but I don't like relying on it.

lgritz:
No, paramlist_test definitely does not need any OpenCV functionality.

PR #1845, opened 2018-01-18
TextureOptBatch: Necessary to have the per-point fields aligned properly

lgritz:


fpsunflower:
LGTM

AlexMWells:
"
alignas(Tex::BatchAlign)  // align the subsequent fields properly  
float sblur[Tex::BatchWidth];    ///< Blur amount  
"
Just to clarify, the alignas will only apply to the sblur data member, subsequent fields will end up aligned because they are all wide themselves, causing each subsequent field to be aligned.  It gets the job done.  Just wondering if each wide field should be marked as alignas, for completeness.

lgritz:
Yes, indeed, I was counting on the fact that I have several adjacent aligned objects of the same size as the alignment, so as long as I aligned the first one, the rest will follow.

Do you think it would be more clear if I put an `alignas` in front of each one? Or a comment explaining?

Also (and this question is meant for @fpsunflower as well), would it be better to put the aligned vectors at the start of the struct? The main reason I didn't is for locality of access -- the last two of those, rblur and rwidth, are only used for volume lookups, i.e., not most of the time, so they seemed like good candidates to have at the end rather than in the middle. But if you guys think that's a silly reason and I should rearrange the structure to have all the width/blur vectors at the front, I can do that.

lgritz:
I'm also tempted, while I'm change this struct anyway, to make sure some of those enum and other fields (like wrap) that currently use 4 bytes use `enum class` to force them to 1 byte each.

Pro or cons? It's not like there's a large array of TextureOptBatch that makes cutting every byte out be at a premium. Is it a minor performance hit to take some of those 4-byte fields and cram them into 1 byte? Or a gain?

I don't have a strong feeling one way or the other, but certainly if we're going to change the struct, we should do it all in one step, to make things easier for Alex on the OSL side.

AlexMWells:
The existing comment made me double check the c++11 alignas specification to make sure it was being applied to a data member not changing the alignment for the rest of the scope of the structure.  I think putting it on the same line as the data member + comment, or just putting it on all of them.

As far as data locality, it shouldn't matter whether these are at the front or back of the structure, what matters is for the other data members that you want locality exist in the same cache line vs. crossing a boundary (micro-optimization).  However putting your wide data types 1st will naturally achieve that (at least for batch size 16) as each is a cache line each.  For rblur/rwidth, does make me wonder if a separate data structure for 3d that extends the 2d one make sense (no is an OK answer).

AlexMWells:
As far as the enums, as this isn't "storage" but more of a service, I don't think it is worth it, will cause more code to be generated to test them.  Counter argument would be if it reduced the cache line set, but I think these all currently fit in 1 cacheline.

I might suggest changing the "bool" to a type with a well defined byte size, like "int", or an int based enum.  On the LLVM side I have to use an int type anyway.

lgritz:
I pushed an update. What do you think of this?

AlexMWells:
I like it, just using the alignas directly is readable and self documenting.

As you mentioned the only downside is a 2 cache line gap for the rwidth,rblur when doing 2d textures.  In practice I think those 2 cache lines will end up getting evicted if not used, so only a super micro downside (vs. explicit 3d texture options).  On the llvm side I'm only to allocate one of these on the llvm stack per function and reuse it for all the texture calls in that function.

lgritz:
I think the simplicity of having one struct layout used for all the texture calls probably outweighs any extremely minor benefit to ensuring that the structs have no unused space whatsoever.
I'm happy with this, too.

AlexMWells:
agreed

fpsunflower:
LGTM2

lgritz:
Alex, are you working out of OIIO master, or shall I backport this to 1.8?

AlexMWells:
For now I'm just matching binary matching it in OSL namespace, then walking over in renderservices.cpp talking to older OIIO 1.7 single sample interface.  My goal from OSL view is to get the batching/binning/mask population correct, then look at moving to newer OIIO once something is on the backside of the TextureOptBatch.  So leaving it in master is probably fine for now.

AlexMWells:
From an OSL shader, or via ShadingSystem public API, how can one control/set the TextureOption::mipmode?

Or from OSL is always default?

lgritz:
Currently, there is not a way to set it from OSL itself, it's always the default. 
But if somebody were to raise a ruckus about wanting it, I suppose it would be via an optional token/value argument to the texture() function, like the way "width", etc., are set.

There are, however, times that it's used by an OSL-using renderer, but not under shader control. For example, in SPI's renderer, RendererServices::texture() checks the ray types to be sneaky, and for "high roughness rays" (think of it as indirect diffuse samples, but that's an oversimplification), the will add a bunch of texture blur and doctor mip and/or interp modes, in an effort to make those texture lookups cheaper and less demanding of the texture cache (for the ray types that will be so blurred out that you won't notice the shortcuts).

AlexMWells:
np, just wanted to make sure I didn't miss processing an optional texture parameter

PR #1849, opened 2018-01-20
OpenEXR reader: gracefully reject subsampled channels

lgritz:
The EXR spec allows for channels to be subsampled, but we don't handle
this properly in the exr ImageInput, and it can lead to a crash.  Since
it's a rarely used feature (if ever), I'm not inclined to fix it for
real until somebody says they actually need it. Instead, let's just
detect this unusual case and correctly issue an error saying that we
don't support it.

This required a bit of housekeeping to allow the errors to be propagated
up from the place where it was convenient to notice this.

### DIFF
--- a/src/iinfo/iinfo.cpp
+++ b/src/iinfo/iinfo.cpp
@@ -689,8 +689,8 @@ main (int argc, const char *argv[])
         if (! in) {
             std::string err = geterror();
             if (err.empty())
-                err = Strutil::format ("Could not open \"%s\"", s.c_str());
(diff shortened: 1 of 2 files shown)


jordibares:
Thank you so much Larry, I am certain none of the outputs we generate with either Arnold or Mantra uses this subsampled method so I am not worried at all.

Thanks a lot for looking at it so promptly.

lgritz:
@jordibares , aside from this spurious crash with subsampled channels, the technique we were discussing about finding incomplete files is otherwise working properly for you?

PR #1853, opened 2018-01-24
TIFF whole-file I/O speedup 5-10x with multithreading

lgritz:
OpenEXR's library has a neat feature where it maintains a thread pool and if you are using the multi-scanline or multi-tile I/O API calls, it will do the compression/decompression of the scanlines or tiles in parallel, even though the disk I/O itself is serial (but the codec work accounts for most of the time, so this is a big speed win). The result is that with many threads, a 5-7x speedup in I/O is typical, versus the single threaded case.

libtiff doesn't do this -- it is not thread-safe or reentrant (for a single file), and lacks API calls to write multiple tiles or scanlines at once (small exception: it will output a "strip" of several scanlines, but not in a flexible way like OpenEXR's). So any straightforward use of libtiff means using one core to do all the work writing (or reading) a file, and all the other cores just waiting around, unable to contribute. 

So with this patch, I take an end run around these libtiff limitations. Instead of using the TIFFWriteScanline or TIFFWriteTile calls, I use the alternative TIFFWriteRaw{Scanline,Tile}, where you have to hand it a "raw" buffer (already compressed, ready for direct write to disk), which means we have to compress it ourselves. But for the very common case of "zip/deflate" compression, we can do this with our own call to zlib and get the same compressed data blob that libtiff would have produced. And in doing so, we can do that compression, and a bunch of other data wrangling and format conversion, in parallel using our thread pool! We can also pipeline the operation, having the serialized write happen even while later chunks are still in the queue waiting to be compressed.

And of course, the same process in reverse for reading.

Some limitations: for now, I only support uint8 and uint16, only deflate/zip compression, only "contig" planarconfig, no weird modes like cmyk or palette images, etc.  But these are the common (nearly exclusive) settings for TIFF files, as we use them.

Note also that this ONLY applies when writing multiple scanlines or tiles at once. It can't parallelize this work if the calling app is only reading or writing (from the OIIO API perspective) a single scanline or tile at a time. But again, a very common use pattern is to read or write the whole image at once. (Important exception: ImageCache/TextureSystem).

The results are stunning: I'm easily getting a 5-10x performance improvement in TIFF input or output time, sometimes even much more. This translates directly into large performance improvements for oiiotool as well as maketx texture creation (of which by far the biggest portion of runtime is the output of the texture, and most of that was due to the compression, which is now multithreaded).

As a specific example, when doing "maketx" of an 8192 x 8192, 4 channel, uint16 data TIFF input file on my 16 core (32 hyperthread) workstation,

    Old:  28.2s total (1.13s read, 23.3s write)
    New:   1.6s total (0.13s read,  1.0s write)

(These times also incorporate a change being reviewed separately as 3 seconds of the original total.)

### DIFF
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -40,6 +40,7 @@
 #include <OpenImageIO/fmath.h>
 #include <OpenImageIO/imageio.h>
 #include <OpenImageIO/deepdata.h>
+#include <OpenImageIO/parallel.h>
 #include "imageio_pvt.h"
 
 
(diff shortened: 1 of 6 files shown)


lgritz:
Beware -- merge of this risky change is imminent!

PR #1858, opened 2018-01-27
ArgParse: modernize, no raw new/delete, use pimpl to hide implementation

lgritz:


lgritz:
My intent here was to make it easier to modify this class by completely hiding its implementation. 

A valid criticism would by to say: "This is exactly backwards. You should be turning ArgParse into a header-only implementation, then you don't have any link or library issues at all."

PR #1863, opened 2018-02-03
Fix for RAW files that are taken with rotated orientation

### LINKED ISSUE #1854: Bad conversion with oiiotool of a vertical raw file
### Problem

Hi,
We get a bad conversion with **oiiotool** of a raw file (tested with `.CR2` and `.RW2`) to `.JPG` when the image is vertical. The output JPG image becomes horizontal and the image is incorrect.
It works fine when the input RAW image is horizontal and it also works fine when the input image is a `.JPG` (horizontal or vertical).
We get the problem when using the `master` branch or release `1.8.7` and the problem was not there in release `1.7.16`.

**Expected behavior:**

The input vertical raw file should be convert to a vertical `.JPG` file without buffer shift.

**Actual behavior:**

Here is the input `.RW2` image (in a zip to be accepted by github): [P1330452.zip](https://github.com/OpenImageIO/oiio/files/1660708/P1330452.zip)

- Using OpenImageIO `master` branch:

`oiiotool -i P1330452.RW2 -o P1330452_oiiotool_master.jpg`
gives this output image (here re-compressed for github):
p1330452_oiiotool_master

- Using OpenImageIO `1.7.16` release:

`oiiotool -i P1330452.RW2 -o P1330452_oiiotool_1_7_16.jpg`
gives this output image: 
p1330452_oiiotool_1_7_16

### Versions

* OIIO branch/version: master (January 24, 2018)
* OS: Linux
* C++ compiler: g++-4.9

lgritz:
Fixes #1854


lgritz:
Pushed updates. I think I have it right now.

lgritz:
OK, updated.

Changed one of the two test images to be a vertical example (the Nikon D3X), per comments on #1854.

Also fixed the Orientation metadata. Please re-test and make sure you are ok with this. If we're letting libraw reorient the image, we need to reset any Orientation metadata that may have been in the Exif so it doesn't contradict what we decoded.

fabiencastan:
We tested it again with the last changes and it works fine!
Thanks

lgritz:
Great, merging!

PR #1864, opened 2018-02-04
Color transform unpremult logic changes.

lgritz:
After some deep thinking and discussion with people who know more color
than me, I have become convinced that you *almost always* want to make
sure to un-premultiply colors (i.e. divide by alpha) before doing color
space transformations with OCIO, and then re-multiply by alpha (to get
back to "associated alpha") immediately after the color transformation.

To support this,

* For the ImageBufAlgo color transformation functions, change the default
  value of the `unpremult` parameter from `false` to `true`.

* `oiiotool --colorconvert` and the various `--ocio` commands will
  also default to doing the unpremult/premult steps, unless a new optional
  modifier is used: `:unpremult=0`.

* `oiiotool --autocc` will do the unpremult/premult steps surrounding any
  automatic color transformations that it does upon input and output.

On the whole, I think this makes the situation less error prone for
casual users. Most oiiotool command lines should just use --autocc and
then not need to worry about any explicit use of `--colorconvert` or
`--unpremult/--premult`. For the rare case when more control is needed,
the individual commands and modifiers can be used.

Note that for oiiotool, `--autocc` is not on by default. If you want
images auto-converted to linear space and ouputs auto-converted from
linear to the desired output space, you need to use `--autocc`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,12 @@ Public API changes:
   only be used to pass into certain IBA functions). The color space names
   "rgb" and "default" are now understood to be synonyms for the default
   "linear" color space. #1788 (1.9.0)
+* `ImageBufAlgo::colorconvert` and various `ocio` transformations have
+  changed the default value of their `unpremult` parameter from `false` to
+  `true`, reflecting the fact that we believe this is almost always the more
+  correct choice. Also, if their input image is clearly marked as having
+  unasociated alpha already, they will not bracket the color conversion with
+  the requested unpremult/premult. #1864 (1.9.2)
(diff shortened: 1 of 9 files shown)


lgritz:
Updated this PR with a few additions:

1. The default for now of the oiiotool colorconvert/ociostuff modifier `:unpremult=` is now 0. Meaning that it preserves the old behavior unless you use `:unpremult=1`. Some time later (but before a major release), we will change the default.

2. I've added warnings to oiiotool in cases where it can tell that you have incorrectly double-unpremultiplied.

PR #1865, opened 2018-02-05
ArgParse: modernize, no raw new/delete, and make header-only.

lgritz:
This is an alternative design to #1858. But instead of pushing all the details to argparse.cpp with PIMPL, in this incarnation we push eliminate argparse.cpp, push all the details to argparse.h, making this a header-only class.

Opinions on which way is preferred?


lgritz:
Chris points out that since this header depends on some other non-header parts of oiio-util (like Strutil), it's not really the kind of header-only library that has no dependencies except for std, so there probably isn't any particular benefit to removing the .cpp.

So I'm going to abandon this PR and instead merge the alternate version from #1858.

PR #1873, opened 2018-02-16
simd.h: speed up some load/store ops especially to/from uint16, uint8

lgritz:
Big improvements on avx512 and in some cases avx2.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -4397,14 +4397,17 @@ OIIO_FORCEINLINE void vint4::store (int *values, int n) const {
         values[i] = m_val[i];
 #endif
 }
-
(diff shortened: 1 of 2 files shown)


lgritz:
My main target with this was speeding up vint -> uint16[] or uint8[].

When running on AVX512vl, this patch speeds it up (compared to pre-patch that only used SSE4) by about 2x for the vint4, about 4x for vint8, and about 8x for vint16. (How does going from 4-wide to 16 wide speed up MORE than 4x? Because it's not just wider, there are also new intrinsics that collapse several steps to one.)

Also some speedup (much more modest, but still helpful) for AVX2 for the vint8 load cases from uint8[] and uint16[], which I hadn't used any intrinsics for previously.

PR #1888, opened 2018-02-28
Modifying fmath.h to allow CUDA compilation

timgrant:
## Description

To support compilation for CUDA targets, we have added device-friendly specializations of a subset of functions in ``fmath.h``. Where necessary, include directives and function definitions that are not able to be compiled in device mode have been guarded with a preprocessor macro.
## Tests

The changes are guarded by a preprocessor macro and are only active when being compiled by a CUDA compiler. Since CPU compilation is not affected, no additional test cases have been added at this time.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -1604,8 +1604,6 @@ inline OIIO_HOSTDEVICE float fast_correct_exp (float x)
     // On x86_64, versions of glibc < 2.16 have an issue where expf is
     // much slower than the double version.  This was fixed in glibc 2.16.
     return static_cast<float>(std::exp(static_cast<double>(x)));
-#elif defined(__CUDACC__)
-    return static_cast<float>(std::exp(static_cast<double>(x)));
 #else
     return std::exp(x);
 #endif
@@ -1713,13 +1711,11 @@ inline OIIO_HOSTDEVICE float fast_safe_pow (float x, float y) {
 }
 
 
-#ifndef __CUDACC__
 // Fast simd pow that only needs to work for positive x
 template<typename T, typename U>
 inline OIIO_HOSTDEVICE T fast_pow_pos (const T& x, const U& y) {
     return fast_exp2(y * fast_log2(x));
 }
-#endif
 
 
 inline OIIO_HOSTDEVICE float fast_erf (float x)
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -62,7 +62,9 @@
 #include <OpenImageIO/platform.h>
 #include <OpenImageIO/dassert.h>
 #include <OpenImageIO/missing_math.h>
+#ifndef __CUDACC__
 #include <OpenImageIO/simd.h>
+#endif
 #include <OpenImageIO/array_view.h>
 
 
@@ -224,6 +226,7 @@ clamp (const T& a, const T& low, const T& high)
 }
 
 
+#ifndef __CUDACC__
 // Specialization of clamp for vfloat4
 template<> inline simd::vfloat4
 clamp (const simd::vfloat4& a, const simd::vfloat4& low, const simd::vfloat4& high)
@@ -236,6 +239,7 @@ clamp (const simd::vfloat8& a, const simd::vfloat8& low, const simd::vfloat8& hi
 {
     return simd::min (high, simd::max (low, a));
 }
+#endif
 
 
 
@@ -473,6 +477,7 @@ floorfrac (float x, int *xint)
 }
 
 
+#ifndef __CUDACC__
 inline simd::vfloat4 floorfrac (const simd::vfloat4& x, simd::vint4 *xint) {
     simd::vfloat4 f = simd::floor(x);
     *xint = simd::vint4(f);
@@ -490,6 +495,7 @@ inline simd::vfloat16 floorfrac (const simd::vfloat16& x, simd::vint16 *xint) {
     *xint = simd::vint16(f);
     return x - f;
 }
+#endif
 
 
 
@@ -687,7 +693,7 @@ void convert_type (const S *src, D *dst, size_t n, D _min, D _max)
 }
 
 
-
+#ifndef __CUDACC__
 template<>
 inline void convert_type<uint8_t,float> (const uint8_t *src,
                                          float *dst, size_t n,
@@ -797,6 +803,7 @@ convert_type<float,half> (const float *src, half *dst, size_t n,
     while (n--)
         *dst++ = *src++;
 }
+#endif
 #endif
 
 
@@ -1141,11 +1148,13 @@ inline int fast_rint (float x) {
 #endif
 }
 
+#ifndef __CUDACC__
 inline simd::vint4 fast_rint (const simd::vfloat4& x) {
     return simd::rint (x);
 }
+#endif
 
-
+#ifndef __CUDACC__
 inline float fast_sin (float x) {
     // very accurate argument reduction from SLEEF
     // starts failing around x=262000
@@ -1347,6 +1356,7 @@ inline float fast_atan2 (float y, float x) {
         r = float(M_PI) - r;
     return copysignf(r, y);
 }
+
 
 template<typename T>
 inline T fast_log2 (const T& xval) {
@@ -1650,6 +1660,7 @@ inline float fast_ierf (float x)
     }
     return p * x;
 }
+#endif
 
 // (end of fast* functions)
 ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


lgritz:
Hi, this LGTM, but I have a couple questions to clear up my own confusion.

So I assume the goal is for this header to be able to be used for both C++ and/or Cuda compilation.

So obviously, you have used `#ifndef __CUDACC__` to hide from Cuda the parts involving explicit CPU SIMD vector types. And I see a couple spots where you've used `#ifdef __CUDACC__` where you need to add some extra things for Cuda.

But what about the __host__ and __device business? I don't see that anywhere in the rest. Are we then to assume that you just don't currently need to use any of those other functions? But if we ever wanted to make the full header completely useful for Cuda as well as C++, we'd need to decorate a whole lot of other functions? (Perhaps using some macro that with nvcc expands to the right keyword but is empty without Cuda?) Ah, I see you've done that in the OSL-side PR with "OSL_HOSTDEVICE". Do we not need that here?

timgrant:
I erred somewhat on the side of keeping the changes isolated and covered (for the most part) only those functions needed to compile llvm_ops.cpp. But you are right that we will eventually cover more functionality, in which case cherry-picking functions probably won't be feasible and decorating more or less everything with HOSTDEVICE might prove to be more sensible.

Eventually we would like to support all of the OSL shadeops, so I will try to ascertain the amount of  OIIO coverage needed and go from there.

src/include/OpenImageIO/fmath.h
```diff
@@ -1742,4 +1753,181 @@ interpolate_linear (float x, array_view_strided<const float> y)
...
 
+
+////////////////////////////////////////////////////////////////////////////
+// CUDA device functions & intrinsics
+//
+// For now, the selection of 'fast' instrinsics is controlled via compiler
+// flags.
+
+#ifdef __CUDACC__
+__device__ inline float fast_sin (float x) {
```
lgritz: I'm wondering if this should either be right next to the other definition of fast_sin, or if there should be ONE fast_sin and the #ifdef should be inside its body?

I'm not sure I have a strong opinion either way (or this way), but I'm trying to think ahead about the long-term maintainability of this file. If we have multiple definitions of the same function for different platforms, I think in general that we want as little textural separation between them as we can do practically. (That said, I can also argue for this approach of just succinctly having all of the Cuda definitions in one compact place. So I'm not sure where that observation gets us. I did say I'm on the fence, right?)

src/include/OpenImageIO/fmath.h
```diff
@@ -1742,4 +1753,181 @@ interpolate_linear (float x, array_view_strided<const float> y)
...
+}
+
+// (end CUDA intrinsics)
+////////////////////////////////////////////////////////////////////////////
+
+////////////////////////////////////////////////////////////////////////////
+// 'Safe' device functions & helpers
+
+template <typename IN_TYPE, typename OUT_TYPE> __device__
+inline OUT_TYPE device_bit_cast (const IN_TYPE in) {
```
lgritz: Is this the SAME definition as the usual bit_cast?

If so, then I would definitely argue that there should be just one. You'll need `__device__ __host__` in front of it, I guess.

I think maybe it's worth the investment in defining OIIO_DEVICEHOST (and maybe also OIIO_DEVICE?), and I think that should be in platform.h -- that's where all the other important compiler-based and platform-based macros are defined, and it will set us up well if we need to have Cuda-defined functions in other headers.

I'll let you make this judgment based on established Cuda practice, but when people want to make a header that can be used for either C++ or Cuda, is the more common idiom (a) 

    #ifdef __CUDACC__
        #define OIIO_DEVICEHOST __device__ __host__
        #define OIIO_DEVICE __device__
    #else 
        #define OIIO_DEVICEHOST 
        #define OIIO_DEVICE
    #endif 

or, (b),

    #ifndef __CUDACC__
        #define __device__
        #define __host__
    #endif 

I've used Cuda some, but not enough to consider myself fluent enough to know which would be the better stylistic choice.

src/include/OpenImageIO/fmath.h
```diff
@@ -1742,4 +1753,181 @@ interpolate_linear (float x, array_view_strided<const float> y)
...
+    // NOTE: this is the only standards compliant way of doing this type of casting,
+    // luckily the compilers we care about know how to optimize away this idiom.
+    OUT_TYPE out;
+    memcpy (&out, &in, sizeof(IN_TYPE));
+    return out;
+}
+
+__device__ inline float safe_acos (float x) {
```
lgritz: Again, here, if this code for safe_acos is identical to the other one, I think that instead of two definitions (one with `__device__` and one without), we should have a single definition and a macro for DEVICE that does the right thing based on the compiler.

src/include/OpenImageIO/fmath.h
```diff
@@ -1742,4 +1753,181 @@ interpolate_linear (float x, array_view_strided<const float> y)
...
+    if (x <= -1.0f) return float(M_PI);
+    if (x >=  1.0f) return 0.0f;
+    return acosf(x);
+}
+
+__device__ inline float fast_safe_pow (float x, float y) {
```
lgritz: same here

src/include/OpenImageIO/fmath.h
timgrant: I have tried the approach of decorating the functions with the HOSTDEVICE macro and selecting the implementation inside the function bodies, and it seems to work well. There is the added benefit that templates and composed functions work with relatively minor changes.

I'll push an updated version of fmath.h (and platform.h with the macro definition) once we have had a chance to review and test the changes.

src/include/OpenImageIO/fmath.h
timgrant: Idiom (a) seems to be more common.

src/include/OpenImageIO/fmath.h
```diff
@@ -1538,29 +1683,44 @@ inline float fast_sinh (float x) {
...
     if (y == 0) return 1.0f; // x^0=1
     if (x == 0) return 0.0f; // 0^y=0
     // be cheap & exact for special case of squaring and identity
     if (y == 1.0f)
         return x;
     if (y == 2.0f)
+#ifndef __CUDACC__
         return std::min (x*x, std::numeric_limits<float>::max());
+#else
+        return min (x*x, std::numeric_limits<float>::max());
```
lgritz: Why does this use `min`, but below in interpolate_linear, you use `fminf`?

src/include/OpenImageIO/fmath.h
```diff
@@ -458,11 +460,15 @@ ifloor (float x)
 /// to the built-in modf, but returns a true int, always rounds down
 /// (compared to modf which rounds toward 0), and always returns
 /// frac >= 0 (comapred to modf which can return <0 if x<0).
-inline float
+inline OIIO_HOSTDEVICE float
 floorfrac (float x, int *xint)
 {
 #if 1
+#ifndef __CUDACC__
     float f = std::floor(x);
```
lgritz: Just curious...

If instead of the whole `#ifndef __CUDACC__` construct, we just had

    inline OIIO_HOSTDEVICE float fast_blah (
        using std::blah;  // or even:  using namespace std
        float f = blah(x);
        ...

would that do the right thing on both/all platforms? That is, would that idiom use std::blah if it exists, and an unqualified `blah` if it didn't? If that works, I think it may be a less cluttered approach for the functions where the only Cuda/C++ difference is prefixing with std::.  Or do I misunderstand how the namespace resolution would work?

lgritz:
LGTM, pending resolving/answering the last two comments I just made. I'm ok with it as it stands, but just looking to find the idioms that leave as little clutter and `#if` branching as possible.

src/include/OpenImageIO/fmath.h
timgrant: Poor temporal locality when editing? I'll make them consistent.

src/include/OpenImageIO/fmath.h
timgrant: Good question. I'll investigate this and get back to you.

lgritz: Something I may not have considered here:  if the Cuda compiler sees a std::floor and also an unqualified floor, and you are purposely selecting the unqualified one on Cuda (is there some reason to not want std?), then I think the `#if` would be necessary. But if Cuda simply lacks the std namespace altogether, then I think my `using` suggestion just might do the right thing.

lgritz: Did that work? Or should I commit as-is?

ee33:
Modified fmath.h coming.

src/include/OpenImageIO/fmath.h
timgrant: It did work for the most part. ```std::min``` seems to be a holdout seemingly based on a template vs. overloading issue. I will push an updated version shortly.

lgritz: Don't over-strain yourself to make the idiom fit where it's difficult. Do it where it simplifies the code easily, just leave what you have to where it doesn't.

src/include/OpenImageIO/fmath.h
```diff
@@ -1637,7 +1605,7 @@ inline OIIO_HOSTDEVICE float fast_correct_exp (float x)
     // much slower than the double version.  This was fixed in glibc 2.16.
     return static_cast<float>(std::exp(static_cast<double>(x)));
 #elif defined(__CUDACC__)
-    return static_cast<float>(exp(static_cast<double>(x)));
+    return static_cast<float>(std::exp(static_cast<double>(x)));
```
lgritz: You sure this one is right? Usually, plain old std::exp(x) would be the right thing. It's only for certain glibc libraries that for obscure reasons the double version is faster. 

I would think cuda would also fall into the "just use std::exp" case?

src/include/OpenImageIO/fmath.h
```diff
@@ -1898,15 +1867,17 @@ T invert (Func &func, T y, T xmin=0.0, T xmax=1.0,
 inline OIIO_HOSTDEVICE float
 interpolate_linear (float x, array_view_strided<const float> y)
 {
+#ifndef __CUDACC__
```
lgritz: Do you think this is something we should address in dassert.h, ensure that the definition of DASSERT is a no-op for Cuda, rather than needing a `#ifndef __CUDACC__` every time it's used? (I realize it's used just once today, but I'm thinking ahead.)

src/include/OpenImageIO/fmath.h
timgrant: Ah, I see. I will revert to the std::exp case.

src/include/OpenImageIO/fmath.h
timgrant: That would certainly be cleaner. I'll make the change when I hit the second instance. :smile:

lgritz: Fair enough

lgritz:
LGTM!

Just waiting for CI tests to finish, then will merge.

src/include/OpenImageIO/fmath.h
lgritz: @timgrant was right, it needed the full separation he originally had. Either one will work when using clang to generate PTX, as he was doing here. But nvcc is much pickier, so when you are using fmath.h in a full nvcc-compiled Cuda program, you need his original approach. Sorry to make you do the extra work. I have refactored with this in mind in #1896.

PR #1890, opened 2018-03-04
Make building and installing of documentation more packaging friendly.

hobbes1069:
## Description

I've been carrying around this patch for a while. It's much easier for me to use rpmbuild to install the documentation in the right places for Fedora. This allows me to build the documentation without installing it. 

It also uses the target to find IV instead of looking for the binary directly as it will always find the installed iv of the previous released package instead of the one in the build tree.


lgritz:
LGTM

PR #1896, opened 2018-03-12
Refactor fmath.h for Cuda.

lgritz:
Turns out that nvcc considers it an error if a `__device__` function is
called from a `__host__ __device__` function, even if the latter is
conditionally compiled only for Cuda mode.  So a function like fast_sin
(which for Cuda calls `__sinf` underneath), this will hit the error.

To make it even more complicated, this is only a problem with nvcc, not
with clang when compiling Cuda, which is how this was overlooked before.

So, as an example, you can't do this:

    __host__ __device__ void myfunc () {
    #ifndef __CUDACC__
        ...generic code...
    #else
        device_only_func();
    #endif
    }

It seems you have to structure it like this:

    #ifndef __CUDACC__
    void myfunc () {
        ...generic code...
    }
    #else
    __device__ void myfunc () {
        device_only_func();
    }
    #endif

I'm sorry, @timgrant , this is the original way you implemented the
Cuda-fication of fmath.h, and I encouraged you to change it. I think
because for that particular use for OSL, you were using clang to generate PTX,
you didn't encounter the error. So your instincts were right, sorry to
have created extra work for you.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -224,7 +224,7 @@ clamp (const T& a, const T& low, const T& high)
 }
 
 
-#ifndef __CUDACC__
+#ifndef __CUDA_ARCH__
 // Specialization of clamp for vfloat4
 template<> inline simd::vfloat4
 clamp (const simd::vfloat4& a, const simd::vfloat4& low, const simd::vfloat4& high)
@@ -475,7 +475,7 @@ floorfrac (float x, int *xint)
 }
 
 
-#ifndef __CUDACC__
+#ifndef __CUDA_ARCH__
 inline simd::vfloat4 floorfrac (const simd::vfloat4& x, simd::vint4 *xint) {
     simd::vfloat4 f = simd::floor(x);
     *xint = simd::vint4(f);
@@ -513,7 +513,7 @@ sincos (float x, float* sine, float* cosine)
 {
 #if defined(__GNUC__) && defined(__linux__) && !defined(__clang__)
     __builtin_sincosf(x, sine, cosine);
-#elif defined(__CUDACC__)
+#elif defined(__CUDA_ARCH__)
     // Explicitly select the single-precision CUDA library function
     sincosf(x, sine, cosine);
 #else
@@ -527,7 +527,7 @@ sincos (double x, double* sine, double* cosine)
 {
 #if defined(__GNUC__) && defined(__linux__) && !defined(__clang__)
     __builtin_sincos(x, sine, cosine);
-#elif defined(__CUDACC__)
+#elif defined(__CUDA_ARCH__)
     // Use of anonymous namespace resolves to the CUDA library function and
     // avoids infinite recursion
     ::sincos(x, sine, cosine);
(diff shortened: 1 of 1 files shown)


lgritz:
This is a correction on top of #1888. In fact, this PR puts the code closer to @timgrant's first stab at it, which I talked him out of.

timgrant:
Having finer-grained control over intrinsic selection (rather than relying on compiler flags) seems like a good thing to me. No worries at all about any "extra work", we appreciate your accommodations.

timgrant:
However, I'm curious about how you hit the NVCC error. In testoptix, NVCC is used to compile the CUDA sources, which include fmath.h and dual.h. I would like to get to the bottom of the error if possible.

lgritz:
I'm not sure what to tell you. My hypothesis was that you were still using clang to generate the ptx, but now I see that you do in fact use nvcc. I'm not sure why you hit that error.

It wasn't in OSL that I was running into the problem; I was doing something else entirely, but happened to use fmath.h in my Cuda project.

timgrant:
I see. Maybe some subset of the NVCC flags from testoptix could work for you?

lgritz:
Maybe? I looked at the OSL CMakeLists that you wrote and the flags therein, and I didn't see any that I wasn't using that looked like they would control something like that. 

I may try to track that down in detail, but I think that for now I'll just merge this version since it seems to work.

ee33:
Larry you might try NOT rearranging the code, but instead replacing __CUDACC__ with __CUDA_ARCH__ .  Does that work?  The advantage would be that nvcc would see both the device and the host versions.  (I think.)

lgritz:
Yeah, I will try this the very first second that I'm not supposed to be in 3 places at once. :-)

lgritz:
Looks like it works, I will update the PR.

lgritz:
I've completely replaced the contents of this PR. Instead of refactoring so extensively, now we merely change the use of `__CUDACC__` to `__CUDA_ARCH__` and that has the effect we were looking for.

lgritz:
So, to summarize my understanding, `__CUDACC__` just tells you if it's nvcc that is compiling, regardless of whether it's making the CPU or GPU side of the code? But `__CUDA_ARCH__` tells you which GPU architecture it's compiling for, and as a by-product, if it's not defined, that indicates that it's the "CPU side" that's being compiled for.

Before I do the merge, do @timgrant or @ee33 have an opinion on whether the two uses of `__CUDACC__` in platform.h should remain as they are or also be switched? One use defines the OIIO_HOSTDEVICE macro, and the other sets the meaning of OIIO_FORCEINLINE. I'm just hazy enough about the subtleties to not be 100% sure.

lgritz:
It seems to me that `__CUDACC__` guarding the HOSTDEVICE definition is the right thing (you want it to see those defs any time it's in nvcc), but I dunno about the inline definition.

timgrant:
Yes, that seems to be the recommended practice. I would use it for the inline definition as well.

lgritz:
@timgrant Sorry, just to be explicit, you mean that you think we should use `__CUDACC__` for both uses in platform.h (HOSTDEVICE and FORCE_INLINE), but use `__CUDA_ARCH__` for separating the code lines themselves in fmath.h?

PR #1899, opened 2018-03-14
Resolve conflict on some platforms where 'inline' and '__optnone__' are incompatible

### LINKED ISSUE #1897: '__optnone__' and 'always_inline' attributes are not compatible
When building on `OS X` (with clang), I get the following compilation error in `benchmark.h`

```
/src/include/OpenImageIO/benchmark.h:410:43: error: '__optnone__' and 'always_inline' attributes are not compatible
```

I am seeing this with `oiio 1.8.6` (however the code in question does not appear to have changed recently). EDIT: I also see the issue in 1.8.8

lgritz:
Fixes #1897

### DIFF
--- a/src/include/OpenImageIO/benchmark.h
+++ b/src/include/OpenImageIO/benchmark.h
@@ -443,7 +443,7 @@ OIIO_FORCEINLINE T const & DoNotOptimize (T const &val) {
 // If __optnone__ attribute is available: make a null function with no
 // optimization, that's all we need.
 template <class T>
-OIIO_FORCEINLINE T const & __attribute__((__optnone__))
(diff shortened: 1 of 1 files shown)


lgritz:
@SylvainCorlay can you verify that this change fixes your build?

SylvainCorlay:
@lgritz it fixes the issue that we have discussed but the build stumbles upon another issue later. Full log available here: https://travis-ci.org/QuantForge/openimageio-recipe/builds/353484690?utm_source=github_status&utm_medium=notification

lgritz:
I'm not sure why this was never merged. Doing it now.

PR #1900, opened 2018-03-14
MSVC: use /bigobj on some other large files

### LINKED ISSUE #1898: Use of /bigobj compiler flag for imagebuf.cpp
Building on windows, I am seeing the following

```
C:\bld\openimageio_1520982480036\work\oiio-Release-1.8.8\src\libOpenImageIO\imagebuf.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

The bigobj compilation flag is already used for [`/src/libOpenImageIO/CMakeLists.txt#L153`](https://github.com/OpenImageIO/oiio/blob/e6d722b4b93986ea3cd8d0ff13e82e7b76d0b089/src/libOpenImageIO/CMakeLists.txt#L153)

lgritz:
Fixes #1898

### DIFF
--- a/src/libOpenImageIO/CMakeLists.txt
+++ b/src/libOpenImageIO/CMakeLists.txt
@@ -151,6 +151,9 @@ endif ()
 
 if (MSVC)
     set_property (SOURCE imagebufalgo_pixelmath.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " /bigobj ")
+    set_property (SOURCE imagebufalgo_addsub.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " /bigobj ")
+    set_property (SOURCE imagebufalgo_muldiv.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " /bigobj ")
+    set_property (SOURCE imagebuf.cpp APPEND_STRING PROPERTY COMPILE_FLAGS " /bigobj ")
 endif ()
 
 if (WIN32)


lgritz:
I've already merged this and backported to RB-1.8, but thinking about it more, @SylvainCorlay, is the best policy going forward to use /bigobj across the board for Windows? Is there any advantage to only doing it for the 2 or 3 files that exceed the default limit?

SylvainCorlay:
Thanks for fixing this!

I don't know about using `/bigobj` across the board in general, sorry. I would say that if the current solution works, I am fine with it.

SylvainCorlay:
I am missing the `/bigobj` on other files apparently.

At the moment, this is failing on `imagebufalgo_copy`.

SylvainCorlay:
I confirm that adding it also to `imagebufalgo_copy.cpp` resolves the issue.

lgritz:
OK, I'll submit a patch that adds that.

What version of MSVC are you using? Is there something unusual about your configuration? 

I'm not sure I understand why you are hitting this, but everybody else is not, and why we don't see a build failure from this on our Appveyor builds. Do you have a hypothesis that explains this?

SylvainCorlay:
> I'm not sure I understand why you are hitting this, but everybody else is not, and why we don't see a build failure from this on our Appveyor builds. Do you have a hypothesis that explains this?

Me neither. I am also building on appveyor, with MSVC 19.0.24215.1

The logs for the last unsuccessful build are here:
https://ci.appveyor.com/project/QuantStack/openimageio-recipe/build/1.0.10

The logs for the first successful build (adding one more bigobj) are here:
https://ci.appveyor.com/project/QuantStack/openimageio-recipe/build/1.0.14

Basically, I build a conda package for 1.8.9 patching the source with the `/bigobj` options in the `CMakeLists`.

PR #1902, opened 2018-03-15
More care with /bigobj on Windows

lgritz:
Expand the set of files to all the imagebufalgo_*.cpp, which are the most likely files to need it because they tend to have template expansion for big sets of data type pairings.

Does anybody know if there is an advantage to limiting it to just the files that most need it? Or is it better to make it totally foolproof by using /bigobj for all our modules on Windows?

### DIFF
--- a/src/libOpenImageIO/CMakeLists.txt
+++ b/src/libOpenImageIO/CMakeLists.txt
@@ -150,10 +150,15 @@ endif ()
 
 
 if (MSVC)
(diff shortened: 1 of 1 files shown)


lgritz:
Does this seem reasonable to you, @SylvainCorlay ?

SylvainCorlay:
> Does this seem reasonable to you, @SylvainCorlay ?

Yes, totally!

lgritz:
OK, merged, and also backported to RB-1.8.

SylvainCorlay:
Thanks! I will remove the patching of the source for the conda package for the next release.

PR #1906, opened 2018-03-22
Move ROI from imagebuf.h to imageio.h

lgritz:
This sets us up for planned changes to use the ROI class more extensively in other imageio.h declarations.

Also add roi(), roi_full(), set_roi(), and set_roi_full() methods to ImageSpec.


src/include/OpenImageIO/imageio.h
```diff
@@ -416,10 +532,45 @@ class OIIO_API ImageSpec {
...
+    }
+
+    /// Return full/display window for this ImageSpec as a ROI.
+    ROI roi_full () const {
+        return ROI (full_x, full_x+full_width, full_y, full_y+full_height,
+                    full_z, full_z+full_depth, 0, nchannels);
+    }
+
+    /// Set pixel data window for this ImageSpec to a ROI.
+    /// Does NOT change the channels of the spec, regardless of newroi.
```
fpsunflower: I'm sure you have a good rationale for ignoring the channel count, but it seems a bit asymmetric.

Perhaps it deserves a bit more explanation in the comment so the user understands why the chbegin/end doesn't get used?

src/include/OpenImageIO/imageio.h
```diff
@@ -107,6 +107,122 @@ typedef ParamValueList ImageIOParameterList;
...
+    friend std::ostream & operator<< (std::ostream &out, const ROI &roi) {
+        out << roi.xbegin << ' ' << roi.xend << ' ' << roi.ybegin << ' '
+            << roi.yend << ' ' << roi.zbegin << ' ' << roi.zend << ' '
+            << roi.chbegin << ' ' << roi.chend;
+        return out;
+    }
+};
+
+
+#if 0
```
fpsunflower: Did you mean to leave this ifdef'd out?

src/include/OpenImageIO/imageio.h
lgritz: The short answer is that I just moved these from imagebuf.h/imagebuf.cpp to imageio.h, I didn't really think about their behavior.

It's a good question, though. The current use pattern is to use an ROI to set the x/y/z dimensions. I don't know how careful the callers are about the channel range or what it would break. I'll see what happens, and at worst, just make the docs clearer.

src/include/OpenImageIO/imageio.h
lgritz: I did, but now I can't remember why.

lgritz:
Pushed update:

* Docs fixup
* constexpr ROI
* Leave the set_roi to not affect channels, but clarify the comments a bit that this was the design.

src/include/OpenImageIO/imageio.h
```diff
@@ -107,6 +107,122 @@ typedef ParamValueList ImageIOParameterList;
...
+    constexpr int height () const { return yend - ybegin; }
+    constexpr int depth () const { return zend - zbegin; }
+
+    /// Number of channels in the region.  Beware -- this defaults to a
+    /// huge number, and to be meaningful you must consider
+    /// std::min (imagebuf.nchannels(), roi.nchannels()).
+    constexpr int nchannels () const { return chend - chbegin; }
+
+    /// Total number of pixels in the region.
+    OIIO_CONSTEXPR14 imagesize_t npixels () const {
```
fpsunflower: Doesn't seem that much harder to write as a C++11 `constexpr`?

src/include/OpenImageIO/imageio.h
```diff
@@ -107,6 +107,122 @@ typedef ParamValueList ImageIOParameterList;
...
+            << roi.yend << ' ' << roi.zbegin << ' ' << roi.zend << ' '
+            << roi.chbegin << ' ' << roi.chend;
+        return out;
+    }
+};
+
+
+
+/// Union of two regions, the smallest region containing both.
+inline OIIO_CONSTEXPR14 ROI roi_union (const ROI &A, const ROI &B) {
```
fpsunflower: Likewise for these two, a few nested ternary operators should do the trick?

lgritz:
Yes, good suggestion, I was able to make those all `constexpr` without the need for 14.
Updated.

fpsunflower:
LGTM!

PR #1910, opened 2018-04-03
PNG: Fix redundant png_write_end call and missing info ptr.

### LINKED ISSUE #1607: PNG: finish_image/png_write_end is called twice
finish_image, whgich calls png_write_end(), is called:
- here https://github.com/OpenImageIO/oiio/blob/master/src/png.imageio/pngoutput.cpp#L218
- and then here https://github.com/OpenImageIO/oiio/blob/master/src/png.imageio/png_pvt.h#L613

resulting in this error when reading the resulting PNG:
`Warning: Extraneous data found after IEND`

besides, finish_image takes the png info struct as second parameter (OIIO passes NULL)

as can be seen from the libpng source code, this info struct is used to write text chunks, time chunk, and custom chunks, i.e. metadata.

lgritz:
Fixes #1607


lgritz:
@devernay Does this look like an improvement to you?

devernay:
looks fine!

PR #1913, opened 2018-04-10
Fix floating point precision issue with "--bumpslopes" exporter

lecocqp:
This patch stores the slope values in (i,j) texture space rather than in normalized s,t space to prevent floating point precision issue when dealing with very high resolution textures.

This was particularly sensitive with half precision format where the square slopes can cause floating point overflow (because the width of 1 texel is 1.0 / res_x in (s,t) space).

To retrieve the slopes in normalized (s,t) space, we just need to scale the slopes with the texture resolution as follow:

dhds = dhds *res_x
dhdt =  dhdt * res_y
dhds2 = dhds2 * res_x * res_x
dhdt2 = dhdt2 * res_y * res_y
dh2dsdt = dh2dsdt * res_x * res_y


lgritz:
Just to be clear, the res is of the top level of the MIP-map, not that it's rescaled separately for each level, right?

lecocqp:
Absolutely right!

lgritz:
OK, LGTM, merging.

PR #1915, opened 2018-04-12
Update detectplatform.mk

kasimir84:
Add aarch64, armv6l, armv7l, ppc, ppc64, ppc64le, s390, s390x to detectplatform.mk

## Description

Added detection for aarch64, armv6l, armv7l, ppc, ppc64, ppc64le, s390, s390x

### DIFF
--- a/src/make/detectplatform.mk
+++ b/src/make/detectplatform.mk
@@ -25,7 +25,23 @@ ifneq (${hw},x86)
   ifneq (${hw},x86_64)
     ifneq (${hw},i386)
       ifneq (${hw},i686)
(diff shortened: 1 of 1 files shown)


lgritz:
Sure. But please note that this is just the "makefile wrapper" for convenience (that figures out some things and launches cmake). It doesn't guarantee that the actual cmake scripts properly build for those systems (let alone that the OIIO software itself builds or runs).

lgritz:
Wait, closed? Did you do that on purpose? Are you withdrawing the PR? You don't want it any more?

Issue #1919, opened 2018-04-16
Very Slowly Opening TIFF images issue 


MyunghoonSukTL:
### Problem

**Expected & Actual behavior:**  
I have my code with oiio 1.7 for opening image files. I can open most tiff files fast without any issues, but sometimes a few tiff files saved by Adobe Photoshop take much longer time to open them. (for example, 40 or more seconds.) 
When I look into the tiffinput.cpp, some of tags set up by find_tag & get_xxx_attribute in such tiff files make longer processing time later. If I disable this block "Use the table for all the obvious things that can be mindlessly shoved into the image spec.", the tiff file is opened very fast, but image is not correct (black & white, wrong width and height, ...) because of no tiff information.
So I just suspect some of tags information in this kind of tiff. But I could not find exactly what tags are.  

Can anyone please help me to figure out what is causing this issue?

### Versions

* OIIO branch/version: 1.7.10
* OS: Mac
* C++ compiler:

lgritz:
Can you allow me to download one of the files that have this problem?

MyunghoonSukTL:
Sure. It's a big size. 108MB. Thanks!
https://www.dropbox.com/s/60kv8i1vucrjvec/slow_opening.tif?dl=0

lgritz:
Thanks, I'll look into this and get back to you here

MyunghoonSukTL:
I really appreciate it!

lgritz:
There are some interesting/unusual things about this file: It's large (you knew that -- 3556 x 4708), and the "rows per strip" is 4708 (the entire image is one single huge strip -- whereas something like 16 or 32 scanlines per strip is more typical), and it uses planarconfig "separate", and it's uncompressed.

I instrumented the code really carefully to time everything, and the bottom line is that all the time is spent inside calls to TIFFReadScanline, 50x longer than the same calls for the equivalent file I create with `tiffcp -r 32 slow_opening.tif slow_rps32.tif` (that is, an equivalent file with scanlines in 32-line strips). I tried rps=1000 and rps=100, and found that the speed seems to be proportional somehow to rps.

[...]

Issue #1754, opened 2017-08-26
Low performances when reading Planar TIF saved from photoshop


edubois:
I'm having very slow read performances when reading a RGB planar 4K TIF image. When saving the image as interleaved, it's way faster.

### Problem
Might come from libtiff, not sure. Profiling is advised.

### Steps to Reproduce

1. Create 4K noisy RGB image in photoshop
2. Save it as Uncompressed planar RGB image
3. Read it using oiio.

### Versions

* OIIO branch/version: master / all versions
* OS: Windows
* C++ compiler: Visual 2015, C++11 with AVX instructions enabled.

lgritz:
Hi, can you provide a little more detail?

Exactly what do you mean by "read it" -- what operation, specifically, and how are you timing it? How much slower is it (with times)? Are you sure that the issue is interleaved/planar versus uncompressed/compressed? Can you share an image (or give us a pure oiiotool command line to generate an image that exhibits the problem)?

lgritz:
I think that #1919 is the same issue, and PR #1925 proposes a fix.

lgritz:
I think this was fixed long ago by #1925. Closing.

PR #1925, opened 2018-04-23
TIFF: Overhaul read_native_scanlines to speed up more cases

lgritz:
Extend the recent work on reading "raw" strips and then decompressing
them ourselves concurrently, to handle "separate" planarconfig as well
as "contiguous". This required another big refactor of
read_native_scanlines.

With this change, even contig images now take the fast path for some
combiations that it didn't before: uncompressed, and zip compressed but
without the horizontal predictor selected.


lgritz:
Fixes #1919 
Fixes #1754

PR #1934, opened 2018-05-07
ImageInput and ImageOutput creators change to return unique_ptr

lgritz:
ALERT! This is a breaking change! But I think it's a simplification,
more modern code style, make using OIIO easier and less error prone.

Previously, ImageInput::create(), ImageInput::open() [the static one],
and ImageOutput::create() returned a raw pointer, allocated by the
plugin's create function, and it was up to the client code to properly
manage and free it.

On Windows, it's not safe to delete in one dll when the new came from a
different dll. That's why we had ImageInput::destroy(), but
unfortunately I realized recently that only works properly for the
readers/writers compiled into libOpenImageIO itself.  It could
potentially fail with mysterious heap corruptions if you used a true dll
plugin to implement a custom format reader/writer.

This patch proposes to change all of these functions to return a
unique_ptr with built-in deleter, so all resources are managed and will
automatically be released when the pointer goes out of scope or is reset.
Because the deleter is supplied by and in each individual module where
the new occurs, the possible heap corruptions are fixed once and for all
for ordinary C++ ImageInput & ImageOutput usage.

It touches a LOT of files, but it's repetition of conceptually simple
changes:

1. Some new type definitions are added to the ImageInput class definition:

        /// Call signature of an ImageInput deleter.
        typedef void (*Deleter)(ImageInput*);
        /// unique_ptr to an ImageInput, with custom deleter.
        using unique_ptr = std::unique_ptr<ImageInput,Deleter>;
        /// Call signature of a function that creates and returns an ImageInput*
        typedef unique_ptr (*Creator)();

2. For each plugin, the creator changes from returning a raw pointer

        OIIO_EXPORT ImageInput *tiff_input_imageio_create () { return new TIFFInput; }

   to making a unique_ptr with deleter:

        OIIO_EXPORT ImageInput::unique_ptr tiff_input_imageio_create () {
            return { new TIFFInput, [](ImageInput *in){ delete in; } };
        }

3. The signature of the static ImageInput::create() and open() methods
   change to returning a unique_ptr, for example, from

        static ImageInput *open (const std::string &filename,
                                 const ImageSpec *config = NULL);

   to

        static unique_ptr open (const std::string &filename,
                                const ImageSpec *config = nullptr);

4. App code therefore gets a lot safer and easier. Old:

       ImageInput *in = ImageInput::open (filename);
       in->read_image (...);
       in->close ();
       ImageInput::destroy (in);

   new:

       ImageInput::unique_ptr in = ImageInput::open (filename);
       in->read_image (...);
       in->close ();

   and if you are a fan of `auto` and take advantage of the fact that
   destroying an ImageInput forces a close if not already done, it's just:

       auto in = ImageInput::open (filename);
       in->read_image (...);

5. Analogous changes for ImageOutput.


src/bmp.imageio/bmpinput.cpp
```diff
@@ -38,8 +38,8 @@ OIIO_PLUGIN_EXPORTS_BEGIN
 
     OIIO_EXPORT int bmp_imageio_version = OIIO_PLUGIN_VERSION;
     OIIO_EXPORT const char* bmp_imageio_library_version () { return NULL; }
-    OIIO_EXPORT ImageInput *bmp_input_imageio_create () {
-        return new BmpInput;
+    OIIO_EXPORT ImageInput::unique_ptr bmp_input_imageio_create () {
```
fpsunflower: Just wondering if instead, we should keep on returning a plain `ImageInput` and wrap into a unique_ptr inside OIIO (possibly pointing the destructor to a `imageio_destroy` function pointer inside the image plugin.

lgritz: Yeah, we could do that. It means that every plugin must supply a new function that returns the local destroy function. You think that would be better?

fpsunflower: Yeah I think that would be clearer/safer.

lgritz:
Updated per @fpsunflower's suggestion: now the plugin's `foo_input_imageio_create()` function is reverted to return an `ImageInput*` once again, and each plugin must also supply a `foo_input_imageio_delete()` function. The assembly of these into a unique_ptr is done in ImageInput::create, not in the plugins.  (And the same for ImageOutput.)

Better?

fpsunflower:
Yes! I like this version.

fpsunflower:
LGTM!

lgritz:
OK, I'll need to touch up the docs to make it reflect the changes. 
Thanks, as usual, for the guidance on this.

lgritz:
updated, I think this is ready to go

(note for @fpsunflower : I have a separate review for Arnold touch-ups coming internally)

PR #1936, opened 2018-05-10
ffmpeg: more precise time for switching symbol to AV_CODEC_CAP_DELAY

### LINKED ISSUE #1935: Support FFMPEG master HEAD
### Problem

Currently, the ffmpeg plugin goes to the trouble of setting some macros and functions for versions 3, 4, etc. of ffmpeg.  However, ffmpeg is usually used from master HEAD and in that case the code fails with:

m_codec_cap_delay = (bool) (m_codec_context->codec->capabilities & CODEC_CAP_DELAY);

**Expected behavior:** 

CODEC_CAP_DELAY is deprecated.  It should use AV_CODEC_CAP_DELAY or set the macro as it does v4.0.

**Actual behavior:**

It did not compile.

### Steps to Reproduce

1. git clone https://git.ffmpeg.org/ffmpeg.git ffmpeg
2. compile ffmpeg
3. install it
4. point oiio to it.
5. compile oiio

### Versions

* OIIO branch/version: 1.9 
* OS: any
* C++ compiler: any

lgritz:
There are some intermediate checkins between ffmpeg 3.4 and 4.0 where my previous attempt at this broke. So switch to the new symbol as soon as it was introduced (a while back). Fixes #1935

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -128,7 +128,7 @@ inline int receive_frame(AVCodecContext *avctx, AVFrame *picture,
 // Changes for ffmpeg 4.0
 #define USE_FFMPEG_4_0 (LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(58, 18, 100))
 
-#if USE_FFMPEG_4_0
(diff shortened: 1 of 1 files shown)


ggarra13:
I verify the change works.

lgritz:
awesome, thanks, merging.

lgritz:
Also backported to RB-1.8

PR #1937, opened 2018-05-11
Adjust __GLIBCXX__ test guarding use of std::atomic_load(ptr)

lgritz:
I didn't realize that the prior number I picked inadvertently included
a 4.9 version that did not in fact support this feature.

Changing to the new value will not have any more "false positives" (error
by incorrectly thinking the feature is present), but does make a "false
negative" (avoid the feature even though it's supported) for gcc 5 and
gcc < 6.2, even though it would work there. That will still behave
correctly, though at a very slight performance penalty, but VFXPlatform
skips from gcc 4.8 to gcc 6.3, thus leapfrogging over those false
negative cases. So I'm going to not care that we get less than maximal
performance from some compiler+library versions that we are unlikely to
use.


ggarra13:
I can verify it works.

PR #1941, opened 2018-05-12
OpenEXR better UTF-8 filename support

lgritz:
We already were careful when opening the file to use the
OpenEXRInputStream to ensure proper UTF-8 filename handling.

But valid_file() used Imf::isOpenExrFile, which is not careful enough.
So we instead switch to making an IStream and then calling the version
of isOpenExrFile that takes the stream. Should be correct.


meshula:
https://github.com/openexr/openexr/issues/292

PR #1945, opened 2018-05-19
Fix unique_ptr use

lgritz:
I really over-engineered the previous commit.  The awkward
unique_ptr-with-deleter is not needed after all. I don't need the
deleter to live in every individual plugin, as long as I can ensure that
the allocation and free both happen in the same place -- libOpenImageIO.
And I can do that by overloading operator new and operator delete, then
using the standard unique_ptr flavor.

This patch backs out of much of the previous one (no individual deleter
functions are needed). Simpler. Not sure why I didn't realize earlier
that this would work.

Can anybody find a flaw in this approach?


fpsunflower:
LGTM!

PR #1946, opened 2018-05-23
Fix iv on OS X with QT5

aconty:
The viewer iv was broken under OS X with QT5. It was not showing the image or responding to user input correctly. This fixes it for me under OS X Sierra

### DIFF
--- a/src/iv/ivgl.cpp
+++ b/src/iv/ivgl.cpp
@@ -1450,14 +1450,11 @@ IvGL::load_texture (int x, int y, int width, int height, float percent)
         }
     }
 
-    // Make it somewhat obvious to the user that some progress is happening
-    // here.
-    m_viewer.statusProgress->setValue ((int)(percent*100));
-    // FIXME: Check whether this works ok (ie, no 'recursive repaint' messages)
-    // on all platforms.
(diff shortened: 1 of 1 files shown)


lgritz:
Works for me on High Sierra! I also confirmed that it doesn't break anything on Linux (in fact, I think it fixes some weird things there as well).

aconty:
I fixed it definitely by stopping progress bar updates inside the paintGL function. It seems clear from experiments and google search that it botches the GL context by repainting. I even tried with the emit signal way, but still bad.

We cannot expect that bar to show anything while we are painting the GL view anyway, can we?

Let me know if you need me to squash the two commits in one.

aconty:
And as a bonus, fixing the pixelview. It is working fine for me on OS X high sierra. Please, check on your end.

lgritz:
Brilliant! It never would have occurred to me that the progress bar would be at all related.

The colors look wrong (too dim?) in the pixelview window.

LGTM otherwise. Should we merge now (several BIG fixes) and come back to that later?

lgritz:
Merging. Will come back to the dim pixelview later. Even so, this is so much less broken than before.
Thanks, Alex!

lgritz:
Also backported to 1.8.

PR #1949, opened 2018-05-25
ImageBuf::set_origin()

lgritz:
We had set_roi_full() to change the display window metadata, needed an
equally easy way to alter the pixel data window origin. (But note that
for pixel data, it's only safe to change the origin with metadata, not
the size, which would change the whole pixel data layout, lead to
reallocations, etc.)


fpsunflower:
LGTM

PR #1950, opened 2018-05-25
IC/TS close() and close_all() methods

lgritz:
The existing methods invalidate() and invalidate_all() fully invalidate
everything the IC/TS knows about a file (or all files) -- the specs,
the pixels, etc.

But sometimes you just want to close any currently-open handles, but not
necessarily forget about what you know about those images, or any pixel
tiles in the cache. That wasn't possible before, but this patch adds the
functionality as close(filename) and close_all() methods to both IC and
TS.

One interesting use case that this enables is a render-paint loop:

    render();
    ts->close_all();
    ... here, it's ok to modify the texture files ...
    ts->invalidate_all (false);
    render();
    ... repeat

The close_all(), performed after the render, closes any open file
handles, which is especially important on Windows where the IC holding
the file open essentially locks the files. Then, right before the
next render, invalidate_all(false) will fully invalidate any files that
have been modified since they were last opened, but not any whose
modification time has not changed.

Note that the act of closing means that if additional tiles need to be
read from disk, it will need to re-open the file. But as long as
any pixels needed are already in the cache from last time, it may not
need to re-open at all.

### DIFF
--- a/src/doc/imagecache.tex
+++ b/src/doc/imagecache.tex
@@ -868,6 +868,16 @@ force is false, in actuality files will only be invalidated if their
 modification times have been changed since they were first opened.
 \apiend
 
+\apiitem{void {\ce close} (ustring filename) \\
+void {\ce close_all} ()}
+\NEW % 1.9
+Close any open file handles associated with a named file, or for all
+files, but do not invalidate any image spec information or pixels
+associated with the files.  A client might do this in order to
+release OS file handle resources, or to make it safe for other
+processes to modify cached files.
+\apiend
+
 \subsection{Seeding the cache}
 \label{sec:imagecache:api:seeding}
 \apiitem{bool {\ce add_file} (ustring filename, \\
--- a/src/doc/texturesys.tex
+++ b/src/doc/texturesys.tex
@@ -1357,6 +1357,16 @@ force is false, in actuality files will only be invalidated if their
 modification times have been changed since they were first opened.
 \apiend
 
+\apiitem{void {\ce close} (ustring filename) \\
+void {\ce close_all} ()}
+\NEW % 1.9
+Close any open file handles associated with a named file, or for all
+files, but do not invalidate any image spec information or pixels
+associated with the files.  A client might do this in order to
+release OS file handle resources, or to make it safe for other
+processes to modify cached files.
+\apiend
+
 \subsection{UDIM and texture atlases}
 \label{sec:texturesys:udim}
 The {\cf texture()} call supports virtual filenames that expand per lookup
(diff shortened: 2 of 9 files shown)


fpsunflower:
The patch looks simple enough.

But this doesn't address the issue of progressive rendering right? You might not have a render that stops to let you call `close_all()` (or it might ''eventually'' stop, but you don't want to wait until it converges to tweak your texture).

I think the way other renderers solve this is by just loading the entire file in memory on first access. Its highly likely that if you are working in photoshop on a texture you are not using a tiled/mipmapped format anyway. It would be cool to have a mode in the texture cache that knew that certain file formats are not worth "streaming".

lgritz:
Interestingly, it is already the case that if you read any untiled file (and have not turned 'autotile' on), it WILL in fact read the whole file as a single tile, and close the file when it's done.  See the tail end of ImageCacheFile::read_untiled(). The purpose for this is just that if you have the whole thing read in, there is no additional benefit to holding the file open, as there would be if you had read some tiles and might need different tiles later.

But the full generality of what you suggest is hard. Suppose you do read in the whole file. As long as you keep asking for pixels that are in cache, it will keep using them if you don't invalidate, no matter what you do to the file. So you *have* to close and/or invalidate, or it won't notice that anything has changed. The only alternative is expecting that for every texture query, it does a 'stat' on the file to see if it has changed. But stat is actually quite expensive (compared to texture reads on pixels already in the cache. That would wreck performance. And gods help you if some other process is actually modifying the files in place and you haven't closed the file first -- on Windows, it will prevent that by locking, but on Linux/OSX it will probably just confuse and corrupt the underlying format library (libtiff, libIlmImf, and so on are not hardened to files changing on disk in the middle of a read).

lgritz:
As an aside, it's worth noting that when we switched to always using progressive rendering, we DID NOT see an epidemic of renders that thrash the texture cache and are I/O bound. Which you might have expected to be a possibility because the old complete-each-bucket-in-turn strategy is more of a "depth first" tile visitation order, whereas progressive rendering "breadth first", if you know what I mean. At least for primary ray shades.

lgritz:
Idea for the future? I guess we could have a "greedy read" mode, where even for tiled images, if there is room in the cache it tries to read the entire texture at once (and as quickly as possible) and then closes. Or where for a particular texture you could kind of tell the cache to pre-fetch the whole thing at once.

This would be a stupid thing for an ordinary render, where you would expect it to result in both a lot of unnecessary I/O and also poor cache performance. But for the particular case of lookdev of one asset, where you *could* fit the whole texture set in cache at once, it wouldn't be too bad and it would mean you were never holding files open unnecessarily. You'd still need to invalidate_all(false) every pass, or periodically, to be sure you re-read any textures that changed. And you'd still need to hope you never had your paint program writing the files at the exact moment that the texture system was reading them, or else ick.

PR #1951, opened 2018-05-26
DESIGN REVIEW: Overhaul ImageBufAlgo to return ImageBuf results

lgritz:
All ImageBufAlgo functions, in both C++ and Python, have signatures that looks like this:

    bool IBA::func (ImageBuf& result, ...args..., ROI roi=All, int nthreads=0);

So, for example, to add two images A and B,

    ImageBuf Result;
    ImageBufAlgo::add (Result, A, B);

and in Python,

    Result = ImageBuf()
    ImageBufAlgo.add (Result, A, B)

There are circumstances in which this approach of writing into an existing    buffers is crucial -- it gives you completely control over allocations and    copying, lets you write partial results into an existing buffer (using    a ROI of a smaller region than the whole), and lets you accumulate results.    As an example, consider the efficiency of:

    ImageBuf images[10];
    ImageBuf sum (images[0]);
    for (int i = 1; i < 10; ++i)
        ImageBufAlgo::add (sum, sum, images[i]);

Note that there are no additional allocations inside the loop.

OK, but this is not the usual case. For the vast majority of uses, you    are just doing a simple

    ImageBuf Result;
    ImageBufAlgo::add (Result, A, B);

and especially in Python, this necessity to pre-declare the result is    very "non-Pythonic". What we really want to say is:

    Result = ImageBufAlgo.add (A, B)

So simple! And why wouldn't we want the same thing in C++?

    auto Result = ImageBufAlgo::add (A, B);

At long last we get to the point of this proposal:

This patch aims to add a variety of each IBA function that returns the    ImageBuf result, rather than take an existing IB destination as a    parameter.

We will still keep the "in-place" methods we've always had (which, as    noted above, are still useful), but for most uses, the new calls    will be preferred for their notational simplicty.

Also, I'm simultaneously replacing the functions that take a raw    `float *` for per-channel colors, with versions that instead take    an array_view<const float>, which should be safer because it guarantees    knowledge of the length of the array.

At this stage, I have only done this for add, sub, absdiff, abs, as a    way to prototype the ideas. Consider this a "design review", circulating    for comments and feedback before I undertake the work to fully overhaul    the rest of the IBA suite of functions.

Feedback?

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -68,6 +68,24 @@ class Filter2D;  // forward declaration
 /// alter all the pixels in dst.  Exceptions will be noted, including 
 /// functions that do not honor their channel range.
 ///
+/// IBA functions tend to come in two varieties:
+/// 1. They return an ImageBuf containing the result.
+///    In this case, an entirely new image will be created to hold the
+///    result. In case of error, the result image returned can have any
+///    error conditions checked with has_error() and geterror().
+/// 2. They take a first parameter that is an ImageBuf& for the destination,
+///    and return a bool indicating whether an error occurred (the message
+///    can be retrieved with dst.geterror()).
+///
+/// The first option (return an ImageBuf) is a more compact and intuitive
+/// notation that is natural for most simple uses. But the second option
+/// (pass an ImageBuf& referring to an existing destination) offers
+/// additional flexibility, including more careful control over allocations,
+/// the ability to partially overwrite regions of an existing image, and the
+/// ability for the destination image to also be one of the input images
+/// (for example, add(A,A,B) adds B into existing image A, with no third
+/// image allocated at all).
+///
 /// In general, IBA functions that are passed an initialized 'dst' or
 /// 'result' image do not reallocate it or alter its existing pixels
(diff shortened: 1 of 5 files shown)


lgritz:
An additional idea I had:

Some IBA functions come in several flavors to cover cases where input parameters may either be an ImageBuf or a per-channel constant value. For example, consider add(A,B):

    add (const ImageBuf& A, const ImageBuf& B)
    add (const ImageBuf& A, array_view<const float> B)
    add (array_view<const float> A, const ImageBuf& B)

It's even worse for mad(), which has three image-or-constant parameters, and therefore 7 useful combinations.

So I was thinking, what if there were a non-owning parameter passing class that could hold EITHER an ImageBuf reference or an array_view<const float>, then you have have just a single function signature:

    add (Image_or_Const A, Image_or_Const B)

and handle sorting out all the cases inside the function implementation. The caller doesn't have to do anything different -- sort of like a string_view or array_view, the constructors handle all the magic of sorting out what types you are passing, but it lets you have a single declaration and implementation, leaving the header (and docs) simple, clear, uncluttered, and self-documenting.

It was something I was fooling around with mentally last night, and it's growing on me.

lgritz:
Closing. I will shortly open the "real" review for this, after having continued to tinker along these lines for the last couple weekend.

PR #1955, opened 2018-05-31
Implement fast version of cbrt

fpsunflower:
This uses an approximation found in the classic Hacker's Delight book. It is slightly less accurate but roughly 3x faster than glibc on my machine.


lgritz:
LGTM!

Please wait until the CI builds have returned before merging, just to be sure one of the compiler combinations doesn't fail.

lgritz:
You didn't merge with the button on this page, did you? It really is easier!

PR #1961, opened 2018-06-09
Overhaul ImageBufAlgo API

lgritz:
We make four large changes across the board in ImageBufAlgo.

1. No more raw pointers.

All functions that take per-channel constants, or destinations for
per-channel outputs, in the form of (pointer, length) -- or worse, just
a pointer and an assumed length big enough to hold all channels -- are
being deprecated, in favor of passing span<>. This will let the IBA
function implementations check lengths and deal appropriately, also
makes it very easy for the caller to pass a single float that will be
understood to be replicated for all channels.

2. ImageBuf return results

All existing functions of the form

    bool IBA::func (ImageBuf& result, ...input args...)

are getting an additional variant,

    ImageBuf IBA::func (...input args...);

For most ordinary uses, this transforms the awkward

    ImageBuf Result;
    ImageBufAlgo::add (Result, A, B);

into the more compact and intuitive

    ImageBuf Result = ImageBufAlgo::add (A, B);

Through the magic of C++11 "move" semantics (aka "rvalue assignment"),
this does not incur an unnecessary copy of the whole ImageBuf.

The old function forms will remain as first class citizens and are stil
useful in cases where you are repeatedly doing operations (so want to avoid
repeated allocation), or you want to use ROI's to write partial results of
the operation into a sub-region of an existing image.

This carries over to the Python API as well, where the old idiom

    Result = ImageBuf()
    ImageBufAlgo.add (Result, A, B)

was especially "non-Pythonic" looking, compared to

    Result = ImageBufAlgo.add (A, B)

3. Reduce combinatorics of Image vs constant.

For some functions, such as mad(), which computes A*B+C, a whole bunch
of clutter is necessary to handle the various combinations of

    image * image + image  ->  mad (const IB&, const IB&, const IB&)
    image * const + image  ->  mad (const IB&, span<const float>, const IB&)
    image * image + const  ->  mad (const IB&, const IB&, span<>)
    const * image + const      ...
    etc.

In fact, we need 7 variants (excluding const*const+const, which is not
useful because you don't know how big the resulting image would be).

These have all been replaced by a single variant,

    mad (Image_or_Const A, Image_or_Const B, Image_or_Const C)

using a special new class just used for parameter passing,
Image_or_Const, which can "wrap" either an ImageBuf or a float or a
span<const float>, and let the implementation sort it all out
internally.  It really de-clutters the declarations! And it also forces
us to ensure in these cases that all the combinations are accounted for.

4. Some IBA functions that generate structs with data, such as compare(),
computePixelStats(), and histogram(), now directly return those structures
instead of needing them pre-declared and passed as reference paramters.
Thanks to the magic of C++11 'move' semantics, this is no longer expensive,
and it's a more intuitive way to structure the API. (The old versions of
the functions still are available, but are considered deprecated and will
be removed in future releases.)

In the process of these changes, I've done a substantial rewrite of both
the IBA and Python chapters of the docs, as well as all the verbose
comments in imagebufalgo.h itself that explains the functions.


hobbes1069:
I can try a test build if it would help but as the main consumer of OIIO on Fedora in Blender. Do you happen to know if it would be affected by this change?

lgritz:
A try would be welcome. It's always better to catch build problems in master rather than as we're preparing releases later.

I don't expect Blender to have any issues. I don't know that it uses ImageBufAlgo at all, and even if it does, I've tried to preserve (though informally deprecate) all of the old functions.

hobbes1069:
Looks like robin-map is a hard dependency? It's not in Fedora yet and the on-the-fly download failed, but builds aren't supposed to require internet when building anyway.

lgritz:
Does that mean that you also don't accept any projects whose build uses CMake's "ExternalProject_Add" feature? Would it be better (vis a vis "the rules") if the external dependencies were implemented as git submodules rather than explicit download?

hobbes1069:
I have started the packaging work for robin-map... I guess it's a good thing I checked out master so I'll be ready when 1.9 is fully released.

I tried building but it seem to accept me setting ROBINMAP_HOME_DIR or whatever it's called on the cmake command line. I was going to refactor it to look for tsl/robin_map.h first before resorting to a download. I'll do a pull request when I think it's complete.

lgritz:
OK, but my question still stands:

We have started using this idiom for directly downloading dependent packages that aren't widely (or at all) included in the usual distros, and to cover the cases where they aren't found (rather than failing a build), and to allow us to force a download and build of a specific version that's different (say, to test a new version ahead if its being in the OS distros, or different build flags).

Is this a bad idea, and we should switch now to relying on git submodule rather than either CMake ExternalProject_Add or rolling our own (as I did for the robinmap)? Or would that be equally problematic for you?

hobbes1069:
Network access is not allowed during official builds, so ironically a bundled copy would be preferred. 

Currently it's using a brute force download instead of ExternalProject_Add and doesn't really give you a way of using a system installed version. Since I'm already touching it I was going to rework it to use ExternalProject_Add but that still only helps people doing personal builds.

lgritz:
The only reason it forces a download is because robinmap was such an obscure project that I assumed it wasn't in any distros.

We have a similar way of getting pybind11, but unless you "force" a download, it does the usual search of system libs to find it first (so, for many/most people, it doesn't need to download).

I'm trying to move away from "bundled" copies of anything. (Plus, depending on something to show up at build time, rather than be bundled into our project, makes a variety of versioning and licensing issues a lot easier to maintain.)

For now, let's just edit the parts of externalpackages.cmake to make robinmap more similar to how we handle pybind11 -- no download if it can be found on the system. And if you want to add this or any other dependent project to Fedora so that we never trigger build-time downloads, that's fine. In the long term, I'm still considering switching from explicit download to relying on git submodule, though that works fine for git-based projects and won't cover all bases if we have dependencies that aren't on git mirrors somewhere. Dunno if that makes anything easier for you or just pushes the headache around to a different spot.

PR #1962, opened 2018-06-11
Update externalpackages to look for robin-map first, and download sec…

hobbes1069:
…ond using ExternalProject_Add.

## Description

Look for robin-map to be locally installed first, then drop back to downloading using ExternalProjects_Add.

## Tests

Tested building with and without a local package of robin-map installed.

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -530,28 +530,24 @@ endmacro()
 set (BUILD_ROBINMAP_VERSION "1e59f1993c7b6eace15032f38b5acb0bc34f530b" CACHE STRING "Preferred Tessil/robin-map version, of downloading/building our own")
 
 macro (find_or_download_robin_map)
-    # Download the headers from github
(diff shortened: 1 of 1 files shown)


hobbes1069:
Noticed after the update ROBINMAP_INSTALL_DIR is superfluous.

lgritz:
It's failing the CI on both platforms.

If you don't want to fully debug whatever's going wrong with the ExternalProject_Add, I would recommend putting it back as it was before, and making just the minimal change to make it more like the code immediately above it, for pybind11, that checks for presence before downloading.

Also I'm happy to give it a try if you don't have time to fool around with this.

lgritz:
In fact, let me do that, I think I can have it fixed in just a couple minutes.

lgritz:
I propose #1964 as an alternative, and it seems to be passing the CI tests.

lgritz:
Closing in favor of #1964

PR #1963, opened 2018-06-12
[cmake] ffmpeg ensure minimal required version

fabiencastan:
## Description

On older platforms (eg ubuntu 14.04) ffmpeg is detected, so it builds the plugin but the build fail as does not meet the minimum version.

## Checklist:

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,7 +42,7 @@ NEW or CHANGED dependencies since the last major release are **bold**.
      * **NumPy**
      * **pybind11** (but OIIO will auto-download it if not found)
  * libRaw >= 0.17 ("RAW" image reading will be disabled if not found)
- * ffmpeg >= 3.0 (tested through 4.0)
+ * ffmpeg >= 2.6 (tested through 4.0)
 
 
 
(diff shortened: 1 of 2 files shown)


lgritz:
I think we support as far back as 2.6 (in fact, we have to, that's what I have installed at work).

lgritz:
I don't think this really works, except in cases where you have pkgconfig.
On OSX, or on any self-built ffmpeg that doesn't have pkgconfig info, it will fail to find it, because FindFFmpeg.cmake doesn't do a proper extraction of the version.

lgritz:
I take back that last comment... it does seem to work on OSX.
Maybe this is ok, though I think you should change the version from 3.0 to 2.6.

fabiencastan:
But [INSTALL.md](https://github.com/OpenImageIO/oiio/blob/master/INSTALL.md) says:
`ffmpeg >= 3.0 (tested through 4.0)`

lgritz:
Yeah, it does. But in retrospect, now that I'm examining both the code and the version I have installed at work, it seems that we actually support 2.6+.

lgritz:
Ping. Care to update for 2.6 and I'll merge?

fabiencastan:
Sorry for the delay. I made the change and updated the doc accordingly.

lgritz:
LGTM, thanks.

PR #1964, opened 2018-06-12
Modify robin-map download to look in system first

lgritz:
Instead of downloading unconditionally, look for system install.

Sorry, this looks like a bigger change than it really is. Most of the "changed lines" are just indented differently because they are enclosed in an `if`.


hobbes1069:
Had some other packages that needed attention. Testing now.

hobbes1069:
Ok, build completed. There are a lot of warnings generated with GCC 8.1 but not related to this PR so I'll post it separately.

PR #1965, opened 2018-06-13
Texture map overscan + clamp

lgritz:
Add a test case for overscanned texture with "clamp" wrap mode. It worked fine on the runtime side.  The maketx side needed a minor fix to know that overscan was ok with clamp (instead of forcing black wrap for all overscan cases).


fpsunflower:
LGTM

PR #1968, opened 2018-06-14
More flexibility for IC::get_image_info / TS::get_texture_info

lgritz:
What happens when you use IC/TS to retrieve metadata and get the type
slightly wrong? Like if the actual metadata is an OpenEXR box2f (which
OIIO represents as TypeDesc(FLOAT,VEC2,NOXFORM,2)), but you have asked
to retrieve it as float[4] (OIIO TypeDesc(FLOAT,4)).

It was failing, because the array lengths didn't match. But I think it
should succeed because they are both 4 floats underneath. This is
especially important if the query is coming from OSL, which doesn't have
a vec2 or a box2f type, and the only way you can ask for this metadata
would be into a float[4].

Simple fix: check for compatibility of basetype and basevalues(), not
for matching arraylength().


fpsunflower:
LGTM

PR #1972, opened 2018-06-21
Fix potential field3d mutex deadlock

lgritz:
Field3DInput::valid_file locked, then called open() which also locked.
How did this ever work? Yet, this was only symptomatic in one internal
renderer test, and only when OIIO was compiled in DEBUG mode.

The simple solution is just to change to a recursive lock. Field3d isn't
an important enough file type to try to be more clever than that.


fpsunflower:
LGTM

PR #1973, opened 2018-06-21
More modern way to test endianness.

lgritz:
Replicate the `endian` enum class that will be in C++20 (eventually we'll
eliminate our own) instead of nonstandard tricks.


fpsunflower:
LGTM!

PR #1979, opened 2018-06-28
Added "zips" to the list of compressors in the OpenEXR plugin documentation

apetrynet:
## Description

I had to create an EXR with single scanline zip compression today and I couldn't find how to do it in the documentation. After some searching online I found the `zips` option. 
This patch adds `zips` to the list of compressors in the built in plugin part of the docs.

PS! Apologize for the unnecessary commit. My intentions were good.. 

## Tests

I have not added any tests for this patch. Neither have I built OIIO with the patch unfortunately.

## Checklist:


lgritz:
So the functionality was there, it was just the documentation that neglected to mention it?

apetrynet:
Yes, exactly. I checked both the 1.8.x version we're using and master and both were lacking the zips compression in the documentation.

lgritz:
Great, thanks for the fix!

PR #1980, opened 2018-07-04
Select gcc string ABI, etc.

lgritz:
Add a build-time option to select which std::string ABI to use.
Building with GLIBCXX_USE_CXX11_ABI=1 uses the new (C++11 compliant) ABI,
building with =0 uses the traditional gcc string ABI for compatibility.
This only affects gcc builds.

Also automatically define _GLIBCXX_SANITIZE_VECTOR=1 when compiling for
sanitize mode, it turns on some extra instrumentation to detect running
off the end of a std::vector but still within its allocated capacity.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -5,7 +5,7 @@ message (STATUS "CMAKE_CXX_COMPILER is ${CMAKE_CXX_COMPILER}")
 message (STATUS "CMAKE_CXX_COMPILER_ID is ${CMAKE_CXX_COMPILER_ID}")
 
 set (USE_CPP 11 CACHE STRING "C++ standard to prefer (11, 14, 17, etc.)")
-option (USE_LIBCPLUSPLUS "Compile with clang libc++")
+option (USE_LIBCPLUSPLUS "Compile with clang libc++" OFF)
 set (USE_SIMD "" CACHE STRING "Use SIMD directives (0, sse2, sse3, ssse3, sse4.1, sse4.2, avx, avx2, avx512f, f16c, aes)")
 option (STOP_ON_WARNING "Stop building if there are any compiler warnings" ON)
 option (HIDE_SYMBOLS "Hide symbols not in the public API" OFF)
@@ -21,6 +21,7 @@ option (CLANG_TIDY "Enable clang-tidy" OFF)
 set (CLANG_TIDY_CHECKS "-*" CACHE STRING "clang-tidy checks to perform")
 set (CLANG_TIDY_ARGS "" CACHE STRING "clang-tidy args")
 option (CLANG_TIDY_FIX "Have clang-tidy fix source" OFF)
+set (GLIBCXX_USE_CXX11_ABI "" CACHE STRING "For gcc, use the new C++11 library ABI (0|1)")
 
 # Figure out which compiler we're using
 if (CMAKE_COMPILER_IS_GNUCC)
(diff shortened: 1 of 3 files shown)


src/cmake/compiler.cmake
```diff
@@ -5,7 +5,7 @@ message (STATUS "CMAKE_CXX_COMPILER is ${CMAKE_CXX_COMPILER}")
 message (STATUS "CMAKE_CXX_COMPILER_ID is ${CMAKE_CXX_COMPILER_ID}")
 
 set (USE_CPP 11 CACHE STRING "C++ standard to prefer (11, 14, 17, etc.)")
-option (USE_LIBCPLUSPLUS "Compile with clang libc++")
+option (USE_LIBCPLUSPLUS "Compile with clang libc++" OFF)
```
fpsunflower: Off by default here, but set to "1" in the helptext above.

src/cmake/compiler.cmake
```diff
@@ -192,6 +193,13 @@ if (USE_LIBCPLUSPLUS AND CMAKE_COMPILER_IS_CLANG)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
 endif ()
 
+# GCC 5+: honor build-time option for whether or not to use new string ABI
+if (CMAKE_COMPILER_IS_GNUCC AND NOT ${GCC_VERSION} VERSION_LESS 5.0)
+    if (NOT ${GLIBCXX_USE_CXX11_ABI} STREQUAL "")
+        add_definitions ("-D_GLIBCXX_USE_CXX11_ABI=${GLIBCXX_USE_CXX11_ABI}")
```
fpsunflower: I don't think this is related to which compiler you use, just if you are using `libstdc++` or not.

It applies just as well to clang or others.

src/cmake/compiler.cmake
```diff
@@ -283,6 +291,10 @@ if (SANITIZE AND (CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG))
         set (SANITIZE_LIBRARIES "asan;pthread")
         # set (SANITIZE_LIBRARIES "asan" "ubsan")
     endif()
+    if (CMAKE_COMPILER_IS_GNUCC)
+        # turn on glibcxx extra annotations to find vector writes past end
+        add_definitions ("-D_GLIBCXX_SANITIZE_VECTOR=1")
```
fpsunflower: Same with this one, it works for clang too, its a `libstdc++` feature, not a gcc one.

src/cmake/compiler.cmake
lgritz: You mean the helptext in the Makefile? You may consider this blasphemy, but those tend to show how to change the defaults, they don't give examples of redundantly setting the defaults. I know that may be backwards, but if we don't like it, we should change it in a separate pass.

src/cmake/compiler.cmake
lgritz: Yes, that's how I originally had it! And it caused endless heartache and build breaks when using clang. The only way I could find to keep this a small task was to restrict it to gcc and resolve to return to the clang issue if and when it popped up as a real problem.

fpsunflower: What kind of errors where you getting?

Maybe leave a comment to that effect at least so we remember why its scoped with the compiler?

lgritz: I was getting errors like headers were strangely mismatched.

I added a big FIXME comment here, so we'll know to come back to this spot if a case arises where we need to fix this for clang.

fpsunflower:
LGTM!

PR #1986, opened 2018-07-18
RefCnt enhancement: release()

lgritz:
refcntptr::release() sets the pointer to null and reduces the ref count, returns the original pointer, but does NOT free the object even if the new ref count is 0.

The motivation for this will be in a forthcoming OSL PR.

### DIFF
--- a/src/include/OpenImageIO/refcnt.h
+++ b/src/include/OpenImageIO/refcnt.h
@@ -118,7 +118,8 @@ public:
     T* release () {
         T* p = m_ptr;
         if (p) {
-            p->_decref ();
+            if (! p->_decref ())
+                DASSERT (0 && "release() when you aren't the sole owner");
             m_ptr = nullptr;
         }
         return p;
--- a/src/include/OpenImageIO/refcnt.h
+++ b/src/include/OpenImageIO/refcnt.h
@@ -105,6 +105,19 @@ public:
         }
     }
 
+    /// Release the reference-counted pointer but DON'T free. Return the raw
+    /// pointer. Use with caution! This can cause a memory leak if the
+    /// caller doesn't eventually reassign the returned pointer to another
+    /// intrusive_ptr!
+    T* release () {
+        T* p = m_ptr;
+        if (p) {
+            intrusive_ptr_release_no_free (m_ptr);
+            m_ptr = nullptr;
+        }
+        return p;
+    }
+
     /// Swap intrusive pointers
     void swap (intrusive_ptr &r) noexcept {
         T *tmp = m_ptr; m_ptr = r.m_ptr; r.m_ptr = tmp;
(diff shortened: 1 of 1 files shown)


fpsunflower:
This seems like a very dangerous construct to expose.

lgritz:
I do document it as dangerous!

lgritz:
Pushed an update. I simplified the implementation (now all the work is done directly in `release()`, no more `intrusive_ptr_release_no_free()` helper) hopefully in a way that minimizes the "surface area" ripe for abuse. And I rewrote the comment, I think it's more clear what it does, its limitations, and the nature of the danger.

At the end of the day, what it does is convert the last managed reference to an object into an unmanaged raw pointer (without freeing the object). This is dangerous, but occasionally useful, and there isn't another easy way to do it without this method (or something equivalent).

src/include/OpenImageIO/refcnt.h
```diff
@@ -105,6 +105,25 @@ class intrusive_ptr
...
+    /// count, return the original raw pointer, but do NOT delete the object
+    /// even if the ref count goes to zero. The only safe use case is to
+    /// convert the sole managed pointer to an object into a raw pointer.
+    /// DANGER -- use with caution! This is only safe to do if no other
+    /// intrusive_ptr refers to the object (such a pointer may subsequently
+    /// reset, decrementing the count to 0, and incorrectly free the
+    /// object), and it can cause a memory leak if the caller isn't careful
+    /// to either reassign the returned pointer to another managed pointer
+    /// or delete it manually.
+    T* release () {
```
fpsunflower: This is mostly intended to be used in cases where we are sure that we are the "last" owner right?

Maybe we could assert the ref count is 1 when we start?

lgritz: Yes, the only safe case is when we are the only owner.
I can assert, but it's still not 100% thread-safe (if it's being misused) because between the time the refcnt is accessed and when it's checked, another thread could add a reference.

lgritz: Added assertion and pushed update.

fpsunflower: You are right. It seems like a `DASSERT` would at least help catch the "normal" uses of this (like in the OSL example) while occasionaly firing if you were using this in some situation you probably shouldn't.

Maybe checking for 0 after the decref is even more reliable? (since no two threads should ever "see" 0 at the same time)

fpsunflower: I think I would suggest the following:

```
if (!p->_decref ())
   DASSERT_MSG(false, "trying to use release without being the last owner");
```

lgritz: I think that's equivalent, but I kind of like that way of using the result of _decref.

lgritz: Updated again

fpsunflower: LGTM!

Issue #1994, opened 2018-08-10
Fails to build with Clang 7.0


jbeich:
CC @sambler @DimitryAndric

### Problem

**Expected behavior:**  Build succeeds.

**Actual behavior:**  

```
$ c++ -v
FreeBSD clang version 7.0.0 (branches/release_70 338892) (based on LLVM 7.0.0svn)
Target: x86_64-unknown-freebsd12.0
Thread model: posix
InstalledDir: /usr/bin
```

```c++
FAILED: src/libutil/CMakeFiles/simd_test.dir/simd_test.cpp.o
/usr/bin/c++  -DUSE_FREETYPE -DUSE_OCIO=1 -DUSE_STD_REGEX -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -Iinclude -Isrc/include -I/usr/local/include/OpenEXR -isystem /usr/local/include -O3 -DNDEBUG   -Wall -Werror -fno-math-errno -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -Wno-expansion-to-defined -std=c++11 -Wno-deprecated-register -UUSE_FIELD3D -MD -MT src/libutil/CMakeFiles/simd_test.dir/simd_test.cpp.o -MF src/libutil/CMakeFiles/simd_test.dir/simd_test.cpp.o.d -o src/libutil/CMakeFiles/simd_test.dir/simd_test.cpp.o -c src/libutil/simd_test.cpp
In file included from src/libutil/simd_test.cpp:37:
src/include/OpenImageIO/simd.h:3229:29: error: argument value 340 is outside the valid range [0, 255] [-Wargument-outside-range]
    return _mm_castsi128_ps(_mm_shuffle_epi32(_mm_castps_si128(a), _MM_SHUFFLE(i3, i2, i1, i0)));
                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/clang/7.0.0/include/emmintrin.h:4347:12: note: expanded from macro '_mm_shuffle_epi32'
  (__m128i)__builtin_ia32_pshufd((__v4si)(__m128i)(a), (int)(imm))
           ^                                           ~~~~~~~~~~
[...]

PR #1995, opened 2018-08-14
Fixes for clang 7.0 compile

lgritz:
1. New warnings about some stuff in our simd_test.

2. What I believe is an error in boost thread, just work around it for now.


lgritz:
Fixes #1994

PR #1996, opened 2018-08-25
Fix ImageBuf::contains_roi always returning true

justinfx:
## Description

While updating my [Go language bindings](https://github.com/justinfx/openimageigo) to oiio 1.7 compatability, I found that `ImageBuf::contains_roi()` was always returning true. There looked to be a bug where the `roi` input parameter was always being compared against itself instead of the `roi` of the `ImageBuf` instance.

## Checklist:


lgritz:
Ouch! That's embarrassing. LGTM, merging. Thanks!

justinfx:
Thanks Larry!

PR #2000, opened 2018-08-29
Fix uninitialized read within texture system

fpsunflower:
Found this bug with valgrind. The `std::atomic` default constructor does not initialize the memory. Therefore this array was not getting properly filled in.

I am not really sure how this didn't cause problems, it could be I am missing something subtle. But the valgrind warning definitely goes away after this fix.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -252,7 +252,10 @@ ImageCacheFile::LevelInfo::LevelInfo (const ImageSpec &spec_,
     }
     int total_tiles = nxtiles * nytiles * nztiles;
     ASSERT (total_tiles >= 1);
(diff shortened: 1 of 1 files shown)


lgritz:
The CI isn't liking that on gcc8

lgritz:
You may need to initialize it with a for loop rather than memset (the compiler will make that as efficient if it can anyway). But when it passes CI, then LGTM.

fpsunflower:
Good catch from the CI's :+1: 

I just pushed a simpler version.

fpsunflower:
FWIW, the previous code could at most report some erroneous stats, which is probably why we never really noticed any issues.

I do wonder how much overhead there is from tracking these fine grained stats though.

PR #2001, opened 2018-08-30
Fix segfault when reading gif with Comment Extension and no Comment Data

SimonBoorer:
## Description

The GIFInput::read_gif_extension method segfaults when reading a GIF that has a Comment Extension with no Comment Data. I don't know how the problem GIF was generated and unfortunately I can't share it but it seems valid and can be opened without issue in other applications.

## Tests

Prior to the change OpenImageIO was segfaulting as soon as the image was opened, now it opens without issue.

## Checklist:


lgritz:
LGTM, good catch!
Merging and also backporting into release branch...

PR #2003, opened 2018-09-01
Fix ImageSpec::erase_attribute using backwards logic for casesensitive

justinfx:
## Description

While testing my [Go language bindings](https://github.com/justinfx/openimageigo) against the 1.8 branch, I found that one of my `ImageSpec::erase_attribute` tests was failing when checking the `casesensitive=true` option. Digging into the history it looks like with the addition of the `std::regex` ifdef, a copy-paste error has left the opposite check for the boolean. So you get case-insensitive when you as for case-sensitive. Again this is only when not using boost::regex. 

One other issue I noticed, after fixing this bug, was that the `TypeDesc` is no longer honored when searching for the attribute. Another test that I have fails on this assumption, but I am not sure if this was the intended new behavior. My test tries to set an int, and then erase it as a float. Previously it will not erase the attribute, but in 1.8 the attribute is erased regardless of type.

## Checklist:

### DIFF
--- a/src/libOpenImageIO/formatspec.cpp
+++ b/src/libOpenImageIO/formatspec.cpp
@@ -351,7 +351,7 @@ ImageSpec::erase_attribute (string_view name, TypeDesc searchtype,
             flag |= boost::regex_constants::icase;
 #else
         std::regex_constants::syntax_option_type flag = std::regex_constants::basic;
-        if (casesensitive)
+        if (! casesensitive)
             flag |= std::regex_constants::icase;
 #endif
         regex re = regex(name.str(), flag);


lgritz:
I'm not sure which is more embarrassing -- that I made this cut-and-paste error, or that for 10 years I thought it was important to provide the flexibility to choose between case-senstitive/insensitive yet clearly nobody need or uses it enough to even notice when it's broken.

Thanks for the fix! I will merge and also forward-port to master.

justinfx:
I thought the same thing, that it seemed no one cared much about that feature anyway :)

On that second point, should I just remove my test that assumes the TypeDesc has any impact on whether it erases the named attribute or not?

lgritz:
I think we should restore the ability to erase only if it's a particular type of attribute. Unless you are of the opinion that nobody would ever want that and it's confusing enough to cause more trouble than it's worth.

justinfx:
I'm actually fine with it not being a thing anymore. It is just that the option is already in the signature. I was just kind of sticking to what my unit tests expected, but it seems completely reasonable to just remove named attributes and not be concerned with their type.

PR #2008, opened 2018-09-14
Use pkgconfig in FindOpenEXR.cmake if available

apetrynet:
## Description
Inspired by the FindFFmpeg.cmake and FindLibRaw.cmake files I added similar use of pkgconfig to locate IlmBase and OpenEXR include directories.
The potentially found paths are appended to `GENERIC_INCLUDE_PATHS` respecting the search order that was before.

## Tests
I have not added any tests, but it seems to build the way it should.

## Checklist:

### DIFF
--- a/src/cmake/modules/FindOpenEXR.cmake
+++ b/src/cmake/modules/FindOpenEXR.cmake
@@ -27,6 +27,13 @@ if (CMAKE_USE_PTHREADS_INIT)
     set (ILMBASE_PTHREADS ${CMAKE_THREAD_LIBS_INIT})
 endif ()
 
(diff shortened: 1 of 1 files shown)


lgritz:
This covers the header include location, but don't you also want to use the information returned by pkgconfig to also influence the finding of the libraries?

apetrynet:
Hmm.. Yes. I took a real naive approach here so missed that after I got it working. Will update and commit changes when I get to work on Monday.

Thanks!

lgritz:
LGTM, merging.
I'll backport to the release branch as well.

PR #2009, opened 2018-09-17
Reorganize how texture3d transforms points to local space of the volume.

lgritz:
What happens when texture3d is called on a point? There was a special
back door for Field3d, but this didn't generalize well. For example, we
want to support texture3d lookups from OpenVDB files (coming soon under
a separate PR -- this is a necessary prelude).

This patch:

* Makes each ImageCacheFile::SubimageInfo contain a unique_ptr to a 4x4
  matrix (so it doesn't use as much space except when needed, which is a
  small minority of all texture files). This also moves (commented-out)
  matrix info for (unsupported) shadow maps from per-file to per-subimage.

* If reading the spec for that subimage yields a "worldtocamera" or
  "worldtolocal" matrix metada, we allocate the space and stash the
  matrix there.

* The texture3d() call looks for the matrix, and if found, uses it to
  transform the input point to the local volume space. Only if this is
  not found does it check for the special field3d back door (which
  exists because f3d supports time-varying transforms that aren't a
  single matrix). If that, too fails, then no transform is applied.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -303,12 +303,6 @@ ImageCacheFile::ImageCacheFile (ImageCacheImpl &imagecache,
     m_filename = imagecache.resolve_filename (m_filename_original.string());
     // N.B. the file is not opened, the ImageInput is NULL.  This is
     // reflected by the fact that m_validspec is false.
-#if USE_SHADOW_MATRICES
-    m_Mlocal.makeIdentity();
-    m_Mproj.makeIdentity();
-    m_Mtex.makeIdentity();
-    m_Mras.makeIdentity();
-#endif
 
     // Figure out if it's a UDIM-like virtual texture
     if (! Filesystem::exists(m_filename.string())
@@ -373,7 +367,8 @@ ImageCacheFile::set_imageinput (std::shared_ptr<ImageInput> newval)
 
 
 void
-ImageCacheFile::SubimageInfo::init (const ImageSpec &spec, bool forcefloat)
+ImageCacheFile::SubimageInfo::init (ImageCacheFile &icfile,
+                                    const ImageSpec &spec, bool forcefloat)
 {
     volume = (spec.depth > 1 || spec.full_depth > 1);
     full_pixel_range = (spec.x == spec.full_x &&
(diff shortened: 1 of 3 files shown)


src/libtexture/imagecache_pvt.h
```diff
@@ -263,7 +263,13 @@ class OIIO_API ImageCacheFile : public RefCnt {
         bool is_constant_image;         ///< Is the image a constant color?
         bool has_average_color;         ///< We have an average color
         std::vector<float> average_color; ///< Average color
-        spin_mutex average_color_mutex; ///< protect average_color
+        spin_mutex mutex;               ///< protect average_color and Mlocal
+        std::unique_ptr<Imath::M44f> Mlocal; ///< shadows/volumes: world-to-local
+#if USE_SHADOW_MATRICES
```
fpsunflower: Not really related to this review, but do you think we'll ever implement shadow map lookups?

fpsunflower:
LGTM

src/libtexture/imagecache_pvt.h
lgritz: Probably not? In the ancient past, I thought we would so I made room, but then never implemented the calls. Then I #ifdef'ed out the storage since we weren't using it.

I guess I should probably just remove it now. It's not like there's a lot to re-add if we change our minds later.

src/libtexture/imagecache.cpp
```diff
@@ -440,6 +435,19 @@ ImageCacheFile::SubimageInfo::init (const ImageSpec &spec, bool forcefloat)
...
+            icfile.m_texformat == TexFormatCubeFaceShadow ||
+            icfile.m_texformat == TexFormatVolumeShadow) {
+        const ParamValue *p = spec.find_attribute ("worldtolocal", TypeMatrix);
+        if (p) {
+            Imath::M44f c2w;
+            icfile.m_imagecache.get_commontoworld (c2w);
+            const Imath::M44f *m = (const Imath::M44f *)p->data();
+            Mlocal.reset (new Imath::M44f (c2w * (*m)));
+        }
+    }
```
marsupial: Slight issue here, the call to `sb.init` from below at line:560 will occur before the call to `ImageCacheFile::init_from_spec`, so icfile.m_texformat has not been set to TexFormatTexture3d yet

marsupial: Oh that looks like it will work, thanks.
If you have a moment tomorrow, maybe can chat?
A few questions/decisions to be made and thinking I'm finally at a point where I won't sound like total buffoon.

PR #2010, opened 2018-09-19
Basic OpenVDB read support

marsupial:
## Description

Add OpenVDB reader to OIIO

## Tests

Not yet, no

## Checklist:

### DIFF
--- a/src/openvdb.imageio/openvdbinput.cpp
+++ b/src/openvdb.imageio/openvdbinput.cpp
@@ -451,8 +451,9 @@ OpenVDBInput::open (const std::string &filename, ImageSpec &newspec)
         close();
 
     auto file = openVDB(filename, this);
-    if (!file || !file->isOpen())
+    if (!file)
         return false;
+    ASSERT(file->isOpen());
 
     try {
         for (io::File::NameIterator name = file->beginName(), end = file->endName();


Makefile
```diff
@@ -119,6 +119,14 @@ ifneq (${FIELD3D_HOME},)
 MY_CMAKE_FLAGS += -DFIELD3D_HOME:STRING=${FIELD3D_HOME}
 endif
 
+ifneq (${USE_OPENVDB},)
+MY_CMAKE_FLAGS += -DUSE_OPENVDB:BOOL=${USE_OPENVDB}
+endif
+
+ifneq (${OPENVDB_HOME},)
+MY_CMAKE_FLAGS += -DOPENVDB_ROOT:STRING=${OPENVDB_ROOT}
```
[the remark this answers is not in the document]
marsupial: OPENVDB_ROOT will do more interesting things...my Makefile knowledge pretty much starts and ends with these changes so probably wise to be thorough there.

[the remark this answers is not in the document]
lgritz: Oh, I just meant to make the names consistent. If the important thing on the cmake side is OPENVDB_ROOT, let's use the same name on the makefile wrapper side.

CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ set (PYTHON_VERSION "2.7" CACHE STRING "Target version of python to find")
 option (PYLIB_INCLUDE_SONAME "If ON, soname/soversion will be set for Python module library" OFF)
 option (PYLIB_LIB_PREFIX "If ON, prefix the Python module with 'lib'" OFF)
 option (USE_FIELD3D "Use Field3D if found" ON)
+option (USE_OPENVDB "Use OpenVDB if found" ON)
 option (USE_FFMPEG "Use FFmpeg if found" ON)
```
lgritz: Should OPENVDB_ROOT be directly exposed as a cached variable here, so that it's more apparent for people running 'ccmake', for example, to interactively see and edit all the build-time options?

site/spi/Makefile-bits-arnold
```diff
@@ -108,6 +111,8 @@ ifeq (${SP_OS}, rhel7)
     MY_CMAKE_FLAGS += \
 	-DOCIO_PATH=${OCIO_PATH} \
 	-DFIELD3D_HOME=${FIELD3D_HOME} \
+    -DOPENVDB_LOCATION=${OPENVDB_LOCATION} \
+    -DTBB_LOCATION=${TBB_LOCATION} \
```
lgritz: I think we should expose a TBB_LOCATION cached variable (and a way to set it via the makefile wrapper?). And maybe a USE_TBB option?

src/cmake/externalpackages.cmake
```diff
@@ -7,6 +7,7 @@ if (NOT VERBOSE)
     set (DCMTK_FIND_QUIETLY true)
     set (FFmpeg_FIND_QUIETLY true)
     set (Field3D_FIND_QUIETLY true)
+    set (OpenVDB_FIND_QUIETLY true)
```
lgritz: Also TBB_FIND_QUIETLY

marsupial:
As I remember TBB was required as somewhere deep in VDB headers it was being pulled in.
But I'll investigate and make sure it wasn't from inclusion of something not included here.

Makefile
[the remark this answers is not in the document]
marsupial: Yes I got that, but wasn't the original just wrong as well:
make OPENVDB_HOME=/somewhere  would have done nothing useful to MY_CMAKE_FLAGS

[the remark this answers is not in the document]
lgritz: Yeah, I think it's all fine as long as it's always ROOT

marsupial:
TBB indeed needs to be visible for compile & link.
Split out TBB to be a separate CMake object.
Added cached variables OPENVDB_LOCATION for TBB_LOCATION.

Might still need to be an update to travis & appveyor to run the test.

src/cmake/externalpackages.cmake
```diff
@@ -301,6 +303,70 @@ endif ()
...
+    if (oiio_tbb_verbose)
+        message (STATUS "TBB_LOCATION=${TBB_LOCATION}")
+    endif ()
+    find_package (TBB)
+
+    if (TBB_INCLUDE_DIR AND Tbb_TBB_LIBRARY)
+        set (TBB_FOUND TRUE)
+        if (oiio_tbb_verbose)
+            message (STATUS "Intel TBB includes = ${TBB_INCLUDE_DIR}")
+            message (STATUS "Intel TBB library = ${Tbb_TBB_LIBRARY}")
```
lgritz: Is that the correct capitalization?

lgritz: Never mind, I see that it is. Weird.

marsupial: Agreed

marsupial:
CI seems to show not having VDB doesn't break the build, is it worth trying to get it supported there?

lgritz:
LGTM, except that I tried to check on my laptop and it failed to find TBB. Offline, Roman and I discussed and he's going to substitute a different FindTBB.cmake, the one from USD, which seems to work better.

lgritz:
Correct, not having VDB should not break the build.

I'll go back later and adjust the travis files to make sure openvdb is installed (at least on Mac, where it's easy thanks to Homebrew) so that this feature is exercised. You can try that, too, if you want, but no big deal if you want to move on and I'll take care of it separately.

marsupial:
Seems good here, probably worth trying to make sure it works as expected on OS X.
Next travis build should give it a try..

lgritz:
On OSX (both on my laptop and Travis), its finding TBB but not finding OpenVDB. Not sure why.

marsupial:
Could **NO_DEFAULT_PATH** in FindOpenVDB.cmake:66 be the culprit? Seems I stripped that from the header portion but not the library.

marsupial:
Updated test to not use blosc, are you going to push something on top with the working CMake or is that something to be done here?

lgritz:
I'm happy to do it (since I already made the changes on my end to test it).
Should I merge what you have and then send a separate PR for that? Or should I push back to your branch to make it all one merge (you need to have checked the box that lets me push to your branch, for that to work)?

marsupial:
I'll push the button..

lgritz:
Builds on OSX and passes the unit test!
I'm going to add some touch-ups and then push back so you can see and approve.

marsupial:
Great thanks...Out of curiosity the VDB library was 4 or 3?

lgritz:
5

lgritz:
OK, I pushed some minor fixes on top of yours. A lot was just minor cleanup I would have done at some later time as I touched up the docs or prepared for a release.

* The original FindOpenVDB.cmake was unreliable. Replaced with one found  elsewhere on the net (from Blender, bearing a BSD license).

* Makefile wrapper: directly use OPENVDB_ROOT_DIR

* Update INSTALL.md and oiiointro.tex to document the new optional      dependencies and how we incorporated those two new Find*.cmake files      from other projects (with appropriate licenses).

* Simplification of the TBB and OpenVDB sections of externalpackages.cmake,      they didn't need to be quite so verbose, didn't need to -UUSE_BLAH if not      found, did not need to set BLAH_FOUND (done by the FindBlah itself),      and I added minimum versions of both packages based on this year's      VFX Platform guidelines (as a forward-looking new feature in master,      I don't feel the need to verify support of older versions).

* Update oiio docs chapter on supported formats to include OpenVDB.

* Some extremely minor formatting and whitespace fixes in      openvdbinput.cpp to conform to prevailing project style.

* Simplification of the SPI-specific Makefile-bits.

* Minor touch of field3d to add "worldtolocal" metadata to conform to the      new convention we're using with vdb.

src/openvdb.imageio/openvdbinput.cpp
```diff
@@ -413,17 +442,18 @@ OpenVDBInput::readMetaData (const openvdb::GridBase& grid,
...
+
 bool
 OpenVDBInput::open (const std::string &filename, ImageSpec &newspec)
 {
     if (m_input)
         close();
 
     auto file = openVDB(filename, this);
-    if (!file)
+    if (!file || !file->isOpen())
```
marsupial: The `ASSERT` removed was because the contract from the `openVDB` function is not to return a `io::File` unless `isOpen` has returned true already.  In the scheme of what's about to be done the additional check likely adds nothing, but I did like the ASSERTION of that promise.

src/cmake/externalpackages.cmake
```diff
@@ -335,35 +324,26 @@ endif ()
...
-        set (OPENVDB_FOUND FALSE)
+        message (STATUS "OpenVDB not found, OPENVDB_ROOT_DIR='${OPENVDB_ROOT_DIR}'")
     endif ()
 else ()
-    add_definitions ("-UUSE_OPENVDB")
-    if (USE_OPENVDB AND NOT TBB_FOUND)
-        set (oiio_vdb_why ", could not find Intel TBB")
-    endif ()
-    message (STATUS "OpenVDB will not be used${oiio_vdb_why}")
+    message (STATUS "OpenVDB will not be used, could not find Intel TBB")
```
marsupial: Won't this spew the message when invoked with `-DUSE_OPENVDB=OFF`

marsupial:
Everything seems to work as was, so I'd say good to go.
Left two little comments above, but totally fine if you want to show them to the waste-bin.

src/openvdb.imageio/openvdbinput.cpp
lgritz: I see. I was just trying to reduce the number of cases where we couldn't recover. From the point of an app, I think it's preferable to return "I can't open this file" whether it be that it's malformed, or OpenVDB is somehow broken, then to terminate the whole process. 

How about if I use DASSERT, which will do the hard crash/error when in debug mode? But for release build, the app will get a failure state rather than terminate?

PR #2012, opened 2018-09-21
ImageCache: add a way to add_tile without a pixel copy

lgritz:
The point of this feature is that your app has a buffer of pixels, and
you want the IC or TextureSystem to use that buffer directly, without
copying the texels or owning the memory. It should also not count the
memory against the limits of the cache.

See imagecache_test.cpp, the test I added as text_app_buffer() which
pretty much spells out how to do it from the app point of view.

From an API standpoint, all that changes is an optional argument to
add_tile that lets you disable the copy.

You need to make it a tile of an image in the cache, though, and the
obvious no-drama choice is just to use a "null" ImageInput as a
placeholder.  In the process, I noticed two helpful enhancement I could
make to the null ImageInput better:

1. Make sure that passing a "config" spec results in the right resolution
   and data type -- in fact, there was a bug where the config's res would
   be overwritten by a 1024x1024x4xUINT8 default. Oops. So only overwrite
   those fields if they are not set in the config.

2. What if you want one of these fake inputs but you don't want the name
   of the texture to be "blah.null"? So I added it so that if the config
   has metadata hint "null:force" set to nonzero, it will always succeed
   in creating a null ImageInput.

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -355,12 +355,15 @@ public:
     /// given format, with supplied strides) which will be copied and
     /// inserted into the cache and made available for future lookups.
     /// If chend < chbegin, it will add a tile containing the full set of
-    /// channels for the image.
+    /// channels for the image. Note that if the 'copy' flag is false, the
+    /// data is assumed to be in some kind of persistent storage and will
+    /// not be copied, nor will its pixels take up additional memory in the
+    /// cache.
     virtual bool add_tile (ustring filename, int subimage, int miplevel,
                      int x, int y, int z, int chbegin, int chend,
                      TypeDesc format, const void *buffer,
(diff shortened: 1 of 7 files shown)


aconty:
How would we get the ImageCache from an existing texture system?

lgritz:
Well, when you create a TextureSystem, you either pass in a specific ImageCache, or else it will use the "global" shared IC (Arnold does the latter).

ImageCache::create() will return the shared cache, which will be the same one the shared TextureSystem used.

lgritz:
In thinking about this, I realize that the way @aconty is about to use this, he's going to run into trouble with the renderer rejecting the file because it looks not-mipmapped. I need to doctor something up to ensure that doesn't happen, so let me tinker with this just a bit more.

lgritz:
I'm going to merge what I've got here, then follow up with a second patch to describe the mip problem.

Issue #1991, opened 2018-08-02
Crash on exit with OSX


ThiagoIze:
While upgrading to the master branch, I came across a very rare crash (sometimes it'll take 10s of thousands of runs before a crash) that happened upon shutdown on my high sierra OSX machine:
```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```
This is a bug in OIIO.  The following patch, which is perfectly valid code, allows it to happen on almost every run:
```diff
diff --git a/src/libutil/thread.cpp b/src/libutil/thread.cpp
index ca2d8024..fe758703 100644
--- a/src/libutil/thread.cpp
+++ b/src/libutil/thread.cpp
@@ -126,7 +126,9 @@ public:
 
     // the destructor waits for all the functions in the queue to be finished
     ~Impl () {
+        printf("~thread_pool::Impl called\n");
         this->stop(true);
+        printf("~thread_pool::Impl manual destruction finished. Automatic destructors happen now\n");
     }
 
     // get the number of running threads in the pool
@@ -300,8 +302,11 @@ private:
                         isPop = this->q.pop(_f);
                 }
                 // the queue is empty here, wait for the next command
+                printf("set_thread: the queue is empty here, wait for the next command\n");
+                sleep(1); // wait for mutex to be destroyed
                 std::unique_lock<std::mutex> lock(this->mutex);
                 ++this->nWaiting;
+                printf("this->cv.wait\n"); // we made it through with no crashes!
                 this->cv.wait(lock, [this, &_f, &isPop, &_flag](){ isPop = this->q.pop(_f); return isPop || this->isDone || _flag; });
                 --this->nWaiting;
                 if (!isPop)
@@ -322,6 +327,9 @@ private:
     std::atomic<bool> isStop;
     std::atomic<int> nWaiting;  // how many threads are waiting
     int m_size {0};             // Number of threads in the queue
+    struct Foo{ ~Foo() {
+       printf("mutex was destructed. pausing ~thread_pool::Impl() for 1s!\n"); sleep(1); }
+    } this_gets_destroyed_after_mutex;
     std::mutex mutex;
     std::condition_variable cv;
```

[...]

PR #2013, opened 2018-09-22
Fixes rare crash when lowering number of threads

ThiagoIze:
## Description
More details, including a hopefully easy repro that works on OSX, are in #1991.

The way this fix works is that instead of detaching threads, which by bad luck might stay running and access thread pool data after the thread pool is destroyed, we stay attached and move the threads into a list of threads that we have to explicitly wait on to be finished, just like we already do for the active threads.

## Tests

The crash is very hard to reproduce without adding a 1s sleep, and even then I suspect the 1s worked for me because my process runs in under 1s. Maybe you'll need to sleep for at least as long as it takes to go from lowering the number of threads to terminating the process so that we can be assured the detached threads survive to the end?

## Checklist:

- [ x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [x ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/libutil/thread.cpp
+++ b/src/libutil/thread.cpp
@@ -159,7 +159,8 @@ public:
             else {  // the number of threads is decreased
                 for (int i = oldNThreads - 1; i >= nThreads; --i) {
                     *this->flags[i] = true;  // this thread will finish
-                    this->threads[i]->detach();
+                    this->terminating_threads.push_back(std::move(this->threads[i]));
+                    this->threads.erase(this->threads.begin() + i);
                 }
                 {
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, thanks for the fix to this tricky issue.
Merging.

lgritz:
Backport to 1.8 as well, I presume?

ThiagoIze:
I haven't tested on 1.8, but looking at the code it does seem like it would benefit as well.

lgritz:
Merged into 1.8.

Issue #1632, opened 2017-03-05
invalidate() on a cached entry doesn't seem to reset its m_configspec?


devernay:
Here's my use case, but I'm not sure if I'm doing something wrong, or OIIO does:
- I want to open a file using the OIIO cache, passing extra config parameters (for RAW files), so I'm using add_file() before get_imagespecs()
- at some point (a decoding parameter changed), I invalidate the file in the cache, and re-call add_file() with the new config spec, gefore the next get_imagespec()
- OIIO doesn't take into account the new the config spec, probably because invalidate_spec() doesn't reset m_configpec

Am I doing something wrong?
Is there a way to reopen a file with a new config spec?

devernay:
btw ImageCacheFile::m_configspec is a memory lead (allocated in the constructor, never deleted)

devernay:
since m_configspec is set in the ImageCacheFile constructor, the only clean way to do it would be that invalidate actually deletes the ImageCacheFile, but it merely seems to keep it in an invalid state, so that the next add_file() does nothing.

lgritz:
I think you are basically correct about all this.

There are performance and safety reasons why we don't want to delete the whole ImageCacheFile upon invalidation.

But, I think that the real problem is that add_file, if the file already exists, just does nothing (and in particular does nothing with the config passed to it). I think a better behavior is that add_file of a file that already exists in the cache ought to invalidate it and use the passed configuration as a replacement for whatever config it may have had before.

I'll see if I can prepare a patch to address this, and also the leak that you point out.

devernay:
I think add_file should do nothing if the file already exists in the cache and is valid, but should load the new config in the file was invalidated.

If add_file invalidates the file, then I need a way in the API to know if a file is present and valid in the cache, because I don't want to invalidate the cache every time I load a file (I do add_file every time I open a file, because I have no way to know if it is already in the cache and valid)

lgritz:
OK, that makes sense to me: add_file should not invalidate, but if the file it mentions is already invalidated, it can replace the config with the new one. Working on it.

lgritz:
I think that this old issue was addressed by the fairly recent PR #2021. That change (which can be found in OIIO 2.0 as well as master) added a `replace` parameter to `add_file()`, which controls what happens if the file being added is already in the cache -- if `replace == false` (the default), the add_file is just a no-op if the file is found, whereas if `replace == true`, a found file will be invalidated and replaced according to the new creator and configuration spec.

[...]

PR #2021, opened 2018-10-01
Fix a double free of imagecache and add a replace mode for add_file()

aconty:
It is split in two commits, see the log. First one fixes a double free bug when the imagecache is provided to the texture system. The second is an option to make add_file() replace any existing entries. Useful for providing user buffers as textures that have already been queried (for instance to get a handle).


lgritz:
LGTM, thanks for the fixes. Merging.

lgritz:
Later realized that this fixes #1632.

PR #2023, opened 2018-10-02
Release an official FindOpenImageIO.cmake

lgritz:
* Make an official FindOpenImageIO.cmake, borrowed from OSL but with some
  minor changes. It will install into INSTALL_PREFIX/share/cmake/Modules

* We hope that downstream projects that have OIIO as a dependency will
  use/copy this file, rather than having a dozen or more slightly
  different versions that downstream projects all write themselves. And
  if there are any deficiencies, they will push fixes or improvements to
  this master copy.

* Change places where we expect the cmake and/or environment variable
  OPENIMAGEIOHOME to be named OPENIMAGEIO_ROOT_DIR (more standard naming
  convention by modern cmake standards).

* Rename the src/cmake/modules directory to the more standard capitalization
  src/cmake/Modules

* Some other very minor fixes to the cmake files (INSTALL_DOCS was
  declared in two different places, etc.)

(Beware, Mac developers: Renaming the modules->Modules directory may
give git a bit of a hiccup if your filesystem is case-independent. When
you switch between branches on either side of this change, you may need
to use `git checkout otherbranch --force` to ensure the old files are
removed and new names used.)


lgritz:
Withdrawing and will submit a new one.

PR #2025, opened 2018-10-03
Export symbols for Windows target

darkhorse64:


lgritz:
LGTM, will merge when CI tests are finished.

PR #2026, opened 2018-10-03
[CMake] Allow builds outside the source tree to pass testsuite.

marsupial:
## Description
Allows builds outside of the source tree to pass. Adds a CMake variable OIIO_TESTSUITE_IMAGEDIR to specify where the image test's can be run from.  The two are currently tied together, in that building from the source tree it is assumed the testsuite should be run as usual/prior.
Can probably split OIIO_TESTSUITE_IMAGEDIR to have meaning when building from the checkout if these changes are desired. 
Also part of the change is explicitly marking some of the -regular- tessuite as requiring oiio-images.

## Tests
The following are still failing locally, but I think unrelated to these changes?
oiiotool-subimage
oiiotool-text
texture-icwrite
texture-icwrite.batch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,8 @@ option (BUILD_OIIOUTIL_ONLY "If ON, will build *only* libOpenImageIO_Util" OFF)
 option (BUILD_DOCS "If ON, build documentation and man pages." ON)
 option (INSTALL_DOCS "If ON, install documentation and man pages." ON)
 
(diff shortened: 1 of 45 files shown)


lgritz:
I'm definitely liking the direction this is going. The ability to point to an arbitrary place to root the image tests, is a good idea. I need to fool around with it a bit to see what it's like in practice.

lgritz:
LGTM, merging!

PR #2027, opened 2018-10-07
Release an official FindOpenImageIO.cmake

lgritz:
* Make an official FindOpenImageIO.cmake, borrowed from OSL but with some
  minor changes. It will install into INSTALL_PREFIX/share/cmake/Modules

* We hope that downstream projects that have OIIO as a dependency will
  use/copy this file, rather than having a dozen or more slightly
  different versions that downstream projects all write themselves. And
  if there are any deficiencies, they will push fixes or improvements to
  this master copy.

* Change places where we expect the cmake and/or environment variable
  OPENIMAGEIOHOME to be named OPENIMAGEIO_ROOT_DIR (more standard naming
  convention by modern cmake standards).

* Simplify FindOpenEXR.cmake, to conform to conventions as well as to
  get rid of some flexibility that is nonstandard and that we didn't use
  anyway.

* Some other very minor fixes to the cmake files (INSTALL_DOCS was
  declared in two different places, etc.)


lgritz:
Rewrite of one that I submitted a few days back, then withdrew for more work.

lgritz:
No objections --> merging

We can revise if people point out ways that this can be improved.

krop:
Bad move, that's like hiding the map to get the treasure chest inside the chest. 
The correct way is to provide an OpenImageIOConfig.cmake file with the actual path for the library and headers.

lgritz:
Care to elaborate for those of us who are obviously less familiar with modern cmake idioms? Maybe a link to an article explaining the best practices here?

krop:
Sure,
I have a project Foo that installs libfoo.so and libfoo.h. My user may install it wherever they want it.
My project Bar has a dependency on Foo and uses CMake, I can write a FindFoo.cmake file to find where the different things are installed. This is a legit case because I don't know where they are.

The alternative:
When I'm working on the Foo project, I *know* where the library and headers will be installed, I also know the project version, so I can provide a FooConfig.cmake file and a FooConfigVersion.cmake file to actually tell where these files are. 
This way, I don't need a FindFoo.cmake anymore and may use the accurate information immediately.
You can see it as a pkgconfig-like file for CMake.

CMake provides a module to write such config files: https://cmake.org/cmake/help/v3.2/module/CMakePackageConfigHelpers.html

lgritz:
I was thinking primarily about the first case, which seems to be the common approach, based on packages I've seen. The prior state-of-the-art for OIIO was that all of the Bar, Baz, and Woo projects each wrote their own FindOpenImageIO.cmake implementation from scratch, they were all different, and none seemed quite up to my standard. So I figured it would at least be a step forward for OIIO to provide a reference implementation of FindOpenImageIO.cmake that I was pretty sure would work under a wide variety of circumstances (*), which other packages could just copy and use as needed.

(*) I'm relatively confident in this implementation because it's just a minor cleanup of the FindOpenImageIO.cmake that we've long used in OSL to find and use OIIO. But we'd never pushed it back to the OIIO project so that people unfamiliar with OSL would be aware of it.

I think I understand the gist of the package config helper files now, thanks for the pointer. But I'm still a little hazy on how it's used by the calling package. Are the generated config files automatically used when the downstream package does a `find_package(OpenImageIO)`? If the user builds OIIO from scratch and they move the files from the "locally installed" location to a system location, how will it know where to look? Or what if they want to install in a nonstandard location (like, maybe they don't even have permissions to install it in a typical system-wide directory), how will cmake in the downstream project they are building know where to look for the config file itself? Do you know an example of another project that you think uses this method particularly well that I can use as a model? (Bonus points if the model is a widely used VFX package so I know oiio users are likely already familiar with the ideas.)

krop:
The calling package just has to find_package(OpenImageIO) (or find_package(OpenImageIO CONFIG) to be sure). CMake then looks for OpenImageIOConfig.cmake in all directories specified in the CMAKE_PREFIX_PATH environment variable and also in common system locations.

We use cmake config files everywhere in the KDE libraries and a simpler example can be found e.g in libical: https://github.com/libical/libical:
The relevant part in the main CMakeLists.txt: https://github.com/libical/libical/blob/master/CMakeLists.txt#L596:L620
The export when installing the library: https://github.com/libical/libical/blob/master/src/libical/CMakeLists.txt#L294:L298
and the LibIcalConfig.cmake.in we're using: https://github.com/libical/libical/blob/master/LibIcalConfig.cmake.in

>  If the user builds OIIO from scratch and they move the files from the "locally installed" location to a system location

This should be discouraged since it can lead to runtime issues if the rpath is set. The best way to change the installation directory is to re-run CMake with another CMAKE_INSTALL_PREFIX value.

lgritz:
Thanks for all the pointers! I'll definitely read up on this and see what I can throw together.

PR #2032, opened 2018-10-12
Fix for TIFF with unassociated alpha

lgritz:
If read_scanlines (et al) was called with the 'format' parameter as
TypeUnknown -- which is supposed to mean "use the native format" -- would
accidentally and silently fall through the switch statement inside
OIIO::premult(), causing it to do nothing.

If called with an explicit type, it all worked find. Subtle to track down.

The particular symptom was that if you were calling ImageBuf::read()
with `force=true` but relying on the default `convert` value of UNKNOWN
(meaning, no conversion, stay in native format), it would trigger this
bug and neglect to premultiply the unassociated colors.


aconty:
LGTM

PR #2033, opened 2018-10-18
Strutil enhancements

lgritz:
Found this while doing fall cleaning. These are some mods I've had sitting around in a private branch for a long time.

1. Turn join() into a template that can act on any iterable container
   of objects that allow stream output.

2. New splits()/splitsv() that directly return a vector of string or
   string_view, respectively (more modern C++ idiom than passing a
   result vector as an argument).

3. A second version of extract_from_list_string that directly returns
   a std::vector (instead of being passed as an argument).

4. Minor fix in unittest.h to allow the OIIO_CHECK_EQUAL macro work to
   compare std::vectors. (This utilizes the new join() to turn them into
   strings and thus allow them to output with <<).

### DIFF
--- a/src/include/OpenImageIO/strutil.h
+++ b/src/include/OpenImageIO/strutil.h
@@ -234,6 +234,8 @@ void OIIO_API to_lower (std::string &a);
 /// a static locale that doesn't require a mutex lock.
 void OIIO_API to_upper (std::string &a);
 
+
+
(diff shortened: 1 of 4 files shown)


hobbes1069:
I'm not in a position to do any testing but it built fine on Fedora 28.

fpsunflower:
LGTM

PR #2035, opened 2018-10-20
Add ImageBufAlgo::fit_range.

marsupial:
## Description
Adds the ability to normalize or fit pixels to a set range and get prior min/max values.

## Tests
Added --fitrange to oiiotool and test in testsuite/oiitool

## Checklist:


lgritz:
I hate to do this to you, but I have an alternate implementation. Shortly before you submitted this review, somebody had contacted me off-list and we had a conversation about contrast remapping that caused me to have already written most of an IBA/oiiotool implementation that largely overlaps with this.

Mine I think is a more flexible remapping -- both domain and range values to define the mapping, any of which can be single floats or per-channel, as well as an optional sigmoidal curve in addition to having a linear ramp. But yours has a nice property of being able to specify the parameters as a percentage, which I kind of like especially from the oiiotool perspective.

I'm going to submit a separate review for mine. I think what I want to do is to go with that version, and then return to it and recreate what you did with the percentages.

lgritz:
I believe that this has been superseded by #2043. Closing.

PR #2038, opened 2018-10-24
Safer thread_pool destruction using native threads status (Windows)

### LINKED ISSUE #1795: OpenImageIO.dll can freeze at app exit
### Problem

OpenImageIO.dll freezes on some Windows 7 machines at application exit. 
I've debug this issue and it freezes in thread.cpp at line 205:
`this->cv.notify_all();  // stop all waiting threads`

- It is not freezing when the threads attribute is set to 1
- There is no problem running the same code on Windows 10
- There is also no problem using a static version of the OIIO library

The problem is similar to:
https://social.msdn.microsoft.com/Forums/en-US/7f91e360-2108-40ca-8728-e295f17cb26b/visual-c-2015-stdconditionvariablenotifyall-may-get-stuck?forum=vcgeneral

It can be related to: #1572 

**Expected behavior:** 
Providing public Initialise/DeInitialise functions could be a stable solution.

### Versions

* OIIO branch/version: 1.8.5 (last stable) - DLL library
* OS: Windows 7 x64
* C++ compiler: Visual Studio 2017 x64

yann-lty:
## Description
Prevent deadlock from illegal communication with already terminated threads when global shared thread_pool is being destroyed during DLL unloading. This scenario is detected by checking native threads status.
Fix #1795 

## Tests
Tested on Windows 7 with Visual Studio 2017. 
Fixes the deadlock issue for global static thread_pool destruction and does not modify the current behavior when thread_pool is used as a non-static variable.

## Checklist:

### DIFF
--- a/src/libutil/thread.cpp
+++ b/src/libutil/thread.cpp
@@ -44,6 +44,10 @@
 #define _ENABLE_ATOMIC_ALIGNMENT_FIX /* Avoid MSVS error, ugh */
 #endif
 
+#if defined(WIN32)
+#include <windows.h>
+#endif
+
 #include <exception>
 #include <functional>
 #include <future>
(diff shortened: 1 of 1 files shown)


lgritz:
LGTM, but I don't know anything about Windows so should not judge.

Does anybody who is a Windows expert else care to have an opinion about this?

lgritz:
Fixes #1795

lgritz:
In #1795, it was reported that this patch fixes things. Good enough for me, merging.

hyperfx:
Excellent news, looking forward to a new build!

PR #2039, opened 2018-10-25
Default to use vanilla boost

marsupial:
Default Makefiles build to use vanilla boost.


lgritz:
Note for readers: this is an SPI only thing.

lgritz:
LGTM.

PR #2042, opened 2018-10-25
null input reader: be more careful with "full" (display) size

lgritz:
For the "null" reader, we were clobbering any configuration hints about
the "full" (i.e. "display window") size parameters. This had bad effects
when using a fake "null" image along with ImageCache::add_file() in
order to set up an in-memory buffer that behaves like a texture, if that
buffer was a crop region or had overscan.


fpsunflower:
LGTM

PR #2043, opened 2018-10-25
IBA::contrast_remap() and oiiotool --contrast

lgritz:
Allows flexible linear or sigmoidal contrast remapping.


lgritz:
@marsupial Any opinions?

marsupial:
Yet to try it out, but liking the curves/ramp and per-channels.
Still into the idea of getting max/min and fitting a range in one call, but maybe the use-case is too specialized, easily added, or I'm reading the docs wrong?

lgritz:
Does it have to be in one call (C++/Python)? Or can it just be super easy on one oiiotool command line? Because the latter is straightforward, I just wanted to do that as a separate PR.

marsupial:
Agree its a separate topic and can be dealt with after.
Think/prefer it would be programmatic though rather than `oiiotool`.
I can just rebase & tinker with #2035, and if you think it's too specific, just use it from the client

PR #2044, opened 2018-10-26
Add a bumpformat option for the bumpslopes exporter

lecocqp:
This option prevents situations where the bumpslopes exporter fails to detect an input bump format as an height map. The typical situation is a 3 channels RGB height map we would expect to be monochromatic (R=G=B for all pixels) but having subtle color variations introduced by a texture publication process (typically a precision or rounding error). In such situation the exporter will automatically interpret the bump map as a false normal map, resulting in a corrupted texture output.

The bumpformat option allows the user to bypass the automatic detection and manually specify the input bump format.

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -480,7 +480,7 @@ normal_gradient (const ImageBuf &src, const ImageBuf::Iterator<float> &dstpix,
 
 template<class SRCTYPE>
 static bool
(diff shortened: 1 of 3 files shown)


lgritz:
The approach LGTM, though it's failing the CI tests on Travis. Do some tests need to be updated?

lecocqp:
An initialization was missing for one test case. Should be better now!

lgritz:
CI passes. LGTM. Merging.

PR #2047, opened 2018-10-29
Remove ustring::operator int()

lgritz:
As pointed out by Thiago Ize and John Haddon (but unfortunately not
heard by me until literally years later), this is error prone because
there are times when a ustring is converted to an int/bool in unexpected
situations. It even led to verifiable errors in OSL that I only
discovered after commenting out this operator to see what would happen.

Neither std::string nor std::string_view have this operator, bolstering
Thiago & John's arguments.

So we are removing it from OIIO 2.0. Hopefully this won't break too many
people's apps that happen to make use of ustring. The fix is simple, just
change:

    if (myustring)

to

    if (! myustring.empty())

Because this is a back-compatibility break, this change will only be merged into master, NOT backported to any current release branches.


fpsunflower:
LGTM!

PR #2048, opened 2018-11-04
Fixes related to debugging half -> float conversion

lgritz:
We found that certain 'half' pixel values that involved "denormalized"
values (very close to zero) were inexplicably getting squashed to exact
0.0f in some circumstances but not others. Long story short, we traced
this to an older version of libopenjpeg (in turn pulled in by ffmpeg's
libraries) that set the x86 DENORMALS_ZERO_MODE upon its DSO load, and
left it like that, which is a truly terrible thing to do because it can
inject math bugs elsewhere into the app, even in parts that seemingly
have nothing to do with decoding jpeg-2000 images.

The code is only symptomatic in OIIO when we build with SSE support but
not with f16c instruction support -- the symptom is specific to the
series of ops we use to accelerate the half->float conversion with SSE
but without f16c. There could be other effects, but we haven't seen
them.

This seems to be fixed in newer versions of libopenjpeg. But we
encountered it, so others may also. It seems equally evil for our
library to reset the flag and leave it that way. However, for a full
application like oiiotool, we can safely set the flag, and so we do (if
the intrinsic is available, which unfortunately it is not in gcc 4.8).
I think that's the best we can do for now, out of several options that
were less than satisfactory. (A different set of instructions for h->f
conversion also avoided the problem, but was much slower.) We also advise
people to use a sufficiently new version of libopenjpeg to avoid this.

Along the way...

* simd.h methods to get/set the flush_zero and denorms_zero modes.
  (Note: known gcc 4.8 bug omits access to those intrinsics.)

* Preemptively turn off the denorms zero mode in oiiotool, in case some
  rogue library has turned it on.

* `oiiotool --stats` : Don't force a conversion to float when reading
  the full image to compute stats on. This change also avoids the bug,
  though in a brittle way. Still probably good to compute the stats on
  the native data.

* fmath_test checks EVERY 'half' value and makes sure the h->f->h round
  trip works properly.

* Change the behavior of `oiiotool --dumpdata` do try to dump the values
  of each data type without conversion, rather than converting all to
  float and printing the float values. This is a more true representation
  of the raw data in the file, which I assume is why somebody wants to
  do the --dumpdata.

* fmath.h changes to speed up bulk h[]<->f[] conversions by using
  8-wide simd ops when f16c is available. (This is just incidental,
  something I saw along the way.)

* `oiiotool --help` prints the number of cores, main mem size, and set
  of processor capabilities (SIMD level, etc.) on the hardware it's
  currently running on, and which set of SIMD capabilities was requested
  at build time.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -744,10 +744,19 @@ inline void convert_type<half,float> (const half *src,
                                       float *dst, size_t n,
                                       float _min, float _max)
 {
+#if OIIO_SIMD >= 8 && OIIO_F16C_ENABLED
+    // If f16c ops are enabled, it's worth doing this by 8's
+    for ( ; n >= 8; n -= 8, src += 8, dst += 8) {
+        simd::vfloat8 s_simd (src);
+        s_simd.store (dst);
+    }
+#endif
+#if OIIO_SIMD >= 4
     for ( ; n >= 4; n -= 4, src += 4, dst += 4) {
         simd::vfloat4 s_simd (src);
         s_simd.store (dst);
     }
+#endif
     while (n--)
         *dst++ = (*src++);
 }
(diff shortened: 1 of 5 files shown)


fpsunflower:
LGTM

PR #2052, opened 2018-11-05
Texture: Fix edge case bug in adjust_blur

lgritz:
Mea culpa. Ramon Montoya alerted me to a problem in 2013 and I either forgot about it, ignored it, or failed to understand it. But I rediscovered his email while trying to purge my inbox of long-ago unresponded emails and decided to look into it again.

The problem situation is that we figure out the filter ellipse based on the derivatives, and then "thicken" it based on the amount of blur requested.  Ramon pointed out that a very uneven amount of blur (very different in s and t) could thicken the minor axis so much more than the major axis that, you know, they aren't the proper axes any more. And that could lead to odd behavior later on, because the samples are laid out along the major axis and assuming in all the math that `majorlegnth >= minorlength`. 

I tried a few approaches, and the most streightforward was just to notice when this happens, and swap the minor/major (rotating the ellipse by PI/2).  This seems to basically work and looks quite good for the common situations, although it's less than satisfactory for the combination of very uneven blurs and extreme sheer that you get when the ds and dt derivatives are nearly colinear (pointed in almost the same direction or almost opposite, instead of the more typical and expected nearly-perpendicular configuration). Though maybe it's right then, too, I dunno, it's very hard to reason about what's "right" in that weird case.

That fix is 3 lines of code in adjust_blur().

The rest of the PR is comments trying to explain what works and doesn't, and various improvements to the unit test visualizations (including adding the ability to turn them on and control the amount of blur from the env OPENIMAGEIO_TEXTURE_OPTIONS, drawing the ellipse axes, and drawing the individual sample footprints in the visualization).

All my testsuite passes. Lookups should all be identical except for that problem case -- certain situations of VERY uneven s and t blur amounts -- and in that case it's possible that you might have renders change the appearance slightly.

### DIFF
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -75,6 +75,8 @@ namespace {  // anonymous
 // contains a few hundred bytes).
 static TextureSystemImpl *shared_texturesys = NULL;
 static spin_mutex shared_texturesys_mutex;
+static bool do_unit_test_texture = false;
+static float unit_test_texture_blur = 0.0f;
 
 static EightBitConverter<float> uchar2float;
 static vfloat4 u8scale (1.0f/255.0f);
@@ -355,9 +357,8 @@ TextureSystemImpl::init ()
     if (options)
         attribute ("options", options);
 
-#if 0
-    unit_test_texture ();
-#endif
+    if (do_unit_test_texture)
+        unit_test_texture ();
 }
 
 
@@ -480,6 +481,14 @@ TextureSystemImpl::attribute (string_view name, TypeDesc type,
     if (name == "statistics:level" && type == TypeInt) {
         m_statslevel = *(const int *)val;
         // DO NOT RETURN! pass the same message to the image cache
+    }
+    if (name == "unit_test" && type == TypeInt) {
+        do_unit_test_texture = *(const int *)val;
+        return true;
+    }
+    if (name == "unit_test_blur" && type == TypeFloat) {
+        unit_test_texture_blur = *(const float *)val;
+        return true;
     }
 
     // Maybe it's meant for the cache?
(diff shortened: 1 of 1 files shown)


src/libtexture/texturesys.cpp
```diff
@@ -1239,17 +1248,30 @@ TextureSystemImpl::texture_lookup_nomip (TextureFile &texturefile,
...
     dsdy *= swidth;
     dtdy *= twidth;
+#if 0
+    // You might think that blur should be added to the original derivs and
+    // then just carry on. And sometimes it looks fine, but for some
+    // situations the results are absurdly wrong, as revealed by the unit
+    // test visualizations. I'm leaving this code here but disabled as a
+    // reminder to myself not to try it again. It's wrong.
+    if (sblur+tblur != 0.0f /* avoid the work when blur is zero */) {
+        dsdx += std::copysign (sblur, dsdx);
```
fpsunflower: I think the right math involves thinking about the two filter kernels as gaussians and looking for the right convolution of the two.

I think there is a closed form for this.

lgritz: I'm sure you're right, but I don't have the bandwidth to chase it down at the moment. Considering that nobody (not even the person who originally reported the problem) has mentioned it in the last 5 years and we've never noticed, I'm very tempted to just do this fix (which at least is a small step over the status quo) for now, and only go down the long rathole of trying to perfectly solve it if somebody either really seems to care or wants to work out the math themselves.

lgritz: So what say you? Is this ok as it stands?

fpsunflower: Yep LGTM, maybe put in a TODO comment about the Gaussian convolution view of the problem for whenever we do look into it again?

lgritz: sure

PR #2053, opened 2018-11-06
Add a .clang-format file and a 'make clang-format' target

lgritz:
clang-format is a part of the LLVM/Clang compiler suite that will
reformat your code -- but much better than most code reformatters,
because this one is based on a compiler toolkit and is working on the
fully understood AST, not just text.

The idea here is that we give up a little bit of individual creativity
in code formatting, in exchange for every pull request trivially being
able to conform to the house style and never having to go through
multiple rounds of PR edits for formatting issues, nor ever have any
squabbles about it. I know that nobody will be 100% happy with the
require formatting (even me), but I think if we can make everybody 80%
happy and avoid those other problems forever after, it's worth it.
People I respect running othe projects consistently swear that adopting
clang-format was one of the best decisions they made.

Basically, after the initial cmake configuration, you can reformat
files with

    make clang-format

Configuration variables CLANG_FORMAT_INCLUDES and CLANG_FORMAT_EXCLUDES
let you customize the lists of glob patterns that get included, or
excluded, from reformatting. For example, if you only want to reformat
one file (and you should try this to see how the whole thing works!):

    make nuke
    make CLANG_FORMAT_INCLUDES="exrinput.cpp" clang-format

Currently, I'm thinking that public header files will NOT be reformatted
(i.e. the exclude list will default to including src/include).  I just
can't seem to get fully satisfied with the transformations, and I think
that's because I think of public headers as documentation, so I'm
extremely picky about the formatting and think that readability is
paramount, so I make little one-off rule breaks all the time to improve
the flow, readability, and appearance. Of public headers. I think for
everything else, I'm willing to live with a set of rigid rules.

I'm still experimenting with getting the parameters right. I may keep
tinkering a bit before committing any actual reformatting. For that reason,
I'm merely proposing the build script and style configuration portions
now. Only after people have tried it out and we think we can live with
it, will I do the great reformatting checkin.

Also, eventually, we'll make one of the Travis-CI tests do a
clang-format and fail if anything needs to be changed (and print the
diffs), thus making the CI make it very clear when any PR fails to
conform, and we will expect the submitter to run clang-format (or change
the offending parts by hand) and resubmit.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,119 @@
+Language:        Cpp
+BasedOnStyle:  WebKit
+SpaceBeforeParens: ControlStatements
+
+#AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+#AlignConsecutiveAssignments: false
+#AlignConsecutiveDeclarations: false
+#AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: true
+#AllowShortFunctionsOnASingleLine: All
+#AllowShortIfStatementsOnASingleLine: false
+#AllowShortLoopsOnASingleLine: false
+#AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: TopLevel
+#AlwaysBreakBeforeMultilineStrings: false
+#AlwaysBreakTemplateDeclarations: MultiLine
+#BinPackArguments: true
+#BinPackParameters: true
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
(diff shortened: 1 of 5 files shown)


kwk:
@lgritz I've used clang format ad McAfee and Intel Security and it was wonderful. People were initially very skeptic about not being able to write code as they were used. Clang format works in every editor which is a nice plus as well.

Nowadays I mostly work with Go and there the discussion about formatting code had almost completely vanished, which let's you focus on the code instead of the formatting and style. It also is great for working in a large team.

scott-wilson:
I haven't really used clang formatter much, but I'm using formatters on my Go and Rust projects, and I love the fact that I don't need to worry about formatting. I just have my editor format on save, and formatting is done for me.

lgritz:
Yes, I was thinking the same thing last night!

I still like the include/exclude list at least while I'm still testing it out, because it makes it really easy to do one directory at a time to sort through the changes and make sure I'm happy with them. But might excise that after we've finished the whole conversion.

lgritz:
I think what I want to do is check in just the make/cmake/.clang-format changes in this review first (now-ish).

Then, when it's time to actually check in the great format change of 2018, the `// clang-format off` will be put in the include files that need it. (I'll probably get to that over the weekend or next week, and give people several days to look it over and object to the formatting choices before we merge.)

Then when it's all over but the shouting, I'll check in the CI test that enforces the new formatting.

Sound reasonable?

Issue #2054, opened 2018-11-08
oiiotool --info does not work for certain windows paths


ThiagoIze:
On Windows we get the following output error (with an audible beep for good measure courtesy of '\a')
```
oiiotool --info somedir\another\foo.tx
somedirnother♀oo.tx : 8192 x 8192, 3 channel, uint8 tiff (+mipmap)
```
I fixed it with
```diff
diff --git a/src/oiiotool/printinfo.cpp b/src/oiiotool/printinfo.cpp
index 0a79bfdf..5dcd6216 100644
--- a/src/oiiotool/printinfo.cpp
+++ b/src/oiiotool/printinfo.cpp
@@ -633,7 +633,7 @@ print_info_subimage (Oiiotool &ot,
     }

     // Unescape the strings if we're printing for human consumption
-    for (size_t i = 0; i < lines.size(); ++i) {
+    for (size_t i = 1; i < lines.size(); ++i) {
         lines[i] = Strutil::unescape_chars (lines[i]);
     }
```
but I'm not sure if this is a good fix, so I'm not making a PR. For instance, maybe lines[0] might not always be a filename or maybe there are other lines that also shouldn't be unescaped?

PR #2055, opened 2018-11-09
fix oiiotool --info windows_path

ThiagoIze:
As mentioned in #2054, windows paths have backslashes which are getting removed by `oiiotool --info`.  This fix solves the filename issue and I verified still works with the maketx "Software" metadata which has backslashed filenames but does not the unescaping to be skipped.

I have not run the testsuite nor made a test.

## Checklist:


lgritz:
LGTM. I will merge and backport.

PR #2057, opened 2018-11-09
Fix FindOpenEXR.cmake to use version detected by pkg-config, if available

devurandom:
Detecting the OpenEXR version by parsing a C header can be unreliable.  Since
 we already use pkg-config to detect the presence of the library, we can also
 use the version number it gives us, if available, and only resort to parsing
 the header file if that failed.

See-Also: https://bugs.gentoo.org/668412
Fixes: ec4f679cd7ba175a5470a6198c07ebc6780f4c29

## Description

See commit message.  I suggest to backport this to branch RB-1.8.

## Tests

I applied this patch to release 1.8.16 and successfully compiled a Gentoo package with it.

## Checklist:

  - N/A
  - N/A
  - N/A


lgritz:
Looking good this time!

lgritz:
Backported to 1.8.

PR #2059, opened 2018-11-10
Reformatting via clang-format (part 1)

lgritz:
This first pass only reformats the contents of src/include/*. We will
tackle other directories in later stages. I wanted to start with just
one directory first, to give people a chance to object to the style
it imposes or tweak the format rules before doing the rest of the
code base.

For a minority of headers, I added a `// clang-format off` switch
because I was just too particular about the formatting of those files,
since I consider those public headers to be "documentation" as much as
working code, and therefore am extremely picky about certain ways I
strive to have them readable even when it involves breaking the usual
rules. Note that in some of these cases, even though I added the switch
to turn off future automated reformatting, I nevertheless made extensive
formatting changes to conform to the new prevailing style where I felt
it did not negatively affect readability.

I also turned clang-format off for a couple of files "imported" from
other projects (such as pugixml) in order to preserve our ability to
diff against changes in the original code. In those cases, I made no
other changes to those files except to add that one line to turn off
future reformatting.

No "code" has changed, except for a few very minor bits of clean-up I
performed at the same time:

* Removed a whole lot of `#ifdef OPENIMAGEIO_FOO_H` include gaurds,
  changing them to use `#pragma once` instead. In a few cases, I did
  leave a `#define` so that certain other source can tell if a file had
  been included. (I have reasons.)

* Change NULL -> nullptr in ustring.h


fpsunflower:
I like where this is going!

LGTM!

PR #2069, opened 2018-11-16
clang-format, final-ish update

lgritz:
This fully uses clang-format on the entire codebase (except for a few
files, and a few sections of files, where we explicitly disable it for
reasons.

Henceforth, we will assume that any new code contributions should be
conformant and reject pull requests that are not. The basic idea is:

1. For large code changes, make sure you have a fairly recent version
   of clang-format installed on your end, and use

        make clang-format

   to format the code before sending off the PR.

2. The TravisCI tests include one entry that merely does a 'make
   clang-format' and considers it an error if there are any resulting
   diffs (i.e. if clang-format needed to reformat anything). In such a
   case, it will also print the diffs themselves if you click on the
   failed test.

Note that for small patches and users who aren't easily able to have a
local clang-format copy, you can essentially skip step 1, submit your PR
as-is and then examine the failure diffs of step 2, making the requested
changes by hand and then updating your PR.

It is also worth noting that as a result of this massive reformat, it is no longer reliable to use 'git blame'  to track down authors on a line-by-line basis (across this boundary), since so very many lines were altered by the reformatting and thus look like they were edited by me, this week, when in fact they are often just reformatted lines written by other people long ago. 

Similarly, certain ways of accounting for how many lines of the code base were authored by each contributor are no longer accurate, since I now have a disproportionate amount attributed to me as a result of the reformatting. So it goes. This is a one-time thing, and the transition to "2.0" seems like the right time to do it.

### DIFF
--- a/src/igrep/igrep.cpp
+++ b/src/igrep/igrep.cpp
@@ -29,41 +29,41 @@
 */
 
 
+#include <cmath>
 #include <cstdio>
 #include <cstdlib>
-#include <cmath>
 #include <ctime>
 #include <iostream>
 #include <iterator>
 #include <memory>
 
 #include <OpenImageIO/argparse.h>
-#include <OpenImageIO/strutil.h>
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/imageio.h>
+#include <OpenImageIO/strutil.h>
 
 #ifdef USE_BOOST_REGEX
-# include <boost/regex.hpp>
-  using boost::regex;
-  using boost::regex_search;
-  using namespace boost::regex_constants;
+#    include <boost/regex.hpp>
+using boost::regex;
+using boost::regex_search;
+using namespace boost::regex_constants;
 #else
-# include <regex>
-  using std::regex;
-  using std::regex_search;
-  using namespace std::regex_constants;
+#    include <regex>
+using std::regex;
+using std::regex_search;
+using namespace std::regex_constants;
 #endif
 
 using namespace OIIO;
 
-static bool help = false;
-static bool invert_match = false;
-static bool ignore_case = false;
-static bool list_files = false;
-static bool recursive = false;
-static bool file_match = false;
-static bool print_dirs = false;
-static bool all_subimages = false;
+static bool help           = false;
+static bool invert_match   = false;
+static bool ignore_case    = false;
+static bool list_files     = false;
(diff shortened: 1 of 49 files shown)


src/igrep/igrep.cpp
```diff
@@ -178,33 +179,34 @@ main (int argc, const char *argv[])
...
-    boost::regex_constants::syntax_option_type flag = boost::regex_constants::grep;
+    boost::regex_constants::syntax_option_type flag
+        = boost::regex_constants::grep;
     if (extended_regex)
         flag = boost::regex::extended;
     if (ignore_case)
         flag |= boost::regex_constants::icase;
 #else
-    std::regex_constants::syntax_option_type flag = std::regex_constants::grep;
+    auto flag = std::regex_constants::grep;
```
fpsunflower: clang-format is also able to translate types to `auto`? Or did you run `clang-tidy` as well?

lgritz: No, that was me. :-)

As I reviewed the diffs to see what it rewrote and why, in a *very* small number of cases of long lines or awkward constructs, where the reformatted code looked especially ugly, I occasionally made a change or two of this nature to coax it into a form that didn't need to re-wrap in a weird way.

fpsunflower: Makes sense. LGTM!

lgritz:
I'll also note that very occasionally, I will have a "perfect" snapshot where running clang-format on my machine yields no diffs, but inexplicably yields a single minor difference when clang-format runs on Travis. I'm not sure why that is, perhaps slightly different clang-format differences between the two machines? (I run the travis test on a Mac instance, as is my laptop, both getting clang-format from homebrew so I don't know why they would be different.) I'm not quite sure how to deal with that yet when it crops up.

src/igrep/igrep.cpp
lgritz: I'm pretty happy about the whole thing overall. The deltas were really extensive, but 90% was totally neutral, stuff like "foo (bar)" -> "foo(bar)" and the like.  Some changes are different than I would have written them and I think they can be improved, but they are within the bounds of what I would accept from a team member contributing to the project, and I think worth accepting my occasional peeves with its choices for the peace of mind of having all future contributors able to trivially 100% conform to a single formatting standard and allow it to be checked automatically.

PR #2070, opened 2018-11-16
zfile fix and test

### LINKED ISSUE #1817: Assert at app exit when opening corrupted image
### Problem

This simple program shows assert at exit:

`#include <OpenImageIO/imageio.h>`
`int main()`
`{`
`  OIIO::ImageInput *in = OIIO::ImageInput::open("obrtest.png");`
`  //_fcloseall();`
`  return 0;`
`}`

obrtest.png:

obrtest

The problem is probably related to Windows - Visual Studio - zlib. 
If the file is invalid, then all plugins are tried (including zfile plugin).
It seems that there can be a problem in zfile.cpp with closing file handles in zlib and msvc.

The included program shows me in debug mode an assert at app exit (or calling _fcloseall manually):
Expression: (_osfile(fh) & FOPEN)

And it also crashes in release mode on some machines.

When I exclude the zfile plugin from the OIIO lib, it works - but there could be a better solution...

### Versions

* OIIO branch/version: 1.8.5 (static lib), zlib (static) - all compiled with /MD (/MDd in debug)
* OS: Windows 10 x64
* C++ compiler: Visual Studio 2017 x64

lgritz:
Spurious problems on Windows seem to happen with gzopen(filename) as we
used in ZfileInput::valid_file, but not with the gzdopen(fileno(fd)) as
we used in ZfileInput::open(). So change valid_file to the method that
seems to work. This has been linked to crashes during app shutdown on
Windows (even when no zfiles are encountered, since it can happen with
invalid files when all input readers are tried).

Also add a zfile testsuite entry.

Fixes #1817.


lgritz:
@riviera-kid

riviera-kid:
I'm still getting the assert. I think that the problem is not in valid_file, but in the ZfileInput::open. BTW the ZfileInput::valid_file is not called when opening the corrupted image - ZfileInput::open is called and return false - error ("Not a valid Zfile")
I think that the code with native zlib file system functions is working (valid_file before the fix)...

lgritz:
Did I do it backwards?

Oh, I see, you're saying that adding an explicit fclose() seems to do the trick?
Let me take another stab at it.

riviera-kid:
There is a problem with mixing stdio with zlib. I think there is no reason to use stdio fopen/fclose in a zfile reader - using only gz*func for IO could solved the issue.

lgritz:
The reason we go through fopen is because of utf filenames. But I see now that there's a gzopen_w that takes wide char filenames.

Update has been pushed. Can you give a try again to the latest version in my PR?

riviera-kid:
The latest version is working well! Thanks

lgritz:
Great to hear! I will merge it later today.

lgritz:
Merged, and also backporting to 1.8.

PR #2079, opened 2018-11-26
Formally bump MSVS minimum to 2015

lgritz:
We think that as a practical matter, this has been the case for a long
time. But the documentation didn't keep up.


lgritz:
The clincher is: MSVS2015 seems to be the minimum for pybind11, so master has been VS2015 minimum for many many months and nobody has complained. Merging.

PR #2081, opened 2018-11-30
m_fingerprint key is the hash, not the filename.

ThiagoIze:
The m_fingerprint key is supposed to be the hash, but we were searching by filename. End result is that no fingerprints were ever removed from the m_fingerprint container. This results in crashes later on if OIIO thinks a duplicate file already exists and so rather than read it in, it tries to use the now non-existent data.


lgritz:
Oh dear, you are right. Thanks for the catch, LGTM.

PR #2082, opened 2018-11-30
vsnprintf only returns -1 on encoding error

ThiagoIze:
I know there's a comment about vsnprintf not working properly on Windows and old linux, but that's no longer the case.  VS2015, along with linux and OSX all return -1 when there's an encoding error. Otherwise they all return the full size.

See: https://msdn.microsoft.com/en-us/library/1kt27hek.aspx and http://www.cplusplus.com/reference/cstdio/vsnprintf/.

## Description
This means that on encoding error we'll always get -1 and OIIO will keep doubling the buffer size until we run out of memory and crash.

## Checklist:

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -175,12 +175,12 @@ Strutil::vsprintf(const char* fmt, va_list ap)
         int needed = vsnprintf(buf, size, fmt, ap);
         va_end(ap);
 
-        // NB. C99 (which modern Linux and OS X follow) says vsnprintf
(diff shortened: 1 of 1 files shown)


ThiagoIze:
Incidentally, I was getting an encoding error on Windows I think because the va_list contains non-ascii wchar (coming from TIFFOpenW) but this function was expecting chars. I just cared about stopping the crashes, so I haven't debugged that part.

lgritz:
LGTM, I will merge as soon as the CI tests finish successfully.

PR #2083, opened 2018-11-30
Python: fix a couple string-returning functions

lgritz:


lgritz:
Amended: found another spot with a similar error.

lgritz:
Amended with another couple spots of the same (very minor) error.
It's not of huge consequence... the worst thing is that a function returning a string to Python, if it's Python 2.7, will return a unicode string. (For Python 3.x, the main string type is unicode, so that's not even a little bit surprising to py3 users, but may be unexpected for py27.)

PR #2086, opened 2018-12-01
Deprecate missing_math.h

lgritz:
Once we were fully C++11 and moved MSVS minimum to >= 2015, the contents of this file were mostly no longer needed. What few things were here that anybody might still rely on, are in fmath.h.


src/include/OpenImageIO/missing_math.h
```diff
@@ -29,132 +29,12 @@
...
-log2f(float val)
-{
-    return logf(val) / static_cast<float>(M_LN2);
-}
-
-#endif
-
-
-OIIO_NAMESPACE_END
+#include <OpenImageIO/fmath.h>
```
fpsunflower: Adding this line broke then OSL build because `simd.h` needs `missing_math.h` which needs `fmath.h` which needs `simd.h` (which hasn't yet been included).

Is it really necessary to bring in `fmath.h` here? or should it be `cmath` instead?

lgritz: I think it helps, because some of those #defines moved to fmath.h. 

The culprit here was that I hadn't removed stray include of missing_math.h from simd.h.

I just pushed that fix, all should be ok onow.

PR #2091, opened 2018-12-06
Fixed a few compiler warnings issued by gcc 7.3.0-27ubuntu1~18.04

norru:
## Description

- fixes build failures of https://github.com/OpenImageIO/oiio.git on Ubuntu 18.04

## Tests

N/A

## Checklist:

- [N/A] If this is more extensive than a small change to existing code, I
- [N/A] I have updated the documentation, if applicable.
- [N/A] I have ensured that the change is tested somewhere in the testsuite
- [N/A] My code follows the prevailing code style of this project.

### DIFF
--- a/src/include/imagebufalgo.h
+++ b/src/include/imagebufalgo.h
@@ -46,7 +46,7 @@
 
 #include <limits>
 
-#ifndef __OPENCV_CORE_TYPES_H__
+#if !defined(__OPENCV_CORE_TYPES_H__) && !defined(OPENCV_CORE_TYPES_H)
 struct IplImage;  // Forward declaration; used by Intel Image lib & OpenCV
 #endif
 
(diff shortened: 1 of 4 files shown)


lgritz:
LGTM, thanks for the fixes.

lgritz:
Whoa. I only realized after I hit the merge button that this is a PR against RB 1.2!

It may be in the tree, but I won't be releasing a tagged release patch to 1.2.  The current supported release is 1.8, which is mere days away from becoming obsolete as we release 2.0. 

I'll make sure these changes are propagated to master (though I suspect they have all long since been fixed in modern versions, which explains why I was surprised that they were still issues).

lgritz:
In fact, I'm going to revert the fix. I don't want any activity on 1.2.

lgritz:
If any of these are still issues in 1.8 or 2.0 (I bet they are not), then please let me know or submit a new PR against one of those branches.

norru:
Hi @lgritz - Ditto! I was using the 1.2 branch because of the "official" build instructions in the README, tried to fix it - then shortly after realized that it was an obsolete one :)

I have successfully built with branch 1.8 - where similar warnings don't cause hard build failure.

Issue #2106, opened 2018-12-16
InputImage::_read_image bug ?


spinell:
I think I found a bug with InputImage::read_image(), read_tiles() and read_scanines()
When you use the paramater xStride, in some situation OIIO ignore the value of the xStride paramater.

Let suppose I have a allocated buffer of unsigned char that will containt RGBA values.
When I call read_image with the format UCHAR and a xStride of 4 bytes to read by example a bmp file that contains RGB, the values goes in my buffer at the right place (OIIO skip the alpha channel in my buffer). But if I'm reading by example an OpenEXR image that only have the red channel and the data in the file are float, OIIO ignore my stride value.

I think the problem is, if the pixel size in the image file is equal to the xStride value, OIIO ignore the stride and write the data in continious memory.

Note: I tried with version 1.8

spinell:
yes my buffer is unsigned char.
the format paramater is UCHAR.
the xStride is 4.

The image file internal data is float and the image only have the red channel.
OIIO convert the float to UCHAR write in my buffer without using the stride.

I can repro the same behaviour with an image file that only have the Red and Green channel
but the image file internal data is half-float.

lgritz:
OK, I think I understand. Looking into it...

lgritz:
Yes, I think I see how this happens. It's late, I will look at it more tomorrow. Should be an easy fix.

lgritz:
Fix proposed in #2108

lgritz:
Not sure why this issue didn't close when we merged the fix of #2108. But we did, so closing the ticket now.

PR #2108, opened 2018-12-17
Fix buffer stride logic in ImageInput read methods

lgritz:
Reported in #2106, there was a subtle bug in read_image/scanline/tile
where we fell into the "contiguous and no data type translation"
case when that was not really the case.

The symptom could be triggered by the following case: You want to read a
1-channel float image into a buffer meant to hold a 4-channel UINT8
image with a 4-byte xstride. The correct behavior is to translate the
floats into uint8's and place them in channel 0 of each 4-channel pixel.
But because the buffer stride length (4 x uint8 = 4 bytes) matched the
native stride length (1 x float = 4 bytes), it was copying the floats
instead of translating and depositing them correctly in channel [0]
(leaving channels [1-3] unchanged).

The fix isn't too hard, but we made the same mistake in several places.
Basically, we were comparing whether the strides matched the native
spacing, instead of whether they matched the buffer spacing. Only
particular unusual combinations would trigger the false positive.

Add an explicit test of this case in imagebuf_test.cpp.

### DIFF
--- a/src/libOpenImageIO/imagebuf_test.cpp
+++ b/src/libOpenImageIO/imagebuf_test.cpp
@@ -500,6 +500,93 @@ test_roi()
 
 
 
+// This tests a particular troublesome case where we got the logic wrong.
+// Read 1-channel float exr into 4-channel uint8 buffer with 4-byte xstride.
+// The correct behavior is to translate the one channel from float to uint8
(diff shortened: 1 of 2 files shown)


lgritz:
Fixes #2106

PR #2109, opened 2018-12-18
Adjust how maketx -u compares command lines

### LINKED ISSUE #1472: maketx -u should ignore paths
maketx -u  will store the command used to run it and then use that in the future to see if the texture should be updated.  This is overly conservative because the paths can benignly differ between uses.  For instance, the following might all be the same but maketx -u will consider them all different:

`/Users/JohnDoe/oiio/maketx -u /full/path/to/file/on/server/foo.jpg`
`/Users/OtherPerson/oiio/maketx -u /full/path/to/file/on/server/foo.jpg`
`/Users/OtherPerson/oiio/maketx -u k:\full\path\to\file\on\server\foo.jpg`
`maketx -u foo.jpg`

The different paths based on OS is especially problematic since I can't see of a reasonable way to avoid that for people that have a heterogeneous mix of machines.

My suggestion for how to solve this is that all the paths could be stripped of the leading parts of the path so that only the final filename is used (what the `basename` tool does).  In the above example, all of these commands would simplify to:  `maketx -u foo.jpg`. 

In fact, it might even be possible to completely get rid of the entire path. I don't think we actually need the filename for the source and destination and the name of the executable is also not needed since we already have the version number encoded in the "Software" string we're examining.  I haven't gone through all the other paths in maketx, but it's possible the others could also be entirely ignored?

This isn't 100% foolproof, for instance, on case insensitive filesystems foo.jpg and Foo.JPG are the same, but I hope that simply removing parts or entirety of paths will be good enough.

lgritz:
maketx -u means "update mode", wherein if it looks like there's already
a texture in place that's the result of the same maketx command, it
won't re-make the texture.

The primary way of doing this is to compare file dates: when using -u,
the output texture will be given the same timestamp as the input source
image. So the next time it runs, it checks the two dates, and if they are
identical, the texture is presumed already made.

But wait, what if they used a different version of OIIO, or had
different command line args, such as choosing a different fileter? Then
you DO want to remake the texture. So we also compare the metadata that
gives the command from last time, with the command line from this time.

But... that can have false negatives, because what if they used an
absolute path one time and a relative path another time?

So in this patch, we add a little trick: before comparing the two
command lines, we take each of them and for any argument that looks like
a file path (has a directory separator in it), we strip out everything
before the last directory separator.  So

    /my/bin/maketx foo.exr -o ./foo.tx
    makex /path/to/foo.exr -o foo.tx

Used to look different, but now look the same.

How do we know that the things we strip are actual file paths? We don't,
and we do not care. No arguments to maketx that we care about are going
to have slashes in them. That won't be part of the filter name or any of
the other things we were trying to distinguish between.

To make this work right, we also modified the `command_line_string` function
in maketx.cpp and oiiotool.cpp, to make it double-quote any arguments with
spaces.

Fixes #1472

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -891,6 +891,39 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
 
 
 
+// Deconstruct the command line string, stripping directory names off of
+// any arguments. This is used for "update mode" to not think it's doing
+// a fresh maketx for relative paths and whatnot.
+static std::string
+strip_cmd_line(string_view cmdline)
+{
+    std::string out;
+    bool firstarg = true;
+    while (!cmdline.empty()) {
+        if (!firstarg)
+            out += ' ';
+        string_view s;
+        if (!Strutil::parse_string(cmdline, s))
+            break;
+        // OK, we're doing something stupid, we're just going to treat every
+        // arg as if it were a filename and if it has a directory separator,
+        // only keep what's left after the last dir separator. This trick
+        // works because there just aren't any expected non-file arguments
+        // to maketx (or oiiotool) are among the ones we are trying to
+        // compare for update mode. Anything that looks like a filename
+        // gets stripped for the comparison, and we're ok with that.
+        std::string stripped = Filesystem::filename(s);
+        if (stripped.find(' ') != std::string::npos)
+            out += Strutil::sprintf("\"%s\"", stripped);
+        else
+            out += stripped;
+        firstarg = false;
+    }
+    return out;
+}
+
+
+
 static bool
 make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
                   std::string filename, std::string outputfilename,
@@ -989,7 +1022,8 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
         }
         std::string newcmdline = configspec.get_string_attribute(
             "maketx:full_command_line");
-        if (lastcmdline.size() && lastcmdline == newcmdline) {
+        if (lastcmdline.size()
+            && strip_cmd_line(lastcmdline) == strip_cmd_line(newcmdline)) {
             outstream << "maketx: no update required for \"" << outputfilename
                       << "\"\n";
             return true;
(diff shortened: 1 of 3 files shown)


ThiagoIze:
It's a little risky, but I like the simplicity of this. I haven't yet compiled it, so I can't confirm if it works properly on Windows, but at least it looks like it should be good.

ThiagoIze:
Turns out in Autodesk's syncolor we can have slashes in the --colorconvert, for instance I think the following might be valid options (I haven't yet gotten syncolor installed so I can verify):
```
 --colorconvert "sRGB" "CanonLog3 / CinemaGamut Rec.709 rendering"
 --colorconvert "sRGB" "CanonLog2 / CinemaGamut Rec.709 rendering"
```
after path stripping both will compare the same.

I'm guessing the easiest fix would be to ignore this check on colorconvert args?

lgritz:
You're killing me, Thiago.

lgritz:
:-) Let me take a look. It's not just colorconvert, there are 4 or 5 that might have color space names.

ThiagoIze:
Yeah, this took my totally by surprise. I'm sure you'll have a better fix, but if it helps, here's the simple fix I went with which is probably good enough for me:
```diff
diff --git a/src/libOpenImageIO/maketexture.cpp b/src/libOpenImageIO/maketexture.cpp
index ecbb93e7..4def7b18 100644
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -905,6 +905,29 @@ strip_cmd_line(string_view cmdline)
         string_view s;
         if (!Strutil::parse_string(cmdline, s))
             break;
+
+        // While we wait for an official fix here's our fix for
+        // colorconvert to not strip false "paths"
+        if (s == "--colorconvert")
+        {
+            if (s.find(' ') != std::string::npos)
+                out += Strutil::sprintf("\"%s\"", s);
+            else
+                out += s;
+
+            if (!Strutil::parse_string(cmdline, s, true, Strutil::KeepQuotes))
+                break;
+            out += ' ';
+            out += s;
+
+            if (!Strutil::parse_string(cmdline, s, true, Strutil::KeepQuotes))
+                break;
+            out += ' ';
+            out += s;
+            firstarg = false;
+            continue;
+        }
+
         // OK, we're doing something stupid, we're just going to treat every
         // arg as if it were a filename and if it has a directory separator,
```

lgritz:
I think I have a more complete solution proposed in #2140

PR #2110, opened 2018-12-18
Allow ImageSpec::find_attribute of "datawindow" and "displaywindow"

lgritz:
These are "pseudo-attributes", they aren't metadata themselves as we store them in an ImageSpec, but
aggreagate from x,y,z,width,height,depth (and the full_ versions). The "datawindow" and "displaywindow" are OpenEXR nomenclature, but it's likely that people will want to refer to them that way.

Also, minor fix to `ImageCache::get_image_info` and `TextureSystem::get_texture_info`, to make the right kind of ImageSpec::find call to retrieve the pseudo-attributes.


lgritz:
Amended with a minor fix to `ImageCache::get_image_info` and  `TextureSystem::get_texture_info`, to make the right kind of ImageSpec::find call to retrieve the pseudo-attributes.

PR #2133, opened 2019-01-07
IC/TS: add optional `force` argument to `invalidate()` method

lgritz:
This works the same way as the `force` argument to `invalidate_all()` --
if true, invalidates unconditionally; if false, invalidates only if the
file's modifiation date has changes since the cache first opened it.

The default is true for back compatibility, but probably you will often
want `false`, only invalidating if the file has changed.

### DIFF
--- a/src/doc/imagecache.tex
+++ b/src/doc/imagecache.tex
@@ -844,7 +844,7 @@ only be called on a tile that has been requested by
 {\cf get_tile()} but has not yet been released with {\cf release_tile()}.
 \apiend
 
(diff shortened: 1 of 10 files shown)


ThiagoIze:
This is causing one of our tests to fail where we load a texture, flip the texture_automip setting, invalidate_all(false), and load the texture again. OIIO has code to check for this in invalidate_all:
```

lgritz:
Oh yes, you're quire right. Fix coming up.

PR #2135, opened 2019-01-09
[OpenVDB] Don't allow any exceptions to escape and use Filesystem::fopen

marsupial:
## Description
Changes for fopen on Windows and containing all exceptions that may be thrown.

### DIFF
--- a/src/openvdb.imageio/openvdbinput.cpp
+++ b/src/openvdb.imageio/openvdbinput.cpp
@@ -317,7 +317,7 @@ openVDB(const std::string& filename, const ImageInput* errReport)
     if (!Filesystem::is_regular(filename))
         return nullptr;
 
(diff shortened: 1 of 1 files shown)


src/openvdb.imageio/openvdbinput.cpp
```diff
@@ -332,22 +332,27 @@ openVDB(const std::string& filename, const ImageInput* errReport)
...
+        } sVDBLib;
+
+        VDBFile file(new io::File(filename));
+
         file->open();
         if (file->isOpen())
             return file;
+
     } catch (const std::exception& e) {
         errReport->error("Could not open '%s': %s", filename, e.what());
```
lgritz: If this is caught here, will you end up with two error messages? One here, then a second below?

lgritz:
LGTM.

PR #2140, opened 2019-01-12
Modify maketx command line directory stripping to be smarter

lgritz:
For the purposes of deciding whether a `maketx -u` command is an
unnecessary repeat, we compare this command line to the last one to see
if it's the same. In so doing, we strip the directories off any
arguments that look like filenames so that changes is how relative path
names are expressed don't cause us to unnecessarily remake the texture.

It's a little naive to do this for all args. Thiago noticed situations
where color space names (args after `--colorconvert`) might have slashes
and so you don't want to strip them of their "directories."

This patch addresses that, and also several other maketx/oiiotool
commands that might have following args that look like filenames but in
fact are not. Also uniformize the interchangeable -cmd and --cmd so that
those pointless differences don't incorrectly trigger rebuilds by
looking different when they are not.


lgritz:
How does this look to you, @ThiagoIze ?

ThiagoIze:
Passes my tests and looks good, thanks!

PR #2144, opened 2019-01-14
pybind11: turns out we can't use < 2.2

lgritz:
So now we check the version, if system install is too old, fall back
to attempting to download a local copy.


lgritz:
Updated, taking more care with old pybind11 that stores common.h in a different place.

PR #2145, opened 2019-01-14
Test for new enough version of pybind11

lgritz:


lgritz:
This is an alternate to #2144, but this one is a much more minimal approach.

lgritz:
Abandoning this approach in favor of #2144

Issue #1844, opened 2018-01-17
maketx for single channel dwaa 


ThiagoIze:
When running maketx on a single channel image and specifying exr dwaa output, I get a fail.

To repro (tested on latest master):
```
% oiiotool --pattern checker 512x512 1 -o test.jpg
% maketx --format exr --compression dwaa test.jpg
maketx ERROR writing "test.8d714779.temp.tx" : Failed OpenEXR write: Failed to write pixel data to image file "test.8d714779.temp.tx". Data compression (zlib) failed.
```
But, if I create the dwaa exr directly through oiiotool then it seems to work:
```
% oiiotool --pattern checker 512x512 1 --compression dwaa -o test.exr
% iinfo -v test.exr
test.exr :  512 x  512, 1 channel, float openexr
    channel list: Y
    oiio:ColorSpace: "Linear"
    compression: "dwaa"
...
```

lgritz:
Accept another apology from me for letting this issue fall through the cracks.

After some investigation, it seems that any attempt to write tiled exr with dwaa compression will hit this error if the tile size is < 16x16. 

In your original example, the reason it failed for maketx but not the oiiotool command is that the maketx, in writing the MIP levels, eventually gets down to 8x8 resolution and that tile size fails.

Here's an even simpler repro case that doesn't involve MIP maps:

    # works
    oiiotool --pattern checker 512x512 1 -d half -compression dwaa -tile 16 16 -o test.exr
    
    # fails
    oiiotool --pattern checker 512x512 1 -d half -compression dwaa -tile 8 8 -o test.exr

Sure seems like an OpenEXR internals bug. I will report to them, I don't think there's much I can do on my side.

lgritz:
Closing. Fixed by #2147.

PR #2147, opened 2019-01-15
OpenEXROutput: skip dwa compression for single channel small tile images

rmv:
dwa? compression fails when saving single channel files with small tile sizes (< 16)

Larry determined in #1844 that the underlying issue is in OpenEXR.

While that gets fixed we should skip dwa compression for the failure cases. This patch replaces dwa compression with the default zip for the failing cases.

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -681,9 +681,9 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
         comp = "zips";
     // For single channel images, dwaa/b compression only seems to work reliably
     // when size > 16 and size is a power of two
-    if (spec.nchannels == 1 && Strutil::istarts_with(comp, "dwa") &&
-        ((spec.tile_width < 16 && spec.tile_height < 16) ||
-         !ispow2(spec.tile_width) || !ispow2(spec.tile_height))) {
+    if (spec.nchannels == 1 && Strutil::istarts_with(comp, "dwa")
+        && ((spec.tile_width < 16 && spec.tile_height < 16)
+            || !ispow2(spec.tile_width) || !ispow2(spec.tile_height))) {
         comp = "zip";
     }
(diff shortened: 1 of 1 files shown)


src/openexr.imageio/exroutput.cpp
```diff
@@ -679,6 +679,11 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
     // on deep files (but allow "none" as well)
     if (spec.deep && comp != "none")
         comp = "zips";
+    // Currently dwaa/b compression does not work on single channel with tile < 16
+    if (spec.nchannels == 1 && spec.tile_width < 16 && spec.tile_height < 16 &&
```
lgritz: Is it if *both* width and height are < 16? Or if either one is?
I don't know/remember, just asking.

lgritz:
LGTM. I'll await your answer to the one question I had, and for the CI tests to return successfully. Then I'll merge. Thanks for the fix.

lgritz:
Fixes #1844

rmv:
For instance 31x15 and 16x8 fail, but 31x16 does not

The patch does not fix all the underlying failure conditions with dwa compression, it fixes the ones that affect maketx (8x8, 4x4, etc), so:

Fixes #1844

lgritz:
I'm not sure we should knowingly still have a bunch of failure cases.

Should we just consider dwa + 1-channel to be unsafe for anything other than power-of-2 square tile sizes >= 16? And make everything else automatically revert to zip, until we know there's a robust fix on the OpenEXR side? (Keeping in mind that this will still allow dwaa for all the maketx and other uses that we truly care about.)

rmv:
Sounds good, there were still quite a lot of failing combinations for non power of 2 tile sizes under 64x64, see new patch.

lgritz:
LGTM, I will merge after the CI tests come back.

rmv:
Thanks!

PR #2148, opened 2019-01-15
Allow optional library name suffix

lgritz:
This patch allows an optional suffix to be specified that will let you
produce something like libOpenImageIO_CIA.so or whatever. Use with
caution, or preferably not at all, if you don't absolutely need it).

Example of how it's used: Suppose you're at a facility that needs (don't
ask why) multiple versions of OpenImageIO to be built. And namespacing
is not enough, because if you have two copies of libOpenImageIO.so, you
can't separately rpath them.


fpsunflower:
LGTM

PR #2149, opened 2019-01-16
Bug: fixed possible null ptr dereference

lgritz:
Found from static analysis!


fpsunflower:
Nice! Which static analyzer did you use?

lgritz:
clang-tidy

Every once in a while, I add one or two more checks (of the dozens and dozens that they have) to the set that I run, to see what it'll turn up.

PR #2160, opened 2019-01-24
HEIF support

lgritz:
HEIF is a container for HEIC-compressed images. HEIC is the still image
sibling to HEVC video compression. Roughly speaking, it produces lossy
compressed images about half the file size of JPEGs but with higher
visual quality.

At the moment, it's hard to find software that will reliably read
heic/heif, except for recent cuts of software within the Apple ecosystem
(the only ones I've encountered so far are iPhone camera images). But
that will surely change, since it's a superior format to JPEG in many
ways.

This is the first stab at supporting heif/heic images using the open
source libheif.  I'm fairly confident in the reading, including exif
metadata, of RGB and RGBA images, including files with multiple images.
I've not yet figured out depth images.

This can also write images (but only single images per file, RGB or
RGBA), and although the exif metadata I write can be read again by
OIIO's heif reader, Apple's Preview.app does not recognize the metadata
for some reason. I don't know why, maybe I'll return to that later.
I also want to fix later the limitation of not being able to output
multi-image heif files.

Along the way, I converted decode_xmp() utility to be based on
string_view, and also changed encode_exif() to take an explicit
'endianreq' parameter so it's possible to request a specific
endianness. (Apple's Preview.app, which is the main way I was able to
test whether I'm writing them correctly, *requires* that the exif be
written as big endian, or it won't recognize the file as proper heic at
all. So that required some extensive changes in my exif blob code.)

The latest release of libheif, 1.3, doesn't fully support color
profiles.  The in-progress libheif master (future 1.4) does, so I'll add
support when that is released. For now, it always looks like sRGB, and
that seems to be the common case, but it can be incorrect.

Currently, we only will read and write uint8 pixel data. The libheif
authors are working on full support for HDR/extended range, but that's
not even merged into its master yet. I'll add support in the OIIO heif
reader as that gets merged into libheif releases.


jaggedmgroup:
Please take me off your email list

lgritz:
@jaggedmgroup This is not an email list. If you're getting notifications about pull requests, it's because you are "watching" the GitHub repository for this project. It's not something we can fix for you, you'll have to click the "unwatch" button on your end.

src/doc/builtinplugins.tex
```diff
@@ -403,6 +403,36 @@ \section{HDR/RGBE}
...
+\index{HEIF} \endex{HEIC}
+
+HEIF is a container format for images compressed with the HEIC compression
+standard (same compression as HEVC/H.265). It is used commonly for iPhone
+camera pictures, but it is not Apple-specific and will probably become more
+popualar on other platforms in coming years. HEIF files usually use the file
+extension {\cf .HEIC}.
+
+HEIC compression is lossy, but is higher visual quality than JPEG while
+taking only half the file size. Currently, OIIO's HEIF reader supports
```
fpsunflower: Missing the end of this sentence?

src/heif.imageio/heifinput.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+    try {
+        auto id     = (subimage == 0) ? m_primary_id : m_item_ids[subimage - 1];
+        m_ihandle   = m_ctx->get_image_handle(id);
+        m_has_alpha = m_ihandle.has_alpha_channel();
+        auto chroma = m_has_alpha ? heif_chroma_interleaved_RGBA
+                                  : heif_chroma_interleaved_RGB;
+        m_himage = m_ihandle.decode_image(heif_colorspace_RGB, chroma);
+
+    } catch (const heif::Error& err) {
```
fpsunflower: Its a shame that the C++ wrapper for `libheif` converts error codes into exceptions and that you have to catch the exception to turn it back into an error code ...

src/doc/builtinplugins.tex
lgritz: Apparently, yes. Will fix.

fpsunflower:
Other than the docs - LGTM!

(the exception observation is more of a rant than a comment, obviously you are stuck with the decision that the `libheif` maintainers made -- unless you wanted to use the C api directly which is probably not worth it at this point)

lgritz:
I could use the C API directly, but the C++ was looked more compact and attractive, and so I was most of the way through implementing it all before I realized how many try/catch blocks I needed.  Didn't feel like rewriting the whole thing, wordier, just to eliminate exceptions.

lgritz:
I pushed an update to fix the docs typos.

src/doc/builtinplugins.tex
fpsunflower: LGTM!

PR #2161, opened 2019-01-26
Add ability to control Python version in CMakeLists.txt

fredrikaverpil:
## Description

Set environment variable `OIIO_PYTHON_VERSION` to e.g. `3.6` to change the default value of "2.7" in CMakeLists.txt, to make it easier to build for Python 3.

I am using `vcpkg install openimageio:x64-windows` but am having a hard time with the fact that this value is hard coded in CMakeLists.txt. I could maintain a patch and sed the [portfile.cmake](https://github.com/Microsoft/vcpkg/blob/master/ports/openimageio/portfile.cmake) to include such a patch, but I would much rather just make sure an environment variable is set.

Before moving on, would you accept this kind of change?

## Checklist:


lgritz:
LGTM. I will merge into master and RB-2.0 after the CI tests finish running.

fredrikaverpil:
Great, thank you!

PR #2162, opened 2019-02-02
SPI only: Coalesce the arnold and spcomp2 scripts into just one

lgritz:
**Non-SPI people: ignore this entirely, it doesn't change anything for you.**

Build for SPI Arnold:

    make

Build for SpComp2:

    make OIIO_SPCOMP2=1 spcomp2
  
or

    make OIIO_SPCOMP2=1 spcomp2_install_local

Build completely generic, avoiding any special Arnold build logic:

    make OIIO_ARNOLD=0

I know that the diffs are difficult to follow in this format, because GitHub is showing you the change relative to Makefile-bits. If only I could tell it to show you the new Makefile-bits compared to the old Makefile-bits-spcomp2, it would all be *so clear* because those diffs are relatively small and easy to understand.  But basically I was starting from a point where site/spi/Makefile-bits just looked at the build targets, and if they looked like spcomp2 stuff, it would include Makefile-bits-spcomp2, and otherwise would include Makefile-bits-arnold.  Those two files were already pretty similar, lined up well in the diffs.  So I just merged them, hiding the few differences between them in conditionals, and then moved the result to just plain old Makefile-bits and got rid of the separate customized files. In reality, if you were to diff the new one to either one of the old ones, it would be pretty close, with the obvious changes to merge in the missing parts of the other.

This whole thing is just for my own personal convenience when maintaining two different schemes of building in our weirdo environment.


lgritz:
YES, that's how I almost always build it since I'm usually building it in a custom location with potentially odd changes in progress.

PR #2166, opened 2019-02-05
Fix bug in IC::invalidate()

lgritz:
When we recently made invalidate(file) have a 'force' parameter, we
botched it by not so forcing a particular call in invalidate_all().
In the line being changed here, we're dealing with a list of files
that definitely need invalidation due to special tests, so once
we've established that, we want to force it.

Pointed out by @ThiagoIze in a post-merge comment in #2133


ThiagoIze:
I can confirm this fixes my tests.

lgritz:
This doesn't need any backports, right? The bug was only in master?

ThiagoIze:
It looks like this was just committed to master, so I think you're good :)
FYI: we've been officially running off of master for a few months now and plan to stick with master for the foreseeable future.

PR #2168, opened 2019-02-05
Color transform by matrix

lgritz:
`IBA::colormatrixtransform()` uses a 4x4 matrix to apply a color
transform to all pixels in the image region. The matrix is, following
Imath, OpenGL, and RenderMan-ish conventions, laid out in the order
that assumes row-vector colors and does math as C' = C*M. (Another way
to put that is that the "translation" part of the matrix is the bottom
row, not the right column.)

`oiiotool --ccmatrix` does the same thing from the command line, and
is followed by an argument containing either 9 or 16 values to form a
3x3 or 4x4 matrix. (If 3x3, it will be extended in the obvious way to
4x4 before passing to the underlying colormatrixtransform() call.
Additionally, the `--ccmatrix` command takes optional arguments that
will first invert the matrix or use its transpose (the latter in case
you are given matrix values in an order that implies column-vector
colors).


lgritz:
@michdolan look ok to you?

src/doc/imagebufalgo.tex
```diff
@@ -2261,6 +2261,51 @@ \section{Color manipulation}
...
+
+\smallskip
+\noindent Example:
+\begin{code}
+    ImageBuf Src ("tahoe.jpg");
+    Imath::M44f M ( .8047379,  .5058794, -.3106172, 0,
+                   -.3106172,  .8047379,  .5058794, 0,
+                    .5058794, -.3106172,  .8047379, 0,
+                     0,         0,         0,       1);
+    ImageBuf dst = ImageBufAlgo::colorconvert (Src, M);
```
michdolan:
```suggestion
    ImageBuf dst = ImageBufAlgo::colormatrixtransform (Src, M);
```

src/doc/pythonbindings.tex
```diff
@@ -3211,6 +3211,29 @@ \subsection{Color manipulation}
...
+
+\smallskip
+\noindent Examples:
+\begin{code}
+    Src = ImageBuf ("tahoe.jpg")
+    M = ( .8047379,  .5058794, -.3106172, 0,
+         -.3106172,  .8047379,  .5058794, 0,
+          .5058794, -.3106172,  .8047379, 0,
+          0,         0,         0,       1)
+    Dst = ImageBufAlgo.colorconvert (Src, M)
```
michdolan:
```suggestion
    Dst = ImageBufAlgo.colormatrixtransform (Src, M)
```

lgritz:
Yes, thanks for the cut-and-paste typo fixes!

Issue #2169, opened 2019-02-08
std::regex throws exception in windows 


ThiagoIze:
More fun in Windows land: We took a multichannel exr file and generated a tx file (.tiff under the hood), so that later when we operate on the .tx file, for instance by using oiiotool --info, we get a crash. In our application we catch exceptions (side note, should oiiotool catch exceptions too?), so instead of crashing we find this exception:
"regex_error(error_stack): There was insufficient memory to determine whether the regular expression
 could match the specified character sequence."

Debugging points to tiffinput.cpp with the exception being thrown from the `regex_replace` call in:
```c++
    found = desc.rfind("oiio:AverageColor=");
    if (found != std::string::npos) {
        size_t begin  = desc.find_first_of('=', found) + 1;
        size_t end    = std::min(desc.find_first_of(' ', begin), desc.size());
        string_view s = string_view(desc.data() + begin, end - begin);
        m_spec.attribute("oiio:AverageColor", s);
        const std::string average_pattern
            = std::string("oiio:AverageColor=(\\[?") + fp_number_pattern
              + ",?)+\\]?[ ]*";
        printf("desc=%s\n", desc.c_str());
        desc        = regex_replace(desc, regex(average_pattern), "");
        updatedDesc = true;
    }
```

This seems like it could be related to https://stackoverflow.com/questions/42909618/a-simple-c11-regex-that-throws-regex-error-while-searching , but for some reason setting a `#define _REGEX_MAX_STACK_COUNT 0` at the very top of tiffinput.cpp doesn't help even though looking at the regex source code makes it look like it should. In fact, doing that in the simpler patch below also doesn't help. Perhaps something sneaky is going on and this isn't getting used by the windows regex header? It'd be nice to figure this out, but I gave up and went with the below fix.

Since we still have the boost code, the following lets us build with boost whenever this is an issue:
```diff
diff --git a/src/cmake/compiler.cmake b/src/cmake/compiler.cmake
index 37a87776..47e2e233 100644
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -267,6 +267,16 @@ set (CMAKE_REQUIRED_DEFINITIONS ${CSTD_FLAGS})
 check_cxx_source_runs("
       #include <regex>
       int main() {
+
+          // This throws an exception on windows, at least with VS2015, when using std_regex
[...]

PR #2173, opened 2019-02-11
Avoid std::regex_replace

lgritz:
Certain regex_replace tasks seem to crash on Windows with MSVS's implementation of std::regex.

After considering a more sophisticated test for when we should use boost.regex instead, I instead settled on just replacing the few uses of regex_replace with our own string processing. None were very    sophisticated or truly needed regex functionality.


lgritz:
Fixes #2169

ThiagoIze:
I can confirm this fixes my problem and side bonus is the code looks much nicer!

PR #2174, opened 2019-02-17
 TextureSystem option to limit mip levels

lgritz:
This patch adds a "max_mip_res" attribute for the TextureSystem that
limits filtered texture access to MIP levels that are no higher than
this resolution in any dimension. The default is 1<<30, which means
there is no effective limit. As an example, setting this attribute to
256 means that all texture access will be limited to the MIP levels
that are 256x256 pixels or lower. This can be useful for preview
renders where you want to cut down on the amount of disk or network
access for the sake of speed.

In a subsequent patch, I intend to also provide a way to set this
on a per-texture basis, but as a first step we'll start with the
global setting.

Along the way, I modified runtest.py to restore its ability to run
by hand from inside the build/PLATFORM/testsuite/TEST directory, which
had been lost recently with some overhaul to the testing scripts.


fpsunflower:
LGTM!

This should allow for some interesting experiments ...

PR #2184, opened 2019-03-02
TS: Fix possible nullptr dereference

### LINKED ISSUE #2179: segfault on get_image_handle
### Problem

**Expected behavior:**  
Calling `ImageCache::get_image_handle(path)` should return NULL or an image handle.

**Actual behavior:**  
`get_image_handle` dereferenced a null pointer and segfaulted.

I'm using OpenImageIO 1.8.16 (installed via vcpkg on windows). I did not have the time to compile and test the latest master.

The null pointer dereference seems to happen in the `find_file` function: 
https://github.com/OpenImageIO/oiio/blob/cbdf585a6b69b6a2a5dcff4be5a36c442960a694/src/libtexture/imagecache.cpp#L1157

The code assumes that `thread_info != NULL`, but the function is called by `get_image_handle` with `thread_info=NULL` (the default argument), which is valid according to the documentation.

At first glance, the latest master also makes the same assumption:
https://github.com/OpenImageIO/oiio/blob/3c4a7af1846918376249b1b12698090c5d79768c/src/libtexture/imagecache.cpp#L1192

### Versions

* OIIO branch/version: 1.8.16
* OS: Windows 10 x64
* C++ compiler: MSVC 2017

lgritz:
find_file() assumes a non-null thread_info parameter.  But it's called
from get_image_handle in a way that could pass a null pointer. Fix.

Fixes #2179

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -943,6 +943,8 @@ public:
     get_image_handle(ustring filename,
                      ImageCachePerThreadInfo* thread_info = NULL)
     {
+        if (!thread_info)
+            thread_info = get_perthread_info();
         ImageCacheFile* file = find_file(filename, thread_info);
         return verify_file(file, thread_info);
     }


lgritz:
Reporter of the problem says this patch fixes the problem. Merging.

PR #2190, opened 2019-03-05
exr read: suppress empty string submage name

lgritz:
It's been reported that V-Ray writes images with this metadata field
present but empty. If you happen to take those images (and their
metadata) and turn them into a multi-part exr, the names all being the
same is problematic. The expedient fix is just to ignore empty names
on input.  Suggested by Chris Allen.


lgritz:
Chris voices approval on the mail ist.

PR #2192, opened 2019-03-13
Argparse -- add test, %1

lgritz:
ArgParse had no unit test! Add one.

Also, add a new feature: %1 is like %* (catch-all for all
miscalleneous arguments that aren't dash-led commands and their
following data).  But if %1 is present, its list will receive any
arguments that come BEFORE any other dash-led args, whereas the %*
will receive all the misc args that happen after the first dash-led
arg. If there is no %1, then %* will capture all the misc args
regardless of whether they are before or after the first dash-led arg.

Why is this useful? So you can have programs that require certain
arguments before any options. Like "git add ..." "git rm ...", you
would achieve something like that via:

    std::string mode;
    std::string files;

    static int
    callback(int argc, const char* argv[])
    {
        for (int i = 0; i < argc; i++)
            files.emplace_back(argv[i]);
        return 0;
    }

    static int
    mode_callback(int argc, const char* argv[])
    {
        if (mode == "" && argc == 2)
            mode = std::string(argv[1]);
        else
            callback(argc, argv);
        return 0;
    }

    ArgParse ap;
    ap.options(
        "My git replacement",
        "%1", initial_callback, "",
        "%*", callback, "",
        ...
        nullptr);
    std::cout << "Mode was " << mode << "\n";
    std::cout << "Files were: " << Strutil::join(files, ", ") << "\n";

### DIFF
--- a/src/include/OpenImageIO/argparse.h
+++ b/src/include/OpenImageIO/argparse.h
@@ -120,6 +120,9 @@ class ArgOption;  // Forward declaration
 ///                  int callback (int argc, char *argv[])
 ///    - \%* - catch all non-options and pass individually as an (argc,argv)
 ///            sublist to a callback, each immediately after it's found
+///    - \%1 - catch all non-options that occur before any option is
+///            encountered (like %*, but only for those prior to the first
+///            real option.
 ///
 /// There are several special format tokens:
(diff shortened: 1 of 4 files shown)


fpsunflower:
LGTM!

PR #2194, opened 2019-03-14
CircleCI - prototype

lgritz:
A bit of material has been refactored out of .travis.yml into separate
scripts so it can be shared with circleci.

CircleCI is an alternative CI system like TravisCI or Appveyor.
Mostly I'm doing this as a prototype and to learn about it, because
it's under discussion as a potential favored CI system for ASWF.

But as it turns out, the VMs provisioned by CircleCI are more capable
CPUs than the ones with Travis. In particular, it looks like they
support AVX-512. Right now, the Travis VMs top out at AVX 1.0, so
although we have Travis test matrix entries that "build" for AVX2 and
AVX512 just to make sure we haven't broken the build, we aren't able
to run the tests when compiled for that architecture. But we can on
CircleCI. So I'm thinking of throwing CircleCI into the mix just for
certain build configurations that aren't possible with Travis, to get
better test coverage.

### DIFF
--- a/src/build-scripts/circleci-installdeps.bash
+++ b/src/build-scripts/circleci-installdeps.bash
@@ -27,7 +27,14 @@ time apt-get install -y \
     libavcodec-dev libavformat-dev libswscale-dev libavutil-dev \
     locales \
     opencolorio-tools \
-    wget
+    wget \
+    libtbb-dev \
+    libopenvdb-dev \
+    libopencv-dev \
+    ptex-base \
+    dcmtk
+
+
 # time apt-get install -y g++-4.8
 #time apt-get install -y g++-7
 #time apt-get install -y g++-8
--- a/src/build-scripts/circleci-installdeps.bash
+++ b/src/build-scripts/circleci-installdeps.bash
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+#
+
+set -ex
+
+#dpkg --list
+
+time apt-get update
+
+time apt-get install -y git
+time apt-get install -y cmake
+time apt-get install -y g++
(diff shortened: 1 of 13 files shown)


src/build-scripts/circleci-installdeps.bash
```diff
@@ -0,0 +1,52 @@
...
+time apt-get install -y libfreetype6-dev
+#time apt-get install -y libjpeg-turbo8-dev
+time apt-get install -y dcmtk
+time apt-get install -y libavcodec-dev
+time apt-get install -y libavformat-dev
+time apt-get install -y libswscale-dev
+time apt-get install -y libavutil-dev
+time apt-get install -y locales
+time apt-get install -y opencolorio-tools
+time apt-get install -y ninja-build
```
zxiiro: All these individual apt-get install calls add a significant amount of time to the build. It would be better to do something like:

```
apt-get install -y \
    package-a \
    package-b \
    package-c
```

lgritz: Oh? It never occurred to me that that doing them on separate lines would have any important performance impact. I assumed that the speed would mostly be limited by the download speed of the packages themselves and their subsequent install. Thanks for the tip!

zxiiro: Yep. The actual call to "apt-get" takes a few seconds (on slow systems much longer) when you're just doing 1 or 2 package installs this way you don't notice it as it's just a couple seconds. But once you get into the 10s of packages the valuable seconds really add up.

Doing a `apt-get install -y package-a package-b package-c` instead causes apt-get to init only once and thus will be much faster if you are installing many packages.

PR #2196, opened 2019-03-18
Forward ProgressCallback in ImageBufImpl::read

marsupial:
My callback was never being called!


lgritz:
Yeah, that'll do it!

LGTM.  I will merge when the rest of the CI tests return.

PR #2197, opened 2019-03-18
Make OIIO:invert a host/device function

marsupial:
## Description
This can get rid of some macro ugliness in https://github.com/imageworks/OpenShadingLanguage/pull/974


lgritz:
LGTM

Aside: I wish Cuda had a way to mark an entire class, header file, or translation unit as "everything here is __HOST__ __DEVICE__ unless I mark it otherwise."

marsupial:
Think this landed in clang:
`#pragma clang force_cuda_host_device_{begin,end}`
http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20161003/173100.html

PR #2199, opened 2019-03-22
Rename pow2roundup/pow2rounddown -> ceil2, floor2

lgritz:
This is in keeping with the nomenclature of the C++20 `<bit>` header which
contains the equivalent functions.

Also give them the constexpr and noexcept treatment.

### DIFF
--- a/src/iv/ivutils.h
+++ b/src/iv/ivutils.h
@@ -10,7 +10,7 @@ OIIO_NAMESPACE_BEGIN
 /// representation.  Once optimized and tested, move to fmath.h
 
 inline float
-pow2roundupf(float f)
+ceil2f(float f)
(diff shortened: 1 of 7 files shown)


src/iv/ivutils.h
```diff
@@ -10,7 +10,7 @@ OIIO_NAMESPACE_BEGIN
 /// representation.  Once optimized and tested, move to fmath.h
 
 inline float
-pow2roundupf(float f)
+ceil2f(float f)
 {
     float logval = logf(f) / logf(2.0f);
```
fpsunflower: Totally unrelated to this review, but this badly calls for a native implementation that just uses the float exponent :)

PR #2202, opened 2019-04-05
Better oiiotool support of multiple subimages

lgritz:
Background:

By default, most oiiotool operations ignore multiple subimages. For
commands that modify metadata in place, that means it will only modify
the first subimage (leaving the rest untouched). For commands that
remove inputs from the stack and produce outputs to be put back on the
stack, generally their output will only be based on the first
subimage, and subsequent subimages will be dropped.

`-a` says that any ops ought to work on all subimages if at all
possible; that means that in-place metadata modifiers will apply to
all subimages, and stack ops should produce outputs with the same
number of subimages as their inputs, doing the right operation to each
subimage in turn.

Most commands (but not 100%) also respect an optional
`:allsubimages=VAL` that override `-a` on that one command, `=1` turns it
on even if you didn't use `-a`, and `=0` turns it off even if you did use
`-a`.

My goal is to make all commands respect the `-a` and `:allsubimages=VAL`
options, in all cases except where there is no sensible interpretation
to multiple subimages. There are a few now where there is a sensible
behavior, but it's just not implemented.

This patch fixes the following:

* These commands previously honored `-a` but did not respect a ":allsubimages="
  modifier, but now they do:
    --sattrib   --attrib   --caption   --clear-keywords
    --iscolorspace  --orientation  --clamp  --fixnan

* These were totally unaware of subimages, neither responding to -a
  nor to ":allsubimages=", now they respect both:
    --crop  --fullsize  --zover  --fill

* These previously always operated on all subimages, but now work on
  only one subimage by default (like the rest) and respect the -a and
  ":allsubimages=" modifier:
    --origin   --fullpixels  --croptofull  --trim

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -877,7 +877,14 @@ set_string_attribute(int argc, const char* argv[])
         ot.warning(argv[0], "no current image available to modify");
         return 0;
     }
-    set_attribute(ot.curimg, argv[1], TypeString, argv[2], ot.allsubimages);
+
+    string_view command = ot.express(argv[0]);
+    std::map<std::string, std::string> options;
+    options["allsubimages"] = std::to_string(ot.allsubimages);
+    ot.extract_options(options, command);
+    bool allsubimages = Strutil::from_string<int>(options["allsubimages"]);
+
+    set_attribute(ot.curimg, argv[1], TypeString, argv[2], allsubimages);
     // N.B. set_attribute does expression expansion on its args
     return 0;
 }
@@ -893,11 +900,14 @@ set_any_attribute(int argc, const char* argv[])
         return 0;
     }
 
+    string_view command = ot.express(argv[0]);
     std::map<std::string, std::string> options;
-    ot.extract_options(options, argv[0]);
+    options["allsubimages"] = std::to_string(ot.allsubimages);
+    ot.extract_options(options, command);
     TypeDesc type(options["type"]);
+    bool allsubimages = Strutil::from_string<int>(options["allsubimages"]);
 
-    set_attribute(ot.curimg, argv[1], type, argv[2], ot.allsubimages);
+    set_attribute(ot.curimg, argv[1], type, argv[2], allsubimages);
     // N.B. set_attribute does expression expansion on its args
     return 0;
 }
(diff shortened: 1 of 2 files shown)


src/oiiotool/oiiotool.cpp
```diff
@@ -1792,9 +1828,15 @@ set_full_to_pixels(int argc, const char* argv[])
     Timer timer(ot.enable_function_timing);
     string_view command = ot.express(argv[0]);
 
+    std::map<std::string, std::string> options;
```
aconty: Any chance these repeated 4 lines could be put into a function somewhere?

fpsunflower:
How do the crop (and related) operations work when the subimages are all of varying sizes? Would it make sense to warn the user at least?

src/oiiotool/oiiotool.cpp
lgritz: Well, I'm planning an overhaul of this soon anyway, and it's going to change a lot, including a lot less repetition of some of these idioms.

lgritz:
Chris, that's an interesting question. The fact is that multi-subimge files aren't super common (and aren't even supported by most file formats). There are only four cases I can think of:

1. TIFF files may contain multiple images, and we hijack this for MIP maps, but I rarely see any tiff images in the wild that have multiple subimages.

2. Multi-part exrs, which I think are not allowed to have differing windows among the parts, and are designed to represent, essentially, different sets of channels of the same view, or multiple views of the same image "format" (like left and right eye, they'll always be the same resolution and aspect).

3. Movie files (which look like multi-subimage files to OIIO), which are always the same size for every frame.

4. Ptex, where each subimage looks like a totally different face of a texture and can have differing resolutions, etc. I don't know if anybody ever uses oiiotool to try to modify ptex images, and surely they wouldn't try to "crop" or do anything like that, because almost no image morphology-changing operation would make sense to do to all the texture faces identically.

So I suppose I could, on these operations, walk through all the subimages, comparing sizes and warn or error if they differ? I don't know if it's a case that would come up in the real world. It could only happen to Ptex textures and true multi-image TIFFs, neither of which intersects commonly with the oiiotool world as far as I know.

Do this now? Or better to sweep it under the rug until somebody complains that they have done something weird?

fpsunflower:
Sounds like it's uncommon enough not to matter right now then. For some reason I thought EXR did allow varying the resolution per sub-image. In any case it will be rather obvious when something does go wrong, and the existing options to limit which subimage gets operated on seem like they'd be flexible enough to express the true intent.


LGTM!

PR #2203, opened 2019-04-05
Make OIIO_DISPATCH_COMMON_TYPES2/3 handle more type combinations.

lgritz:
Internal to ImageBufAlgo functions, we are faced with the fact that we
have 1, 2, or 3 input images in addition to an output image, and they
each could be any of the 9 pixel data types we support (float, uint8,
uint16, half, int8, int16, uint, int, double).  This means that the
template expansion of all possible combination would tend to be 9^2=81
specializations for unary functions, 9^3=729 for binary functions, and
9^4=6561 for ternary functions (I think mad() is the only ternary
function we have).

This is unworkable, leads to infeasible compilation times, so for a
long time we have made the following shortcut: handle the full
cross-product of just the 4 commonly used types (uint8, uint16, float,
half), and in all other cases convert to and from float. So that means
unary needs 16 template specializations, binary 64, and ternary 256.

But... other cases do crop up, so with this patch, I'm expanding the
range of cases that OIIO_DISPATCH_COMMON handles without
conversion. If all of the types (result, and inputs) are identical, we
handle it with the native type. This is only slightly more expensive:
unary functions need 9+12=21 template specializations (why 12 instead
of 16? because 4 combinations are already handled by the "all same type"
case, and therefore don't need to be instantiated again), binary
functions need 9+60=69, and ternary 9+252=261.

We think that, outside the "common" cases, the most common uncommon
ones are the "all types the same" cases. So accommodating this seems
worth the slight increase in compile time, which is nowhere near the
combinatorial explosion of the full cross-product of all types,
and in the process we make it even more rare that we fall into the
"plan B" case of converting to and from float to handle the rare
oddball cases.

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo_util.h
+++ b/src/include/OpenImageIO/imagebufalgo_util.h
@@ -339,31 +339,60 @@ inline TypeDesc type_merge (TypeDesc a, TypeDesc b, TypeDesc c)
     }
 
 // Macro to call a type-specialzed version func<Rtype,Atype>(R,A,...) for
-// the most common types. It will auto-convert other cases to/from float.
+// the most common types, and even for uncommon types when src and dst types
+// are identical. It will auto-convert remaining rare cases to float.
 #define OIIO_DISPATCH_COMMON_TYPES2(ret,name,func,Rtype,Atype,R,A,...)  \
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

PR #2204, opened 2019-04-08
Enhancements to the syntax for setting and retrieving attributes.

lgritz:
Lots of important classes have "attributes", which are used as option
settings and the like, basically any kind of named/typed data that
objects carry around. The prototypical way we set and retrieve them
is:

    obj.attribute (name, typedesc, &data);
    found_ok = obj.getattribute (name, typedesc, &data);

This is very general, but it's kind of a wonky syntax. So this patch
introduces an additional synax that is super handy and aesthetically
pleasing:

    obj[name] = data;     // set, works for nearly any data!

    data = obj[name]->get<type>();   // any type
    mystring = obj[name];            // turns any data into a string

This magic is accomplished with a helper template called an
AttrDelegate, which we define in attrdelegate.h. This is pretty handy
and we anticipate using it more widely over time, as we add this syntax
to additional classes.

The first class we rig up in this way is ParamValueList, which
necessitates adding attribute() and getattribute() methods to begin
with. It's a natural fit, and you can get a feel for how it looks when
used by examining the new unit tests in paramlist_test.cpp.

Also added to ImageSpec, for which I think this will be a handy new
way to set and retrieve metadata.

There's a bunch more cleanup that should be fairly straightforward to
understand.

Also, we expose two new functions (pieces of which were used internally
before, though we have now made them public and refactored them):

    std::string tostring (typedesc, void* data,
                          const tostring_foratting& fmt={})

        Turns any data (described by the typedesc) into a string,
        with optional fine-grained control over the formatting.

    bool convert_type (TypeDesc srctype, const void* src,
                       TypeDesc dsttype, void* dst)

        Copies with data conversion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,7 +319,8 @@ endif()
 if (USE_PYTHON AND NOT BUILD_OIIOUTIL_ONLY)
     if (NOT SANITIZE_ON_LINUX)
         oiio_add_tests (
-                python-typedesc python-imagespec python-roi python-deep
+                python-typedesc python-paramlist
+                python-imagespec python-roi python-deep
                 python-imageinput python-imageoutput
                 python-imagebuf python-imagebufalgo
                 IMAGEDIR oiio-images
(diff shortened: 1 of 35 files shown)


lgritz:
If nobody objects to this within a couple days, I will merge it later this week.

Issue #2205, opened 2019-04-08
crash reading certain tiff files


ThiagoIze:
If you create a tiff file using an older version of nconvert (try [5.0.2](https://download.xnview.com/old_versions/)) and then try to use it in OIIO we get a segfault:

```
% oiiotool --info -v  uv_test_nconvert.tif
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x2)
  * frame #0: 0x00000001007a1c2a oiiotool`_TIFFVGetField + 458
    frame #1: 0x000000010079f1e1 oiiotool`TIFFGetField + 209
    frame #2: 0x00000001006679b0 oiiotool`OpenImageIO_v2_1::TIFFInput::readspec(bool) + 3264
    frame #3: 0x000000010066670a oiiotool`OpenImageIO_v2_1::TIFFInput::seek_subimage(int, int) + 330
    frame #4: 0x000000010066650d oiiotool`OpenImageIO_v2_1::TIFFInput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, OpenImageIO_v2_1::ImageSpec&) + 61
    frame #5: 0x0000000100666cd5 oiiotool`OpenImageIO_v2_1::TIFFInput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, OpenImageIO_v2_1::ImageSpec&, OpenImageIO_v2_1::ImageSpec const&) + 181
    frame #6: 0x00000001005a634d oiiotool`OpenImageIO_v2_1::pvt::ImageCacheFile::open(OpenImageIO_v2_1::pvt::ImageCachePerThreadInfo*) + 1837
    frame #7: 0x00000001005abc94 oiiotool`OpenImageIO_v2_1::pvt::ImageCacheImpl::verify_file(OpenImageIO_v2_1::pvt::ImageCacheFile*, OpenImageIO_v2_1::pvt::ImageCachePerThreadInfo*, bool) + 356
    frame #8: 0x00000001005b509a oiiotool`OpenImageIO_v2_1::pvt::ImageCacheImpl::get_image_info(OpenImageIO_v2_1::pvt::ImageCacheFile*, OpenImageIO_v2_1::pvt::ImageCachePerThreadInfo*, int, int, OpenImageIO_v2_1::ustring, OpenImageIO_v2_1::TypeDesc, void*) + 74
[...]

PR #2206, opened 2019-04-10
Fix crash reading certain old nconvert-written tiff files

lgritz:
I don't claim to understand exactly why is is problematic, but certain
files written by an older versin of "nconvert" have something strange
about their tags that was leading to a crash. Solved with use of
our existing safe_tiffgetfield() versus TIFFGetField when retrieving
the PREDICTOR tag.

Also refactored safe_tiffgetfield a bit so that for the calls that
"fail" (results are ignored), it doesn't overwrite the results. Do this
by retrieving into a temp buffer and then copying to the caller's
memory upon success.

TODO: How many other TIFFGetField calls ought to be replaced by
safe_tiffgetfield? All? Some? What are the circumstances that lead to
this inexplicable problem? Is there a better tag-reading idiom we
should be using? Investigating this on all on the side, but it need
not hold up this immediate fix to the one known problem case.

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -239,17 +239,22 @@ private:
     // program!  Ick.  So to avoid this, we always push a pointer, which
     // we expect NOT to be altered, and if it is, it's a danger sign (plus
     // we didn't crash).
-    bool safe_tiffgetfield(const std::string& name, int tag, void* dest)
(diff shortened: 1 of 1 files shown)


lgritz:
Fixes #2205

lgritz:
Oh no, this is very broken!
Sorry, @ThiagoIze 
I am going to immediately revert this broken fix, then make a revised PR tonight.

lgritz:
Fixed in #2207 and re-merged.

PR #2211, opened 2019-04-17
IBA::resize -- fix precision for double images

### LINKED ISSUE #2210: resizing images of type double
https://github.com/OpenImageIO/oiio/blob/b32cc4eddef22f84ee6b92a821dad452b331650e/src/libOpenImageIO/imagebufalgo_xform.cpp#L235

Hey,

Thanks again for the wonderful library. We're probably outliers using doubles in our particular workflow.

This pel variable being type float* seems to cause just enough of a precision error in my resizing an image of type double to show up in the unit tests. (A value is 1.0000012 instead of 1.0000000). If I change it to type double it works great. 

Would making it double cause too many issues elsewhere? It seems to scale with the number of channels rather than the number of pixels, but maybe there's memory issues I'm not thinking of. It could also possibly be templated to DSTTYPE, with possible issues with lower bit-depth types.

                      -- Alex

lgritz:
The innards of IBA::resize() has to accumulate the weighted
contributions of source pixels under the filter window, and that
accumulator used 'float'. This is fine except when the input image is
'double', in which case you might want just a touch more precision.

So we define an Accum_t typedef template that is double when double is
needed, and float for all other cases. Adds the precision that double
image resize needs, but retains the speed and precision of the prior
float behavior in all other cases.

Fixes #2210

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_xform.cpp
+++ b/src/libOpenImageIO/imagebufalgo_xform.cpp
@@ -52,6 +52,13 @@ OIIO_NAMESPACE_BEGIN
 
 namespace {
 
+// Define a templated Accumulator type that's float, except in the case
+// of double input, in which case it's double.
+template <typename T> struct Accum_t { typedef float  type; };
+template<> struct Accum_t<double>    { typedef double type; };
+
+
+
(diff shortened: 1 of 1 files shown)


asuter:
This change fixes it for my use case. Thanks!

PR #2215, opened 2019-04-19
Fix a number of performance inefficiencies

DennisOSRM:
## Description

Fixes a number of performance-related issues, namely pass by value that should be const refs, avoiding unnecessary temporaries.

## Tests

No further tests added.

## Checklist:

### DIFF
--- a/src/fits.imageio/fits_pvt.cpp
+++ b/src/fits.imageio/fits_pvt.cpp
@@ -101,7 +101,7 @@ unpack_card(const std::string& card, std::string& keyname, std::string& value)
     size_t begin = 0, end = std::string::npos;
     if (card_cpy[0] == '\'') {
         begin = 1;
-        end   = card_cpy.find(''', 1);
+        end   = card_cpy.find('\'', 1);
     } else {
         end = card_cpy.find('/', 1);
     }
--- a/src/fits.imageio/fits_pvt.cpp
+++ b/src/fits.imageio/fits_pvt.cpp
@@ -101,9 +101,9 @@ unpack_card(const std::string& card, std::string& keyname, std::string& value)
     size_t begin = 0, end = std::string::npos;
     if (card_cpy[0] == '\'') {
         begin = 1;
-        end   = card_cpy.find("'", 1);
+        end   = card_cpy.find(''', 1);
     } else {
-        end = card_cpy.find("/", 1);
+        end = card_cpy.find('/', 1);
     }
 
     // after creating substring we strip NULL chars from the end
--- a/src/iinfo/iinfo.cpp
+++ b/src/iinfo/iinfo.cpp
@@ -418,7 +418,7 @@ print_metadata(const ImageSpec& spec, const std::string& filename)
 
 
 static const char*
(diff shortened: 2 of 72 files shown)


lgritz:
Do you have benchmarks to prove that your changes measurably improve performance? Because these are not in fact inefficiencies at all, and the changes in this PR will likely slow things down.

BIG objects or objects that have nontrivial copy constructors should certainly be passed by const reference instead of copying the objects onto the stack, because it's less expensive to push an 8 byte pointer onto the stack rather than, say, the 200 byte structure it points to. But this general rule is not true for small objects that are themselves not much larger than a pointer, and that have trivial constructors.

Passing by value places the object you want right on the stack, in commonly-used memory and sharing a cache line with the other parameters or local variables of the called function. Also, because that makes it local to the function, the compiler knows it can't be modified by other threads or by other functions your function might call (in both cases, because if you don't explicitly pass its address, no other function or thread can possibly know where it is in the local storage of your function's stack frame), and can therefore do a variety of optimizations based on this knowledge. (I can explain examples if you have no idea what I'm talking about.)

On the other hand, anything you pass by reference costs an extra pointer indirection every time you access it, plus the penalty from the fact that the memory being accessed could be anywhere (and in particular not likely near the other things you will be using frequently, so may be a cache miss to access), and also the compiler has potentially no idea what other functions (or concurrent threads) might be modifying the memory your reference points to, and therefore will have to generate much more defensive and inefficient code in order to guarantee correctness.

TypeDesc is 64 bits, by design. Pointers are also 64 bits (on any CPU you would use if you care about performance). So there is no possible savings from passing a pointer to a TypeDesc instead of the TypeDesc itself. In fact, it's surely worse, because you then will have to pay for the extra pointer indirection every time you access it, and all the other problems outlined above.

string_view is a pointer and a length (total: 16 bytes). It, too, should be passed by value. It's true that passing by reference (i.e. a pointer, 8 bytes) saves *slightly* (pushing an extra 8 bytes on the stack), but you will come out behind because every single time you use it, you will incur an extra memory read because you've inserted an extra level of indirection.

And, sigh, also ustring! A ustring is, underneath, just a pointer to the characters. It's 64 bits. There is, I guarantee, no gain from passing a pointer to the pointer to the characters, instead of just the pointer to the characters, and that's exactly what you have done by changing `ustring` to `const ustring&`. You don't pass a reference to a ustring for exactly the same reason you wouldn't pass a reference to a `char*` -- the pointer itself is already the smallest and least expensive thing you can pass to convey where the characters are, there is nothing to gain by passing a pointer to the pointer.

I'm sorry about this, I feel bad that you made such extensive changes only to get it shot down. It really would have been better if you'd proposed on the mail list what you intended to do, or submitted a "preliminary" PR where you do the change to JUST ONE PLACE and say "hey, do you like this? Because if it's the efficiency gain I suspect, I can do it all over the code base." And with that tiny investment, we could have set you straight before you did any more of the work.

DennisOSRM:
Thanks for the elaborate explanation. This is helpful.

Issue #2198, opened 2019-03-22
CMAKE error 'Could NOT find ZLIB (missing: ZLIB_INCLUDE_DIR)'


Nuclearfossil:
Hey all,

I've taken a look through the closed (and open) issues, but I don't see any reference to the issue I'm currently having ...

### Problem
I'm trying to get a build of OIIO running under windows, Visual Studio 2015/2017.

This _could_ be me just being a noob with CMAKE, but I've been getting the following:

```
CMake Error at C:/Program Files/CMake/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find ZLIB (missing: ZLIB_INCLUDE_DIR)
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.12/Modules/FindZLIB.cmake:112 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/cmake/externalpackages.cmake:56 (find_package)
  CMakeLists.txt:137 (include)
```

I've grabbed the pre-built dependencies, but I ran into an issue where OIIO is looking for a newer version of the TIFF library (3.9 minimum and a much earlier version is in the pre-build package).  However, the real issue I'm having is getting cmake to find zlib. 

I've already added a `ZLIB_LIBRARY` that points to the path to the zlib 'lib' folder but every time I define a `ZLIB_INCLUDE_DIR`, it gets overwritten each time I configure the cmake project (I'm using the CMAKE gui).

I had a similar problem with the TIFF library reference, but was able to get past it by updating the `TIFF_INCLUDE_DIR` and `TIFF_LIBRARY` vars.

Has anyone had any luck building a recent version for windows? I'm seeing this both in the master and RB-2.0 branches

for reference:
![image](https://user-images.githubusercontent.com/398853/54844099-456a2b00-4c93-11e9-8c89-2d2929a129cc.png)

ttddee:
I had a similar problem where CMake would not accept ZLIB_LIBRARY and ZLIB_INCLUDE_DIR.

Did you try setting ZLIB_ROOT?

lgritz:
[...]

PR #2216, opened 2019-04-19
Fix ZLIB references

lgritz:
* Correct use of ZLIB_INCLUDE_DIRS (not ZLIB_INCLUDE_DIR!).
* Only add include dirs for modules that need it.
* Also, change from obsolete PNG_INCLUDE_DIR to correct PNG_INCLUDE_DIRS.


lgritz:
Fixes #2198

PR #2220, opened 2019-04-27
Fix memory leak in AttrDelegate::as_string

lgritz:


src/include/OpenImageIO/attrdelegate.h
```diff
@@ -185,7 +186,14 @@ template<class C> class AttrDelegate {
         if (t == TypeString) {  // Attrib is a string? Return it.
             s = get<std::string>();
         } else if (t != TypeUnknown) {  // Non-string attrib? Convert.
-            char* buffer = new char[t.size()];
+            const int localsize = 64;   // Small types copy to stack, avoid new
+            char localbuffer[localsize];
+            char* buffer = localbuffer;
+            std::unique_ptr<char> allocbuffer;
```
fpsunflower: I think you want `std::unique_ptr<char[]>` (otherwise you mix `new char[]` with a plain `delete`)

lgritz: Quite right, I will fix.

lgritz: Quite right, I will fix.

PR #2229, opened 2019-04-30
Sysutil: add utilities to get stack traces

lgritz:
Set up signal handlers in all the command line utils so that if they
crash, they will print stack traces that will help track down CI or
user-side failure cases.

This relies on Boost >= 1.65 for its stacktrace library. When built
against older versions of boost, stacktrace() will return an empty
string and the stacktrace-printing signal handles will have no effect.
C'est la vie.

### DIFF
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -51,6 +51,10 @@
 #include <OpenImageIO/platform.h>
 #include <OpenImageIO/string_view.h>
 
+// Allow client software to know if this version has Sysutil::stacktrace().
+#define OIIO_HAS_STACKTRACE 1
+
+
 
 OIIO_NAMESPACE_BEGIN
 
@@ -115,6 +119,19 @@ physical_concurrency();
 /// Get the maximum number of open file handles allowed on this system.
 OIIO_API size_t
 max_open_files();
+
+/// Return a string containing a readable stack trace from the point where
+/// it was called. Return an empty string if not supported on this platform.
+OIIO_API std::string
+stacktrace();
+
+/// Turn on automatic stacktrace dump to the named file if the program
+/// crashes. Return true if this is properly set up, false if it is not
+/// possible on this platform. The name may be "stdout" or "stderr" to
+/// merely print the trace to stdout or stderr, respectively. If the name
+/// is "", it will disable the auto-stacktrace printing.
+OIIO_API bool
+setup_crash_stacktrace(string_view filename);
 
 /// Try to figure out how many columns wide the terminal window is. May not
 /// be correct on all systems, will default to 80 if it can't figure it out.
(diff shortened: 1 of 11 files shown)


src/libutil/sysutil.cpp
```diff
@@ -604,4 +616,60 @@ aligned_free(void* ptr)
 }
 
 
+
+std::string
+Sysutil::stacktrace()
+{
+#if BOOST_VERSION >= 106500
+    std::stringstream out;
+    out << boost::stacktrace::stacktrace();
```
fpsunflower: Seems like a shame to require memory allocation (however small it may be) inside the signal handler.

In what case do you think the user would want to access this string as memory instead of relying on the installed handler?

src/include/OpenImageIO/sysutil.h
```diff
@@ -116,6 +120,19 @@ physical_concurrency();
 OIIO_API size_t
 max_open_files();
 
+/// Return a string containing a readable stack trace from the point where
+/// it was called. Return an empty string if not supported on this platform.
+OIIO_API std::string
```
fpsunflower: I'm not totally clear how the user can gain access to this string.

If the program crashes, the default handler is invoked and the program exits immediately afterwards, no?

src/libutil/sysutil.cpp
lgritz: Well, I don't consider it ok to install signal handlers as a library, unless explicitly told to do so. So at most I can guarantee that it's set for the standalone binaries (oiiotool, etc).

I can imagine all sorts of nonfatal situations in which it would be useful to print some kind of diagnostic or debugging message that can include the full stack trace, at times when no signal-triggering fault has occurred.

As far as the allocation... maybe? So what? It does seem to work, and it's a situation where we're about to crash anyway.

src/include/OpenImageIO/sysutil.h
lgritz: You don't have to wait for a crashing fault to ask for a stack trace!

fpsunflower: Oh I see, you are thinking of regular debugging cases too.

fpsunflower:
LGTM!

src/include/OpenImageIO/sysutil.h
lgritz: The main reason I did this is to be able to increase the set of circumstances in which users can give me clues about where something crashed, even if they are not debugger-savvy or I can't reproduce the crash on my end.

For the command line tools, installing a signal handler is enough.

But it was so little extra work to full expose it as a string, I thought it would be helpful to open up the possibilities of a new kind of rich diagnostic error for certain nonfatal error situations (but that are unexpected enough that we'd really want to know more about the circumstances that led to them). You could build an interesting kind of non-aborting assertion out of it, for example, that tells you much more about exactly what the call sequence was, without having to catch it live in a debugger at the moment of the abort().

Issue #2230, opened 2019-05-02
Reading PNG or JPG fails on Windows


ttddee:
### Problem

Reading PNG or JPG images fails on Windows, starting with version OIIO 2.0.7.

**Expected behavior:**  
```
shared_ptr<ImageBuf> inputImage (new ImageBuf());
inputImage->init_spec("test.jpg", 0, 0);
bool ok = inputImage->read(0, 0, TypeDesc::FLOAT);
```

`ok` should be true

**Actual behavior:**  

`ok` is false and `inputImage->geterror()` returns the following for JPG images:

`Invalid image file "test.jpg": JPEG error: Not a JPEG file: starts with 0xff 0xe1`

And for PNG images:

```
libpng error: Not a PNG file
Invalid image file "azul.png": PNG read error: Not a PNG file
```

Other image formats work fine.

I have gone through the commits and this change introduces the problem:

https://github.com/OpenImageIO/oiio/commit/7677d498b599295fa8277d050ef994efbd297b55

Anything before works fine.

### Steps to Reproduce

1. Build OIIO master on Windows x64
2. Run the code above on any PNG or JPG

### Versions

* OIIO branch/version: master
* OS: Windows x64 
* C++ compiler: MSVC2017 toolset v141

lgritz:
Can you directly email me a small JPEG file that fails?

lgritz:
Some follow-up questions for you:

* You say that OIIO 2.0.7 is where it first broke. But the commit 7677d49, is on master, not 2.0. The equivalent change was backported to RB-2.0, but it was in what became release 2.0.8, not 2.0.7. 

So I'm confused, I don't see how the changes in that commit could affect release 2.0.7.

Neither the PNG nor JPEG modules were altered between 2.0.6 and 2.0.7.

That commit you cite, though -- that does look suspicious! It touches both PNG and JPEG in similar ways that are related to how the files are opened and read from.

Can you possibly verify if things still seem broken on (a) TODAY's top of RB-2.0, which is imminently about to become 2.0.8, and (b) TODAY's "master" branch?

lgritz:
Can you send me a specific JPG file that exhibits this problem?

ttddee:
Yes I will verify both 2.0 and master and get back to you.

Also sent you a test image.

lgritz:
OK, I finally have a hypothesis about what's going on.

Can you try an experiment for me?

[...]

Issue #2218, opened 2019-04-24
Most PNG images can no longer be read


gregcotten:
### Problem

Opening a PNG used to work, now it doesn't.

### Steps to Reproduce

1. Run ```iinfo``` on [this png](https://www.dropbox.com/s/fhsywqydg4dstm5/Rectangle.png?dl=0) or [this png](https://www.dropbox.com/s/0zxe5gasvw173u4/4x4Checker.png?dl=0)
2. Get error: ```iinfo ERROR: Could not create PNG read structure```

### Versions

* OIIO branch/version: 2.0.7
* libpng version: 1.6.37
* OS: macOS 10.13 or 10.14
* C++ compiler: Apple LLVM version 10.0.0 (clang-1000.11.45.5)

gregcotten:
Strangely is not due to #2167 as I can get the PNG to read in 2.0.6

lgritz:
Awesome, I will take a look right away.

gregcotten:
Looks like the PNG test images on the olio-images repo don't load either.

lgritz:
I can't reproduce with current master, current RB-2.0 head, or at the Release-2.0.6 tag, with my MacOS 10.14.4, LLVM 10.0.1, libpng 1.6.37 (from homebrew), either release or debug build. :-(

Can you post me the full (VERBOSE=1) log that shows the cmake setup and any additional options you used? For example,

    make nuke
    make VERBOSE=1 > build.log

Also, does it fail for you on the current master as well?

lgritz:
I'm sorry to ask you to do this, but If it fails for you at the 2.0.6 tag, can you try 2.0.5, etc., and narrow it down to a release? Or, if you really are a saint, bisect and narrow it down to a commit?

When you say "iinfo" has that error -- with any particular flags? You are just running `iinfo Rectangle.png` to get that message?

gregcotten:
Oh yea ```iinfo Rectangle.png```. This is with OIIO 2.0.7. 2.0.6 has no such issue. I could bisect, but I'll try master first and see if it's fixed.

lgritz:
The weird thing is that between 2.0.6 and 2.0.7, there were no direct changes to the png code.

The only thing that looks suspicious is commit 5579d9, which is PR 2185 -- that changed some of the logic of deciding which format reader to invoke.

Could you test at the point of commit 5579d9 versus the commit immediately prior to it?

gregcotten:
[...]

PR #2231, opened 2019-05-03
Fix IOProxy::pread on Windows that botched PNG & JPEG read

lgritz:
We've been plagued by recently-introduced bugs with PNG and JPEG
reading that stemmed from the first IOProxy-related patch.  The
problem was our use of the Windows ReadFile call, which seems to not
preserve the file pointer as we'd assumed (the MSDN docs are ambiguous
about this). I also began to doubt its thread safety, so in the end I
just replace the code with a simpler approach with a mutex and calls
to tell/seek/read.


lgritz:
Fixes #2230

lgritz:
Fixes #2218

PR #2237, opened 2019-05-11
Straighten out construction of ImageBuf from spec -- zero or not?

lgritz:
* Docs, header, and implementation were not all consistent about
  whether the ImageBuf construct-from-ImageSpec (or the `reset()` call
  of the same flavor) zeroes out the newly allocated buffer, or if the
  pixel values are left uninitialized.

* Change the call signature to take an optional `zero` parameter -- which
  defaults to true, meaning that the pixels are zeroed out when allocated
  (this is the foolproof approach, at the cost of writing all the memory
  of the buffer). If zero==false, the newly allocated pixels will remain
  uninitialized (this is for power users who want to maximize performance
  when they know that they will soon overwrite all the vaues in the buffer
  and want to avoid any unnecessary writes).

* Touch up the docs and header comments to be consistent and explain the
  behavior correctly.

### DIFF
--- a/src/doc/imagebuf.tex
+++ b/src/doc/imagebuf.tex
@@ -60,12 +60,15 @@ constructed \ImageBuf using the default constructor.
 
 \subsection*{Constructing and initializing a writeable \ImageBuf}
 
-\apiitem{{\ce ImageBuf} (const ImageSpec \&spec, bool zero = true) \\
-{\ce ImageBuf} (string_view name, const ImageSpec \&spec, bool zero = true)}
+\apiitem{{\ce ImageBuf} (const ImageSpec \&spec, \\
+\big\big InitializePixels zero = InitializePixels::Yes) \\
+{\ce ImageBuf} (string_view name, const ImageSpec \&spec,
+\big\big InitializePixels zero = InitializePixels::Yes)}
 Constructs a writeable \ImageBuf with the given specification (including
 resolution, data type, metadata, etc.).
-The pixels will be initialized to black/empty if `zero` is true, otherwise the
-pixel values will remain uninitialized.
+The {\cf zero} parameter controls whether the pixel values are filled with
+black/empty, or are left uninitialized after being allocated (if set to
+{\cf InitializePixels::No}).
 Optionally, you may name the \ImageBuf.
 \apiend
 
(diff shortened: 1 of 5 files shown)


fpsunflower:
Doesn't a bunch of internal code create `ImageBuf` objects? Do you think all of them should gain this new zero-initialization?

lgritz:
We can always pass zero=false in specific cases that we're trying to optimize. But clearing automatically is definitely the safe/foolproof way to go.

This change was prompted by a (user) error, provoked by the inconsistencies in the docs and header. And the bug was intermittent -- many times (in fact always for simple uses) the uninitialized buffer was filled with zeroes initially. But when used in a loop, so that the memory was recycled between a few different IB's as they were created and destroyed repeatedly, strange things were happening due to the un-zeroed buffers.

Since it was news to me that the docs weren't consistent, and I had to look at the code to figure out whether it zeroed or not (and wonder which way I really intended), it kind of made me wonder if *my* code using that constructor had the bugs, perhaps unnoticed. I feel better knowing I'm zeroing it out by default, and I can change performance-critical sections later to disable the initialization.

fpsunflower:
Makes sense. LGTM!

lgritz:
Revised... replaced the 'bool' parameter with an enum that is self-documenting. Do people like this idiom, or not? I can keep it or back out to what I had before. Just trying this on for size.

fpsunflower:
Its definitely more explicit which is nice when reading user code.

Is it possible to propagate that same idea to the python bindings?

lgritz:
In C++, you would worry about a function call that looks like

    x = myfunction (true, false, true);

what the heck is that? Perhaps the function signature looked like

    int myfunction (bool colorful, bool big, bool female);

So maybe you make it less error prone by changing the bools to enums, and end up with

    enum class Colorful { No, Yes };
    enum class Size { Normal, Big };
    enum class Gender { Male, Female };
    int myfunction (Colorful c, Size s, Gender g);

    x = myfunction (Colorful::Yes, Size::Normal, Gender::Female);

which is self-documenting at the call point and makes it very hard to accidentally switch the order of arguments and end up with subtle bugs.

But in python, you could write

    x = myfunction (colorful=True, big=True, female=True)

and that serves of the same purpose. In fact, if you name all the arguments like that, you can swap the argument order around as well.

It seems to me that adding a bunch of enums is largely unnecessary in Python, since you can name arguments when you pass it. But I don't know. Is there a consensus on what the most "pythonic" way to do it is?

fpsunflower:
That makes sense. I don't write enough Python to know what the preferred idiom would be, but it seems reasonable to keep the boolean there (as long as you don't forsee any new enums added to the set in the near future).

lgritz:
I checked with Jon Ware and he said that bools with named params is the most pythonic way to to it in a case like this.

If we end up adding a bunch more choices, we can revisit. But in this case, it's just "zero it out" or not.

fpsunflower:
SGTM. LGTM!

PR #2242, opened 2019-05-24
oiiotool --ch did not honor `allsubimages=1` modifier

lgritz:
It worked with the global -a, but not the customary `:allsubimages=1`
modifier to that individual command. Either is supposed to make it
apply the channel shuffling to all subimages of a multi-subimage file
(for example, a multipart exr file).


aconty:
LGTM!

PR #2247, opened 2019-05-31
Experiment: Try the combination of Doxygen + Sphinx + Breathe + rst

lgritz:
Years ago we made Doxygen docs. I hated it, I think they look ugly and
are a poor user experience trying to make sense of them. So at some
point we stopped the charade and let it rust away. Though it lives on
a bit by the fact that I'd already established the habit of using
`///` for documentary comments.

This leaves us with the problem of how to make the header files
self-documenting, while also maintaining the beautiful LaTeX PDF book
that I consider the primary documentation of OIIO. I've been keeping
up as best as I can, but this leads to redundancy of the writing (at
best), and divergence (at worst, which is to say, always).

I'm not a huge fan of "restructured text" (rst), but the documentation
that Sphinx makes is fairly pleasing, it can generate html web pages
and also a PDF (via LaTeX) from the same source material, and I'm
really pleased with other projects that I see use sphinx-generated
documentation hosted on readthedocs.org.

Turns out that there is a Sphinx extension called "Breathe" that will
take Doxygen-generated XML and weave it into a Sphinx doc pretty
nicely.  In short, this allows me to write the book I want (in rst,
though, rather than LaTeX) and simply include by reference the
documentation for specific API calls (anything from a single function
to an entire class), which are taken directly from the header
declarations via Doxygen+Breathe. And in addition to the PDF book, it
simultaneously makes nice looking HTML version of the book, which can
be hosted on readthedocs.

For this PR, I'm not replacing any of the old stuff yet, but I set up
the skeleton of this scheme and convert just a single chapter of the
book.  I'm liking what I see, so I want to commit what I have so
far. Over time, I will convert the other chapters and eventually will
switch permanently to this scheme, assuming I don't discover a
horrible unintended consequence of this.

In the process of getting this first chapter done -- which happens to
be the one about helper classes like ImageSpec, TypeDesc, ustring,
etc. -- I did some very heavy editing of the header comments. No code
changes, but a LOT of beefed up comments, brining them into sync with
what was in the full latex documentation.

### DIFF
--- a/src/doc/conf.py
+++ b/src/doc/conf.py
@@ -1,0 +1,77 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# http://www.sphinx-doc.org/en/master/config
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+# The master toctree document.
+master_doc = 'index'
+
+project = 'OpenImageIO'
+copyright = 'Contributors to OpenImageIO'
+author = 'Larry Gritz'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+version = '2.1'
+# The full version, including alpha/beta/rc tags.
+release = '2.1.1'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+#'sphinx.ext.pngmath', 
+#'sphinx.ext.todo', 
+'breathe' ]
+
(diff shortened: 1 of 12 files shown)


src/doc/conf.py
```diff
@@ -0,0 +1,77 @@
...
+
+project = 'OpenImageIO'
+copyright = 'Contributors to OpenImageIO'
+author = 'Larry Gritz'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+version = '2.1'
```
fpsunflower: WBN to grab this from the header somehow so that it stays in sync ..

src/include/OpenImageIO/deepdata.h
```diff
@@ -60,17 +62,21 @@ class OIIO_API DeepData {
...
-    /// Clear the vectors and reset size to 0.
+    /// Reset the `DeepData` to be equivalent to its empty initial state.
     void clear();
-    /// Deallocate all space in the vectors
+    /// In addition to performing the tasks of `clear()`, also ensure that
+    /// all allocated memory has been truly freed.
     void free();
 
-    /// Initialize size and allocate nsamples, pointers.
+    /// Initialize the `DeepData` with git specified number of pixels,
```
fpsunflower: typo: `git` ?

lgritz: um... must have thought the focus was on a terminal window...

src/doc/conf.py
lgritz: Yes, I expect a bunch of tweaking of the tooling I'm building around this.

PR #2252, opened 2019-06-04
Fix broken ability to specify compression of multipart exr files

### LINKED ISSUE #2251: oiiotool ignores --compression option when outputting multipart EXRs
When using EXR output and saving a multipart EXR, (either via building one via --siappend in oiiotool, or using as input an EXR that already is multipart) the --compression option is ignored and all EXR parts that are saved in any output file use default (zip) compression.

This occurs under current release (2.0.8) as well as the master branch (2.1.1dev). (running debian stretch, built for C++11)  bug is also present in iconvert.

We found this when trying to optimize some of our EXRs for read performance on another system.  When taking 'sample.exr' that has say R,G,B channels, plus a number of other named channels in the same view/part, let's call them SKY, WATER, FG.  We wanted to move the named channels into the second part of a new, multi-part file. (sample.exr has no compression in this case) 

`oiiotool -i sample.exr --dup --ch "R,G,B" --swap --ch "SKY, WATER, FG" --siappend --compression "none" -o output.exr`

Will yield a file with zip compression (but, curiously a "compressionName" header field that says "none" despite the zip compression.)

lgritz:
Typo! Was grabbing from this->spec before it was assigned, should have
looked in the parameter spec (to OpenEXROutput::spec_to_header).

Fixes #2251

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -674,7 +674,7 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
 
     string_view comp;
     int qual;
-    std::tie(comp, qual) = m_spec.decode_compression_metadata("zip");
+    std::tie(comp, qual) = spec.decode_compression_metadata("zip");
     // It seems that zips is the only compression that can reliably work
     // on deep files (but allow "none" as well)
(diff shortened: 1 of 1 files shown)


fpsunflower:
LGTM

PR #2260, opened 2019-06-12
Revise raw processing

### LINKED ISSUE #2256: Gamma issue when loading RAW images
### Problem

There is an inconsistency between the metatada `oiio:Colorspace` and the gamma applied on the image when loading a RAW image.

The problem has evolved over time with the different versions of OIIO:
- in previous versions, there was both a color shift and a gamma problem
- in version 1.8.12, everything was working fine and both problems were solved. The `oiio:Colorspace` returned when loading a raw file was `Linear` (and `raw:Colorspace` was `sRGB`).
- in 2.0.x, there is a gamma problem again. The `oiio:Colorspace` returned when loading a raw file is `sRGB` (no more `raw:Colorspace` metadata).

[Here](https://github.com/OpenImageIO/oiio/blob/master/src/raw.imageio/rawinput.cpp#L409) we set:
```
    // Set the gamma curve to Linear
    m_processor->imgdata.params.gamm[0] = 1.0;
    m_processor->imgdata.params.gamm[1] = 1.0;
```
And [few lines later](https://github.com/OpenImageIO/oiio/blob/master/src/raw.imageio/rawinput.cpp#L468) we set the colorspace:
```
m_processor->imgdata.params.output_color = c;
```
And the `oiio:Colorspace` is set to the same value (`c` is the index for libraw and `cs` is the corresponding colorspace in oiio as a string):
```
        m_spec.attribute("oiio:ColorSpace", cs);
```
But AFAIU, the `output_color` only controls the primaries and not the gamma curve that we imposed to linear in all cases by setting the parameter `params.gamm` to 1.
So I think that the `oiio:Colorspace` should be set to linear in all cases.

For instance in the release 1.8.16, the code was:
```
    // Set the gamma curve to Linear
    m_spec.attribute("oiio:ColorSpace","Linear"); <<<< Set to linear in all cases
    m_processor->imgdata.params.gamm[0] = 1.0;
    m_processor->imgdata.params.gamm[1] = 1.0;
```

### Versions

* OIIO branch/version: 1.8.12 was working fine, 2.0.X doesn't
* OS: tested on both Linux and Windows
* C++ compiler: gcc/VS

lgritz:
We were not using certain libraw parameters correctly.

New default behavior is to decode to sRGB (both primaries and
response) -- because for a naive user that just reads a raw file and
slaps it up for display, this behaves like a jpeg.

But configuration hints "raw:ColorSpace" can request various other
things, like "ACES" (linear ACES), "Adobe" (Adobe gamut with 1.8
gamma), and so on. Asking for "linear" gives you sRGB primaries
with linear response.

Fixes #2256

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -368,7 +368,6 @@ RawInput::open_raw(bool unpack, const std::string& name,
                    const ImageSpec& config)
 {
     // std::cout << "open_raw " << name << " unpack=" << unpack << "\n";
-    ASSERT(!m_processor);
     m_processor.reset(new LibRaw);
 
     // Temp spec for exif parser callback to dump into
@@ -406,10 +405,6 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // Output 16 bit images
     m_processor->imgdata.params.output_bps = 16;
 
-    // Set the gamma curve to Linear
-    m_processor->imgdata.params.gamm[0] = 1.0;
-    m_processor->imgdata.params.gamm[1] = 1.0;
-
     // Disable exposure correction (unless config "raw:auto_bright" == 1)
     m_processor->imgdata.params.no_auto_bright
         = !config.get_int_attribute("raw:auto_bright", 0);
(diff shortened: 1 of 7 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -443,47 +438,63 @@ RawInput::open_raw(bool unpack, const std::string& name,
...
 #endif
-                error("raw:ColorSpace set to unknown value");
-            return false;
-        }
-        m_spec.attribute("oiio:ColorSpace", cs);
-    } else {
-        // By default we use sRGB primaries for simplicity
+    } else if (Strutil::iequals(cs, "linear")) {
+        // Request "sRGB" primaries, linear reponse
         m_processor->imgdata.params.output_color = 1;
```
fabiencastan:
```
} else if (Strutil::iequals(cs, "linear")) {
        // Request "sRGB" primaries, linear response // << typo on response
        m_processor->imgdata.params.output_color = 1;
        m_processor->imgdata.params.gamm[0]      = 1.0; //<<< seems to be missing, isn't it?
        m_processor->imgdata.params.gamm[1]      = 1.0;
```

src/raw.imageio/rawinput.cpp
```diff
@@ -443,47 +438,63 @@ RawInput::open_raw(bool unpack, const std::string& name,
...
         = config.get_int_attribute("raw:use_camera_matrix", 1);
 
-
-    // Check to see if the user has explicitly set the output colorspace primaries.
-    // The default is to ask it to convert to sRGB space.
+    // Check to see if the user has explicitly requested output colorspace
+    // primaries via a configuration hinnt "raw:ColorSpace". The default if
+    // there is no such hint is convert to sRGB, so that if somebody just
+    // naively reads a raw image and slaps it into a framebuffer for
+    // display, it'lll work just like a jpeg. More sophisticated users might
```
fabiencastan: typo: `it will`

fabiencastan:
I like the idea of just setting "raw:Colorspace" and let oiio setup libraw correctly!
That's great!

src/raw.imageio/rawinput.cpp
lgritz: good catch

src/raw.imageio/rawinput.cpp
lgritz: correct, thanks for the catch

lgritz:
Small revision posted. Will merge when I get all the CI tests passing.

PR #2262, opened 2019-06-15
Spruce up ASSERT macros

lgritz:
* Use pretty function printing to show what func it failed in.

* Changing wording to more closely match built in assert().

* Make DASSERT simply define to built in assert(), which has the
 advantage compiling without warnings and running correctly with Cuda!
 The ASSERT (works in debug mode as well) and [D]ASSERT_MSG macros
 still are not Cuda friendly because they specifically mention stderr.

### DIFF
--- a/src/include/OpenImageIO/dassert.h
+++ b/src/include/OpenImageIO/dassert.h
@@ -31,6 +31,7 @@
 
 #pragma once
 
+#include <cassert>
 #include <cstdio>
 #include <cstdlib>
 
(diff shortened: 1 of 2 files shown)


src/include/OpenImageIO/dassert.h
```diff
@@ -31,6 +31,7 @@
 
 #pragma once
 
+#include <assert.h>
```
fpsunflower: `<cassert>` ?

src/include/OpenImageIO/dassert.h
```diff
@@ -92,15 +96,10 @@
...
-/// (optimized, shipping) build.
-#ifndef NDEBUG
-#    define DASSERT(x) ASSERT(x)
-#else
-/* DASSERT does nothing when not debugging; sizeof trick prevents warnings */
-#    define DASSERT(x) ((void)sizeof(x)) /*NOLINT*/
-#endif
+/// DASSERT(condition) is just an alias for the usual assert() macro.
+/// It does nothing when in a non-DEBUG (optimized, shipping) build.
+#define DASSERT(x) assert(x)
```
fpsunflower: The `sizeof` trick was useful in preventing warnings about unused variables in some cases. Most assert implementations (sadly) don't implement this.

Maybe it's worth keeping the `#ifdef NDEBUG` case ?

lgritz: So the main point of all this was to want to find a way that a function using DASSERT can be decorated with `__host__ __device__` without generating warnings.

My ASSERT when used in Cuda code will generate warnings upon compile (because of its referencing of stderr), and not work properly when executed. Whereas if you use assert(), there is a Cuda implementation that not only works but prints some useful info on the threads, etc. It uses some trickery of substituting a different assert.h or some other override of a true assert() function that is Cuda-aware.

So what I was trying to do is make DASSERT just use the built in assert(), and only for my unconditional ASSERT have our own implementation.

I guess I can go back to the original and have an even more complicated `#if` that falls back to assert() when in NVCC only, or something like that?

lgritz: OH, I see your point about the `#if`. Let me give it another try...

src/include/OpenImageIO/dassert.h
lgritz: I was worried that would thwart whatever cuda is doing, but this does seem ok so I will fix.

lgritz:
Updated to address @fpsunflower's comments.

lgritz:
better now?

fpsunflower:
LGTM!

PR #2272, opened 2019-06-27
Protect simd.h against stray #define of True or False

lgritz:
This proposed patch just #undef's it. 

Optionally, I could also re-define it at the end, like:

    #ifdef _X11_XLIB_H_ /* regrettable */
        #define True 1
        #define False 1
    #endif 

I'm not sure if that's clever, or "too clever" and will cause more trouble.


fpsunflower:
Redefining seems like trouble for anyone who would want to use `vbool*::True` down the line.

LGTM!

PR #2277, opened 2019-07-01
Add thread_info to texture related calls that only take a filename

fpsunflower:
This fixes a slight assymetry between the calls that accepts a handle and those which only accept a filename.

The real motivation for this is to avoid incorrectly counting the number of threads (which is inferred by the number of created thread_info structures) in applications that mix both styles of calls (such as OSL based renderers that use OIIO's texture system).

Along for the ride is a cleanup of the private texture header to add `override` annotations to prevent typos when implementing virtual calls.

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -619,6 +619,10 @@ public:
     ///             should be stored. It is the caller's responsibility to
     ///             ensure that `data` points to a large enough storage area
     ///             to accommodate the `datatype` requested.
+    /// @param  thread_info
+    ///             Pointer to a pre-allocated thread_info structure. If this
+    ///             is NULL, one will be created automatically with a modest
+    ///             performance cost.
     ///
     /// @returns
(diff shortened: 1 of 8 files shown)


lgritz:
So currently, there are two sets of calls:

* Naive and simple: pass a filename only.
* Advanced: pass a handle and a thread info.

You are noticing a missing scenario, which is when the app has a thread info, but still a filename (not a handle). But I'm not sure that we need to add new methods or change the call signatures to handle this case.

I want to suggest that a simpler solution is fully on the app side, by calling the existing advanced version, and simply generating a handle from the filename with

    get_texture_handle(filename, thread_info)

That's all that happens inside the implementation, so there is no performance penalty for doing so. That would avoid needing to change any of the call signatures.

fpsunflower:
Hmmm that would indeed save a lot of work.

I think I got thrown off by the extra logic in `ImageCacheImpl::get_image_info` for when the handle is `null`, but on second look I don't think that code can be triggered. A valid (but broken) handle is created for missing files.

Let me revisit all of this then.

PR #2278, opened 2019-07-02
Prevent a doubled-up error message in get_image_info

fpsunflower:
Prevent error messages from getting doubled-up in get_image_info if a handle is created during the call

Minor optimization to how errors are cleared in exists queries


lgritz:
LGTM

fpsunflower:
Merged.

lgritz:
I think you merged a version that fails the clang-format test.

lgritz:
Don't worry, I'll fix it.

PR #2280, opened 2019-07-03
Update fmt implementation to fix windows warnings

### LINKED ISSUE #2276: [BUILD] embedded fmtlib link errors on Windows
### Problem
OIIO version: 2.0.8
The embedded version of fmtlib contains definitions of isnan/_isnan functions that clash with Windows' ones. It does not prevent OIIO from building, but can cause link errors when using OIIO as a library (probably due to Windows' <math.h> also being included at some point):
```
error C2084: function 'fmt::v5::internal::dummy_int fmt::v5::internal::_isnan(...)' already has a body (compiling source file ...) ...\fmt\format.h(278): note: see previous definition of '_isnan'.
```
This was reported on fmtlib (https://github.com/fmtlib/fmt/issues/1104) and a fix has been merged in master since. No official release including this fix has been made yet, but using the last master snapshot (https://github.com/fmtlib/fmt/tree/2711cb1672543cc67b5f9a2b1ec900d0b333f93c ATM) solves the problem.
Would it be possible to upgrade the embedded version of fmtlib despite the lack of an official release?

lgritz:
Resolves #2276 

I don't expect anybody to "review" the code itself, it's just syncing with the latest fmtlib.

But perhaps @yann-lty can try building from this PR and see if it avoids the error? (If it works, I will also backport to RB-2.0 if you need it there.)

### DIFF
--- a/src/include/OpenImageIO/fmt/core.h
+++ b/src/include/OpenImageIO/fmt/core.h
@@ -16,190 +16,206 @@
 #include <type_traits>
 
 // The fmt library version in the form major * 10000 + minor * 100 + patch.
-#define FMT_VERSION 50202
+#define FMT_VERSION 50301
 
 #ifdef __has_feature
-# define FMT_HAS_FEATURE(x) __has_feature(x)
+#  define FMT_HAS_FEATURE(x) __has_feature(x)
 #else
-# define FMT_HAS_FEATURE(x) 0
+#  define FMT_HAS_FEATURE(x) 0
 #endif
 
 #if defined(__has_include) && !defined(__INTELLISENSE__) && \
     !(defined(__INTEL_COMPILER) && __INTEL_COMPILER < 1600)
-# define FMT_HAS_INCLUDE(x) __has_include(x)
+#  define FMT_HAS_INCLUDE(x) __has_include(x)
 #else
-# define FMT_HAS_INCLUDE(x) 0
+#  define FMT_HAS_INCLUDE(x) 0
 #endif
 
 #ifdef __has_cpp_attribute
-# define FMT_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
+#  define FMT_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
 #else
-# define FMT_HAS_CPP_ATTRIBUTE(x) 0
+#  define FMT_HAS_CPP_ATTRIBUTE(x) 0
 #endif
 
 #if defined(__GNUC__) && !defined(__clang__)
-# define FMT_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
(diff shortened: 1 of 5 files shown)


fpsunflower:
Maybe we should just accept that `fmt` is a dependency and handle it externally?

Like `pybind11`, its not a particularly onerous dependency since its header only. I'm guessing we could handle it in a similar way.

yann-lty:
Turns out that the link errors were actually due to another external library that was redefining "isnan", more details here: https://github.com/OpenImageIO/oiio/issues/2276#issuecomment-508040022. 
This PR builds and runs fine on my side, but since updating fmt is not required after all - and not in the "bugfix" scope anymore -, I don't know if it's still wanted to make changes about this right now ?

lgritz:
I think since I've already done the (small amount of) work, and it passes the tests, and presumably has some improvements to fmt's internals, I may as well merge it. I mean, I would/should have done it sooner or later anyway. 

But I certainly won't backport it to 2.0 if there's not a pressing need. (And I also realized now, maybe if there was a need for a fix in 2.0, it would have been better to just `#undef isnan` in strutil.h and not update any code.)

lgritz:
@fpsunflower : I'm still trying to decide what to do long term. pybind11 is strictly a build-time dependency, but with fmt it needs to have the headers available when people (it's included by strutil.h) so we need to copy some of the headers into wherever include/OpenImageIO dists to. So I'm a little leery of just relying on a system install at build time. 

Maybe it's still better to have it external like we do pybind11 (even if we copy parts to `dist`), rather than actually check the files into our repo. I'm not sure I want to tackle that today, need to think about it. Do you have any current thinking/preferences about "downloading" (like we do with pybind11) versus using git submodules?

fpsunflower:
Oh right, I forgot we have to expose that API as part of ours. That's definitely unfortunate (but unavoidable since its all template driven).

I'm not sure I want to get into recursive submodules at the moment. Its few enough headers that copying is probably the best solution for now. I'm just anticipating that at some point `fmtlib` will be popular enough that conflicts may arise for people that use their own copy elsewhere? It doesn't look like they allow for customizing the namespace wrapping.

I do miss the simplicity of `printf`  ...

lgritz:
We could change the namespace just in our copy, I guess.

lgritz:
There is an inner namespace with a version number, so presumably it should not be a problem for multiple libraries to be using fmtlib, assuming that they are properly updating the version if ABI compatibility breaks.

fpsunflower:
I think its more complicated than that since the headers are using include guards instead of `#pragma once`, so just including the headers from two different copies in the wrong order could mix up which ones get used.

To be safe, I think both things should be patched (header guards and namespace).

lgritz:
I think I'm going to merge what I have (nothing wrong with updating but keeping the status quo on inclusion strategy), and punt for now on changes to fmt. Will revisit another day if it pops up as an actual problem.

PR #2281, opened 2019-07-04
Reorder includes for Python.h and POSIX_C_SOURCE macro

marsupial:
## Description
Python.h will set POSIX_C_SOURCE it was built with and py_oiio.h includes that first

### DIFF
--- a/src/python/py_colorconfig.cpp
+++ b/src/python/py_colorconfig.cpp
@@ -28,10 +28,9 @@
   (This is the Modified BSD License)
 */
 
-#include <utility>
-
 #include "py_oiio.h"
 #include <OpenImageIO/color.h>
+#include <utility>
 
 namespace PyOpenImageIO {
 
--- a/src/python/py_oiio.h
+++ b/src/python/py_oiio.h
@@ -12,6 +12,9 @@
 #endif
 
 // Must include Python.h first to avoid certain warnings
+#ifdef _POSIX_C_SOURCE
+#  error "You must include Python.h (and therefore py_oiio.h) BEFORE anything that defines _POSIX_C_SOURCE"
+#endif 
 #include <Python.h>
 
 #include <memory>


src/python/py_colorconfig.cpp
```diff
@@ -28,10 +28,9 @@
   (This is the Modified BSD License)
 */
 
-#include <utility>
-
 #include "py_oiio.h"
 #include <OpenImageIO/color.h>
+#include <utility>
```
lgritz: I'm not sure I understand what the issue here is.

Python.h, which is included by py_oiio.h, sets that macro and... what does that have to do with <utility>?

lgritz: Ugh, damn GitHub formatting...

what does that have to do with `<utility>`?

marsupial: Python.h sets **POSIX_C_SOURCE**, but it's unconditional and the same as what Python was compiled with.

`<utility>` can wind up included std-c headers that also set `POSIX_C_SOURCE`, but those tend to be conditional (adapting to incoming values).

So if Python.h is first, it's `POSIX_C_SOURCE` is used later in `<utlity>` and no redefinition occurs.
When `<utility>` is first, it chooses `POSIX_C_SOURCE` however it wants, which may conflict with `Python.h`

lgritz: This seems very brittle with respect to future additions of #include which most people think doesn't matter the order.

Perhaps a cleaner solution is to put this in py_oiio.h:

    #ifdef POSIX_C_SOURCE
        #undef POSIX_C_SOURCE
    #endif

lgritz: But with a comment:

    // It's bad news if POSIX_C_SOURCE is already defined with Python.h is included
    #ifdef POSIX_C_SOURCE
        #undef POSIX_C_SOURCE
    #endif 
    #include <Python.h>

marsupial: The Python guys seem pretty adamant their POSIX_C_SOURCE should rule for any modules compiled with it...whether that's valid or not...

My only thought is that doing the undef is making a new issue (albeit rare) where one might need to define `POSIX_C_SOURCE` rather keeping the it squarely in Python's court?

That said, don't have that strong of an opinion, I just though the comment below was pretty much because of this issue and the <utility> just wound up getting misplaced: https://github.com/OpenImageIO/oiio/blob/master/src/python/py_oiio.h#L14

lgritz: How about, in py_oiio.h :  

    #ifdef POSIX_C_SOURCE
    #  error "You must include Python.h (and therefore py_oiio.h) BEFORE anything that defines POSIX_C_SOURCE"
    #endif 

That will ensure that if somebody working on OIIO python bindings years from now changes the order of include files, they'll see this message and know what to do.

marsupial: LGTM (and certainly explicit!)
Should I add it on top ?

lgritz: Yeah, just put in py_oiio.h, right before including python.h, as an extra safety measure.

lgritz: In addition to your original reordering, I mean.

lgritz:
LGTM

PR #2283, opened 2019-07-08
Fix embarrassing ustring::compare(string_view) bug

lgritz:
The logic was wrong in cases where the string_view was longer than the
ustring, but they has the same character sequence up to the length of
the ustring.

Add some unit tests to ferret out this case.

To do: fully flesh out unit tests for every last bit of functionality
of ustring and string_view.


fpsunflower:
Why not just use:
```c
return string_view(m_chars, size()).compare(str);
```

which already has the correct logic?

lgritz:
I think you're right, and I can eliminate a few other cases as well. Updated.

fpsunflower:
LGTM!

PR #2286, opened 2019-07-10
iv: Add -rawcolor option to skip auto-convert to RGB

lgritz:
For better inspection of raw values of CMYK or other non-RGB color
spaces that would ordinarily be automatically converted.


fpsunflower:
LGTM

PR #2288, opened 2019-07-10
oiiotool -stats not communicating the input configuration hints

lgritz:
Like, let's say it's a CMYK file, and you want to know the stats about the original CMYK values, not the stats about the emulated RGB translation...


fpsunflower:
LGTM

PR #2289, opened 2019-07-10
IBA::computePixelStats was not properly controlling the number of threads

lgritz:
By not passing the nthreads param to the parallel_for_chunked loop,
it was always running fully threaded instead of being controlled by
the nthreads passed to it.


fpsunflower:
LGTM

PR #2292, opened 2019-07-13
Fix not allocating enough scratch space for cmyk tiff reads

lgritz:
Related to the fixes I did a couple commits ago. There was one use of a temp buffer that was allocating based on the 3-channel size for translation to RGB, but it was being used to read the raw 4 channnel data into it, overrunning the bounds and screwing up the heap.


lgritz:
Added a unit test

lgritz:
Original reporter says this works

PR #2300, opened 2019-07-23
FindOpenEXR.cmake improvements (ported from master)

lgritz:


lgritz:
This is a patch against RB-2.0, just porting some fixes already in master.

PR #2301, opened 2019-07-24
ImageBufAlgo: Fix implicit float/double casts to half incorrectly flagging values as non-finite

nrusch:
## Description

I recently discovered that, when running `maketx` on an HDR that included the sun, the extremely high channel values (> 145,000) were being rolled off to just over 65,000. Eventually, I was able to track the cause of the "rolloff" down to the use of `--fixnan box3`, and from there, I found that float/double values were being incorrectly flagged as non-finite due to an implicit cast to half.

Basically, because a cast to half is possible, the local `isfinite` [in `imagebufalgo_pixelmath.cpp`](https://github.com/OpenImageIO/oiio/blob/master/src/libOpenImageIO/imagebufalgo_pixelmath.cpp#L1003) is being preferred over `std::isfinite` for float and double values.

The fix uses a templated shim around `std::isfinite`, with an explicit specialization for `half`.

## Tests

I can add an additional `oiiotool-fixnan` test if you think it's valuable.

## Checklist:

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_pixelmath.cpp
+++ b/src/libOpenImageIO/imagebufalgo_pixelmath.cpp
@@ -998,9 +998,17 @@ ImageBufAlgo::color_map(const ImageBuf& src, int srcchannel,
 
 namespace {
 
-// Make sure isfinite is defined for 'half'
+template<typename T>
 inline bool
-isfinite(half h)
+isfinite_(T v)
+{
+    return std::isfinite(v);
+}
+
+// Make sure isfinite is defined for 'half'
+template<>
+inline bool
+isfinite_<half>(half h)
 {
     return h.isFinite();
 }
@@ -1022,7 +1030,7 @@ fixNonFinite_(ImageBuf& dst, ImageBufAlgo::NonFiniteFixMode mode,
                  ++pixel) {
                 for (int c = roi.chbegin; c < roi.chend; ++c) {
                     T value = pixel[c];
-                    if (!isfinite(value)) {
+                    if (!isfinite_(value)) {
                         ++count;
                         break;  // only count one per pixel
                     }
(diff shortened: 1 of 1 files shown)


ThiagoIze:
I just came across this same problem, but I went with a different fix because I didn't realize it was already fixed in master. While the current code looks like it should work just fine, I wonder if maybe we should use the below code instead since it results in what I think is cleaner code.  Notice it's just one line instead of a bunch of new lines and a new function, not to mention it prevents possible bugs that could later arise if someone uses an isfinite() by accident (which compiles) instead of isfinite_()? Here's what it looks like if applied to the old broken code:

```diff
diff --git a/src/libOpenImageIO/imagebufalgo_pixelmath.cpp b/src/libOpenImageIO/imagebufalgo_pixelmath.cpp
index ba245d7df..b8a019af3 100644
--- a/src/libOpenImageIO/imagebufalgo_pixelmath.cpp
+++ b/src/libOpenImageIO/imagebufalgo_pixelmath.cpp
@@ -1000,16 +1000,18 @@ namespace {
 
 // Make sure isfinite is defined for 'half'
 inline bool
 isfinite(half h)
 {
     return h.isFinite();
 }
 
+// Make sure isfinite is defined for 'float' and other standard types
+using std::isfinite;
 
 template<typename T>
 bool
 fixNonFinite_(ImageBuf& dst, ImageBufAlgo::NonFiniteFixMode mode,
               int* pixelsFixed, ROI roi, int nthreads)
 {
```

Side note: the half(float) constructor should be made explicit so that these kinds of bugs can't happen. I verified that at least in my slightly older OIIO this compiles just fine after using the above fix and has a compile time error without any fixes. Of course that's an OpenEXR issue, so I'll officially raise this issue there, but I figured I'd mention it here too.

lgritz:
Oh, that's a more elegant solution, Thiago. Should I go ahead and make that change to fmath.h?

lgritz:
N/M, I misunderstood. Not a change to fmath.h, is it?

Still, is it better for me to add this to imagebufalgo_pixelmath.cpp and get rid of the slightly more awkward extra template?

ThiagoIze:
Yes, I think it would be, provided my patch makes sense to others.

nrusch:
If it works reliably, then I'm all for cleaner code.

PR #2304, opened 2019-08-01
Some fixes to allow build with mingw

Alexpux:
Small fixes allowing to build OpenImageIO with mingw


lgritz:
LGTM, thanks! Merging, and I will also backport to RB-2.0.

PR #2314, opened 2019-08-17
Small improvements concurrent map

fpsunflower:
The hash table underlying `unordered_map_concurrent` has support for queries using a precomputed hash value.

Since we need to hash once to figure out which bin to use, we may as well re-use that value for the lookup when possible, which avoids an extra hash operation while we are holding the lock.

Benchmarks show a 1-2% speedup in some cases (as part of a whole renderer).

Also swapped the `FingerprintMap` to  `tsl::robin_map` while I was in there, there's no good reason to use `std:unordered_map`.

### DIFF
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -264,11 +264,12 @@ public:
     /// lock.
     iterator find(const KEY& key, bool do_lock = true)
     {
-        size_t b = whichbin(key);
+        size_t hash = m_hash(key);
+        size_t b = whichbin(hash);
         Bin& bin(m_bins[b]);
         if (do_lock)
             bin.lock();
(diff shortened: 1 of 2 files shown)


lgritz:
It's failing the clang-format check. Can you enable the setting that lets me push changes to your submission? I'm happy to pull, clang-format it, and then update.

lgritz:
N/M, I see I already can do so. Hang on, I'll fix it.

lgritz:
I fixed the formatting for you. I will merge when the new CI tests come back. Good fixes.

lgritz:
I have some ideas for what to try next, BTW.

lgritz:
It's failing CI on OSX, but it doesn't have anything to do with you. Looks like just this morning homebrew updated libheif (which we use to read HEIF/HEIC files) and it's changing some of the image hashes for that format. I will deal with that separately. Merging your code...

PR #2315, opened 2019-08-17
Performance tuning for ustring table

fpsunflower:
The main change here is to tune the default size and shard count for the ustring table. The average workstation has changed quite a bit since I last measured the performance of this code, and a simple tweak to these constants gives a nice speed boost on the `ustring_test` stress test.

On a machine with 100+ hardware threads, the new code stays efficient much longer. Previously the wedge test maxed out around 14% efficiency, and its close to 100% under the new code.

Measuring the cost of making 2M ustrings with a single thread is around 0.5s, while with all threads running at once (one will make progress while the others wait) it takes 4.0s (vs. 24.2s before this change).

On a more modest machine with 32 hardware threads, the tuned code is also faster (27% efficiency for the wedge test becomes 90%). Cost of making 2M strings with all threads at once goes from 6.0s down to 2.5s.

There are lots of different possible usage patterns and obviously `ustring_test` is not perfect, but I think this will be a win overall.

I did a bit of housecleaning on the file at the same time, switching to C++11 idioms in a few places.

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -34,20 +34,18 @@ typedef null_lock<null_mutex> ustring_read_lock_t;
 typedef null_lock<null_mutex> ustring_write_lock_t;
 #endif
 
+// #define USTRING_TRACK_NUM_LOOKUPS
 
-// NOTE: BASE_CAPACITY must be a power of 2
-template<unsigned BASE_CAPACITY = 1 << 20, unsigned POOL_SIZE = 4 << 20>
-struct TableRepMap {
+template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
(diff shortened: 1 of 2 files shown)


fpsunflower:
I think this one is also just down to Mac libheif issues and clang-format.

@lgritz is there a simple way to run the `clang-format` step locally?

lgritz:
Yes!

    make clang-format

fpsunflower:
`make clang-format` doesn't seem to be working on my (spi) machine, its finding the system `clang-format` binary instead of the one that lives alongside the compiler. It also is not able to pick up the `.clang-format` project file and therefore trying to reformat the whole project with LLVM style :(

fpsunflower:
Running `clang-format` manually seems to have done the right thing ... I think this one should pass.

lgritz:
Weird. I usually do it on my laptop, so maybe I need to try it on the work Linux and see what's going wrong.

PR #2316, opened 2019-08-19
Improve the hash function for the tile ID

fpsunflower:
Use all bits of the hash to pick the bin in unordered concurrent map instead of just the low 32 bits we used before (because the previous hash function was 32 bits only).

Minor cleanups to the TileID struct

In preliminary benchmarks, this simple change has a dramatic impact on some scenes (from 10% to close to 2x faster), while having a relatively no measurable impact on scene that are not stressing the texture system.

Something I've yet to play with: instead of mixing in the filename hash, just use the file pointer itself (which should be unique as well). This would avoid the two caches misses it takes to lookup the filename's hash at the expense of making the hash function non-deterministic.

### DIFF
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -416,19 +416,16 @@ private:
     size_t whichbin(size_t hash)
     {
         constexpr int LOG2_BINS = log2(BINS);
-        constexpr int BIN_SHIFT = 32 - LOG2_BINS;
+        constexpr int BIN_SHIFT = 8 * sizeof(size_t) - LOG2_BINS;
 
         static_assert(1 << LOG2_BINS == BINS,
                       "Number of bins must be a power of two");
-        static_assert(~uint32_t(0) >> BIN_SHIFT == (BINS - 1), "Hash overflow");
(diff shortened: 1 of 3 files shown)


lgritz:
Some pattern must have had really uneven map bucket occupancy with the old hash?

fpsunflower:
Thats what I'm thinking. The pattern of `a * x + b * y + c * z` is risky because even though you are mixing with a "good" hash function afterwards, you can still get "aliases" where different values of `x` and `y` wind up with the same hash.

Also just adding the filename hash is also a bit risky since its a "constant" shift, so the same tiles of different maps (imagine a color/spec/roughness map of all the same resolution) can wind up colliding frequently if you are unlucky.

The next set of measurements is to check `m_file->filename().hash()` against `(size_t) m_file` to see if reducing the cache misses there is worth it.

lgritz:
Yes, I was thinking that, too. Or store the filename hash in the tileid?

fpsunflower:
For the sake of documenting what I tried. Using `(size_t) m_file` only helped in a few cases, and sometimes made things measurably much worse.

In repeated runs, the swings in performance also seemed much wider, which I attribute at least partially to the collisions changing from run to run.

It might be worth revisiting after we further improve the concurrent hash map.

PR #2327, opened 2019-09-03
Fix potential deadlock in TIFF I/O: minor flaw with threadpool method

lgritz:
This has *very* occasionally popped up for a while. I've only seen it
on OSX for some reason, too hard to reproduce but once every several
days.  But a totally unrelated set of changes caused me to see it
happen quite regularly and it's icky.

Long story short, our thread pool seemed to have two different ways of
finding out whether the current thread is a member of the pool, one of
them only used from the TIFF reader/writer, and they didn't match!
Specifically, the correct one tried to know if the calling thread was
doing work for the pool, either being a member of the pool itself or a
nonmember that was stealing work because the pool was too full to add
to. The other one merely said whether it was a fully fledged pool
worker thread. Turns out this difference could get you into trouble,
because you don't want to spawn new tasks from a work stealer, either.

It took me hours of debugging to isolate the trouble, it's too hard to
summarize the specific set of coincidences necessary for this
circumstance to lead to a deadlock.

I also simplified some of the TIFF code just a hair, was a redundant
check for read_raw_strips.

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -715,7 +715,10 @@ public:
     /// Return true if the calling thread is part of the thread pool. This
     /// can be used to limit a pool thread from inadvisedly adding its own
     /// subtasks to clog up the pool.
-    bool this_thread_is_in_pool() const;
+    /// DEPRECATED(2.1) -- use is_worker() instead.
+    bool this_thread_is_in_pool() const {
+        return is_worker();
+    }
 
     /// Register a thread (not already in the thread pool itself) as working
     /// on tasks in the pool. This is used to avoid recursion.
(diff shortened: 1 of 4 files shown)


src/include/OpenImageIO/thread.h
```diff
@@ -725,8 +728,10 @@ class OIIO_API thread_pool {
     void deregister_worker(std::thread::id id);
     /// Is the thread in the pool or currently engaged in taking tasks from
     /// the pool?
+    bool is_worker(std::thread::id id) const;
```
fpsunflower: Do you need both the `const` and non-`const` versions?

fpsunflower: NVM, I guess this is purely for backwards compatibility?

lgritz: Yes, because it fixes a potential deadlock, I want to immediately backport it to the release branch, where I can't break link compatibility. I'm marking it as deprecated, so I'll remove the non-const next time we have a fully compatibility-breaking release.

src/include/OpenImageIO/thread.h
```diff
@@ -715,7 +715,10 @@ class OIIO_API thread_pool {
     /// Return true if the calling thread is part of the thread pool. This
     /// can be used to limit a pool thread from inadvisedly adding its own
     /// subtasks to clog up the pool.
-    bool this_thread_is_in_pool() const;
+    /// DEPRECATED(2.1) -- use is_worker() instead.
```
lgritz: Oh, I just realized, I need to make this non-inline in order to not break compatibility when I backport to the release branch.

fpsunflower:
LGTM!

PR #2330, opened 2019-09-04
Update dpxinput.cpp to fix issue with poor DPX read performance

talonrenaud:
Patch gracefully provided by Michael McCartney (mmccartney@stereodllc.com)
I don't think this patch is the right way to fix the problem. It seems to re-implements the read_image() method and successfully fix OIIO 2.x DPX read speed issue for us but this is most likely a band-aid for a wider problem.
This should be at the very least a good starting point in terms of figuring out what caused the problem in the first place.
OIO 1.8.5 didn't have this read speed problem, I started running into this problem since OIIO 2.0.x.

## Description

DPX file are extremely slow to read (mostly when reading in parallel in separate threads)
See more detailed problem description here:
http://lists.openimageio.org/pipermail/oiio-dev-openimageio.org/2019-September/001707.html

## Tests

If there's a DPX test suite then timing the read speed with a speed range might be a good thing to detect this sort of issue in the future. (testing hardware may vary though so this could be tricky)

## Checklist:

Most check points below are not applicable since pull request is provided as a guideline for further troubleshooting. Code isn't requested to be pulled as in into project's code.

- [ ] I have read the [contribution guidelines](../CONTRIBUTING.md).

### DIFF
--- a/src/dpx.imageio/dpxinput.cpp
+++ b/src/dpx.imageio/dpxinput.cpp
@@ -34,6 +34,15 @@ public:
     virtual bool read_native_scanline(int subimage, int miplevel, int y, int z,
                                       void* data) override;
 
+    /*
+        Custom read_image for slurping the whole thing
+    */
+    virtual bool read_image(int subimage, int miplevel, int chbegin, int chend,
+                            TypeDesc format, void* data, stride_t xstride,
+                            stride_t ystride, stride_t zstride,
+                            ProgressCallback progress_callback,
+                            void* progress_callback_data) override;
+
 private:
     int m_subimage;
     InStream* m_stream = nullptr;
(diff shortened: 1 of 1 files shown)


lgritz:
Thanks, that gives me a good starting point. I don't really see why this should be a dramatic speedup. Before committing, I want to investigate a little and try to understand when and why performance degraded for the individual scanline reads.

mccartnm:
Saw this _after_ creating a PR with the same code. 👍 Nailed it.

@menssimia is the best person to answer any questions as he identified the performance killer in the first place - I just implemented the "fix". I will ping him and see if there's any extra info he has on it.

lgritz:
I propose #2333 as the correct fix.

PR #2333, opened 2019-09-04
Fix DPX performance regression

lgritz:
Quite some time ago (early in OIIO 2.0 development, after it was
separated from 1.8), we changed the ImageInput::read_native functions
to be thread-safe and stateless. This meant that a seek_subimage would
be called inside the read (which now had a lock) whereas it was
previously assumed that the user had already called it.

All good, as long as your seek_subimage checks whether it's seeking to the
subimage it's already on, in which case no work need be done.

DPX failed to do this, so every scanline read involved a costly header
examination.  Somehow nobody noticed this until now, but it's fairly
icky. It can slow down DPX reads by 2x, and this is greatly exacerbated
in a heavily multithreaded environment.

I found and fixed a similar error for psd files.

While I was at it, I also added a read_native_scanlines so it can read
multiple scanlines (the building block of read_image working efficiently)
with less work, though in my timings it's only barely worth it.

### DIFF
--- a/src/dpx.imageio/dpxinput.cpp
+++ b/src/dpx.imageio/dpxinput.cpp
@@ -33,6 +33,8 @@ public:
     virtual bool seek_subimage(int subimage, int miplevel) override;
     virtual bool read_native_scanline(int subimage, int miplevel, int y, int z,
                                       void* data) override;
+    virtual bool read_native_scanlines(int subimage, int miplevel, int ybegin,
+                                       int yend, int z, void* data) override;
 
 private:
     int m_subimage;
@@ -40,19 +42,18 @@ private:
     dpx::Reader m_dpx;
     std::vector<unsigned char> m_userBuf;
     bool m_rawcolor;
-    unsigned char* m_dataPtr = nullptr;
+    std::vector<unsigned char> m_decodebuf;  // temporary decode buffer
 
     /// Reset everything to initial state
     ///
     void init()
     {
+        m_subimage = -1;
         if (m_stream) {
             m_stream->Close();
             delete m_stream;
             m_stream = nullptr;
         }
-        delete[] m_dataPtr;
-        m_dataPtr = nullptr;
         m_userBuf.clear();
         m_rawcolor = false;
     }
@@ -163,6 +164,8 @@ DPXInput::seek_subimage(int subimage, int miplevel)
 {
     if (miplevel != 0)
         return false;
+    if (subimage == m_subimage)
+        return true;
     if (subimage < 0 || subimage >= m_dpx.header.ImageElementCount())
         return false;
 
@@ -537,17 +540,6 @@ DPXInput::seek_subimage(int subimage, int miplevel)
     if (m_spec.nchannels == 1)
         m_rawcolor = true;
 
-    dpx::Block block(0, 0, m_dpx.header.Width() - 1, m_dpx.header.Height() - 1);
-    int bufsize = dpx::QueryRGBBufferSize(m_dpx.header, subimage, block);
-    if (bufsize == 0 && !m_rawcolor) {
-        error("Unable to deliver RGB data from source data");
-        return false;
-    } else if (!m_rawcolor && bufsize > 0)
-        m_dataPtr = new unsigned char[bufsize];
-    else
-        // no need to allocate another buffer
-        m_dataPtr = NULL;
-
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/dpx.imageio/dpxinput.cpp
```diff
@@ -163,6 +164,8 @@ DPXInput::seek_subimage(int subimage, int miplevel)
 {
     if (miplevel != 0)
         return false;
+    if (subimage == m_subimage)
```
lgritz: THESE TWO LINES ARE THE **REAL** FIX.

lgritz:
@mccartnm @talonrenaud Can you pull this patch and try on your end?

If you don't want to worry about patches, you could even just type in by hand the two lines that I pointed out in the comment (the addition to seek_subimage) which account for almost all of the performance. The rest is just cleanup and squeezing out a bit more.

talonrenaud:
I re-compiled with the updated dpxinput.cpp and psdinput.cpp (patching 2.0.10 release) and it works beautifully.
My small test app now shows DPX load time 4 times faster than 1.8.5. This is fantastic !
Thank you both for your help this is greatly appreciated, I will sleep better tonight.
I always assume I’m doing something wrong when compiling everything so I’m glad it wasn’t me this time around.
Thanks again !

mccartnm:
A much better of a solution than our giant-hammer-band-aid. Thanks! Interesting that we didn't find that when doing our dig. I remember looking at the subimage stuff but coming up empty. Glad it was such a minor tweak! Compiled as well - performance looks really good 👍

menssimia:
The thing that we noticed was that OIIO's DPX read performance over CIFS (e.g. reading from an Isilon on a Windows box) was quite poor, since OIIO reads the DPX file backward, and reverse reads over CIFS have terrible performance characteristics (CIFS seems to strongly favour optimistic forward reads). This was in stark contrast with reading from local storage.

This may also be true of other formats that define their scanline storage in reverse order from OIIO, but it was especially noticeable with DPX images.

We'll do another set of tests once we have this update on the production network and report back.

talonrenaud:
> The thing that we noticed was that OIIO's DPX read performance over CIFS (e.g. reading from an Isilon on a Windows box) was quite poor, since OIIO reads the DPX file backward, and reverse reads over CIFS have terrible performance characteristics (CIFS seems to strongly favour optimistic forward reads). This was in stark contrast with reading from local storage.
> 
> This may also be true of other formats that define their scanline storage in reverse order from OIIO, but it was especially noticeable with DPX images.
> 
> We'll do another set of tests once we have this update on the production network and report back.

I thought menssimia's comment was interesting and we forwarded the comment to EMC and Qumulo. See their response below.

**Qumulo:**
> We have a pretty aggressive prefetcher by default but it switches strategies when it 'regrets' lots of prefetches (it reads data into cache that is never requested by a client). This is actually a thing that we could build for: recognition that we're reading files backwards. I will have a chat with Tommy about this. It's also possible that they're not just reading backwards, but they're reading in very small chunks. Both of those things would be suboptimal with our current strategy, but we can improve the experience of reading backwards.
> 
> OOC, why read these files backwards with OIIO?
> 
> Mike Bott
> 
> Principal Systems Engineer

**EMC (Isilon)**

> Can the application be flipped to read the DPX top to bottom? It’s on github so I can assume its modifiable on a fork, maybe not easy but could be done.
> 
> Brett Dellandre
> Advisory Systems Engineer
> Dell EMC | Unstructured Data Solutions

lgritz:
In what sense are DPX files "read backwards?" Can you point to the lines of code where you think this is happening?

Here is where read_native_scanlines sets up the dpx::Block describing the region, and there is no reversal: https://github.com/OpenImageIO/oiio/blob/master/src/dpx.imageio/dpxinput.cpp#L574

Here is where dpx::Reader::ReadBlock does the actual file seek to the start of the region, for the non-RLE case, just a straightforward seek of the y scanline multiplied by scanline size:
https://github.com/OpenImageIO/oiio/blob/master/src/dpx.imageio/libdpx/Reader.cpp#L145

For the RLE case, it actually reads the whole image it at once (in order) and then just copies out the individual scanlines as they are needed. This code starts here: https://github.com/OpenImageIO/oiio/blob/master/src/dpx.imageio/libdpx/RunLengthEncoding.cpp#L130

Am I missing something?

talonrenaud:
I was just reacting to @menssimia comment, he would be the one who could answer this question.
If the scanlines are read in order in the code then there shouldn't be anything to worry about. The patch you created works great for us so this is considered resolved on our end. Thanks.

PR #2336, opened 2019-09-06
Debian -- certain architectures seem to need -latomic with gcc9

lgritz:


lgritz:
Withdrawing. I don't think this is the best solution. A followup may come later.

PR #2341, opened 2019-09-11
Address gcc8 warnings for SHA1.cpp

### LINKED ISSUE #2337: BUILD: -Wall causing fits for GCC 8.3.0-6
`error: ‘char* strncpy(char*, const char*, size_t)’ output may be truncated copying 15 bytes from a string of length 15 [-Werror=stringop-truncation]
   strncpy(szReport, szTemp, 15);`

Any way to get the default build to negotiate this one?

lgritz:
Fixes #2337


lgritz:
Abandoning in favor of #2342

PR #2344, opened 2019-09-19
New attrib "oiio:subimages" says number of subimages, if known

lgritz:
It's the number of frames for movie files, number of subimages for
other files.

Set and positive only if it can be known inexpensively from the file
header. If absent or zero, it means there was no easy way to determine
the number of frames or subimages from just reading the header (in that
case, you'll have to discover the subimages the old fashioned way,
repeatedly seeking to subimages until you fail).


talonrenaud:
This is great. I'll try to implement it this week end if I can compile the master branch.
Thank you.

PR #2345, opened 2019-09-20
Fixes that make multi-subimage and movie files faster

lgritz:
Several small, subtle, interacting issues were making it particularly
difficult for oiiotool to process frames within movie files (and,
less noticeably, other multi-subimage files). The fixes:

* The ffmpeg reader had no valid_file() method, which meant it fell
  back to the default of doing a full open() and looking for success,
  which is expensive in this case. Replace it with the cheap and naive
  test of just ensuring that the file exists and its name has one of
  the proper file extensions.

* ImageCache::read_untiled(), when reading an untiled image as one
  single tile, closed the file immediately thereafter, which seems
  like a prudent thing to do because once you have read the whole
  image into the cache, you may as well free the resources associated
  with the open file, since you don't need to read any more from
  it. Except... if it's a multi-subimage file, you may need to read
  other subimages.  Especially for a movie file with many frames, you
  don't want to do a full open/close cycle for every frame
  individually just because you are using an ImageCache.

* In oiiotool, ImageRec::read was applying a heuristic to directly and
  fully read "small" images, but fall back on the ImageCache to manage
  "large" images. But it was just looking at the each subimage
  individually. For a movie file, that means that it will force read
  all the frames up front, which is silly. So adjust the heuristic to
  multiply the frame size by number of frames as a better estimate for
  the memory cost of the up-front read. This results in most movie
  files of moderate size falling back to the cache, and thus deferring
  the read until the frames are actually needed (especially helpful if
  not all frames will eventually need to be read).

* Oh boy, oiiotool was always keeping handy a config spec to pass to
  the image open, whether it was needed or not. But also some logic in
  the ImageCache was causing certain kinds of "find file" operations
  (which when asked to find an already in-cache file should just
  return it) to invalidate that previously read file and close+reopen
  when a config spec was supplied (in case the configuration hints
  were meant to change the read behavior or interpretation of the
  file from how it was opened before). These combine badly. So we
  change oiiotool to only pass the config spec when opening via the
  cache if there was a user option to actually set a configuration
  hint, and therefore when that doesn't happen, many needless cache
  invalidations were avoided.

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -123,6 +123,7 @@ public:
     {
         return "FFmpeg movie";
     }
+    virtual bool valid_file(const std::string& name) const override;
     virtual bool open(const std::string& name, ImageSpec& spec) override;
     virtual bool close(void) override;
     virtual int current_subimage(void) const override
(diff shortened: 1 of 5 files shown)


lgritz:
With this patch, doing a certain frame-by-frame extraction and processing of a 100 frame 2048x1152 quicktime with oiiotool went from 24 minutes down to 8 seconds.

PR #2375, opened 2019-10-27
Fix compiler warning about doubly defined symbol

lgritz:


lgritz:
I see now that this was already done in master a while back and just not backported to 2.0.

PR #2378, opened 2019-10-30
Rig exr texture output to avoid dwa compression for 1-channel images

lgritz:
We've run into bugs with DWAA/DWAB compression of OpenEXR images in
the specific case of 1-channel images where tile size is either < 16
or not a power of 2.  There was already special logic in the exr
output module that notices this problem case and switches to zip
compression.

However, in the case of outputting MIPmaps, the top level may satisfy
the conditions under which DWA compression works fine (power of 2,
tile size >= 16). But in the course of outputting the MIP levels, we
will eventually reach a tile size < 16, and the OpenEXR format doesn't
allow us to select a different tile size per MIP level (it's just
stored once in the header for the whole file).

So add some logic to our MIPmap generation (maketx, etc), so that
single channel exr texture maps preemptively switch from dwaa/dwab to
something safe. This still only affects 1-channel images, multiple
channels don't appear to tickle the bug.

I couldn't help myself, along the way I did a little refactoring,
coalescing several "if the texture is an exr..." clauses together
under one conditional, and also modernizing an initialization loop
and fixing a typo.

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -611,6 +611,7 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
         return false;
     }
 
+    bool verbose = configspec.get_int_attribute("maketx:verbose") != 0;
     bool src_samples_border = false;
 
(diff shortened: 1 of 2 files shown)


rmv:
LGTM!

PR #2379, opened 2019-10-31
ImageCache per-thread file cache to deal with thread contention

lgritz:
It's a wash for small thread counts, but for highly threaded renders,
I'm seeing 1-2% speedups across real world scenes.

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -710,7 +710,7 @@ public:
     // Add a new filename/fileptr pair to our microcache
     void remember_filename(ustring n, ImageCacheFile* f)
     {
-        m_thread_files[n] = f;
+        m_thread_files.emplace(n, f);
     }
 
     // See if a filename has a fileptr in the microcache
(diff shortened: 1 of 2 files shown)


fpsunflower:
The code looks good, curious to see the benchmark results.

src/libtexture/imagecache_pvt.h
```diff
@@ -709,21 +708,16 @@ class ImageCachePerThreadInfo {
...
 
     // Add a new filename/fileptr pair to our microcache
-    void filename(ustring n, ImageCacheFile* f)
+    void remember_filename(ustring n, ImageCacheFile* f)
     {
-        last_filename[next_last_file] = n;
-        last_file[next_last_file]     = f;
-        ++next_last_file;
-        next_last_file %= nlastfile;
+        m_thread_files[n] = f;
```
fpsunflower: `m_thread_files.emplace(n, f)` may be a smidge more efficient

lgritz: Sure, I'll change to that.

lgritz:
For the curious, we saw about 6% average speedup in overall rendering time, across a suite of 28 production scenes, in a highly threaded situation (a machine with 52 cores + hyperthread).

src/libtexture/imagecache_pvt.h
lgritz: pushed an update with this fix

fpsunflower:
LGTM!

PR #2381, opened 2019-11-04
Bug fix in deep image compare when pixels had different # of samples

lgritz:
The IBA::compare for deep images A and B would loop over pixels in the
union of A's and B's ROIs (so far, so good), then for each pixel would
loop from 0 to A's number of samples in that pixel. Oops! That means
that if A had fewer samples than B (or if A had no samples but B had
samples) in that pixel, it would not notice that the images were
different.


aconty:
LGTM!

PR #2382, opened 2019-11-05
New function: default_thread_pool_shutdown()

lgritz:
This cleanly terminates all worker threads of the default thread pool
by resizing the pool to 0 threads... but only if the pool has been
created in the first place. Note that the alternate idiom

    default_thread_pool()->resize(0);

seems like it should accomplish the same thing, but in fact will start
by creating a pool and its threads, if it has not done so before, only
to immediately terminate them.


fpsunflower:
I'm not sure I follow the logic.

If the thread pool already exists, it will stop all the threads. But if the thread pool hasn't been created yet, it isn't going to prevent them from being created in the future.

So someone might call `default_thread_pool_shutdown()` only to see the default thread pool get created in the future.

Am I missing something?

lgritz:
Abandoning this, it seems not to be needed at this time.

lgritz:
This look ok to you, @LazyDodo ?

LazyDodo:
LGTM!

PR #2384, opened 2019-11-06
Cleaner PIMPL idiom for ImageBuf

lgritz:
This is a revision of #2350. I like this solution better.

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -24,14 +24,7 @@
 OIIO_NAMESPACE_BEGIN
 
 class ImageBuf;
-
-// Opaque type for the unique_ptr.
-class ImageBufImpl;
-class ImageBufImplBase {
-public:
-    virtual ~ImageBufImplBase() {}
-    void operator delete(void* todel) { ::operator delete(todel); }
-};
+class ImageBufImpl;  // Opaque type for the unique_ptr.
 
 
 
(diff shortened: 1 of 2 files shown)


src/include/OpenImageIO/imagebuf.h
```diff
@@ -1408,13 +1401,12 @@ class OIIO_API ImageBuf {
...
+    // PIMPL idiom
+    static void impl_deleter(ImageBufImpl*);
+    std::unique_ptr<ImageBufImpl, decltype(&impl_deleter)> m_impl;
 
-    ImageBufImpl* impl() { return (ImageBufImpl*)m_impl.get(); }
-    const ImageBufImpl* impl() const
-    {
-        return (const ImageBufImpl*)m_impl.get();
-    }
+    ImageBufImpl* impl() { return m_impl.get(); }
```
fpsunflower: Does this help anything? Anything that can access `impl()` can also access `m_impl` which is the same length.

I doesn't look like `impl()` is used too many times in the code either ...

fpsunflower:
I like this cleanup!

src/include/OpenImageIO/imagebuf.h
lgritz: Oh, you might be right. I'll take a look and eliminate it if it is only in a few places.

lgritz: It was a LOT of places. I just pushed an update, see if you are comfortable with that much change.

fpsunflower: I like it, I think this will be cleaner in the long run.

fpsunflower:
LGTM!

PR #2385, opened 2019-11-06
Reduce the number of stat calls when opening files

fpsunflower:
These are just a few simple changes that avoid an excessive number of syscalls to "stat" when opening a file for the first time (especially through the `ImageCache` which the texture system uses internally).

Noticed these while using `strace` on a render.

There are 2 `stat` calls left, both of which seem useful. One gets the size of the file for the stats, the other gets the last modification time for the invalidation logic.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -298,14 +298,12 @@ ImageCacheFile::ImageCacheFile(ImageCacheImpl& imagecache,
     // reflected by the fact that m_validspec is false.
 
     // Figure out if it's a UDIM-like virtual texture
-    if (!Filesystem::exists(m_filename.string())
-        && (m_filename.find("<UDIM>") != m_filename.npos
-            || m_filename.find("<U>") != m_filename.npos
-            || m_filename.find("<V>") != m_filename.npos
-            || m_filename.find("<u>") != m_filename.npos
-            || m_filename.find("<v>") != m_filename.npos)) {
-        m_is_udim = true;
-    }
+    m_is_udim = (m_filename.find("<UDIM>") != m_filename.npos
+                 || m_filename.find("<U>") != m_filename.npos
+                 || m_filename.find("<V>") != m_filename.npos
+                 || m_filename.find("<u>") != m_filename.npos
+                 || m_filename.find("<v>") != m_filename.npos)
+                && !Filesystem::exists(m_filename.string());
 }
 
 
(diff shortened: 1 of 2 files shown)


src/libtexture/imagecache.cpp
```diff
@@ -298,14 +298,11 @@ ImageCacheFile::ImageCacheFile(ImageCacheImpl& imagecache,
...
     // Figure out if it's a UDIM-like virtual texture
-    if (!Filesystem::exists(m_filename.string())
-        && (m_filename.find("<UDIM>") != m_filename.npos
-            || m_filename.find("<U>") != m_filename.npos
-            || m_filename.find("<V>") != m_filename.npos
-            || m_filename.find("<u>") != m_filename.npos
-            || m_filename.find("<v>") != m_filename.npos)) {
-        m_is_udim = true;
-    }
+    m_is_udim = m_filename.find("<UDIM>") != m_filename.npos
```
lgritz: Here you have lost the test for whether the file exists. That may be ok, but what it means is that you can never open a texture that is actually called `blah<u>.exr`. I don't know if anybody will ever want that, but it will not be possible. The original logic was that it only considered it a potential udim filename template if there was not an existing file with that literal name.

fpsunflower: If you think this is important, I would probably just restore the test, with the `&&` flipped so that it only checks for file existence if the filename really looks like a udim. String searching an in-memory string should be cheaper that asking the filesystem a question.

Given that `<` and `>` are illegal on most windows filesystems, the probability of someone truly wanting to name their textures with such tokens embedded seems small.

In fact, I would argue that even *if* somehow such a file existed on disk, the more likely case is that the file got "touched" by some other process not realizing it got a udim token string instead of a real filename.

fpsunflower: Another slightly more complicated proposal would be to add an option to the `ImageCache` to enable/disable the udim feature (off by default).

If you explicitly opt-in to udims, the proposed behavior here would be reasonable I think.

fpsunflower:
Fixed `clang-format` errors.

src/libtexture/imagecache.cpp
lgritz: I don't want to have to opt in, that will break the behavior of current renderers.

I'm ok with the assumption that nobody will use those names, and if it becomes a problem, we can add an "opt out of udim" option later.

Another compromise is to leave the existence test but move it to the end of the clauses,

    if ((m_filename.find("<UDIM>") != m_filename.npos
            || m_filename.find("<U>") != m_filename.npos
            || m_filename.find("<V>") != m_filename.npos
            || m_filename.find("<u>") != m_filename.npos
            || m_filename.find("<v>") != m_filename.npos)
        && !Filesystem::exists(m_filename.string())) {
        m_is_udim = true;
    }

That leaves behavior as it is for udims, but avoids the extra stat for the vast majority of files that are not udim. Remember that this expense is small, this is inside the ImageCacheFile ctr, so it's one extra stat per opened udim pattern in the scene.

fpsunflower: Right, that's what I was saying above.

So leave it as is for now?

lgritz: I'm on the fence.

I'm 100% enthusiastic for rearranging the "if" so that the exists() test only happens if it appears to be a udim name pattern.

I'm middle of the road with assuming that if it's a udim pattern, it's udim files, so don't bother to check if the file literally exists. It makes me a little nervous, but not enough to really object if you think that's best.

fpsunflower: I'll re-arrange, might as well preserve existing behavior.

fpsunflower:
Restored the existence check on udims, only if the filename include a udim token.

PR #2386, opened 2019-11-07
Fix bugs in Strutil::parse_string that would fail for escaped quotes

lgritz:
Two subtle bugs in Strutil::parse_string:

1. Toggling whether we are in an escape code based on the first
   character in the original string, rather than the next character to
   scan.  Big oops!

2. Not realizing that a second backslash in a row should *clear* the
   flag (i.e. `\\` does not leave you in an escaped state).

I beefed up the unit tests, looking at the example I added should
explain the edge case where this crops up -- a double-quoted string
with embedded \\\" within it, if you catch my drift. Just look at the
unit test, it explains better than I can do here. It's very hard to
talk and write (and reason correctly) about escaped strings with
embedded quotes!

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -771,13 +771,13 @@ Strutil::parse_string(string_view& str, string_view& val, bool eat,
     char lead_char    = p.front();
     bool quoted       = parse_char(p, '\"') || parse_char(p, '\'');
     const char *begin = p.begin(), *end = p.begin();
-    bool escaped = false;
+    bool escaped = false;  // was the prior character a backslash
     while (end != p.end()) {
         if (isspace(*end) && !quoted)
             break;  // not quoted and we hit whitespace: we're done
         if (quoted && *end == lead_char && !escaped)
-            break;  // closing quite -- we're done (beware embedded quote)
-        escaped = (p[0] == '\\');
+            break;  // closing quote -- we're done (beware embedded quote)
+        escaped = (end[0] == '\\') && (!escaped);
         ++end;
     }
     if (quoted && keep_quotes == KeepQuotes) {
--- a/src/libutil/strutil_test.cpp
+++ b/src/libutil/strutil_test.cpp
@@ -166,11 +166,20 @@ test_get_rest_arguments()
 
 
 void
+test_escape(string_view raw, string_view escaped)
+{
+    Strutil::printf ("escape '%s' <-> '%s'\n", raw, escaped);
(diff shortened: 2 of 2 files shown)


fpsunflower:
LGTM

MasterZap:
LGBTQ :)

/Z

src/libutil/strutil.cpp
```diff
@@ -771,13 +771,13 @@ Strutil::parse_string(string_view& str, string_view& val, bool eat,
     char lead_char    = p.front();
     bool quoted       = parse_char(p, '\"') || parse_char(p, '\'');
     const char *begin = p.begin(), *end = p.begin();
-    bool escaped = false;
+    bool escaped = false;  // was the prior character a backslash
     while (end != p.end()) {
         if (isspace(*end) && !quoted)
             break;  // not quoted and we hit whitespace: we're done
         if (quoted && *end == lead_char && !escaped)
-            break;  // closing quite -- we're done (beware embedded quote)
```
MasterZap: Not quite a quote :P

lgritz:
The hardest part of this was constructing the test -- it was really hard to reason about the right numbers of backslashes in different places to express in C++ the string that was causing the trouble.

MasterZap:
Can we have this in a 1.10.8 perchance?

MasterZap:
Oh... just realizing this is in OIIO code not OSL code... well...  is it in a release?

lgritz:
There were some escaping fixes in OIIO, and some in OSL. Both are needed, but I incorporated the important OIIO fix into OSL itself (inside an `ifdef OIIO is older than...`), in case you aren't able to upgrade both. :-)

Both are currently backported to their respective RB-?.? branches, but have not yet have official tagged releases (which I usually do on the first of the month). If it's really important for you to have a tagged release before then (like, if you have an imminent ship deadline and are only allowed to use tagged releases of dependencies), let me know and I can do it for you now.

MasterZap:
Omg no, I'm not in THAT much of a hurry./ZOn Nov 20, 2019 17:32, Larry Gritz <notifications@github.com> wrote:There were some escaping fixes in OIIO, and some in OSL. Both are needed, but I incorporated the important OIIO fix into OSL itself (inside an ifdef OIIO is older than...), in case you aren't able to upgrade both. :-)
Both are currently backported to their respective RB-?.? branches, but have not yet have official tagged releases (which I usually do on the first of the month). If it's really important for you to have a tagged release before then (like, if you have an imminent ship deadline and are only allowed to use tagged releases of dependencies), let me know and I can do it for you now.

PR #2387, opened 2019-11-07
[cmake] add config based install and usage

simogasp:
## Description

This PR replaces #2374 after the discussion with @lgritz and @kdt3rd.
It introduces the modern cmake way of exporting the project and using it as third party importing the targets in the host project.
It generates and installs 
* `OpenImageIOConfig.cmake` it's the config file. For the moment it's a basic template, but it must be updated whenever any dependency of OIIO will be switched to cmake config
* `OpenImageIOConfigVersion.cmake` Automatically generated to manage the version asked by the host project
* `OpenImageIOTargets.cmake` Automatically generated with all the targets with an `OpenImageIO` namespace.

In the host project OIIO can be used like this:

```cmake
find_package(OpenImageIO CONFIG)

add_executable(test test.cpp)
target_link_libraries(test PUBLIC OpenImageIO::OpenImageIO)
# note that target_include_directory() is not necessary
```

## Checklist:


lgritz:
I like! And thanks so much for the help. I really like how much you tried to make the code clear and commented, especially helpful for those of us new to the config generation.

But it's failing to build for most of the CI matrix tests. In particular, against CMake 3.12 (which is our advertised minimum), you aren't allowed to use install(TARGETS) on targets created outside the current directory, that wasn't added until CMake 3.13.

simogasp:
Hi sorry, i forgot about that. I'll get back with the solution. What I usually do is to have a global list variable containing the targets to export: they are appended at each declaration of each target.

simogasp:
@lgritz I used the existing oiio macro `install_targets()` to collect all the targets inside `OIIO_EXPORTED_TARGETS`, that way we have no problem with cmake < 3.13.
There are still failures in CircleCI but I think it's normal as the cmake does not meet the minimum required version.

Let me know what you think! :-)

lgritz:
Don't worry about the circleci failures -- something broke on their end, I don't need circleci enough to fix it, I'm eliminating it from the testing lineup now that we're leaning heavily on GH CI.

PR #2390, opened 2019-11-07
Guarantee that Release builds use -O3 for gcc/clang

lgritz:
We found at least some versions/distros of cmake that default to -O2,
but we think people always want -O3 for release builds.


lgritz:
I'm going to abandon this. Nobody had reported it as a problem, nobody commented on the PR, it only happens on some systems (maybe rare), and our benchmarks show negligible speed benefits in our tests.

PR #2391, opened 2019-11-08
strutil: specify the global namespace for strtof and strtod calls.

korli:
## Description

strutil: specify the global namespace for strtof and strtod calls. avoids infinite loop.
## Tests

Did run the testsuite on Haiku, without python tests. Some tests were previously crashing with an infinite loop in strtof.

## Checklist:

- [x] this is a small change to existing code.
- [x] I have ensured that the change is tested somewhere in the testsuite (no new test cases needed).
- [x] My code follows the prevailing code style of this project (I didn't used make clang-format though)


lgritz:
Yes, good catch, thanks!

PR #2400, opened 2019-11-13
upgrade to 3 channels when doing color conversions

rmv:
When color correcting single channel images, oiio just color corrects them as (y, 0, 0), it's better to use (y, y, y). We also fix a couple of spots in maketx with the same issue.


lgritz:
LGTM, thanks!

PR #2401, opened 2019-11-13
Bump robin map version to latest release

lgritz:
We were pegged to a fairly old commit. Also, modern versions have
changed their directory layout a bit (tsl/... -> include/tsl/..., so
compensate for that.


fpsunflower:
LGTM

hobbes1069:
Ok, good. I just built 0.6.2 (that's right, correct?) for Fedora.

PR #2404, opened 2019-11-14
BMP safety for big files and int32 overflow for offsets, and speedups

lgritz:
* Fix places that could overflow int32.

* Speed up read/write by about 40%, mostly by avoiding the pointless
  zeroing out of allocated memory in cerain std::vector temporary buffers.

### DIFF
--- a/src/bmp.imageio/bmpinput.cpp
+++ b/src/bmp.imageio/bmpinput.cpp
@@ -136,12 +136,13 @@ BmpInput::read_native_scanline(int subimage, int miplevel, int y, int z,
     // if the height is positive scanlines are stored bottom-up
     if (m_dib_header.width >= 0)
         y = m_spec.height - y - 1;
-    const int scanline_off = y * m_padded_scanline_size;
+    const int64_t scanline_off = y * m_padded_scanline_size;
 
-    std::vector<unsigned char> fscanline(m_padded_scanline_size);
+    std::unique_ptr<unsigned char[]> fscanline(
+        new unsigned char[m_padded_scanline_size]);
     fsetpos(m_fd, &m_image_start);
-    fseek(m_fd, scanline_off, SEEK_CUR);
-    size_t n = fread(&fscanline[0], 1, m_padded_scanline_size, m_fd);
+    Filesystem::fseek(m_fd, scanline_off, SEEK_CUR);
+    size_t n = fread(fscanline.get(), 1, m_padded_scanline_size, m_fd);
     if (n != (size_t)m_padded_scanline_size) {
         if (feof(m_fd))
             errorf("Hit end of file unexpectedly");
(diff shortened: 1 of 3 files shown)


src/bmp.imageio/bmpinput.cpp
```diff
@@ -136,12 +136,12 @@ BmpInput::read_native_scanline(int subimage, int miplevel, int y, int z,
     // if the height is positive scanlines are stored bottom-up
     if (m_dib_header.width >= 0)
         y = m_spec.height - y - 1;
-    const int scanline_off = y * m_padded_scanline_size;
+    const int64_t scanline_off = y * m_padded_scanline_size;
 
-    std::vector<unsigned char> fscanline(m_padded_scanline_size);
+    std::unique_ptr<unsigned char[]> fscanline(new unsigned char[m_padded_scanline_size]);
     fsetpos(m_fd, &m_image_start);
```
fpsunflower: Shouldn't calls to `fsetpos` also be adjusted? (Even if in practice its likely to be near the start of the file and shouldn't need a 64-bit offset)

Also, wouldn't it be possible to combine the `fsetpos` and `fseek` calls into one?

src/bmp.imageio/bmpinput.cpp
```diff
@@ -157,16 +157,17 @@ BmpInput::read_native_scanline(int subimage, int miplevel, int y, int z,
...
              i += m_spec.nchannels)
             std::swap(fscanline[i], fscanline[i + 2]);
-        memcpy(data, &fscanline[0], m_spec.scanline_bytes());
+        memcpy(data, fscanline.get(), m_spec.scanline_bytes());
         return true;
     }
 
-    std::vector<unsigned char> mscanline(m_spec.scanline_bytes());
+    size_t scanline_bytes = m_spec.scanline_bytes();
+    std::unique_ptr<unsigned char[]> mscanline(new unsigned char[scanline_bytes]);
```
fpsunflower: Are you sure you need this copy? At the end, `mscanline` goes straight into `data`. Am I missing something subtle?

src/bmp.imageio/bmpinput.cpp
lgritz: `man fsetpos/fgetpos` says: " These functions provide  a portable way to seek to offsets larger than those that can be represented by a long int." So they are 64 bit safe already. 

I didn't write this code originally, so I was trying to make my changes minimal lest I break something. But I think you're right that they can probably be eliminated in favor of 64 bit safe fseek/ftell. Let me take a look.

src/bmp.imageio/bmpinput.cpp
lgritz: No, I'm not sure, you're probably right. I was just chasing the low-hanging fruit on these second tier to make them 64 bit safe and, in a couple of spots, making the most obvious perf improvements when I didn't have to think about it hard.

lgritz:
By the way, the background here is that after finding that bug in openexr deep files with large files, I wrote a test harness that creates, writes, reads (and checks that everything is still the same) a > 4GB file for each format we support. I'm saving the PR for that test until the end. But it identified problems in almost every file format, which I spent last weekend fixing one by one and now am doling out the PRs individually. For most of these formats, I wasn't the person who wrote them, and many of them are not the important formats we use daily in VFX. So quite often I was purposely just trying to make the minimally intrusive fix to make them demonstrably big-safe.

src/bmp.imageio/bmpinput.cpp
lgritz: Yes, replaced and update pushed.

src/bmp.imageio/bmpinput.cpp
fpsunflower: Yeah totally unrelated to this review - just looked like another easy potential speedup.

lgritz: When I was testing enormous 64k x 64k images, some formats took seconds to do a write/read cycle, others took minutes. In a couple of the latter instances that were most egregious, I spent a few minutes looking for obvious inefficiencies that could be fixed with just a couple lines and not much thought, just because it grated on me that this test overall took too long to run to be incorporated into the default CI suite. But it was a struggle, because I also knew that the bad ones were file formats that we never use, and that would never plausibly be used for images of that size, so it didn't really matter. Does BMP need to be fast? I think the answer is a clear no. But the perfectionist in me is upset by the inefficiency and needs to be pinned down long enough let the feeling pass and move on to things that actually matter.

Formats using RLE compression are the worst -- separate decisions and branching on every, inherently serial, can't easily be multithreaded or take advantage of SIMD.

PR #2406, opened 2019-11-15
Speed up hdr/rgbe read and write uncompressed by 5-10x

lgritz:
It was calling fread/fwrite for each pixel! Allocate a little memory and
do just one write per scanline.

### DIFF
--- a/src/hdr.imageio/rgbe.cpp
+++ b/src/hdr.imageio/rgbe.cpp
@@ -106,6 +106,28 @@ float2rgbe(unsigned char rgbe[4], float red, float green, float blue)
   }
 }
 
+
+inline void
+float2rgbe(unsigned char* rgbe, const float* rgb)
+{
+    float red = rgb[0], green = rgb[1], blue = rgb[2];
+    float v = red;
+    if (green > v) v = green;
+    if (blue > v) v = blue;
+    if (v < 1e-32) {
+        rgbe[0] = rgbe[1] = rgbe[2] = rgbe[3] = 0;
+    } else {
+        int e;
+        v = frexpf(v,&e) * 256.0f/v;
+        rgbe[0] = (unsigned char) (red * v);
+        rgbe[1] = (unsigned char) (green * v);
+        rgbe[2] = (unsigned char) (blue * v);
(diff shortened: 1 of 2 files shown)


src/hdr.imageio/rgbe.cpp
```diff
@@ -106,6 +106,28 @@ float2rgbe(unsigned char rgbe[4], float red, float green, float blue)
...
+{
+    float red = rgb[0], green = rgb[1], blue = rgb[2];
+    float v = red;
+    if (green > v) v = green;
+    if (blue > v) v = blue;
+    if (v < 1e-32) {
+        rgbe[0] = rgbe[1] = rgbe[2] = rgbe[3] = 0;
+    } else {
+        int e;
+        v = frexpf(v,&e) * 256.0f/v;
```
fpsunflower: Not related to this review, but these encoding/decoding method cannot be optimized by the compiler (just checked gcc/clang in godbolt).

Its fairly easy to swap in some bit-twiddling in its place.

fpsunflower:
LGTM. The call can technically happen for multiple scanlines at once, no? (I didn't trace it all the way, but it looked like `num_pixels=width * num_scanlines`)

src/hdr.imageio/rgbe.cpp
lgritz: Certainly true. It's just a matter of how much effort to put into a format that is never really on a critical performance path.

lgritz:
Yes, I'd need to add a read_scanlines() method to this class -- right now, since it's not a very important format, it just supplies read_scanline and relies on the parent class default implementation of read_scanlines to loop over the scanlines. The same for writing.

lgritz:
I'm trying to exercise the discipline to NOT crack this open any farther right now, because these are not important formats in production (except to occasionally convert to or from them). I just did one batch of fixes to satisfy myself that I squashed most of the int32 overflow bugs (and in a couple spots, fixed very low hanging perf in cases where it took minutes to run my write/read test of gigapixel files). In fact, this was all one-ish day's fooling around last weekend, but I split it into a different PR for each format and spread posting them over many days. I don't consider this ongoing work, unless somebody complains about the real world performance of these formats.

PR #2407, opened 2019-11-15
Reduce tile cache entry 'used' contention

lgritz:
Each texture query may perform many samples, and each sample may fetch a
tile (or more). EVERY time it was making sure that the 'used' atomic int
was set to 1. That's potentially a lot of contention.

Idea: Only set it once for all those samples. We really only need to mark
it 'used' per texture call, not per sample.

So I augmented find_tile() with an optional parameter mark_same_tile_used
(defaulting to true), and the effect is that if it happens to be the
most recently used tile in the microcache, it will only call use() if
mark_same_tile_used is true. So then, when looping over sampling, it
passes sample==1 for that param, having the effect that for the first
sample only (or of course if any other tile is needed than the last one)
will it mark used and diddle the atomic variable.

In our production scene benchmarks at SGI, when run in a highly
threaded configuration with 52 cores, I'm seeing a 2% average speedup
in overall rendering time (geometric mean of many tests; on a few
tests I'm seeing up to 20% gain, which is crazy, but many other scenes
merely break even).

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -916,14 +916,18 @@ public:
     /// Try to avoid looking to the big cache (and locking) most of the
     /// time for fairly coherent tile access patterns, by using the
     /// per-thread microcache to boost our hit rate over the big cache.
-    /// Inlined for speed.  The tile is marked as 'used'.
-    bool find_tile(const TileID& id, ImageCachePerThreadInfo* thread_info)
+    /// Inlined for speed.  The tile is marked as 'used' if it wasn't the
+    /// very last one used, or if it was the same as the last used and
+    /// mark_same_tile_used is true.
+    bool find_tile(const TileID& id, ImageCachePerThreadInfo* thread_info,
+                   bool mark_same_tile_used)
     {
         ++thread_info->m_stats.find_tile_calls;
         ImageCacheTileRef& tile(thread_info->tile);
         if (tile) {
             if (tile->id() == id) {
-                tile->use();
+                if (mark_same_tile_used)
+                    tile->use();
                 return true;  // already have the tile we want
             }
             // Tile didn't match, maybe lasttile will?  Swap tile
--- a/src/libtexture/texture_pvt.h
+++ b/src/libtexture/texture_pvt.h
@@ -374,16 +374,12 @@ private:
         return texturefile;
     }
 
-    /// Find the tile specified by id.  If found, return true and place
-    /// the tile ref in thread_info->tile; if not found, return false.
-    /// This is more efficient than find_tile_main_cache() because it
-    /// avoids looking to the big cache (and locking) most of the time
-    /// for fairly coherent tile access patterns, by using the
-    /// per-thread microcache to boost our hit rate over the big cache.
-    /// Inlined for speed.
-    bool find_tile(const TileID& id, PerThreadInfo* thread_info)
+    /// Find the tile specified by id.  Just a pass-through to the
+    /// underlying ImageCache.
+    bool find_tile(const TileID& id, PerThreadInfo* thread_info,
+                   bool mark_same_tile_used = true)
     {
-        return m_imagecache->find_tile(id, thread_info);
+        return m_imagecache->find_tile(id, thread_info, mark_same_tile_used);
     }
 
     // Define a prototype of a member function pointer for texture
--- a/src/libtexture/texture_pvt.h
+++ b/src/libtexture/texture_pvt.h
@@ -377,7 +377,7 @@ private:
     /// Find the tile specified by id.  Just a pass-through to the
     /// underlying ImageCache.
     bool find_tile(const TileID& id, PerThreadInfo* thread_info,
-                   bool mark_same_tile_used = true)
+                   bool mark_same_tile_used)
     {
         return m_imagecache->find_tile(id, thread_info, mark_same_tile_used);
     }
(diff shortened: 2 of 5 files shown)


src/libtexture/imagecache_pvt.h
```diff
@@ -916,14 +916,18 @@ class ImageCacheImpl : public ImageCache {
     /// Try to avoid looking to the big cache (and locking) most of the
     /// time for fairly coherent tile access patterns, by using the
     /// per-thread microcache to boost our hit rate over the big cache.
-    /// Inlined for speed.  The tile is marked as 'used'.
-    bool find_tile(const TileID& id, ImageCachePerThreadInfo* thread_info)
+    /// Inlined for speed.  The tile is marked as 'used' if it wasn't the
+    /// very last one used, or if it was the same as the last used and
+    /// mark_same_tile_used is true.
+    bool find_tile(const TileID& id, ImageCachePerThreadInfo* thread_info,
+                   bool mark_same_tile_used = true)
```
fpsunflower: This seems like something sensitive enough that it might be worth not  defaulting the argument so that we can truly see everywhere that calls it?

fpsunflower:
Impressive results from such a small change! I've been wondering about the cost of the texture reference counts for similar reasons.

> In our production scene benchmarks at SGI, 

:smiley:

src/libtexture/texturesys.cpp
```diff
@@ -2208,7 +2208,7 @@ TextureSystemImpl::sample_bilinear(
                     // iteration, as long as we aren't using mirror wrap mode!
                     if (i == 0 || tile_s == 0 || noreusetile) {
                         id.xy(tile_edge[S0 + i], tile_edge[T0 + j]);
-                        bool ok = find_tile(id, thread_info);
+                        bool ok = find_tile(id, thread_info, sample == 0);
```
fpsunflower: Why is it ok for only `sample=0` to mark as used? Can't different samples probe into different tiles?

src/libtexture/imagecache_pvt.h
lgritz: sure

src/libtexture/texturesys.cpp
```diff
@@ -2147,7 +2147,7 @@ TextureSystemImpl::sample_bilinear(
         if (onetile && all(stvalid)) {
             // Shortcut if all the texels we need are on the same tile
             id.xy(sttex[S0] - tile_st[S0], sttex[T0] - tile_st[T0]);
-            bool ok = find_tile(id, thread_info);
+            bool ok = find_tile(id, thread_info, sample == 0);
```
fpsunflower: Same question here actually -- if the `find_tile` calls would always hit the same tile, the call would be outside the loop, no?

So why is it ok to only flag the first tile we found as used?

Is there maybe a different way to structure all of this to recognize when the `id` is changing from iteration to iteration?

lgritz: The thread remembers which tile it saw last... but "last" may be from a previous texture call quite a while ago. So we want to poke the "used" flag on THIS call, but we need not do it multiple times. So this means that the first (of maybe several) probes we do on this call will be sure to set "used", even if it's the same tile we last accessed on a prior call.

If this texture call ends up accessing more than one tile for this call, then that later probe (even if not probe number 0) will be one where the tile it needs isn't the "last" tile, so it too will poke the new tile's used flag.

fpsunflower: Let me repeat this in my own words just to make sure I have it right...

The code in `master` calls `use()` on every access. This patch only calls `use()` for the first sample - and if subsequent calls hit the same tile, they will skip the `use()` call (because it knows it can't be the same one).

If we cross over into another tile, `find_tile_main_cache` will do the call to `use()` on its own. And any subsequent probes into that tile would by definition not need to twiddle `use()` again.

I'm having a hard time reasoning about the call to `use()` after the swap from `lasttile`, but I think it doesn't need the same logic, right? By definition it can only happen "once" for a sequence of probes that hit the same tile.

src/libtexture/texture_pvt.h
```diff
@@ -374,16 +374,12 @@ class TextureSystemImpl : public TextureSystem {
...
-    /// This is more efficient than find_tile_main_cache() because it
-    /// avoids looking to the big cache (and locking) most of the time
-    /// for fairly coherent tile access patterns, by using the
-    /// per-thread microcache to boost our hit rate over the big cache.
-    /// Inlined for speed.
-    bool find_tile(const TileID& id, PerThreadInfo* thread_info)
+    /// Find the tile specified by id.  Just a pass-through to the
+    /// underlying ImageCache.
+    bool find_tile(const TileID& id, PerThreadInfo* thread_info,
+                   bool mark_same_tile_used = true)
```
fpsunflower: Change this one too?

src/libtexture/texturesys.cpp
lgritz: yeah, I think you have it

src/libtexture/texture_pvt.h
lgritz: yes, and oops, I found a place where I didn't do the optimization. Results may be even better, will have to run the benchmarks again!

lgritz: Results were virtually the same. The vast majority of the contention was apparently covered by the first set of changes.

fpsunflower:
LGTM!

PR #2410, opened 2019-11-18
More care with locales -- Strutil::vsprintf was not locale independent

lgritz:
We relied on vsnprintf underneath, which is locale dependent.  This is
important for OSL, which generates code that uses this (it will botch
the formatting of floating point values in OSL printf, format, error,
and other similar calls, if the user has their system set to a locale
with commas as decimal separators).

I don't yet have an all-platforms solution, but for OSX, FreeBSD, and Windows,
we can use vsnprintf_l which allows passing a locale (for which we use
the "C" locale).

It makes me very unhappy that I don't yet have a solution for Linux, which
lacks the *printf_l family of functions. I'm keeping my eyes open for
a fix here. But in the mean time, I think Windows is the platform
most likely to have users whose machines are set to foreign locale
without realizing the implications, so at least it will be fixed there.

Also found a spot in rla output that could botch certain metadata when
in the wrong locale.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -406,6 +406,7 @@ set (CLANG_FORMAT_EXCLUDES "src/include/OpenImageIO/fmt/*.h"
                            "src/dds.imageio/squish/*"
                            "src/gif.imageio/gif.h"
                            "src/hdr.imageio/rgbe.cpp"
+                           "src/libutil/stb_sprintf.h"
      CACHE STRING "Glob patterns to exclude for clang-format")
 find_program (CLANG_FORMAT_EXE
               NAMES clang-format bin/clang-format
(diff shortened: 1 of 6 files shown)


fpsunflower:
I thought all our `*printf` logic was being handled by `fmtlib` ?

lgritz:
Technicality: we still use tinyformat for the printf-like formatting, fmtlib for the python-like formatting). But back to your question...

ALMOST all. The fmtlib/tinyformat are template based, so at build time they have to generate the version specific to the types involved.

When OSL generates code on the fly, this can't be done, so we must rely on something that takes a va_list, since we can generate LLVM IR to put arbitrary data on the stack. So we still retained a vestigial Strutil::vsnprintf that was not type safe, not based on tinyformat/fmtlib, but can be used in situations like that. Neither tinyformat nor fmtlib have a version that takes a va_list.

I'm still mulling over how to solve this fully in the long run. But for now, at least this fixes it for OSX and Windows -- they supply explicit-local-based versions of vsnprintf_l. Linux does not.

fpsunflower:
You could always grab this one:

https://github.com/nothings/stb/blob/master/stb_sprintf.h

which is pretty fast, portable, and completely locale independent.

lgritz:
Ah, maybe that's a solution, thanks.

lgritz:
Pushed an update in which we use stb_sprintf.h as the underlying implementation of Strutil::vsnprintf, since not all platforms natively have vsnprintf_l.

fpsunflower:
LGTM

PR #2421, opened 2019-11-28
IC/TS option "trust_file_extensions"

lgritz:
The default is 0, old behavior. If you set to nonzero, you are
promising the ImageCache that it will always be able to identify a
texture file's format based on the filename extension (e.g. ".exr"
always means it's an openexr file, ".tif" or "tx" for TIFF files, and
you will NEVER call a file ".tx" if it's really openexr, or if you do,
you are ok with it being unable to open it). When you enable it with
that promise, it can eliminate a few redundant file opens and header
reads, by not needing to verify file format types at certain stages of
processing.

I'm not sure I would recommend enabling this in general. But if you
are operating in a closed environment where you are sure that file
extensions will match their types (such as a studio with rigid naming
conventions, where all possible textures have been converted with
maketx), this can eliminate some unnecessary network traffic and file
server workload.

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -204,6 +204,12 @@ public:
     ///           aren't getting any helpful additional information, this
     ///           can cut down on the clutter and the runtime. (default:
     ///           100)
+    /// - `int trust_file_extensions` :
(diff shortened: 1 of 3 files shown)


fpsunflower:
LGTM!

PR #2423, opened 2019-11-29
Update embedded fmtlib to v6.0.0

lgritz:
This updated version of fmt has its floating point conversions to be
locale-independent as we wanted. But we are still not (for this
commit) chaning which places we use fmt versus which we use
tinyformat.

### DIFF
--- a/LICENSE-THIRD-PARTY.md
+++ b/LICENSE-THIRD-PARTY.md
@@ -416,29 +416,28 @@ SOFTWARE.
 
 {fmt} library - https://github.com/fmtlib/fmt
 
-Copyright (c) 2012 - 2016, Victor Zverovich
+The MIT license. https://github.com/fmtlib/fmt/blob/master/LICENSE.rst
+SPDX-License-Identifier: MIT
 
-All rights reserved.
+Copyright (c) 2012 - present, Victor Zverovich
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to
+deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+sell copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
-1. Redistributions of source code must retain the above copyright notice, this
-   list of conditions and the following disclaimer.
-2. Redistributions in binary form must reproduce the above copyright notice,
-   this list of conditions and the following disclaimer in the documentation
-   and/or other materials provided with the distribution.
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
 
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
-ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
-WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
-ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
-(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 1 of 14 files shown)


fpsunflower:
LGTM

lgritz:
Wait, another update coming... they released 6.1 since I first posted this (you can see I already updated that part) and it looks like if I switch the "sprintf" and "to_string" implementations from tinyformat to fmt (essentially, no more tinyformat), I get like a 8x speed boost on float conversions. Testing now...

lgritz:
Yeah, results are very good for not only upgrading fmt to the newest (6.1) release, but also throwing the switches to use it internally even for the sprintf-like formatting. Here are the latest timings:

First, baseline of regular sprintf to a static char buffer (no std::string allocation):

    std::sprintf("%g"): 470 ns

    std::sprintf("%d"):  75 ns

    std::sprintf("%g %d %s %d %s %g", f, i, s, i, s, f): 874 ns

Current master, which uses tinyformat for sprintf() and most to_string() varieties, fmt for format() and to_string(int) (remember that these all return allocated std::string's):

    Strutil::sprintf("%g"):      1340 ns
    Strutil::fmt::format("{:g}"): 180 ns
    Strutil::to_string(float):   1380 ns

    Strutil::sprintf("%d"):          75 ns
    Strutil::fmt::format("{}", int): 71 ns
    Strutil::to_string(int):         70 ns

    Strutil::sprintf("%g %d %s %d %s %g", f, i, s, i, s, f):      2060 ns
    Strutil::fmt::format("{} {} {} {} {} {}", f, i, s, i, s, f):   477 ns

You can see that fmt has been faster than tinyformat for most things already. But now it's even faster, and can take over the printf-style stuff...

Same tests, after this full patch (including the full switchover that I haven't posted yet):

    Strutil::sprintf("%g"):       135 ns
    Strutil::fmt::format("{:g}"): 150 ns
    Strutil::to_string(float):    130 ns

    Strutil::sprintf("%d"):          74 ns
    Strutil::fmt::format("{}", int): 68 ns
    Strutil::to_string(int):         68 ns

    Strutil::sprintf("%g %d %s %d %s %g", f, i, s, i, s, f):      363 ns
    Strutil::fmt::format("{} {} {} {} {} {}", f, i, s, i, s, f):  325 ns

Summary:

* float->string (printf style and to_string) is almost 10x faster than before, and now is ~3x faster than C++ std::sprintf to buffer.

* int->string is as fast or faster than std::sprintf to a buffer (despite also doing allocation!).

* A 6-item conversion to string of mixed int/float/string values is 5.6x faster than before when using printf-style, 1.5x faster for fmt-style, and those two now approximately match and are each 2.7x faster than std::sprintf to pre-allocated buffer.

lgritz:
I've got a few things to clean up before I can update... it seems that tinyformat didn't care about %d vs %u -- in both case it would base the formatting on the actual type (and for that matter, even %s would be accepted for non-strings and give default formatting). Whereas fmt::sprintf actually does the "right thing", treating %u things as unsigned and %d things as signed. I was sloppy (or, alternately, too advantage of tinyformat's odd interpretation) in some sections, am working on straightening these out.

lgritz:
OK, updated with the full set of changes that fix the signed-vs-unsigned sloppiness from tinyformat and pass all tests.

lgritz:
Assuming that you still consider this a LGTM with the additional full upgrade to using fmt for all the formatting (even printf-style, to get full performance)...

I have an interesting dilemma, which is that we were supposed to declare 2.1 released yesterday. I hate to break any ABIs after doing this, nor do I want this to completely slip until next year before getting into a release branch. So here's what I propose: (1) immediately backport JUST the "upgrade which version of fmt we use" to 2.1, but no changes in which functionality uses fmt vs tinyformat, keep that the same; (2) cut the supported 2.1 release (technically 2.1.8); (3) give ourselves at least a few weeks with the full switch thrown in master only to shake out any issues; (4) then backport the rest of the full switchover to RB-2.1 to become part of some subsequent 2.1.9+, which shouldn't be an ABI breakage because the only things changing are inline functions and we aren't taking away any public link symbols.

Does that sound reasonable? 

The more conservative alternative is just to make the whole shebang master only, won't be in a supported release until 2.2. That is definitely safer, but feels disappointing compared to being able to somehow get this upgrade (and the promise of a perf increase for a point release) in under the wire.

fpsunflower:
LGTM, excited to see this switchover happen


In this case I'd be inclined to stick to master only since its a fairly big change. I don't think the performance of formatting is critical enough to warrant being aggressive. Am I forgetting some format call that happens really often ?

lgritz:
So you think I shouldn't even upgrade which version of fmt we bundle, except in master, even if no other behavior changes?

fpsunflower:
I don't think there's any major benefit to trying to shoehorn it into 2.1 if it isn't exercised heavily.

lgritz:
Well, one benefit is that the old version was not properly locale independent, but the new version is. So if anybody starts switching from the printf-based to the fmt-based error functions, say, the new version will avoid a lot of potential subtle bugs.

fpsunflower:
That does make sense then.

lgritz:
Maybe I'll update 2.1 to the new fmt version but leave it as a RC for one more week just in case something turns up.

PR #2427, opened 2019-12-05
Upgrade to fmt 6.1.1 for some minor bug fixes

lgritz:


lgritz:
Update with a fix to a totally wrong format code in oiiotool, which would be harmless with tinyformat but threw an exception for fmt!

lgritz:
Updated again with the magic that disables exceptions. Will still assert for debug builds if you use a nonsensical format spec for the parameters, and in release builds will skip those arguments.

PR #2431, opened 2019-12-16
Replace most uses of boost::thread_specific_ptr with C++11 thread_local

lgritz:
There is still one use left, in ImageCache. I'm not sure how to get rid
of that yet, because the C++11 std version only works for global or static
variables, but we are using it in a class. We'll come back to that later.
In the mean time, this gets rid of all other uses.

One thing this does is fix a link error I get on MacOS when building
with EMBEDPLUGINS=0.


fpsunflower:
LGTM

Issue #2440, opened 2019-12-24
[BUG] pkg-config file missing version


jtojnar:
**Describe the bug**
After [upgrade](https://github.com/NixOS/nixpkgs/pull/76358#pullrequestreview-336278981) to openimageio 2.1.9.0, the `Version` field in the `OpenImageIO.pc` only contains `..`.

**To Reproduce**
1. Install OpenImageIO

**Expected behavior**
`Version` field should contain `2.1.9.0`.

**Platform information:**
 - OIIO branch/version: 2.1.9.0
 - OS: NixOS (same issue on [Debian](https://packages.debian.org/experimental/amd64/libopenimageio-dev/download))
 - C++ compiler: GNU 8.3.0
[...]

PR #2441, opened 2019-12-25
Fix broken version in the built .pc file

lgritz:


lgritz:
Fixes #2440

jtojnar:
Can confirm applying https://github.com/OpenImageIO/oiio/pull/2441/commits/e9bdd69596103edf41b659ad8ab0ca4ce002f6f5.patch fixes the issue.

PR #2449, opened 2020-01-01
Debian -- certain architectures seem to need -latomic with gcc9

lgritz:


lgritz:
I think that #2450, if it works, is a better approach than this one. Closing.

PR #2450, opened 2020-01-01
Detect whether we need to link libatomic

lgritz:


lgritz:
@mfvescovi reports that this fixes several architectures. Merging and backporting.

PR #2453, opened 2020-01-04
Bump the minimum pybind11 that we need

lgritz:
Recent changes (2347) seem to rely on some pybind declarations that
are newer than the minimum version we checked for. Bump the pybind11
required version.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -41,10 +41,8 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * If you are building the Python bindings or running the testsuite:
      * Python >= 2.7 (tested against 2.7, 3.6, 3.7)
      * NumPy
-     * pybind11 >= 2.2.0 (Tested through 2.4.2. It is known that 2.4.0
-       and 2.4.1 have bugs and don't build properly for C++11. For this
-       case, or if no pybind11 is found already on the system, OIIO will
-       auto-download it.)
+     * pybind11 >= 2.4.2 (Tested through 2.4.3. If no adequate pybind11 is
+       found already on the system, OIIO will auto-download it.
  * If you want support for camera "RAW" formats:
      * libRaw >= 0.15 (tested 0.15 - 0.19; libRaw >= 0.18 is necessary for
        ACES support and much better recognition of camera metadata)
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -281,7 +281,7 @@ macro (find_or_download_robin_map)
     # packages be built, or we we are forcing a local copy to be built, then
     # download and build it.
     # Download the headers from github
-    if ((BUILD_MISSING_ROBINMAP AND NOT ROBINMAP_INCLUDES) OR BUILD_ROBINMAP_FORCE)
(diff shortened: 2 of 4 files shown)


lgritz:
Did this work, @sobotka ?

sobotka:
Checking now. Hold tight.

sobotka:
@lgritz clean cmake directly to your `lg-oiio` repository using the lg-pybind11 branch still seems to bomb out.

Topmost error:
```
[ 95%] Building CXX object src/python/CMakeFiles/PyOpenImageIO.dir/py_colorconfig.cpp.o
In file included from /lg-oiio/src/python/py_colorconfig.cpp:31:
/lg-oiio/src/python/py_oiio.h: In function ‘bool PyOpenImageIO::py_buffer_to_stdvector(std::vector<T>&, const pybind11::buffer&)’:
/lg-oiio/src/python/py_oiio.h:267:36: error: no matching function for call to ‘pybind11::buffer::request() const’
     oiio_bufinfo binfo(obj.request());
```

lgritz:
Can you send me the full verbose output of a fresh build from scratch? If you use the makefile wrapper:

    make nuke ; make | tee build.log

lgritz:
Sorry, I meant:

    make nuke ; make VERBOSE=1 | tee build.log

lgritz:
It's not enough just to install the patch, you need to start from the initial cmake config step in order to get it to download the right pybind11 version.

sobotka:
Trying another clean directory cmake build. Will ship log if error.

sobotka:
From a clean directory, did a cmake to source and built.

Build log attached.

[build.log](https://github.com/OpenImageIO/oiio/files/4022439/build.log)

lgritz:
That can't be the whole thing, it doesn't show the cmake config or generation steps at all.

sobotka:
Replacing
```
    if ((BUILD_MISSING_PYBIND11 AND NOT PYBIND11_INCLUDES) OR BUILD_PYBIND11_FORCE)
```
with
```
    if ((BUILD_MISSING_PYBIND11 AND NOT PYBIND11_FOUND) OR BUILD_PYBIND11_FORCE)
```

in `pythonutils.cmake` fixes this, as per @lgritz

lgritz:
Yes, Troy and I had a long off-ticket conversation. There was a bug, as cited above, where if you had pybind11 on your system but it was too old a version, it wouldn't do the auto-download like it was supposed to (and like it did, when it didn't find any pybind11 on your system).

I will amend the PR with that fix.

lgritz:
Amended...


Fix logic in pybind11 auto-download:

If no pybind11 was found on the system, it would correctly auto-download.
I botched the test where, if a pybind11 was found but was an insufficient
version, it would not auto-download as intended. This was just an incorrect
test of PYBIND11_INCLUDES when it should have tested PYBIND11_FOUND.

I also augmented the logic to print a more helpful message pybind11
is found but is insufficient version, and furthermore if it's found
in the ext/ area (the auto-downloaded one) but still too old, it will
remind you that it needs to be blown away and refreshed. You'll see
something like this in the cmake configure output:

    -- Pybind11 was found in /Users/lg/code/oiio/oiio.lg/ext/pybind11/include
    -- ... but was version 2.4.0 (minimum is 2.4.2)
    -- Try removing ext/pybind11 and let me download a newer version.

PR #2459, opened 2020-01-10
Extend OpenEXR to handle float vector metadata (in and out)

lgritz:


lgritz:
Pushed an update -- I hadn't realized that the FloatVectorAttribute this depends on was only introduced in OpenEXR >= 2.2.

Issue #2458, opened 2020-01-09
[BUG] Unexpected behavior of ImageBuf::reset()


ttddee:
**Describe the bug**

In the documentation for ImageBuf::reset() it says:

> Re-initialize as a writeable ImageBuf with the given specification (including resolution, data type, metadata, etc.). The optional zero parameter controls whether the pixel values are filled with black/empty, or are left uninitialized after being allocated (if set to InitializePixels::No).

Maybe it is just the wording in the documentation but whether or not you initialize the pixels does not affect if the ImageBuf itself gets initialized. 

If you want the ImageBuf itself to be initialized you need to pass it a ImageSpec that has the format set, although you might be resetting a Buf that knows its format already.

Passing a Spec without the format set will not initialize the ImageBuf, no matter what the other options are.

**Steps to reproduce the behavior:**

    std::shared_ptr<ImageBuf> img (new ImageBuf("/tmp/test.jpg"));

    std::cout << "Initialized: " << img->initialized() << std::endl;

    ImageSpec sp;
    sp.width = 400;
    sp.height = 400;
    sp.nchannels = 3;
    img->reset(sp, InitializePixels::Yes);

    std::cout << "Initialized: " << img->initialized() << std::endl;

    ImageSpec sp2;
    sp2.set_format(TypeDesc::FLOAT);
    img->reset(sp2, InitializePixels::No);

    std::cout << "Initialized: " << img->initialized() << std::endl;

Output:

    Initialized: 1
    Initialized: 0
    Initialized: 1

**Expected behavior**

    Initialized: 1
    Initialized: 1
    Initialized: 1

lgritz:
I think that maybe we can chalk this one up to unclear documentation, and an unfortunate use of the word "initialize" to mean two different things.

An ImageBuf currently can be in one of four states: 
* LOCALBUFFER -- a read/write ImageBuf that internally owns allocated memory to store its pixels.
* APPBUFFER -- a read/write ImageBuf that refers (in a non-owning way) to pixel memory that lives somewhere else. This is how we make a light-weight ImageBuf that "wraps" some other application-owned memory. The only difference between LOCALBUFFER and APPBUFFER is that the latter does not free the internal pointer when it is destructed or reset.
[...]

PR #2460, opened 2020-01-12
Improve docs for ImageBuf, and some error detection and missing methods

lgritz:
* Tried to expand and clarify explanations of the constructors and
  reset methods in particular, after an issue was filed with some
  confusion.

* Added checks to ensure that for the ctrs and reset methods that take
  an ImageSpec and allocate (or wrap) a buffer, if the spec doesn't
  have enough information to do the allocation or know the size of the
  wrapped memory (spec lacking nonzero width, height, depth,
  nchannels, and a concrete data format), the buffer stays marked as
  UNINITIALIZED.

* Added `ImageBuf::reset()` as a synonym for `clear()` and a reset
  variety corresponding to the "wrapped app buffer" constructor.
  This makes it so that every ctr has a corresponding reset method, and
  vice versa. As I was clarifying the docs, I realized that this would
  be better if fully symmetric. Because these are not virtual functions,
  these changes should not break ABI compatibility and should be safe to
  backport.

* In a few spots nearby I renamed NULL to nullptr, couldn't help myself.

### DIFF
--- a/src/doc/imagebuf.rst
+++ b/src/doc/imagebuf.rst
@@ -11,38 +11,32 @@ ImageBuf is a utility class that stores an entire image.  It provides a
 nice API for reading, writing, and manipulating images as a single unit,
 without needing to worry about any of the details of storage or I/O.
 
-An ImageBuf can store its pixels in one of several ways:
-
-* Allocate "local storage" to hold the image pixels internal to the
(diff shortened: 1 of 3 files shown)


lgritz:
Fixes #2458

PR #2478, opened 2020-01-27
Add concat function to ustring and Strutil

lgritz:
`concat(s,t)` is semantically equivalent to `sprintf("%s%s",s,t)`, but
concat carefully avoids copying any of the characters more than once,
and it performs at most one temporary allocation (typically it does
not allocate temp memory from the heap at all, if the sum of the
strings' lengths are not too long).

Added both a std::string-returning and a ustring-returning version.

Also reworked the implementation of `Strutil::repeat()` to similarly
perform at most one one temporary allocation (typically none) and
not perform any unnecessary copies.

Inspired by work Chris Kulla did in OSL.

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -629,7 +629,7 @@ Strutil::concat(string_view s, string_view t)
     }
     memcpy(buf, s.data(), sl);
     memcpy(buf + sl, t.data(), tl);
-    return std::string(local_buf, len);
+    return std::string(buf, len);
 }
 
 
@@ -648,7 +648,7 @@ Strutil::repeat(string_view str, int n)
     }
     for (int i = 0; i < n; ++i)
         memcpy(buf + i * sl, str.data(), sl);
-    return std::string(local_buf, len);
+    return std::string(buf, len);
 }
 
 
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -615,12 +615,40 @@ Strutil::split(string_view str, std::vector<string_view>& result,
 
 
 std::string
+Strutil::concat(string_view s, string_view t)
+{
+    size_t sl  = s.size();
+    size_t tl  = t.size();
+    size_t len = sl + tl;
+    std::unique_ptr<char[]> heap_buf;
+    char local_buf[256];
+    char* buf = local_buf;
+    if (len > sizeof(local_buf)) {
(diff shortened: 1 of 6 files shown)


src/libutil/strutil.cpp
```diff
@@ -614,13 +614,41 @@ Strutil::split(string_view str, std::vector<string_view>& result,
...
+    std::unique_ptr<char[]> heap_buf;
+    char local_buf[256];
+    char* buf = local_buf;
+    if (len > sizeof(local_buf)) {
+        heap_buf.reset(new char[len]);
+        buf = heap_buf.get();
+    }
+    memcpy(buf, s.data(), sl);
+    memcpy(buf + sl, t.data(), tl);
+    return std::string(local_buf, len);
```
fpsunflower: You want `buf` here, not `local_buf`

src/libutil/strutil.cpp
```diff
@@ -614,13 +614,41 @@ Strutil::split(string_view str, std::vector<string_view>& result,
...
+    std::unique_ptr<char[]> heap_buf;
+    char local_buf[256];
+    char* buf = local_buf;
+    if (len > sizeof(local_buf)) {
+        heap_buf.reset(new char[len]);
+        buf = heap_buf.get();
+    }
+    for (int i = 0; i < n; ++i)
+        memcpy(buf + i * sl, str.data(), sl);
+    return std::string(local_buf, len);
```
fpsunflower: Same here.

lgritz: You are right I hate to tell you this, Chris, but... can you guess?

src/libutil/strutil.cpp
fpsunflower: And *ahem* it looks like I was  the one who wrote the code incorrectly in the first place :(

lgritz: That will teach me an important lesson about cut and paste from untrusted sources...

lgritz: And I am similarly embarrassed about not only failing to notice it while reviewing your code, but failing TWICE MORE when putting the equivalent into OIIO, and then one final failure by actually making test cases, but that do not cover the "long string" case.

lgritz:
Updated. 

"Untested code is buggy code" has proven itself once again.

fpsunflower:
LGTM!

Issue #2480, opened 2020-01-29
[FEATURE REQUEST] Ability to provide a WrapMode to ImageBufAlgo::fit


alister-chowdhury:
Hello again,

This is pretty much in the same vein as https://github.com/OpenImageIO/oiio/issues/2479
Whereby using ImageBufAlgo::fit, with exact=true, results in warp being called with a wrap mode of WrapBlack.

And my proposed solution would also be to allow an optional argument of wrap=WrapBlack for ImageBufAlgo::fit.

If you'd agree with that, and are comftable, I could make a PR with both changes at the same time?

Many thanks

lgritz:
Yes, a PR implementing this would be very welcome!

lgritz:
Closed due to merge of #2341

Issue #2479, opened 2020-01-28
[FEATURE REQUEST] Ability to provide a WrapMode to ImageBufAlgo::resample


alister-chowdhury:
Hello,

When using resample over fit, resize etc because of its performance benefits.
However when upscaling small images, the default WrapBlack used in interppixel causes an unwanted black haloing effect around the borders.

My proposition would be to simply add wrap=WrapBlack as an argument to resample, and have it feed wrap to interppixel.

Many thanks.

lgritz:
That's a really good idea.

alister-chowdhury:
@lgritz 
Would you like me to close this ticket and #2480 ?

lgritz:
Thanks for reminding me, I'll care of it.

Closed due to merge of #2341

PR #2481, opened 2020-01-30
Added wrapmode arguments to ImageBufAlgo::fit and ImageBufAlgo::resample

alister-chowdhury:
## Description

This PR addresses #2480 and #2479 

As mentioned in the tickets, I simply added a wrap argument.
This should hopefully not break any existing code, as I've added wrappers to match the previous signatures.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -622,13 +622,17 @@ bool OIIO_API rotate (ImageBuf &dst, const ImageBuf &src,
 
 ImageBuf OIIO_API resize (const ImageBuf &src,
                           string_view filtername = "", float filterwidth=0.0f,
+                          ImageBuf::WrapMode wrap=ImageBuf::WrapMode::WrapClamp,
                           ROI roi={}, int nthreads=0);
 ImageBuf OIIO_API resize (const ImageBuf &src, Filter2D *filter,
+                          ImageBuf::WrapMode wrap=ImageBuf::WrapMode::WrapClamp,
                           ROI roi={}, int nthreads=0);
 bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src,
                       string_view filtername = "", float filterwidth=0.0f,
+                      ImageBuf::WrapMode wrap=ImageBuf::WrapMode::WrapClamp,
                       ROI roi={}, int nthreads=0);
 bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
+                      ImageBuf::WrapMode wrap=ImageBuf::WrapMode::WrapClamp,
                       ROI roi={}, int nthreads=0);
 /// @}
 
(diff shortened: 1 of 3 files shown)


src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -649,10 +649,23 @@ bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
 /// needed, rather than attempting to interpolate deep pixels (regardless of
 /// the value of `interpolate`).
 ImageBuf OIIO_API resample (const ImageBuf &src, bool interpolate = true,
-                        ROI roi={}, int nthreads=0);
+                            ROI roi={},
+                            ImageBuf::WrapMode wrap=ImageBuf::WrapDefault,
```
lgritz: Can you please swap the order here (and everywhere), so that roi and nthreads are always the last two parameters, for uniformity with absolutely everything else?

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -649,10 +649,23 @@ bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
...
+                            int nthreads=0);
 /// Write to an exsisting image `dst` (allocating if it is uninitialized).
 bool OIIO_API resample (ImageBuf &dst, const ImageBuf &src,
-                        bool interpolate = true, ROI roi={}, int nthreads=0);
+                        bool interpolate = true, ROI roi={},
+                        ImageBuf::WrapMode wrap=ImageBuf::WrapDefault,
+                        int nthreads=0);
+inline ImageBuf OIIO_API resample (const ImageBuf &src, bool interpolate,
```
lgritz: What you've done here adding inline functions with the old signatures is good, I'm glad you thought of that.

However, this only preserves source compatibility (with a recompile). That is totally fine if you intend for these changes to only be in master (i.e., "someday 2.2").

If you would like the new kind of resample/resize/fit to be backported to release branch 2.1, then we need to preserve link compatibility as well. For that, you'd need to make this here in the header just be the declaration, and then put the body (which still just trivially calls the new, more complete call) in the same source module as the existing one.

I hope you understand what I'm getting at. But again, if you don't need this backported, then this inline approach is fine.

Sorry, one more thing... Can you please move the "back compatibility" versions (whether they be inline or just declarations) down to the bottom of imagebufalgo.h? You can see there is already a big section with this comment: 

    // DEPRECATED(1.9): These are all ...

I would add these ones below it, with a brief comment:

    // DEPRECATED(2.2): moving forward, please favor the ones with a 'wrap' parameter.

src/python/py_imagebufalgo.cpp
```diff
@@ -2795,17 +2800,19 @@ declare_imagebufalgo(py::module& m)
 
         .def_static("resample", &IBA_resample, "dst"_a, "src"_a,
                     "interpolate"_a = true, "roi"_a = ROI::All(),
-                    "nthreads"_a = 0)
+                    "wrapname"_a = "default", "nthreads"_a = 0)
```
lgritz: To match the C++ interface, let's call this parameter "wrap" as well (and the same comment as before about the parameter order).

lgritz:
Thanks! This is very close to finished and I appreciate your sending the PR.

alister-chowdhury:
I switched roi and wrap arguments, changed it to be 'wrap' in python, added the inline functions to the bottom (I'm not in a hurry :)), and added wrap to resize.

- Switched roi and wrap ordering.
- Moved inline functions to the bottom under a deprecated heading (I'm in not specific hurry :))
- Added wrap to resize (with a default of clamp)

alister-chowdhury:
I might be wrong, but the tests may be failing (it's somewhat hard for me to run them where I am currently), because I'm passing the wrap mode into resize via fit, rather than hardcoding the wrap to be WrapClamp.

If this is the case, I'm guessing, it would be preferable to hardcode WrapClamp?

lgritz:
Hmmm, now this is all less obvious.

So the original problem you started with is that resample was getting black fringing because -- unlike resize, which was hard-coded to use clamp wrap mode -- resample was not specifying wrap and thus getting black wrap.

Fit was even weirder -- it looks like it was calling warp with black wrap for the "exact" mode, but calling resize (and thus implicitly using clamp) for the default "inexact" mode.

Now I don't know the right answer. Outstanding questions:

* resample -- Hard code to clamp to match resize? Or provide the option as you did?

* fit -- Was I thinking clearly when I wrote this, or did I just miss the asymmetry? Should it stay this way, or always use clamp, or allow a call-time option?

* resize -- is clamp always the desired result, or should it be an option?

I **think** that if more than one of them allow an optional user-set wrap mode, they should all have the same default, or it's confusing. People think of resample as a less exact and faster version of resample, and fit as a version of resize that forces an overall frame aspect ratio and letterboxes to make it right.

Was resize right all along and the others should just hard-code clamp as well to match (leaving only the fully general `warp` having a user-adjustable wrap mode)? Or should all of them have had the option all along? (and if the latter, what's the default that will give people the result they think is right in the widest range of cases?)

lgritz:
I'm going to need to think about this one...

lgritz:
I think... maybe the best thing to do is radically scale back the ambition of this PR, and just make it so that resample() hard-codes its call to interppixel to pass the 'WrapClamp' wrap mode, to match the behavior of resize(). I don't think that we need to change any of the public APIs or add new parameters, unless somebody make a strong case for the current behavior of resize or fit is not sensible or needs another parameter that uses will need to ponder every time they use it.

Sorry to ask you to back off much of the work you did. Sometimes that's just the way it works out. I think I've thrown away more of my own code that I've committed over the years.

alister-chowdhury:
I have to admit, that was my initial thought, but figured someone somewhere may want black edges to contribute?

Would it be worth updating fit aswell (so as to stop asymmetry)?

That's absolutely fine, I'm the first to throw away everything in favour of a better solution :)

lgritz:
I edited the subject title of the ticket so that it would be more clear what it ultimately did, for anybody browsing the closed PRs.

PR #2491, opened 2020-02-19
Rewrite of clamp() from Alex Wells generates 2x faster code.

lgritz:
Also add clamp unit test and benchmark, and the previously missing
implementation of clamp for vfloat16.


lgritz:
@AlexMWells

AlexMWells:
LGTM.  Should the unit test check NAN returns the low value (as that is a feature/requirement)?

lgritz:
That unit test has existed all along. Its failure with the first implementation you suggested is how the NaN issue came to my attention.

PR #2492, opened 2020-02-20
Improve fmath.h madd function

lgritz:
I was previously trying to use std::fma, but upon further investigation
it was turning into a single vfmadd instruction a lot less than I thought,
in many cases ending up as a function call (which itself was extra slow
when emulating fma on hardware that lacked it).

The better strategy seems to be just saying `a*b+c`, which on gcc &
icc compilers automatically turns into vfmadd when available on the
hardware, and adding a clang-specific pragma ensures this behavior for
clang as well.

Thanks to Alex Wells of Intel for pointing this out to me (in OSL land).


lgritz:
@alexmwells

PR #2494, opened 2020-02-21
fmath.h bit_cast improvements

lgritz:
* static_assert to check sizes
* use force inline
* use intrinsics where available

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -660,17 +660,56 @@ inline OIIO_HOSTDEVICE float sign (float x)
 
 
 template <typename IN_TYPE, typename OUT_TYPE>
-inline OIIO_HOSTDEVICE OUT_TYPE bit_cast (const IN_TYPE in) {
+OIIO_FORCEINLINE OIIO_HOSTDEVICE OUT_TYPE bit_cast (const IN_TYPE in) {
     // NOTE: this is the only standards compliant way of doing this type of casting,
     // luckily the compilers we care about know how to optimize away this idiom.
+    static_assert(sizeof(IN_TYPE) == sizeof(OUT_TYPE),
+                  "bit_cast must be between objects of the same size");
     OUT_TYPE out;
     memcpy ((void *)&out, &in, sizeof(IN_TYPE));
     return out;
 }
 
+#if defined(__INTEL_COMPILER)
+    // On x86/x86_64 for certain compilers we can use Intel CPU intrinsics
+    // for some common bit_cast cases that might be even more understandable
+    // to the compiler and generate better code without its getting confused
+    // about the memcpy in the general case.
+    // FIXME: The intrinsics are not in clang <= 9 nor gcc <= 9.1. Check
+    // future releases.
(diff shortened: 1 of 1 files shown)


lgritz:
@alexmwells

src/include/OpenImageIO/fmath.h
```diff
@@ -660,17 +660,56 @@ inline OIIO_HOSTDEVICE float sign (float x)
...
     memcpy ((void *)&out, &in, sizeof(IN_TYPE));
     return out;
 }
 
+#if defined(__INTEL_COMPILER)
+    // On x86/x86_64 for certain compilers we can use Intel CPU intrinsics
+    // for some common bit_cast cases that might be even more understandable
+    // to the compiler and generate better code without its getting confused
+    // about the memcpy in the general case.
+    // FIXME: The intrinsics are not in clang <= 9 nor gcc <= 9.1. Check
```
fpsunflower: Looks like they were added recently to LLVM:

https://reviews.llvm.org/rL372802

So I'm guessing they will show up in LLVM 10+. However the implementation is just like the original bitcast, so I don't think there will be any practical advantage to these overloads.

Do we have reasons to believe the Intel compiler really needs these specializations?

lgritz: I did see that review. When LLVM/clang 10 is out, we'll augment the conditional to include it. I don't know if it will actually help LLVM or not. As for Intel, I'm basing this on the work of Alex Wells, who spent a long time poring over the assembly code and trying to find every last spot where OSL was failing to vectorize SIMD loops because of their content. I guess Intel understands something about those intrinsics (or the code they expand to).

lgritz: You ok with this, @fpsunflower ?

fpsunflower: Yep, LGTM!

AlexMWells: The typical memcpy version of std::bit_cast takes an address of the object.  During optimization in a SIMD loop, taking an address of an object could force the compiler to use a standard data layout.  This means the object in question would need to exist in AOS format and possibly require gathers/strided load to access it.  The intrinsic overloads are value based and allow the underlying object's data layout to remain "private".  In short by using these specializations a lot better code generation, the kind one would expect from a "bit_cast" happens in a SIMD loop.
Now another compiler (perhaps clang) might be able to recognize memcpy of built_in data types and covert them a value based cast.  It all depends on optimization passes and the order they are applied.

fpsunflower: That makes sense. Since LLVM implements the bitcast intrinsic with memcpy, that idiom is probably well supported (the current implementation certainly has worked well for us for a while now in scalar paths).

I will be curious to see how well the auto-vectorization stuff pans out on non-intel compilers.

AlexMWells: Most all of the Wide(SIMD) OSL library functions vectorize well with clang 9 (at least after removing some stumbling blocks (like inlining limits and the fmath changes coming in).  There are still the most complex gabor noise functions it has issues with, but all the rest are working.

PR #2495, opened 2020-02-23
Backport fmath performance and other fixes from OSL

lgritz:
A variety of minor rewrites of certain math functions to ensure they
generate better machine code and auto-vectorize more cleanly.

* Lots of inline -> OIIO_FORCEINLINE

* Improve clamp used in fast_sin and cos

* fast_safe_pow improve code gen with OIIO_UNLIKELY

* Introduce `OIIO_FMATH_SIMD_FRIENDLY` to allow application switching
  between implementations that give the best scalar performance versus
  sacrificing scalar perf to have the best SIMD vectorization of loops
  containing the fmath function. This is anticipated to be very rare,
  and of course we strive to be simultaneously fastest on scalar & simd,
  but we have one or two cases where such a tradeoff exists.

* Lots of additional fmath benchmarks to help us judge how good they are
  compared to std functions.

* New safe_fmod which not only prevents division by zero but even in other
  cases is much faster than std::fmod.

* New fast_neg is faster than `-float`, in cases where you are ok with
  -(0.0f) being 0.0f instead of actual floating point -0.0f. If you have
  no idea what I'm talking about or why it matters, you definitely will
  like this function!

* Most of these improvements were backported from OSL, made by Alex Wells,
  Intel.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -46,6 +46,28 @@
 
 OIIO_NAMESPACE_BEGIN
 
+/// Occasionally there may be a tradeoff where the best/fastest
+/// implementation of a math function in an ordinary scalar context does
+/// things that prevent autovectorization (or OMP pragma vectorization) of a
+/// loop containing that operation, and the only way to make it SIMD
+/// friendly slows it down as a scalar op.
+///
+/// By default, we always prefer doing it as correctly and quickly as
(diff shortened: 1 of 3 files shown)


lgritz:
@alexmwells

lgritz:
Any objections to any of this?

lgritz:
Merging. If anything turns up to be an issue later, we can always amend.

PR #2512, opened 2020-03-08
Adjust testsuite after adding tga samples to oiio-images

lgritz:
The place where the TGAUTILS samples were stored seems to have disappeared
from the web. So put them in oiio-images. Adjusted the targa-tgautils test
for the new location.

Also took the opportunity to clean up the pnm test and simplify how it's
declared.


etheory:
THANKS!

PR #2516, opened 2020-03-11
Fix resolution unit not properly set in JPEG export

Willem871:
## Description
To give an indication of the physical dimensions (pixels per inch, pixels per mm,...) of the image we can set the "ResolutionUnit", "XResolution" and "YResolution" attributes in the image spec.
For instance for PNG and TIFF this works perfectly, for JPEG however the resolution unit is not properly written.

In jpegoutput.cpp in the open() function the "ResolutionUnit" attribute was read and set in the m_cinfo.density_unit entry. 
However this was overwritten by the jpeg_set_defaults(&m_cinfo) call. 
As a fix I moved the "ResolutionUnit" code to resmeta_to_density().

## Checklist:

### DIFF
--- a/src/jpeg.imageio/jpegoutput.cpp
+++ b/src/jpeg.imageio/jpegoutput.cpp
@@ -307,7 +307,7 @@ JpgOutput::resmeta_to_density()
         m_cinfo.density_unit = 2;
     else
         m_cinfo.density_unit = 0;
-    
+
     int X_density = int(m_spec.get_float_attribute("XResolution"));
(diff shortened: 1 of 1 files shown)


Willem871:
> LGTM. I fixed a whitespace issue for you that was causing it to fail the clang-format test.

Sorry, I thought I applied clang-format correctly. Thanks for fixing the formatting!

lgritz:
No worries. It was just a blank line between functions that I think had a couple spaces instead of being empty. Easy to fix on my end.

PR #2530, opened 2020-03-27
Build correctly against OCIO 2.x (in progress)

lgritz:
And incorporate a build against OCIO master into the "bleeding edge" CI test, so we don't drift out of sync again.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -591,12 +591,39 @@ ColorConfig::configname() const
 
 
 #ifdef USE_OCIO
+
+#    if OCIO_VERSION_HEX >= 0x02000000
+inline OCIO::BitDepth
+ocio_bitdepth(TypeDesc type)
+{
+    if (type == TypeDesc::UINT8)
+        return OCIO::BIT_DEPTH_UINT8;
+    if (type == TypeDesc::UINT16)
+        return OCIO::BIT_DEPTH_UINT16;
+    if (type == TypeDesc::UINT32)
+        return OCIO::BIT_DEPTH_UINT32;
+    // N.B.: OCIOv2 also supports 10, 12, and 14 bit int, but we won't
+    // ever have data in that format at this stage.
+    if (type == TypeDesc::HALF)
+        return OCIO::BIT_DEPTH_F16;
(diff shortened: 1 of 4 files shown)


lgritz:
Just to clarify, OCIO v2 won't be released for months. This is just ensuring that we can build OIIO against OCIO's master at all times.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -591,12 +591,39 @@ ColorConfig::configname() const
...
+inline OCIO::BitDepth
+ocio_bitdepth(TypeDesc type)
+{
+    if (type == TypeDesc::UINT8)
+        return OCIO::BIT_DEPTH_UINT8;
+    if (type == TypeDesc::UINT16)
+        return OCIO::BIT_DEPTH_UINT16;
+    if (type == TypeDesc::UINT32)
+        return OCIO::BIT_DEPTH_UINT32;
+    // N.B.: OCIOv2 also supports 10, 12, and 14 bit int, but we won't
```
doug-walker: Not sure exactly what you use this function for, but I just wanted to clarify something in case it may be helpful ... The OCIO BitDepth enum is unchanged from v1.  What has changed is that we can now process image buffers that are not F32.  However we did not add support for all of the BitDepths that v1 declared.  The pixel types we support in v2 are UINT8, UINT10, UINT12, UINT16, F16, and F32.  Note that we do not support (that is, are unable to process) the UINT14 or UINT32 pixel types.

lgritz: OK, that's good to know. I will adjust the comments.

Currently, the rest of OIIO's interface with OCIO assumes (from v1) that it can only handle float, so that's the only kind of buffer that's passed into this section. But when v2 is released, there are probably places where I can accommodate that without needing to do data type conversions first.

hodoulp: In case it could help OCIOv2 now supports the channel ordering for `PackedImageDesc`:
```
enum ChannelOrdering
{
    CHANNEL_ORDERING_RGBA = 0,
    CHANNEL_ORDERING_BGRA,
    CHANNEL_ORDERING_ABGR,
    CHANNEL_ORDERING_RGB,
    CHANNEL_ORDERING_BGR
};
```

PR #2535, opened 2020-03-31
Fix potential tls errors

lgritz:
It came to my attention that thread_local variables are not allowed to
have initializers that are not constant. But we had two instances of
thread_local std::string, whose default ctr is not constexpr (until
C++20), so that seems suspicious, right?

Replace with unique_ptr<string>, and hide inside a static function so
that we have better control over when it initializes.

### DIFF
--- a/src/libOpenImageIO/imageio.cpp
+++ b/src/libOpenImageIO/imageio.cpp
@@ -215,13 +215,21 @@ openimageio_version()
 
 // To avoid thread oddities, we have the storage area buffering error
 // messages for seterror()/geterror() be thread-specific.
-static thread_local std::string error_msg;
+
+static std::string&
+error_msg()
+{
+    static thread_local std::unique_ptr<std::string> thread_error_msg;
+    if (!thread_error_msg)
+        thread_error_msg.reset(new std::string);
+    return *thread_error_msg;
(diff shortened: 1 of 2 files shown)


fpsunflower:
Can you provide a link to this constant requirement? I can't find it in:

https://en.cppreference.com/w/cpp/language/storage_duration

Also - are you sure that `unique_ptr` satisfies the requirement? The constructor is `constexpr`, but I don't believe it qualifies as a POD type, and it still has a destructor that needs to be invoked. So I'm not sure if its any simpler than `std::string` in the end?

lgritz:
Golly, now I'm having trouble finding those references that I found so easily last week. Hmmm...

I'll look around tonight and see if I can reconstruct how I got to this last week.

Nothing would make me happier than to abandon this. Though I do wonder if the problem is not string vs unique_ptr, but rather just putting it inside the static function. Something was going wrong on dso load related to the tls.

lgritz:
I'm going to close this PR. There is apparently a known problem when libraries linked against jemalloc are demand-loaded into a process that didn't start up with it. No amount of this fiddling is going to fix things.

The real solution is to fix OpenVDB and that's underway.

PR #2537, opened 2020-04-01
externalpackages.cmake: Fix static boost to not overlink

Shootfast:
When compiling OIIO against a static version of boost where only a subset of boost libs have been built:
(eg, boost configured with --link=static --with-filesystem --with-system --with-thread --with-headers)
We want to disable the automatic MSVC linking, which will otherwise also try to link against unused and non-existant boost libraries (such as date_time).

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -146,6 +146,9 @@ option (BUILD_MISSING_DEPS "Try to download and build any missing dependencies" 
 # Boost setup
 if (LINKSTATIC)
     set (Boost_USE_STATIC_LIBS ON)
+    if (MSVC)
+        add_definitions (-DBOOST_ALL_NO_LIB=1)
+    endif ()
 else ()
     if (MSVC)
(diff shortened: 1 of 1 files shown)


fpsunflower:
Forgive me since I am not a cmake expert -- but isn't this sort of detail supposed to be handled by cmake's `FindBoost.cmake` ? Or by the cmake config generated out of boost's build system?

Or is this file for working around known limitations of those?

Shootfast:
In the newer style of FindBoost.cmake this is enacted by adding a dependency to Boost::disable_autolinking, but as our Cmake setup doesn't use the new constructs (and already had a similar workaround for BOOST_ALL_DYN_LINK below) I didn't want to rock the boat too much.

I have been experimenting with updating the CMake builder across the board, but this was just an immediate hurdle on my windows static build.

fpsunflower:
That all make sense. Thanks!

lgritz:
Dunno how new it is, but our advertised cmake minimum is 3.12, so newer additions to Boot modules that ship with cmake can't be counted on.

PR #2543, opened 2020-04-07
Make the 'bleeding edge' CI test build libtiff and openexr from master

lgritz:
The idea of this "bleeding edge" entry in the CI test matrix is to
really use the very latest toolchain, in order to spot problems long
before users will encounter them.

To that end, let's make sure we are building and testing the current
master of libtiff and OpenEXR.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -69,7 +69,7 @@ jobs:
             source src/build-scripts/ci-build-and-test.bash
 
   linux-bleeding:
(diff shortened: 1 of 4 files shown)


hobbes1069:
Not that I'm suggesting you actually go to the trouble to do this, but if you moved over to Fedora Rawhide you would also get the latest GCC. Rawhide is basically the test ground for gcc pre-releases.

lgritz:
GitHub CI doesn't give me the option to switch to a different distro.

hobbes1069:
Yeah, that's a shame too. I think there's a way to do it through docker but that's probably more trouble than it's worth.

Issue #2544, opened 2020-04-07
[BUG]Error with concurrent reading jpeg files from IOMemReader


etsach:
**Describe the bug**
I'm accessing a JPEG texure through TextureSystem and ImageCache multiple times, in several concurrent threads. The JPEG file is first loaded to a memory buffer, and added to the image cache using IOMemReader proxy.

**To Reproduce**
Simple test program attached : it loads the JPEG file in memory, and adds it with a proxy to the TextureSystem underlying ImageCache. Then, multiple threads are started (as much as hardware concurrency) and lookup the texture through the TextureSystem.

[bug-2544.zip](https://github.com/OpenImageIO/oiio/files/4444783/bug-2544.zip)

**Expected behavior**
It shouldn't produce any error and always give the same output

**Evidence**
Random error messages : 

> Could not open file "membuffer-grid.jpg"

The code works sometimes. It always works properly with a single thread, the problems get more frequent with more threads.
The issue also happens when OSL is in the loop, with more explicit error messages through the error handler, such as:

> Error:[RendererServices::texture] Could not open file "membuffer-grid.jpg"
> Error:[RendererServices::texture] JPEG error: Improper call to JPEG library in state 0 ("membuffer-grid.jpg")
> JPEG error: Improper call to JPEG library in state 0 ("membuffer-grid.jpg")
> JPEG error: Improper call to JPEG library in state 0 ("membuffer-grid.jpg")

**Platform information:**
 - OIIO 2.1.10
 - OS: ubuntu 18.04
 - C++ compiler: clang 7
 - Any non-default build flags when you build OIIO:
`cmake -DCMAKE_CXX_FLAGS="-Wno-macro-redefined"`

etsach:
I believe this issue is due to IOProxy which cannot be shared between threads. The internal state (file handle or memory pointers) can not be used by multiple threads, even if we add locking.
It's not an issue in the common use case, when no oiio::ioproxy parameter is set, since the InputImage then creates a local, non shared IOFile class.

Here are two possible fixes I can  see:

1. Create a IOProxyFactory class. Add an  "oiio::ioproxyfactory" parameter to ImageSpec to pass a pointer to an IOProxyFactory. Modify input plugins, so that they look first for an IOProxyFactory before an IOProxy, and if they find one, use it to create an IOProxy local to the ImageInput. Specify in the documentation that oiio::ioproxy can't be used if the texture will be read from multiple threads, for instance through an ImageCache

[...]

PR #2549, opened 2020-04-11
JPEG & PNG: Fix loss of config info upon close and reopen

lgritz:
Both the JPEG and PNG readers, because the can only read scanlines in
strictly ascending order, have a section where if they need to backtrack,
they do so by closing the file and re-opening.

But in this close/open, any 'config' settings they were passed when they
opened the first time would get lost. And this could include the IOProxy
they were supposed to be using.

The config info needs to be remembered and applied again if the file
is closed and reopened internal to the reader's operations.


lgritz:
Closes #2544

lgritz:
Original reporter of the bug says that this patch fixes it. Merging.

PR #2555, opened 2020-04-13
Raise openjpeg minimum to 2.0

lgritz:
Openjpeg releases 2.0 in 2015. Continuing to support 1.5 should not be
necessary for future releases.


lgritz:
I'll let this sit here for several days. If anybody objects to raising the openjpeg minimum to 2.0+, please comment here. Otherwise, I'll merge it next week.

PR #2564, opened 2020-04-24
Avoid pulling in tbbmalloc

fpsunflower:
I have no idea if this if the right "cmake" way to fix this, but we've been trying to flush out tbbmalloc from our link path and this was one of the places it was being added.

### DIFF
--- a/src/openvdb.imageio/CMakeLists.txt
+++ b/src/openvdb.imageio/CMakeLists.txt
@@ -5,5 +5,5 @@
 if (OPENVDB_FOUND)
     add_oiio_plugin (openvdbinput.cpp
                      INCLUDE_DIRS ${OPENVDB_INCLUDE_DIRS} ${TBB_INCLUDE_DIRS}
-                     LINK_LIBRARIES ${OPENVDB_LIBRARIES} ${TBB_LIBRARIES} ${BOOST_LIBRARIES})
+                     LINK_LIBRARIES ${OPENVDB_LIBRARIES} ${TBB_tbb_LIBRARY} ${BOOST_LIBRARIES})
 endif()


fpsunflower:
I'm mainly submitting this PR as a way to see what the CI reports back :)

lgritz:
I wonder if it's better to have something like

    if (NOT TBB_COMPONENTS)
        set (TBB_COMPONENTS tbb)
    endif ()
    foreach(_comp ${TBB_COMPONENTS})

and then, if we want other components (or if somebody else reuses this FindTBB.cmake module), it's easy to modify it from the outside to add other tbb components. In particular, it would make a config time `-DTBB_COMPONENTS="tbb;tbbmalloc_proxy"` be able to add the proxy library with no other modifications.

fpsunflower:
Sure, let me add that patch. At least the CI seems happy with taking the others out!

lgritz:
You know, I think also a valid solution is not to modify this file at all, but instead, in src/openvdb.imageio/CMakeLists.txt, change the reference to `${TBB_LIBRARIES}` to `${TBB_tbb_LIBRARY}`.  Maybe the bug is that we were inadvertently pulling in all the components when we just needed the one?

fpsunflower:
I can try that ..

fpsunflower:
Even better -- that line doesn't need to link to any libraries at all it looks like!

fpsunflower:
I opted for doing both. I think in most cases you only need the "tbb" component. "preview" is for stuff that is so bleeding edge its not part of the main library, and the malloc library is better handled at an application level as we've discussed on the OSL side.

Using VDB requires being able to find the tbb headers , but doesn't seem to need to link against them.

fpsunflower:
If in the future we have another use for tbb (please no) then at least we won't re-introduce the malloc library this way.

fpsunflower:
Not all platforms were happy about that ... lets see if third time's the charm?

lgritz:
TBH, I'm on the fence about the FindTBB.cmake change. We were using it incorrectly. We never should have asked for all the libraries.

src/openvdb.imageio/CMakeLists.txt
```diff
@@ -5,5 +5,5 @@
 if (OPENVDB_FOUND)
     add_oiio_plugin (openvdbinput.cpp
                      INCLUDE_DIRS ${OPENVDB_INCLUDE_DIRS} ${TBB_INCLUDE_DIRS}
-                     LINK_LIBRARIES ${OPENVDB_LIBRARIES} ${TBB_LIBRARIES} ${BOOST_LIBRARIES})
+                     LINK_LIBRARIES ${OPENVDB_LIBRARIES} ${TBB_tbb_LIBRARIES} ${BOOST_LIBRARIES})
```
lgritz, line 8: I think it's `${TBB_tbb_LIBRARY}`

lgritz: Never mind! I see now that you are correct.

fpsunflower:
So back out the other change then?

lgritz:
I think that might be best.

Now, on my Mac (homebrew), there is a /usr/local/lib/cmake/TBB/TBBConfig.cmake. If all proper TBB installs include this (and did so dating back to the oldest TBB people are likely to use), then we don't need FindTBB.cmake at all. I can investigate that separately, maybe this file will go away eventually anyway.

PR #2567, opened 2020-04-26
writeable -> writable

lgritz:
Nick Black points out that in many places we have used "writeable", which
is (depending on which dictionary you check) is either a less commonly
used variant of "writable", or just an outright misspelling.

Even charitably interpreting it as an alternate spelling, let's standardize
on the more common variant.

There's only one API function where this comes into play, so to preserve
link compatibility, we'll have both ImageBuf::make_writable() and
make_writeable() as a deprecated synonym that will eventually disappear.


lgritz:
@dankamongmen

PR #2572, opened 2020-04-29
Add the raw:flip attribute

servantf:
## Description

Adding the "raw:flip" attribute. 

This field is set by LibRaw when the image has been rotated internally. 
From my understanding, this is an equivalent to the exif:orientation (Don't know why it is not in this exif).

## Tests

I don't think a test is needed as it is only an export of a known LibRaw structure's field.

## Checklist:

- [ X My code follows the prevailing code style of this project.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -601,7 +601,9 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // The "flip" field gives the information about this rotation.
     // This rotation is dependent on the camera orientation sensor.
     // This information may be important for the user.
-    m_spec.attribute("raw:flip", sizes.flip);
+    if (sizes.flip != 0) {
+        m_spec.attribute("raw:flip", sizes.flip);
+    }
 
     // FIXME: sizes. top_margin, left_margin, raw_pitch, mask?
 
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -596,6 +596,13 @@ RawInput::open_raw(bool unpack, const std::string& name,
 
     const libraw_image_sizes_t& sizes(m_processor->imgdata.sizes);
     m_spec.attribute("PixelAspectRatio", (float)sizes.pixel_aspect);
+    
+    // Libraw rotate the pixels automatically. 
+    // The "flip" field gives the information about this rotation.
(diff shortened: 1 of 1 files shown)


lgritz:
Most of the CI tests are failing, because all the raw-related tests now have additional "raw:flip" data in their metadata printing output. Maybe you should only set the attribute if the field is set to a non-default value?

But I'm also wondering if instead of returning this value as a separate field -- which would go unrecognized by any app that wasn't intentionally looking for it -- maybe a better approach would be to simply incorporate it into the "Orientation" that is reported for the image? In other words, the raw file may be in some sense have both the exif orientation and the flip value, and the combination of those tells you how to view the image, but from OIIO's perspective it can fold both into the orientation and then any app that already responds to orientation will automatically also understand the raw flip.

servantf:
Definitely a good idea to merge it with orientation. However i am not confident currently with the "universal" meaning of the flip values. I only have Canon cameras and their values seems very similar to the exif orientation possible values.

lgritz:
Usually with libraw, if the field is not in one of the "maker" structures, it tends to mean the same thing for all cameras.

lgritz:
https://www.libraw.org/docs/API-datastruct-eng.html  That says it's the same as orientation? Or does it override orientation?

https://www.libraw.org/node/2445  (read the whole thread) that's also a little alarming, it may be the 1-8 orientation tag, or it may be a degree value?

servantf:
https://github.com/LibRaw/LibRaw/blob/master/src/x3f/x3f_parse_process.cpp#L119

I found an example. However, there is this function after : 

https://github.com/LibRaw/LibRaw/blob/master/src/preprocessing/raw2image.cpp#L31

lgritz:
OK, so our code should look at the flip value, if it's a multiple of 90 then we convert to a proper orientation. And then... put the results into Orientation, overriding any Exif value?

servantf:
I am not sure you're right.

For me, orientation describe a rotation to be done by the viewer. That is, the array described by the image is still the "sensor array".

Flip describe a rotation already applied to the image. The image is not to be done by the viewer. However, it may still be a valuable information for some applications.

lgritz:
OK, in that case, then let's go ahead and set "raw:flip" without altering "Orientation." But I think you should only set the flip attribute when it's a non-default value. There's no need to clutter the metadata (and alter all the unit test reference output) for an attribute that is almost never used, except in cases where it is.

PR #2582, opened 2020-05-12
oiiotool: allow optional subimages modifier for many ops

lgritz:
For a large number of oiiotool commands (anything whose implementation
inherits from the OiiotoolOp class), allow the optional modifier
`:subimages=...` to control how the operation applies to different
subimages (when dealing with multi-subimage images).

Of course, previously, the default is that most commands operate only
on the first subimage, discarding the others. The global `-a` (and for
many commands the optional `:allsubimages=1` modifier) cause all
subimages to be preserved and the operation to be performed on all of
them.

This new `subimages=...` allows more fine-grained control by
specifying which subimages will be active, as a comma-separated list
of either numeric indices (starting at 0) or subimage names. The
special string "all" means all subimages (thus `:subimages=all` is
equivalent to `:allsubimages=1`). Any subimage number or name preceded
by a "-" excludes that subimage from the active list.  The operation
is performed on only on the active subimages, with excluded subimages
merely being copied from the first input image. The presence of any
`subimages` modifier implies that all subimages should be preserved.

For example, if a.exr is a "multi-part" OpenEXR image (what OIIO calls
multi-image):

    # color convert just the first subimage, discard the rest
    oiiotool a.exr --colorconvert linear sRGB -o out.exr

    # color convert and output all subimages: -a
    oiiotool -a a.exr --colorconvert linear sRGB -o out.exr

    # color convert just the subimage named "beauty", but output them all
    oiiotool a.exr --colorconvert:subimages=beauty linear sRGB -o out.exr

    # color convert all EXCEPT the subimage named "normal"
    oiiotool a.exr --colorconvert:subimages=-normal linear sRGB -o out.exr

    # color convert subimages 1, 2, and 5, leave the others unmodified
    oiiotool a.exr --colorconvert:subimages=1,3,5 linear sRGB -o out.exr


SolenOchHavet:
Love it!! :)

lgritz:
Thanks for the prompt to consider this issue. I'm really pleased with this solution, and it also inspires me to convert all the remaining ops to be subclasses of OiiotoolOp. It's very powerful to have a central place where I can make an interesting option apply to all commands. It would have taken a long time (and be very error-prone) to do this to each op individually.

PR #2593, opened 2020-05-27
Extra test for multiplication

lgritz:
SSE2 lacks an instruction for 32 bit SIMD int multiply. SSE4 has
it. We simulate it with a trick using unsigned multiply on SSE2. Chris
wondered if this alternate code path was adequately tested for
negative numbers, so I added a test to be sure.


fpsunflower:
Thanks!

PR #2594, opened 2020-05-28
Fix minor issue with fmath's OIIO::clamp

lgritz:
OIIO::clamp had the property that if `val <= low`, it sets `low`.
But actually, when `val == low`, we want to use `val`, not `low`.

This may seem like the same thing, and it is for ordinary types.  But
OSL uses this clamp template for "dual" types (implementing automatic
differentiation), and it causes a case where `val == low` but `val`
carries derivatives that we want preserved, not written over by the
0-derivatives of the fixed `low` value.


lgritz:
@AlexMWells does this look right to you now?

PR #2606, opened 2020-06-11
Use GH CI to upload test output for any job that fails tests

lgritz:
This will make it much easier to investigate failures or update ref
output when failures only happen on CI -- any test that fail tests
will produce a downloadable "artifact" consisting of all tests'
captured console output.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -40,7 +40,7 @@ jobs:
         if: failure()
         with:
           name: ${{ github.job }}
-          path: build/*/testsuite/*/out.txt
+          path: build/*/testsuite/*/*.*
 
   vfxplatform-2019:
     name: "Linux VFX Platform 2019: gcc6/C++14 py2.7 boost-1.66 exr-2.3"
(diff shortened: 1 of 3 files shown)


fpsunflower:
Do we have any tests that output images? Or might we have any in the future? Maybe capture `out.*` instead of only `out.txt` ?

lgritz:
Most tests (other than a few utility unit tests) output images. But then they usually compare to a reference image and print the results. Lots of tests are printing metadata or hashes to verify images.

Because of how large the artifact would be if I included all image output, and that my first important use of this was for text output (separate PR coming soon), I omitted it for now, thinking that I could add if/when we need it. But if you think we should just be proactive and bundle them all now, it's an easy change.

Yeah, sure, that makes sense. I'll amend.

I'm gonna add the same bits to OSL's CI as well, this is super useful, especially for those test differences that are acceptable differences, but just ever so slightly varying for each platform, so I can't quite get the same results on the machines I have direct access to.

lgritz:
Amended: We now upload all outputs (including images), but only from the actual tests that fail.

fpsunflower:
Sweet - will be very handy on the OSL side too. I was just thinking about the sampler change there and how to update the windows/mac ref images ...

lgritz:
I'll submit the OSL version of this right away.

PR #2610, opened 2020-06-18
Updated vcpkg instructions to aid discoverability

tlm2021:
## Description
Added a line for vcpkg users to discover additional opencolorio features they can build with that tool. This will help avoid people trying to build manually to enable features available via that relatively obscure package manager. Looking in the vcpkg repo isn't as helpful for people unfamiliar with that package manager or how it is configured.

## Tests
None. Documentation patch only.


lgritz:
That's a great idea, thanks.

PR #2613, opened 2020-06-25
Update LibRaw longitude

JenusL:
## Description

LibRaw recently fixed their long lasting spelling error on longitude (https://github.com/LibRaw/LibRaw/pull/304)
This means that any version of LibRaw from 0.20-Beta3 and newer has this fix and it will break oiio which uses the old misspelled parameter name.

This fix adds a LibRaw version check to know if it should use the misspelled parameter or the correct one.
Just keep in mind that this will cause LibRaw 0.20 versions older than Beta3 (i.e.. Beta1 and Beta2) to not work any more.

## Tests
Don't think a test is needed for this small fix.

## Checklist:


lgritz:
Thanks! I did notice that our "bleeding edge" CI test (which builds against libraw top of tree master) was failing in the last couple days, related to compilation errors in the libraw code, but I hadn't quite gotten to look into it yet. This is obviously the root of the problem.

I'll merge as soon as the CI tests are finished.

PR #2618, opened 2020-06-27
Avoid unexplained build break with VS2015 -- presumed compiler bug

### LINKED ISSUE #2611: new ArgParse code does not build with VS2015
After commit e81036b6559 (#2531), I can no longer build on Windows using VS2015 because I get build errors of the form:
```
oiio\src\oiiotool\oiiotool.cpp(5244): error C2664: 'OpenImageIO_v2_2::ArgParse::Arg &OpenImageIO_v2_2::ArgParse::Arg::action(int (__cdecl *)(int,const char **))': cannot convert argument 1 from 'void (__cdecl *)(OpenImageIO_v2_2::span<const char *const ,-1>)' to 'OpenImageIO_v2_2::ArgParse::ArgAction &&'
oiio\src\oiiotool\oiiotool.cpp(5244): note: Reason: cannot convert from 'overloaded-function' to 'OpenImageIO_v2_2::ArgParse::ArgAction'
oiio\src\oiiotool\oiiotool.cpp(5244): note: No constructor could take the source type, or constructor overload resolution was ambiguous
```
 Looking at the code, I can't see how this could be considered ambiguous and this should work, so I suspect a compiler bug. And in fact, if I build with VS2019 I get no error.

I'm hoping very soon we'll be building OIIO with a newer compiler than VS2015, so in my particular case I don't think I need this fixed in OIIO. However, currently the minimum visual studio required, according to the following is 2015:
```
src/include/OpenImageIO/platform.h
# if _MSC_VER < 1900
 # error "This version of OIIO is meant to work only with Visual Studio 2015 or later"
# endif
```
So I think either we'd want to raise the minimum VS (I haven't tested 2017, but I do know 2019 works) or implement a temporary work-around? There's also the possibility I'm messing something up, so of course would be good if someone else could do a sanity check?

lgritz:
Doesn't trigger an error on other compilers, or on newer MSVS. But this
minor rearrangement makes VS2015 happy, so let's just go with it and
not look under that rug.

Closes #2611

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5307,8 +5307,8 @@ getargs(int argc, char* argv[])
     ap.arg("--metamerge", &ot.metamerge)
       .help("Always merge metadata of all inputs into output");
     ap.arg("--crash")
-      .action(crash_me)
-      .hidden();
+      .hidden()
+      .action(crash_me);
     ap.separator("Commands that read images:");
     ap.arg("-i %s:FILENAME")
       .help("Input file (options: now=, printinfo=, autocc=, type=, ch=)")


JenusL:
This just fixes one of the several compile errors reported in #2611. We should keep this PR open until we have fixed all of them.

PR #2619, opened 2020-06-28
oiiotool - remove argc/argv and substitute cspan

lgritz:
This allows us to remove some older-style vestiges of ArgParse, and in
the process -- hopefully -- avoid some ambiguous polymorphism that older
MSVS seems unable to resolve.

Possibly addresses #2611 


lgritz:
abandoning for now

PR #2624, opened 2020-07-02
unordered_map_concurrent: fix missing size decrement upon erase

### LINKED ISSUE #2623: [BUG] unordered_map_concurrent does not decrement size in erase
**Describe the bug**
In `unordered_map_concurrent::erase`, should add `--m_size` if an element is really erased.

**To Reproduce**
Check the code logic of `unordered_map_concurrent`

**Expected behavior**
The `unordered_map_concurrent::m_size` should reflect the actual size in container.

**Evidence**
```
unordered_map_concurrent<int, int> my_map;
int key=1;
int value=2;
my_map.insert(key, value);
my_map.erase(key);
std::cout << my_map.size() << std::endl; // should be 0 here!
```

**Platform information:**
 - OIIO branch/version: Release 1.8.17
 - OS: CentOS 7.5
 - C++ compiler: GCC 8.1

lgritz:
Closes #2623

### DIFF
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -416,6 +416,7 @@ public:
         if (do_lock)
             bin.lock();
         bin.map.erase(key, hash);
+        --m_size;
         if (do_lock)
             bin.unlock();
     }


fpsunflower:
Did you benchmark this? Any reason to think this might influence the texture system?

ElaraFX:
> Did you benchmark this? Any reason to think this might influence the texture system?

I reviewed the places in OIIO that use this class, it does not affect the correctness of OIIO's texture system itself. However, we re-used OIIO's infrastructure to build some other cache system, which requires this class to be correctly implemented.

lgritz:
It only fixes the fact that as we called erase(), it was not decrementing the overall count of entries in the map (which we were incrementing when we add entries). In our code base, there are no consequences, which is why we never noticed the bug.

fpsunflower:
LGTM then!

PR #2635, opened 2020-07-08
Fix typo in `src/raw.imageio/rawinput.cpp`

zshipko:
## Description

Fixes a typo in `src/raw.imageio/rawinput.cpp`

## Tests

N/A

## Checklist:

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -665,7 +665,7 @@ RawInput::open_raw(bool unpack, const std::string& name,
 #    if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0, 20, 0)
         add("GPS", "Longitude", other.parsed_gps.longitude, false, 0.0f);
 #    else
-        add("GPS", "Longitude", other.parsed_gps.longtitude, false,
+        add("GPS", "Longitude", other.parsed_gps.longitude, false,
             0.0f);  // N.B. wrong spelling!
 #    endif
         add("GPS", "TimeStamp", other.parsed_gps.gpstimestamp, false, 0.0f);


lgritz:
@zshipko It may help to explain the backstory here.

Apparently, for many years, this field in libraw headers was misspelled "longtitude." In the very most recent beta (they are preparing for a 0.20 release), they fixed it. So now we have an `#if` on our side that says "longitude" for >= 0.20 and the old "longtitude" for < 0.20.

But the problem is... if you are building against an *earlier* beta of 0.20, it says it's 0.20 but still has the old misspelling. There's no way for us to distinguish between 0.20-beta2 and 0.20-beta3, unfortunately. So we just had to assume that soon 0.20 will be released, everything on our end will be fine with either 0.19 releases or 0.20 releases, and nobody will care that it's broken in particular early betas.

Pro tip for library authors: EVEN if you're in a beta or prerelease, you need to find some way to detectably bump a version nubmer if you make a change that breaks back-compatibility.

zshipko:
After a while I remembered building a development version of libraw and figured all of this out - but I agree that this is kind of a pain! Perhaps the overlap between 0.20 before the field name was changed is a short enough period of time that it's not really an issue.

Thanks for the background, I felt pretty silly once I realized what was going on. I have also been really enjoying digging through the OIIO codebase!

lgritz:
I figure that 99.9% of people will only every encounter a release version of LibRaw. Just a few people, comparatively, will ever build a beta, and the ones who do are very likely to upgrade to the next beta as soon as it's released. The only people who will build an early beta and keep using it long term are those who forgot they have done so. :-) 

That's what I'm hoping, anyway. I can't think of any way to address it on our side if they have not provided some testable preprocessor symbol that changes at exactly the time that they renamed the field.

zshipko:
Yeah that makes sense - especially since anyone affected by this has already built a beta version of libraw so it shouldn’t be too inconvenient to pull the latest and rebuild.

PR #2636, opened 2020-07-08
Adjust to API changes in OpenColorIO v2

lgritz:

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -19,11 +19,19 @@
 
 #include "imageio_pvt.h"
 
+#define MAKE_OCIO_VERSION_HEX(maj, min, patch)                                 \
+    (((maj) << 24) | ((min) << 16) | (patch))
+
 #ifdef USE_OCIO
 #    include <OpenColorIO/OpenColorIO.h>
+#    if OCIO_VERSION_HEX >= MAKE_OCIO_VERSION_HEX(2, 0, 0)
+#        define OCIO_v2 1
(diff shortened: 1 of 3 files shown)


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -1200,21 +1208,29 @@ ColorConfig::createDisplayTransform(ustring display, ustring view,
...
     if (getImpl()->config_) {
-        OCIO::ConstConfigRcPtr config         = getImpl()->config_;
-        OCIO::DisplayTransformRcPtr transform = OCIO::DisplayTransform::Create();
+        OCIO::ConstConfigRcPtr config = getImpl()->config_;
+#    ifdef OCIO_v2
+        auto transform = OCIO::DisplayViewTransform::Create();
+        transform->setSrc(inputColorSpace.c_str());
+        if (looks.size()) {
+            getImpl()->error(
+                "createDisplayTransform: looks overrides are now allowed in OpenColorIO v2");
```
fpsunflower, line 1217: typo: not allowed?

lgritz: yes, thanks

lgritz:
By the way, I'm expecting a CI failure here because the "bleeding edge" test that is building against OCIO master (where this comes into play) also builds against pybind11 master, which just a few days ago introduced a change that breaks our build. I'm trying to fix that independently of this PR. When I get that fixed, I will rebase this PR on top of that fix, at which point I expect the testsuite to pass.

src/libOpenImageIO/color_ocio.cpp
lgritz: fixed

PR #2639, opened 2020-07-10
Build against more recent versions of {fmt}

lgritz:
fmt is supposed to (if you include fmt/ostream.h) automatically
support any user type that defines an `operator<<()`. And so it did
through version 6.1.2.

Starting with 6.2.x, the presence of `operator bool()` throws it off,
particularly for our TypeDesc, which started rendering as "true" rather
than "uint8" or whatever. It seems that the fact that it can convert
to bool takes precedence, it formats as a bool, rather than notice that
there's a stream ouput option.

In this patch, we add a custom formatter for TypeDesc, which solves
the problem for `format("{}", typedesc)`, but unfortunately not for
`sprintf("%s", typedesc)`. I've filed an issue, but in any case, the
workaround to avoid incorrect formatting when using sprintf is to be
sure to use typedesc.c_str().

### DIFF
--- a/src/include/OpenImageIO/typedesc.h
+++ b/src/include/OpenImageIO/typedesc.h
@@ -171,7 +171,7 @@ struct OIIO_API TypeDesc {
     /// "float", "int[5]", "normal"
     const char *c_str() const;
 
-    friend std::ostream& operator<< (std::ostream& o, TypeDesc t) {
+    friend std::ostream& operator<< (std::ostream& o, const TypeDesc& t) {
         o << t.c_str();  return o;
     }
 
@@ -529,3 +529,34 @@ convert_type(TypeDesc srctype, const void* src,
 
 
 OIIO_NAMESPACE_END
+
+
+
+#if OIIO_USE_FMT
+// Supply a fmtlib compatible custom formatter for TypeDesc.
+FMT_BEGIN_NAMESPACE
+template <>
+struct formatter<OIIO::TypeDesc> {
+    // Parses format specification
+    // C++14: constexpr auto parse(format_parse_context& ctx) {
+    auto parse(format_parse_context &ctx) -> decltype(ctx.begin()) // c++11
+    {
+        // Get the presentation type, if any. Required to be 's'.
+        auto it = ctx.begin(), end = ctx.end();
+        if (it != end && (*it == 's')) ++it;
+        // Check if reached the end of the range:
+        if (it != end && *it != '}')
+            throw format_error("invalid format");
+        // Return an iterator past the end of the parsed range:
+        return it;
+    }
+
+    template <typename FormatContext>
+    auto format(const OIIO::TypeDesc& t, FormatContext& ctx) -> decltype(ctx.out()){
+        // C++14:   auto format(const point& p, FormatContext& ctx) {
+        // ctx.out() is an output iterator to write to.
+        return format_to(ctx.out(), "{}", t.c_str());
+    }
+};
+FMT_END_NAMESPACE
+#endif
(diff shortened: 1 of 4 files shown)


src/include/OpenImageIO/typedesc.h
```diff
@@ -529,3 +529,34 @@ convert_type(TypeDesc srctype, const void* src,
...
+
+
+
+#if OIIO_USE_FMT
+// Supply a fmtlib compatible custom formatter for TypeDesc.
+FMT_BEGIN_NAMESPACE
+template <>
+struct formatter<OIIO::TypeDesc> {
+    // Parses format specification
+    // C++14: constexpr auto parse(format_parse_context& ctx) {
```
fpsunflower, line 541: Did you mean to add `contexpr` here in c++14 mode ?

lgritz: No, I just meant for the comment to remind myself that when C++14 is our minimum, I can switch to the more compact declaration. I barely noticed the constexpr part, the big difference was that C++11 needs the "-> decltype(...)" to go along with the "auto", but C++14 can deduce the return type.

src/cmake/externalpackages.cmake
```diff
@@ -348,7 +348,7 @@ endif ()
 
 option (BUILD_FMT_FORCE "Force local download/build of fmt even if installed" OFF)
 option (BUILD_MISSING_FMT "Local download/build of fmt if not installed" ON)
-set (BUILD_FMT_VERSION "6.1.2" CACHE STRING "Preferred fmtlib/fmt version, when downloading/building our own")
+set (BUILD_FMT_VERSION "7.0.1" CACHE STRING "Preferred fmtlib/fmt version, when downloading/building our own")
```
lgritz, line 351: I'm going to revert this line. I'm not 100% confident I've found every possible place I need to change to work around the "bool" bug, so I'll keep the default back at a safe version of fmt, and bump up just for some of the CI tests and for my own use until I'm sure it's fine.

fpsunflower: Do you think we should stick to 6.1.2 for the renderer as well?

src/include/OpenImageIO/typedesc.h
fpsunflower: I'm assuming that `constexpr` is desirable for when you want to use the compile time format string parsing feature - but I guess we are not using this widely right now?

lgritz: Not using it at all for the moment.

src/cmake/externalpackages.cmake
lgritz: Did it pass the whole renderer and shading testsuites?

fpsunflower: Renderer yes, shading I don't know yet since we haven't done a build.

I'm not sure what kind of `sprintf` calls of `TypeDesc` going through fmtlib would be on a critical path, but it's hard to predict.

lgritz: So far I've just fixed the obvious bugs on the OIIO side, and I haven't yet fully explored what could break on the OSL side. That's next.

lgritz: Like, could it corrupt an oso file as it compiles?

fpsunflower: We compile plenty of shaders as part of the renderer testsuite, but its certainly worth seeing what the impact is on the OSL side.

fpsunflower: In any case - LGTM to this patch!

Issue #2261, opened 2019-06-14
Test OIIO version at compile time


fabiencastan:
### Problem

Is there a good way to easily tests oiio versions at compile time for dealing with compatibility?

After looking at the code in [oiioversion.h.in](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/oiioversion.h.in#L42):
```
#define OIIO_VERSION (10000 * OIIO_VERSION_MAJOR + \
                        100 * OIIO_VERSION_MINOR + \
                              OIIO_VERSION_PATCH)
```

The only solution in user code seems to be:
```
#if OIIO_VERSION <= (10000 * 2 + 100 * 0 + 8)
  // code dealing with compatibility for oiio <= 2.0.8
#else
  // more recent code
#endif
```
It would more readable to be able to do something like:
```
#if OIIO_VERSION <= OIIO_MAKE_VERSION(2, 0, 8)
  // code dealing with compatibility for oiio <= 2.0.8
#else
  // more recent code
#endif
```

lgritz:
We could easily add an OIIO_MAKE_VERSION macro (feel free to make a PR!), but to my personal aesthetic, I'm not sure it's really any more clear overall than the idiom I tend to use:

    #if OIIO_VERSION <= 20008

I think it's compact, easy to type, and easy to read.

lgritz:
Long time coming, but this is implemented by PR #2641

lgritz:
Belatedly closing, this has been implemented.

PR #2641, opened 2020-07-13
Handy macro to construct an integer version code

lgritz:
People have occasionally asked for this.


lgritz:
Closes #2261

PR #2646, opened 2020-07-24
Eliminate some needlessly included headers to speed up build time

lgritz:


hobbes1069:
Did you use any kind of tool to determine this?

lgritz:
Not really.

I did run https://github.com/aras-p/ClangBuildAnalyzer to see what was expensive to compile, and I was a little surprised about how over-included fmath.h was. But then I did the rest by hand.

I know there are automated tools to do this more comprehensively, but this was just a quick stab at some low hanging fruit.

PR #2649, opened 2020-07-28
More gradual fmt conversion

lgritz:
Another step toward switching to std::format notation.

The basic strategy is to make sure that all the classes that have
errorf (etc) methods using printf notation will also have errorfmt
(etc) methods using std::format notation. These will coexist for a
while to let people switch, before the printf ones will be marked
deprecated and then eventually removed.

In some cases where the methods aren't really part of the public
API, but only are used by interns, we are already skipping to the
deprecation marking or just changing them completely.

### DIFF
--- a/src/libutil/typedesc.cpp
+++ b/src/libutil/typedesc.cpp
@@ -343,9 +343,9 @@ tostring_formatting::tostring_formatting(
 
 
 template<class T>
-inline std::string
-sprintt(TypeDesc type, const char* format, const tostring_formatting& fmt,
-        const T* v)
+static std::string
+sprint_type(TypeDesc type, const char* format, const tostring_formatting& fmt,
+            const T* v)
 {
     std::string val;
     if (type.arraylen)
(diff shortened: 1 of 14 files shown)


src/libutil/typedesc.cpp
```diff
@@ -388,6 +403,68 @@ sprintt(TypeDesc type, const char* format, const tostring_formatting& fmt,
 
 
 
+template<class T>
+inline std::string
+formatt(TypeDesc type, const char* format, const tostring_formatting& fmt,
```
fpsunflower, line 408: Maybe call this `format_type`?

lgritz: I can, though it's a strictly internal utility function to that module, and was named that way to be symmetric to the old one called sprintt.

fpsunflower: Right, same thought for `sprintt` -- it just a bit hard to read visually I think with that double `t`. At first I thought it was some other variant of `format` rather than a local function. `sprint_type` is more verbose but more immediately clear what it does.

PR #2657, opened 2020-08-06
Fix mingw compiler warnings

Biswa96:


src/ptex.imageio/ptexinput.cpp
```diff
@@ -105,7 +105,7 @@ PtexInput::open(const std::string& name, ImageSpec& newspec)
             m_ptex->release();
             m_ptex = NULL;
         }
-        errorf("%s", perr);
+        errorf("%s", perr.c_str());
```
lgritz, line 108: What's warning did this one generate?

Biswa96: It produces this error: undefined reference to `Ptex::v2_3::operator<<(std::ostream&, Ptex::v2_3::String const&)`

Biswa96:
One query. In `src/field3d.imageio/CMakeLists.txt` file, is it possible to uncomment `${HDF5_LIBRARIES}` argument? It is required to build in msys2/mingw environment.

lgritz:
> One query. In `src/field3d.imageio/CMakeLists.txt` file, is it possible to uncomment `${HDF5_LIBRARIES}` argument? It is required to build in msys2/mingw environment.

It shouldn't be necessary. The field3d library itself should be linked to hdf5, we shouldn't need to do it directly. Is it just because it's static libraries?

Really, the right solution is for Field3D to have its CMakeLists.txt updated so that it exports a "config" cmake file that sets the dependencies correctly, including the libraries we need to link against, so that we can just do

    target_link_libraries ( blah blah PRIVATE Field3D::Field3D )

and it knows everything.

lgritz:
Another thing to consider is if any users on this platform would care at all about Field3D support. If not, then just disable it and call it a day.

PR #2658, opened 2020-08-06
Reading dpx from mem

ra-mt:
## Description

Add support for reading DPX files from memory. I think it's obvious from the code that I've transformed the LibDpx InStream class into an interface, moved its implementation to InStreamFile and added an InStreamMem class that reads from memory; also updated the DPX OIIO code to allow for an IOProxy when reading.

## Tests

I have separately tested the LibDpx InStreamMem class in a small project, but it is not straightforward to port and I have little experience with OIIO's testing environment.
But I have seen there are tests in place that test reading from memory if an OIIO ImageInput supports "ioproxy" attribute, so reading DPX from memory is tested automatically.

## Checklist:


ra-mt:
I will cancel

PR #2660, opened 2020-08-06
Fix compilation error with armv7, x86

lmarz:
## Description
Fixes a compilation error, that appears on alpine linux.
```
 /builds/lmarz/aports/testing/openimageio/src/oiio-Release-2.1.18.0/src/libOpenImageIO/exif.cpp: In function 'void OpenImageIO_v2_1::read_exif_tag(OpenImageIO_v2_1::ImageSpec&, const TIFFDirEntry*, OpenImageIO_v2_1::cspan<unsigned char>, bool, int, std::set<unsigned int>&, const OpenImageIO_v2_1::pvt::TagMap&)':
 /builds/lmarz/aports/testing/openimageio/src/oiio-Release-2.1.18.0/src/libOpenImageIO/exif.cpp:789:20: error: comparison of integer expressions of different signedness: 'unsigned int' and 'OpenImageIO_v2_1::span<const unsigned char, -1>::index_type' {aka 'int'} [-Werror=sign-compare]
   789 |         if (offset >= buf.size()) {
       |
```

## Checklist:

### DIFF
--- a/src/libOpenImageIO/exif.cpp
+++ b/src/libOpenImageIO/exif.cpp
@@ -786,7 +786,7 @@ read_exif_tag(ImageSpec& spec, const TIFFDirEntry* dirp, cspan<uint8_t> buf,
         unsigned int offset = dirp->tdir_offset;  // int stored in offset itself
         if (swab)
             swap_endian(&offset);
-        if (offset >= buf.size()) {
+        if (offset >= size_t(buf.size())) {
 #if DEBUG_EXIF_READ
             unsigned int off2 = offset;
             swap_endian(&off2);


hobbes1069:
Interesting, it hasn't shown up in the Fedora armv7 builds.

src/libOpenImageIO/exif.cpp
```diff
@@ -783,7 +783,7 @@ read_exif_tag(ImageSpec& spec, const TIFFDirEntry* dirp, cspan<uint8_t> buf,
     if (dir.tdir_tag == TIFFTAG_EXIFIFD || dir.tdir_tag == TIFFTAG_GPSIFD) {
         // Special case: It's a pointer to a private EXIF directory.
         // Handle the whole thing recursively.
-        unsigned int offset = dirp->tdir_offset;  // int stored in offset itself
+        int offset = dirp->tdir_offset;  // int stored in offset itself
```
lgritz, line 786: That doesn't make sense. If you look in tiffutils.h, `tdir_offset` is clearly unsigned.

What version of libtiff are you building against?

lgritz:
Wait a minute, the warning you quoted was about line 789, but your edit is to line 786. Did you change the wrong thing?

lmarz:
I didn't. Alpine Linux is a bit ... weird. For instance, it uses musl libc instead of the GNU libc library. It quite often happens, that the compiler outputs some weird error messages. This error only appears on 2 architectures, the others are fine. It isn't even specifically arch related.  
The error appears a few lines below my change. The compiler says that it's comparing an unsigned int against a signed int (i dont know why this is an error). So I simply changed the offset from unsigned int to int, which eliminates the error and doesn't produce a warning, so I guess it's fine.

src/libOpenImageIO/exif.cpp
lmarz: I'm using tiff version 4.1.0

lgritz:
> The error appears a few lines below my change. The compiler says that it's comparing an unsigned int against a signed int (i dont know why this is an error). So I simply changed the offset from unsigned int to int, which eliminates the error and doesn't produce a warning, so I guess it's fine.

Aah, no, that will just cause other warnings, maybe for people with different compiler versions.

I think the correct fix is to change 

        if (offset >= buf.size()) {

to

        if (offset >= size_t(buf.size())) {

I understand why it's happening. It's because `buf` is a cspan, and its size() returns a signed value, because at the time I made our `span` class, the draft of C++20's span had its size returning a signed value, but it has since been changed to unsigned, much like the other STL types. We'll do this eventually, too, but it's a compatibility change, so I'm waiting until the kind of release where I'm allowed to introduce potential backwards incompatibilities.

PR #2661, opened 2020-08-06
Fix Windows long paths when opening files

ra-mt:
## Description

In order to support Windows extended-length / long paths ("\\?\" style), I've added code to disregard the long path prefix when getting the REST arguments, for opening files.

## Tests

Added 3 test cases in `test_get_rest_arguments()` of "strutil_test.cpp".

## Checklist:

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -265,7 +265,7 @@ Strutil::get_rest_arguments(const std::string& str, std::string& base,
                             std::map<std::string, std::string>& result)
 {
     // Disregard the Windows long path question style prefix "\\?\"
-    const char longPathQPrefix[]    = "\\\\?\\";
+    const char longPathQPrefix[] = "\\\\?\\";
     const size_t longPathQPrefixLen
         = sizeof(longPathQPrefix) / sizeof(longPathQPrefix[0]) - 1;
     std::string::size_type find_start_pos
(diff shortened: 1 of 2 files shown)


lgritz:
In strutil_test.cpp, there is a function test_get_rest_arguments()... can you please add a test case there that exercises this code?

ra-mt:
Thanks! I'll add the test in "strutil_test.cpp" tomorrow -- I've finished for today :)

src/libutil/strutil.cpp
```diff
@@ -264,7 +264,16 @@ bool
...
+    // Disregard the Windows long path question style prefix "\\?\"
+    const char longPathQPrefix[] = "\\\\?\\";
+    const size_t longPathQPrefixLen
+        = sizeof(longPathQPrefix) / sizeof(longPathQPrefix[0]) - 1;
+    std::string::size_type find_start_pos
+        = (str.compare(0, longPathQPrefixLen, longPathQPrefix) == 0)
+              ? longPathQPrefixLen
+              : 0;
+
+    std::string::size_type mark_pos = str.find_first_of("?", find_start_pos);
```
lgritz, line 276: I think this might be more clear as

    static const std::string longPathQPrefix ("\\\\?\\");
    auto find_start_pos = starts_with(str, longPathQPrefix) ? longPathQPrefix.size() : size_t(0);

ra-mt: Yes, It's shorter and more clear. I'll add it tomorrow.

ra-mt:
I've added the clearer code and the test cases. I hope everything is ok now.

PR #2662, opened 2020-08-06
Fix poor performance with texture cache and HDR files

brechtvl:
## Description

When the texture cache reads scanlines in random order, the same pixels could
be read many times over. Now the file offset for scanlines is cached for quick
random access.

Example command that is 18x faster here with this change:
```
oiiotool 8k.hdr --autotile 16 --flip -o out.exr
```

## Tests

I'm not aware of an existing mechanism to test such performance issues.

## Checklist:


lgritz:
That's awesome!

lgritz:
How far back would you like this backported, @brechtvl ? Obvs 2.2. Do you want a 2.1 backport also, considering that 2.2 will be the supported release at the end of the month?

brechtvl:
We don't need a backport to 2.1 then.

PR #2668, opened 2020-08-10
bit_cast should take a reference, not a value

lgritz:

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -733,7 +733,7 @@ inline OIIO_HOSTDEVICE float sign (float x)
 
 
 template <typename IN_TYPE, typename OUT_TYPE>
-OIIO_FORCEINLINE OIIO_HOSTDEVICE OUT_TYPE bit_cast (const IN_TYPE in) {
+OIIO_FORCEINLINE OIIO_HOSTDEVICE OUT_TYPE bit_cast (const IN_TYPE& in) {
     // NOTE: this is the only standards compliant way of doing this type of casting,
     // luckily the compilers we care about know how to optimize away this idiom.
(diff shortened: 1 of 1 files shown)


fpsunflower:
Why? It's expected that this only gets used for simple types like `int` and `float` -- so it shouldn't make any difference? In fact I'd expect that taking the address would only be more confusing for the optimizer?

lgritz:
I looked at the equivalent implementation in a few other projects and they all use a reference. In case somebody uses it for a more complex type, this avoids the possibility of accidentally copying. The optimizer should be really good with references like this, and it has to take the address anyway for the memcpy. I think this common idiom is well understood by today's C++ optimizers.

AlexMWells:
Just found an issue with the change to bit_cast to  a reference, the following specializations for __INTEL_COMPILER no longer compile because they were not updated to take a reference.
Results in compilation issues:
```
/dist/include/OpenImageIO/fmath.h(771): error: no instance of function template "OpenImageIO_v2_4_0::bit_cast" matches the specified type
      template<> OIIO_FORCEINLINE uint32_t bit_cast<float, uint32_t> (const float val) {
```

Simple fix is to just make all the bit_cast specialization accept a reference.

I would echo @fpsunflower that perhaps nothing that can't afford to be copied should be bit_cast.  I've just noticed funny behavior with compilers and references where entire AST trees get moved down into functions possibly causing the the AST segment to be evaluated(inlined) multiple times if passed into multiple functions that take references.

lgritz:
I was just conforming what what the C++20 std::bit_cast does: https://en.cppreference.com/w/cpp/numeric/bit_cast

But you're right, it's a mistake for us to use reference for the template, but not for the specializations. I'll submit a fix right away.

lgritz:
Fix proposed in #3213

PR #2695, opened 2020-09-03
Python: be more forgiving in what's passed to attribute()

### LINKED ISSUE #2674: [BUG] ImageSpec.attribute() always expects a tuple when explicitly setting the data type
**Describe the bug**
I'm getting a `TypeError` when explicitly specifying attribute's data type when copying attributes between specs. In oiio-1.8 there doesn't seem to be any validation on the passed type of the `value` so I'm not quite sure if our code is correct. Now `attribute()` recognizes the data type on it's own but we used to pass a the source attribute's TypeDesc and convert double attributes to floats - though it's not clear to me why this was done and if ImageSpecs ever hold double attributes.

**To Reproduce**

```python
>>> spec = ImageSpec()
>>> spec.attribute("name", oiio.TypeDesc("string"),  "value")
  File "<input>", line 1, in <module>
TypeError: attribute(): incompatible function arguments. The following argument types are supported:
    1. (self: OpenImageIO.ImageSpec, arg0: unicode, arg1: float) -> None
    2. (self: OpenImageIO.ImageSpec, arg0: unicode, arg1: int) -> None
    3. (self: OpenImageIO.ImageSpec, arg0: unicode, arg1: unicode) -> None
    4. (self: OpenImageIO.ImageSpec, arg0: unicode, arg1: OpenImageIO.TypeDesc, arg2: tuple) -> None
Invoked with: <OpenImageIO.ImageSpec object at 0x7f02140caf10>, 'name', <TypeDesc 'string'>, 'value'
```

**Expected behavior**
No error is raised.

**Evidence**
See above.

**Platform information:**
 - OIIO branch/version: 2.2 beta1
 - OS: Cent7
 - C++ compiler: gcc 6.4.0

Any advice would be appreciated.

lgritz:
We have many classes that have methods

    .attribute(name, intval)
    .attribute(name, floatval)
    .attribute(name, stringval)
    .attribute(name, TypeDesc, anyval)

For ImageSpec.attribute() and OIIO.attribute(), the anyval was
restricted to only accept tuples. Fix these to bring them in line with
many other classes' attribute calls in which the 'anyval' really can
be just about anything -- tuples, lists, numpy arrays or other buffer
objects, or scalar values.

Fixes #2674 

### DIFF
--- a/src/python/py_imagespec.cpp
+++ b/src/python/py_imagespec.cpp
@@ -198,7 +198,7 @@ declare_imagespec(py::module& m)
                 const std::string& val) { spec.attribute(name, val); })
         .def("attribute",
              [](ImageSpec& spec, const std::string& name, TypeDesc type,
-                const py::tuple& obj) {
+                const py::object& obj) {
                  attribute_typed(spec, name, type, obj);
              })
         // .def("attribute", [](ImageSpec &spec, const std::string &name, TypeDesc type, const py::list &obj) {
--- a/src/python/py_oiio.cpp
+++ b/src/python/py_oiio.cpp
@@ -173,7 +173,7 @@ oiio_bufinfo::oiio_bufinfo(const py::buffer_info& pybuf, int nchans, int width,
 
 bool
 oiio_attribute_typed(const std::string& name, TypeDesc type,
-                     const py::tuple& obj)
+                     const py::object& obj)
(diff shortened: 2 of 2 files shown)


lgritz:
Reported in #2674 that this fixes the problem.

PR #2696, opened 2020-09-03
Fix ImageBuf vs ImageCache out-of-date image problem

### LINKED ISSUE #1987: OIIO python api is not properly overwriting an image
### Problem

We are running into an issue where we are writing an image "A" with an external DCC and then using oiio to manipulate the file and rewrite over top of the same image with the same file name which I will call image "B".  When we view the file with iv, I see image A.

**Expected behavior:**  
Image B should be in place, not A.

**Actual behavior:**  
I see image A.

### Steps to Reproduce
Make two jpg images(checkerboard) with different sizes (tested with both jpg & exr).
dog.jpg = 640x480
cow.jpg = 1280x720

This will make it easy to verify whether the OIIO python API has the correct image without opening the image in a viewer.

We will use an oiio.ImageBuf to repeatedly open a file called animal.jpg which is created by copying either dog.jpg or cow.jpg to a file with the name animal.jpg

# python code below:
```

import OpenImageIO as oiio

# before proceeding
# in an external shell, copy dog.jpg to animal.jpg
filepath_in = "/var/tmp/animal.jpg"

image = oiio.ImageBuf(filepath_in)
image.read()
print image.spec().width

# should print 640 to the console here

# wipe the contents of the ImageBuf
image.clear()

# now, in an external shell, copy cow.jpg over top of animal.jpg

image = oiio.ImageBuf(filepath_in)

image.read()
print image.spec().width

# should print 1280 to the console here, but prints 640 instead.

# end of code
```

You'll notice that not only are the width and height of the original incarnation of animal.jpg frozen for eternity, the image content is as well. In addition to ImageBuf.clear() we've also tried reset() and read(0,0,True) to force the flushing of the cache. But OIIO python API never seems to be able to dissociate the file name animal.jpg with anything but its first incarnation. Is this a bug? Or is there a documented way to re-read a file whose content has changed. Thanks in advance.

### Versions

* OIIO 1.8.5 (also tested with 1.8.10)
* OS: Centos 6.4
* C++ compiler: gcc 4.8.5

lgritz:
ImageBuf uses ImageCache underneath. But this can lead to the following
unexpected behavior if you don't understand this:

1. Make an ImageBuf for an image, read it (even just header info).
2. Done with it, destroy the ImageBuf.
3. Some other part of the program, or even a different process, changes
   that file on disk.
4. Make another ImageBuf with the same filename, thinking you're getting
   the updated image, but in fact it's relying on the cached representation
   which doesn't contain the changes made in step 3.

So these tiny changes try to make this much harder to shoot yourself in the
foot:

* New ImageBufs automatically call cache->invalidate on the filename
  before they start reading, to ensure that if the file has changed on
  disk, anything in the cache is invalidated.
* When you destroy or clear an ImageBuf, it calls cache->close() on the
  file, to release the handle (especially helpful on Windows, where an
  open file handle can lock the file to other processes).

There is a chance that some image operations will have degraded
performance, if you are mixing direct ImageCache use (including
TextureSystem) and ImageBuf reads of the same files -- since creating
and destroying the IB's will flush tiles from the cache and close cached
file handles. But I think that this should not be a problem in practice,
since most apps that are heavy users of ImageBuf (like oiiotool) aren't
heavey users of IC/TS, and vice versa. And eliminating this confusing
behavior is probably more important than occasional over-eager
invalidation of the cache.

Also clarify in the docs (and imagebuf_test) that if you use the
variety of ImageBuf ctr where you give it a custom ImageCache, it's
important that the ImageBuf outlives the ImageCache, as it is a
non-owning reference.

Fixes #1987

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -117,7 +117,10 @@ public:
     ///             first subimage of the file, highest-res MIP level).
     /// @param imagecache
     ///             Optionally, a particular ImageCache to use. If nullptr,
-    ///             the default global/shared image cache will be used.
+    ///             the default global/shared image cache will be used. If
+    ///             a custom ImageCache (not the global/shared one), it is
+    ///             important that the IC should not be destroyed while the
+    ///             ImageBuf is still alive.
     /// @param config
     ///             Optionally, a pointer to an ImageSpec whose metadata
     ///             contains configuration hints that set options related
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -233,6 +233,17 @@ public:
         return (z * m_spec.height + y) * m_spec.width + x;
     }
 
+    // Invalidate the file in our imagecache and the shared one
+    void invalidate(ustring filename, bool force)
+    {
+        ImageCache* shared_imagecache = ImageCache::create(true);
+        OIIO_DASSERT(shared_imagecache);
+        if (m_imagecache)
+            m_imagecache->invalidate(filename, force);  // *our* IC
+        if (m_imagecache != shared_imagecache)
+            shared_imagecache->invalidate(filename, force);  // the shared IC
+    }
+
 private:
     ImageBuf::IBStorage m_storage;  ///< Pixel storage class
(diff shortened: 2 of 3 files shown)


lgritz:
Does anybody think this is a bad idea?

PR #2698, opened 2020-09-03
Fix pugixml linkage and some mingw build problems

patlefort:
New versions of pugixml no longer generates variables like PUGIXML_LIBRARIES. The proper way is to link to `pugixml`. I also removed the `BEFORE` statement as it is unnecessary as far as I can tell.

I fixed a few problems for cross compiling with mingw from linux. All windows include files should be in lower case. I replaced some _MSC_VER macro checks with _WIN32.


lgritz (approved):
LGTM, thanks!

PR #2704, opened 2020-09-07
IBA::contrast_remap -- move alloca outside loop

### LINKED ISSUE #2703: [BUG] Crash in contrast_remap
Contrast_remap is throwing a segmentation fault on my large images.

I did some code review and saw :

https://github.com/OpenImageIO/oiio/blob/19b7fefdffa8d272aba2cf2ac6ac5a0d3503d8b0/src/libOpenImageIO/imagebufalgo_pixelmath.cpp#L974
https://github.com/OpenImageIO/oiio/blob/19b7fefdffa8d272aba2cf2ac6ac5a0d3503d8b0/src/libOpenImageIO/imagebufalgo_pixelmath.cpp#L975

alloca is called inside a loop :
 - useless as array's length is not changing inside loop
 - causing crash as alloca-ed memory is not freed at scope end (Stack overflow)

I tested my assertion by moving allocation outside of loop. It worked. Not sure why alloca is used here.

lgritz:
This could stack overflow for large images.

Fixes #2703


lgritz:
@servantf

PR #2708, opened 2020-09-13
oiiotool --pastemeta

lgritz:
This new command takes two images as arguments, and appends all the
metadata (only) from the first image onto the second image's pixels
and metadata, producing a combined image.

This can be useful for users who have some (either non-OIIO-based or
just badly implemented) application that operates on image files but
fails to propagate the metadata, by having a simple way to transfer
metadata from one file to another:

    bad-app in.exr -o out.exr
    # oops, out.exr no longer has the crucial metadata from in.exr

    oiiotool in.exr out.exr --pastemeta -o out2.exr
    # Now out2.exr has the pixels of out.exr, but with all metadata
    # from in.exr added to or overwriting that found in out.exr.


lgritz:
@meshula

meshula (commented):
Nice! I love to see how it's a straight forward wrapping of functionality that was already in OIIO :)

meshula:
Thanks for adding this!

lgritz:
@meshula, I'll backport to the release branch (2.2) in time for the October 1 monthly patch release.

PR #2727, opened 2020-09-28
require boost date_time explicitly in cmake

debaetsd:
## Description

This PR adds date_time to the required boost components in cmake.

In order to save on CI, I compile/stage only the (shared lib) boost components needed for OIIO (and other deps).
Based on _src/cmake/externalpackages.cmake_, those should have been filesystem, system, thread & regex.
Without this change, cmake setup is completing successfully yet results in linker errors (see below).
With this change, cmake setup fails to detect the correct boost libraries while also documenting what libraries are actually required.

Linker erros without this change
> -- Build files have been written to: snip/oiio-Release-2.2.6.1/cmBuild
Microsoft (R) Build Engine version 16.7.0+b89cb5fde for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.
> LINK : fatal error LNK1181: cannot open input file 'snip\lib\boost_date_time-vc142-mt-x64-1_7
3.lib' [snip\oiio-Release-2.2.6.1\cmBuild\src\libOpenImageIO\OpenImageIO.vcxproj]
> LINK : fatal error LNK1181: cannot open input file 'snip\lib\boost_date_time-vc142-mt-x64-1_7
3.lib' [snip\oiio-Release-2.2.6.1\cmBuild\src\libutil\OpenImageIO_Util.vcxproj]

## Tests

Compiling only the minimum required subset of boost now works.
Tested on Windows 10. Due to limited availability of other setups, relying on CI to verify other platforms

## Checklist:

- [ x] I have read the [contribution guidelines](../CONTRIBUTING.md).
- [ x] My code follows the prevailing code style of this project.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -47,7 +47,7 @@ if (BOOST_CUSTOM)
     # N.B. For a custom version, the caller had better set up the variables
     # Boost_VERSION, Boost_INCLUDE_DIRS, Boost_LIBRARY_DIRS, Boost_LIBRARIES.
 else ()
-    set (Boost_COMPONENTS filesystem system thread)
+    set (Boost_COMPONENTS filesystem system thread date_time)
     if (NOT USE_STD_REGEX)
         list (APPEND Boost_COMPONENTS regex)
     endif ()
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -32,11 +32,12 @@ option (BUILD_MISSING_DEPS "Try to download and build any missing dependencies" 
 
 ###########################################################################
 # Boost setup
+if (MSVC)
+    # Disable automatic linking using pragma comment(lib,...) of boost libraries upon including of a header
+    add_definitions (-DBOOST_ALL_NO_LIB=1)
+endif ()
 if (LINKSTATIC)
-    set (Boost_USE_STATIC_LIBS ON)
-    if (MSVC)
-        add_definitions (-DBOOST_ALL_NO_LIB=1)
-    endif ()
+    set (Boost_USE_STATIC_LIBS ON)    
 else ()
     if (MSVC)
         add_definitions (-DBOOST_ALL_DYN_LINK=1)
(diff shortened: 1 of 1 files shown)


lgritz:
Where do we use date_time?

debaetsd:
@lgritz 

Well I don't know but it seemed reasonable (and so only a minor cmake config error) :)

But after some digging around, it turns out that OIIO does indeed not uses date_time, it gets auto linked by boost thread.
If you follow some of the includes starting from thread.hpp, you end up in a date_time configured auto_link.hpp (see below for details).
This will try to link the binaries using a pragma comment.

There are solutions to prevent this and I believe expanding the solution in #2537 to be the cleanest
If we define `BOOST_ALL_NO_LIB` (for both static and dynamic linking), it will not attempt to do any kind of auto linking.
Boost docs about this : `Tells the config system not to automatically select which libraries to link against. Normally if a compiler supports #pragma lib, then the correct library build variant will be automatically selected and linked against, simply by the act of including one of that library's headers. This macro turns that feature off.`
As cmake is taking care of linking the libraries, it should not be needed to have boost interfere with it.
This is safe to do as boost.thread is only needed for compiling `Sysutil::physical_concurrency` on clang (and only requiring boost.thread on that specific setup is perhaps overkill)

I locally confirmed that the `BOOST_ALL_NO_LIB` cmake tweak works (on windows).

Alternative we could limit it to `BOOST_DATE_TIME_NO_LIB` to only exclude date_time (though that could repeat these kind of discussion in the future)

How would you like to continue with this? 

This makes this PR in itself completely bogus :) 
I can update this PR/create a new one/something else/... ??


> Flow on how libs get linked (boost 1.73)
> boost\thread.hpp
> boost\thread\thread_time.hpp
> boost\date_time\time_clock.hpp
> boost\date_time\c_time.hpp
> boost\date_time\compiler_config.hpp
> 	-> sets BOOST_LIB_NAME
> boost\config\auto_link.hpp
> 	-> pragma comment(lib,....

lgritz:
Some of this is a bit over my head, but I think that maybe you should submit whatever the minimal BOOST_ALL_NO_LIB tweak you think will work?

If all else fails, I'm also happy to add date_time to our list to capture the apparent transitive dependency, but that seems a little odd because we don't use it directly.

Whatever you think is the least intrusive solution.

debaetsd:
@lgritz I added a new commit moving the BOOST_ALL_NO_LIB. 
I have been applying this exact change on a local build for the past week or so.
This is a MSVC only change.

lgritz (approved):
So in the end, all we needed to do is `BOOST_ALL_NO_LIB=1` on Windows for both dynamic as well as static library boost, instead of only for static?

LGTM.

lgritz:
Merged. I changed the title and explanation a bit to make it reflect where we ended up with the solution.
I will backport to the release branch.

Thanks for the fix!

PR #2730, opened 2020-09-29
WebP input overhaul (including animated webp and metadata)

### LINKED ISSUE #2721: Animated webp images unsupported
OpenImageIO does not support Animated webp images. Not only does it not support the Animation, not even the first frame is read. Instead, there is an error about the image not being valid GIF.

Single image webp images work fine.

Are there plans to add support for animated wbp images?

lgritz:
Huge overhaul of webp reading. Since webp support was first added in
2011 (with no major changes since, other than a couple minor bug
fixes), we have been using the basic/simplified webp API, which had
many limitations.  This patch rewrites it to use the more complete
API. This results in the following changes and new capabilities:

* RGB images are kept as 3-channel instead of forced to be RGBA upon
  input.

* Skip alpha premultiplication for RGB files. And for RGBA, skip the
  dubious color space transformation when doing premultiplication of
  RGBA files. I think this has been wrong all along.

* We can now read animated webp images (akin to animated GIFs, they look
  like multi-subimage files).

* Much more efficient to read just the header, the pixels are read and
  decoded lazily.

* Switch to the WebPDecode*Into API calls, which lets us preallocate
  and manage the decoded image buffer (as a unique_ptr). Also switch
  the encoded buffer from a std::vector to std::unique_ptr, which
  avoids an unnecessary zeroing out of the buffer.

* Close the file handle immediately after we're done reading, rather
  than leaving the handle open until the client is done with the
  ImageInput.

* Read Exif metadata properly from webp files.

* To use the more sophisticated API, we need to find and use the
  libwebpdemux library.

* Since there is a "LoopCount" metadata here, similarly to with GIF, I
  decided to make an OIIO metadata "oiio:LoopCount" that will work
  with both. GIF saves with this, too, and "gif:LoopCount" is now
  considered a deprecated synonym.

Fixes #2721

N.B. This only covers webp input, not output. A similar overhaul should be
done eventually on the output side.


lgritz:
@MickeyKnox are you still out there? Does this fix the problem you reported?

MickeyKnox:
Sorry for the late answer, I've been sick. I've tried this version but unfortunately I could not witness any difference to the behavior of the released version. Animated webp images are still not displayed, not even the first frame, and I continue to face stability issues when there are webp images involved.

lgritz:
Sorry to hear you haven't been feeling well. There is no rush on this, I just didn't want it to be forgotten.

When you say "I've tried this version" do you mean you pulled this patch and applied it, rebuilt, and still have problems? Because it has not yet been merged, so no ordinary checkout or release will get you the fixes from this PR.

There is no doubt that the prior code could not read multiple subimages in a webp file -- it made no attempt to do so, and was completely unaware of the possibility. (That said, it should have allowed you to read the first image, it did so for me, and I never experienced any crashes or instability.) The new code in this PR does attempt to read and handle multiple webp subimages, and I can confirm that it does so, at least for a few webp animations I found on the web, as confirmed by `iinfo -v -a` showing details about all the subimages, and `iv` being able to display the image and page from subimage to subimage via the `>` hotkey.

I'm tempted to merge what we have since it's a clear step forward (I think), though if even with the new code you are seeing instability or are unable to read the subimages of a webp animation, perhaps there is more work to do.

Is there a specific webp image that seems to crash or that even with the code in this PR does not let you read the subimages? If so, can you send it to me (private email is fine), just in case the test images I found on the net are now working but some other detail in other images is triggering the issues?

lgritz:
Addendum: I was incorrect in my statement above, after a couple weeks my memory was faded, but I just double checked.

Before applying this patch, animated webp files were rejected entirely by OIIO's webp reader, it would just fail, claiming it wasn't a valid webp at all. With the code in this PR, it reads it correctly, knows it's a multi-subimage webp file, and appears to correctly read the individual subimages.

MickeyKnox:
My bad. Turns out, I didn't checkout the correct branch (lg-webp). I can confirm that animated webp images work now. The instability issues I was facing are also gone. Thank you.

lgritz:
Woo-hoo, that's great news!

Thanks for checking. I'll merge this into master immediately, and also see if I can get it safely backported to 2.2 in time for the Nov 1 monthly update release.

lgritz:
One thing I should note: It turns out that the way webp stores animated images, the frames only store only a rectangular region containing the pixels that change compared to the previous image. 

So reading sequentially (subimage 0, 1, 2, 3...) is very efficient, it just reads the changed pixel region and applies it to a cached buffer it stores internally.

Reading forward arbitrarily (0, 1, 4, 7) will take a bit longer when you skip frames, because basically it needs to read those intermediate frames along the way.

Truly random access or moving backwards requires (internally) going all the way back to the beginning and then progressing forward to where you are aiming, and so will tend to be low performance if there are many frames (also uneven performance, being more expensive to seek to later in the sequence).

PR #2735, opened 2020-10-06
StrongParam helper for disambiguating parameters

lgritz:
Consider this a design review. The point is all explained in depth in the comments, I won't restate it here.

This is an idea I've been contemplating for a long time. I want people to think about whether using this idiom -- sparingly, in   the kinds of places where we may have used an enum to bring more clarity than a bare bool or int -- would help improve code readability and reduce parameter passing errors. If so, I'd like to start using it. But if people think it's a needless  flourish that detracts from code readability, then I can be persuaded to abandon it.

### DIFF
--- a/src/include/OpenImageIO/strongparam.h
+++ b/src/include/OpenImageIO/strongparam.h
@@ -1,0 +1,116 @@
+// Copyright Contributors to the OpenImageIO project.
+// SPDX-License-Identifier: BSD-3-Clause
+// https://github.com/OpenImageIO/oiio
+
+
+#pragma once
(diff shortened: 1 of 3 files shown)


lgritz:
Does anybody have thoughts, pro or con, about this kind of idiom to clarify parameter passing in some cases?

lgritz:
I've sat on this for a long time, but I'm going to merge now because I just had to fix a really embarrassing bug that would not have happened if I'd used this idiom: https://github.com/OpenImageIO/oiio/pull/2844

PR #2736, opened 2020-10-06
Fix broken read of 16 bit iff files

lgritz:
This used to work, was inadvertently broken in #2398, in Nov 2019, and
no release of OIIO 2.1 ever had it working!  (I guess people don't
come across 16 bit iff files very often?)


lgritz:
Original reporter of the bug says this works. Also will backport to 2.2.

PR #2743, opened 2020-10-14
Fix Cuda build break

lgritz:
Cuda 10.1 seems to have trouble because of the `__attribute__` used in
`OPENIMAGEIO_PRINTF_ARGS`. Declan Russell suggests modifying the `#if`
to avoid using it for `__CUDACC__`.

LG further observes that `OPENIMAGEIO_PRINTF_ARGS` is only used for the
declaration of two functions, it seems like a nice decluttering to
just directly use the short attribute exactly where it is needed
instead of having a lengthy explanation and possibly defining
`__attribute__` everywhere as a side effect.

### DIFF
--- a/src/include/OpenImageIO/strutil.h
+++ b/src/include/OpenImageIO/strutil.h
@@ -59,24 +59,6 @@
 // was locale dependent.
 #define OIIO_FMT_LOCALE_INDEPENDENT 1
 
-#ifndef OPENIMAGEIO_PRINTF_ARGS
-#   ifndef __GNUC__
-#       define __attribute__(x)
-#   endif
-    // Enable printf-like warnings with gcc by attaching
-    // OPENIMAGEIO_PRINTF_ARGS to printf-like functions.  Eg:
-    //
-    // void foo (const char* fmt, ...) OPENIMAGEIO_PRINTF_ARGS(1,2);
-    //
-    // The arguments specify the positions of the format string and the the
-    // beginning of the varargs parameter list respectively.
-    //
-    // For member functions with arguments like the example above, you need
-    // OPENIMAGEIO_PRINTF_ARGS(2,3) instead.  (gcc includes the implicit this
-    // pointer when it counts member function arguments.)
-#   define OPENIMAGEIO_PRINTF_ARGS(fmtarg_pos, vararg_pos) \
-        __attribute__ ((format (printf, fmtarg_pos, vararg_pos) ))
-#endif
 
 
 OIIO_NAMESPACE_BEGIN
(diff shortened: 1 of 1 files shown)


DeclanRussell:
It looks like later down the line we actually changed the ifdef to `if !defined(__GNUC__) && !defined(__clang__)`, as not having the attribute defined I recall caused issues with nvcc. Though I can see this breaking in some cases. I think a more complete solution could be something along the lines of `if !defined(__GNUC__) && !defined(__clang__) && !defined(__CUDACC__)`. In that case it would only be clangs CUDA compiler that excludes the attribute. I'll need to run a build to check this though.

lgritz:
I guess we could remove the attribute entirely. We don't encourage people to use this function, the error checking it provides is of extremely marginal utility.

DeclanRussell:
I don't really have a strong opinion about removing it or not. I tested again with the following condition and it worked fine. So I guess as we have a solution it doesn't hurt to keep it.

`if !defined(__GNUC__) && !(defined(__CUDACC__) && defined(__clang__))`

lgritz:
@DeclanRussell Are you sure that's right? It's specifically a gcc attribute, why would you have `!defined(__GNUC__)`?

DeclanRussell:
I was just going off the original code
```
#   ifndef __GNUC__
#       define __attribute__(x)
#   endif
```

lgritz:
OH, so this is the negative case?

Because in my rewrite of this PR, I never defined `__attribute__` -- that seems unwise -- but instead only include the `__attribute__` in the positive case.

What does the positive case need to be to only include the attribute when it's safe?

I was using

    #if defined(__GNUC__) && !defined(__CUDACC__)

Are you saying it's not ok for clang?

DeclanRussell:
Yeah so if this is already defined by GCC, then we don't need to redefine this when using GCC. However it seems that enabling the positive case and defining `__attribute__` breaks clangs CUDA compiler. So this would fix that,
```
#if !defined(__GNUC__) && !defined(__CUDACC__)
```
Unfortunately this break when compiling with NVCC though (mo' compilers mo'problems :( ). So we need the definition for NVCC. Thats where I came up with,
```
if !defined(__GNUC__) && !(defined(__CUDACC__) && defined(__clang__))
```
This then means that we define `__attribute__` only when we're not using GCC or clangs CUDA backend.

lgritz:
This is hurting my brain. You're still talking about the negative case, right? The one where the argument-checking attribute is not available, so we (previously) defined `__attribute__` to be empty so it wouldn't be an error?

If I understand your guard correctly, it means: "if this is not GCC, and also not Cuda code being compiled by clang, then define `__attribute__` to be nothing so it's harmless."

I don't ever want to redefine `__attribute__`. I only want to have a "positive guard" to include the `__attribute__` in the two places where it's used. So I think the positive case is:

    #if defined(__GNUC__) || (defined(__CUDACC__) && defined(__clang__))

would that work? And for that matter, does it even need the second clause at all? This is all statically checkable, so as far as I can tell, it's fine to only do the `__attribute__` check in one CI test and that will fully catch any problems. So the even simpler construct I proposed earlier,

    #if defined(__GNUC__) && !defined(__CUDACC__)

should be sufficient for the positive case, no? It means "if this is GCC and you're not compiling for Cuda..." and that's the only  case we want the `__attribute__`, right?

I'm so tempted to just rip this whole thing out entirely.

DeclanRussell:
Ahh right I get you. Yup you're right, the current patch is fine. I even tested it to make sure and it works great. Sorry for all the confusion.

PR #2748, opened 2020-10-22
Sketch of a C API for comment

anderslanglands:
## Description

This adds a minimal C API called "coiio" currently that is built as an additional shared library, the implementation of which is under src/coiio and gated behind option BUILD_COIIO (defaults to ON in this PR). I've also added a test under testsuite/coiio that exercises it by loading a tiff, adding a couple of attributes to the header, writing it out as an exr, then reading the exr back and getting the attribute values.

Notable things are right now I've only wrapped `TypeDesc::TypeDesc(string_view)` as `TypeDesc_from_string` in order to generate TypeDescs because it was quick. I imagine we'll ultimately want to declare prebuilt extern TypeDesc objects as a replacement for the static constexpr ones for commonly used types in C++.

One slight wrinkle is in all the *_geterror() functions. In C++ they return a copy of an internally owned std::string. I've worked around this for now by caching the string in a thread_local and returning a pointer from that to C. I'm not completely up to date on the semantics of thread_local (does it even work in Apple's libc++ yet?) so not sure if this is that great.

An alternative would be to add a new function, e.g. `const char* ImageInput::geterrorc()` that would just call m_errmsg.c_str(), then `ImageInput::geterror()` could call _that_ and make a copy of the string to return, while the C API would just call geterrorc, with an explicit warning that the pointer is only valid until the ImageInput is mutated again. Obviously if ImageInput were to be mutated from another thread then all bets are off and thus I don't think this idea works.

Yet another way, and an idiom you see often in C APIs would be to provide a couple more functions:
```
int ImageInput::geterrorlength() { return m_errmsg.size(); }
void ImageInput::geterrormsg(char* msg) { strcpy(msg, m_errmsg.c_str(), m_errmsg.size()); }
```
and have the caller manage the memory. In the multithreaded case this also doesn't work, but another way of doing this is:
```
void ImageInput::geterrormsg(char* msg, int buffer_len) { 
    lock_guard lock(m_mutex);
    strcpy(msg, m_errmsg.c_str(), min(buffer_len, m_errmsg.size()));
}
````

Plenty of opportunity for bikeshedding here too. The name is just what I called the library in my rust crate, so can be whatever you prefer. I would suggest making it different enough from OpenImageIO so that you don't mistake the headers in include statements though, e.g.:
```
#include <COpenImageIO/imageio.h> //< oops, but probably won't lead to subtle bugs so may be ok?
```

## Tests

New test is testsuite/coiio. I also copied over run_app() from OSL to the testsuite to allow building the test binary as part of the test.

## Checklist:

## TODO checklist:

- [x]  Add licence blurb to each file
- [x] 3 lines between functions
- [x] extra newline before closing namespace
- [x] rename files to c-imageio.h (or imageio-c.h?)
- [x] fix doc comments to refer to underlying C++ instead of copying full doc comments
- [x] Resolve geterror() situation

### CODE REVIEW

**Comment by lgritz** [comment]
Some possible alternate designs to consider for the geterror issue:

1. You could have the C version of geterror do a strdup to return, and it be understood that the caller of geterror is responsible for freeing it.

2. You  could have C geterror call the underlying C++ `geterror(/*clear=*/ false)`, returning a  `char*` for the error but not clearing it, and add a second C-only clearerror() that you subsequently call when you're done using the error message.

File: src/coiio/imageio.cpp [hunk 1]
```diff
@@ -0,0 +1,201 @@
...
+}
+
+const char*
+openimageio_geterror(bool clear)
+{
+    thread_local std::string errorstring;
+    errorstring = OIIO::geterror(clear);
+    return errorstring.c_str();
+}
+}
```
lgritz, line 201: I think there should be one more linefeed here

anderslanglands:
1. I had considered strdup but I seem to remember that calling free on
something malloced in another module is bad news on Windows though I don’t
remember the exact rules around that.

2. This would still explode if another thread assigned to m_errmsg between
the first thread calling geterror() and using the returned char* wouldn’t
it?

src/coiio/imageio.h
```diff
@@ -0,0 +1,337 @@
...
+#include "imageio_defines.h"
+#include "typedesc.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// ImageSpec describes the data format of an image -- dimensions, layout,
+/// number and meanings of image channels.
+///
```
lgritz, line 12: I'm really not sure yet how we should end up documenting all of this.

But I would advise against copying the whole comment blocks from the C++ side. At the very least, that's going to be a nightmare to keep synchronized.

For now, let's keep the the comments minimal, merely referring to the equivalent C++ as the primary documentation, and only having extended comments when they are needed to describe differences from the C++. For example, maybe here we would have:

```
/// ImageSpec describes the data format of an image -- dimensions, layout,
/// number and meanings of image channels. It is equivalent to the C++
/// Imagespec, but it is only exposed to C as a pointer to an opaque structure.
typedef struct ImageSpec ImageSpec;

/// Create a new ImageSpec. The caller takes ownership and is responsible
/// for eventually calling ImageSpec_delete.
///
/// Equivalent C++:  `new Imagespec`
///
OIIO_API ImageSpec*
ImageSpec_new();

/// Delete an ImageSpec.
///
/// Equivalent C++: `delete is;`
///
OIIO_API void
ImageSpec_delete(const ImageSpec* is);

/// Create a new ImageSpec with explicit dimensions.
///
/// Equivalent C++:  `new ImageSpec(xres, yres, nchans, fmt)`
///
OIIO_API ImageSpec*
ImageSpec_new_with_dimensions(int xres, int yres, int nchans, TypeDesc fmt);

/// Width of the pixel data window. In C++ we would directly access the structure members,
/// but in C we only have a pointer to an opaque ImageSpec structure.
///
/// Equivalent C++:  `is->width`
///
OIIO_API int
ImageSpec_width(const ImageSpec* is);

```

src/coiio/imageio_defines.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include <OpenImageIO/export.h>
+
+typedef int64_t stride_t;
+typedef uint64_t imagesize_t;
+
+#define AUTOSTRIDE 0x8000000000000000L  // int64_t min
```
lgritz, line 11: Does modern C have the equivalent of `const size_t AutoStride = ...` ?

I'm worried about `#define`ing a generic term like AUTOSTRIDE. Or is that typical of C these days?

src/coiio/typedesc.cpp
```diff
@@ -0,0 +1,27 @@
+#include <OpenImageIO/typedesc.h>
+
+#include "typedesc.h"
+#include "util.hpp"
+
+// Sanity check that our types are equivalent
+static_assert(sizeof(TypeDesc) == sizeof(OIIO::TypeDesc));
+static_assert(alignof(TypeDesc) == alignof(OIIO::TypeDesc));
+static_assert(offsetof(TypeDesc, basetype)
+              == offsetof(OIIO::TypeDesc, basetype));
```
lgritz, line 10: I like the use of static_assert here to ensure that the C and C++ structs stay exactly equivalent.

testsuite/coiio/coiiotest.cpp
```diff
@@ -0,0 +1,147 @@
+#include <coiio/imageio.h>
```
lgritz, line 1: I'd propose that maybe this would be a better test if it were `coiiotest.c` -- C -- so we don't make any mistakes that would make it not work with a true C compiler.

lgritz (commented):
I like the way this is going.

I know this is a "draft", but I added a whole lot of minor comments and suggestions for you to consider as you work on this.

lgritz:
> had considered strdup but I seem to remember that calling free on something malloced in another module is bad news on Windows though I don’t remember the exact rules around that.

The rule is that you can't allocate and free on opposite sides of a DLL boundary, that's right. So it's safe to have ImageSpec_new and ImageSpec_delete, because the underlying malloc and free are both in the coiio library, not in the caller's module. But you're right, we can't return a malloced string and expect the app to free() it directly.

So I think it's better if geterror() just returns a non-owning `char*` to the string still stored in OIIO (and be sure to call C++ geterror(false) so the string isn't freed!), then we'll just have to add a separate `clearerror()` on the C side. We can add clearerror on the C++ side also if it bugs you that we have a C function with no direct equivalent in C++.

src/coiio/typedesc.cpp
lgritz: OH, these aren't legal C++11 (only 14) because they lack an explanation string.
Consider using the `OIIO_STATIC_ASSERT` macro in dassert.h, which takes care of this for you.

anderslanglands:
> > had considered strdup but I seem to remember that calling free on something malloced in another module is bad news on Windows though I don’t remember the exact rules around that.
> 
> The rule is that you can't allocate and free on opposite sides of a DLL boundary, that's right. So it's safe to have ImageSpec_new and ImageSpec_delete, because the underlying malloc and free are both in the coiio library, not in the caller's module. But you're right, we can't return a malloced string and expect the app to free() it directly.
> 
> So I think it's better if geterror() just returns a non-owning `char*` to the string still stored in OIIO (and be sure to call C++ geterror(false) so the string isn't freed!), then we'll just have to add a separate `clearerror()` on the C side. We can add clearerror on the C++ side also if it bugs you that we have a C function with no direct equivalent in C++.

I think that's not going to be threadsafe though - when geterror() returns it will release the lock and then another thread could happily write to the internal string while something on the C side is trying to read from it.

src/coiio/imageio_defines.h
anderslanglands: I don't think so I'm afraid. It's either #define or doing `extern stride_t AUTOSTRIDE;` and defining it in one (and only one) of the cpp files. Either way we might want to use OIIO_AUTOSTRIDE to guard against possible name clashes.

src/coiio/imageio.h
anderslanglands: Yeah, that make sense.

src/coiio/imageio.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    return new ImageSpec;
+}
+
+void
+ImageSpec_delete(const ImageSpec* is)
+{
+    delete is;
+}
+
```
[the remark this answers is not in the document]
anderslanglands: No apology necessary! I'll fix this. Incidentally, what version of clang-format are you using? Mine (9.0.0) barfs on the SpaceInEmptyBlock line.

src/coiio/imageio.cpp
```diff
@@ -0,0 +1,201 @@
...
+const char*
+ImageOutput_geterror(const ImageOutput* io)
+{
+    // If we have multiple ImageOutputs that we want to get errors for, we're
+    // still passing through this one function. We have to cache the string
+    // in order to return a char*, so it needs to be thread_local to avoid issues.
+    // We force clear here so that multiple calls from different ImageOutputs
+    // don't get each others' errors.
+    thread_local std::string errorstring;
```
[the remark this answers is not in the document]
anderslanglands: I'm 99% sure thread_local implies static, such that `static thread_local std::string` and `thread_local std::string` are equivalent. That said, it would't therefore hurt to add `static` so that others (and future me) don't have the same reaction.

src/coiio/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+add_library(coiio SHARED 
+    imageio.cpp
+    imageio.h
+    imageio_defines.h
+    typedesc.cpp
+    typedesc.h
```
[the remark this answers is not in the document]
anderslanglands: Heh, yeah it's already started to irritate me a little actually. I like the c-XXX.h. I don't know of many libraries structured like this. OpenTimelineIO that Scott pointed out takes the same approach I've done here and would suffer from the same project search issue. Others I've seen use .h for C and .hh or .hpp for C++ and disambiguate that way.

c-imageio.h seems fine to me. Any thoughts on the name of the library itself? Would you like libc-OpenImageIO.so? or libOpenImageIO-c.so perhaps?

src/coiio/imageio.cpp
anderslanglands: Actually this appears to be done by clang-format. It's controlled with BreakBeforeBraces which is currently set to WebKit.

lgritz: I meant at the end of the file... usually the last character of the file is a linefeed, not the brace itself.

src/coiio/imageio.cpp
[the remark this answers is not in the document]
lgritz: Yeah, you need to use clang10 for clang-format, I guess. FWIW, clang-format 11 fixes this -- ignoring unknown options instead of considering them errors.

src/coiio/CMakeLists.txt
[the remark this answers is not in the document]
lgritz: I like libOpenImageIO-C.so best, I think? Not by a huge margin. Do what you think is best and/or least surprising to people familiar with C-wrapped C++ libraries.

src/coiio/imageio.cpp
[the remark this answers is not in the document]
lgritz: Oh, you could be right. But in any case, the thread safety, where it's needed, should all be handled on the C++ side. I don't think you should ever need to impose additional thread safety on the C side.

ImageOutput is not thread safe at all, it doesn't mean anything useful for threads to simultaneously output to one file. ImageInput, though, should be thread safe (makes perfect sense for multiple threads to read simultaneously from the same open file), so if there's someplace there where we slipped up on the C++ side, we should address that separately.

src/coiio/imageio_defines.h
lgritz: If it's a macro, then let's prefix with OIIO_ (I just hate when other headers pollute the preprocessor symbols). If it's an extern, then let's use the same capitalization as we used in C++: AutoStride.

I think I'm preferring the extern solution, but I don't feel especially strongly about it.

lgritz:
> I think that's not going to be threadsafe though - when geterror() returns it will release the lock and then another thread could happily write to the internal string while something on the C side is trying to read from it.

Oh, I see your point. In C++, the retrieval and clear happen within the same lock.

But, I think that this just points out a problem with C++, too. Consider this sequence:

    thread 1 reads tile A -- error
    thread 2 reads tile B
    thread 2 calls geterror, and incorrectly retrieves the other thread's error

I think that like the global OIIO::geterror, the ImageInput needs a thread-specific error storage.

You should assume that this all works, and I'll separately fix the C++ side. It's going to be trickier than you thought. The simple thread_local is not good enough, because the error needs also to be specific to that particular ImageInput!

anderslanglands:
Hi Larry, just so I understand correctly, the only time multiple threads
are involved is “behind” the C++ API: e.g. someone calls read_image() and
that spawns multiple threads to perform the read and then joins before
returning? So from the caller’s point of view order of execution is serial
and calling methods on the same ImageInput from multiple threads on the
caller’s side is explicitly not supported?

lgritz:
> Hi Larry, just so I understand correctly, the only time multiple threads are involved is “behind” the C++ API: e.g. someone calls read_image() and that spawns multiple threads to perform the read and then joins before returning? So from the caller’s point of view order of execution is serial and calling methods on the same ImageInput from multiple threads on the caller’s side is explicitly not supported?

No, not at all. We should assume that multiple reads could be requested by different app threads simultaneously -- in fact, this is the usual behavior we should expect from the TextureSystem.

Multiple threads writing is not supported -- an ImageOutput is assumed to not be shared among threads.

lgritz:
> No, not at all. We should assume that multiple reads could be requested by different app threads simultaneously -- in fact, this is the usual behavior we should expect from the TextureSystem.

I think you have managed to highlight that although we have taken great care to ensure thread safety for multiple simultaneous read_scanlines/read_tiles/read_image calls, we still have an issue with the error reporting that slipped by our notice: there's only one error message, so retrieving the message after a failed read could get you the wrong error message, if failed reads happened in another thread between your read and when you tried to access the message. Like this:

    thread 1 reads tile A, fails
    thread 2 reads tile B, fails
    thread 1 calls geterror() gets both messages!
    thread 2 calls geterror(), gets no message

This is going to be tricky to fix. You needn't concern yourself with it in the C bindings. I will separately attack this problem and change C++ geterror() to retrieve the last error that happened in read calls made *by your thread*.

src/coiio/util.h
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+// https://github.com/OpenImageIO/oiio
+
+
+#pragma once
+
+// Utility function to pun one type to another safely via memcpy
+template<class T, class U>
+T
+bit_cast(const U& tpp)
```
lgritz, line 11: We already have this in fmath.h

anderslanglands:
> > Hi Larry, just so I understand correctly, the only time multiple threads are involved is “behind” the C++ API: e.g. someone calls read_image() and that spawns multiple threads to perform the read and then joins before returning? So from the caller’s point of view order of execution is serial and calling methods on the same ImageInput from multiple threads on the caller’s side is explicitly not supported?
> 
> No, not at all. We should assume that multiple reads could be requested by different app threads simultaneously -- in fact, this is the usual behavior we should expect from the TextureSystem.
> 
> Multiple threads writing is not supported -- an ImageOutput is assumed to not be shared among threads.

Right, OK, that's what I was safeguarding against with the thread_local. Accessing this from C is even worse than C++ because with the char* we're holding a reference into the memory owned by m_errmsg, which could easily be freed by a concurrent write to that variable, and we cannot lock:
```
[thread 1] ImageInput* ii = imageInput_open(...);
[thread 1] ImageInput_read_image(ii, ...);
[thread 1] const char* error_message = ImageInput_geterror(ii); // error_message points into ii->m_errmsg
    [thread 2] ImageInput_read_image(...);                                    // an error occurs, ImageInput sets m_errmsg, causing it to reallocate
[thread 1] printf("Error was: %s\n", error_mssage);                      // boom!
```

lgritz:
Yes. Just sweep under the rug for now, I'll work on the threading fix. When I'm done, your implementation of ImageInput_geterror(ii) can be `ii->geterror(false).c_str()`, and you can be assured that this pointer is valid until your own thread calls geterror(true) to clear it.

lgritz:
This is going to be very hard to fix in the release branch. To do it well, I need to change the data layout of ImageInput in a way that we can't backport safely. 

Can we assume that the C (and therefore Rust) bindings will be master only, that is, they will be a feature of future OIIO 2.3 but will not be supported in the existing 2.2 release? 

If making this safe to backport and work with 2.2 is a requirement, then I'm going to need a different and probably much uglier design.

lgritz:
> Yes. Just sweep under the rug for now, I'll work on the threading fix. When I'm done, your implementation of ImageInput_geterror(ii) can be `ii->geterror(false).c_str()`, and you can be assured that this pointer is valid until your own thread calls geterror(true) to clear it.

Ugh, I'm wrong here, too. C++ geterror() returns a `std::string`, not a `std::string&`. So you can't just return its c_str().

anderslanglands:
> > Yes. Just sweep under the rug for now, I'll work on the threading fix. When I'm done, your implementation of ImageInput_geterror(ii) can be `ii->geterror(false).c_str()`, and you can be assured that this pointer is valid until your own thread calls geterror(true) to clear it.
> 
> Ugh, I'm wrong here, too. C++ geterror() returns a `std::string`, not a `std::string&`. So you can't just return its c_str().

If you changed it to return a const& that would work and have the same end result for existing code wouldn't it?

In answer to your earlier question, yes I think targeting master and not attempting to support current or old releases is fine.

lgritz:
> If you changed it to return a const& that would work and have the same end result for existing code wouldn't it?

I'm afraid not. In the usual case (in C++ anyway) where we simultaneously retrieve and clear, there is nothing left to return the reference of. That's why we returned a string outright.

meshula:
Heh, I feel like I am late to a conversation that only started a half-day ago :) 

I personally very much dislike C APIs that, more than necessary, require fancy memory management around strings. 

Almost universally these days, I prefer this pattern for strings:

```
void func(char* resultStr, int resultLen); // resultLen is maximum capacity of resultStr buffer
int len = 1024;
char* mybuff = some_allocator(len);
if (mybuff)
   func(mybuff, len);
else
   mybuff[0] = '\0'; 
```
No ambiguity or rules that are easy to forget, or header comments I forgot to read about who owns the string, or concerns about whether the string memory will outlive a dynamic library's unloading.

lgritz:
@meshula Are you suggesting that the C version of geterror should take a pointer to a caller-owned block of memory and also pass a maximum length and it just gets truncated if the pending error messages are longer than the buffer?

anderslanglands:
> @meshula Are you suggesting that the C version of geterror should take a pointer to a caller-owned block of memory and also pass a maximum length and it just gets truncated if the pending error messages are longer than the buffer?

Yep, that was the third option in my original post up top^. I would lean towards that as well because of its simplicity and obviousness.

Still doesn't solve the multi-threading issue of course.

lgritz:
I'll have a PR tonight that I think completely fixes the threading problem.

src/coiio/imageio_defines.h
anderslanglands: I still feel like we ought to prepend a namespace there to be hygenic since it's essentially a global. The same goes for enums and their variants (which should also really have the enum name embedded too). So we have a few options:

1. Shouty case. This _feels_ the most c-like to me, but maybe that's because I've been spending a lot of time in CUDA and OptiX recently and this is what they do.

```
extern stride_t OIIO_AUTOSTRIDE;      //< static constexpr stride_t AutoStride;
extern OIIO_TypeDesc OIIO_TYPE_STRING; //< static constexpr TypeDesc TypeString;

enum OIIO_OpenMode {
    OIIO_OPENMODE_CREATE,
    OIIO_OPENMODE_APPENDSUBIMAGE,
    OIIO_OPENMODE_APPENDMIPLEVEL,
};
```

2. Pascal case. This is most similar to what's in C++ and is basically just removing the colons from the C++ name.
```
extern stride_t OIIOAutoStride;
extern OIIOTypeDesc OIIOTypeString;

enum OIIOOpenMode {
    OIIOOpenModeCreate,
    OIIOOpenModeAppendSubImage,
    OIIOOpenModeAppendMipLevel,
};
```

3. Pascalsnake case. Same as 2 but replacing the colons with underscores:
```
extern stride_t OIIO_AutoStride;
extern OIIO_TypeDesc OIIO_TypeString;

enum OIIO_OpenMode {
    OIIO_OpenMode_Create,
    OIIO_OpenMode_AppendSubImage,
    OIIO_OpenMode_AppendMipLevel,
};
```

src/coiio/imageio.cpp
anderslanglands: Ahh gotcha.

src/coiio/imageio_defines.h
lgritz: >  3. Pascalsnake case. Same as 2 but replacing the colons with underscores:

This one is my preference, I think it's the easiest to read and also the easiest to mentally infer the correct C names from C++ and vice versa.

lgritz:
Threading issue with ImageInput error message is address here: https://github.com/OpenImageIO/oiio/pull/2752

anderslanglands:
Updated everything based on what we've discussed here, including error handling. Don't know about you but I'm not massively fond of:
```
#include <OpenImageIO-C/c-imageio.h>
```
Do we want to put all the headers in the same directory given that all the C ones are prefixed with `c-`?

lgritz:
> Do we want to put all the headers in the same directory given that all the C ones are prefixed with `c-`?

Yeah, I like that. Less complexity.

Also, I don't know how strongly I care, but all the rest of the headers that we export to the install tend to live in src/include/OpenImageIO, whereas "internal" headers live in the same directory as the implementation cpp files that use them.

src/libOpenImageIO-C/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright 2008-present Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: BSD-3-Clause
+# https://github.com/OpenImageIO/oiio/blob/master/LICENSE.md
+
+add_library(OpenImageIO-C SHARED 
+    c-imageio.cpp
+    c-imageio.h
+    c-imageio_defines.h
```
lgritz, line 8: I *think* that you don't need to list the .h files here in `add_library`, because they are not files to be compiled.

src/libOpenImageIO-C/c-imageio.cpp
```diff
@@ -0,0 +1,257 @@
...
+const ImageSpec*
+ImageInput_spec(ImageInput* ii)
+{
+    return &ii->spec();
+}
+
+
+
+bool
+ImageInput_read_image(ImageInput* ii, int subimage, int miplevel, int chbegin,
```
lgritz, line 142: I assume this is just a first stab, and eventually we'll probably want somebody to go back and add read/write scanlines and tiles?

src/libOpenImageIO-C/c-imageio.cpp
```diff
@@ -0,0 +1,257 @@
...
+ImageOutput_write_image(ImageOutput* io, TypeDesc format, const void* data,
+                        stride_t xstride, stride_t ystride, stride_t zstride,
+                        ProgressCallback progress_callback,
+                        void* progress_callback_data)
+{
+    return io->write_image(bit_cast<TypeDesc, OIIO::TypeDesc>(format), data,
+                           xstride, ystride, zstride, progress_callback,
+                           progress_callback_data);
+}
+int
```
lgritz, line 233: blank lines go here

src/libOpenImageIO-C/c-imageio_defines.h
```diff
@@ -0,0 +1,37 @@
+// Copyright Contributors to the OpenImageIO project.
```
lgritz, line 1: c-imageio_defines.h: We can always straighten this out later, but I wonder if this should just go in c-imageio.h rather than being a separate file that has no direct analogy in the C++ API? Was there a reason to break it into a separate header?

testsuite/coiio/coiiotest.c
```diff
@@ -0,0 +1,158 @@
+#include <OpenImageIO-C/c-imageio.h>
```
lgritz, line 1: Missing copyright block

src/libOpenImageIO-C/c-imageio.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+
+void
+ImageInput_geterror(const ImageInput* ii, char* msg, int buffer_length,
+                    bool clear)
+{
+    std::string errorstring = ii->geterror(clear);
+    int length = std::min(buffer_length, (int)errorstring.size() + 1);
+    memcpy(msg, errorstring.c_str(), length);
+    msg[length - 1] = '\0';
```
lgritz, line 171: Is this more compactly expressed by a call to Strutil::safe_strcpy() ?

lgritz:
This is looking great, @anderslanglands .  I've made a couple additional notes/questions upon rereading, but they are very minor and could be addressed later if you want (knowing that this is the first of several PRs that will be necessary to add the C interface).

Is there anything left to do on this one, or should we merge this and move on to the next set of tasks?

Also, before completing the merge, I would like to know that you've made your employers aware of this effort and they are ok with it. I don't anticipate your having any problems getting that permission, but I will sleep better knowing you're totally safe on your end and there will be no surprises later.

lgritz:
@anderslanglands : I think what you've done here is really clean and although we've had a lot of back-and-forth over the details, I believe that in the process we've hashed out a lot of design elements and conventions that will apply to the whole C binding effort. I think that with this in place as a clear example of what we're going for and the conventions we are using, it should be easy for the rest of the work to be distributed among several people, each taking different classes, and if they stick to your model they should all be pretty straightforward to review and integrate.

src/libOpenImageIO-C/CMakeLists.txt
meshula: If you have the IDE folder organization option turned on in CMake, listing the H files will cause them to appear in a Headers folder in the IDE. In Xcode at least, project searches are restricted to files actually referenced in an xcodeproj file, so IDE users probably prefer files being exhaustively listed like this.

lgritz: Ah, something I was not aware of! I thought this was just needless listing that had no effect. By all means, leave it be if it serves a purpose.

Aside: I've used Macs as my primary at-home and carry-around machines for the last 12 years and I've never run xcode.

anderslanglands: I was just copying what's done in libOpenImageIO :) (CMakeLists.txt, line 87)

src/libOpenImageIO-C/c-imageio.cpp
anderslanglands: Yep sorry about that, I was having a weird issue in vscode yesterday where clang-format was barfing when I had a trailing newline. I've found the appropriate setting to add those automatically and all seems good now.

src/libOpenImageIO-C/c-imageio_defines.h
anderslanglands: At the time I did it it was to separate definitions into things that would be visible to C and to C++ to avoid redefining symbols, but that doesn't seem necessary anymore so yeah I'll collapse this into c-imageio.h

src/libOpenImageIO-C/CMakeLists.txt
lgritz: > I was just copying what's done in libOpenImageIO :) (CMakeLists.txt, line 87)

Aha, indeed you were.

src/libOpenImageIO-C/c-imageio.cpp
anderslanglands: Yes absolutely, I was just doing the bare minimum to get the test working. I'll hopefully get the rest of imageio.h bound over the weekend and submit a complete PR your Saturday night.

src/libOpenImageIO-C/c-imageio.cpp
anderslanglands: Yes it is! I really need to read through all your utility functions before rolling my own...

anderslanglands:
> This is looking great, @anderslanglands . I've made a couple additional notes/questions upon rereading, but they are very minor and could be addressed later if you want (knowing that this is the first of several PRs that will be necessary to add the C interface).
> 
> Is there anything left to do on this one, or should we merge this and move on to the next set of tasks?
> 
> Also, before completing the merge, I would like to know that you've made your employers aware of this effort and they are ok with it. I don't anticipate your having any problems getting that permission, but I will sleep better knowing you're totally safe on your end and there will be no surprises later.

Yeah I've mentioned it to them already and all good, just going to get it in writing next week so there's no chance of surprises later.

I'm just making the rest of the changes in your comments, then I'd like to see if I can get the rest of imageio.h fleshed out so we have a solid "block" done without any holes in it, then I'll submit this PR for merging. Would you like the commits in here squashed?

After that I need to finish off that OSL example...

src/libOpenImageIO-C/c-imageio_defines.h
anderslanglands: Ahh no that definitely was still necessary. Basically we can't include the c-imageio.h in c-imageio.cpp because the C interface forward declares opaque ImageSpec, ImageInput etc., and then puns them in c-imageio.cpp by importing OIIO::ImageInput etc into the TU namespace. If we include the opaque definitions then we get lots of compiler errors because I'm essentially defining ImageInput twice. 

I'm halfway through doing it "properly" and removing the pointer punning and it's turning into a bit of a casting nightmare:
```
ImageInput*
ImageInput_open(const char* filename, const ImageSpec* config, void* _ioproxy)
{
    auto ii = OIIO::ImageInput::open(
        filename, reinterpret_cast<const OIIO::ImageSpec*>(config), nullptr);
    return reinterpret_cast<ImageInput*>(ii.release());
}
```

I think a better solution here might be to return to punning but change the C type names to prepend `OIIO_`, which to be honest we should probably do anyway in the name of hygeine. So the above would then become:
```
using OIIO_ImageInput = OIIO::ImageInput;
using OIIO_ImageSpec = OIIO::ImageSpec;

OIIO_ImageInput*
ImageInput_open(const char* filename, const OIIO_ImageSpec* config, void* _ioproxy)
{
    return OIIO::ImageInput::open(filename, config, nullptr).release();
}
```
That then begs the question of should we be prepending `OIIO_` to all the function names as well, which again my gut tells me yes, even though it's getting a little verbose.

lgritz: Yeah, that probably the right thing to do. 

It's easy to hate on C++, and I can complain plenty, but it's been interesting seeing this process and being reminded of all the parts that I would be very frustrated by missing if I were working in C again. Namespaces and function overloading by argument is definitely something I really like. I really wish the C++ language itself dictated the name mangling, so that you could easily construct a C linkage entry point name from a C++ function. That would dramatically simplify the amount of new code required.

lgritz:
OK, just let me when you feel it's ready to merge and you're squared away at work. As far as I'm concerned, it's ready to merge at any time and at worst we can return to touch things up or add the missing methods.

src/libOpenImageIO-C/c-imageio_defines.h
anderslanglands: Namespaces I’m with you on for sure. Not so convinced about function overloading any more. Rust doesn’t allow it either and despite some initial frustration I’ve actually come to really like it. I find it much easier to read as you don’t have to mentally type-check every function call to know what it’s doing.

lgritz:
For anybody following the discussion in this PR, at your leisure please also read #2754, I'd be eager to hear anyone's thoughts on it.

anderslanglands:
The latest commits make the naming changes we discussed, and finish off binding everything in imageio.h and dependent types. Specifically the following should now be complete bar a couple of functions (marked `TODO: (AL)`): 
```
ROI
ImageSpec
ImageInput
ImageOutput
ParamValue
DeepData
```

Probably the only contentious thing I've added is short pointer conversion functions to make the code readable instead of a sea of `reinterpret_cast<OIIO::ImageInput*>()` everywhere, as well as a macro to make defining them easier.

I'll need to give some though to adding some tests as well. Currently it's still just the same simple round-trip test, so tests for the rest of the functionality will need adding at some point.

alexxbb:
I've been working on Rust bindings for Houdini Engine and I think it might be a source of inspiration, for some of its simplicity of design, for example this is how they handle retrieving errors(and strings in general): [link](https://www.sidefx.com/docs/hengine/_h_a_p_i__fundamentals.html#HAPI_Fundamentals_ReturnCodesErrorStrings)

So the host does malloc/free in their case and this is easy to wrap in Rust too:
```rust
// allocate
let vec = Vec::<u8>::with_capacity(cap);
let cstr = CString::from_vec_unchecked(vec);
// pass this to C
let char_ptr: *const c_char = cst.into_raw();
// take ownership back
let result = CString::from_raw(char_ptr);
```

lgritz:
There are a number of errors popping up on the Windows CI tests. 
I think that this is probably about [this definition](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/export.h#L62) and how the right dllimport/dllexport is not set correctly for the new library.

anderslanglands:
Ah so it looks like I need to "#define OpenImageIO_exports" before
including export.h? Also looks like I'm missing OIIO_API on some of the
function definitions. I'll update that now.

lgritz:
I think it's trickier than that.

On Windows (grrr...) you need `__declspec(dllexport)` on the declarations in the header when you're compiling the code *for the library*, and `__declspec(dllimport)` on the declarations in the header when you're compiling an app that *uses* the library.

So the gist is that when you are building the openimageio-c files, its own symbols are exported but the symbols it needs from libOpenImageIO are imported. Therefore you can't just use OIIO_API.

I think what you need to do is add this to exports.h:

```
#if defined(OpenImageIO_C_EXPORTS)
#    define OIIOC_API OIIO_EXPORT
#else
#    define OIIOC_API OIIO_IMPORT
#endif
```

And then use OIIOC_API for all the C library public declarations.

Remove the line in libOpenImageIO-C/CMakeLists.txt -- don't have it define OpenImageIO_EXPORTS -- either define OpenImageIO_C_EXPORTS instead, or maybe CMake already does this automatically (though I'm not quite sure how it will mangle the name if it contains a dash `-` which is allowed in a library name but not in a preprocessor symbol, I'm only guessing that it will convert it to an underscore, but I'm not positive).

anderslanglands:
Ahhh gotcha. God I hate windows. I'll give this a go today.

lgritz:
CLA received. Now we're just waiting for fixes to that Windows import/export problem and with that we're ready to merge.

(The test failure on the clang-format verification looks unrelated to this patch, and I think has already been fixed. If you rebase this on top of the current master, I think that will go away.)

anderslanglands:
Looks like this one's good to go! Thanks for the help with the windows stuff.

lgritz (approved):

lgritz:
Merged. Congrats!

aghiles:
Great work Anders.

PR #2749, opened 2020-10-22
Add support for saving progressive jpegs

SRHMorris:
## Description

Allows saving progressive jpegs by adding the attribute "jpeg:progressive", an int which if non-zero will tell libjpeg to create a progressive file. 

## Tests

TODO

## Checklist:

### DIFF
--- a/src/jpeg.imageio/jpegoutput.cpp
+++ b/src/jpeg.imageio/jpegoutput.cpp
@@ -186,8 +186,7 @@ JpgOutput::open(const std::string& name, const ImageSpec& newspec,
         DBG std::cout << "out open: set_colorspace\n";
 
         // Save as a progressive jpeg if requested by the user
-        ParamValue* progressive = m_spec.find_attribute("jpeg:progressive", TypeDesc::INT);
-        if (progressive && *(int*)progressive->data() != 0) {
+        if (m_spec.get_int_attribute("jpeg:progressive")) {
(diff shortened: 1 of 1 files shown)


src/jpeg.imageio/jpegoutput.cpp
```diff
@@ -185,6 +185,12 @@ JpgOutput::open(const std::string& name, const ImageSpec& newspec,
         }
         DBG std::cout << "out open: set_colorspace\n";
 
+        // Save as a progressive jpeg if requested by the user
+        ParamValue* progressive = m_spec.find_attribute("jpeg:progressive", TypeDesc::INT);
+        if (progressive && *(int*)progressive->data() != 0) {
```
lgritz, line 190: There's a new idiom that is a more compact way to express this!

    if (m_spec.get_int_attribute("jpeg:progressive")) {

(The value it returns if the attrib doesn't exist is 0, so this works out perfectly in this case. Though you can also pass a second argument that is the default to use if there is no attribute.)

Or, since 2.1, this also works (which style you prefer might perhaps be a matter of personal taste):

     if (m_spec["jpeg:progressive"].get<int>()) {

SRHMorris: Nice, thank you. I went with the first option as it looks a little cleaner to me.

lgritz (approved):
LGTM!

I'll merge after the CI tests complete.

PR #2750, opened 2020-10-23
Work around GHCI Mac problem with homebrew openssl

lgritz:


lgritz:
CI tests were failing as of a couple days ago. It was a problem with a broken openssl package on the Homebrew side. This fixes by avoiding the broken package.

PR #2752, opened 2020-10-23
Fix thread safety in ImageInput error reporting

lgritz:
Although we were careful to make sure that the new versions of
ImageInput `read_scanlines`/`read_tiles`/`read_image` methods were
thread-safe when called concurrently, one thing that slipped through
the cracks was error reporting. A failed read (or other) call would
return an error flag and use error() to set the error message, which
could be retrieved via geterror()/has_error().

Except the flaw was that the error itself was a single std::string in
the class. It was guarded by a mutex, so at least the threads wouldn't
clobber each other over access to the string. But it means that the
following call sequence would do the wrong thing:

    thread 1:   read_tile A  --> fails
    thread 2:   read_tile B  --> fails
    thread 1:   geterror()   --> gets both errors appended, oops!
    thread 2:   geterror()   --> gets an empty error message, oops!

Why hadn't we noticed until now? Because read failures of individual
scanlines or tiles are rare -- usually the failure or discovery of a
missing or corrupt file is in open().  So I guess it's not often that two
concurrent reads would both have just the right timing to have the
error retrieval problem described above.

There is no easy fix, so a bit of a bigger, ABI-breaking, change was
necessary to make this work properly once and for all:

* ImageInput and ImageOutput error messages are now thread-specific. That
  is, geterror() retrieves errors generated by method calls made by the
  same thread.

* ImageInput and ImgageOutput now use a PIMPL idiom to store some data
  behind an opaque pointer. This is good -- it means that the details
  are hidden from the public API, and can use classes we don't wish to
  expose (which we will do here with boost::thread_specific_pointer) as
  well as allow us to change what's behind the pointer without breaking
  the ABI.

* The error message uses a thread_specific_pointer.

* We therefore removed the m_errormessage member that was unfortunately
  in the public class. Also removed the mutex and the m_threads, which
  can now be safely tucked behind the PIMPL.

* With the mutex hidden from view, we now re-jigger references to it
  to use the lock/unlock methods we had already. In fact, instead of
  using lock_guard<mutex>, we can use lock_guard<ImageInput> because
  the presence of lock/unlock methods make ImageInput (and ImageOutput)
  themselves compliant with the BasicLockable concept.

Note that this changes the data layout of ImageInput and ImageOutput,
and therefore is an ABI break. So this is a change meant for master
only (future OIIO 2.3+), and will not be backported to the 2.2 release
branch.

### DIFF
--- a/src/bmp.imageio/bmpinput.cpp
+++ b/src/bmp.imageio/bmpinput.cpp
@@ -126,7 +126,7 @@ bool
 BmpInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
                                void* data)
 {
-    lock_guard lock(m_mutex);
+    lock_guard lock(*this);
     if (!seek_subimage(subimage, miplevel))
         return false;
 
--- a/src/cineon.imageio/cineoninput.cpp
+++ b/src/cineon.imageio/cineoninput.cpp
@@ -369,7 +369,7 @@ bool
 CineonInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
                                   void* data)
 {
-    lock_guard lock(m_mutex);
+    lock_guard lock(*this);
     if (!seek_subimage(subimage, miplevel))
         return false;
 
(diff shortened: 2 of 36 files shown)


lgritz:
@anderslanglands does this look reasonable to you (both on its own, and in terms of getting this code into shape to make the C bindings safe)?

anderslanglands:
Yes it does.

Issue #2744, opened 2020-10-17
[FEATURE REQUEST] Separate per-channel buffers


lgritz:
All of the basic ImageInput/ImageOutput functionality assumes that the in-memory buffer has "contiguous" pixels with interleaved channels (RGB RGB RGB...). It is currently impossible to read channels into a buffer with a "separate" configuration (RRRRR GGGGG BBBBB), nor to generally put each channel into a different buffer entirely. There is no combination of strides that will do this properly in a single read or write. You could read into a temporary contiguous buffer and then make nchannels calls to OIIO::convert_image (or parallel_convert_image) to copy each channel out into a separate permanent area.

That is certainly a relatively rare desire (evidenced by how few people have requested this over the years), but not an unreasonable one for some applications.

This issue is a placeholder for this work. I don't have time to tackle it right now, but I would welcome a PR that adds this capability in a clean way. Below I will sketch out how I think it would work and even break it into pieces which can be done one at a time.

- [ ] The first step is to nail down the user API by adding ImageInput/ImageOutput API calls declarations to imageio.h for this mode. I'll describe what I would recommend for ImageInput::read_scanlines, and let you extrapolate that into read_tiles, read_image, and the equivalent methods for ImageOutput.

     Note that there is a simple read_scanline() which reads only one scanline, all channels. And there is also a more complicated read_scanlines() that has many more parameters and has the ability to read any number of contiguous scanlines, as well as a range of channels which may be a contiguous subset of the total channel set. I recommend that we only add new versions of read_scanlines(), keep the simple read_scanline super simple. Same with read_tile vs read_tiles.

    So I would add a new method, `read_scanlines_separate()` which mimics read_scanlines, except that instead of parameter `void *data`, it takes `cspan<void*> data`.  The `data` span needs to have either size=1 or size = chend-chbegin, or it's an error. That is, a single data pointer that implies that the separate channel planes should be adjacent to each other in memory, or one data pointer for each channel which could point to totally arbitrary areas. Similarly, the `format` and `stride` parameters should also be cspan's, again where each one of which could take a single scalar value (meaning it's the same for each channel plane) but optionally allowing the span to contain (chend - chbegin) elements, giving potentially a different format or set of strides for each channel.

- [ ] Similarly, there will need to be an additional virtual method `read_native_scanlines_separate` which is like `read_native_scanlines` except with the data pointer now a cspan rather than a scalar. (Same for tiles.)

[...]

PR #2756, opened 2020-10-25
Docs: illustrate reading image data into separate per-channel buffers

JanHett:
Related to #2744

## Description

The discussion around #2744 came to the conclusion that the status quo already allows relatively efficient reading of data into separate per-channel buffers. This PR adds a section to the docs to make this approach more discoverable.

## Tests

I didn't add a test, but since adding this to the docs makes this kind of use of the API "official" I maybe should...?

## Checklist:

### DIFF
--- a/src/doc/imageinput.rst
+++ b/src/doc/imageinput.rst
@@ -299,6 +299,37 @@ Please consult Section :ref:`ImageInput Class Reference` for detailed
 descriptions of the stride parameters to each ``read`` function.
 
 
+Reading channels to separate buffers
+------------------------------------
+
+While specifying data strides allows writing entire pixels to buffers with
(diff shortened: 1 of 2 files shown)


lgritz:
This is great, thanks! So few people alter the docs, even when they have a PR that adds features to the code. I really appreciate your effort to explain this situation in the docs.

By the way, the .tex changes are unnecessary, they aren't used any more. All the docs are based on the rst files now. I probably should have deleted the .tex a long time ago, but I was keeping them around so I could refer to them as I cleaned up the last of the cross references and whatnot that weren't quite right in the new docs. But we haven't been keeping the tex up to date.

lgritz (approved):

PR #2757, opened 2020-10-26
Fix reading TIFF files with "separate" planarconfig

lgritz:
There is a big bug lurking in how we read TIFF files with "separate"
planarconfig (RRRGGGBBB, versus "contig" RGBRGBRGB). I knew that
libtiff when reading scanline files doesn't support true "random
access" to scanline for most compression methods -- because you can't
seek to the start of a scanline in the middle of a compressed "strip,"
you have to read whole strips, or read individual scanlines
consecutively. But what somehow I've misunderstood for years (if not
decades) is that for "separate" planarconfig, this means that you
can't jump from scanline i in plane 0 to scanline i in plane 1! You
have to read all the scanlines (at least within one strip) in a single
plane before moving to the next plane.

This means that we get all sorts of read failures when reading TIFF
files meeting all of these conditions:

* "separate" planarconfig (these are rare in general)
* Scanline-oriented (tiles are fine, because they're read whole)
* More than one channel (with one channel, separate and contig are
  basically the same)
* Certain compression codecs (turns out the separate+scanline combo
  tests in the libtiff example suite are uncompressed, which is how
  our testsuite failed to catch this).
* rowsperstrip > 1 (if rps == 1, every scanline is a new strip, and
  that worked)
* written by an app that doesn't use OpenImageIO (because OIIO TIFF
  output forced rps=1 for separate planarconfig)

Things were very broken. There is a comment in tiffoutput.cpp that says

    // I can only seem to make separate planarconfig work when
    // rowsperstrip is 1.

and forces rps=1 for separate planarconfig output. So at some point
many years ago I realized that something odd was happening here, but I
apparently I thought it was a bug in libtiff when writing with rps > 1.
Probably we were writing correctly all along, and the real bug was in
the reading.  This also explains why we had mostly not seen this
problem before, which is because all OIIO-written files avoided
producing tiff files that fell into the failures case.

This patch mostly fixes it and also makes a number of performance
enhancements:

* Fix misunderstanding in read_scanlines where it was loose with thread
  safety in retrieving the rows per strip.

* In read_scanlines when deciding the "chunk size" to limit temp
  buffer size, and in read_image when deciding how many scanlines to
  read at a time via read_scanlines, these computations weren't the
  same, leaving slivers to be read alone instead of staying aligned
  with strips. Make those both compute the same way, so from read_image
  on down, we end up reading chunks that are always aligned with
  strips.

* TIFF read_scanlines used to try to read whole strips, except for
  reading individual scanlines that didn't comprise a whole strip
  length. Modify it to recognize that the one exception is the last
  strip in the file, in which the strip is shorter than usual if the
  total number of scanlines is not an even multiple of rps. This
  ensures that the last strip is read as a strip, not as individual
  scanlines.

* Imagecache autotile behavior is modified to force the pseudo-tile
  size to be a multiple of the strip length. Again, that keeps the
  resulting reads to always be in units of strips, which is much more
  efficient and is the safe way to seek and read separate
  planarconfig.  Also, always turn on autotile for any tiff file with
  rowsperstrip > 1, because it's already organized that way in the
  file. This is not only more efficient generally, but also makes it
  so that ImageCache will only do reads aligned safely to strip
  boundaries.

This is not a complete solution, in the sense that an app making
individual read_scanline calls for solo scanlines, or read_scanlines()
that is not aligned to strips, can still have read failures for TIFF
files that are just the right combination of separate planarconfig +
certain compressions + scanline order.

However, with this patch, the following common cases will ALWAYS work
correctly now: (a) read_image or read_scanlines of the whole image
size; (b) any reads mediated by ImageCache or TextureSystem; (c) all
images read by ImageBuf (which always use either (a) or (b)); (d) all
operations of oiiotool (which use ImageBuf).

After this patch is merged, solving the common cases, I'll return to
the topic to tackle the remaining issue of broken reads for these
files when doing individual ImageInput::read_scanline reads or
read_scanlines with ranges not aligned to strips. To make those work,
I'll need some kind of ugly internal strip buffering.

I'm going to leave the "force rps = 1 when writing planar separate tiff"
as it is, because even after this set of fixes, software using old
versions of OIIO will not be able to correctly read rps>1 planar files,
so we wouldn't want to be inadvertently writing them, at least not
until all the pre-fix OIIO releases are unlikely to be used by
anyone. Not writing planar files with rps>1 is not much of a
sacrifice, considering how rarely planar files are used at all.


JanHett:
I just tested this on a few files that crashed the previous version and they open perfectly with this fix.

Of course I'm not nearly familiar enough with the code itself to make an authoritative statement about it, but FWIW I couldn't see anything negative that stood out to me there either.

In short, all looks good to me and many thanks for having a fix in hand in such a short time!

PR #2769, opened 2020-11-07
Servantf/libraw flip option

servantf:
Add an option "raw:user_flip" to force LibRaw to use a specific flip value. 

- LibRaw already have this input field available in imgdata.params.user_flip
- It overrides the "flip" field found in some raw images such as .CR2 files 

Why ?

Libraw will read the "flip" field and rotate the image buffer. As opposite to the orientation EXIF tag, this will change the result buffer. 
Other (all except libraw ?) raw reader such as Adobe DNG converter, will ignore this field and will not propagate this information in the result buffer.

This means that if you take a CR2 and read it with libraw, or converting it to DNG then read it with libraw will not give you the same image. Being able to disable the flip value is a way to make sure the behavior is the same.


lgritz (approved):
LGTM, thanks!

Can you please edit to fix the broken clang-format test? Just click on the details of the broken test, look at the end, and the diff will tell you exactly what to change to confirm to the formatting rules.

servantf:
I think it's done

lgritz:
Works!

The failure on the "bleeding edge" case is unrelated to your change. I'm looking at that separately.

PR #2777, opened 2020-11-26
Support UTF-8 filenames for GIF

### LINKED ISSUE #2775: [BUG] GIFs with non ascii filenames fail to open on windows
**Describe the bug**
Calling `OIIO::ImageInput::open(filename);` fails for gif images with non ascii filenames on windows. Other image plugins appear to work fine though I've not tested all of them, presumably because they use the OIIO::Filesystem stuff.

I assume this works correctly on any os that uses utf8. Renaming the file so that it contains only ascii characters works as expected.

Github will have renamed the file I guess, so just rename it to something like "test×.gif".

testgif×

**To Reproduce**
1. call OIIO::ImageInput::open() on the attached image, which has a non ascii '×' character in the filename
2. OIIO::geterror() returns "Failed to open given file"

**Platform information:**
 - OIIO branch/version: 2.2.7.0
 - OS: windows 10
 - C++ compiler: msvc

lgritz:
Fixes #2775

### DIFF
--- a/src/gif.imageio/gif.h
+++ b/src/gif.imageio/gif.h
@@ -767,7 +767,11 @@ struct GifWriter
 bool GifBegin( GifWriter* writer, const char* filename, uint32_t width, uint32_t height, uint32_t delay, int32_t bitDepth = 8, bool dither = false )
 {
     (void)bitDepth; (void)dither; // Mute "Unused argument" warnings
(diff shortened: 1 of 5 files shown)


lgritz:
Original reporter (#2775) says that this patch fixes the problem. Merging.

PR #2780, opened 2020-12-02
Better error reporting for IOFile, esp. for OpenEXR writes

lgritz:
Filesystem::IOFile constructor now upon an error opening the file will
set the error message appropriately.

Move IOProxy::error() implementation from inline to be in
filesystem.cpp to hide its details, use a static mutex to make error
set and retrieval be thread-safe (we presume that the rare chance that
multiple threads will be setting or getting IOProxy errors
simultaneously are rare enough that it's not a thread performance
issue, and if the user is getting file I/O errors, a bit of minor
locking is the least of their worries).

Better checking of these errors in OpenEXR input and output open
operations.  Hopefully this means that exr file I/O errors will
percolate up more helpful error messages.

### DIFF
--- a/src/include/OpenImageIO/filesystem.h
+++ b/src/include/OpenImageIO/filesystem.h
@@ -368,22 +368,15 @@ public:
                      (origin == SEEK_END ? offset+int64_t(size()) : 0));
     }
 
-#if OIIO_VERSION >= 20101
     #define OIIO_IOPROXY_HAS_ERROR 1
-    std::string error() const { return m_error; }
-    void error(string_view e) { m_error = e; }
-#else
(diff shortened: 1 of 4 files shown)


fabiencastan:
I'm using vcpkg to build oiio on windows, but the version available on vcpkg is still 2.1.16.0.
I tried to update but there are conflicts with 2 patches:
https://github.com/microsoft/vcpkg/blob/master/ports/openimageio/fix-dependency.patch
https://github.com/microsoft/vcpkg/blob/master/ports/openimageio/fix-tools-path.patch

It would be good for the long term to see if the vcpkg patches could be integrated in oiio.

PR #2783, opened 2020-12-07
Fix webp compile break on 32 bit systems

lgritz:
The WebPData::size is a size_t, on 32 bit systems it will balk at our
putting a uint64_t in it. Downcast explicitly so the warning doesn't
flag us.

Also some changes to ensure adequate testing of webp in CI. (Even though
we only CI test 64 bit systems, so this was not related to the specific
build break being fixed... but an excuse to make sure we are hitting it
a reasonable amount in CI.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
           CMAKE_CXX_STANDARD: 14
           PYTHON_VERSION: 3.7
           USE_SIMD: avx
+          WEBP_VERSION: v1.1.0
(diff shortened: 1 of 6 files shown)


lgritz:
@hobbes1069

hobbes1069:
RHEL/CentOS has pretty much abandoned 32bit systems. Fedora still builds for i686 and armv7hl. I'm not seeing any breakage but there are some unrelated warnings if you want to take a look:

https://kojipkgs.fedoraproject.org//packages/OpenImageIO/2.2.9.0/1.fc33/data/logs/i686/build.log

PR #2785, opened 2020-12-12
Removed some duplicates in maketx.rst doc

Xelt:
Removed some duplicates in documentation


lgritz:
Strange! Not quite sure how that happened.

Good catch, thanks for the fix.

Xelt:
How can we make the same behaviour of maketx --constant-color-detect with oiiotool? -otex:constant_color_detect=1?

lgritz:
Yes, I think that works, though now I see that it's not in the documentation. I'll add it.

This also works, and is equivalent:

    oiiotool in.tif -attrib "maketx:constant_color_detect" 1 -otex out.tx

PR #2788, opened 2020-12-16
Add ICC profile support for PSD input

SRHMorris:
## Description

Loads the icc profile from the file and adds it to the common specs.

## Checklist:

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -1173,7 +1173,8 @@ PSDInput::load_resource_1039(uint32_t length)
     if (!m_file.read(icc_buf.get(), length))
         return false;
 
-    common_attribute("ICCProfile", TypeDesc(TypeDesc::UINT8, length), icc_buf.get());
+    TypeDesc type(TypeDesc::UINT8, length);
+    common_attribute("ICCProfile", type, icc_buf.get());
     return true;
 }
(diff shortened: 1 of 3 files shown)


lgritz (changes requested):
This is great, thanks!

The CI tests are failing... but for the good reason! -- ICC profiles are being read now, and so the reference output on the "psd" testsuite entry is no longer matching. Can you please update the reference output?

The trick is that there are TWO reference output files -- testsuite/psd/ref/out.txt and out-alt.txt. They differ slightly because the exact values in the thumbnail seem to depend on libjpeg version, so there are two possible valid outputs! You'll have to update both, but only the line that changes because of ICC, please don't end up accidentally making those files the same in the ways that they are currently (and supposed to be) different. Sorry, that's a little confusing, I hope you know what I'm talking about.

There is an extremely minor code formatting fix needed that you can also see from the "clang-format" CI test.

SRHMorris:
Yeah I'd noticed that certain jpegs looked slightly different between libjpeg and libjpeg-turbo.

Don't think the last test is me!

lgritz:
No, that Mac failure is not you, something changed on the GitHub Mac instances (or on homebrew packaging), and I have a separate PR to address that. It looks like you've got everything passing now.

lgritz (approved):

PR #2801, opened 2020-12-30
Fix crash parsing OpenEXR header.

boolFighter:
## Description

Fix for a few issues parsing rational values with large denominator in OpenEXR header.

Incorrect use of array index can cause divide by zero crash (observed consistently on Windows with repro image).

Conditional comparison was assigning 1 or 0 to int f declaration instead of the actual gcd value.

Changed gcd type parameter to long long to avoid uint->long int overflow on windows (long int is only 4-bytes on MSVC).

## Checklist:

### DIFF
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -825,8 +825,7 @@ OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in,
                 r[0] = n;
                 r[1] = static_cast<int>(d);
                 spec.attribute(oname, TypeRational, r);
-            }
-            else {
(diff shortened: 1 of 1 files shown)


lgritz:
Hi, thanks for this! Yes, looks like a bug, obviously a confusion between the `Rational*` and the individual integers.

LGTM, but can you add the one-line fix that brings it into formatting compliance (see the bottom of the output of https://github.com/OpenImageIO/oiio/pull/2801/checks?check_run_id=1627391413)? Thanks.

lgritz:
I will backport to 2.2 and have it incorporated into the Jan 1 release.

PR #2805, opened 2021-01-04
CI: fix Windows python tests

lgritz:
* Change to finding a consistent Python version (3.7)

* Install numpy

* On Windows, convert crlf to lf before diffing reference output, so we
  don't get mismatches just because of LF vs CRLF.

* Update ref output on testsuite/python-imageinput

* Fix windows detection in runtest.py oiio_relpath

* Fix some semicolon problems that were making bash command lines conused

Along for the ride:

* Have ci-build-and-test.bash send OIIO_CTEST_FLAGS as extra args to ctest

* Remove some needless verbosity from the dependency building scripts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,10 +322,9 @@ jobs:
       - name: all
         shell: bash
         env:
-          PYTHON_VERSION: 3.6
+          PYTHON_VERSION: 3.7
           CMAKE_GENERATOR: "Visual Studio 16 2019"
           OPENEXR_VERSION: v2.4.1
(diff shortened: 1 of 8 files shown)


lgritz:
@debaetsd

lgritz:
I set out here just to make the Python tests work, but the crlf->lf conversion on all text outputs seems to have brought several other tests into conformance as well.

lgritz:
@debaetsd does this look ok to you?

debaetsd:
Better late then never, but no I did not have comments on this. LGTM

PR #2809, opened 2021-01-04
Fix error decoding Apple HEIF files

### LINKED ISSUE #2794: [BUG] Unable to open Apple HEIC images
**Describe the bug**
With OpenImageIO compiled with libheif 1.9.1 I am able to open HEIC/HEIF images. However, Apple HEIC files from various iPhones and iPads do not load.

**To Reproduce**
1. Build OIIO with libheif
2. Call `ImageInput::open` with one of the examples below
3. ImageInput will throw an unknown exception

**Expected behavior**
I expect it to handle the HEIC files like other non-Apple HEIC files. For non-Apple samples, I've used files from [this repository](https://github.com/tigranbs/test-heic-images) which do load correctly.

**Evidence**
The images that cause these errors come from iPhones and iPads. I've attached one of each [here](https://github.com/OpenImageIO/oiio/files/5736369/Examples.zip).

One thing I've noticed is that non-loading HEIC files have an additional brand `MiPr`, which according to my searching is for "Multi-Image Application format brand for progressive decoding and rendering". Maybe this is not supported?

For reference, the code throwing the error is this:
```cpp
OIIO::ImageInput::unique_ptr in_p = nullptr;
OIIO::ImageSpec config;
setupRawSpec(config);  // This is a series of config.attribute calls to set stuff like "raw:use_camera_wb", etc

try {
    in_p = OIIO::ImageInput::open(fileName.toStdString(), &config);
} catch (std::exception& ex) {
    qCritical() << "ImageInput::open Exception:" << ex.what();
    // ...
} catch (...) {
    qCritical() << "ImageInput::open Unknown Exception";  // This is what gets hit when opening Apple HEIC files
    // ...
}
```

**Platform information:**
 - OIIO branch/version: 2.2.9.0
 - OS: Windows and Mac
 - C++ compiler: MSVC 2019 on Windows, XCode 12.1 on Mac
 - Any non-default build flags when you build OIIO: `USE_LIBHEIF=1` and `USE_LIBRAW=1` I believe are the only non-defaults.

hjmallon:
## Description

* 0 size metadata causes an issue in the libheif cxx wrapper
* fixes #2794

## Tests

I have tested with the samples from #2794 

## Checklist:

Upstream fix: https://github.com/strukturag/libheif/pull/415

### DIFF
--- a/src/heif.imageio/heifinput.cpp
+++ b/src/heif.imageio/heifinput.cpp
@@ -200,7 +200,17 @@ HeifInput::seek_subimage(int subimage, int miplevel)
     auto meta_ids = m_ihandle.get_list_of_metadata_block_IDs();
     // std::cout << "nmeta? " << meta_ids.size() << "\n";
     for (auto m : meta_ids) {
-        auto metacontents = m_ihandle.get_metadata(m);
+        std::vector<uint8_t> metacontents;
+        try {
+            metacontents = m_ihandle.get_metadata(m);
+        } catch (const heif::Error& err) {
+            if (err.get_code() == heif_error_Usage_error
+                && err.get_subcode() == heif_suberror_Null_pointer_argument) {
+                // a bug in heif_cxx.h means a 0 byte metadata causes a null
+                // ptr error code, which we ignore
+                continue;
+            }
+        }
         if (Strutil::iequals(m_ihandle.get_metadata_type(m), "Exif")
             && metacontents.size() >= 10) {
             cspan<uint8_t> s(&metacontents[10], metacontents.size() - 10);


lgritz (commented):
LGTM! I will merge when the CI tests finish (assuming they don't turn up anything unexpected).

lgritz (approved):

lgritz:
What leads to this 0 byte metadata? Is there any metadata expected to have 0 size, or is that an oddity of the file itself?

hjmallon:
> What leads to this 0 byte metadata? Is there any metadata expected to have 0 size, or is that an oddity of the file itself?

I honestly have no idea. The example code in libheif checks for 0 size before copying the data so works fine, but that check isn't in the cxx bindings. Maybe the original issue creator @AdamMainsTL could say?

AdamMainsTL:
Thank you very much for the quick fix! I'll try to update soon.

For the images as far as I'm aware they were just taken straight from an iPhone (12 pro max), another iPhone (non-12) that I didn't upload, and one from an iPad. I'm asking the person who took them all whether they have some privacy setting or something that would cause this as I didn't strip anything on my end. The entire chain of custody for the images would have been Apple device -> Slack -> attempt to load, so it's also possible Slack removed data.

AdamMainsTL:
Got an update. The full chain was actually iPhone -> Airdrop to Mac -> Slack. It seems like Airdrop is the weak link according to this thread: https://discussions.apple.com/thread/251040579

lgritz:
Interesting, thanks for the investigation.

PR #2811, opened 2021-01-04
Enabled AVIF decoding with libheif

hjmallon:
## Description

Requires libheif built with libaom or dav1d/rav1e. Changes to the code just required adding '.avif' to the extensions.

Also updated as much of the docs as I could find to mention AVIF.

## Tests

I have added a testsuite file, but it only uses the info command like the other heic file.

## Checklist:


INSTALL.md
```diff
@@ -52,6 +52,8 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * If you want support for HEIF/HEIC images:
      * libheif >= 1.3 (tested through 1.9; older versions may also work, we
        haven't tested)
+ * If you want support for AVIF images:
+     * libheif >= 1.7 (tested through 1.10)
```
lgritz, line 56: Can you combine?

     * If you want support for HEIF/HEIC or AVIF images:
         * libheif >= 1.3 (1.7 required for AVIF support, tested through 1.10)

src/cmake/externalpackages.cmake
```diff
@@ -163,7 +163,7 @@ checked_find_package (GIF
                       VERSION_MIN 4
                       RECOMMEND_MIN 5.0
                       RECOMMEND_MIN_REASON "for stability and thread safety")
-checked_find_package (Libheif VERSION_MIN 1.3)  # For HEIF/HEIC format
+checked_find_package (Libheif VERSION_MIN 1.3)  # For HEIF/HEIC/AVIF formats
```
lgritz, line 166: # For HEIF/HEIC/AVIF formats
    checked_find_package (Libheif VERSION_MIN 1.3
                          RECOMMENDED_MIN 1.7
                          RECOMMENDED_MIN_REASON "for AVIF support")

src/doc/builtinplugins.rst
```diff
@@ -568,17 +568,18 @@ preferred except when legacy file access is required.
 
 .. _sec-bundledplugins-heif:
 
-HEIF/HEIC
+HEIF/HEIC/AVIF
```
lgritz, line 571: I really appreciate your editing the docs, thanks!

testsuite/heif/run.py
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env python
 
 imagedir = "ref/"
-files = [ "IMG_7702_small.heic" ]
+files = [ "IMG_7702_small.heic", "Chimera-AV1-8bit-162.avif" ]
 for f in files:
     command = command + info_command (os.path.join(imagedir, f))
+
+# AVIF file is expected to fail with libheif without AV1 decoder
+failureok = 1
```
lgritz, line 9: Let's remove this (failureok=1 masks any real problems, so what's the point of the test?) and just ensure that CI and any developers who run tests and care about this will have libheif >= 1.7.

lgritz (changes requested):

testsuite/heif/run.py
hjmallon: brew currently ships libheif without AV1 enabled (I have sent a PR for that too) so the macOS tests fail. CI/developers will have to make sure that they have libheif >1.7 with AV1 enabled. Is that OK?

lgritz: I think that's fine.

FWIW, Brew currently ships libheif 1.10 and I found that it's utterly broken on MacOS when writing heif file. I reported the problem and pointed them to the exact commit where it started failing. Awaiting a resolution. But one more problem with the current Brew default build of libheif is a drop in the bucket at the moment.

lgritz (approved):

lgritz:
@hjmallon You were totally correct with the failureok trick that I asked you to remove. I realize now that I was confused, thinking that you were marking the entire test as "ok to fail" (and thus meaningless), but in fact I had forgotten that what that flag actually indicates is that it's ok for any individual command within the test to exit with an error return code, but the test overall will pass or fail based on whether the outputs matched any of the available references.

So sorry to have steered you wrong.

Would you mind restoring what you were originally trying to do with the test, so that the heif test will pass whether or not the avif support is enabled (as long as the output matches either the "heic+avif" or "heic only" reference outputs)?

hjmallon:
See https://github.com/OpenImageIO/oiio/pull/2818

PR #2812, opened 2021-01-05
Enabled AVIF encoding with libheif (#2811)

hjmallon:
## Description

Requires libheif built with libaom or dav1d/rav1e. I have tested this, but really not very extensively. It shouldn't cause any regressions though.

## Checklist:


lgritz (approved):
LGTM!

(Don't worry about the CI failures, they are all for reasons unrelated to this patch.)

PR #2814, opened 2021-01-05
Fix build break for older libheif

lgritz:
With some recent heic related packages, we lost the ability to build
against some older libheif versions that we purport to support. We may
eventually want to raise the minimum supported libheif, but in the
mean time, this restores the prior build abilities.

### DIFF
--- a/src/heif.imageio/heifinput.cpp
+++ b/src/heif.imageio/heifinput.cpp
@@ -29,7 +29,9 @@ public:
     {
         return feature == "exif";
     }
+#if LIBHEIF_HAVE_VERSION(1, 4, 0)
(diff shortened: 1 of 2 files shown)


lgritz:
@hjmallon @cmstein

lgritz:
@cmstein confirms that this fixes his build break

hjmallon:
Thanks @lgritz I did have a horrible thought that something like this would be needed.

lgritz:
We finally got bitten by the fact that most of our CI configurations don't have libheif installed at all. We should fix that, and make sure that the "oldest everything", "current releases", and "bleeding edge" CI tests are all using appropriate versions of libheif.

PR #2824, opened 2021-01-13
[exif] memory leak in exif parse

servantf:
I had a crash on reading a JPEG image coming from a Canon EOS1200D.

After investigation, it appears that an IFD offset was larger than the buffer size. This is tested for various Exif IFD but not for the one spotted in this commit (INTEROP).

I am not sure why I have this problem in the first place (Bad header ? or memory leak ?)  but I assume this check won't hurt.


lgritz (approved):

lgritz (approved):
LGTM

PR #2827, opened 2021-01-15
Start staging span and string_view size and ssize changes

lgritz:
When we wrote our `span<>`, we based it off the draft for std::span,
which at that time specified span::size() as ptrdiff_t, which is
signed. By the time std::span got finalized in C++20, it was changed
to size_t, which is unsigned. Ick. Switching to size_t will be an API
break, so we can't do that until we're at a version where it's ok to
make breaking changes.

But we can stage a few changes that will make that future transition
easier.

* Add std::size() and std::ssize() wrappers for both span and
  string_view.  This gives downstream apps the opportunity to avoid
  span::size(), instead using std::size(the_span), which is guaranteed
  to be unsigned, and std::ssize(the_span), which is guaranteed to be
  signed, and both of these will work now as well as after we make the
  other breaking changes.

* Doing this properly required adding a couple 'constexpr' in places
  where they should have been all along.

* Define macros OIIO_SPAN_HAS_STD_SIZE and OIIO_STRING_VIEW_HAS_STD_SIZE
  making it easy to know if std::size and std::ssize are defined for
  span and string_view

* OIIO_SPAN_SIZE_IS_UNSIGNED will be defined when OIIO::span switches
  to having an unsigned size like std::span (making it easy for client
  software to know if span::size is signed or unsigned).

  Turn this on automatically for OIIO >= 3.0 so we don't forget.

  If client code predefines OIIO_SPAN_SIZE_IS_UNSIGNED before
  including span.h, that forces span::size to unsigned now (good for
  experimentation and compatibility testing, but I don't recommend
  anybody ship software in that state unless they really know what
  they are doing and don't mind potential incompatibility with the
  out-of-the-box OIIO).

These SHOULD NOT break any existing code or ABIs that uses our
existing span and string_view classes or their size() methods. It only
adds options that will help bridge compatibility.

### DIFF
--- a/src/include/OpenImageIO/span.h
+++ b/src/include/OpenImageIO/span.h
@@ -21,8 +21,25 @@
 
 OIIO_NAMESPACE_BEGIN
 
+// By default, our span::size() is a signed value, because we wrote this at
+// a time that the draft of std::span said it should be signed. The final
+// C++20 std::span ended up with an unsigned size, like all the other STL
+// classes. We will eventually conform by switching, but not until we are at
+// OIIO 3.0, allowing source code API-breaking incompatibilities. In the
+// mean time, we allow a back door to experiment with standard conformance
+// by pre-defining OIIO_SPAN_SIZE_IS_UNSIGNED=1.
+#if OIIO_VERSION_GREATER_EQUAL(3,0,0)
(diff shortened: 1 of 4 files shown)


lgritz:
Updated with revisions

PR #2830, opened 2021-01-16
Changing pybind11 bindings to used SHARED rather than using MODULE

alister-chowdhury:
This subtle change basically allows the python bindings to work correctly on windows.

For some background, I don't think I've ever gotten OIIO python bindings to work on windows, I've always chalked this up to using different versions of MSVC and never thought much of it (and just learned to only use the bindings at work).

All dependencies would have been installed via `vcpkg`.

The sort of error I would typically see was:
```python
>>> import OpenImageIO
  File "<stdin>", line 1, in <module>
ImportError: DLL load failed while importing OpenImageIO: The specified module could not be found.
```

Not terribly informative, but if instead I do:
```python
from ctypes import cdll
cdll.LoadLibrary("./OpenImageIO.cp38-win_amd64.pyd")
```

We get a popup message saying the following symbol couldn't be resolved:
`??9TypeDesc@OpenImageIO_v2_3@@QEBA_NAEBU01@@Z`

Running that through `https://demangler.com/` we get:
`public: BOOL __cdecl OpenImageIO_v2_3::TypeDesc::operator!=(struct OpenImageIO_v2_3::TypeDesc const & __ptr64)const __ptr64`

The actual definition of which is:
```cpp
    constexpr bool operator!= (const TypeDesc &t) const noexcept { return ! (*this == t); }
```

And if we run dumpbin on the python module we can see:
```
    OpenImageIO.dll
             1800D3280 Import Address Table
             1800EDBB8 Import Name Table
                     0 time date stamp
                     0 Index of first forwarder reference

                         14A ??9TypeDesc@OpenImageIO_v2_3@@QEBA_NAEBU01@@Z
                         ...
```

Somehow the python shared object, has a dependency on a constexpr function, which could (and I haven't looked into this) have something to do with declaring `OIIO_API` at the class level.

In any case, changing `pybind11_add_module` to use `SHARED` rather than implicitly `MODULE` fixes this (which is something I've seen using GCC/Linux on in-house codebases).

### DIFF
--- a/src/cmake/pythonutils.cmake
+++ b/src/cmake/pythonutils.cmake
@@ -69,7 +69,7 @@ macro (setup_python_module)
         set_property (SOURCE ${lib_SOURCES} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-macro-redefined ")
     endif ()
 
-    pybind11_add_module(${target_name} ${lib_SOURCES})
+    pybind11_add_module(${target_name} SHARED ${lib_SOURCES})
 
 #    # Add the library itself
 #    add_library (${target_name} MODULE ${lib_SOURCES})
--- a/src/cmake/pythonutils.cmake
+++ b/src/cmake/pythonutils.cmake
@@ -8,6 +8,11 @@ set (PYTHON_VERSION "" CACHE STRING "Target version of python to find")
 option (PYLIB_INCLUDE_SONAME "If ON, soname/soversion will be set for Python module library" OFF)
 option (PYLIB_LIB_PREFIX "If ON, prefix the Python module with 'lib'" OFF)
 set (PYMODULE_SUFFIX "" CACHE STRING "Suffix to add to Python module init namespace")
+if (WIN32)
+    set (PYLIB_LIB_TYPE SHARED CACHE STRING "Type of library to build for python module (MODULE or SHARED)")
+else ()
+    set (PYLIB_LIB_TYPE MODULE CACHE STRING "Type of library to build for python module (MODULE or SHARED)")
+endif ()
 
 
 # Find Python. This macro should only be called if python is required. If
@@ -69,7 +74,7 @@ macro (setup_python_module)
         set_property (SOURCE ${lib_SOURCES} APPEND_STRING PROPERTY COMPILE_FLAGS " -Wno-macro-redefined ")
     endif ()
 
-    pybind11_add_module(${target_name} SHARED ${lib_SOURCES})
+    pybind11_add_module(${target_name} ${PYLIB_LIB_TYPE} ${lib_SOURCES})
 
 #    # Add the library itself
 #    add_library (${target_name} MODULE ${lib_SOURCES})


lgritz:
Hmmm. 

So, we recently got the Python bindings working for the Windows CI build. A handful of tests in Windows CI are still failing and we're working our way through them, but the Python bindings definitely can work without this fix.

So I'm not sure that it's inherently broken with MODULE. But I don't know enough about Windows dev. Could it depend on which MSVS? Could the VCPkg setup have something different that is relevant? Is there a problem with the symbol visibility where we're doing something that's subtly wrong in this case?

If I understand the docs, SHARED is supposed to be for shared libraries that you link against, whereas MODULE is specifically for a runtime-loadable library, like a plugin (or python module). MODULE being the default that pybind11's utility functions impose is also strong circumstantial evidence that this is the usual choice for building loadable python modules.

I don't know all the ways that SHARED vs MODULE  differ on each OS. I know that for MacOS, they imply different file extension (by common convention, anyway) of .dylib vs .so. So I wouldn't want to end up with any oddities on other platforms by making a change here.

So here is how I would like to propose proceeding:

1. Since the docs appear to make it clear that this is more of a MODULE situation, let's be careful not to change or break any platforms where it currently works. So at the very least, it should only use SHARED -- if indeed that is the right solution anywhere -- on Windows.

2. Are we *sure* that the problem isn't a missing OIIO_API, or an OIIO_API where it shouldn't be, or some symbol visibility or compiler flag set incorrectly, that SHARED jiggles something that makes it work coincidentally, but that's not the most correct solution?

alister-chowdhury:
Yeah it does appear pybind11 does just pass it directly to `add_library`.
https://github.com/pybind/pybind11/blob/master/tools/pybind11Tools.cmake#L145-L151

And I'll be honest, before pybind11 I wasn't even aware of `MODULE` (and from what I can see, it does sound like it should result in just a .dll/.so with no .lib/.a).
Not really sure about the extension changing although on Windows, it appears to always result in a .pyd.


I managed to capture a diff between some of the generated things that sounded relevant (left is `MODULE` and right is `SHARED`):
`PyOpenImageIO.vcxproj`: https://editor.mergely.com/l9zdnZBA/
`CL.command.1.tlog`: https://editor.mergely.com/fpdqbBBP/
`link.command.1.tlog`: https://editor.mergely.com/LWgyhoCe/


Nothing really seems (to my eyes anyway) to standout here, which makes this even more confusing.
I attempted to use `THIN_LTO` and `NO_EXTRAS`, both of which didn't seem to do anything of note.

I removed the OIIO_API call from being at `struct OIIO_API TypeDesc` and sprinkled them into the respective implementation calls, and I can confirm that this causes the new unresolved symbol to move to:
`public: int __cdecl OpenImageIO_v2_3::ParamValue::nvalues(void)const __ptr64`
And `TypeDesc` references are no longer found as imports (via dumpbin).


So I did the same thing to `ParamValue` and `ParamValueList`, which then pushed the list of unresolved things onto:
`public: unsigned __int64 __cdecl OpenImageIO_v2_3::ImageSpec::scanline_bytes(BOOL)const __ptr64`


It would appear that putting `OIIO_API` explicitly on every function rather than at the class level fixes whatever symbol lookup problem is being casued (although I guess it wouldn't protect from a third-party library that might be in a header doing the same thing).
I'm not entirely sure WHY this is the case, because presumably, it's been like this for a while.

lgritz:
Can you check something briefly for me? Instead of relying on the default behavior of pybind11_add_module implying MODULE, what happens if you pass MODULE explicitly (instead of SHARED) to pybind11_add_module?

I was looking at the source of pybind11_add_module and there is something that looks suspiciously like there might be a bug in how this is handled.  (... haha, yes, and it has been fixed already in their master branch, though I'm not sure it affects us here or not)

lgritz:
"blanket using SHARED":

I don't like this solution because it seems to contradict the stated meaning of SHARED/MODULE. The [cmake docs](https://cmake.org/cmake/help/latest/command/add_library.html) say

> SHARED libraries are linked dynamically and loaded at runtime. MODULE libraries are plugins that are not linked into other targets but may be loaded dynamically at runtime using dlopen-like functionality.

That really makes it sound like python modules (not linked against explicitly, but runtime loaded by the python interpreter) is precisely the case where MODULE is intended to be used. 

In practice, on some platforms (including Linux), there is no effective difference between the two, but on others there may be, so this just doesn't seem safe or future-proof to me.

"Keeping MODULE but changing OIIO_API declerations."

If it were just one or two small classes where moving the OIIO_API from the class def to every non-inline method declaration might fix, I would say that sounds like the best fix. But it looks like we have lots of classes like that, so I'm not quite ready to encourage such an extensive change unless it's unnecessary

"Adding SHARED to just Windows":

This is certainly the simplest change (one word!), and has the merit of being certain not to break any platforms that are currently working.

The only thing I worry about is that, because we don't fully understand why this works when it seems to contradict the CMake docs -- and that other Windows users including our own Windows CI seem to have it working just fine now -- is that adding the SHARED will break some other Windows case. Maybe whatever is going wrong is particular to the specific version of MSVS, or cmake (maybe a bug was added, or fixed, deep in cmake's handling of SHARED vs MODULE for Windows).

But at least if, after making this change, if it turns out that other people complain that it breaks for them, it's also extremely easy to reverse and then try one of the more difficult solutions or to keep searching to understand the real underlying issue.

I do have one more alternate idea, will put in a different comment after I check a couple things.

lgritz:
The "other thing" I mentioned didn't pan out.

After a little more thought, I think we should go ahead with the simple "use SHARED on Windows" strategy, using a modified version of this PR:

Near the top of the pythonutils.cmake, where the other options are, put this:

    if (WINDOWS)
        set (PYLIB_LIB_TYPE SHARED CACHE STRING "Type of library to build for python module (MODULE or SHARED)")
    else ()
        set (PYLIB_LIB_TYPE MODULE CACHE STRING "Type of library to build for python module (MODULE or SHARED)")
    endif ()

and change the call to

    pybind11_add_module(${target_name} ${PYLIB_LIB_TYPE} ${lib_SOURCES})

This makes it very easy to override with a config-time `-DPYLIB_LIB_TYPE=MODULE`, in case anybody else's build is broken by your change.

If lots of people report that using SHARED breaks and it's only you and a minority of other people with a particular situation for whom MODULE is not the right choice, we could later change the default for Windows (like for the other OS's) back to MODULE, but then you would have a way to easily override to fix your builds.

What do you think?

alister-chowdhury:
I think this makes perfect sense!
PR updated.

lgritz (approved):
LGTM. Even if this breaks somebody, we know to just tell them to set that one build flag to get unstuck.

Issue #2828, opened 2021-01-15
[FEATURE REQUEST] Expose TextureSys in the Python bindings.


alister-chowdhury:
Hey there,

While implementing a mini-renderer using the OIIO texture system, I've found myself thinking, man this was be awesome if there were python bindings, so I didn't need to keep recompiling to run mini-renderer experiments.

Obviously, one can't expect the performance to be anything close to what you gain when you use C++, and probably shouldn't use this in production code.

But I can't be the first person to want to quickly test out sampling textures with mipblending when writing proof-of-concept prototypes.

I can't see a very obvious reason why (looking at the code and whats already exposed) this wouldn't make sense either.

Many thanks,
Alister

lgritz:
Yeah, should be straightforward. The only reason it wasn't ever done is because I never imagined that anybody would find TextureSystem useful from Python.

alister-chowdhury:
> Yeah, should be straightforward. The only reason it wasn't ever done is because I never imagined that anybody would find TextureSystem useful from Python.

Is this something you'd be comfortable with me having a crack at?

lgritz:
Oh yes, definitely!

alister-chowdhury:
Closing, sibling already merged

PR #2842, opened 2021-01-25
Adding basic TextureSystem support for python.

alister-chowdhury:
This PR contains some basic TextureSystem python bindings (#2828).

I haven't added any support for `PerThread`, `TextureHandle` or `TextureBatchOpt` function overloads.
Nor have I added `shadow`, since it appears to just be a dummy that returns `false`?
Additionally I haven't added a way to provide a `ImageCache` to the `TextureSystem`, mainly due to the way wrapping around `create` and `destroy` works, which sounds like opening a can of worms in regards to pointer management.

I couldn't quite figure out the most appropriate way to setup tests for these, I was more-or-less hoping to copy C++ tests for the `TextureSystem`, but was unable to find anything that looked like a good starting point?


lgritz (commented):
This looks great, thanks!

lgritz (changes requested):
To complete this PR, would you mind adding a test to the testsuite, akin to the other python-* tests, just to have a way to verify that the basic functionality works? I don't think it needs to do much other than create a TS, do a couple lookups from a known small texture and verify the results. But just to have some sanity check that the basics must be working and callable from Python.

lgritz:
I don't think you should bother with the PerThread, handles, or the batched stuff.  Let's start with the basics, it's not clear that anybody using this from Python will need any of those. This looks great except for lack of test. Look at a few of the testsuite/python-* tests to see how I did it for the other python classes, it's fairly straightforward.

alister-chowdhury:
Hrm...
`Linux oldest/hobbled: gcc4.8/C++11 py2.7 boost-1.66 exr-2.2 no-sse no-ocio`, `Linux static libs: gcc7 C++14 exr2.4` and `Linux 2017-ish: gcc4.8/C++11 py2.7 boost-1.66 exr-2.3` are falling over due to a segfault...

I'm not sure what the best way to debug this (since I don't have a setup which matches these enviroments).
Is there a way of extracting out the segfault from the tests?

lgritz:
Sorry for the delay. I'll try on my end and see if I can reproduce the failure.

lgritz:
I was able to reproduce this crash (I think only with combination of python 2.7 and somewhat older pybind11). Testing a fix now. If it works, I'll try to amend your PR by pushing the fix to your branch.

lgritz:
I pushed fixes to your branch.

The fix to the crashes was some minor rearrangement of how we release the GIL. Some combination of python2.7 and older pybind11 was not liking our constructing the result tuple of the texture call inside the block where we had gil_scoped_release.

I also changed some `TextureOptWrap*` to `TextureOptWrap&`.

And a couple minor renames, for example testsuite/python-texture-sys to testsuite/python-texturesys, to stick to the prevailing convention of using the class name.

If CI tests pass, I'll merge. I'm expecting a couple to fail for unreleated reasons -- your changes were based off a master of a few weeks ago, and I subsequently fixed some totally unrelated CI breaks.

lgritz (approved):
Looks good pending completion of CI tests.

alister-chowdhury:
Oh I see, I wouldn't have though the release gil needed to be scoped!
Thanks Larry!

PR #2843, opened 2021-01-25
make hash functions constexpr friendly so we can compute string hashe…

cmstein:
This makes the hash functions `constexpr` friendly.  If you're compiling for a CUDA device, then they will also be declared with `__device__` as well.

## Description

I have pretty much excised all the hashing code from `farmhash.cpp` and have moved it to a new file, `farmhash.h`.  As such, I suppose we should change the copyright notice in `farmhash.cpp` since it no longer contains any google code.

## Tests

I have a local OSL branch that uses ustring hashes in lieu of `const char *` for CUDA/OptiX code and it successfully grabs the the constexpr ustring hash at compile time when building the CUDA code.  Also, I don't see any changes in the testsuite in this branch compared to 'master' on my machine.

## Checklist:

- [ ] I have read the [contribution guidelines](../CONTRIBUTING.md).


lgritz (changes requested):
This is great. I have just a few suggestions, which are optional for you versus merging this and I can always patch it up as part of subsequent work that I'll do...

1. As discussed in chat, I'm curious to know if there's a reason why on the Cuda side you needed `__forceinline__` versus regular `inline`, and if there is not a good reason, let's prefer the latter, as it gives the compiler a little more leeway in deciding when to fully inline. (Truly forcing inline should be used sparingly because there are times it can reduce perf.)

2. I somewhat prefer OIIO::farmhash::inlined::foo() over OIIO::inlined_hashes::farmhash::foo? It feels like it should be more things inside the farmhash namespace, if you know what I mean.

3. I'm wondering if we should move farmhash.h into the include/OpenImageIO/detail/  subdirectory, because we don't really intend users to directly include it or call the functions therein. It has to be installed, but it's not really part of the direct public API, and we're leaning toward denoting that by the use of the detail subdir or namespace. I think I just want your opinion on this, you need not make that move right now; I can do it subsequently because it also will require a change to some of the CMake files to get it installed in the right way.

cmstein:
1)  regarding `__forceinline__` vs `inline`:  I ran a quick test over here and `inline` seems to suffice.  I'll try again just to verify, but I think this will be fine.  I'll clean it up.

2)  Yes, I prefer your "ordering" of the namespace.  I'll fix that up.

3)  Yes, I agree that this should be hidden from the user and so I like your proposal.  I'll probably leave it as is for now and let you fix things up once it's checked in.

lgritz:
That all sounds good to me.

cmstein:
just pushed an update PR addressing the comments

lgritz (approved):

PR #2847, opened 2021-01-28
Don't use the version of libheif that's known to be broken on Mac

lgritz:
This was causing failed CI tests and probably confused users.
libheif 1.10 is very broken on MacOS. The author says that 1.11 will come soon.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -163,10 +163,16 @@ checked_find_package (GIF
                       VERSION_MIN 4
                       RECOMMEND_MIN 5.0
                       RECOMMEND_MIN_REASON "for stability and thread safety")
+
 # For HEIF/HEIC/AVIF formats
 checked_find_package (Libheif VERSION_MIN 1.3
                       RECOMMEND_MIN 1.7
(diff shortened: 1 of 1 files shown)


EwoutH:
Is it correct that libheif is [still disabled in the CI]https://github.com/OpenImageIO/oiio/blob/master/src/build-scripts/gh-installdeps.bash#L40)? Or did I read that incorrect?

lgritz:
It's enabled for Mac, disabled for Linux and Windows.

I'll try turning on for Linux again and see what happens.

lgritz:
Seems to work, so I will re-enable for several of the Linux CI tests.

https://github.com/OpenImageIO/oiio/pull/2866

EwoutH:
Great, thanks! Any chance it could also be turned on for Windows and macOS?

libheif [v1.11.0](https://github.com/strukturag/libheif/releases/tag/v1.11.0) should have some fixes for macOS.

lgritz:
It has always been enabled for MacOS.

Enabling on Windows would require building it from scratch, or having the Windows CI tests download a pre-built version, if that's easily available. It would be great to have a volunteer to do that CI work, I'm really strapped for time at the moment.

PR #2853, opened 2021-02-03
Modify hash_test to verify correctness and stability of the hashes

lgritz:
Do some adjustment/simplification of the macros that set up the right
decorators on the new inline farmhash.

Some additional constexpr's, including Strutil::strhash().

Add to hash_test verification of the actual results of each of the
hashes, so we know that we don't break them over time, that they give
the same result on all platforms, and most immediately, to 100% verify
that the new constexpr inline version of the hash precisely matches
the old procedure call one.

Augment the existing benchmarking of the hashes to include
farmhash::inline.  It tries a variety of string sizes. I was worried
that the constexpr/inline version would be a performance hit because
it can't use any of the SSE intrinsics (which, ugh, aren't declared
constexpr in the headers).  But it turns out that the inline (but
SSE-free0 version is a little faster on some tests, a little slower on
other tests, overall a wash and never differing by more than
10%. Maybe that means that the inlining/constexpr makes up for the
lack of SSE? Maybe it means that the compiler optimization is adding
SSE back in well enough that losing the intrinsics isn't as bad as we
would have thought? In any case, the upshot is that this doesn't
represent a performance it, so I may consider a future change to just
go strictly with the inline constexpr versions and eliminating the
separately compiled version altogether. Will mull that over.

### DIFF
--- a/src/include/OpenImageIO/detail/farmhash.h
+++ b/src/include/OpenImageIO/detail/farmhash.h
@@ -78,7 +78,7 @@
 // Make static inline 'const expr, if possible.  Also, try to make CUDA friendly
 // for device code.
 #undef STATIC_INLINE
-#if  __cplusplus >= 201402L
+#if OIIO_CPLUSPLUS_VERSION >= 14
 #  define HASH_CAN_USE_CONSTEXPR 1
 #endif
 #define STATIC_INLINE OIIO_HOSTDEVICE inline OIIO_CONSTEXPR14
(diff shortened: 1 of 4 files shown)


lgritz:
I pushed some fixes that were necessary to make it fully happy on Windows.

cmstein:
LGTM!  I tested this branch locally with my updated OSL ustring/hash PR and it worked fine.

PR #2854, opened 2021-02-03
Static building

Shootfast:
## Description

Numerous small fixes for issues discovered when statically linking on Windows.

## Checklist:


lgritz (approved):
LGTM, thanks for the fixes.

Do you think we should aim to add a CI test to verify LINKSTATIC=1?

Shootfast:
A CI test for the static builds would be amazing, but I think it will be quite hard to set up.

On Windows (where I'm currently building a static OIIO) there is a lot of CMake mangling I am doing to point/trick all of the find_package scripts into working. Not to mention ensuring all the dependencies are built with the correct Debug/Release modes.

PR #2865, opened 2021-02-16
Improve parsing of xmp records

lgritz:
* Better handling of lists/sequences -- append into attribute arrays
  instead of allowing each value to overwrite the previous one.

* Better inference of unknown attributes that look like int, float, or
  rational, rather than cramming into strings.

* Fixed what I now think were bugs in handling rational values.

Fixes #2874


lgritz:
Fixes #2874

PR #2869, opened 2021-02-18
pkgconfig OpenImageIO.pc was setting include dir one level too deep

### LINKED ISSUE #2868: [HELP] Do you have any feedback about this ? (Homebrew/pkg-config --cflags)
Hi all !

I wanted to have your feedback about this issue in homebrew with the openimageio package.
https://github.com/Homebrew/homebrew-core/issues/71360

You may spare me a lot of time of my free time if you do so :)

Thanks in advance.

lgritz:
Closes #2868

### DIFF
--- a/src/build-scripts/OpenImageIO.pc.in
+++ b/src/build-scripts/OpenImageIO.pc.in
@@ -7,4 +7,4 @@ Name: OpenImageIO
 Description: OpenImageIO is a library for reading and writing images.
 Version: @PROJECT_VERSION_MAJOR@.@PROJECT_VERSION_MINOR@.@PROJECT_VERSION_PATCH@.@PROJECT_VERSION_TWEAK@
 Libs: -L${libdir} -lOpenImageIO 
-Cflags: -I${includedir}/OpenImageIO
+Cflags: -I${includedir}


lgritz:
@barbak Does this look right to you?

barbak:
For my use case it's great. Hoping it's not breaking any projects.
I saw that other lib often double the include dirs:
```
% pkg-config --cflags IlmBase
-I/usr/local/Cellar/ilmbase/2.5.5/include -I/usr/local/Cellar/ilmbase/2.5.5/include/OpenEXR
```
Not sure it is a good practice though ...

lgritz:
No, I think that's bad practice. I think all includes should look like `#include <OpenImageIO/foo.h>`, never just a raw `<foo.h>`, and so we definitely want this change. I'm willing to allow it to slightly break some people because it's definitely a bug as far as I'm concerned.

lgritz:
I don't know how many people use the pkgconfig. I certainly don't, which is why I had not noticed this before. So maybe this will only fix people, and not break anybody.

barbak:
This practice is fine with me. For the context, I was exhuming an old project that uses cmake and pkg-config. I don't think many projects use these kind of detection nowadays but it's always good to have this old pkg-config working correctly.
BTW, I think the problem is also in the OpenEXR library, https://github.com/AcademySoftwareFoundation/openexr/blob/master/cmake/OpenEXR.pc.in
I will contact them to see if they can fix that too !
Thanks again for your time.
Regards

Issue #2880, opened 2021-03-03
[BUG] Converting FLOAT to UINT16 causes segfault


apetrynet:
**Describe the bug**
In python, when attempting to write a `FLOAT` buffer to disk that has set write format to `UINT16` I get a segfault.
 
**To Reproduce**
``` python
import OpenImageIO as oiio
roi = oiio.ROI(0, 1920, 0, 1080, 0, 1, 0, 3)
buf = oiio.ImageBufAlgo.fill((0, 0, 0), roi=roi)
buf.set_write_format(oiio.UINT16)
buf.write('uint16.dpx')
```
I tested the equivalent code in oiio1.8.x it works.

**Expected behavior**
I expect OIIO to write a dpx file in UINT16

**Evidence**
``` bash
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import OpenImageIO as oiio
>>> roi = oiio.ROI(0, 1920, 0, 1080, 0, 1, 0, 3)
>>> buf = oiio.ImageBufAlgo.fill((0, 0, 0), roi=roi)
>>> buf.set_write_format(oiio.UINT16)
>>> buf.write('uint16.dpx')
Segmentation fault (core dumped)
```
``` bash
Python 2.7.5 (default, Nov 16 2020, 22:23:17) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import OpenImageIO as oiio
>>> roi = oiio.ROI(0, 1920, 0, 1080, 0, 1, 0, 4)
>>> buf = oiio.ImageBufAlgo.fill((0, 0, 0, 1), roi=roi)
>>> buf.set_write_format(oiio.UINT16)
>>> buf.write('uint16.dpx')
Segmentation fault
```

**Platform information:**
 - OIIO branch/version: 2.2.10.0 (Ubuntu python3.8) and 2.2.10.1 (CentOS7 python2.7)
 - OS: Ubuntu20.04/CentOS7
 - C++ compiler: 9.3.0 (Ubuntu) and 4.8.5 (CentOS7)

apetrynet:
I found a workaround that works in the meantime.

``` python
import OpenImageIO as oiio
roi = oiio.ROI(0, 1920, 0, 1080, 0, 1, 0, 3)
[...]

PR #2885, opened 2021-03-05
Fix crash related to ImageBuf::set_write_format

lgritz:
There were some flaws in the logic of IB::set_write_format that could
cause a crash when used in conjunction with a file format that doesn't
support per-channel data types (even if the set_write_format call
specified a single type rather than per-channel types!).

There are basically two problems I fixed here.

First, in ImageBuf::write, where we are dealing with the fact that
set_write_format previously registered a request, be extra careful
that the "default" data format is the "biggest" type out of the
per-channel choices, and also make sure that if all the channels are
identical, we just use the default type and clear the channelformats
vector.

Secondly -- and more importantly to the crash itself -- in
ImageOutput::to_native_rectangle, if the output file format does not
suport per-channel data types, just use the main format.  Do NOT
consider this case a hard error, which previously took an early out by
returning NULL for the location where the native data was copied.  It
was that NULL that, when subsequently copied from (not noticing that
it was NULL indicating an error), actually produced the segfault that
caused the crash.  In other words, don't worry, be happy -- just do
your best by treating the unsatisfiable request as the next best
thing, as if it were the same type for all channels.

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -1328,15 +1328,25 @@ ImageBuf::write(string_view _filename, TypeDesc dtype, string_view _fileformat,
         newspec.channelformats.clear();
     } else if (m_impl->m_write_format.size() != 0) {
         // If set_write_format was called for the ImageBuf, it overrides
-        if (m_impl->m_write_format.size())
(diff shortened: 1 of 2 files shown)


lgritz:
Fixes #2880

PR #2886, opened 2021-03-05
Fix error checking for file descriptor in gifinput on windows

SRHMorris:
## Description

Calling ImageInput::open() with a file path that doesn't exist will cause OIIO to attempt to load the missing file with all plugins. When it gets to the gif plugin the file descriptor is not checked correctly (windows only). This causes an assert inside DGifOpenFileHandle() when it calls _setmode().

## Checklist:


lgritz (approved):
Good catch!

PR #2888, opened 2021-03-06
Make RAW files handle Unicode filenames on Windows

### LINKED ISSUE #2881: [BUG] RAW image paths with ö do not open, or open broken
**Describe the bug**
Opening a RAW image with a filename or folder path which includes ö (and some other similar characters) either doesn't work, or opens broken. For example, `image.orf` will load file, but `imäge.orf` will fail with an IO error. Likewise, a CR2 with ä in it will open completely solid green but will open correctly if the ä is removed.

**To Reproduce**
1. Find any RAW file
2. Ensure it opens correctly
3. Rename the file or folder it's in to include ö
4. Try opening it again and get an error/bad image

**Expected behavior**
I expect the image to open correctly. Non-RAW images with the same names do work, so it's something in the RAW side of things that is breaking.

**Evidence**
Going to just point to some raw files I found online as examples since I don't know what files I have that I can give out.

CR2 opens fine, but when renamed shows up as pure green: http://www.rawsamples.ch/raws/canon/1dm2/RAW_CANON_1DM2.CR2
ORF opens fine, but when renamed gives an IO error: http://www.rawsamples.ch/raws/olympus/e3/RAW_OLYMPUS_E3.ORF

Error for when it doesn't load:
> Could not open file "C:/Users/Admin/Pictures/Debug/CR3/IMG_0482_räw.CR3", Input/output error

**Platform information:**
 - OIIO branch/version: 2.2.11.1
 - OS: Windows
 - C++ compiler: MSVC 2019

lgritz:
Fixes #2881

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -368,8 +368,13 @@ RawInput::open_raw(bool unpack, const std::string& name,
     m_processor->imgdata.params.user_flip
         = config.get_int_attribute("raw:user_flip", -1);
 
-    int ret;
-    if ((ret = m_processor->open_file(name.c_str())) != LIBRAW_SUCCESS) {
+#ifdef _WIN32
+    // Convert to wide chars, just on Windows.
+    int ret = m_processor->open_file(Strutil::utf8_to_utf16(name).c_str());
+#else
+    int ret = m_processor->open_file(name.c_str());
+#endif
+    if (ret != LIBRAW_SUCCESS) {
         errorf("Could not open file \"%s\", %s", m_filename,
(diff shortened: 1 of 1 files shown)


lgritz:
@AdamMainsTL are you able to apply this patch on your end and verify that it fixes the specific files you were having trouble reading?

AdamMainsTL:
> 
> 
> @AdamMainsTL are you able to apply this patch on your end and verify that it fixes the specific files you were having trouble reading?

Sorry didn't see this response. I can try it out maybe later today or tomorrow.

AdamMainsTL:
This seems to work with my previously broken files. Thank you very much for the fix!

Issue #2887, opened 2021-03-05
[BUG] OIIO seems to consider that the alpha channel is gamma-compressed when reading/writing 8-bit formats


devernay:
**Describe the bug**

Use the attached images `ramp.png` and `solidblack.png`.
ramp.png has a linear alpha variation from left to right (8-bit values checked with the GIMP)
and do:
```
oiiotool ramp.png --ch "R,G,B" -o rampoiiorgb.png # result way too dark
oiiotool --no-autopremult ramp.png --tocolorspace linear --premult --ch "R,G,B" --tocolorspace sRGB -o rampoiioopaque.png # GOOD expected result
oiiotool ramp.png --ch "R,G,B" -o rampoiiorgb.exr # result not fully OK, the red and black bands are too narrow
oiiotool rampoiiorgb.exr --tocolorspace sRGB -o rampoiiorgbviaexr.png # same

# composite over solid black
oiiotool ramp.png solidblack.png -over -o rampoversolidblackoiio.png # result too dark
oiiotool --no-autopremult ramp.png --tocolorspace linear --premult solidblack.png -over --tocolorspace sRGB -o rampoversolisblackoiio2.png # GOOF expected result
oiiotool ramp.png solidblack.png -over -o rampoversolidblackoiio.exr # result not fully OK
oiiotool rampoversolidblackoiio.exr --tocolorspace sRGB -o rampoversolidblackoiioviaexr.png # result OK

```

I checked manually the values at x=960, y=540:
- ramp.png has RGBA value (177,0,196,128) (50% alpha, unassociated because PNG)
- rampoiiorgb.png and rampoversolidblackoiio.png have RGB value (89,0,98)
- rampopaque.png (attached image) has RGB value (129,0,143) (it was obtained by linearizing RGB from ramp.png, then compositing over black with the linear alpha or equivalently multiply by alpha, then delinearizing RGB)
- rampoiioopaque.png and rampoversolisblackoiio2.png have RGB value (129,0,144), close to what I compute by hand
- rampoiiorgbviaexr.png and rampoversolidblackoiioviaexr.png have RGB value (159,0,167), which I cannot explain, where do these values come from?
- R_good = ((177/255)^2.2 * (128/255))^(1/2.2) * 255 = 129 <-- I should get this value
- B_good = ((196/255)^2.2 * (128/255))^(1/2.2) * 255 = 143 <-- id
[...]

PR #2890, opened 2021-03-08
oiiotool --autocc now has optional :unpremult=1 modifier

lgritz:
The oiiotool --colorconvert operator (and the other color conversion ops)
have an optional modifier ":unpremult=1" that will bracket the color
transformation with an unpremultiply/premltiply. But the implied
colorconvert that happens upon input and output with --autocc has no such
option, which is an oversight. This patch adds the ability to control
whether autocc transformations happen in an unpremultiplied state (which I
believe they SHOULD, generally speaking).

Also beef up the docs and help messages in some areas where I
coincidentally realized that -i and -o explanations omitted some of
the possible options.


lgritz:
Related to #2887

michdolan:
LGTM!

PR #2893, opened 2021-03-14
Docs: Fix doxygen mismatch, bump breathe version

lgritz:


lgritz:
oops, this was a merge, not a close, I just did i wrong

PR #2895, opened 2021-03-16
Fix reading tiff files when size is not multiple of tile size

nayyden:
## Description

The problem occurs when a tiff file has vertical size which is not multiple of the tile size.
For example tile size 64x64 and size 1240x1240. In this case, when `libtiff` reads the tiles on the last row, it moves the data to the bottom of the tile (vertically), but oiio starts reading from the top of the tile.

## Tests

I have a pre-existing .tx file which is not loaded correctly.

This is the result without the fix:
problem

And this is with the fix:
fixed

I also tested with `zackthecat.tif` from the tiff test suite. In this case the whole image is shifted because the tile is bigger than the image: resolution 234x213 with tile size 240x224

with 2.2.11.1
image
vs with the fix
image

I have not run the tests locally - not sure how yet, but I suppose the fix works and test cases need to be updated

command used:
```
./oiiotool -i zackthecat.tif -o zackthecat.jpg
```

## Checklist:

- [x ] I have read the [contribution guidelines](../CONTRIBUTING.md).

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -1843,10 +1843,10 @@ TIFFInput::read_native_tile(int subimage, int miplevel, int x, int y, int z,
         int vert_offset = m_spec.tile_height - th;
 
         copy_image(m_spec.nchannels, tw, th, 1,
-                   &m_rgbadata[vert_offset * m_spec.tile_width + (th - 1) * m_spec.tile_width], m_spec.nchannels,
-                   4, -m_spec.tile_width * 4, AutoStride, data,
-                   m_spec.nchannels, m_spec.nchannels * m_spec.tile_width,
-                   AutoStride);
+                   &m_rgbadata[vert_offset * m_spec.tile_width
+                               + (th - 1) * m_spec.tile_width],
+                   m_spec.nchannels, 4, -m_spec.tile_width * 4, AutoStride,
+                   data, m_spec.nchannels, m_spec.nchannels * m_spec.tile_width,
         return true;
     }
 
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -1833,13 +1833,15 @@ TIFFInput::read_native_tile(int subimage, int miplevel, int x, int y, int z,
             errorf("Unknown error trying to read TIFF as RGBA");
             return false;
         }
-        // Copy, and use stride magic to reverse top-to-bottom
+        // Copy, and use stride magic to reverse top-to-bottom, because
+        // TIFFReadRGBATile always returns data in bottom-to-top order.
         int tw = std::min(m_spec.tile_width, m_spec.width - x);
         int th = std::min(m_spec.tile_height, m_spec.height - y);
 
         // When the vertical read size is smaller that the tile size
         // the actual data is in the bottom end of the tile
-        // so copy_image should start from tile_height - read_height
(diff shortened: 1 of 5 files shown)


lgritz:
Are we doing the writing incorrectly as well?

nayyden:
I just tried with `oiiotool` on arch linux using the stock openimageio package and the problem exists - same result with `zackthecat.tif`.

About writing I don't think so.
I have tried various combinations of `maketx` parameters and it always worked in both oiio and Gimp.

Maybe the real reason it works is that I can't find a way to instruct `maketx` to create a file in the way the problematic images were created but I might be wrong.

lgritz:
The zackthecat image has a few other things that are unusual about it, and I am worried about confounding factors -- it's an older jpeg compression scheme, it's using subsampling, it's in YCbCr space. It may be using a variety of unusual code paths inside libtiff as well as oiio. Also, because of those oddities, Apple's Preview.app refuses to read it, which is another sanity check I would like to use.

Can you send me (privately is ok) a different image that exhibits this problem?

I'm wondering if these images have been created incorrectly, rather than that we're reading it incorrectly.

Reading the TIFF 6.0 spec, it says this about tiled images when the image size is not a whole multiple of the tile size:

>Boundary tiles are padded to the tile boundaries. For example, if TileWidth is 64 and ImageWidth is 129, then the image is 3 tiles wide and 63 pixels of padding must be added to fill the rightmost column of tiles. The same holds for TileLength and ImageLength. It doesn’t matter what value is used for padding, because good TIFF readers display only the pixels defined by ImageWidth and ImageLength and ignore any padded pixels. Some compression schemes work best if the padding is accomplished by replicating the last column and last row instead of pad- ding with 0’s.
>
>The price for padding the image out to tile boundaries is that some space is wasted. But compression usually shrinks the padded areas to almost nothing. Even if data is not compressed, remember that tiling is intended for large images. Large images have lots of comparatively small tiles, so that the percentage of wasted space will be very small, generally on the order of a few percent or less.
>
>The advantages of padding an image to the tile boundaries are that implementations can be simpler and faster and that it is more compatible with tile-oriented compression schemes such as JPEG. See Section 22.

To me, that sure sounds like any padding is to the right, and below, the valid parts of the boundary tiles. It does not sound like that last tile row has the valid pixels shifted to the bottom.

lgritz:
I see that the change you propose is only to the clause `if (m_use_rgba_interface)`. This is only used for a few obscure corners of the TIFF spec (such as when YCbCr and subsampling are used, as in the zackthecat image). So I wonder now if maybe this shift is something that TIFFReadRGBATile is doing, either purposefully or buggily. That's why it's so hard to generate files that exhibit the problem -- these are obscure cases that we (I think) never write.

lgritz:
Aha yes!

From the documentation of TIFFReadRGBATile:

> Note that the raster is assume to be organized such that the pixel at location (x,y) is raster[y*width+x]; with the raster origin in the lower-left hand corner of the tile. That is bottom to top organization. Edge tiles which partly fall off the image will be filled out with appropriate zeroed areas.

So I think that your fix is correct!

lgritz:
I pushed an update to your tree that added a bit more comment to explain what I learned about TIFFReadRGBATile, and also update the testsuite/tiff-suite reference output now that we have fixed the reading of the zackthecat image.

Merging now.

Thanks so much for this fix! It's always nice to discover a longstanding problem that we never noticed and have a complete fix for it.

nayyden:
Great, thanks!

PR #2899, opened 2021-03-18
BMP: fix bug reading older variety

### LINKED ISSUE #2898: [BUG] Loading BMP crashes in open
[PRINTER.zip](https://github.com/OpenImageIO/oiio/files/6160445/PRINTER.zip)

_open_ crashes on the attached bmp.

lgritz:
The very old OS2_V1 variety of BMP files -- which dates from the OS/2
and Windows 2.0 circa 1987 -- could have undefined results by relying
on the cpalete header field that didn't exist in that early
version. Our failure to initialize it led to expecting color palettes
of undefined size (which could either cause a read failure or even a
crash if it vastly over-allocated).

While we're at it, also rig the BMP reader to know the "dib" is another
extension often associated with BMP files.

Also beef up one of the error messages invoked for failures while reading
the color palette.

Fixes #2898


lgritz:
I pushed an update after finding some other problems with BMP.

* Added the test case from @DougRogers
* Noticed that our CI tests were not downloading the bmpsuite.zip and therefore not testing bmp at all.
* Noticed that our test, even when it did run, had (not previously noticed) incorrect results for one test, which revealed a bug wherein 4bpp BMP images whose horizontal width was an odd number of pixels would fail to fill in the correct value for the last column of pixels!
* Found a latent bug for reading BMP files of the very old OS2 variety when executing on big endian machines.

PR #2906, opened 2021-03-21
Fully separate libOpenImageIO and libOpenImageIO_Util

lgritz:
On a personal project a while back, I wasted a full day chasing down
inexplicable behavior that turned out to stem from TWO implementations
of ustring being in the same application, a result of inadvertently
linking against both libOpenImageIO and libOpenImageIO_Util. This led
to two ustrings that contain the same character sequence, but with
different pointers because they resided in different ustring tables.

There is no way that anybody without very deep understanding of ustring
and a lot of patience could have tracked down the underlying issue.

This patch eliminates the possibility of duplicated modules in
libOpenImageIO and libOpenImageIO_Util.  Now libOpenImageIO depends on
and links to the latter. This should all be automatic for users -- as
before, link against libOpenImageIO if you want everything, link
against libOpenImageIO if you just want the util helper classes but
don't need any of the image I/O functionality.  But now, if you
accidentally specify both... well, that's just the same thing as
asking for libOpenImageIO.


EvanAW:
> link against libOpenImageIO if you want everything, link
> against libOpenImageIO if you just want the util helper classes but
> don't need any of the image I/O functionality.

Minor nitpick: shouldn't the second one be `libOpenImageIO_Util`?

lgritz:
@EvanAW Yes. Typo.

PR #2912, opened 2021-03-25
Fix missing boost path on recently broken GHA CI

lgritz:
GitHub Actions in its wisdom has removed pre-installed Boost from the Windows
images, causing our CI to fail early in the cmake configure stage when finding
dependencies, doesn't even get as far as starting to compile any modules.

Instead, this change gets boost from vcpkg, just the pieces we need.

Unfortunately, although this approach finds the headers and completes the build,
several testsuite tests fail.

I need to track down the test failures, but even as it stands, this is still a step forward,
compared to having no idea at all if a PR going through CI breaks the Windows build.


lgritz:
Fixed the test failures, will merge.

PR #2921, opened 2021-03-28
TIFF: support IOProxy for input

### LINKED ISSUE #2920: [FEATURE REQUEST] Read TIFF from memory
Some raw formats have TIFF previews that need to read from memory.  It would be nice to have the ability to read these using OIIO.  As it is now, I have to have a separate TIFF reader that can do this.

This function may be helpful:

extern TIFF* TIFFStreamOpen(const char*, std::ostream *);

https://stackoverflow.com/questions/4624144/c-libtiff-read-and-save-file-from-and-to-memory

lgritz:
Also beef up docs, which had neglected to mention that proxy was
supported for a few other formats.

Closes #2920


lgritz:
@DougRogers

PR #2930, opened 2021-04-02
fix c++11 build

devernay:
## Description

C++11 build of OIIO failed with:
```
[  1%] Building CXX object src/libutil/CMakeFiles/OpenImageIO_Util.dir/argparse.cpp.o
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/libutil/argparse.cpp:16:
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/argparse.h:20:
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/paramlist.h:17:
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/attrdelegate.h:13:
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/typedesc.h:29:
In file included from /Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/strutil.h:24:
/Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/hash.h:334:44: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
OIIO_HOSTDEVICE inline constexpr uint128_t Uint128(uint64_t lo, uint64_t hi) {
                                           ^
/Users/devernay/Development/oiio-2.2.13.0/src/include/OpenImageIO/hash.h:335:12: note: non-constexpr constructor 'pair<unsigned long long &, unsigned long long &, false>' cannot be used in a constant expression
    return uint128_t(lo, hi);
           ^
/usr/local/opt/llvm/bin/../include/c++/v1/utility:444:5: note: declared here
    pair(_U1&& __u1, _U2&& __u2)
    ^
1 error generated.
```

## Tests

1. Build and install a C++11 OpenEXR (OpenEXR 2.5 is C++14 by default and must be compiled with `-DOPENEXR_CXX_STANDARD=11`) - if OpenEXR is C++14, then OIIO will also compile in C++14 mode, whatever the `CMAKE_CXX_STANDARD` value is. Maybe OIIO should check for this and trigger an error if `OPENEXR_CXX_STANDARD` > `CMAKE_CXX_STANDARD`?
2. Configure a C++11 oiio
3. Compile with make VERBOSE=1, and check that the c++11 standard is used (`-std=c++11`)

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/hash.h
+++ b/src/include/OpenImageIO/hash.h
@@ -331,7 +331,7 @@ OIIO_HOSTDEVICE inline constexpr uint64_t Uint128High64(const uint128_t x) {
     return x.second;
 }
 
-OIIO_HOSTDEVICE inline constexpr uint128_t Uint128(uint64_t lo, uint64_t hi) {
+OIIO_HOSTDEVICE inline OIIO_CONSTEXPR14 uint128_t Uint128(uint64_t lo, uint64_t hi) {
     return uint128_t(lo, hi);
 }
 


lgritz:
>  if OpenEXR is C++14, then OIIO will also compile in C++14 mode, whatever the CMAKE_CXX_STANDARD value is. 

Whoa, that's alarming. 

I think it's probably true that our CI tests the combo of gcc-4.8, C++11, OpenEXR 2.4, and that all the CI tests against OpenEXR 2.5 probably are ones where we intentionally are building OIIO with C++14 as well, so I can see how this would go unnoticed in our tests.

I'm not 100% sure I know why this pollution would happen, I will investigate.

In the mean time, I think this patch looks reasonable. Thanks for the fix.

devernay:
> I'm not 100% sure I know why this pollution would happen, I will investigate.

I know that these two lines triggered libutil to be compiled in C++14:
https://github.com/OpenImageIO/oiio/blob/master/src/libutil/CMakeLists.txt#L17

and traced it down to $prefix/lib/cmake/OpenEXR/OpenEXRTargets.cmake which contains a bunch of `INTERFACE_COMPILE_FEATURES "cxx_std_14"` if OpenEXR was compiled in C++14.

lgritz:
Aha, I figured it was something like that.

I feel like there is a bug in OpenEXR here -- there's nothing in the OpenEXR headers that should force downstream projects to need C++14, even if OpenEXR itself is compiled with C++14. I'm trying to figure out how it ends up with `INTERFACE_COMPILE_FEATURES "cxx_std_14"`. OIIO does not end up with this in its exported cmake configs, as far as I can tell (and also doesn't want it, for the same reasons). I'm trying to understand what the difference is.

lgritz:
I'll merge this into the release branch and make a release patch, but first I need to track down a couple other related things to fold into the same patch.

lgritz:
I believe this is the necessary fix on the OpenEXR side:  https://github.com/AcademySoftwareFoundation/openexr/pull/995

PR #2934, opened 2021-04-07
BMP: eliminate wasteful allocation and copying

lgritz:
Temporary allocations were made and freed for each scanline read or
written. Do it once per image and reuse. When reading, eliminate one
temp buffer by writing data into its final location, also eliminating
a needless extra copy of each scanline.


lgritz:
@pwm1234-sri reports that this works

Issue #2937, opened 2021-04-08
[BUG] fillholes_pushpull seems to be doing something funky when the input image isn't 4 channels.


alister-chowdhury:
So, we typically use fillholes_pushpull to bleed edges of textures, at some point (I suspect recently) this seems to have stopped working correctly.

When applied to say 2 channels (1 colour + 1 alpha) it doesn't seem to apply the triangle filter correctly.
This was originally discovered on Linux (2.2.7) via C++ but I have been able to replicate it with the master branch on Windows via Python.
Which does indicate this isn't related to bindings or any specific platform.

Here's some replication steps, I've gotten it to write out the images at every stage for convenience.
```python
import OpenImageIO

# 1 channel texture (lets say something a bump map or something from Substance or Mari)
texture = OpenImageIO.ImageBuf(OpenImageIO.ImageSpec(256, 256, 1, OpenImageIO.UINT8))
OpenImageIO.ImageBufAlgo.checker(texture, 32, 32, 1, (0.0,), (1.0,))

# Stencil we want to bleed edges from (lets say a models UVs rasterized)
stencil = OpenImageIO.ImageBuf(OpenImageIO.ImageSpec(256, 256, 1, OpenImageIO.UINT8))
OpenImageIO.ImageBufAlgo.render_box(stencil, 0, 0, 128, 256, (1.0,), True)

# Make a temporary buffer which contains the texture + stencil
tmp_spec = OpenImageIO.ImageSpec(texture.spec())
tmp_spec.nchannels += 1
tmp_spec.channelnames = tmp_spec.channelnames + ("injected",)
tmp_spec.alpha_channel = tmp_spec.nchannels - 1
tmp = OpenImageIO.ImageBuf(tmp_spec)

OpenImageIO.ImageBufAlgo.paste(tmp, 0, 0, 0, 0, texture)
OpenImageIO.ImageBufAlgo.paste(tmp, 0, 0, 0, tmp_spec.alpha_channel, stencil)
tmp = OpenImageIO.ImageBufAlgo.premult(tmp)

# Bleed the edges
bled = OpenImageIO.ImageBufAlgo.fillholes_pushpull(tmp)

# Extract the original channels and discard the stencil
bled_texture = OpenImageIO.ImageBuf(texture.spec())
OpenImageIO.ImageBufAlgo.paste(bled_texture, 0, 0, 0, 0, bled)

texture.write("texture.tif")
stencil.write("stencil.tif")
[...]

PR #2939, opened 2021-04-08
Added logic to copy along the alpha_channel when generating sub-images for fillholes_pushpull

alister-chowdhury:
Fixes the bug reported in #2937, whereby non RGBA images were effectively reading into negative memory when passed to `fillholes_pushpull`


lgritz (approved):
LGTM, thanks!

PR #2940, opened 2021-04-09
Adjust openexr includes

lgritz:
We broke things subtly for older OpenEXR when we added the 3.x
support.  Without this fix, our export CMake config files weren't
properly exporting the OpenEXR include path as the interface includes
for OIIO libraries. This doesn't affect the OIIO build itself, but it
can affect downstream libraries that use OIIO as a dependency and
rely on our exported cmake config files.

You might think that we wouldn't need this since we don't use any
OpenEXR classes in our APIs (though we do use Imath ones). Well, turns
out that our Imath.h file includes OpenEXR/OpenEXRConfigh in order to
find out which version it's using, so it can include the Imath headers
from the right place (which differs for Imath/OpenEXR 3.x versus 2.x).
In other words, we do leak an interface need for OpenEXR headers, even
though we don't use any interface need for OpenEXR classes.


lgritz:
Retracting, I have a better solution coming

PR #2942, opened 2021-04-12
Yet again, cmake adjustments to openexr, etc.

lgritz:
Sorry this has taken so much tinkering to get right for downstream
projects. It's tricky business.

First, we start by adding testsuite/cmake-consumer -- a test that
consumes OIIO and its exported cmake modules, and is independent of
the rest of OIIO's build system, more accurately simulating whether
things will work for a downstream client.

Adjust our exported module, it needs to break out the OpenEXR 3 vs 2
cases separately, as well as whether it found an OpenEXR/Imath config
file or not. I think I have all the cases covered now. I also needed
to specify find_dependency of IlmBase and Threads, for the sake of
openexr 2.4 with config, and add the INCLUDES directory lists in the
case of no config files found.

Some adjustment of our library target_link_libraries -- better PRIVATE
hiding of things not needed in the interface by downstream builds.
In particular, the OpenEXR components and OpenCV should be private, not
public.


lgritz:
@nrusch Can I ask you to please try this out and make sure I haven't re-broken things for you?

nrusch:
Just tested this from your PR branch, and by cherry-picking https://github.com/OpenImageIO/oiio/pull/2941 and these changes onto 2.2.13.1, and both work as expected against my split IlmBase/OpenEXR 2.5.5 install.

However, this does not fix https://github.com/OpenImageIO/oiio/issues/2936 (though I'm not sure if it's intended to).

lgritz:
Thanks, @nrusch. This wasn't related to 2936. It was yet another adjustment to how *downstream* projects consume our exported cmake configs, in situations where they are dealing with OpenEXR 2.4. I wanted to be sure that I wasn't re-breaking you (with OpenEXR 2.5) in the process. Whew!

PR #2943, opened 2021-04-12
support single channel 8-bit bmp

pwm1234-sri:
## Description

Support single channel, 8-bit grayscale bmp files. 

## Tests

I did not add a test.

## Checklist:

This is a small bug fix that should not need a CLA, updated documentation or tests. The code, except for the color_table_is_all_gray method, is copy/paste from an email exchange with Lary Gritz or from Larry's own lg-bmp branch where he cleaned up memory allocations when reading bmp files.

### DIFF
--- a/src/bmp.imageio/bmpinput.cpp
+++ b/src/bmp.imageio/bmpinput.cpp
@@ -99,6 +99,10 @@ BmpInput::open(const std::string& name, ImageSpec& spec)
         m_padded_scanline_size = (m_spec.width + 3) & ~3;
         if (!read_color_table())
             return false;
+        // m_allgray is a new BMPInput data member, initialized to false
+        m_allgray = color_table_is_all_gray();
+        if (m_allgray)
+            m_spec.nchannels = 1;   // make it look like a 1-channel image
         break;
     case 4:
         swidth                 = (m_spec.width + 1) / 2;
@@ -184,10 +188,18 @@ BmpInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         }
     }
     if (m_dib_header.bpp == 8) {
-        for (unsigned int i = 0, j = 0; j < scanline_bytes; ++i, j += 3) {
-            mscanline[j]     = m_colortable[fscanline[i]].r;
-            mscanline[j + 1] = m_colortable[fscanline[i]].g;
-            mscanline[j + 2] = m_colortable[fscanline[i]].b;
+        if (m_allgray) {
+            // Keep it as 1-channel image because all colors are gray
+            for (unsigned int i = 0; i < scanline_bytes; ++i) {
+                mscanline[i] = m_colortable[fscanline[i]].r;
+            }
+        } else {
+            // Expand palette image into 3-channel RGB (existing code)
+            for (unsigned int i = 0, j = 0; j < scanline_bytes; ++i, j += 3) {
+                mscanline[j]     = m_colortable[fscanline[i]].r;
+                mscanline[j + 1] = m_colortable[fscanline[i]].g;
+                mscanline[j + 2] = m_colortable[fscanline[i]].b;
+            }
         }
     }
     if (m_dib_header.bpp == 4) {
@@ -267,6 +279,21 @@ BmpInput::read_color_table(void)
         }
     }
     return true;  // ok
+}
+
+bool
+BmpInput::color_table_is_all_gray(void)
+{
+    size_t ncolors = m_colortable.size();
+    for (size_t i = 0; i < ncolors; i++)
+    {
+        color_table& color = m_colortable[i];
+        if ((color.b == color.g) && (color.g == color.r))
+            continue;
+        else
+            return false;
+    }
+    return true;
 }
 
 OIIO_PLUGIN_NAMESPACE_END
--- a/src/bmp.imageio/bmpinput.cpp
+++ b/src/bmp.imageio/bmpinput.cpp
@@ -99,7 +99,6 @@ BmpInput::open(const std::string& name, ImageSpec& spec)
         m_padded_scanline_size = (m_spec.width + 3) & ~3;
         if (!read_color_table())
             return false;
-        // m_allgray is a new BMPInput data member, initialized to false
         m_allgray = color_table_is_all_gray();
         if (m_allgray)
             m_spec.nchannels = 1;  // make it look like a 1-channel image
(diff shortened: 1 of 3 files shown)


lgritz:
This seems to be on top of my PR for reducing allocations (#2934). Can I assume that means that you have tried it and are happy with it?

If that's the case, then I'd like to suggest that I go ahead and merge that PR, and then you should rebase on top of it (and re-push this branch with --force). That will update this PR and make it so the diffs are only your new part, will make it a bit easier to review than having my changes mixed in.

pwm1234-sri:
Yes, my PR is on top of (#2934). And, yes, I have tried it and am happy with the reduced allocation changes you made for bmp files. (I tested using ImageInput and see that no allocations are made when reading into an external buffer - thank you!)

lgritz:
OK, I merged the other PR. I suggest that you rebased this one on top of it.

src/bmp.imageio/bmpinput.cpp
```diff
@@ -99,6 +99,10 @@ BmpInput::open(const std::string& name, ImageSpec& spec)
         m_padded_scanline_size = (m_spec.width + 3) & ~3;
         if (!read_color_table())
             return false;
+        // m_allgray is a new BMPInput data member, initialized to false
```
lgritz, line 102: This comment can be eliminated, it was just explaining what m_allgray was for when I was explaining this idea in the ticket

src/bmp.imageio/bmpinput.cpp
```diff
@@ -184,10 +188,18 @@ BmpInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+        } else {
+            // Expand palette image into 3-channel RGB (existing code)
+            for (unsigned int i = 0, j = 0; j < scanline_bytes; ++i, j += 3) {
+                mscanline[j]     = m_colortable[fscanline[i]].r;
+                mscanline[j + 1] = m_colortable[fscanline[i]].g;
+                mscanline[j + 2] = m_colortable[fscanline[i]].b;
+            }
         }
     }
     if (m_dib_header.bpp == 4) {
```
lgritz, line 205: I don't know if you care (nor if I care), but this only fixes the 8 bits per pixel case. Are you concerned with 4bpp and 1bpp cases to also ensure that if their (even more limited) color map entries are all gray that it should appear to OIIO as if it were a one channel image?

src/bmp.imageio/bmpinput.cpp
```diff
@@ -269,4 +281,18 @@ BmpInput::read_color_table(void)
...
+bool
+BmpInput::color_table_is_all_gray(void)
+{
+    size_t ncolors = m_colortable.size();
+    for (size_t i = 0; i < ncolors; i++) {
+        color_table& color = m_colortable[i];
+        if ((color.b == color.g) && (color.g == color.r))
+            continue;
+        else
+            return false;
```
lgritz, line 293: Since the `continue` is exactly what happens at the end of the loop anyway, can/should this be simplified to:

    if (color.r != color.g || color.r != color.b)
        return false;

lgritz (approved):
LGTM. Two very minor suggestions for improvement in the comments.

src/bmp.imageio/bmpinput.cpp
pwm1234-sri: Removed in latest commit.

src/bmp.imageio/bmpinput.cpp
pwm1234-sri: No, I do not work with 4bpp or 1bpp images.

src/bmp.imageio/bmpinput.cpp
pwm1234-sri: I applied this simplification, which is better than my initial verbose implementation.

pwm1234-sri:
Is there anything else you want me to do to this PR?

lgritz:
No, it looks great. I will merge after the CI tests on the latest push are finished.

pwm1234-sri:
Thanks! 

I have another PR I can submit if you are interested. It is for ``to_OpenCV`` [line 313](https://github.com/OpenImageIO/oiio/blob/665af526dfb0beec1dfc08e9c5b281f59eb06ede/src/libOpenImageIO/imagebufalgo_opencv.cpp#L313) and following.  Instead of ``cv::Mat mat(roi.height(), roi.width(), dstFormat);`` with ``dst = std::move(mat);`` at the end, you can eliminate ``mat`` altogether and use ``dst.create(roi.height(), roi.width(), dstFormat);`` since the opencv Mat ``create`` method does the same thing as the constructor if the arguments are for something different than the instance is already for. But, it does no change if the instance already matches the spec. Eliminating this object avoids another unnecessary memory allocation for me where I am using opencv for the processing but oiio for image reading (and soon to be writing). Of course I will submit a PR if this sounds reasonable (or even if you are midly interested). But I will not bother if this is not a change that you think you would want.

pwm1234-sri:
I do not think my last commit had anything to do with the test failure:

```
99% tests passed, 1 tests failed out of 139

Total Test time (real) = 490.77 sec

The following tests FAILED:
	100 - openexr-chroma (Failed)
```

lgritz:
No, I'm seeing a few intermittent failures on Windows, and it's definitely not related to this PR.

lgritz:
Are you building out of master, or do you need this backported to the release branch?

pwm1234-sri:
I have been using a release build, but I can use master or a tagged commit on master, unless you think there is a reason not to. I do not have any thing forcing me to a release branch other than the desire for stability. (My platform is currently windows VS2019, but I also use Ubuntu 18, and any commit you point me to will be fine.) So I am fine with leading edge but not bleeding edge. Master works okay for me, but I would to a release-2.3.x-dev tag when it is available.

lgritz:
This is the kind of change to existing behavior that I prefer not to backport to existing releases unless somebody makes the specific request for a good reason. So if you are not desperate to have this in 2.2.x, then I recommend you work from master now (and whatever commit you want to lock down on) and then can peg to the next dev tag I make.

My studio does production work from master (or close to it), so I'm not worried about stability or robustness. The most important difference between master and release branches is simply that release branches promise not to break backward ABI and API compatibility.

pwm1234-sri:
Okay, sounds great! I would prefer to stay closer to the leading than the trailing edge. And backward compatibility, and particularly, ABI compatibility is not important to me.

By the way, did you get anyone to take you up on your request to take on the vcpkg build of oiio? I do not think I can take that on myself, but I might be able to lend a hand. There are still many things about vcpkg I do not understand, but I am able to use it. In fact for these two PRs, I used my vcpkg build (on windows) of the oiio dependencies (but without the vcpkg build of openimageio itself). And then I was able to build the oiio master branch using the vcpkg toolchain file. (I had to first remove the find modules for ffmpeg, libraw, and opencv from src/cmake/modules as is done in the openimageio port. ) There are still a couple of kinks with the build of the debug configuration, so it would be much better if I could build the head of the master branch (or something much newer than what is on vcpkg), but the current patch files in the vcpkg port make that difficult (at least for me).

PR #2944, opened 2021-04-13
use dst.create rather than constructor w/move

pwm1234-sri:
## Description

Avoid unnecessary allocation in to_OpenCV 
## Tests

No, I did not add documentation or a test. 

## Checklist:

This is a minor change that does not warrant a CLA.

The [cv::Mat::create](https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a55ced2c8d844d683ea9a725c60037ad0) method does the same thing as the constructor but "Allocates new array data if needed." And when not needed, which is my use case when all the images I process are the same size, ``create`` does nothing. This is more efficient than the implementation where a new mat is constructed and then moved to the destination. This is a minor optimization, but it does avoid allocating an image size block of pixels when it is not necessary.

Thanks for your time and consideration!


lgritz (approved):
LGTM

Issue #2959, opened 2021-05-01
Crash in TextureSystem::texture with automip=1 and TextureOpt::firstchannel>0, when sampling mip levels


ix-dcourtois:
Edit: I forgot to mention we're running on OIIO 2.1.15. I'll try to run the same test on a more recent version.
Edit2: Still happening with 2.2.13.1

**Describe the bug**
We have an OpenEXR file with lots of AOVs, and we noticed some crashes when we tried to read any of those AOVs with the `automip` option set to `1`

I recompiled OIIO in debug to try and understand the cause of the problem, and I think I narrowed it down to https://github.com/OpenImageIO/oiio/blob/4824fc7457b698bdd6b3264b9b1d72c46143aec4/src/libOpenImageIO/imagebuf.cpp#L2062

In our case, the ROI has `chbegin=10` and `chend=13`. The result of that line is that the ROI now has `chbegin=10` and `chend=3`, resulting in the next calls to `roi.chnnels()` to return `-7`, leading to crashes (access violation) in the `parallel_convert_image` call happening just after.

At first, I tried to simply comment out that line. It fixes the crashes. BUT then the mipmaps seems incorrect:
![mipmaps](https://user-images.githubusercontent.com/29434962/116780074-7922fc00-aa7a-11eb-9d4e-7cd06fd6fd3a.gif)
When we zoom out, the colors shift, and there's even a sort of black band appearing.
This is why I'm opening this issue instead of a pull request.

**To Reproduce**
Just compile the following sample, and put the attached .exr file ([aov_multiple00000.zip](https://github.com/OpenImageIO/oiio/files/6409409/aov_multiple00000.zip))
 next to it:
```cpp
#include <OpenImageIO/texture.h>

int main(int argc, char ** argv) {
    OIIO::ustring filename("aov_multiple00000.exr");

    OIIO::TextureSystem *texture_system = OIIO::TextureSystem::create();
    texture_system->attribute("automip", 1);

    OIIO::TextureOpt options;
    options.firstchannel = 10;

    float result[4] = { 0.0f, 0.0f, 0.0f, 0.0f };
    texture_system->texture(filename, options, 0.2, 0.2, 0.003, -0.003, 0.003, 0.003, 3, result);

    OIIO::TextureSystem::destroy(texture_system);
    return 0;
}
```

[...]

PR #2962, opened 2021-05-03
fixed a crash in ImageCacheFile::read_unmipped

ix-dcourtois:
Fixed a crash in ImageCacheFile::read_unmipped when sampling a channel index greater than the number of channels we're sampling (see Description below)

## Description

This is a proposed fix for issue #2959 
To give a bit more information: `ImageCacheFile::read_unmipped` crashes when trying to read from a generated mipmap, when its called with a `chbegin` greater than the number of channels we're sampling.
This is because that method creates a temp image buffer with `chend - chbegin` channels only, and then at the end of the method, it will get its pixels by calling its `get_pixels` methods, and by providing a ROI with unchanged `chbegin` and `chend`.
And in the beginning of `ImageBuf::get_pixels` this line result in a crash in our case: `roi.chend = std::min(roi.chend, nchannels());`
Let's say we were sampling channel 10 to 13, we're creating an image buffer with only 3 channels, and when we get the pixels, that line will set `roi.chend` to `3`.. which leads to negative values in strides and crashes.

Anyway this pull request fixes this by changing the ROI instance passed to `ImageBuf::get_pixels` in order to match how that image buf was created.

## Tests

The only test I did for now was to test the fix of the crash through the sample attached to the issue, and also check that mipmaps are correct through our software.

I'm going to let the CI run on this PR and update this once the results are available.

## Checklist:

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -934,7 +934,7 @@ ImageCacheFile::read_unmipped(ImageCachePerThreadInfo* thread_info,
     }
 
     // Now convert and copy those values out to the caller's buffer
-    lores.get_pixels(ROI(0, tw, 0, th, 0, 1, chbegin, chend), format, data);
+    lores.get_pixels(ROI(0, tw, 0, th, 0, 1, 0, nchans), format, data);
 
     // Restore the microcache to the way it was before.
     thread_info->tile     = oldtile;


lgritz (approved):
Aha, yes, you are right. We've already copied [chbegin,chend) by using get_pixels(), so now they appear in the lores buffer as [0,nchans).

Good catch!

ix-dcourtois:
Hi, thanks for approving. I had a look at tests, but `ImageCache` doesn't seem to have ones.
Not sure if I should add ones (and if so, how) Let me know if something needs to be done there.

lgritz:
ImageCache lies behind all TextureSystems operations and quite a bit of ImageBuf (except for fairly small images), so it's indirectly tested in a large portion of the testsuite. There's almost never a need to write separate ImageCache tests. For this change, it's testing a tiny edge case that's apparently not tested elsewhere, but my inclination is not to make a separate test, because this is a one-liner that is so obviously the right fix. I wouldn't worry about it.

Issue #2969, opened 2021-05-13
[BUG] BMP fails in read_image


DougRogers:
[rle.zip](https://github.com/OpenImageIO/oiio/files/6475294/rle.zip)

Crashes in _read_image_

```
int main(int argc, char *argv[])
{
    std::string filename = "F:/rle.bmp";

    auto in = ImageInput::open(filename);
    if (in)
    {
        const ImageSpec &spec = in->spec();
        int xres              = spec.width;
        int yres              = spec.height;
        int channels          = spec.nchannels;
        std::vector<unsigned char> pixels(xres * yres * channels);
        in->read_image(TypeDesc::UINT8, &pixels[0]);
        in->close();
    }
```

Windows x64, Visual Studop 2019

DougRogers:
![image](https://user-images.githubusercontent.com/7563851/118215336-3112ba80-b426-11eb-87ea-27689da847b2.png)

DougRogers:
![BMPError](https://user-images.githubusercontent.com/7563851/118215531-8353db80-b426-11eb-9b7e-d5ea64f3897c.png)

lgritz:
I'm afraid it seems like our current implementation of BMP does not support RLE compression. 

It's a little surprising that nobody has noticed this or complained in all these years. On the other hand, I did a bit of google searching in the hopes of finding an existing open source BMP reader we could simply use (if it had the right license terms) rather than having to implement it from scratch, but after skimming the code of the first dozen hits, I did not encounter a single one that handled RLE compressed files. They just ignored the issue entirely, didn't even have a comment like `/* TODO: Add RLE support */`. So maybe RLE compressed BMPs aren't super common in the wild? At least, not common enough that most people who try to implement BMP readers add support for RLE.

I offer PR #2972 as a temporary measure -- at least this checks the compression and gives a helpful error message (without reading more of the file, crashing, or returning the wrong image).

I would love for OIIO to support RLE-compressed BMP files. We would welcome a PR that added this capability. It's probably not very difficult, but I just don't currently have sufficient need/time ratio to do it myself.

DougRogers:
I agree with you.  Returning an error for this case is fine until/if it is
addressed.

The bmp encoder/decoder is MIT licensed.

https://github.com/fabvalaaah/rlec

-- 
-Doug

lgritz:
[...]

PR #2972, opened 2021-05-16
Clarify that our BMP implementation does not support RLE files.

lgritz:
We would welcome a patch that adds this capability.


lgritz:
This is related to #2969. It's not a complete fix, which would fully implement support for RLE files. But maybe somebody will be interested in implementing support.

PR #2979, opened 2021-05-24
Build: FindOpenJPEG.cmake improvements

lgritz:
Rename FindOpenJpeg -> FindOpenJPEG to match the conventions used by
that project, in particular for its exported CMake config files.

(We don't yet rely on their exported configs, I need to do some
homework to see if they exist and are reliable as far back as the
oldest OpenJPEG release that we support.)


lgritz:
Amended with a build_OpenJPEG.bash script that can help users (or our CI) build a missing OpenJPEG.

PR #2985, opened 2021-05-30
Fix broken build against TBB 2021

### LINKED ISSUE #2963: [BUILD] Can't build against TBB 2021
I can't build the 2.2.14.0 against TBB 2021.x, only with 2020_U2, because OpenImageIO can't find it, I think a little tweak need be done on [FindTBB.cmake](https://github.com/OpenImageIO/oiio/blob/master/src/cmake/modules/FindTBB.cmake) file.

I am using Slackware 15.0 beta
GCC 10.3.0
Cmake 3.20.2
Boost 1.76

lgritz:
For TBB 2021, headers moved around, and our FindTBB.cmake was failing.
But actually, starting with TBB 2021, TBB exports a proper cmake config!
So take advantage of that by preferring to load TBB from a config file.

Fixes #2963

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,7 +42,7 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
      * Field3D (tested through 1.7.3)
  * If you want support for OpenVDB files:
      * OpenVDB >= 5.0 (tested through 8.0) and Intel TBB >= 2018 (tested
-       through 2020_U3)
+       through 2021)
      * Note that OpenVDB 8.0+ requires C++14 or higher.
  * If you want support for converting to and from OpenCV data structures,
    or for capturing images from a camera:
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -161,8 +161,8 @@ checked_find_package (OpenCV
 # Intel TBB
 set (TBB_USE_DEBUG_BUILD OFF)
 checked_find_package (TBB 2017
-                   DEFINITIONS  -DUSE_TBB=1
-                   ISDEPOF      OpenVDB)
+                      DEFINITIONS  -DUSE_TBB=1
+                      PREFER_CONFIG)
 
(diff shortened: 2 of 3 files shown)


MehdiChinoune:
I don't think, this is enough, openvdb.imageio should be also linked to TBB::tbb. ${TBB_LIBRARY} is now empty with this change.

lgritz:
Are you using a static library for OpenVDB?

MehdiChinoune:
> Are you using a static library for OpenVDB?

No, I am shared library, assuming you mean OpenVDB_DLL

lgritz:
That's odd, because the OpenVDB library itself should be linked to TBB so that merely pulling in OpenVDB ought to be enough and OIIO shouldn't have to explicitly list OpenVDB's upstream dependencies like TBB.

BTW, which OpenVDB release are you using?

lgritz:
And which version of TBB?

MehdiChinoune:
It happened with both TBB-2020-03/OpenVDB-8.0 and TBB-2021.04/OpenVDB-9.0, BTW I am using MINGW.

lgritz:
I think I see what's happening.

MehdiChinoune:
This is the log for TBB-2020.03/OpenVDB-8.0 https://github.com/MehdiChinoune/MINGW-packages/runs/4415702900?check_suite_focus=true
and this one for TBB-2021.04/OpenVDB-9.0 https://github.com/MehdiChinoune/MINGW-packages/runs/4418013268?check_suite_focus=true
This is the log after I have changed (patched oiio) ${TBB_tbb_LIBRARY} to TBB::tbb https://github.com/MehdiChinoune/MINGW-packages/runs/4418326677?check_suite_focus=true

lgritz:
I think I understand. 

For older TBB, we were using our own FindTBB.cmake, which set that TBB_tbb_LIBRARY. Newer TBB exports cmake config files properly, but it defines the target rather than that library. Fix on the way.

lgritz:
Does this do the trick for you?  https://github.com/OpenImageIO/oiio/pull/3214

PR #2991, opened 2021-06-05
make_texture now detects and rejects deep images

### LINKED ISSUE #2950: [BUG] Texture conversion causes null pointer dereference with test .exr files
**Describe the bug**
When using Windows 10 with OpenImageIO-2.2.10.0 results in certain .exr files causing segfaults during texture conversion.

**To Reproduce**
1. Download https://github.com/AcademySoftwareFoundation/openexr-images
2. With regards to the images, texture convert them using the options:
```
    config.attribute("maketx:nomipmap", false);
    config.attribute("maketx:oiio_options", true);
    config.attribute("maketx:checknan", true);
    config.attribute("maketx:filtername", "lanczos3");
    config.attribute("maketx:fileformatname", "tx");
```
Then any of the following textures:

- openexr-images\v2\LeftView\Ground.tx
- openexr-images\v2\LeftView\Leaves.tx
- openexr-images\v2\LeftView\Trunks.tx
- openexr-images\v2\LowResLeftView\composited.tx

3. Result in a segfault here: https://github.com/OpenImageIO/oiio/blob/5c96e9df6b3af74dd751bbc176290559cbca18d9/src/include/OpenImageIO/fmath.h#L1145 due to src being a nullptr

**Expected behavior**
I don't think any texture conversion should result in a segfault, instead, it should result in a warning indicating the conversion issue.

**Evidence**
image

**Platform information:**
 - OIIO branch/version: 2.2.10.0
 - OS: OS Name: Microsoft Windows 10 Home - Version 10.0.19041 Build 19041
 - C++ compiler: Visual Studio Community 2019
 - Any non-default build flags when you build OIIO: default build for vcpkg x64-windows

lgritz:
It's meaningless to try to use make_texture on input images that are
"deep".  We never even thought to check for this, which led to a
cascade of problems resulting in a segfault.

Fixes #2950

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -1004,6 +1004,11 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
     }
     OIIO_DASSERT(src.get());
 
+    if (src->deep()) {
+        errorfmt("Deep images cannot be turned into textures.");
+        return false;
+    }
+
     if (!outputfilename.length()) {
         std::string fn = src->name();
         if (fn.length()) {


etheory (approved):
LGTM! Thanks for finding the root of this issue.
Since the code I was running was just attempting to load a bunch of random textures, this never even occurred to me.
Thanks!

PR #2993, opened 2021-06-06
Raise giflib minimum to 5.0 and spruce up

lgritz:
* Raise minimum GIFLib to 5.0. That was back in 2012, so there's really no
  reason for us to support 4.x any longer.
* Remove cruft no longer necessary.
* While I was in there, update error calls to the new format convention,
  and change some old NULL to nullptr.

This will only go into master / 2.3, it will not change the minimum
dependency of released branches.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -48,8 +48,7 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
    or for capturing images from a camera:
      * OpenCV 2.x, 3.x, or 4.x (tested through 4.5)
  * If you want support for GIF images:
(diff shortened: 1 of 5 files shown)


lgritz:
I'm having second thoughts about this. It seems that all varieties of CentOS 7 (the most common Linux distro family in the VFX world) are inexplicably pegged to giflib 4.1.

Now I'm worried that removing support for 4.x (despite 5.0 being released 9 years ago!) has very little benefit in terms of code simplification, but will surely result in an increased support burden from people whose OIIO builds are unable to read such a common file format.

I'll sleep on it, but probably will end up withdrawing this PR.

PR #2998, opened 2021-06-11
More exception safety in Filesystem directory iteration

lgritz:
When using filesystem::directory_iterator, I was using operator++,
which can throw exceptions for OS errors. I should have used
increment(ec) instead. Throw in a surrounding try/catch just in case,
since even increment is not marked as noexcept.


etheory:
LGTM, thanks @lgritz!

PR #3009, opened 2021-06-23
[WIP] Experimental use of openexr C layer in openimageio

kdt3rd:
Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>

## Description

Experimental only version of use of the C layer for reading images, to be used for evaluating the new API, and testing improved thread behavior when higher level application (i.e. renderer) has it's own pile of threads hammering away at textures.

## Tests

At this point, just smoke test that iv can load a couple of images.

## Checklist:

### DIFF
--- a/src/openexr.imageio/exrinput_c.cpp
+++ b/src/openexr.imageio/exrinput_c.cpp
@@ -1241,7 +1241,7 @@ OpenEXRInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
     }
 #endif
     int endy = spec.y + spec.height;
-    yend = std::min(endy, yend);
+    yend     = std::min(endy, yend);
     std::vector<uint8_t> fullchunk;
     bool first = true;
     int nlines = scansperchunk;
--- a/src/openexr.imageio/exrinput_c.cpp
+++ b/src/openexr.imageio/exrinput_c.cpp
@@ -210,7 +210,8 @@ private:
         {
         }
         ~PartInfo() {}
-        bool parse_header(OpenEXRInput* in, exr_context_t ctxt, int subimage);
+        bool parse_header(OpenEXRInput* in, exr_context_t ctxt, int subimage,
+                          int miplevel);
         bool query_channels(OpenEXRInput* in, exr_context_t ctxt, int subimage);
         void compute_mipres(int miplevel, ImageSpec& spec) const;
     };
@@ -455,7 +456,9 @@ OpenEXRInput::open(const std::string& name, ImageSpec& newspec,
         m_userdata.m_io = nullptr;
         return false;
     }
-
+#if ENABLE_READ_DEBUG_PRINTS
+    exr_print_context_info(m_exr_context, 1);
+#endif
     rv = exr_get_count(m_exr_context, &m_nsubimages);
     if (rv != EXR_ERR_SUCCESS) {
         m_local_io.reset();
@@ -502,7 +505,7 @@ OpenEXRInput::init_part(int subimage, int miplevel)
 
 bool
 OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in, exr_context_t ctxt,
-                                     int subimage)
+                                     int subimage, int miplevel)
 {
     bool ok = true;
     if (initialized)
@@ -1076,7 +1079,7 @@ OpenEXRInput::seek_subimage(int subimage, int miplevel)
 
     PartInfo& part(m_parts[subimage]);
     if (!part.initialized) {
-        if (!part.parse_header(this, m_exr_context, subimage))
+        if (!part.parse_header(this, m_exr_context, subimage, miplevel))
             return false;
         part.initialized = true;
     }
@@ -1207,6 +1210,9 @@ OpenEXRInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
         return false;
     }
 
+    // NB: to prevent locking, we use the SUBIMAGE spec, so the mip
+    // information is not valid!!!! instead, we will use the library
+    // which has an internal thread-safe cache of the sizes if needed
     const ImageSpec& spec = init_part(subimage, miplevel);
 
     chend = clamp(chend, chbegin + 1, spec.nchannels);
@@ -1224,11 +1230,15 @@ OpenEXRInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
         return false;
 
 #if ENABLE_READ_DEBUG_PRINTS
-    std::cerr << "openexr read scans " << m_userdata.m_io->filename() << " si "
-              << subimage << " mip " << miplevel << " pos " << ybegin << ' '
-              << yend << " chans " << chbegin << "-" << (chend - 1)
-              << "\n pixbytes " << pixelbytes << " scan " << scanlinebytes
-              << " scans per chunk: " << scansperchunk << std::endl;
+    {
+        lock_guard lock(*this);
+
+        std::cerr << "exr rns " << m_userdata.m_io->filename() << ":"
+                  << subimage << ":" << miplevel << " scans (" << ybegin << '-'
+                  << yend << "|" << (yend - ybegin) << ")[" << chbegin << "-"
+                  << (chend - 1) << "] -> pb " << pixelbytes << " sb "
+                  << scanlinebytes << " spc " << scansperchunk << std::endl;
+    }
 #endif
 
     std::vector<uint8_t> fullchunk;
@@ -1271,8 +1281,8 @@ OpenEXRInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
             for (int dc = 0; dc < decoder.channel_count; ++dc) {
                 exr_coding_channel_info_t& curchan = decoder.channels[dc];
 #if ENABLE_READ_DEBUG_PRINTS
-                std::cerr << " looking for " << cname.c_str() << ": dc " << dc
-                          << " curchan " << curchan.channel_name << std::endl;
+                //std::cerr << " looking for " << cname.c_str() << ": dc " << dc
+                //          << " curchan " << curchan.channel_name << std::endl;
 #endif
                 if (cname == curchan.channel_name) {
                     curchan.decode_to_ptr     = cdata + chanoffset;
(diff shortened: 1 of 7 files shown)


kdt3rd:
lg - I believe I've re-done the locks that'll be used in exrinput_c.cpp to be more conducive to not locking during texturesys, except when the internal cache of the imagespec needs to be built (although we could in theory get rid of that and do the channel remap slightly differently)

src/openexr.imageio/exrinput_c.cpp
```diff
@@ -0,0 +1,1576 @@
...
+    spec = ImageSpec();
+
+    exr_result_t rv = exr_get_data_window(ctxt, subimage, &top_datawindow);
+    if (rv != EXR_ERR_SUCCESS)
+        return false;
+    rv = exr_get_display_window(ctxt, subimage, &top_displaywindow);
+    if (rv != EXR_ERR_SUCCESS)
+        return false;
+    spec.x           = top_datawindow.x_min;
+    spec.y           = top_datawindow.y_min;
```
lgritz, line 538: The diffs in this area make me wonder: Would it be a better choice if the C version uses structs in such a way that you say `top_datawindow.min.x` rather than `.x_min`, and then there's a lot more code that looks identical between the C and C++ versions? And you could even have handy functions (on the C++ side) that convert between the C and C++ equivalents of many of the imath classes.

lgritz (approved):
Looks good. I'm looking forward to trying this out.

src/openexr.imageio/exrinput_c.cpp
```diff
@@ -0,0 +1,1576 @@
...
+        m_local_io.reset();
+        m_userdata.m_io = nullptr;
+        errorf("OpenEXR exception: %s", e.what());
+        return false;
+    } catch (...) {  // catch-all for edge cases or compiler bugs
+        m_local_io.reset();
+        m_userdata.m_io = nullptr;
+        errorf("OpenEXR exception: unknown");
+        return false;
+    }
```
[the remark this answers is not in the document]
kdt3rd: awesome, will clean that out then :)

src/openexr.imageio/exrinput_c.cpp
```diff
@@ -0,0 +1,1576 @@
...
+
+    m_userdata.m_img                = this;
+    exr_context_initializer_t cinit = EXR_DEFAULT_CONTEXT_INITIALIZER;
+
+    cinit.error_handler_fn = &oiio_exr_error_handler;
+    cinit.user_data        = &m_userdata;
+    // KDT TODO: We could change the above to not always create an
+    // ioproxy? Although if we do that, we'll have to change the logic
+    // in the error handler where right now we assume that only
+    // happens in the valid file check
```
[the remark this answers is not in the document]
kdt3rd: yeah, that comment was around when the user does not provide a proxy - I like to think about less code, less bugs as a first course of action. But I suppose if the implementor changes the default behaviour of IOProxy, this will continue to leave this behaving consistently. Given I can remove the exception bits, this will look much simpler after

src/openexr.imageio/exrinput_c.cpp
kdt3rd: i debated that. You can certainly memory cast the structs, and I have unit tests to do so, and with your new constructors to imath, the vectors should just copy construct. I chose the "different" route to drive home that I'm not providing any of the other functionality in the c++ object, but maybe that is not the best way to indicate that

lgritz: It's your code and your call. But I think I would have chosen to make the analogs of the C++ structures organized such that in client code, the API calls obviously would be different between the C and C++ interfaces, but as much as possible of the secondary, more pedestrian code were identical for C and C++ in all the places that they could be  (like how you access struct members). It's not something I'm going to lose sleep over, but because it would change source compatibility, it's pretty much now or never to lock these choices in or rethink them.

BTW, I added the interop utilities only for Vec and Matrix types, but maybe it would be handy for Box, too.

lgritz:
Should we merge this, or do you have other changes coming right away?

Are you brave enough to edit `.github/workflows/ci.yml` to enable this code path in the "bleeding edge" test (which pulls from OpenEXR master) so that this is going through CI henceforth? We'd also need to exclude any testsuite entries that use deep images. If you don't want to mess with that, I can do it after the merge and submit a separate PR.

kthurston:
I can add a CI path for this - wasn't sure whether you wanted to keep it as a separate file (i.e. two implementations)? I can put it as ifdef blocks in exrinput if that would be better? Just thought it was easier to compare side-by-side while we play. Did you have a look at what I did for the tile reads? I never made a multi-tile read call in oiio (only ever single tile at a time), so wasn't sure I interpreted the desired memory layout (interleaved channels, tiles are successive in memory in y, x order. Also - in the new world, if you are at a small mip level (i.e. smaller than the tile size, it only negotiates that tile size in memory inside the C code, and hadn't validated the spec is correct for that yet. - I have the accessors I made to query the sizes from the internal structs in exr, probably should use those to ensure that. Further, wasn't sure that is what the API is expected to be filling in, or if it's expecting to always deliver a full tilesize (i.e. 32x32) regardless of mip level

lgritz:
> wasn't sure whether you wanted to keep it as a separate file (i.e. two implementations)? I can put it as ifdef blocks in exrinput if that would be better? Just thought it was easier to compare side-by-side while we play. 

I think this is great, and at this stage made it easy to diff the two files. Let's keep it separate until it's complete and as we fiddle around, but then eventually when it just does this by default, we'll want to merge this all into one file with appropriate `#if` guards.

> I never made a multi-tile read call in oiio (only ever single tile at a time), 

Putting it into CI will uncover any problems immediately, as there are plenty of tests involving tiled images and the multi-tile reads happen whenever we read whole images rather than single tiles in the texture system.

Also, in `iv`, the `<` and `>` will show you the different MIP levels of an image.

> But at least let's get the exception bits cleaned up - I started that this morning, but meetings are starting now...

Up to you. Nothing wrong with coming back to simplify that part later.

lgritz:
@etheory We are getting close to being able to benchmark the new OpenEXR code for better threaded tile performance.

kthurston:
ok, I'll push up the exception cleanup and the CI path tonight / tomorrow morning, then we can unleash the hounds as they say

kdt3rd:
got things patched up, started running tests, there are a couple of problems, let me get those patched up prior to merging this in

kdt3rd:
fyi, when I run the tests, I see a number of failures with tiff and with freetype, it seems to use a different font on my system than for you (ubuntu?) by default. However, there are a couple of bad / incomplete tests that fail because the error messages are different. I'm happy to update the references, but wanted you to have a look prior to doing so to make sure I am interpreting it correctly.

kdt3rd:
additionally, I did visually audit the texture sys chain, so other than the lock that happens when you first access a subimage / miplevel to update the ImageSpec cache in the input file, I believe all the locking should be removed, so should be ready for some baseline tests. Will love to hear how it helps, or if there are other issues.

lgritz:
Freetype has the maddening property of rasterizing fonts ever so slightly differently from version to version. This is one case where we tend to just check multiple reference images into the testsuite/NAME/ref directory (the testsuite scripts will allow a pass if the output matches *any* of the reference images in that directory). If your Linux distro has a different version of Freetype than any of the ones we've encountered in CI or our personal machines, it may indeed differ in output. Feel free (maybe in a separate PR?) to add new reference output for your version of freetype. You can see how we usually name the reference files to make it clear which freetype versions required new reference images.
```
$ ls testsuite/oiiotool-text/ref
aligned-freetype2.4.11.tif  text-freetype2.7.tif
aligned.tif                 text.tif
out.txt                     textalpha.tif
text-alt.tif                textshadowed-freetype2.4.11.tif
text-freetype2.6.5.tif      textshadowed.tif
```

kdt3rd:
fyi, the image tests that OpenImageIO does has uncovered an issue with the piz decompressor under C beyond the current unit tests in the exr tree. Ignoring that, I believe I have fixed the issues I've found with the texture reading in OpenImageIO and the new C library.

I will fix the piz decompression tomorrow and wire in deep tomorrow. But I don't want to push up reference output changes from the different error text, so we'll continue to have errors even after I fix those.

side note: I found out why the text tests were failing under Arch: I didn't have the droid fonts installed, so it wasn't using DroidSans. So seems like the newer freetype (2.10.4) hasn't changed much. I don't seem to have the correct DroidSerif font for one of the oiiotool tests (also not found), but not going to worry too much about that right now.

lgritz:
The Droid fonts are installed as part of OIIO itself. I suspect that if you only build (but don't install) and then run tests, maybe it can't find them. Try just setting OpenImageIO_ROOT to wherever you have your test install of OIIO, and that should be enough to find them. I'll try to make the testsuite itself a little more robust to not needing that set if you have a build area.

kdt3rd:
@lgritz - There may still be issues, but this (with latest openexr master) seems close enough for now to enable wider testing

kthurston:
Huh, I see that unit_imageinout is crashing, will take a look at that tonight. I believe the other two failing tests are differences in the error messages.

kdt3rd:
good to go whenever I think - the C layer isn't designed to read non-chunk-sized blocks, but think I've fixed the various corner cases I've seen

lgritz:
I have a "fix" for the error messages. By which I mean, a patch that adds an additional reference output for it to match to for these two tests. I'm testing it now and will push to your branch if it succeeds.

But in the mean time, I do urge you to look at the new error output reference versus the old one from the same tests. IMHO, the old error messages are MUCH more understandable (by the users who will see them as passed up via oiio) than the new ones. For example,

Old: `Failed OpenEXR read: Error reading pixel data from image file "src/incomplete.exr". Scan line 126 is missing.`

("Oh, a scanline is missing and I know which one. It's an incompletely written file.")

New: `EXR Error (src/incomplete.exr): EXR_ERR_BAD_CHUNK_DATA Request for scanline 0 found scanline 20000630, not 0 at chunk 0`

("What is a chunk and why does it think there is a scanline 20000630?")

lgritz:
With the extra output reference checked in, CI is fully passing, so I'm going to merge.

Related items to-do:

* Benchmark to see if this has the performance gains (especially for texturing) that we hope for.
* Contemplate when we are comfortable using the core api automatically any time OpenEXR 3.1 is found.
* Should exrinput.cpp and exrinput_c.cpp be merged, with appropriate `#if`, rather than separate files? (Pro: will eliminate headache of edit in two places as other OIIO things change or we improve openexr handling. Con: possibly just as confusing to have to do it in two clauses of an `#if` and may look cluttered.)

PR #3012, opened 2021-06-26
Add a scaled slopes UV normalization options for --bumpslopes.

lecocqp:
## Description

This patch provides an optional parameter `--uvslopes_scale` for the `--bumpslopes` exporter to output the derivatives in the normalized UV space and scale them by a user defined constant. 

Originally, we decided to store the derivatives in texel space rather than the UV space to prevent floating point overflow when using the half float format (67c863d). However, expressing the derivatives in UV space is interesting because it is resolution independent and convenient when the resolution is not accessible. We restore this ability thanks to this optional parameter and a scale factor to prevent the overflow when using the half format. A typical value of 256 is recommended to prevent half floating point overflow with texture resolutions up to 32k. The value is stored as an extra attribute in the output file to revert the scaling on slopes when the texture is used. Note that this options is only meaningful with height map inputs.

## Checklist:

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -469,7 +469,7 @@ bump_to_bumpslopes(ImageBuf& dst, const ImageBuf& src,
     float res_x = 1.0f;
     float res_y = 1.0f;
 
-    const std::string& bumpformat = configspec.get_string_attribute(
+    string_view bumpformat = configspec.get_string_attribute(
         "maketx:bumpformat");
 
     if (Strutil::iequals(bumpformat, "height"))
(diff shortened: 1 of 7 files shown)


src/libOpenImageIO/maketexture.cpp
```diff
@@ -466,6 +466,12 @@ bump_to_bumpslopes(ImageBuf& dst, const ImageBuf& src,
 
     bump_filter = &sobel_gradient<SRCTYPE>;
 
+    float res_x = 1.0f;
+    float res_y = 1.0f;
+
+    const std::string& bumpformat = configspec.get_string_attribute(
```
lgritz, line 472: I'm a little suspicious of this line. get_string_attribute returns a string_view, not a `string&`, so I think this is making a temporary std::string and then having bumpformat be a const ref to it? But the lifetime of that string is a little worrying. I would maybe sleep better at night if this line were

    string_view bumpformat = configspec.get_string_attribute ...

lgritz (approved):
LGTM, with one minor optional suggestion.

src/libOpenImageIO/maketexture.cpp
lecocqp: fixed in 6736d236ae375501eaf5a6bf6fb698c92746d04c

lgritz:
I think the remaining windows failures are spurious and unrelated to your changes, so I will merge.

PR #3013, opened 2021-06-26
Convert dpx output to be ioproxy enabled

kdt3rd:
Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>

## Description

Converts dpx output (input is already) to use the IO proxy layer (I recently
had a reason to write dpx, but couldn't make a memory stream).

## Tests

just tested with dpx image tests

## Checklist:

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -226,12 +226,17 @@ control aspects of the writing itself:
        will keep unaltered pixel values (versus the default OIIO behavior
        of automatically converting from RGB to the designated color space
        as the pixels are written).
+   * - ``oiio:ioproxy``
+     - ptr
+     - Pointer to a ``Filesystem::IOProxy`` that will handle the I/O, for
+       example by writing to memory rather than the file system.
 
 **Custom I/O Overrides**
 
-DPX input (but, currently, not output) supports the "custom I/O" feature
-via the `ImageInput::set_ioproxy()` method and the special
-``"oiio:ioproxy"`` attributes (see Section :ref:`sec-imageinput-ioproxy`).
+DPX input and output both support the "custom I/O" feature via the
+special ``"oiio:ioproxy"`` attributes (see Sections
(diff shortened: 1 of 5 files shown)


lgritz:
Did you clang-format OutStream.h by hand? We had excluded it from "make clang-format" because it's imported from another project and we wanted to retain the ability to easily diff between them. I'm tempted to recommend reverting the format-only changes to existing code.

lgritz:
But I'm not really going to get bent out of shape about it. In truth, it has been extremely rare over the years for us to need to diff these files versus the originals. And with IOProxy, the files really do have differences, whereas they used to be identical.

lgritz:
Looks like CI is failing the "imageinout_test", which (among other things) for all file types supporting proxies, will write a file and write a proxy and then verify that they are byte for byte identical.

kdt3rd:
I did reformat the outstream when I did the pull request because I looked at the code after the comment in the PR template - and it wasn't following the same style, so I thought it just hadn't been visited in a while to have been formatted. Happy to revert that if you'd prefer.

I ran ctest (i.e. ninja -C <buildfolder> test) and that succeeded for me, I wonder what I'm missing that imageinout_test doesn't run...

lgritz:
Try running by hand, it lives in `BUILDAREA/bin/imageinout_test`

lgritz:
I mildly prefer that you remove the formatting change, but only if it's super easy. It's not worth the hassle of extra work, I'm just being superstitious about wanting "imported" source to restict diffs versus upstream to only the things we purposely change.

kdt3rd:
ah, I forgot dpx puts the filename in the header. I've changed the test  write proxy to just use the disk filename, do you know of any issues with that? I believe I reverted the change to formatting OutputStream.cpp

lgritz:
That sounds reasonable. In the imageinout_test we are also careful to set the DateTime attribute to something fixed so that files don't differ because of that metadata.

lgritz:
Would you mind also fixing builtinplugins.rst to note that DPX takes proxies for output as well as input now?

lgritz (approved):
LGTM.

Issue #2902, opened 2021-03-19
[FEATURE REQUEST]Read as little as possible from image during 'open' call


DougRogers:
When scanning a drive for images, I am interested in gathering all images that are larger than a specific size.  During the open call, more info than just the header is read in, slowing the call significantly.  Calling open on thousands or millions of images is very time-consuming.

For example, in 
```
auto in = ImageInput::open (filename);
if (! in)
    return;
const ImageSpec &spec = in->spec();
int xres = spec.width;
int yres = spec.height;
```
the TGA processing reads in the thumbnail data.

I would like to have all pixel processing, even the thumbnails, removed from the open call and just have the minimal spec information read so that this call is performant.

lgritz:
This is an interesting topic, and it's not obvious what the right approach is, so I am definitely interested in feedback from you and anybody else reading.

So, just as background, a common theme in OIIO's APIs is that the whole image can be separated into the pixel data itself, and everything else is stored in ImageSpec. The idea is that the result of a full read cycle is an ImageSpec + a pixel buffer, and so when you write an image, you just pass an ImageSpec and a pixel buffer. The specific thing we tried to avoid was having to read item after item through separate API calls, have the user code carefully tend to those separate pieces of data, and pass them though a succession of special API calls when writing or copying one image to another.

So for that reason, I historically had hesitated to make a separate call to read the thumbnail, or to write it. And it's definitely a priority for me that the simple "read an image from one file, write it out to a different file" be really, really simple, a small number of simple API calls.

But I also see that there are times when you don't want all that expense. And thumbnails aren't the only case like this: ICC profiles can be quite big. PTex files have a big set of information giving the connectivity of the mesh. OpenEXR just merged a patch that adds (enormous!) metadata that associates a manifest of object names with integer "object ID" images. So I think this issue will keep popping up in various forms and that there is merit to addressing it intelligently.

So I've been mulling this over for a few days, and I can think of a few strategies we might take here. These are not necessarily mutually exclusive and are in no particular order:

* Add a new hint for the "open with config" call, which indicates that any "big/expensive" metadata should not be read or stored in the spec. You'd tend to use this hint only for cases when you were opening the image for the sole purpose of getting the basic resolution and type info. This has the nice property of not breaking any existing code or ABIs, it's optional (for the author of the format reader) and we can implement it selectively, and it's very straightforward.

[...]

PR #3021, opened 2021-07-02
Overhaul of how to read/write thumbnails

lgritz:
A few image formats (but not most) support "thumbnails" -- low-res preview
images.

Traditionally, we have shoved them in the ImageSpec metadata as a byte
array (and ints for the width, height, nchannels). This is less than
ideal: It makes the metadata very bloated, taking more memory and
being more expensive to copy metadata, and potentially wastes time
doing I/O on this data that most applications don't want or
need. (Remember that any alterations to the image will make the
thumbnail "wrong", also. So it's rarely useful except for applications
that specifically want a preview image.)

Discussion in Issue #2902 led us to think that thumbnails should be
treated explicitly by the API, rather than shoved into the metadata in
this way. This patch is the attempt to overhaul thumbnail handling:

* ImageInput has a new `get_thumbnail()` method that can retrieve the
  thumbnail (if it exists) into an ImageBuf. The base class / default
  implementation does nothing; the image formats that support
  thumbnails should have their readers override this method. They are
  encouraged to read the thumbnails "lazily" if possible -- only
  if/when this method is called, and avoid the cost, memory, and I/O
  if not requested.

* ImageOutput likewise has a new `set_thumbnail` method.

* These are implemented for Targa and PSD, the only two formats for
  which we previously had implemented thumbnail functionality. (There
  are probably other formats that support thumbnails, but for which we
  never added OIIO support. Leave that for another day.)

* ImageBuf has new `has_thumbnail`, `thumbnail`, `set_thumbnail`, and
  `clear_thumbnail` methods.

* ImageCache has a new `get_thumbnail` method.

* oiiotool and iconvert attempt to preserve a thumbnail from input to
  output (but of course, only if both file types support it).

* ImageBufAlgo::IBAPrep clears the thumbnail in the destination image,
  on the theory that pretty much everything IBA would ever do will
  likely invalidate any preview image that it started with. This gets
  us to the point with oiiotool where a straight copy (like a format
  conversion) `oiiotool in.tga -o out.tga` will preserve the thumbnail,
  but just about anything else `oiiotool in.tga --stuff -o out.tga`
  will not transmit the original thumbnail to the output (which would
  almost certainly be wrong).

There is more work to do in the future. Some file formats that support
thumbnails don't yet have OIIO support for it. I'm not sure we ever
really want (except in the case of a truly straight copy) for oiiotool
to keep any input thumbnails; maybe if a thumbnail is wanted for
output, it should always be created anew (by downscaling) so that it
definitely matches the main image? Maybe also oiiotool needs a command
to explicitly extract and save the thumbnail (is that needed for any
legit use? or only as a debugging tool?).


apetrynet:
These changes make sense to me :) 

One question: 
I'm sure there's a valid reason for this, but `ImageInput` and `ImageCache` have methods named `get_thumbnail`,  why is the "same" method named simply `thumbnail` in `ImageBuf`?

lgritz:
A function/method whose primary task is to return a value is usually given a NOUN name that describes the thing.

    string name();  // returns the name

A function/method whose task is to performs a procedure (it may be void, or it may have a return value that is incidental, like an error code) is usually given a VERB name that describes the procedure.

    void get_name(string& name_storage); 

Now, this isn't the kind of rule that you make the hill you're going to die on. The fact is that if you swapped those names around, people would still understand what you were doing in the code. I'm just justifying how lots of people pick function names as parts of speech, all other things being equal.

In this case, `thumbnail()` actually returns an image, whereas the `get_thumbnail()` varieties don't return a thumbnail, they copy the thumbnail into memory referenced by a pointer passed by the user.

Now, I'm justifying how I wrote it this way, but not necessarily defending it very hard. I stand by it only 70%. If people think inconsistency in naming is worse than inconsistency in grammar (and I did notice and struggle with the choice in this case), I can change the name to get_thumbnail without regret.

apetrynet:
Thanks for the detailed answer! All valid points.
I have faith in your judgement and am fine with the way it is now. Don't change the name if I'm the only one who reacted slightly to it.

One more thing. (Apologize if this is obvious)
I noticed the definition of `get_thumbnail` is commented out in `py_imagecache.cpp`, is that deliberate?

lgritz:
It's one of many things that are not exposed by the python bindings of ImageCache, because it doesn't map to Python well, so I'm deferring until (if every) it appears that somebody really needs to use it from Python. I put it in there as a comment just to remind myself where it should be, if we implemented it.

lgritz:
I took @apetrynet's comments to heart, updated the PR with the IB method called get_thumbnail to match the others. I figured that it's probably easier for people if the analogous method is called the same thing in all the classes, even if they take differing parameters.

nrusch:
Just out of curiosity, how many formats does OIIO support that support embedded thumbnails (regardless of whether or not the actual thumbnail reading is implemented in OIIO)?

lgritz:
@nrusch Not a lot do. PSD and Targa do, which we support. Apparently OpenEXR does! Which I did not know until recently, and so have never supported it. The fact that nobody has complained in 13 years probably speaks to the frequency with which exr files in the wild have thumbnails (that anybody would miss if they disappeared when travelling through OIIO).

malcolmhumphreys:
Would it be possible to support multiple thumbnails of different
resolutions in the API? many raw formats embed multiple resolutions of
thumbnails. These are super handy and fast way to preview images before
doing any slower processing.

It would be great to be able to do this directly in the API vs running a
tool like exiv2 to extract all the previews and then pick a low resolution.

$ exiv2 extract -e p IMG_1892.CR2

$ exiv2 -h

   -e tgt  Extract target(s) for the 'extract' action. Possible targets
           are the same as those for the -d option, plus a target to extract
           preview images and a modifier to generate an XMP sidecar file:
             p[<n>[,<m> ...]] : Extract preview images.

https://benjaminhorn.io/code/extracting-thumbnails-from-camera-raw-files-cr2-and-nef-with-php/

// This commands shows the available previews that are embedded
exiv2 -pp IMG_2205.CR2
// Returns something like:
Preview 1: image/jpeg, 160x120 pixels, 15888 bytes
Preview 2: image/tiff, 668x432 pixels, 1731456 bytes
Preview 3: image/jpeg, 5184x3456 pixels, 2235171 bytes


Malcolm

lgritz:
Thanks, @malcolmhumphreys, that's the kind of feedback I needed to test my design assumtions.

My belief was that there would be at most one thumbnail, and that it would be substantially lower resolution (preview 1 in your example is what I imagined).

I think there are two ways to handle this, and am curious about your opinions.

One approach could be to fully generalize the thumbnail interface. We could extend the proposed thumbnail API to... I guess tell you how many thumbnails there are and their resolutions, let you choose which one (or perhaps request a resolution and it returns the closest?). That seems like a lot of client-side complexity -- instead of "give me the thumbnail (or nothing)", now we're talking a protracted back-and-forth, do you have thumbnails, tell me the inventory of all available, oh boy which one do I want, ok give me the 160x120 one."

The alternate approach is to stick to the original assumption -- a single thumbnail that's assumed to be fairly low res -- and for formats like the raw example where there were 3 different preview images (up to 5184x3456!), maybe thumbnail isn't the right interface at all. Perhaps in that case, the raw reader ought to be considering the file to be either multi-image (subimages) or multi-rez (like a MIPmap, though sparser -- our MIP-maps aren't strict about the multiresolution representation being a power of 2)?

(For the sake of exhaustively exploring the design space, I'll point out that yet another approach is to get rid of the idea of thumbnails entirely, and utilize the existing concept of multiple representations of the same image at different sizes -- MIPmaps. Even if there was just one thumbnail, we could demand that format readers present thumbnails or preview images as if they were MIP-maps, even if there are only two levels spaced quite widely apart. Hmmm, what happens / is it possible, to have an image that is MIPmapped and also has a thumbnail? Would anybody ever do that?)

SRHMorris:
About the EXR support lacking thumbnails, actually I had missed this feature. I always meant to submit a patch to implement this but never got around to it. When you have to load many files at a time and display them to a user thumbnails definitely help performance a huge amount.

While I'm here, one downside of implementing it as mipmaps is that often the embedded thumbnail is a different format to the overall file type (e.g. an embedded jpeg in a psd). It probably doesn't matter for my use case, but could that matter to someone, going from a lossless to a lossy format?

lgritz:
@SRHMorris Once this is in, we can easily add thumbnail support for exr. Thanks for letting us know it's something that was missed.

The codec (lossless or not) is already allowed to vary between MIP levels, I don't know why that would cause a problem.

It is true that a thumbnail might always be 3 channel uint8, whereas the main image may have a different data format or number of channels. That's something to watch out for, if software that is MIPmap aware makes the assumption that data format or nchannels can't change between MIP levels. I don't think we ever promise that this has to be the case. The TextureSystem does assume and rely on it! But not all MIPmaps are textures, they have to be marked as such, and so we wouldn't mark multi-preview images as textures.

Remember, we're still talking about a single ordinary thumbnail being accessed with special thumbnail methods. The MIPmap idea was just a possibility for how we might handle a file format that allows many preview images. I think DSLR  camera raw is the only one where that's a possibility, and one would never use that as a texture.

lgritz:
@malcolmhumphreys, @SRHMorris, @Shootfast, and anybody else who cares how we handle thumbnails, particularly for camera raw formats:

The deadline for branching 2.3 and locking down its API is fast approaching. I see the following possibilities and am interested in your opinions. In no particular order,

1. Let the deadline pass, keep everything as it is for 2.3. Rather than be hasty, we'll figure this out at our leisure for 2.4 (down side: won't be in a stable supported release until next year).

2. Commit this new proposed API for a single thumbnail image, which covers all cases except for multi-preview raw files (which we don't handle at all now). Later on, we can figure out what to do for that case, which can be added at any time if it doesn't further change APIs (like if we treat them as MIPmaps), or as a 2.4 feature if it requires non-back-compatible API changes.

3. *Quickly* CHANGE or ADD TO the proposal to handle the multi-preview case -- which I only want to do if it's clear how it should be used and isn't a huge burden on client software to do something complicated. I don't want the common case of single low-rez thumbnail which is now a single, simple call to retrieve, into a protracted series of calls or complex logic to figure out how many thumbnails there are, what their resolutions are, which one is desired, retrieve that one, etc.

SRHMorris:
From my perspective the proposed api (single call to get_thumbnail() or whatever) sounds like it fulfills the common usecase. I was under the impression that the majority of camera raws just use a single embedded jpeg, but could be wrong. Certainly it roughly fits with the way my code currently works via the spec attribute!

That said I don't mind if you don't want to rush.

Shootfast:
To me option 2 sounds ok (I like the API as is), but am not screaming out for it, so option 1 is also ok.
I'm not sure how useful a multi thumbnail API is (especially seeing how little thumbnails currently are used), but I wouldn't be unhappy if we eventually morphed the API towards option 3, I just think it's a good "later" feature.

lgritz:
OK, I'm leaning toward option 2 -- let's accept this PR as a clear step forward for 2.3 that handles the common case simply, and later we can assess if there are enough exceptions to warrant refinement of the approach in future releases.

nrusch:
I'm also in the boat of not really needing this feature myself, but I would agree that option 2 seems like a good step forward.

PR #3025, opened 2021-07-06
Allow oiiotool to print metadata of computed images

lgritz:
The existing semantics of `oiiotool -info`, `-info -v` and `-stats`
are that these options set flags so that each time a file is opened
for the first time, it will print the metadata and/or stats. But these
do not allow you to create new images and print their metadata or stats
(at least, not without writing to disk and reading those files).

This patch extends the expression substition syntax to recognize the
following new metadata notations:

* `METABRIEF` : a string containing the brief one-line description that
  would be printed with `oiiotool -info`.
* `META` : a multi-line string containing the full metadata that would
  be printed with `oiiotool -info -v`.
* `STATS` : a multi-line string containing the image statistics that would
  be printed with `oiiotool -stats`.

Some interesting use cases are illustrated with these examples:

    # Roughly the same behavior as `oiiotool -info img.exr`:
    oiiotool img.exr -echo "{TOP.filename} : {TOP.METABRIEF}"

    # Perform computations on an image and print statistics about
    # the computed image (not the original):
    oiiotool img.exr -blur 3x3 -echo "blurred stats: {TOP.STATS}"

    # Print statistics on intermediate results... and keep going with
    # more operations:
    oiiotool fg.exr bg.exr -over -echo "{TOP.STATS}" -o comp.exr

    # Write metdata as "burn-in" on an image:
    oiiotool in.exr -text:x=10:y=20 "{TOP.META}" -o burned.exr

The implementation in the business logic of oiiotool.cpp was very
simple.  The more extensive change behind the scenes to make this
possible was a massive refactoring of printinfo.cpp, which previously
just wrote everything to stdout in a totally undisciplined fashion,
and transforming it to very carefully write to output streams instead,
which could be passed a stringstream to render the metadata listing or
stats output as a string. Also needed some refactoring to allow it to
take either the name of a file to read anew (as it did before), or an
existing ImageBuf/ImageRec (to support the new feature).

### DIFF
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -225,6 +225,12 @@ contents of an expression may be any of:
     comma-separated)
   * `AVGCOLOR` : the average pixel value of the image (channels are
     comma-separated)
+  * `METABRIEF` : a string containing the brief one-line description that
+    would be printed with `oiiotool -info`.
+  * `META` : a multi-line string containing the full metadata that would
+    be printed with `oiiotool -info -v`.
+  * `STATS` : a multi-line string containing the image statistics that would
+    be printed with `oiiotool -stats`.
 
 * *imagename.'metadata'*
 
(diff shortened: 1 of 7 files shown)


lgritz:
This is for you, @christopherdavies

Issue #2956, opened 2021-04-29
[FEATURE REQUEST] Make finding Boost a bit more flexible


zomgrolf:
**Is your feature request related to a problem? Please describe.**
When building OIIO on Windows we tend to link almost all of its dependencies statically (i.e. with `LINKSTATIC` set to true). However, in some cases we want to link Boost dynamically (while other dependencies are still linked as static libs). While it's not too difficult to set up, it still requires some wrangling with the build system and I'd really like to make it a bit more straightforward.

**Describe the solution you'd like**
There is the `BOOST_CUSTOM` option, but using it requires carefully setting up all the variables, rather than letting the FindBoost module take care of that. Looking at [externalpackages.cmake](https://github.com/OpenImageIO/oiio/blob/f46c8b9a0f62738f1699c57a73368772c71f3b6c/src/cmake/externalpackages.cmake#L39) a fairly small change would enable the scenario I'm interested in (being able to choose the way Boost is linked independently from the `LINKSTATIC` setting) without the need for the fully manual setup.
The script uses the `Boost_USE_STATIC_LIBS` variable to tell FindBoost which type of libraries to look for. If `LINKSTATIC` is true, then it gets set to true as well, which means FindBoost will try to find static boost libraries. If it's not true, it seems to assume that `Boost_USE_STATIC_LIBS` is false, but that's not necessarily the case -- someone could still set it to true. 

I'd like to propose the following change: check if `Boost_USE_STATIC_LIBS` has already been defined, and if not, set it to the current value of the `LINKSTATIC` variable. Then, based on the value of `Boost_USE_STATIC_LIBS` add the necessary definitions for the MSVC compiler:
```
if (NOT DEFINED Boost_USE_STATIC_LIBS)
    # Set Boost_USE_STATIC_LIBS to true, but only if linking dependencies statically
    if (LINKSTATIC)
        set (Boost_USE_STATIC_LIBS ON)
    endif()
    # Or perhaps: set (Boost_USE_STATIC_LIBS "${LINKSTATIC}")	
endif ()

if (MSVC)
    if (NOT Boost_USE_STATIC_LIBS)
        add_definitions (-DBOOST_ALL_DYN_LINK=1)
    endif ()
endif () 
```
[...]

PR #3031, opened 2021-07-14
Make finding Boost more flexible.

zomgrolf:
## Description

Addresses #2956. Apologies for the long radio silence on this, I've been quite busy lately...

This patch modifies the build system to make it possible to control
whether OIIO should link Boost in the form of static or dynamic
libraries, independently from the LINKSTATIC setting, without
having to use the fully custom route (BOOST_CUSTOM).

This is done by preserving the value of the Boost_USE_STATIC_LIBS
variable, if it has been defined by the user. If the value
of Boost_USE_STATIC_LIBS has NOT been explicitly defined, it is
set to match the value of the LINKSTATIC variable, which should
result in Boost being linked in the same way as other dependencies.

## Tests

The existing CI workflows should (hopefully) be able to detect any anomalies in the common build scenarios.
To test this specific scenario a new CI workflow would be needed, but it seems such a niche use case, that I don't really expect it to be included in the standard test suite.

## Checklist:

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -36,13 +36,20 @@ if (MSVC)
     # Disable automatic linking using pragma comment(lib,...) of boost libraries upon including of a header
     add_definitions (-DBOOST_ALL_NO_LIB=1)
 endif ()
-if (LINKSTATIC)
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM, thanks

PR #3035, opened 2021-07-25
ImageCache: add a variety of invalidate() that takes an ImageHandle*

### LINKED ISSUE #3034: [FEATURE REQUEST] Add overload of ImageCache::invalidate(ImageHandle*)
ImageSpec* ImageCache::imagespec(ImageHandle*, ...) returns a pointer to an ImageSpec that is valid until the ImageCache is destroyed, or until invalidate() is called on the file that the ImageHandle refers to. In the Rust bindings I could enforce this in the type system by consuming the ImageHandle when calling invalidate(), only there's no overload of invalidate() that takes an ImageHandle* instead of a ustring. 

I would like an overload of ImageCache::invalidate that takes an ImageHandle* 

Unless I'm missing it, there doesn't appear to be a way of getting the filename from the ImageHandle, in which case my only option would be to store the filename in my ImageHandle wrapper itself (not the end of the world I suppose).

**Additional context**

lgritz:
Most of the rest of the IC methods that take a filename, also have
a version that takes a handle.

Closes #3034

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -821,6 +821,10 @@ public:
     /// since it was first opened by the ImageCache.
     virtual void invalidate(ustring filename, bool force = true) = 0;
 
+    /// A more efficient variety of `invalidate()` for cases where you
+    /// already have an `ImageHandle*` for the file you want to invalidate.
+    virtual void invalidate(ImageHandle* file, bool force = true) = 0;
+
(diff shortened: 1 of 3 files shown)


lgritz:
@anderslanglands Is this what you need?

lgritz:
Closes #3034

anderslanglands:
Yes that looks like it, thanks Larry!

PR #3036, opened 2021-07-25
imagebuf.h: remove inclusion of imagecache.h

lgritz:
I noticed when working on an ImageCache thing that even the smallest
change to imagecache.h was causing a substantial amount of the
codebase to needlessly recompile even though very few places really
need to use the IC APIs directly.

It took a little bit of juggling of some declarations, but this patch
allows imagebuf.h to be able to treat ImageCache and ImageCacheTile as
opaque, forward-declared types, without the need for every file that
needs imagebuf.h to also have to parse imagecache.h and everything it
might pull in.

This speeds up my development cycle substantially in many cases, and
it also speeds up compilation of user code (by a small amount) by
avoiding parsing of imagecache.h in circumstances where the user code
only needed to know about ImageBuf but not ImageCache.

There's a downside to this -- it's possible that user code that truly
needs both imagebuf.h and imagecache.h only includes the former
(inadvertently relying on the fact that the latter was transitively
included). These apps may need `#include <OpenImageIO/imagecache.h>`
to be added after upgrading to OIIO 2.3. Is that really really bad?

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -14,7 +14,6 @@
 
 #include <OpenImageIO/dassert.h>
 #include <OpenImageIO/fmath.h>
-#include <OpenImageIO/imagecache.h>
 #include <OpenImageIO/imageio.h>
 
 #include <limits>
@@ -25,6 +24,11 @@ OIIO_NAMESPACE_BEGIN
 
 class ImageBuf;
 class ImageBufImpl;  // Opaque type for the unique_ptr.
+class ImageCache;
+
+namespace pvt {
+class ImageCacheTile;
+};  // namespace pvt
 
 
 
@@ -1219,7 +1223,7 @@ public:
         ~IteratorBase()
         {
             if (m_tile)
-                m_ib->imagecache()->release_tile(m_tile);
+                release_tile();
         }
 
         /// Assign one IteratorBase to another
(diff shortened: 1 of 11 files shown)


EvanAW:
> These apps may need `#include <OpenImageIO/imagecache.h>`
> to be added after upgrading to OIIO 2.3. Is that really really bad?

In my opinion, no, of course not. Adding some `#include`s is a minor user inconvenience compared to the benefit of speeding up the development cycle of the developer who contributes 99% of the code, and therefore increases the speed of the development cycle of library we all get to use for free. I can’t see anyone rationally weighing the cost vs the benefit (assuming that I’m not missing some large factor) and being against this change.  

Just document the possibility of needing to add said `#include`s in 2.3’s release notes, both in the change log and in an additional “Upgrading to 2.3” section if you like, and you’re covered in terms of what I expect as a user.

lgritz:
Oh, thanks. It's not about inconvenience per se, just that generally we try hard to make it so that 2nd-digit changes (say, 2.3 -> 2.4) is ok to need recompile, but we try very hard not to require any changes to downstream source. Whereas 2.4 -> 3.0 might be expected to have downstream source changes required.

`#include` is a little bit of a gray area. It's SO minor, it's a change to the source file, but it's not a change to the call points.

I'm leaning toward merging, thanks for the feedback. I'll give people another day or two to object.

PR #3038, opened 2021-07-27
Let CUE_THREADS help determine default number of threads for OIIO

lgritz:
OpenCue-launched apps are encouraged to honor the setting of
CUE_THREADS to limit the number of threads they use. This change makes
oiiotool a better citizen when run as part of an OpenCue job.

CUE_THREADS has lower precedence than the more OIIO-specific
`OPENIMAGEIO_THREADS`.

For ordinary command line usage, this should not change behavior, since
you wouldn't expect CUE_THREADS to be set.

(Also fix docs which incorrectly named `OPENIMAGEIO_TEXTURE_OPTIONS`.)

### DIFF
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -316,7 +316,7 @@ inside the source code.
         imagecache->attribute ("options", value);
 
 
-``OPENIMAGEIO_TEXTURESYSTEM_OPTIONS``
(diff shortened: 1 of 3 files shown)


lgritz:
I know this will fail CI, until #3037 is merged (which this depends on). I wanted to separate that into a different PR.

PR #3045, opened 2021-08-01
RAW plugin improvements

Shootfast:
## Description

This PR introduces a number of fixes/updates for the RAW plugin.

1. It fixes the output when raw:Demosaic is set to "none". Previously pixel output was indexed incorrectly into the raw buffer and did not take into account rotation or possible "junk" pixels.
2. Implements raw:balance_clamped mode that will apply a per-channel clamp on all pixel values that are over the saturation level of the sensor. This eliminates issues where clipped highlights take on an unwanted hue shift. This is especially useful when stacking HDRI exposures.
3. Implements raw:apply_scene_linear_scale and raw:camera_to_scene_linear_scale, which allows users to scale libraw output to expected levels. The default scale factors are currently hard coded, but experimentally these settings have proven "good enough" over many years of use.

## Checklist:

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -66,6 +66,8 @@ private:
     bool m_unpacked = false;
     std::unique_ptr<LibRaw> m_processor;
     libraw_processed_image_t* m_image = nullptr;
+    bool m_do_balance_clamped = false;
+    float m_balanced_max = 1.0;
     std::string m_filename;
     ImageSpec m_config;  // save config requests
     std::string m_make;
@@ -434,9 +436,6 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // Disable exposure correction (unless config "raw:auto_bright" == 1)
     m_processor->imgdata.params.no_auto_bright
         = !config.get_int_attribute("raw:auto_bright", 0);
-    // Use camera white balance if "raw:use_camera_wb" is not 0
-    m_processor->imgdata.params.use_camera_wb
-        = config.get_int_attribute("raw:use_camera_wb", 1);
     // Turn off maximum threshold value (unless set to non-zero)
     m_processor->imgdata.params.adjust_maximum_thr
         = config.get_float_attribute("raw:adjust_maximum_thr", 0.0f);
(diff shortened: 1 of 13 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -612,6 +660,114 @@ RawInput::open_raw(bool unpack, const std::string& name,
...
+            float unadjusted = m_processor->imgdata.color.maximum;
+
+            // Set the max threshold to either the default 1.0, or user requested max
+            m_processor->imgdata.params.adjust_maximum_thr =
+                (old_max_thr == 0.0f) ? 1.0 : old_max_thr;
+
+            // Get new max value
+            if (m_processor->adjust_maximum() != LIBRAW_SUCCESS){
+                errorf("HighlightMode maximum adjustment failed");
+                errorf(libraw_strerror(ret));
```
lgritz, line 753:
```suggestion
                errorf("%s", libraw_strerror(ret));
```

lgritz (changes requested):
Code LGTM other than the one problematic errorf call that I caught.

Needs clang-format (see log of the clang-format CI test). I can do that on my end if you want.

Output reference need updating, but I can add that myself as the last step.

lgritz:
Also: would you mind adding any new configuration hints to the documentation in the RAW section of builtinplugins.rst? Thanks.

Shootfast:
Great, I'll update the docs and fix that error!

lgritz:
I pushed additions that fix the formatting and update the testsuite reference output.

Shootfast:
I've fixed the errorf calls and updated the docs

lgritz:
Thanks for adding the docs, this is fine and I'm ready to merge.

But just for my own edification, I'm a little confused about how these controls work and have a few questions. (Which may or may not prompt you to want to change something, it's up to you. I'm primarily indicating that I don't fully understand how these are used.)

1. If apply_scene_linear_scale merely enables the scale of camera_to_scene_linear_scale, why have two controls? Would it be simpler if camera_to_scene_linear_scale served both purposes, where it defaults to 1.0, in which case no scale is applied, and any other value both enables the scaling and also specifies the scale factor?
2. I wonder if the "changes output datatype to half" is confusing. A subsequent call to `read_scanlines()` et al still lets you ask for any data type you want. It may be that you are (internally) retrieving the data as half, but that's invisible to the user. Maybe it would be more clear if you said "enabling this option might result in pixel values that might be > 1.0, thus you should retrieve pixels as half or float values to ensure that values are not clipped by a data type conversion to integer types."

lgritz (approved):
LGTM, but I'll wait to merge until you see my minor questions, just in case they prompt you to want to change anything.

Shootfast:
>     1. If apply_scene_linear_scale merely enables the scale of camera_to_scene_linear_scale, why have two controls? Would it be simpler if camera_to_scene_linear_scale served both purposes, where it defaults to 1.0, in which case no scale is applied, and any other value both enables the scaling and also specifies the scale factor?

I am not expecting people to need to customize the scale value very frequently, nor expecting them to remember the magic number of 1.0 / 0.45. So I was planning on just saying "use apply_scene_linear_scale" in my general guidelines.
 
>     2. I wonder if the "changes output datatype to half" is confusing. A subsequent call to `read_scanlines()` et al still lets you ask for any data type you want. It may be that you are (internally) retrieving the data as half, but that's invisible to the user. Maybe it would be more clear if you said "enabling this option might result in pixel values that are > 1.0, thus you should retrieve pixels as half or float values to ensure that values are not clipped by a data type conversion to integer types."

Hmm, I'm not sure I think that's a good solution. The user doesn't choose the datatype of values that come out of the file per se, that is set by the plugin in the format (or per-channel format) of the spec. When a user asks for a specific datatype, the expectation is that the native type will be converted for them automagically. I would think that the ideal behavior would be that no data is lost by default, so changing the type returned by the plugin made the most sense. That way if you just go raw->exr (which I imagine is a very common use case) there is no loss.

I am open to any further discussion or feedback from the community before we merge this though.

lgritz:
> I am not expecting people to need to customize the scale value very frequently, nor expecting them to remember the magic number of 1.0 / 0.45. So I was planning on just saying "use apply_scene_linear_scale" in my general guidelines.

I see. So you are saying that it's two controls so that people can turn without needing to know what the magic value is for the common use case? Makes sense.


> Hmm, I'm not sure I think that's a good solution. The user doesn't choose the datatype of values that come out of the file per se, that is set by the plugin in the format (or per-channel format) of the spec. When a user asks for a specific datatype, the expectation is that the native type will be converted for them automagically. I would think that the ideal behavior would be that no data is lost by default, so changing the type returned by the plugin made the most sense. That way if you just go raw->exr (which I imagine is a very common use case) there is no loss.

Oh, I think I understand now. You're saying that if the options are used, the ImageSpec will report that the data is HALF, rather than uint16. I forgot that aspect of the situation. So it's not truly internal.

OK, all sounds fine to me.

PR #3049, opened 2021-08-03
Udim improvements

lgritz:
The primary issue being addressed here was that get_texture_info() on a
udim pattern filename (like `"foo_<UDIM>.exr"`) was unable to retrieve
info, except for an extremely narrow range of queries, because of course,
different UDIM "tiles" might have different attributes or be missing.

So in consultation with Lee Kerley, we decided that get_texture_info should
work as follows:

* If the attribute exists and has the same value for all of the
  matching UDIM file files, then succeed and return the value.

* If any of the tile files does not contain the attribute, or if the
  attribute doesn't have the same value in all found files, then it's
  a failure and the value cannot be retrieved.

This necessitated a more rebust accounting of all the tile files
comprising a UDIM pattern, by doing a full inventory the first time
the UDIM pattern is encountered, rather than (ick) iterating through
all possible tile files every time we retrieve an attribute, as we
were apparently doing before. And so the following improvements were
also made:

* get_texture_info now works for more queries -- in fact, any query
  that would be possible to ask about on an individual file, without
  restriction.  (Though, as I said, the query will fail if the value
  isn't identical across all the found tile files.)

* The info queries (even the few that were previously supported) will
  be MUCH FASTER than before, because we are doing the tile file
  inventory once per udim pattern, rather than on every query.

* That inventory will also pre-build the association between u,v tiles
  and the concrete filename (and texture handle!), so that each
  individual texture call does a simple map lookup to get all the way
  to the texture handle, without any string manipulation to first
  construct a concrete filename, then look that new filename up in the
  cache to get a handle, etc. I'm expecting this to reduce some of the
  overhead of using UDIM files with the TextureSystem.

Note: I added an ImageCacheImpl::find_file_no_add at some point in
this work, eventually not needing it by the time I got to the final
version of my implementation of this PR, but I'm going to keep it
because I can forsee its usefulness and I don't want to delete it only
to write it again some day.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -5,6 +5,7 @@
 
 #include <cstring>
 #include <memory>
+#include <regex>
 #include <sstream>
 #include <string>
 #include <vector>
@@ -296,14 +297,7 @@ ImageCacheFile::ImageCacheFile(ImageCacheImpl& imagecache,
     // reflected by the fact that m_validspec is false.
 
     // Figure out if it's a UDIM-like virtual texture
-    m_is_udim = (m_filename.find("<UDIM>") != m_filename.npos
-                 || m_filename.find("<U>") != m_filename.npos
-                 || m_filename.find("<V>") != m_filename.npos
-                 || m_filename.find("<u>") != m_filename.npos
-                 || m_filename.find("<v>") != m_filename.npos
-                 || m_filename.find("_u##v##") != m_filename.npos
-                 || m_filename.find("%(UDIM)d") != m_filename.npos)
-                && !Filesystem::exists(m_filename);
+    udim_setup();
 
     // If the config has an IOProxy, remember that we should never actually
     // release() it, because the proxy can't be reopened.
@@ -1164,6 +1158,36 @@ ImageCacheFile::mark_broken(string_view error)
     m_broken_message = error;
     imagecache().error("{}", error);
     invalidate_spec();
+}
+
+
+
+ImageCacheFile*
+ImageCacheImpl::find_file_no_add(ustring filename,
+                                 ImageCachePerThreadInfo* thread_info)
+{
+    // Shortcut - check the per-thread microcache before grabbing a more
+    // expensive lock on the shared file cache.
+    ImageCacheFile* tf = thread_info->find_file(filename);
+
+    // Check the man file cache. For this part, we need to lock the file cache.
+    if (!tf) {  // was not found in microcache
+#if IMAGECACHE_TIME_STATS
+        Timer timer;
+#endif
+        size_t bin                  = m_files.lock_bin(filename);
+        FilenameMap::iterator found = m_files.find(filename, false);
+        if (found) {
+            tf = found->second.get();
+        }
+        m_files.unlock_bin(bin);
+        if (tf)  // add to the microcache
+            thread_info->remember_filename(filename, tf);
+#if IMAGECACHE_TIME_STATS
+        thread_info->m_stats.find_file_time += timer();
+#endif
+    }
+    return tf;
 }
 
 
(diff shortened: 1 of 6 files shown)


lgritz:
@lkerley this is a feature you requested.

@etheory would like your eyes on it, too, since you recently touched the udim code.

etheory:
It's funny that you tagged me @lgritz since the second I saw this PR I thought to myself. This PR implements what I was literally about to ask you about when I next talked to you ;-). This is actually how we do it in our modified OIIO version in glimpse, and it works really well. Yes, it's a lot faster than a dynamic locked map lookup. I'll browse through the code now, but I like the gist of this from the description.

src/libtexture/imagecache.cpp
```diff
@@ -1168,6 +1162,36 @@ ImageCacheFile::mark_broken(string_view error)
...
 
+ImageCacheFile*
+ImageCacheImpl::find_file_no_add(ustring filename,
+                                 ImageCachePerThreadInfo* thread_info)
+{
+    // Shortcut - check the per-thread microcache before grabbing a more
+    // expensive lock on the shared file cache.
+    ImageCacheFile* tf = thread_info->find_file(filename);
+
+    // Check the man file cache. For this part, we need to lock the file cache.
```
etheory, line 1173: Check the main.
You could check the man, but I'm sure he's fine.

src/libtexture/imagecache.cpp
```diff
@@ -2591,33 +2615,49 @@ ImageCacheImpl::get_image_info(ImageCacheFile* file,
...
-                                            miplevel, dataname, datatype, data);
-                    // suppress errors from missing tiles we encountered
-                    // prior to finding the one that worked.
-                    (void)geterror();
-                    return r;
+
+    if (file->is_udim()) {
+        // UDIMs are tricky. What we want to do here is successfully
+        // retrieve the attribute if it's the same for all UDIM tiles, but
+        // fail if it's not the same for all tiles.
```
etheory, line 2622: The one place where I strongly disagree with this for is resolution.
Resolution can be, and often is, very different for different UDIMs. So if you can query resolution differently for each UDIM, then that's probably fine. But that restriction seems unneccessarily restrictive. At least for how we work at AL.

src/libtexture/imagecache.cpp
```diff
@@ -3438,18 +3478,154 @@ namespace {
 
 
 
+void
+ImageCacheFile::udim_setup()
+{
+    // If it's a literal, existing file, always treat it as a regular
+    // texture, even if it has what looks like udim pattern markers.
+    if (Filesystem::exists(m_filename)) {
```
etheory, line 3486: Yep, nice, we have this same feature too, and it allows users to define explicit UDIM's as filenames transparently. Nicely done, nicely done.

src/libtexture/imagecache.cpp
```diff
@@ -3438,18 +3478,154 @@ namespace {
...
+        m_is_udim = false;
+        return;
+    }
+
+    // Look for udim pattern markers, which identify this as a udim pattern.
+    // We separately identify whether it's the kind of pattern that has a
+    // single index, or separate u and v indices (and for separate, whether
+    // we use the convention of adding 1 to the tile index) -- we're going
+    // to need to know that later.
+    bool udim1   = (Strutil::contains(m_filename, "<UDIM>")
```
etheory, line 3496: I mean, this is splitting hairs, but it's important to note that the patterns are always just before the file extension (almost in every case).
For for a little more more performance, ensure you reverse search for the pattern, or reverse search from the dot before the file extension.
But we're talking tiny savings overall here.

src/libtexture/imagecache.cpp
```diff
@@ -3438,18 +3478,154 @@ namespace {
...
+ImageCacheFile*
+ImageCacheImpl::resolve_udim(ImageCacheFile* udimfile, Perthread* thread_info,
+                             int utile, int vtile)
+{
+    // Synthesize a single combined ID that we'll use as an index into
+    // m_udim_lookup.
+    uint64_t id = udim_id(utile, vtile);
 
     // Which is our mutex from the pool? Use a hash baseed on the filename.
     spin_rw_mutex& udim_lookup_mutex(
```
etheory, line 3631: Rather than all this spinlock stuff, couldn't you just resolve all the UDIM's of a file the first time it's requested, in full, store them in a vector (just store null entries for non-existent tiles) and then use constant-time lookup at runtime? The spin_rw_mutex seems expensive to have here.

testsuite/texture-udim/run.py
```diff
@@ -1,13 +1,20 @@
 #!/usr/bin/env python
 
-
-command += oiiotool ("-pattern constant:color=.5,.1,.1 256x256 3 -text:size=50:x=75:y=140 1001 -d uint8 -otex file.1001.tx")
-command += oiiotool ("-pattern constant:color=.1,.5,.1 256x256 3 -text:size=50:x=75:y=140 1002 -d uint8 -otex file.1002.tx")
-command += oiiotool ("-pattern constant:color=.1,.1,.5 256x256 3 -text:size=50:x=75:y=140 1011 -d uint8 -otex file.1011.tx")
-command += oiiotool ("-pattern constant:color=.1,.5,.5 256x256 3 -text:size=50:x=75:y=140 1012 -d uint8 -otex file.1012.tx")
+command += oiiotool ("-pattern constant:color=.5,.1,.1 256x256 3 -text:size=50:x=75:y=140 1001 --attrib Make pet --attrib Model dog -d uint8 -otex file.1001.tx")
```
etheory, line 3: It'd be good to add tests for all the patterns.

src/libtexture/imagecache.cpp
```diff
@@ -296,14 +297,7 @@ ImageCacheFile::ImageCacheFile(ImageCacheImpl& imagecache,
...
     // Figure out if it's a UDIM-like virtual texture
-    m_is_udim = (m_filename.find("<UDIM>") != m_filename.npos
-                 || m_filename.find("<U>") != m_filename.npos
-                 || m_filename.find("<V>") != m_filename.npos
-                 || m_filename.find("<u>") != m_filename.npos
-                 || m_filename.find("<v>") != m_filename.npos
-                 || m_filename.find("_u##v##") != m_filename.npos
-                 || m_filename.find("%(UDIM)d") != m_filename.npos)
-                && !Filesystem::exists(m_filename);
+    udim_setup();
```
etheory, line 300: I really like that this now happens early and once only. The run-time spin rw lock always seemed like an unneccessary bottleneck to me.

src/libtexture/imagecache.cpp
lgritz: You can certainly have different UDIM file each with different resolution. You just can't ask for the resolution (if they differ) in this get_image_info() call that doesn't know at u,v panel you want to know about. Maybe we need a get_image_info variety that also takes texture coordinates?

src/libtexture/imagecache.cpp
lgritz: fixing, thanks

src/libtexture/imagecache.cpp
lgritz: I don't feel comfortable making too any assumptions about how people will structure the `<UDIM>` marker within their filenames, that sounds inherently risky. At our place, I think the UDIM designation might be before the color space and the frame number if there is one (because why not, you could have an animated texture even in a udim context).

This section of code only happens once per UDIM pattern for the whole render, I don't see how it could possibly be a performance issue.

testsuite/texture-udim/run.py
lgritz: Yeah, that's a good point, thanks for holding my feet to the fire on this.

lgritz:
Thanks so much for the detailed review, @etheory. 

I think that in the interest of not letting the scope of this PR grow too much, and getting the basic feature of the expanded gettextureinfo ability to our shader writers quickly, I'd like to merge what I have now (which is a definite step forward in capabilities and performance), and then prepare a follow-on PR that also tackles the following:

* More comprehensive tests that exercise every naming scheme.
* Explore the idea of an O(1) dense vector lookup instead of the O(log n) flat_map I'm using here.
* Add a new API call that is a variety of gettextureinfo that takes s,t texture coordinates and thus is requesting information about the specific concrete panel that will be used at that location. (Strong caveat: this will not be constant foldable in a shader, since the s,t can't be known at JIT time!)
* At my team's request, add public API calls that expose some of the internal logic of "what is the inventory of panels for this pattern", "which concrete file corresponds to these texture coordinates", and "what is the opaque file handle for this texture coordinate" -- all of which are handy for other apps to know, so let's avoid making them duplicate the logic that's already in IC.

src/libtexture/imagecache.cpp
etheory: Maybe it's easier to jist say that a reverse search here is likely to be faster than a forward search. I.e. rfind vs find. Yes it's not performance critical code, but every wasted cycle is another dead tree.

lgritz: OH, I see what you're saying now. That's an interesting observation! I'll tackle that in the next round as well.

lgritz:
You ok with it as it stands, for now?

src/libtexture/imagecache.cpp
etheory: I guess my point is that the number of empty entries should be "somewhat sane" in all cases.
If users put a single UDIM at 1000 and one more at 1, when it's a little bit silly already.... but even in that case, 999 empty udim entries in a vector won't use all their available RAM either.

etheory:
Yep, LGTM @lgritz. I'm good with this.
For the next PR, one obvious thing I just realized is that my suggestion for a lock-free udim vector only makes sense if you pre-resolve the udim tiles, which isn't necessarily a given in the way most people use OIIO. In our renderer we pre-resolve all file handles so we have zero-cost lookups at render time. So maybe the option to choose both makes sense. i.e. some way to say "I've pre-resolved, use no locks" vs "I'm doing dynamic stuff, keep the locks". But we can nut that out in the next round.

PR #3051, opened 2021-08-04
Turn off building the experimental C APIs by default

lgritz:
We now think that this C API will be removed, and replaced by a
different one. Please don't use this one, it was never included in a
supported release, and could disappear at any time.

We're TEMPORARILY still allowing the ability to turn it on as a
built-time option, just to help people who need a bridge between now
and the final C API. But we do not consider this part of the public
API and will make it disappear at some point (without regard to
whether it's marked as a breaking release -- so please don't rely on
it).


lgritz:
@anderslanglands Do you think this goes far enough? Should we just rip the whole thing out now?

anderslanglands:
I’d rip the whole thing out. When we commit the auto-generated API it will
be a wholesale rewrite of each file anyway.

lgritz:
OK, will amend with a more complete removal of it all.

lgritz:
Amended: More completely remove the C API.

Merely commented out the option and other lines in the top-level
CMakeLists.txt, since almost the exact thing will be needed eventually.
Also kept libOpenImageIO-C/CMakeLists.txt since we'll eventually need
one much like it. But deleted the .cpp and .h files, since those will
be totally replaced.

PR #3052, opened 2021-08-05
WebpInput: Fixed mismatched new[] and delete

pumkinpal:
## Description
This request changes the type of `WebpInput::m_decoded_image` from `std::unique_ptr<uint8_t>` to `std::unique_ptr<uint8_t[]>` which ensures `operator delete[]` is called as the variable is initialized with a `operator new uint8_t[]` expression in `WebpInput::open()`.. This mismatched new/delete pair was causing errors when compiling programs with ASAN enabled.

## Tests

When ASAN was enabled the webp portion of imageinout_test was triggering the sanitizer. After this change the test completes successfully.

## Checklist:

### DIFF
--- a/src/webp.imageio/webpinput.cpp
+++ b/src/webp.imageio/webpinput.cpp
@@ -36,7 +36,7 @@ public:
 private:
     std::string m_filename;
     std::unique_ptr<uint8_t[]> m_encoded_image;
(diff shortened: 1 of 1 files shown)


lgritz (approved):
Good catch, thanks!

PR #3054, opened 2021-08-08
Udim refactor and expose more helper functions

lgritz:
To help applications that need to deal with UDIM textures, the TextureSystem
now exposes the following public API calls:

* is_udim() : is a filename (or a `TextureHandle*`) referring to a UDIM
  pattern / virtual file?

* resolve_udim() : given a filename (or handle) and s,t coordinates, return
  the handle for the concrete file corresponding to those coordinates.

* inventory_udim(): retrieve the full list of concrete filenames, as a
  vector indexed by `(utile + vtile * nvtiles)`, and the nutiles, nvtiles
  counts. Empty filenames in the vector indicate UDIM tiles that are not
  populated.

Some extra functionality falls out naturally from these and existing
ingredients:

* inventory_udim + get_texture_handle together can get you the list of
  handles corresponding to each found udim tile.

* resolve_udim + get_texture_info together can retrieve the attributes
  of the udim tile corresponding to a particular texture coordinate.

Additionally, I have done some refactoring and performance
improvements of the internals:

* The internal UdimInfo that's stored inside the ImageCacheFile record
  is now a vector, not a flat_map, so conversion of texture
  coordinates to a handle is now O(1) rather than O(log n).

* Per Luke Emrose's suggestion, use `rcontains()` when searching for
  UDIM pattern markers, since they will almost always occur close to the
  end of the full pathname.

* Added TextureSystem and ImageCache methods that lets you find out
  the filename that corresponds to a handle. (That's bound to be handy
  for somebody.)

* Beefed up the udim tests to be sure to test every udim notation that
  we support.


lgritz:
@aconty @etheory @lkerley Please review. I think the tests show that this works, but primarily I want to make sure that before I freeze the APIs for the upcoming OIIO 2.3 release, these additions seem to cover the things we want to expose to help apps deal with udims efficiently.

@etheory: the only thing you commented on previously that I haven't tackled yet is that there's still a lock involved. But at least it's a reader/writer lock, so it should be very rare for a thread to block on it (the write and full lock only happens the first time an individual UDIM tile is opened for the first time). I may take another look at this later to try to turn it into a single atomic ptr rather than a lock. I'll come back to that part.

After this is merged, I'll make a change on the OSL side as well, to add a new variety of gettextureinfo() that takes an s,t and retrieves the info of a specific UDIM panel.

src/include/OpenImageIO/texture.h
```diff
@@ -1659,6 +1664,63 @@ class OIIO_API TextureSystem {
...
+    /// and `nvtiles`, respectively. The vector `filenames` will be sized
+    /// to `ntiles * nvtiles` and filled with the the names of the concrete
+    /// files comprising the atlas, with an empty ustring corresponding to
+    /// any unpopulated tiles (the UDIM set is allowed to be sparse). The
+    /// filename list is indexed as `utile + vtile * nvtiles`.
+    ///
+    /// This method was added in OpenImageIO 2.3.
+    virtual void inventory_udim(ustring udimpattern,
+                                std::vector<ustring>& filenames,
+                                int& nutiles, int& nvtiles) = 0;
```
aconty, line 1711: This looks appropriate for our current GPU preview shader. We use cuda's textures and the on demand library, so we assign consecutive ids to these tiles. At render time ids are computed from a pattern base_id uv's on the GPU. Requests for pixels are requested with these ids and the CPU then gets the filename and loads the data with the ImageCache.

For OSL we could query the shader for used patterns. Then map the pattern ustring hash to a base id and add the tile offset from uv's. This is because the peculiarities of the optix on demand library. I think we can work with this.

Don't know if there are plans for making oiio texture lookups work on GPU, for now we are reading pixels with oiio and sampling with cuda.

lgritz: I would like to work toward a future where OIIO migrates its TextureSystem to an "XPU" style where it will just work either from C++/CPU or from Cuda (minimizing how much code and data is redundantly/independently implemented for each path), and the Cuda side fully implements all current features of the TS. One step at a time.

PR #3058, opened 2021-08-14
add flags to control stages of dependency build

anderslanglands:
## Description

For building the Rust bindings we need to provide crates.io with a complete source tree that does not depend on external downloads. So during the process of generating the bindings we want to download all the dependencies, then when building the crate we do no further network communication and jsut build the already downloaded dependencies.

To do this I've added support for two environment variables to the relevant build scripts: OIIO_DEP_BUILD_ONLY and OIIO_DEP_DOWNLOAD_ONLY that do what you would expect.

## Tests

N/A

## Checklist:

### DIFF
--- a/src/build-scripts/build_OpenJPEG.bash
+++ b/src/build-scripts/build_OpenJPEG.bash
@@ -33,10 +33,8 @@ if [[ ! -e ${OPENJPEG_SRC_DIR} ]] ; then
 fi
 cd ${OPENJPEG_SRC_DIR}
 
-if [[ -z $OIIO_DEP_BUILD_ONLY ]]; then
-    echo "git checkout ${OPENJPEG_VERSION} --force"
-    git checkout ${OPENJPEG_VERSION} --force
-fi
+echo "git checkout ${OPENJPEG_VERSION} --force"
+git checkout ${OPENJPEG_VERSION} --force
 
 mkdir -p ${OPENJPEG_BUILD_DIR} && true
 cd ${OPENJPEG_BUILD_DIR}
--- a/src/build-scripts/build_OpenJPEG.bash
+++ b/src/build-scripts/build_OpenJPEG.bash
@@ -32,16 +32,22 @@ if [[ ! -e ${OPENJPEG_SRC_DIR} ]] ; then
     git clone ${OPENJPEG_REPO} ${OPENJPEG_SRC_DIR}
 fi
 cd ${OPENJPEG_SRC_DIR}
-echo "git checkout ${OPENJPEG_VERSION} --force"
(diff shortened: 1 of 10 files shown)


src/build-scripts/build_OpenJPEG.bash
```diff
@@ -32,16 +32,22 @@ if [[ ! -e ${OPENJPEG_SRC_DIR} ]] ; then
     git clone ${OPENJPEG_REPO} ${OPENJPEG_SRC_DIR}
 fi
 cd ${OPENJPEG_SRC_DIR}
-echo "git checkout ${OPENJPEG_VERSION} --force"
-git checkout ${OPENJPEG_VERSION} --force
+
+if [[ -z $OIIO_DEP_BUILD_ONLY ]]; then
+    echo "git checkout ${OPENJPEG_VERSION} --force"
+    git checkout ${OPENJPEG_VERSION} --force
+fi
```
lgritz, line 39: I think you may have missed a bit here. You're guarding the "git checkout", which only syncs which commit is active in the already-local repo. It's the "git clone" several lines above that does the I/O.

But is OIIO_DEP_BUILD_ONLY really necessary? The clone only happens if the directory isn't already there. Like, if you run the existing script twice, I believe that the second time it wouldn't download anything.

lgritz (changes requested):
I'm not sure the "build only" is necessary -- it should already work that way because the clones don't happen if the directories. Maybe the one script that does a `curl` instead of a guarded git clone needs a similar check for whether the file already has been downloaded?

anderslanglands:
Umm yeah. Duh. I removed the BUILD_ONLY stuff and added the missing guard to build_gif.bash

lgritz (approved):
LGTM

lgritz:
Looks good, I'll merge it. 

There are two missing pieces that you'll need, though. The cmake build system itself still downloads fmt and robin-map (see the end of src/cmake/externalpackages.cmake). They only download if the packages are not found installed. But still, I think what we want is to make `build_fmt.bash` and `build_robinmap.bash` scripts so that they can be easily pre-installed to avoid any possible download. (And maybe afterwards, we should also simplify the cmake to assume that they are present instead of doing downloads as part of the build.)

anderslanglands:
Yep currently I just run a configure to get robin-map and fmt. Which works,
but is ugly because of the error messages. I can look into adding specific
download scripts for them too.

lgritz:
I can cobble them together also if you don't have the time. I mainly just wanted to be sure you were aware that the current set of build_*.bash scripts don't 100% get everything you need to avoid network access during any part of the cmake process.

PR #3072, opened 2021-08-23
Build scripts: rename OIIO_DEP_DOWNLOAD_ONLY to DEP_DOWNLOAD_ONLY

lgritz:
I realized after #3058 was merged that I often share and copy a lot of
these little dependency build scripts across multiple projects, so I
wanted to remove the OIIO-specific naming.


lgritz:
@anderslanglands I hope this doesn't mess you up too much.

anderslanglands:
No that should be fine.

PR #3077, opened 2021-08-30
Additional TIFF IOProxy bug fix -- we don't want a full init inside open

lgritz:
Calling close() in open will forget too much state. All we want to do
here is make sure the TIFF file itself is closed at that point (so it
can be re-opened in append mode to add a subsequent subimage).


lgritz:
@MrKepzie

MrKepzie:
good catch

PR #3089, opened 2021-09-03
OCIO: better catching of exceptions.

lgritz:
When OIIO that was built against OCIO 1.x, but encounters an OCIO 2.x
config file, OCIO could throw an exception that was not caught by
OIIO.  The problem was the call to OCIO::GetCurrentConfig(), it should
have been inside the try/catch block.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -322,9 +322,9 @@ ColorConfig::reset(string_view filename)
     m_impl.reset(new ColorConfig::Impl);
 #ifdef USE_OCIO
     OCIO::SetLoggingLevel(OCIO::LOGGING_LEVEL_NONE);
-    if (!ocio_current_config)
-        ocio_current_config = OCIO::GetCurrentConfig();
     try {
+        if (!ocio_current_config)
+            ocio_current_config = OCIO::GetCurrentConfig();
         if (filename.empty()) {
             getImpl()->config_  = ocio_current_config;
             string_view ocioenv = Sysutil::getenv("OCIO");
--- a/testsuite/oiiotool-color/run.py
+++ b/testsuite/oiiotool-color/run.py
@@ -62,6 +62,11 @@ command += oiiotool ("--autocc " + oiiotoolsrcdir+"/tahoe-tiny.tif"+
                      + "--ccmatrix 0.805,0.506,-0.311,0,-0.311,0.805,0.506,0,0.506,-0.311,0.805,0,0,0,0,1 "
                      + "-d uint8 -o tahoe-ccmatrix.tif")
 
+# test what happens when we read an OCIOv2 config. In particular, when
+# building against OCIOv1, we should at worst have an error message, not crash
+# with an uncaught exception.
+command += oiiotool ("--colorconfig ../common/OpenColorIO/ocio-v2_demo.ocio -echo 'OCIOv2 config Ok'")
+
 
 # To add more tests, just append more lines like the above and also add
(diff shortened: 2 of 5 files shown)


lgritz:
@ hodoulp  Can you verify that this patch fixes the crash you were seeing?

lgritz:
@hodoulp Can you verify that this patch fixes the crash you were seeing?

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -322,9 +322,9 @@ ColorConfig::reset(string_view filename)
...
 #ifdef USE_OCIO
     OCIO::SetLoggingLevel(OCIO::LOGGING_LEVEL_NONE);
-    if (!ocio_current_config)
-        ocio_current_config = OCIO::GetCurrentConfig();
     try {
+        if (!ocio_current_config)
+            ocio_current_config = OCIO::GetCurrentConfig();
         if (filename.empty()) {
             getImpl()->config_  = ocio_current_config;
             string_view ocioenv = Sysutil::getenv("OCIO");
```
hodoulp, line 330: The fix is the right one to avoid the unexpected OCIO exception.

[Comment] The code could end up loading two different configs if OCIO exists and filename is not empty. And the `OCIO` env. variable must have the priority. So, the code could be improved:
1. **if** the `OCIO` env. variable exists the code calls `OCIO::GetCurrentConfig()`
2. **else if** filename not empty the code calls `config = OCIO::Config::CreateFromFile()`, and `OCIO::SetCurrentConfig(config)`
3. **else** `config = OCIO::Config::CreateRaw()`, and `OCIO::SetCurrentConfig(config)`

It now means that the OIIO code only needs to call `OCIO::GetCurrentConfig()` to access the current config instance.

hodoulp: At line 338 and 341, the code could do `getImpl()->config_ = OCIO::Config::CreateRaw()` or `config = OCIO::Config::CreateRaw(); OCIO::SetCurrentConfig(config);` if a config instance is mandatory

testsuite/oiiotool-color/run.py
```diff
@@ -62,6 +62,10 @@
                      + "--ccmatrix 0.805,0.506,-0.311,0,-0.311,0.805,0.506,0,0.506,-0.311,0.805,0,0,0,0,1 "
                      + "-d uint8 -o tahoe-ccmatrix.tif")
 
+# test test we can read an OCIOv2 config -- when building against OCIOv1,
```
hodoulp, line 65: 1. There is a  typo 'test test' -> 'test'
2. Reading a v2 config using OCIOv1 must fail so, the test description and the test result message are ambiguous !

src/libOpenImageIO/color_ocio.cpp
lgritz: I see, thanks. So the issue is that GetCurrentConfig() will read a config if $OCIO is set? The filename is definitely intended to override $OCIO (potentially per color conversion operation). Let me take a stab at this and see if I can fix the logic so that it won't read two configs.

testsuite/oiiotool-color/run.py
lgritz: will fix, thanks

lgritz:
I think what I'd like to do, Patrick, is merge this simple fix (don't crash because an uncaught exception) because that can and should be backported cleanly.

Then I'll tackle some of these other changes you suggest, which I think I should evaluate more carefully if they will change behavior in ways that shouldn't be backported to certain releases.

PR #3092, opened 2021-09-05
Rewrite ColorConfig::reset for clearer logic and better err handling

lgritz:
Add tests to be sure we have reasonable error reporting for at least
two important cases: a specified OCIO config file is not found at all,
and an OCIO config is present but cannot be read (in particular, if
the config is newer than the library). Because the exact error spewage
is different for each OCIO version, and apparently also depending on
the exact version of the yaml library that OCIO used, we have a few
different possible output references.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -11,6 +11,7 @@
 #include <boost/container/flat_map.hpp>
 
 #include <OpenImageIO/color.h>
+#include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/imagebufalgo.h>
 #include <OpenImageIO/imagebufalgo_util.h>
 #include <OpenImageIO/strutil.h>
@@ -231,18 +232,19 @@ public:
         return handle;
     }
 
-    void error(const std::string& err)
+    // Note: Uses std::format syntax
+    template<typename... Args>
+    void error(const char* fmt, const Args&... args) const
     {
         spin_rw_write_lock lock(m_mutex);
-        m_error = err;
+        m_error = Strutil::fmt::format(fmt, args...);
     }
     std::string geterror(bool clear = true)
     {
         std::string err;
         if (clear) {
             spin_rw_write_lock lock(m_mutex);
-            err = m_error;
-            m_error.clear();
+            std::swap(err, m_error);
         } else {
             spin_rw_read_lock lock(m_mutex);
             err = m_error;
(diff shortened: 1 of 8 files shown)


lgritz:
@hodoulp This is the rewrite that addresses some of the other suggestions you made, among other things.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -317,48 +319,70 @@ ColorConfig::~ColorConfig() {}
...
+    // If no filename was specified, use env $OCIO
+    if (filename.empty())
+        filename = Sysutil::getenv("OCIO");
+
+    if (filename.size() && !Filesystem::exists(filename)) {
+        getImpl()->error("Requested non-existant OCIO config \"{}\"", filename);
+    } else {
+        // Either filename passed, or taken from $OCIO, and it seems to exist
+        try {
             getImpl()->config_ = OCIO::Config::CreateFromFile(filename.c_str());
```
hodoulp, line 341: - If this call succeeds then the next line should be `OCIO:SetCurrentConfig(getImpl()->config_);` so that `OCIO::GetCurrentConfig()` can be used from anywhere (i.e. `ocio_current_config` becomes useless).
- if this call fails then the line 354 will also fail. In fact, I do not understand the need of lines 351 to line 363.

lgritz: I assume this is part of a big multithreaded app, and I don't want OIIO to change what I assume is OCIO global state of a "current config" that other parts of an app may be depending on. So here (in English) is the logic I'm shooting for:

* If a specific config file was named (or, failing that, if a config file is named by $OCIO), try to instantiate a config based on that. (lines 333-348)
* Otherwise, just ask OCIO what it thinks is the the current config and use that (maybe it was already directed at a different file, or had a config set up by the API). (lines 351-365)
* If neither of those works, or no OCIO support is available, then just fall back on some known common transformations, such as being able to go from sRGB to a presumed linear with the same color primaries. (happens inside the call to inventory())

hodoulp: I now better understand the global context, and the code implements the algorithm.

But I have two comments:
- By using shared libraries for `OpenImageIO` and/or `OpenColorIO`, the 'current' config instance from the app could be different from the 'current' config instance from `OpenImageIO` if there is a version mismatch between the two. If the `OpenColorIO` versions are different (e.g. `2.0 vs. 2.1` or `2.1 vs. 2.2`) it means that the `OpenColorIO` symbol names will be different so, there will be two `OpenColorIO` instances in memory.
     -  If the config load fails at line 341, it could be because of a version mismatch.
- Trying to guess what the apps is actually using could be difficult. The only deterministic solution for OpenImageIO is to always receive the OCIO config file path from the app. I suggest that the logic be moved from the library to the command line tools & app if possible.

As the code implements the expected behavior, I will approve the pull request.

hodoulp (approved):

src/libOpenImageIO/color_ocio.cpp
lgritz: Yes, I understand. As you point out, if you want well-defined behavior, you name a config explicitly or you make sure $OCIO is right. The fallback to the current config of the app is already only used if neither of those things is true. We hope that the app only has one OCIO linked in (as is guaranteed for people correctly using packaging systems like Rez or Spk, which don't allow more than one at a time in a consistent environment). The fact that things could be unpredictable if there are multiple OCIO's of different versions in the same app is an edge case that I'm not going to worry about for the moment.

lgritz:
Amended. I didn't change the code as reviewed, but I did have to make some adjustments to the test reference approach. 

It seems that for the case of OCIOv1 reading a v2 config, OCIO itself spews a bunch of warnings to stderr. And apparently the order is nondeterministic, because I just could not get any set of reference error output to match stably from run to run in CI. But the problem is, if I don't capture stderr, I also don't get the oiiotool error messages, which are the very thing I'm trying to test!

So what I did is add a new flag to oiiotool, `--nostderr`, which causes oiiotool's error and warnings to get sent to stdout rather than stderr. This lets me capture the usual output and the error messages from oiiotool, for this test, while allowing OCIO's stray stderr output to go into the ether where we don't care about its order.

I'm not sure if there's any non-CI, non-debugging application for oiiotool --nostderr, but maybe it'll turn out to be generally useful.

src/libOpenImageIO/color_ocio.cpp
hodoulp: >  We hope that the app only has one OCIO linked in (as is guaranteed for people correctly using packaging systems like Rez or Spk, which don't allow more than one at a time in a consistent environment). The fact that things could be unpredictable if there are multiple OCIO's of different versions in the same app is an edge case that I'm not going to worry about for the moment.

Apps with internal (e.g. different groups) and external (e.g. different companies) third-party libraries & plugins can easily fall in that problem because of desynchronized update/release policies. So, that's not an edge case. But that's not a problem to fix in oiio i.e. that's something to keep in mind when directly or indirectly integrating third-party libraries. It explains why I push for explicit dependencies and behaviours to help integrators/devs understand the relationship between the libraries, plugins, apps, etc.

PR #3100, opened 2021-09-20
Always have OpenEXRCore C library on tap -- with option

lgritz:
Instead of a build-time option for enabling (and always using) the
experimental OpenEXR "core" library...

* Build support for it by default any time we're using a new enough
  OpenEXR version (>= 3.1). You can still hard-disable it being compiled
  in at all with `-DOIIO_USE_EXR_C_API=OFF`.

* Select whether it's actually used with the global OIIO attribute
  "openexr:core" (default to 0, disabled, for now).

This easily lets an app select support at runtime, and also may be
selected by user with environment variable
`OPENIMAGEIO_OPTIONS="openexr:core=1".

This all is intended to make it easier to benchmark use of this new
feature, switching back and forth without a rebuild.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -391,7 +391,8 @@ jobs:
       PYBIND11_VERSION: master
       PYTHON_VERSION: 3.8
       WEBP_VERSION: master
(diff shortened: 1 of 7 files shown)


lgritz:
@kdt3rd

PR #3107, opened 2021-09-22
Add multithreading to exr core read_scanlines and read_tiles

lgritz:
The old libOpenEXR/libIlmImf (default) implementation handled
multithreading on the exr side of things, inside the readPixels method.

The new (experimental, enabled only by option) OpenEXRCore
implementation leaves the threading to the caller, and we weren't
doing it, so the performance was unsurprisingly awful.

This patch turns the implementations of the exrcore-based
read_native_scanlines and read_native_tiles (and also the deep
versions thereof) to use parallel-for loops, restoring parallel
processing to the exr reader when using OpenEXRCore.

Reading of whole scanline and tiled images now is on par or slightly
faster (a few %) than when using the main OpenEXR library. Reading of whole
scanline and tiled *deep* images is substantially faster (2-3x!) than
before.

I did some extra refactoring while I was there, simplifying and
eliminating some redundant error handling and using a new
DecoderDestroyer utility class.

STILL TO DO:

* See the long comment at the end of read_native_scanlines -- the error
  message strategy implemented here is temporary, I'd like to improve it,
  will probably involve both some work here as well as on the openexr
  side.

### DIFF
--- a/src/openexr.imageio/exrinput_c.cpp
+++ b/src/openexr.imageio/exrinput_c.cpp
@@ -63,7 +63,7 @@ struct oiioexr_filebuf_struct {
 
 static void
 oiio_exr_error_handler(exr_const_context_t ctxt, exr_result_t code,
-                       const char* msg)
+                       const char* msg = nullptr)
 {
     void* userdata;
     if (EXR_ERR_SUCCESS == exr_get_user_data(ctxt, &userdata)) {
@@ -74,7 +74,9 @@ oiio_exr_error_handler(exr_const_context_t ctxt, exr_result_t code,
                 fb->m_img->errorf("EXR Error (%s): %s %s",
                                   (fb->m_io ? fb->m_io->filename().c_str()
                                             : "<unknown>"),
-                                  exr_get_error_code_as_string(code), msg);
+                                  exr_get_error_code_as_string(code),
+                                  msg ? msg
+                                      : exr_get_default_error_message(code));
                 return;
             }
         }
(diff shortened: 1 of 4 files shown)


lgritz:
@kdt3rd

lgritz:
Updated the PR:

parallelize read_native_deep_scanlines and read_native_deep_tiles

This completes the parallelization work of the new exrcore-based reader.

Reading of whole scanline and tiled images now is on par or slightly
faster (a few %) than the main OpenEXR library. Reading of whole
scanline and tiled *deep* images is substantially faster (2-3x!) than
before.

lgritz:
Any opinions on this, @kdt3rd ?

I've also since done some more extensive time trials for the "read whole image" use case (i.e., calling read_scanlines or read_tiles to slurp up the whole image at once) across a big selection of ~4k images, and I'm seeing an across-the-board speedup with this PR and exr core of around 7% for scanline zip-compressed files, and about 14% for tiled. Reading large deep files is more like 2x faster using exrcore.

I'm moving on to testing texture performance (that is, single tile reads from many threads at once) and will report those as soon as I have reliable numbers.

src/openexr.imageio/exrinput_c.cpp
```diff
@@ -1922,124 +1943,99 @@ OpenEXRCoreInput::read_native_deep_tiles(int subimage, int miplevel, int xbegin,
...
-            for (int tx = 0; tx < nxtiles; ++tx, ++curxtile) {
-                rv = exr_read_tile_chunk_info(m_exr_context, subimage, curxtile,
-                                              curytile, miplevel, miplevel,
-                                              &cinfo);
-                if (rv != EXR_ERR_SUCCESS)
-                    break;
-
-                if (first) {
+        std::vector<uint32_t> all_samples(npixels);
+        parallel_for_2D(
```
kdt3rd, line 1952: not sure if parallel_for_2D already specializes the case of pulling one tile or not, but the extra thread dispatch that is in ilmthread was definitely noticeable in the testing I had done...

kdt3rd:
In general, this looks fine - I didn't have a chance to recurse and look at the parallel loop functions to see if they specialize such that if there's only going to be one thread, it just does it in the current thread - probably more of a big deal for the tile reader than anything, once you're talking about it in the belly of the renderer

src/openexr.imageio/exrinput_c.cpp
lgritz: As a matter of fact, yes, if there is only one task in the loop, it's just performed by the calling thread.

lgritz:
Correct, if there is only one task in the loop, it's just performed by the calling thread.

PR #3116, opened 2021-09-29
Guard against invalid UDIM tile numbering

lgritz:
If think that our inventory of UDIM tile files won't find names with
wildly out-of-range tile indices, you haven't been working in a
production studio! When those indices imply tile numbers < 0, we were
running off the beginning of the vector when filling in the udim
table, to hilarious results.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -3551,6 +3551,8 @@ ImageCacheFile::udim_setup()
                     v -= 1;
                 }
             }
+            if (u < 0 || v < 0)
+                continue;  // invalid tile
             udim_list.emplace_back(ustring(udim_tile_name), nullptr, u, v);
             m_udim_nutiles = std::max(m_udim_nutiles, short(u + 1));
             m_udim_nvtiles = std::max(m_udim_nvtiles, short(v + 1));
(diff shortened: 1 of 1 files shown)


src/libtexture/imagecache.cpp
```diff
@@ -3551,6 +3551,8 @@ ImageCacheFile::udim_setup()
                     v -= 1;
                 }
             }
+            if (u < 0 || v < 0)
+                continue;  // invalid tile
```
etheory, line 3555: It would probably pay to spit out a warning here to the user to tell them.
Allowing the process to continue, un-notified, could be rather perplexing if the user is unaware of the issue, and the renderer also doesn't know what just happened.

lgritz:
I can add a warning, but here's the dilemma that makes it complicated:

* Out-of-bound texture coordinates on a texture() call have never triggered a warning -- they just use the wrap mode to adjudicate the situation and go on with their day.
* All along, out-of-range UDIM tile indices have been silently rejected, basically treating them as out-of-range texture coords on ordinary textures.
* The *NEW* feature that recently introduced this bug is that when we first encounter a UDIM pattern, we search for all existing files that match the pattern, and create a flat array mapping tile number to handle, to speed up the texture lookups we'll get later, and in this initial inventorying, a file that matches the pattern but is out of range was overwriting array bounds.
* We don't know if we'll ever ask for that tile, all that's happening at the point of this bug is that we notice that there is a filename on disk that matches the UDIM pattern and yet is not a valid UDIM. There is no other instance where we issue errors or warnings about files that we aren't actually asked to use, but merely happen to exist on disk.
* At least as I was encountering the error, the inventory was being done as the result of an OSL call to get_texture_info, and that call was part of the runtime optimization / constant folding -- for which we usually suppress all errors from being reported to the renderer, because again, the shader hasn't actually asked for the texture yet and may never do so (we may be looking at a texture used in a code branch that will be removed, and don't want spurious errors). So this mechanism for error suppression in OSL would suppress the error we'd get in this case anyway.

lgritz:
I'd like to get this fix (which stops a potentially important crash) into a release, so I'm going to merge this as-is.

I'm happy to continue the discussion about whether we should warn when we discover this, and if we decide to do so, I'll add that change in a separate PR. I've noted above my reservations about why such warnings may not be the behavior we really want, but I am willing to be convinced otherwise.

PR #3119, opened 2021-10-01
Use `inline constexpr` for C++17

### LINKED ISSUE #3117: [BUG] Unfortunate use of `static constexpr` in header files instead of `inline constexpr`
**Describe the bug**
The use of `static constexpr` in headers, while better than older usages of `static const` is still incorrect in the modern c++ world.

Static constexpr types will obey internal linkage semantics and, as such, every translation unit that includes said headers will now have their own copy of this struct placed in the binary. If you try printing out the address of one of these types in each translation unit you'll see a different pointer address listed in each.

The correct way to be ODR-safe AND have exactly 1 copy of the data in the final binary is to use `inline constexpr` for c++17 onwards. Right now large apps that include OIIO often have hundreds of bytes of storage wasted due to types like `TypeVector` (and friends) being duplicated in every implementation file that includes OIIO headers.

**Expected behavior**
Header files that declare static/const arrays and other "const" types do so in a manner that does not waste bytes in the resulting final executable.

**Evidence**
Example output from test application comparing a custom `inline constexpr` type defined in a header with OIIO's `TypeMatrix44`. 

Even with a full release build, the different addresses means different copies in the binary:

```
Inside file ConstExprTest.cpp
  &inline_constexpr               == 00007FF6075522D0
  &OpenImageIO_v2_3::TypeMatrix44 == 00007FF607552310
-------------
Inside module1.cpp
  &inline_constexpr               == 00007FF6075522D0
  &OpenImageIO_v2_3::TypeMatrix44 == 00007FF607552358
```

**Platform information:**
 - OIIO branch/version: 2.3
 - OS: Windows
 - C++ compiler: VS2019 (16.11) and VS2022

lgritz:
Owing to our C++11 focus in previous years, we had a number of places
were constants were declared as `static constexpr`. When done in
headers, this potentially causes separate copies of those constants to
be instantiated in each module that includes the header. It's not
really harmful the way we're using it for small constand objects, but
it's wasteful to potentially have dozens of copies of each those
constants by having it repeated in each module.

C++17 has a new idiom, `inline constexpr` that is more like we want --
makes it a true compiler constant that doesn't get instantiated in the
memory of each module.

We still have to cater to C++14, and in platform.h we already had a
C++-standard-dependent definition of OIIO_INLINE_CONSTEXPR that was
almost correct. Fix it (it wasn't previously used, but I must have
seen this coming at some point) so that it expands to
`inline constexpr` for C++17 and beyond, but stays as
`static constexpr` for C++14, and then use it for these constants.

Fixes #3117

### DIFF
--- a/src/doc/Doxyfile
+++ b/src/doc/Doxyfile
@@ -2166,6 +2166,7 @@ PREDEFINED             = DOXYGEN_SHOULD_SKIP_THIS \
                          OIIO_CONSTEXPR14=constexpr \
                          OIIO_CONSTEXPR17=constexpr \
                          OIIO_CONSTEXPR20=constexpr \
+                         OIIO_INLINE_CONSTEXPR="inline constexpr" \
                          OIIO_PURE_FUNC= \
                          OIIO_CONST_FUNC= \
                          OIIO_MAYBE_UNUSED= \
--- a/src/include/OpenImageIO/platform.h
+++ b/src/include/OpenImageIO/platform.h
@@ -68,7 +68,7 @@
 //                    nothing (this is useful for things that can only be
 //                    constexpr for particular versions or greater).
 // OIIO_INLINE_CONSTEXPR : inline constexpr variables, added in C++17. For
-//                         older C++, just constexpr.
+//                         older C++, static constexpr.
 //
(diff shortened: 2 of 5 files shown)


lgritz:
@jessey-git

jessey-git:
This seems ok to me.

PR #3133, opened 2021-10-16
Fix building with Clang on big endian archs

brad0:
```
error: expected value in expression
#    if _LIBCPP_BIG_ENDIAN
```

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -308,7 +308,7 @@ struct libcpp_string__long {
     std::string::size_type __size_;
     std::string::size_type __cap_;
 };
-#    if _LIBCPP_BIG_ENDIAN
+#    ifdef _LIBCPP_BIG_ENDIAN
 enum { libcpp_string__long_mask = 0x1ul };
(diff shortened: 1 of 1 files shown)


brad0:
Please backport this to the 2.3 and 2.2 branches.

lgritz:
Yes, will do.

hobbes1069:
This is interesting. I've been building OIIO for s390x for 10 years which is a big endian system without an issue.

brad0:
> This is interesting. I've been building OIIO for s390x for 10 years which is a big endian system without an issue.

You have been building with Clang for 10 years? I doubt that.

lgritz:
Fun trivia: the first OIIO commit that mentions clang is 8eb94c9362, the commit message is "Patch to be able to build with clang", and the date is "2010-10-06 22:51:37".

brad0:
> Strange. I wonder why we haven't seen this problem before? In any case, the fix is definitely more correct, thanks for submitting it!

When I looked back when this chunk of code was added it was back in 2014 or so. So I am surprised. But working with Clang especially and newer GCC releases I have seen a lot of newer warnings finding little niggling issues like this.

PR #3159, opened 2021-11-01
Add support for writing Gaussian forward and inverse transform lookup tables

olegul:
This patch is an attempt to implement the pre-processing parts from Brent Burley's Histogram-Preserving Blending paper. It generates two float array lookup tables - one for transforming each image channel to a Gaussian distribution (the invCDF float array) and one for transforming from the Gaussian distribution back to the image's distribution (the CDF array). 

This will let whoever is consuming the image (like a shader) do the blending on Gaussian distributed data and transform the blended result back to the image distribution, preserving details better than a regular blend.

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -1186,19 +1186,21 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
     }
 
     if (configspec.get_int_attribute("maketx:cdf")) {
-        // Writes Gaussian CDF and Inverse Gaussian CDF as per-channel metadata. We provide both
-        // the inverse transfrom and forward transfrom, so in theory we're free to change
-        // the distribution.
+        // Writes Gaussian CDF and Inverse Gaussian CDF as per-channel
+        // metadata. We provide both the inverse transfrom and forward
+        // transfrom, so in theory we're free to change the distribution.
         //
         // References:
         //
-        // Brent Burley, On Histogram-Preserving Blending for Randomized Texture Tiling,
-        // Journal of Computer Graphics Techniques (JCGT), vol. 8, no. 4, 31-53, 2019
+        // Brent Burley, On Histogram-Preserving Blending for Randomized
+        // Texture Tiling, Journal of Computer Graphics Techniques (JCGT),
+        // vol. 8, no. 4, 31-53, 2019
         //
-        // Eric Heitz and Fabrice Neyret, High-Performance By-Example Noise using a Histogram-Preserving Blending Operator,
-        // https://hal.inria.fr/hal-01824773},
-        // Proceedings of the ACM on Computer Graphics and Interactive Techniques,
-        // ACM SIGGRAPH / Eurographics Symposium on High-Performance Graphics 2018,
+        // Eric Heitz and Fabrice Neyret, High-Performance By-Example Noise
+        // using a Histogram-Preserving Blending Operator,
+        // https://hal.inria.fr/hal-01824773}, Proceedings of the ACM on
+        // Computer Graphics and Interactive Techniques, ACM SIGGRAPH /
+        // Eurographics Symposium on High-Performance Graphics 2018,
         //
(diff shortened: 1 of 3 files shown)


src/libOpenImageIO/maketexture.cpp
```diff
@@ -1184,6 +1184,91 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
...
+            for (int j = 0; j < bins; j++) {
+                upper  = std::upper_bound(invCDF.begin(), invCDF.end(),
+                                         float(j) / (float(bins - 1)));
+                CDF[j] = std::min(1.0f,
+                                  std::max(0.0f, float(upper - invCDF.begin())
+                                                     / float(bins - 1)));
+            }
+
+            configspec.attribute("CDF_" + std::to_string(i),
+                                 TypeDesc("float[" + std::to_string(bins) + "]"),
```
lgritz, line 1263: I think this is more easily expressed as

    TypeDesc(TypeDesc::FLOAT, bins)

rather than constructing a string in multiple steps and then having to re-parse that string in the TypeDesc constructor.

src/libOpenImageIO/maketexture.cpp
```diff
@@ -1184,6 +1184,91 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
...
+                                 CDF.data());
+        }
+
+        configspec.attribute("CDF_bits", TypeDesc("int"), &bits);
```
lgritz, line 1267: More compactly as

    configspec.attribute("CDF_bits", bits);

(If bits is an int, it knows what to do.)

lgritz: Also, just FYI, for these simple cases there is yet another syntax (which may or may not be to your taste):

    configspec["CDF_bits"] = bits;

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -1925,6 +1925,7 @@ enum MakeTextureMode {
     MakeTxTexture, MakeTxShadow, MakeTxEnvLatl,
     MakeTxEnvLatlFromLightProbe,
     MakeTxBumpWithSlopes,
+    MakeTxCDF,
```
lgritz, line 1928: I wonder if this is the right approach here at all -- it's not really a different texture type, it's an ordinary TxTexture, but with some extra metadata, right?

Perhaps a simpler way to express this is to treat it as TxTexture, but if the attributes contain either of the `maketx:cdfsigma*` attribute, then make_texture_impl knows to turn on this feature?

That would also let it fully work from `oiiotool ... -otex`, which I think your current implementation overlooked.

lgritz (changes requested):

src/libOpenImageIO/maketexture.cpp
olegul: that does look a lot nicer, yes :)

src/include/OpenImageIO/imagebufalgo.h
olegul: Yes, you're right, and I think that makes prefect sense; I've moved them over to be all attributes now. I ended up with cdf as a on/off toggle attribute, so cdfbits and cdfsigma can have default values.

lgritz (approved):
LGTM now, thanks.

Were you planning at any point to add documentation or any tests? Or did you want me to try to do that after this gets merged?

lgritz (changes requested):
Looks like a bunch of the CI tests are failing their builds, can you fix?

lgritz:
I took the liberty of pushing fixes for the signed/unsigned comparison warnings, and a few minor formatting fixes.

lgritz:
@olegul  I merged this so that we can do builds this week that include it, but I think it needs a touch-up to actually document it (i.e., amend the comments for make_texture in imagebufalgo.h, where it explains all the controls it responds to). I can do this if you don't have the time. And ideally -- though perhaps on the OSL side -- we should publish a small example shader that shows how to use this information to do the thing it's designed for.

Issue #3166, opened 2021-11-02
[BUG] fancy_add_executable.cmake does not set DEFINITIONS


SebTV:
When adding a compile-definition to the fancy_add_executable() call like
 fancy_add_executable (
            ....
            DEFINITIONS -DQT_DLL
            )
cmake ends with `Cannot specify compile definitions for target "PRIVATE" which is not built`
The reason is maybe a spelling-error:
```
--- oiio-2.3.7.2/src/cmake/fancy_add_executable.cmake      2021-11-02 17:23:34.539999100 +0100
+++ oiio-2.3.7.2/src/cmake/fancy_add_executable.cmake  2021-11-02 17:23:46.228477800 +0100
@@ -42,7 +42,7 @@
             target_include_directories (${_target_NAME} SYSTEM PRIVATE ${_target_SYSTEM_INCLUDE_DIRS})
         endif ()
         if (_target_DEFINITIONS)
-            target_compile_definitions (${_target_name} PRIVATE ${_target_DEFINITIONS})
+            target_compile_definitions (${_target_NAME} PRIVATE ${_target_DEFINITIONS})
         endif ()
         if (_target_LINK_LIBRARIES)
             target_link_libraries (${_target_NAME} PRIVATE ${_target_LINK_LIBRARIES})

```

lgritz:
Yes, that looks correct. I don't suppose you could submit the fix as a PR so I can just push a button to merge it?

SebTV:
Did a pull-request. I am new to this git-thing, so it might be wrong.

lgritz:
Looks like you did it right to me.

PR #3167, opened 2021-11-03
Fix Typo that prevents setting Compile Definitions

SebTV:
By a Typo the Macro ${target_name} is not expanded. target_link_libraries will be missing the "target" parameter and lead to an error _Cannot specify compile definitions for target "PRIVATE" which is not built_
The correct Macro is ${target_NAME}

Tested using a local compile on Windows shows the vars beeing set after the patch


lgritz:
Fixes #3166

lgritz (approved):

PR #3176, opened 2021-11-11
Fix excessive memory usage when saving EXR file with many channels

brechtvl:
## Description

Scratch memory allocation in `ImageOutput::to_native_rectangle` was accidentally using the number of channels squared.

We encountered a case where saving a 2K EXR file with 107 channels would use more than 50GB.

I don't have a simple isolated repro case. It requires writing tiles with a stride that does not match the stride in the image file. But hopefully the error in the code is clear.

## Tests

I did not add a test, it's not clear there is a good mechanism to verify memory usage like this with a test.

## Checklist:


lgritz (approved):
Holy crap! That's an embarrassing one. Good catch!

I will merge as soon as the CI finishes.

PR #3179, opened 2021-11-13
hdr overhaul

lgritz:
* Add full IOProxy support to both input and output.

* This necessitated a major refactor, including sucking the I/O specific
  code from rgbe.{cpp,h} into hdr{input,output}.cpp and getting rid of
  the old files.

* Add a testsuite entry for hdr (shockingly, there was none before)


lgritz:
Closing, I'm going to implement this a totally different way.

PR #3186, opened 2021-11-16
TypeDesc ctr from string: accept "box2f", "box3f"

lgritz:
The official names we use are "box2" and "box3", but users of Imath
know these as box2f/box3f, so accept these as synonyms when
constructing a TypeDesc from a string.


lgritz:
@grdanny

PR #3196, opened 2021-11-24
IOProxyMixin helper class

lgritz:
IOProxyMixin helper class
    
When implementing a format-specialized derived class of ImageInput or
ImageOutput, now you can inherit from `IOProxyMixin<ImageInput>` or
`IOProxyMixin<ImageOutput>` that gives you the data members and many
convenience functions you'll need for using the IOProxy, without
needing to repeat it in every format plugin that supports proxies.

This coalesces some redundant code that would need to appear
separately in most of the plugins that support IOProxy, helps to
reduce cut-and-paste errors from having to make these changes
separately in many places, and centralizes the logic so that as our
ideas of how to implement IOProxy evolve, we can easily change it in
just one place and have it apply to all plugins.

Revise the JPEG and GIF IOProxy handling to use this new mixin class,
so you can see how it works, and the bits of redundant code that it
eliminates.

Also add IOProxy support to DDS input using the new mixin.

### DIFF
--- a/src/dds.imageio/ddsinput.cpp
+++ b/src/dds.imageio/ddsinput.cpp
@@ -10,6 +10,7 @@
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/fmath.h>
 #include <OpenImageIO/imageio.h>
+#include <OpenImageIO/ioproxymixin.h>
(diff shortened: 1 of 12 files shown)


lgritz:
Abandoning. I had a better implementation idea.

PR #3205, opened 2021-11-30
make_texture()/maketx fix to channel name with monochrome_detect

### LINKED ISSUE #3204: [BUG] maketx --monochrome-detect should generate Y not R image for exr formats
**Describe the bug**
The `maketx --monochrome-detect` generates a single channel texture when it's safe to do so. Currently, if the output format of the .tx file is an exr, it names this channel R instead of Y even though Y was previously established to be the OIIO single channel name in #1434.  I guess this happens because exr allows named channels, but in this case we never specified what this name should be, so R is incorrectly used. This is a minor problem for us because some DCCs will display these textures as red images instead of grey.

**To Reproduce**
run `maketx --monochrome-detect --format exr` on a 3 channel texture where all channels are the same. Then run `oiiotool --info -v your_texture.tx` and you'll see results such as:

```
> oiiotool.exe --info -v Abstract_Soft_Left_sm.exr
Reading Abstract_Soft_Left_sm.exr
Abstract_Soft_Left_sm.exr :  512 x  512, 3 channel, float openexr
    channel list: R, G, B

> oiiotool.exe --info -v Abstract_Soft_Left_sm.tx
Reading Abstract_Soft_Left_sm.tx
Abstract_Soft_Left_sm.tx :  512 x  512, 1 channel, float openexr
    MIP-map levels: 512x512 256x256 128x128 64x64 32x32 16x16 8x8 4x4 2x2 1x1
    channel list: R
```

**Expected behavior**
`channel list: R` should have specified Y.

**Evidence**
See above for an example.

**Platform information:**
This was tested with a forked branch of 2.3.2.dev on Windows.

lgritz:
When the make_texture option monochrome_detect is used to collapse 3
identical channels to 1, it was leaving the first channel name in
place (for example, "R"). That was confusing, because at this point,
we're not trying to represent a red image, we're trying to represent a
grayscale image. A strategic call to `default_channel_names()` clears
it up, imposing the standard "Y" we use for a single channel luminance
image.

Also amended the documentation of IBA::channels(). It turned out not
to be the way I called channels() that was the problem here, but by
the time I figured that out, I had already seen that its docs could be
improved.

Fixes #3204

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -405,9 +405,11 @@ ROI OIIO_API text_size (string_view text, int fontsize=16,
 ///             `channelvalues[i]` rather than copy any channel from `src`.
 ///             If `channelorder` itself is empty, the implied channel order
 ///             will be `{0, 1, ..., nchannels-1}`, meaning that it's only
-///             renaming channels, not reordering them.
+///             renaming, truncating, or extending channels, not reordering
+///             the channels that are already present.
 /// @param  channelvalues Fill values for color channels in which
-///             `channelorder[i]` < 0.
+///             `channelorder[i]` < 0. This can be empty if no channels are
+///             to be filled with constant values.
 /// @param  newchannelnames
 ///             An array of new channel names. Channels for which this
 ///             specifies an empty string will have their name taken from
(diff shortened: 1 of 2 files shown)


ThiagoIze (approved):
Proposed code produced a Y channel for both tif and exr outputs when monochrome detect was set. When not using monochrome detection, it still does RGB, so patch looks good to me. Thanks!

PR #3232, opened 2021-12-14
Make texture system 64-bit safe for tile access

lgritz:
TextureSystem is designed primarily for tiled images (typically 64x64
pixels each), but to handle untiled images, it just treats the whole
image as one big tile. That means for very large untiled images, you
might end up with one VERY large tile, and if the image is large
enough, the size and/or offset computations therein may overflow a
32 bit int.

This patch attempts to make all the offset computations in the texture
system be 64-bit safe.


MasterZap (approved):
Looks good to moi :)

PR #3243, opened 2021-12-25
oiiotool: control flow

lgritz:
How do you craft an oiiotool command that will examine a file and add
an alpha=1 channel if it's a 3-channel file, but leave the existing
alpha alone if it's already a 4 channel file?

This patch adds control flow, so you could express the above like this:

    oiiotool in.exr --if "{TOP.nchannels < 4}" --ch R,G,B,A=1 --endif -o out.exr

The following slate of new fetures is added:

* `--set name value` sets a "user variable".

* Expression evaluation will expand identifiers that are the names of
  user variables. For example, `{i}` is replaced by the value of user
  variable `i`, and `{i+3}` returns 3 more than the current numerical
  value of `i`.

* Expression evaluation has been extended to support the following
  numerical comparison operators: `<` `>` `<=` `>=` `==` `!=` and
  `<=>` (the last one being the 3-way comparison, like in C++20).
  Note that the comprison operators have the same precedence as `+`
  and `-`, so use parentheses liberally to disambiguate complex mixed
  expressions.

* `--if condition commands... [--else commands...] --endif`

* `--while condition commands... --endwhile`

* `--for variable range commands... --endfor` where the range may be,
  like Python, any of `end` (implied begin=0, step=1), `begin,end`
  (implied step=1), or `begin,end,step`.

Also needed to modify ArgParse so that it has the notion of whether
it's "running" -- actually calling the actions for each argument, or
if it's merely parsing but taking no action -- and to be able to set
and retrieve which command line argument index an action is responding
to.


lgritz:
Fun with loops:

```
oiiotool --autocc --create 400x400 3 \
    -for j 4 \
        -for i 4 \
            "--box:fill=1:color={i/3},{j/3},0.5" "{50+100*i},{50+100*j},{50+100*i+25},{50+100*j+25}" \
        -endfor \
    -endfor \
    -d uint8 -o out.jpg
```
out

lgritz:
I beg you to reconsider your life choices if you find yourself doing anything like this:

```
oiiotool --autocc --create 500x500 1 \
    -set w "{TOP.width}" \
    -set h "{TOP.height}" \
    -set maxiters 100 \
    -for j "{h}" \
        -echo "scanline {j}" \
        -set ystart "{-1.5+3*j/h}" \
        -for i "{w}" \
            -set xstart "{-1.5+3*i/w}" \
            -set iters 1 \
            -set x "{xstart}" \
            -set y "{ystart}" \
            -while "{((x*x+y*y) < 4) * (iters <= maxiters)}" \
                -set newx "{x*x-y*y+xstart}" \
                -set newy "{2*x*y+ystart}" \
                -set x "{newx}" \
                -set y "{newy}" \
                -set iters "{iters + 1}" \
            -endwhile \
            -if "{iters < maxiters}" \
                "--box:fill=1:color={iters/maxiters}" "{i},{j},{i},{j}" \
            -endif \
        -endfor \
    -endfor \
    -powc 0.5 \
    -colormap inferno \
    -o mandel.jpg
```
mandel

lgritz:
Update: Added `&&` and `||` to expression evaluation.

lgritz:
There is a special notoriety, and a surprise gift from me, for the first person who posts the code for a working ray tracer implemented entirely as a single oiiotool command line.

lgritz:
Updated PR to add new expression functionality: 
- `!` for logical not, and also a functional synonym `not(x)` which may be handy if you're using a shell that thinks `!` is special and hard to represent literally on the command line.
- `var(x)` as another way to get the value held by user variable x (to use in cases where using the name directly could somehow be ambiguous).
- `eq(x,y)` and `neq(x,y)` for string comparisons.

PR #3261, opened 2022-01-03
Fix variable usage example in docs

AdamMainsTL:
I found this while reading through the image output section and it seems wrong given the following examples.

## Description

Fixes the docs regarding environmental variable usage to match the examples used.

## Tests

Didn't add tests

## Checklist:


lgritz (approved):
Yes, quite right! Thanks.

PR #3280, opened 2022-01-17
Add UDIM pattern `<UVTILE>` (specified by MaterialX)

### LINKED ISSUE #3275: [FEATURE REQUEST] Support the MaterialX `<UVTILE>` substitution marker
MaterialX introduced a particular substitution marker[1] that is currently not very well supported in the ecosystem. It's actual unclear on what the future of it might be[2]. As I cannot predict the outcome of what might happen here :), I suppose the only recourse is to support it.

Add the `<UVTILE>` substitution pattern to `ImageCacheFile` processing. This is a 1-based format and is equivalent to `<U>_<V>` according to the current MaterialX spec.

Currently, in the renderer where this was noticed, we will manually manipulate the incoming filename and replace the `"<UVTILE>"` substring with `"<U>_<V>"`.

[1] Page 18 of http://www.materialx.org/assets/MaterialX.v1.38.Spec.pdf 
[2] USD only supports `<UDIM>` but they wish to extend that in the future. It's unclear if that means to extend to only `<UVTILE>` or extend to all OIIO tokens as I can find no public discussion on the details.

lgritz:
MaterialX documents a udim pattern that we hadn't previously supported,
`<UVTILE>`, which is equivalent to our `u<U>_v<V>`.

Fixes #3275


src/doc/texturesys.rst
```diff
@@ -209,15 +209,16 @@ UDIM tiled texture atlases. The substitutions will occur if the texture
...
-`_u##v##`  utile, vtile
-`%(UDIM)d` 1001 + utile + vtile*10
-========== ========================
+==========  ================================================
+`<UDIM>`    1001 + utile + vtile*10
+`<u>`       utile
+`<v>`       vtile
+`<U>`       utile + 1
+`<V>`       vtile + 1
+`<UVTILE>`  equivalent to `u<U>_v<V>`
```
jessey-git, line 218: This is a little confusing since `<U>` already implies the 'u' prefix -- I had to read the source code to know that though. Same for the other singular tokens like `<u>` etc.  It's really equivalent to `<U>_<V>`.

jessey-git:
Looks ok to me save for my 1 minor comment above.

src/doc/texturesys.rst
lgritz: Oh, you are right! I had forgotten that. I fill fix.

Issue #3276, opened 2022-01-16
[BUG] OIIO uses the AVPicture API, which has been removed from ffmpeg in 5.0


berolinux:
**Describe the bug**
OIIO makes use of the ffmpeg AVPicture API (src/ffmpeg.imageio/ffmpeginput.cpp).
This API has been deprecated in 2017, and has been removed in the recently released ffmpeg 5.0.

**To Reproduce**
1. Update ffmpeg to 5.0
2. Try to build OIIO

**Expected behavior**
Compiles

**Evidence**
```
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:35:16: error: unknown type name 'AVPicture'
avpicture_fill(AVPicture* picture, uint8_t* ptr, enum AVPixelFormat pix_fmt,
               ^
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:241:31: error: use of undeclared identifier 'av_register_all'
    std::call_once(init_flag, av_register_all);
                              ^
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:272:15: error: assigning to 'AVCodec *' from 'const AVCodec *' discards qualifiers
    m_codec = avcodec_find_decoder(par->codec_id);
              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:611:47: error: no member named 'pkt_pts' in 'AVFrame'
            if (static_cast<int64_t>(m_frame->pkt_pts)
                                     ~~~~~~~  ^
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:615:34: error: no member named 'pkt_pts' in 'AVFrame'
                      * m_frame->pkt_pts;
                        ~~~~~~~  ^
/home/bero/abf/OpenImageIO/BUILD/oiio-2.3.10.1/src/ffmpeg.imageio/ffmpeginput.cpp:623:49: error: unknown type name 'AVPicture'
                avpicture_fill(reinterpret_cast<AVPicture*>(m_rgb_frame),
                                                ^
[...]

PR #3282, opened 2022-01-18
Port to ffmpeg 5.0

berolinux:
Port to ffmpeg 5.0 -- av_register_all removed, AVPicture
removed, pkt_pts changed to pts

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>
Fixes: #3276

## Checklist:


lgritz:
Does this change the oldest version of ffmpeg that we support?

src/ffmpeg.imageio/ffmpeginput.cpp
```diff
@@ -238,7 +238,6 @@ FFmpegInput::open(const std::string& name, ImageSpec& spec)
     }
 
     static std::once_flag init_flag;
```
lgritz, line 240: If you get rid of the call_once, you should also get rid of those once_flag.

PR #3303, opened 2022-01-24
Test Filesystem::searchpath_split behavior on an empty string

lgritz:
It should make a vector with no entries, not a vector with one entry
that is an empty string. Verify.


lgritz:
Withdrawing, I have a replacement coming.

PR #3305, opened 2022-01-25
Embed Imath version symbols in oiioversion.h

lgritz:
This makes sure that two new preprocessor symbols are defined in
oiioversion.h: `OIIO_USING_IMATH_VERSION_MAJOR` and
`OIIO_USING_IMATH_VERSION_MINOR`, revealing the version of Imath being
used by the build of OIIO (which may not be the same as the version
found or being used by the app that's calling OIIO). They also get
defined as CMake variables in our exported cmake configs, allowng
downstream cmake build systems to be able to easily test whether the
Imath they found is compatible with the version that we use in our
APIs.

### DIFF
--- a/src/cmake/modules/FindOpenEXR.cmake
+++ b/src/cmake/modules/FindOpenEXR.cmake
@@ -46,7 +46,7 @@ if (TARGET OpenEXR::OpenEXR AND TARGET Imath::Imath)
     # OpenEXR 3.x if both of these targets are found
     set (FOUND_OPENEXR_WITH_CONFIG 1)
     if (NOT OpenEXR_FIND_QUIETLY)
-        message (STATUS "Found CONFIG for OpenEXR 3 (OPENEXR_VERSION=${OpenEXR_VERSION})")
+        message (STATUS "Found CONFIG for OpenEXR 3 (OpenEXR_VERSION=${OpenEXR_VERSION})")
     endif ()
 
     # Mimic old style variables
@@ -76,7 +76,7 @@ elseif (TARGET OpenEXR::IlmImf AND TARGET IlmBase::Imath AND
     # OpenEXR 2.4 or 2.5 with exported config
     set (FOUND_OPENEXR_WITH_CONFIG 1)
     if (NOT OpenEXR_FIND_QUIETLY)
-        message (STATUS "Found CONFIG for OpenEXR 2 (OPENEXR_VERSION=${OpenEXR_VERSION})")
+        message (STATUS "Found CONFIG for OpenEXR 2 (OpenEXR_VERSION=${OpenEXR_VERSION})")
     endif ()
 
     # Mimic old style variables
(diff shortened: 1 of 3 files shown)


src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -166,21 +166,23 @@ find_path (OPENEXR_INCLUDE_PATH OpenEXR/OpenEXRConfig.h
...
+    string (REGEX MATCHALL "[0-9]+" OpenEXR_VERSION_MAJOR ${TMP})
     file(STRINGS "${OPENEXR_INCLUDE_PATH}/OpenEXR/OpenEXRConfig.h" TMP REGEX "^#define OPENEXR_VERSION_MINOR .*$")
-    string (REGEX MATCHALL "[0-9]+" OPENEXR_VERSION_MINOR ${TMP})
+    string (REGEX MATCHALL "[0-9]+" OpenEXR_VERSION_MINOR ${TMP})
 else ()
     # Assume an old one, predates 2.x that had versions
     set (OPENEXR_VERSION 1.6.1)
-    set (OPENEXR_MAJOR 1)
-    set (OPENEXR_MINOR 6)
+    set (OpenEXR_VERSION 1.6.1)
```
EvanAW, line 183: Duplicate line?

src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -166,21 +166,23 @@ find_path (OPENEXR_INCLUDE_PATH OpenEXR/OpenEXRConfig.h
 # Try to figure out version number
 if (DEFINED _OPENEXR_VERSION AND NOT "${_OPENEXR_VERSION}" STREQUAL "")
     set (OPENEXR_VERSION "${_OPENEXR_VERSION}")
```
EvanAW, line 168: Shouldn’t this line be removed as well?

src/cmake/modules/FindOpenEXR.cmake
```diff
@@ -166,21 +166,23 @@ find_path (OPENEXR_INCLUDE_PATH OpenEXR/OpenEXRConfig.h
...
+    string (REGEX MATCHALL "[0-9]+[.0-9]+" OpenEXR_VERSION ${TMP})
     file(STRINGS "${OPENEXR_INCLUDE_PATH}/OpenEXR/OpenEXRConfig.h" TMP REGEX "^#define OPENEXR_VERSION_MAJOR .*$")
-    string (REGEX MATCHALL "[0-9]+" OPENEXR_VERSION_MAJOR ${TMP})
+    string (REGEX MATCHALL "[0-9]+" OpenEXR_VERSION_MAJOR ${TMP})
     file(STRINGS "${OPENEXR_INCLUDE_PATH}/OpenEXR/OpenEXRConfig.h" TMP REGEX "^#define OPENEXR_VERSION_MINOR .*$")
-    string (REGEX MATCHALL "[0-9]+" OPENEXR_VERSION_MINOR ${TMP})
+    string (REGEX MATCHALL "[0-9]+" OpenEXR_VERSION_MINOR ${TMP})
 else ()
     # Assume an old one, predates 2.x that had versions
     set (OPENEXR_VERSION 1.6.1)
```
EvanAW, line 182: Remove this line as well?

src/cmake/modules/FindOpenEXR.cmake
lgritz: I'm setting both OPENEXR_VERSION and also OpenEXR_VERSION, in case somebody is depending on the old one.

src/cmake/modules/FindOpenEXR.cmake
lgritz: Again, I'm following two conventions so I don't break anything downstream that might be using this module.

PR #3306, opened 2022-01-25
Filesystem::searchpath_split handle empty paths

### LINKED ISSUE #3274: [BUG] searchpath_split contains empty paths
**Describe the bug**
I haven't verified if it was this comment, but I believe the Nov 30 commit that removed using boost for `searchpath_split()` introduced a regression where now we can get empty paths returned. This was detected by our unit tests.

**To Reproduce**
Try `OIIO::Filesystem::searchpath_split(std::string(""), paths, false);` and then you should find that older OIIO returned an empty vector while the new OIIO has a vector that contains an empty string. This happens with non-empty search paths as well.

I worked around this in our code by removing empty strings, but it'd probably be better to fix this in OIIO itself, either by using my fix or by identifying what part of OIIO is allowing empty paths through:

```
      OIIO::Filesystem::searchpath_split(the_string, paths, false);
      auto it = std::remove(paths.begin(), paths.end(), std::string());
      std::distance(it, paths.end());
      paths.erase(it, paths.end());
```

**Expected behavior**
I would have expected there to be no change in behavior.

**Platform information:**
I tested this on the lastest OIIO master and on an OIIO master that's maybe a year old, so somewhere between these the regression happened.

lgritz:
Fixes #3274

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -190,6 +190,8 @@ Filesystem::searchpath_split(string_view searchpath, bool validonly)
         size_t len = path.size();
         while (len > 1 && (path.back() == '/' || path.back() == '\\'))
             path.erase(--len);
+        if (path.empty())
+            continue;
         // If it's a valid directory, or if validonly is false, add it
         // to the list
         if (!validonly || Filesystem::is_directory(path))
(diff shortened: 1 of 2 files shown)


ThiagoIze:
This passes my testsuite -- thanks!

PR #3312, opened 2022-01-27
wstring (almost) all the things

### LINKED ISSUE #3304: Add support for std::wstring for open
Windows supported files that cannot be represented in std::sting. so the attached file cannot be loaded.

[test.zip](https://github.com/OpenImageIO/oiio/files/7930273/test.zip)

lgritz:
As Doug Rogers points out in Issue #3304, it is customary for Windows
applications to use wstring extensively for Unicode support. But it's
especially critical for unicode filenames, because unlike Unix-like
OS's where fopen() and similar functions support UTF-8, on Windows
those functions are strictly ASCII, and Windows users are used to
calling _wfopen() and other nonstandard functions that expect
filenames to be std::wstring if they want to support Unicode.

This patch:

* More clearly documents through all of our classes when parameters that
  represent filenames or searchpaths are assumed to support UTF-8.

* Augments ImageInput, ImageOutput, and ImageBuf so that their methods
  that take file names or path lists have additional flavors that take
  wstring (always implemented as an inline shim that translates the
  UTF-16 wstring to a std::string with UTF-8 encoding to pass on to the
  usual method).

* Augments ImageCache and TextureSystem in just a couple places, the
  functions where a filename is turned into a handle. All the many
  other functions that take ustring for function names were left alone,
  because they are called very frequently and are performance-sensitive,
  so we don't want any translation from wstring to string happening on
  every call.

We do not attempt to add wstring-based versions of functions that take
strings for other purposes -- that is, that are not intended to
represent filenames.

Fixes #3304

### DIFF
--- a/src/include/OpenImageIO/filesystem.h
+++ b/src/include/OpenImageIO/filesystem.h
@@ -29,6 +29,7 @@
 #include <OpenImageIO/export.h>
 #include <OpenImageIO/oiioversion.h>
 #include <OpenImageIO/span.h>
+#include <OpenImageIO/strutil.h>
 #include <OpenImageIO/string_view.h>
 
 #if defined(_WIN32) && defined(__GLIBCXX__)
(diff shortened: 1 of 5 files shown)


lgritz:
Is this approximately what you had in mind, @DougRogers ?

DougRogers:
Yes.  This is exactly wheat I had in mind.  Thank you.

PR #3320, opened 2022-02-02
update python documentation for ImageBuf.set_full() 

apetrynet:
## Description

The python documentation for `ImageBuf.set_full()` states it requires a `ROI` which doesn't seem to be the case.
This PR updates the documentation and example to match C++
  
## Tests

This is a pure documentation update. No additional tests needed 

## Checklist:


lgritz (approved):
Quite right! I'm not sure how that happened. Thanks for the fix

PR #3321, opened 2022-02-03
Color space adjustment for PPM and PNG files.

### LINKED ISSUE #3084: [BUG] png and pnm plug-ins do not set default ColorSpace
**Describe the bug**
When a PNG file without any color space chunks (no sRGB, iCCP or gAMA) is read the plug-in does not set the `oiio:ColorSpace` attribute.

The pnm plug-in also does not set `oiio:ColorSpace`.

**To Reproduce**
1. Set a breakpoint at the line after a program calls `ImageSpec::find_attribute("oiio:ColorSpace")`
2. Open a .png file with no color space info or open a pnm file.
3. Observe that return value from `find_attribute` is `nullptr`.

**Expected behavior**
I expect the plug-ins to set the default color space information specified by the file format. Both the jpeg and webp plug-ins set `oiio:ColorSpace` to "sRGB" when there is no information in the file, the usual case. Like those specs, the PNG spec says sRGB is to be assumed in the absence of color space info. The PNM spec says that the images are in Rec709 space so that plug-in should set `oiio:ColorSpace` to "rec709".

**Evidence**
I'm attaching a png file with no color space info. No PNM (.pam, .pgm, .ppm) files have color space info so you can try with any file you can find.

color_grid

**Platform information:**
 - OIIO branch/version: 2.1.17
 - OS: macOS 11.5.2
 - C++ compiler: Clang

lgritz:
PPM by specification is Rec709! Make it so.

PNG files were getting marked as sRGB or Gamma if certain header
attributes were found in the file, but if they were absent, we didn't
designate a color space at all. Change so that the default assumes
sRGB, in the absence of any metadata telling us for sure.

Fixes #3084

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5531,6 +5531,10 @@ output_file(int /*argc*/, const char* argv[])
                 || Strutil::iends_with(filename, ".gif")
                 || Strutil::iends_with(filename, ".webp")))
             outcolorspace = string_view("sRGB");
+        if (outcolorspace.empty()
+            && (Strutil::iends_with(filename, ".ppm")
+                || Strutil::iends_with(filename, ".pnm")))
+            outcolorspace = string_view("Rec709");
         if (outcolorspace.size() && currentspace != outcolorspace) {
             if (ot.debug)
                 std::cout << "  Converting from " << currentspace << " to "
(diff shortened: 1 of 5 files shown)


MarkCallow:
For the sake of thoroughness we should note that the PNG spec merely says "choose a likely default gamma value" in the absence of any color space information. sRGB is the right likely value, given that web browsers all make the same choice.

PR #3323, opened 2022-02-04
Targa: Fix parsing of TGA 2.0 extension area.

jreichel-nvidia:
Skip 3 bytes between "software" and "background (key) colour" for the "software
version" (was simply missing). Avoids that m_alpha_type is overwritten with
wrong values.

## Tests

Did not create a new test case. Existing targa test fails for me (probably unrelated) with:

```131/161 Testing: targa
131/161 Test: targa
Command: "/usr/bin/python3.9" "/some_prefix/src/testsuite/runtest.py" "/some_prefix/src/build-debug/testsuite/targa"
Directory: /some_prefix/src/build-debug
"targa" start time: Feb 04 09:19 CET
Output:
----------------------------------------------------------
  File "/some_prefix/src/testsuite/runtest.py", line 452, in <module>
    exec (code)
  File "run.py", line 3, in <module>
ModuleNotFoundError: No module named 'OpenImageIO'
<end of output>
Test time =   0.04 sec
----------------------------------------------------------
Test Failed.
"targa" end time: Feb 04 09:19 CET
"targa" time elapsed: 00:00:00
----------------------------------------------------------
```

## Checklist:

### DIFF
--- a/src/targa.imageio/targainput.cpp
+++ b/src/targa.imageio/targainput.cpp
@@ -342,20 +342,16 @@ TGAInput::open(const std::string& name, ImageSpec& newspec)
             // software
             if (!ioread(buf.c, 41, 1))
                 return false;
+            uint16_t n;
+            char l;
+            if (!read(n) || !read(l))
+                return false;
             if (buf.c[0]) {
                 // tack on the version number and letter
-                uint16_t n;
-                char l;
-                if (!read(n) || !read(l))
-                    return false;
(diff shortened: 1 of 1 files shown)


lgritz:
CI fails testsuite, is printing different metadata than before. Either it was wrong before and right now, and the reference output from the targa test needs to be updating as part of the PR, or else it indicates that something about this patch is not correct.

lgritz (approved):
Aah, that makes a lot more sense!
Thanks for the fix.

PR #3330, opened 2022-02-07
V3fParam

lgritz:
Speculative work -- create a small shim class template called
Vec3Param (and its specialization, V3fParam) that does nothing but
implicitly convert from and to things that looks like 3-vectors
(thanks to some template magic I developed a while back for the Imath
project). And similar for template MatrixParam (and specializations
M33fParam and M44fParam) to pass a square matrix.

Basically the idea is that we can replace the public API function
parameters where we take an Imath::V3f and instead use a
V3fParam. User code that passes an Imath::V3f still works because
V3fParam accepts and converts it.  Implementation code can implicitly
convert the V3fParam to an Imath::V3f, so it doesn't need to be
changed, either. Only the function declaration itself. (Same basic
situation for MatrixParam and Imath::M33f/M44f.)

So in the process, now we have laundered our public API so that it
doesn't reference the Imath types and headers any longer. This confers
two big advantages:

1. It allows OIIO itself and any downstream client code calling OIIO
   to use *different* versions of Imath if they want.

2. It allows client code to not need Imath at all, for example if they
   have their own proprietary vector type that they prefer to use
   internally.

I'm not necessarily 100% sold on this approach, but I think it's got
promise and am seeking feedback. It seems to work, and I was able
to compile OSL against this altered OIIO with *no changes*.

### DIFF
--- a/THIRD-PARTY.md
+++ b/THIRD-PARTY.md
@@ -34,6 +34,11 @@ SPDX-License-Identifier: BSD-3-Clause
   The sample OpenColorIO configurations in our testsuite are borrowed from
   this ASWF project, also BSD-3-Clause licensed.
 
+* OpenEXR/Imath (c) Copyright contributors to the OpenEXR Project.
+  https://github.com/AcademySoftwareFoundation/Imath
+
+  Some templates in vecparam.h were first developed for Imath.
+
 * DPX reader/writer (c) Copyright 2009 Patrick A. Palmer.
   https://github.com/patrickpalmer/dpx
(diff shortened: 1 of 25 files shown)


lgritz:
Does anybody have thoughts about this? 

Does it seem like a good idea to use a minimal parameter interface class like this in order to completely excise Imath types from our public headers (even if we still use them extensively internally) and make OIIO totally agnostic to the vector/matrix types (or Imath version) favored by its downstream client?

PR #3331, opened 2022-02-08
Add texture creation metadata: "handed"

lgritz:
For vector displacement maps or normal maps that are in tangent space,
it is heplful to be able to record whether the data are intended to use
a local left-handed or right-handed coordinate system, and for that to
be querably from the shader accessing the map.

The `maketx --handed` option, or `oiiotool -attrib -otex:handed=...`,
or adding "handed" metadata to the configuration ImageSpec being
passed to `IBA::make_texture()` does this. It's easy for texture
formats like openexr that allow arbitrary metadata. For TIFF textures,
we embed it in the ImageDescription tag (as we do for several other
texture-specific hints).

### DIFF
--- a/src/doc/maketx.rst
+++ b/src/doc/maketx.rst
@@ -490,6 +490,14 @@ Command-line arguments are:
 
    These options were first added in OpenImageIO 2.3.10.
 
+.. option:: --handed <value>
+
+   Adds a "handed" metadata to the resulting texture, which reveals the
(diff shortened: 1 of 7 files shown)


olegul:
nice - I tested it with OSL reading both exrs and tiff, lgtm!

olegul (approved):
I did a quick build with it in OSL, reading both tx/tiff and exr - lgtm!

PR #3343, opened 2022-02-20
Wean ourselves off of string_view::c_str()

lgritz:
c_str() was cute, but it's a little dicey and it's not present in
the C++20 std::string_view. We didn't use it in a ton of places,
so replace with other idioms so that in the future we are better
able to interchange with std::string_view or replace our own entirely
once we have flying cars and C++20 minimum.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -2789,7 +2789,8 @@ inline bool attribute (string_view name, float val) {
 }
 /// Shortcut attribute() for setting a single string.
 inline bool attribute (string_view name, string_view val) {
-    const char *s = val.c_str();
+    std::string valstr = val;
+    const char *s = valstr.c_str();
     return attribute (name, TypeString, &s);
 }
 
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -789,7 +789,8 @@ public:
     }
     virtual bool attribute(string_view name, string_view val)
     {
-        const char* s = val.c_str();
+        std::string valstr(val);
+        const char* s = valstr.c_str();
         return attribute(name, TypeDesc::STRING, &s);
     }
 
--- a/src/libtexture/texture_pvt.h
+++ b/src/libtexture/texture_pvt.h
@@ -56,7 +56,8 @@ public:
     }
     virtual bool attribute(string_view name, string_view val)
     {
-        const char* s = val.c_str();
+        std::string valstr(val);
+        const char* s = valstr.c_str();
         return attribute(name, TypeDesc::STRING, &s);
     }
 
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -881,6 +881,9 @@ public:
     /// read from disk).
     string_view name(void) const;
 
+    /// Return the name of the buffer as a ustring.
+    ustring uname(void) const;
+
     /// Return the name of the image file format of the file this ImageBuf
     /// refers to (for example `"openexr"`).  Returns an empty string for an
     /// ImageBuf that was not constructed as a direct reference to a file.
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -814,7 +814,7 @@ ImageViewer::saveAs()
         return;
     QString name;
     name = QFileDialog::getSaveFileName(this, tr("Save Image"),
-                                        QString(img->name().c_str()),
+                                        QString(img->uname().c_str()),
                                         tr(s_file_filters));
     if (name.isEmpty())
         return;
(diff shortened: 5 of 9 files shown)


src/include/OpenImageIO/imageio.h
```diff
@@ -2789,7 +2789,8 @@ inline bool attribute (string_view name, float val) {
 }
 /// Shortcut attribute() for setting a single string.
 inline bool attribute (string_view name, string_view val) {
-    const char *s = val.c_str();
+    std::string valstr = val;
+    const char *s = valstr.c_str();
```
etheory, line 2793: Why the temporary std::string here and not:
`const char *s = valstr.str().c_str();`
like you do throughout the rest of the PR?

src/libtexture/imagecache_pvt.h
```diff
@@ -789,7 +789,8 @@ class ImageCacheImpl final : public ImageCache {
     }
     virtual bool attribute(string_view name, string_view val)
     {
-        const char* s = val.c_str();
+        std::string valstr(val);
```
etheory, line 792: Same comment as above.

src/libtexture/texture_pvt.h
```diff
@@ -56,7 +56,8 @@ class TextureSystemImpl final : public TextureSystem {
     }
     virtual bool attribute(string_view name, string_view val)
     {
-        const char* s = val.c_str();
+        std::string valstr(val);
```
etheory, line 59: Again same as above.

src/include/OpenImageIO/imageio.h
lgritz: No reason. They mean the same thing. `valstr.str().c_str()` still makes a temporary string, it just doesn't have a name.

This entire set of changes is a actually from my vacation in December when I tinkered with a number of infrastructure oddities, just because (and when) I felt like it. I suspect that what happened here was that I worked on this a little bit at a time, fixed different places on different days and didn't check for absolutely consistency.

If you prefer the more compact one-expression idiom without the named temporary, I can change them all to be consistent.

lgritz: Actually, now that I look at it, C++17 std::string_view doesn't have the str() method either. So it turns out that this assignment to a std::string has the std::string_view equivalency I was aiming for, but the `sv.str().c_str()` idiom does not. The most compact 1-expression idiom that is fully interchangeable with std::string_view would be `std::string(sv).c_str()`.

I will fix this up to be std-equivalent everywhere.

lgritz: Notwithstanding the prior statement about how `sv.str().c_str()` should be the one that changes if we desire full interchangeability with std::string_view, now that I've had my coffee, I can see that there REALLY WAS a reason why we used the temporary in this place!

You need to look at the NEXT line, which is

    return attribute (name, TypeString, &s);

We need the *address* of the `char*`, which means the std::string that we called c_str() on needs to survive until after that call.

This particular spot was the correct idiom all along.

lgritz:
If you're ok with my explanation, I'm going to let this PR stand as it is. In a subsequent set of changes, I'll similarly find all the places where we use the nonstandard str() methods and replace it with the proper std::string constructor.

etheory:
Sounds good to me @lgritz though I'm not to keen on the extra copies this is making and the potential performance decrease it could produce, unless I'm mistaken?

src/include/OpenImageIO/imageio.h
etheory: @lgritz isn't this returning the address of a temporary in that case? Isn't that bad?

lgritz: Safety: It's taking the address of a named local variable, to pass to a function that will make a more permanent copy. So it's safe.

Copying: The old c_str() that this replaces made a ustring (!) to have something stable to take the c_str() of. Now it's just making a std::string, which is probably better performance, in general.

But most importantly, there is a second PR in flight, #3342, that makes an orthogonal set of changes that will eliminate the extra copies entirely.

etheory: Thanks for the clarification and answering my silly questions @lgritz ;-). LGTM!

lgritz:
I appreciate the second pair of eyes on this.

lgritz:
I should also point out, in terms of performance, that these lines you mentioned in your comments were specifically for:

* Setting global oiio attributes
* Setting ImageCache attributes
* Setting TextureSystem attributes

These are calls that are expected to be made AT MOST once per global attribute, once per ImageCache attribute when it is set up, and once per TextureSystem attribute, respectively. And all three only have a handful each of attributes that are string data. So we are talking about maybe a dozen total calls (and total extra string copies) per execution of the whole program. It is just not a performance issue here.

On the other hand, the string attribute setting of an individual ImageSpec or setting values in a ParamValueList might happens tens to hundreds of times per spec. But like I said, #3342 specifically addresses those cases in a way that eliminates the extra copies for string data.

PR #3344, opened 2022-02-20
Fully templatize our string_view

lgritz:
Brings us more in line with C++17 std::string_view.


lgritz:
If nobody has any reservations about this, I will probably merge it this weekend.

PR #3361, opened 2022-03-04
simd.h fix for armv7 -- vaddvq_s32 is only aarch64

### LINKED ISSUE #3360: [BUILD] issue on aarch64 neon asm enabled on arm32
**Describe the bug**
When building a dependency with armv7 neon enabled (-fpu=neon), oiio fail with the following error:

```
In file included from /builddir/build/BUILD/olive-41a49c4880e21b11b08de94d2df1b227bf219fc6/app/codec/ffmpeg/ffmpegdecoder.cpp:32:
In file included from /usr/include/OpenImageIO/imagebuf.h:16:
In file included from /usr/include/OpenImageIO/fmath.h:45:
/usr/include/OpenImageIO/simd.h:4753:12: error: use of undeclared identifier 'vaddvq_s32'; did you mean 'vpaddlq_s32'?
    return vaddvq_s32(v);
           ^
/usr/lib/clang/13.0.1/include/arm_neon.h:18534:16: note: 'vpaddlq_s32' declared here
__ai int64x2_t vpaddlq_s32(int32x4_t __p0) {
```

**To Reproduce**
1. build (or cross-compile) oiio on armv7 with neon enabled.

**Expected behavior**
Build should succeed, but fails because vaddvq_s32 is only available on aarch64 and not on armv7.
Most of the time armv7 (distro/)users don't explicitly enable neon.

**Evidence**
vaddvq_s32 is tagged as aarch64 only in this documentation.
https://arm-software.github.io/acle/neon_intrinsics/advsimd.html

**Platform information:**
 - OIIO branch/version: current tag 
 - OS: Linux
 - C++ compiler: clang 13 on Fedora 36
 - Any non-default build flags when you build OIIO: -fpu=neon

lgritz:
Fixes #3360

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -4751,7 +4751,7 @@ OIIO_FORCEINLINE vint4 vreduce_add (const vint4& v) {
 OIIO_FORCEINLINE int reduce_add (const vint4& v) {
 #if OIIO_SIMD_SSE
     return extract<0> (vreduce_add(v));
-#elif OIIO_SIMD_NEON
+#elif OIIO_SIMD_NEON && defined(__aarch64__)
     return vaddvq_s32(v);
 #else
     SIMD_RETURN_REDUCE (int, 0, r += v[i]);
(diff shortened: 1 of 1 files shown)


lgritz:
No response on the issue #3360 about whether this works... but it looks like a step forward, so I'm going to commit it and we can revisit if it doesn't completely solve the problem.

PR #3363, opened 2022-03-06
Build and CI and support for the icx compiler

lgritz:
icx is the *new* Intel LLVM-based compiler, and it behaves a lot like
clang (versus icc, which is the "classic" Intel compiler).

* An additional CI test in ci.yml for icx.

* Tweaks in compiler.cmake and platform.h to detect and set options for
  this compiler.

* A variety of warning fixes encountered for this compiler.

* Add a new cmake build-time option `ENABLE_FAST_MATH` (default: OFF)
  for enabling fast-math optimizations that trade IEEE compliance for
  speed. The Intel compiler enables it by default (!), so this build
  option makes icx behave like the other compilers (not fast math
  unless explicitly asked for), which makes the CI tests and other
  behavior more consisent across compilers and platforms.

* Note: The bug that prevents some versions of fmtlib from working
  with icc does not appear to be a problem for icx.

### DIFF
--- a/src/build-scripts/gh-installdeps.bash
+++ b/src/build-scripts/gh-installdeps.bash
@@ -21,12 +21,10 @@ if [[ "$ASWF_ORG" != ""  ]] ; then
         time sudo yum install -y ${EXTRA_DEP_PACKAGES}
     fi
 
-    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != "" ]] ; then
+    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != "" || "$CXX" == "icpx" || "$CC" == "icx" || "$USE_ICX" != "" ]] ; then
         sudo cp src/build-scripts/oneAPI.repo /etc/yum.repos.d
-        # sudo yum install -y intel-hpckit
-        time sudo yum install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
+        sudo yum install -y intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic
         set +e; source /opt/intel/oneapi/setvars.sh; set -e
-        icpc --version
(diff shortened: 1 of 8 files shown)


src/build-scripts/gh-installdeps.bash
```diff
@@ -86,14 +84,13 @@ else
         time sudo apt-get install -y g++-11
     fi
 
-    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != "" ]] ; then
+    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != "" || "$USE_ICX" != "" ]] ; then
```
AlexMWells, line 87: Is this missing checks for icx?
```suggestion
    if [[ "$CXX" == "icpc" || "$CC" == "icc" || "$USE_ICC" != ""  || "$CXX" == "icpx" || "$CC" == "icx" || "$USE_ICX" != ""  ]] ; then
```

AlexMWells (approved):
LGTM

src/build-scripts/gh-installdeps.bash
lgritz: I don't use USE_ICX.

Ideally, all we should do is set CXX=icpc CC=icc, right? But for icc, I was (at least at first) having trouble building the dependencies with icc, which of course would happen if we set CXX and CC. So I switched to a method of setting USE_ICC, and in gh-installdeps.bash, that is the signal, including to switch CXX and CC at the last minute, after all dependencies are built but before we move on to building the main package.

None of the appeared necessary for icx, so I skipped the whole charade and just directly set CXX=icpx and let everything build that way.

I should go back and double check icc again. It may be that a better approach is to use icc on the dependencies, but instead make sure they are not using -Werror (or patch those projects to fix the icc-specific errors). But I was saving that for another day.

PR #3372, opened 2022-03-25
Fix conflict between RESTful and Windows long path notation

lgritz:
When trying to break filenames into the base name and any "RESTful"
arguments (such as "/foo/bar.exr?name=value&dog=cat"), we were
just looking for presence of the "?" to determine if it was a RESTful
thingie and using that as the split position.

But this is thrown off by an odd Windows notation where names that
start with (get this) `\\?\` somehow designate long filepaths. See
See: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

So the solution is to change Strutil::get_rest_arguments to explicitly
check for this pattern at the start of the string.

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -269,7 +269,11 @@ bool
 Strutil::get_rest_arguments(const std::string& str, std::string& base,
                             std::map<std::string, std::string>& result)
 {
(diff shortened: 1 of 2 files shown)


lgritz:
Closed accidentally. Reopening.

Issue #3347, opened 2022-02-24
Legal BMP file cannot be opened


DougRogers:
ImageInput::open fails for this fil
[gracehopper.zip](https://github.com/OpenImageIO/oiio/files/8128893/gracehopper.zip)
e

lgritz:
Aha!  Solution here: https://github.com/OpenImageIO/oiio/pull/3375

It was an "undocumented" header configuration that PhotoShop is happy to write despite not being described in MSDN.

lgritz:
@DougRogers Is that bmp file ok to check into our testsuite? It doesn't seem to have any attached license or ownership information, though I assume that the likeness of her is from a government photo and thus probably in the public domain.

DougRogers:
Should be fine.  The image is posted here, albeit in a different format.

https://commons.wikimedia.org/wiki/File:Grace_Hopper.jpg

DougRogers:
I think that bmp is poster here:
https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/platform/testing/data/

PR #3375, opened 2022-03-27
BMP files - read additional varieties

lgritz:
There are two undocumented varieties of BMP files only written by
certain Adobe apps that we did not properly handle, identified by the
size value in their header.

This is the only explanation I could find about it:
https://web.archive.org/web/20150127132443/https://forums.adobe.com/message/3272950

Fixes https://github.com/OpenImageIO/oiio/issues/3347


lgritz:
Fixes #3347

PR #3379, opened 2022-03-30
ImageBufAlgo: Initial implementation of `st_warp` transform function

nrusch:
## Description

Add a new `ImageBufAlgo::st_warp` function, intended to closely match the behavior of Nuke's `STMap` node.

The function takes a source `ImageBuf` to warp, and a secondary `ImageBuf` containing channels to treat as normalized image space (i.e. `st`) coordinates. For each pixel in the `st` buffer, the coordinates are used to look up a source pixel position, which is then sampled using a filter to produce a new output pixel.

The transform is currently only defined over the geometry of the `st` buffer (again, matching the behavior of the Nuke node), with the expectation that any desired scaling should be done to the `st` buffer prior to warping.

One notable difference between the Nuke node and this implementation is the filtering behavior: for better or worse, Nuke's filtering API does some extra level of interpolation/smoothing on top of the raw filter results (possibly just an extra linear filtering pass).

This set of changes has been developed off of `dev-2.3` and then rebased onto `master` for the PR, so it is known to be portable between those branches.

## Tests

I have not added any new tests at this point. I didn't see any examples of testing the spatial IBA functions in `imagebufalgo_test.cpp`, so I'm not sure what the policy is there, or how best to ensure strict correctness.

## Checklist:


src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -790,6 +790,68 @@ bool OIIO_API warp (ImageBuf &dst, const ImageBuf &src, M33fParam M,
 /// @}
 
 
+/// @defgroup st_warp (st_warp: warp an image using per-pixel st coordinates)
```
nrusch, line 793: Is there anything I need to do to update the docs besides adding the Doxygen comments here?

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1144,261 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+        roi         = roi_intersection(inputROI, stSpec.roi());
+        roi.chbegin = std::max(roi.chbegin, 0);
+        roi.chend   = std::min(roi.chend, src.spec().nchannels);
+
+        ImageSpec destSpec(dst.spec());
+        destSpec.set_roi(roi);
+        destSpec.set_roi_full(stSpec.roi_full());
+        // XXX: It would be nice to be able to tell `IBAprep` to skip the buffer
+        // initialization. Worth adding a new prep flag?
+        dst.reset(destSpec);
```
nrusch, line 1311: I don't love this double-reset behavior, but I was trying to lean on `IBAprep` as much as possible, while still applying my own ROI overrides.

I'd like to improve this in some way, particularly since I think the worst-case scenario is the buffer storage being reallocated three times (since this function gets called twice when using the overloads that take a filter name and width).

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1144,261 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+            float total_weight = 0.0f;
+            for (; !src_iter.done(); ++src_iter) {
+                const float weight = (*filter)(src_iter.x() - src_x + 0.5f,
+                                               src_iter.y() - src_y + 0.5f);
+                total_weight += weight;
+                for (int idx = 0, chan = roi.chbegin; chan < roi.chend;
+                     ++chan, ++idx) {
+                    sample_accum[idx] += src_iter[chan] * weight;
+                }
+            }
```
nrusch, line 1231: Since this is my first time using the low-level filter primitives directly, it would be great to know if I'm overlooking anything here (or just doing something dumb).

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1144,261 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+    if (!check_st_warp_args(dst, src, stbuf, chan_s, chan_t, roi)) {
+        return false;
+    }
+
+    // Resize ratios
+    float wratio = float(dst.spec().full_width) / src.spec().full_width;
+    float hratio = float(dst.spec().full_height) / src.spec().full_height;
+
+    auto filter = get_resize_filter(filtername, filterwidth, dst, wratio,
+                                    hratio);
```
nrusch, line 1360: I might end up canning the use of `get_resize_filter` in favor of a uniform default filter, at least in part to remove the extra call to `check_st_warp_args`.

src/include/OpenImageIO/imagebufalgo.h
lgritz: No, that's probably enough for the IBA part.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: OK, I ended up pulling out a static `get_warp_filter` helper, which is like `get_resize_filter`, but always defaults to `lanczos3`, and is used by both `warp` and `st_warp`.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1155,255 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+            ++out_iter;
+        }
+    });  // end of parallel_image
+    return true;
+}
+
+
+
+static bool
+check_st_warp_args(ImageBuf& dst, const ImageBuf& src, const ImageBuf& stbuf,
```
nrusch, line 1263: I'm at least going to add some unit tests for the extra validation done by this function.

lgritz:
I would recommend (if you think you can tackle it) adding `oiiotool --st_warp`, and then it becomes much easier to have robust tests for this by adding something appropriate to testsuite/oiiotool-xform/run.py.

nrusch:
I've added `oiiotool --st_warp`, and also added a new test to `testsuite/oiiotool-xform/run.py`.

However, I wanted to run something strange by you.

Initially, I generated my test reference from the existing reference `resize.tif` in the OIIO repo:
`oiiotool -i testsuite/oiiotool-xform/ref/resize.tif --pattern fill:topleft=0,0,0:topright=1,0,0:bottomleft=0,1,0:bottomright=1,1,0 256x256 3 --powc 1.2 --st_warp -o testsuite/oiiotool-xform/ref/st_warped.tif`

However, using that as my reference image was causing consistent test failures due to differences above the minimum difference threshold when compared with the output that used the `resize.tif` generated by the testsuite:

```
Comparing "st_warped.tif" and "ref/st_warped.tif"
256 x 256, 4 channel
  Mean error = 0.000463329
  RMS error = 0.00135965
  Peak SNR = 57.3314
  Max error  = 0.0078432 @ (13, 41, R)  values are 0.752941, 0.423529, 0.423529, 0 vs 0.745098, 0.423529, 0.423529, 0
  0 pixels (0%) over 0.008
  190 pixels (0.29%) over 0.004
FAILURE
```

I checked out the raw diff in Nuke, and the pixel differences are all 0-2 8-bit code values apart.

I then went back and regenerated my reference image from the `resize.tif` generated by the testsuite:

`oiiotool -i build/platform-linux/gcc-6.3.1/boost\=\=1.70.0/python-2.7/\!nuke/testsuite/oiiotool-xform/resize.tif --pattern fill:topleft=0,0,0:topright=1,0,0:bottomleft=0,1,0:bottomright=1,1,0 256x256 3 --powc 1.2 --st_warp -o src/testsuite/oiiotool-xform/ref/st_warped.tif`

Using this new reference image does not cause the same test failure, and this is the reference image I have added to the PR.

I wanted to mention this because I don't know how to explain it; the reference `testsuite/oiiotool-xform/ref/resize.tif` and the `resize.tif` generated by the testsuite on my end compare identical in Nuke and using `idiff -warn 1e-08`, so I'm not sure how they could be producing different results after the `st_warp`.

I also tested what would happen if I generated my reference image directly from the original `grid.tif` found in the `oiio-images` repo:
`oiiotool -i ../oiio-images/grid.tif --resize 256x256 --pattern fill:topleft=0,0,0:topright=1,0,0:bottomleft=0,1,0:bottomright=1,1,0 256x256 3 --powc 1.2 --st_warp -o testsuite/oiiotool-xform/ref/st_warped.tif`

This also causes test failures when compared with the `resize.tif` generated by the testsuite, with even greater pixel differences. However, if I change my test case to *also* start from `grid.tif` and do the resize on the fly before warping, it works.

Let me know if you have any insights or concerns.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1155,255 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+        // If a null filter was provided, fall back to a reasonable default.
+        filterptr.reset(Filter2D::create("lanczos3", 6.0f, 6.0f));
+        filter = filterptr.get();
+    }
+
+    bool ok;
+    OIIO_DISPATCH_COMMON_TYPES3(ok, "st_warp", st_warp_, dst.spec().format,
+                                src.spec().format, stbuf.spec().format, dst,
+                                src, stbuf, chan_s, chan_t, flip_s, flip_t,
+                                filter, roi, nthreads);
```
lgritz, line 1353: `OIIO_DISPATCH_COMMON_TYPES3` will handle all combinations of the types of the three images. But if I understand correctly, you already disallow stbuf being anything other than float. So that means that `OIIO_DISPATCH_COMMON_TYPES2` should work and will make many fewer template expansions. (Or else, stick with 3, and then you can remove the restriction on the type of stbuf, and I think it should all work.)

lgritz (changes requested):
This looks really good. Some minor changes are suggested (see comments inline).

I would encourage you to also add docs for the new oiiotool command in oiiotool.rst (though I can give that a stab after this merges, if you just don't have the time for it).

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1155,255 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+
+    if (dst.initialized()) {
+        // XXX: Currently requiring the pixel scales of ST and a preallocated
+        // output buffer to match.
+        if (dst.spec().full_width != stSpec.full_width
+            || dst.spec().full_height != stSpec.full_height) {
+            dst.error("ImageBufAlgo::st_warp : Output and ST buffers must have "
+                      "the same full width and height");
+            return false;
+        }
```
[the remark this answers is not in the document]
nrusch: To be honest, I'm not entirely sure. My instinct is that we can't rely on iterator clamping, since we want to be able to write outside of the display window.

That said, I think this is currently somewhat meaningless at the moment, since I'm unconditionally resetting `dst` below, so this may go away. I'll circle back to this after figuring out how best to handle/skip the buffer allocation below.

src/oiiotool/oiiotool.cpp
```diff
@@ -3674,6 +3674,19 @@ OIIOTOOL_OP(warp, 1, [](OiiotoolOp& op, span<ImageBuf*> img) {
 
 
 
+// --st_warp
+OIIOTOOL_OP(st_warp, 2, [](OiiotoolOp& op, span<ImageBuf*> img) {
+    std::string filtername = op.options()["filter"];
+    int chan_s             = op.options().get_int("chan_s");
+    int chan_t             = op.options().get_int("chan_t", 1);
+    bool flip_s            = op.options().get_int("flip_s");
+    bool flip_t            = op.options().get_int("flip_t");
```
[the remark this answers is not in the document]
nrusch: Can do. I was basically just copying the behavior of the existing ops.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: Thanks, good catch. I originally thought it would be important to maintain the exact pixel datatype in the ST buffer when computing the lookup coordinates, but since I decided to punt and just use float across the board, I'll definitely change this.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: > If dst is already initialized and we're writing into it, won't this throw that out and reallocate a new blank image?

Yes, it will.

> What are the circumstances in which there's an extra reset here?

`IBAprep` will reset it if it's not already initialized, and then I come along and do it again.

I think this is something I could use your perspective on.

The basic issue is that I need to make sure `dst` is "big" enough to fit the ST buffer, and this was just the blunt-instrument solution.

For an already-initialized `dst` buffer, there are two possibilities:

1. The ST buffer's ROI is a subset of `dst`'s.
    * In this case, we should hypothetically be able to continue on without doing any more sanity-checking, and the actual `st_warp` operation would only fill in the subset of the image area defined by the ST buffer's ROI (if that sounds like reasonable behavior).
    * I think `IBAprep` will still alter the ROI argument to match the destination buffer's, which isn't actually what I want to operate on, so I would still need to modify it after the fact.
2. The ST buffer's ROI is a proper superset of `dst`'s (that is, it contains coordinates that are outside of `dst`'s ROI).
    * Since I don't want the operation to be clamped to `dst`'s display window, I think using `ImageBuf::Iterator` to avoid out-of-bounds writes is off the table here.

Maybe the right thing to do in either case would just be to clamp the loop iterator to the intersection of the `dst` and ST buffer ROIs? I think that would allow me to be a little less heavy-handed in this validation/setup code, but maybe I'm under- or overthinking this?

> Yes, it's totally fine to make a new flag value for IBAprep_NO_RESET_DST or something like that. That's not an external API, we can change it however we need to make implementation of these functions easier.

OK, good to know. For context This idea stems from the behavior mentioned in `1.` above, where `IBAprep` will basically use the `dst` buffer to resize the ROI argument. In the cast of an uninitialized `dst` buffer, I want to set up the ROI of the `dst` buffer myself and only allocate the storage one time, but I'd still like to be able to take advantage of some of what `IBAprep` does in terms of validation, metadata copying, etc.

I will say that I'm not sure how `IBAprep` would go about setting up a spec for `dst` without allocating local storage for it (keeping in mind that I'd like to potentially modify the assigned spec further before allocating local storage). Maybe just using `dst.specmod() = spec;` would be appropriate? If so, what would be the most idiomatic way for the caller to actually allocate the underlying storage later on? `ImageBufAlgo::zero`?

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -790,6 +790,68 @@ bool OIIO_API warp (ImageBuf &dst, const ImageBuf &src, M33fParam M,
...
+///             coordinates are defined in terms of a different image origin 
+///             than OpenImageIO's.
+
+ImageBuf OIIO_API st_warp (const ImageBuf &src, const ImageBuf& stbuf,
+                           string_view filtername=string_view(),
+                           float filterwidth=0.0f, int chan_s=0, int chan_t=1,
+                           bool flip_s=false, bool flip_t=false, ROI roi={},
+                           int nthreads=0);
+ImageBuf OIIO_API st_warp (const ImageBuf &src, const ImageBuf& stbuf,
+                           Filter2D *filter, int chan_s=0, int chan_t=1,
```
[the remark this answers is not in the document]
nrusch: Can do.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: OK, after taking another look at this, I'm not sure I can trivially switch to `OIIO_DISPATCH_COMMON_TYPES2` here.

> But if I understand correctly, you already disallow stbuf being anything other than float.

I only check that `stbuf` is a floating-point datatype, so `half`, `float`, and `double` are all acceptable. Thus, I can't just convert the `stbuf` iterator to `ImageBuf::ConstIterator<float>`, since that will misrepresent the buffer's source data and lead to garbage results (e.g. when using half ST data).

I'm not sure there's a way out of this without potentially adding another special-purpose dispatch template/macro, but let me know what you think.

src/libOpenImageIO/imagebufalgo_xform.cpp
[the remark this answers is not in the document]
lgritz: What I meant is: iterate over dst, which will be initialized to src's size if it doesn't already exist, but there's no reason for stbuf to need to be exactly the same size. If there's a part of dst (or the iteration roi) that's outside the range of stbuf, then you just use the stbuf's iterator's ability to have a wrap mode. You need not consider it an error for stbuf's full_width to be different from dst. I think?

src/libOpenImageIO/imagebufalgo_xform.cpp
lgritz: I think there are other examples of IBA function that take 2 input images and make one output, and I'm not sure why this shouldn't just use the same logic about how big to make dst.

src/libOpenImageIO/imagebufalgo_xform.cpp
lgritz: Ah, I see now. You're right, I misread the is_float_point as a hard test for float.

If you use COMMON_TYPES3, then it will automatically handle any time for stbuf, it doesn't really have to be float-like.

lgritz: ...automatically handle any TYPE for stbuf...

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: OK thanks, I'll have another look.

One thing I keep coming back to when it comes to the idea of `dst`-centric iteration is that the transform is undefined outside the ROI of `stbuf`. Thus, I really only want to operate within the bounds of the ST ROI no matter what, and I don't want to rely on any kind of auto-clamping/wrapping of the `stbuf` iterator to give me default pixel values, etc.

I think pre-clamping the operating ROI to the intersection of the `dst` and `stbuf` ROIs sounds like it should give me the right balance here, as well as getting rid of this extra `dst` reallocation. I'm working on this approach right now.

lgritz: OK, I defer to your best judgment. Clamp the iteration to the well-defined part of stbuf, and we can always revisit this to make it more flexible if user experience shows that to be an issue.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: I just realized that I think I misunderstood how the auto-conversion done by `ImageBuf::Iterator` works between fixed- and floating-point types...

The extra `is_floating_point` check I've been doing was so I could avoid having to normalize fixed-point values from the ST buffer, but I see now that the `conver_type` helpers look to be scaling fixed-point values to/from the normalized floating-point range (at least for unsigned fixed-point types). Thus, I think that extra restriction can go, and I can just let accept any datatype for `stbuf`.

lgritz: Yes, exactly, the normalization of integer types to [0.0-1.0] floating point values automatically happens when you dereference the iterator, that's why it's templated on the underlying data type! You can, in effect, write your "kernel" of an IBA function as if everything were floating point, and the iterators handle the conversion in and out (at the expense of the big template expansion handled by the `OIIO_DISPATCH_*` macros).

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: OK, I've simplified this logic and removed the extra buffer reset.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -1144,4 +1155,232 @@ ImageBufAlgo::rotate(const ImageBuf& src, float angle, string_view filtername,
...
+        const int ybegin = src.ybegin();
+        const int yend   = src.yend();
+
+        // The horizontal and vertical filter radii, in source pixels.
+        // We will sample and filter the source over
+        //   [x-filterrad_x, x+filterrad_x] X [y-filterrad_y,y+filterrad_y].
+        const int filterrad_x = (int)ceilf(filter->width() / 2.0f / xscale);
+        const int filterrad_y = (int)ceilf(filter->height() / 2.0f / yscale);
+
+        ImageBuf::ConstIterator<SRCTYPE> src_iter(src);
```
nrusch, line 1188: Question: Since the data from `src_iter` is being accumulated directly into a buffer typed to match `DSTTYPE`, should this be typed as `ImageBuf::ConstIterator<SRCTYPE, DSTTYPE>` instead?

nrusch:
> I would encourage you to also add docs for the new oiiotool command in oiiotool.rst (though I can give that a stab after this merges, if you just don't have the time for it).

Whoops, I missed this one. I'll take a look at this tomorrow.

src/libOpenImageIO/imagebufalgo_xform.cpp
nrusch: I realized that this should probably actually be `ImageBuf::ConstIterator<SRCTYPE, Acc_t>`.

lgritz: yes, exactly. Treat float (or double, for doubles) as the type you do all the real math in, and then the iterators just convert between the buffer format and float.

nrusch: Thanks for confirming.

nrusch:
OK, I've added an `--st_warp` section to `oiiotool.rst` (and hopefully fixed the issue that broke the CI jobs).

lgritz (approved):

lgritz:
LGTM!

lgritz (approved):

lgritz:
@nrusch, do you want this backported to 2.3?

nrusch:
Yes please, that would be great. Thanks!

PR #3381, opened 2022-04-08
Lay groundwork for unity builds

lgritz:
As I learned recently, a "unity" (aka "jumbo") build is where multiple
.cpp files are combined into one translation unit -- imagine a unity.cpp
that simply has

    #include "file1.cpp"
    #include "file2.cpp"
    ...
    #include "fileN.cpp"

and so you compile unity.cpp instead of the separate file?.cpp files
individually.

Turns out that CMake understands this concept and can do it for you
automatically!

The benefit of a unity build is that file1...fileN probably include
most of the same headers, expand the same templates, etc., so a bunch
of the per-file work of the compiler can be done once rather than
redundantly for each file.

There are two potential downsides, however:

1. It may not be safe to concatenate your cpp files! For example, if
   both file1.cpp and file2.cpp each contain a `static int foo;`, that
   may have been safe when compiled separately, but is not allowed to
   happen twice in one compilation unit. Similarly, if you have headers
   that don't have proper guards against multiple includes, etc. So one
   should expect a whole lot of little fix-ups are needed for this to
   work properly. (We'll come back to that topic.)

2. Combining source file into these "jumbo" modules can make heavily
   parallelized builds on many-core machine not be able to load balance
   or keep all the cores busy. (Simplified examples: if you have 16 .cpp
   files on a 16 core machine, each core can compile one cpp file in
   parallel with the others. But if you mash the modules into just one
   huge cpp file, give that to one core, and your other 15 cores are
   idle, so the full build probably takes much longer.)

I tried this out, including the many fixups implied by (1) above, and
at first the unity builds were slower on both my laptop (8 cores) and
workstation at work (32 cores), because of downside (2) explained
earlier -- harder to take advantage of parallel builds when there are
fewer, bigger, compilation units. Some tweaking of strategy got me to
the point where I could always get the unity builds to go faster, but
not by a whole lot when many core were available. Slighty faster, but
not worth the trouble. A bit disappointing, nearly abandoned the whole
idea.

HOWEVER, in situation where you are limited to very few cores (like in our
CI, which allocates 2 cores, and it sure seems more like 1-1.5 for the
Windows and Mac CI runners), the unity builds are substantially faster --
there's already not much parallelism to exploit, so you come out ahead
with the savings of that redundant per-file work we talked about.

So my current thinking is to go ahead and make the changes that allow
unity builds. I don't particularly recommend them when highly parallel
builds are available to you, but it might help to cut our CI
turnaround time down on the GitHub runners. And maybe it will help in
other situations for other people.

Ok, then. The present patch introduces these concepts and makes the
CMake and other build system changes to allow unity builds. (N.B. It
won't WORK yet, so don't try it!) After we get that out of the way, in
subsequent PRs I'll show all the changs to the code that were
necessary to fix all the little things that went wrong when source
files got combined.


lgritz:
To expand a bit on the explanation of the parallelism problem:

Imagine you have 4 .cpp files that take 1 minute each to compile. Compiling them in sequence takes 4 minutes. If you have at least 4 cores and do a parallel build, you can compile them at the same time, taking 4 minutes of CPU time but with an elapsed "wall clock" time of 1 minute.

If you turn those 4 modules into one jumbo module, let's say that it only takes 2 minutes of CPU time to compile -- 50% reduction in total work. But now there's nothing to parallelize, so that 2 minute compile actually takes TWICE as long as did the normal parallel build you were doing originally. (Half the work, but ironically twice the time.) But if you were in a situation (like free CI on GitHub) where you were already forced to use only 1 core to compile, so wall clock time was == total work, then it could turn out to be a big win (2 minutes rather than 4 minutes).

So that's the strategy I'm currently looking at -- enable the ability for unity/jumbo builds, do not advise people to use it under normal circumstances, but we will perhaps use it for some of our CI runs (Windows and Mac, where the CI takes longest and seems to be pretty close to single core performance).

My plan is as follows:

Step 0 (this PR) -- build system changes that allow us to start and control unity builds.

Step 1...N -- fix all the things in the source modules that break in unity builds (split into several PRs to keep the changes simple and reviewable).

Step N+1 -- make the changes to ci.yml to enable it for Windows and Mac CI builds to speed them up.

PR #3382, opened 2022-04-08
Fix a number of small warnings on windows

fpsunflower:
Fixing a few small things that MSVC complains about when building oiio.

### DIFF
--- a/src/term.imageio/termoutput.cpp
+++ b/src/term.imageio/termoutput.cpp
@@ -204,7 +204,7 @@ TermOutput::output()
                                  TypeDesc::UINT8, &rgb);
                 std::cout << term.ansi_fgcolor(rgb[0][0], rgb[0][1], rgb[0][2]);
                 std::cout << term.ansi_bgcolor(rgb[1][0], rgb[1][1], rgb[1][2])
-                          << "\x5C\x75\x32\x35\x38\x30";  // utf8 encoding
+                          << "\x5C\x75\x32\x35\x38\x30";  // utf8 encoding of "\u2580"
             }
             std::cout << term.ansi("default") << "\n";
         }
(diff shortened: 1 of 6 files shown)


fpsunflower:
@lg - not sure what my CLA status is for OIIO, let me know if you need paperwork from my employer ...

lgritz:
On OIIO, it's always been the case that I've strongly encouraged/insisted on CLAs for large new features, but generally I consider it optional for small fixes to existing features -- i.e., changes that seem too small to plausibly be the subject of a serious IP dispute and that we could revert if we really needed to without a major disruption. 

I trust you to be the judge of what you need to arrange with your employer to make it ok on your end (I don't want you to get into hot water because of this project). I don't seem to have a CLA from Epic, so it might not be the worst idea to set one up to pave the way for any future bigger contributions. I'm not sure, but I think I've accepted changes from Epic folks before, so maybe it's just that they were all small enough to fall below the CLA insistence threshold.

(Aside: After 14 years of nobody raising one iota of an objection about anybody's contributions, and a growing number of major projects in the world using DCO instead of CLA, I have wondered if we should eliminate the CLA and go DCO only. Something to think about moving forward.)

fpsunflower:
Bleeding edge CI doesn't like the utf8 change -- will try to find a cleaner workaround ...

fpsunflower:
CI is passing now. I will tackle the windows CI tweaks in another PR. It would be nice to add VS2022 to the windows matrix.

lgritz:
I tried to set up VS2022 in the matrix and had some trouble and had to leave it aside for the moment. I can't remember now exactly what the issue was, but maybe I'll try it again with your fixes in, in case it was related.

src/term.imageio/termoutput.cpp
```diff
@@ -204,7 +204,7 @@ TermOutput::output()
                                  TypeDesc::UINT8, &rgb);
                 std::cout << term.ansi_fgcolor(rgb[0][0], rgb[0][1], rgb[0][2]);
                 std::cout << term.ansi_bgcolor(rgb[1][0], rgb[1][1], rgb[1][2])
-                          << "\u2580";
+                          << "\x5C\x75\x32\x35\x38\x30";  // utf8 encoding
```
lgritz, line 207:
```suggestion
                          << "\x5C\x75\x32\x35\x38\x30";  // utf8 encoding of "\u2580"
```

lgritz (approved):
Approved, pending the suggestion

PR #3388, opened 2022-04-13
Speed up case insensitive string comparisons

lgritz:
Oh boy, never leave anything unbenchmarked.

Turns out the boost::algorithm functions we were relying on underneath
many Strutil "case-insensitive" comparisons were ridiculously slow.
We thought they were good enough because they allowed specification of
locale, so we could just pass the static classic locale, and so they
would be inexpensive because they didn't have to query the current
locale.  But this is wrong, they were still ghastly.

So here I rewrite Strutil::iequals, iless, starts_with, istarts_with,
ends_with, iends_with in terms of a new (internal) Strutil::strcasecmp
and strncasecmp (which underneath handle differences in platform, and
use the locale-independent versions). The net result is that most of
those case-independent comparisons speed up by a factor of
50-100x. Wow.

I still need to tackle the family of 'ifind' related functions. They
are a bit trickier. But I'll leave them for another time, because I
need to roll this present fix out right away to fix a real production
bottleneck.

(Worth noting: iequals is instrumental when you're searching a
ParamValueList for a particular name, which is what happens when you
look up attributes from an ImageSpec, which is what happens when you
call get_texture_info(), which is what underlies OSL gettextureinfo()
calls in the cases that they cannot be constant-folded during runtime
optimization. So this came to my attention when analyzing a slow scene
whose shaders had a pathological explosion of gettextureinfo calls that
couldn't be optimized away.)

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -399,13 +399,37 @@ Strutil::wordwrap(string_view src, int columns, int prefix, string_view sep,
 
 
 
-// DEPRECATED(2.0) -- for link compatibility
 namespace Strutil {
-std::string
-wordwrap(string_view src, int columns, int prefix)
+
+// Define a locale-independent, case-insensitive comparison for all platforms.
+
+inline int
+strcasecmp(const char* a, const char* b)
 {
-    return wordwrap(src, columns, prefix, " ", "");
+#if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) \
+    || defined(__FreeBSD_kernel__) || defined(__GLIBC__)
+    return strcasecmp_l(a, b, c_loc);
+#elif defined(_WIN32)
+    return _stricmp_l(a, b, c_loc);
+#else
+#    error("need equivalent of strcasecmp_l on this platform");
+#endif
 }
+
+
+inline int
+strncasecmp(const char* a, const char* b, size_t size)
+{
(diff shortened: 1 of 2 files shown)


aconty (approved):

vinniefalco:
Now this is very interesting :) I have a need for case-insensitive algorithms on low-ASCII strings myself!! (no `locale`, ewww)

vinniefalco:
Oh....eh... I looked through the changes. I was hoping for something really cool like some algorithms that worked 4 or 8 bytes at a time or some kind of interesting mathematical relation that greatly simplified the emitted assembly for case-insensitive operations.

Maybe these could be of some help to you:

https://lemire.me/blog/2020/04/30/for-case-insensitive-string-comparisons-avoid-char-by-char-functions/

https://github.com/lemire/Code-used-on-Daniel-Lemire-s-blog/blob/master/2020/04/30/tolower.cpp

lgritz:
Hi, @vinniefalco 

Because the implementation in this PR uses glibc's strncasecmp_l underneath (or _strnicmp_l for Windows, they always have to be a little bit different), I assume that their underlying implementation already is fairly smart with not operating one char at a time.

To the best of my knowledge, not that this PR is merged, string compares are no longer a bottleneck in OIIO. If anybody comes up with examples that show it's still a problem, I'm happy to look into further refinements.

PR #3396, opened 2022-04-21
py_imagespec: added get_bytes_attribute to bypass decoding as utf-8

### LINKED ISSUE #3395: [BUG] python3 metadata values not valid utf-8
**Describe the bug**
In Python3, when a metadata value is of type string, `py::str` is used to convert from `char *` to Python3 `str`.
That conversion decodes the byte string as utf-8.
If the byte string is not valid utf-8, in my python code I have raised:
`RuntimeError: Could not allocate string object!`

**To Reproduce**
1. In Python2 use oiio to write out an image (e.g. EXR) having manually set a metadata key to a value like `"\xff\xff\xff\xff"`
2. In Python3 use oiio thus:
```
img = oiio.ImageInput(path)
print(img.spec().getattribute(metadata_key))
```
3. The RuntimeError above will be raised 

**Suggested behavior**
I've classified this as a bug rather than a feature request, because I can't currently find a way to get to the raw metadata data using the python bindings. Also, although I can catch the above exception in my code and ignore broken metadata values, it's not a particularly informative exception to catch.

I'm aware in C++ world there is only one type for string, so in Python that should one-to-one map to a Python type. I'm also aware that str is probably the friendliest choice, as having to deal with bytes objects every time would be a bit of a nightmare.

So I guess I'm not entirely sure what I'd suggest. Ideas are:

1. always return bytes for oiio type string (deal with utf-8 in user Python code)
2. validate the byte string, valid utf-8 returns str, invalid returns bytes (feels a little badly defined. User Python code can't guarantee what the type returned will be until runtime)
3. add an extra method to the Python bindings which always returns bytes for type string. e.g. `getattribute_raw`. Maybe `ParamValue` also has a `value_raw` too

splidje:
py_imagespec: added get_bytes_attribute to bypass decoding as utf-8 and instead returning as a Python3 bytes object

Fixes #3395

### DIFF
--- a/src/doc/pythonbindings.rst
+++ b/src/doc/pythonbindings.rst
@@ -587,10 +587,15 @@ Section :ref:`sec-ImageSpec`, is replicated for Python.
 .. py:method:: ImageSpec.get_int_attribute (name, defaultval=0)
                   ImageSpec.get_float_attribute (name, defaultval=0.0)
                   ImageSpec.get_string_attribute (name, defaultval="")
+                  ImageSpec.get_bytes_attribute (name, defaultval="")
 
     Retrieves a named metadata value from `extra_attribs`, if it is
     found and is of the given type; returns the default value (or a passed
     value) if not found.
+
+    For an attribute of type STRING, get_bytes_attribute in Python3 skips
+    decoding the underlying C string as UTF-8 and returns a `bytes` object
+    containing the raw byte string.
 
     Example:
 
@@ -3816,6 +3821,7 @@ details.
                get_int_attribute (name, defaultval=0)
                get_float_attribute (name, defaultval=0.0)
                get_string_attribute (name, defaultval="")
+               get_bytes_attribute (name, defaultval="")
 
     Retrieves an attribute value from the named set of global OIIO options.
     (See Section :ref:`sec-globalattribs`.) The `getattribute()` function
@@ -3823,6 +3829,10 @@ details.
     not exist.  The typed variety will only succeed if the attribute is
     actually of that type specified. Type varity with the type in the name
     also takes a default value.
+
+    For an attribute of type STRING, get_bytes_attribute in Python3 skips
+    decoding the underlying C string as UTF-8 and returns a `bytes` object
+    containing the raw byte string.
 
     Example:
 
--- a/src/python/py_imagespec.cpp
+++ b/src/python/py_imagespec.cpp
@@ -223,6 +223,14 @@ declare_imagespec(py::module& m)
                     std::string(spec.get_string_attribute(name, def)));
             },
             "name"_a, "defaultval"_a = "")
+        .def(
+            "get_bytes_attribute",
+            [](const ImageSpec& spec, const std::string& name,
+               const std::string& def) {
+                return py::bytes(
+                    std::string(spec.get_string_attribute(name, def)));
+            },
+            "name"_a, "defaultval"_a = "")
         .def("getattribute", &ImageSpec_getattribute_typed, "name"_a,
              "type"_a = TypeUnknown)
         .def(
--- a/src/python/py_oiio.cpp
+++ b/src/python/py_oiio.cpp
@@ -307,6 +307,12 @@ OIIO_DECLARE_PYMODULE(PYMODULE_NAME)
             return PY_STR(std::string(OIIO::get_string_attribute(name, def)));
         },
         py::arg("name"), py::arg("defaultval") = "");
+    m.def(
+        "get_bytes_attribute",
+        [](const std::string& name, const std::string& def) {
+            return py::bytes(std::string(OIIO::get_string_attribute(name, def)));
+        },
+        py::arg("name"), py::arg("defaultval") = "");
     m.def("getattribute", &oiio_getattribute_typed);
     m.def(
         "is_imageio_format_name",
--- a/src/python/py_oiio.cpp
+++ b/src/python/py_oiio.cpp
@@ -310,7 +310,8 @@ OIIO_DECLARE_PYMODULE(PYMODULE_NAME)
     m.def(
         "get_bytes_attribute",
         [](const std::string& name, const std::string& def) {
-            return py::bytes(std::string(OIIO::get_string_attribute(name, def)));
+            return py::bytes(
+                std::string(OIIO::get_string_attribute(name, def)));
         },
         py::arg("name"), py::arg("defaultval") = "");
     m.def("getattribute", &oiio_getattribute_typed);


lgritz (changes requested):
I think this is a good compromise to solve the problem you reported without changing behavior of the existing getattribute.

Can you also take a stab at adding this to the documentation (src/doc/pythonbindings.rst)?

src/doc/pythonbindings.rst
```diff
@@ -587,11 +587,16 @@ Section :ref:`sec-ImageSpec`, is replicated for Python.
...
+                  ImageSpec.get_bytes_attribute (name, defaultval="")
 
     Retrieves a named metadata value from `extra_attribs`, if it is
     found and is of the given type; returns the default value (or a passed
     value) if not found.
 
+    For an attribute of type STRING, get_bytes_attribute in Python3 skips
+    decoding the underlying C string as UTF-8 and returns a `bytes` object
+    containing the raw byte string.
+
```
lgritz, line 599: Thanks.

Just for my own understanding, how is the behavior different in Python 2? 

And exactly when would a user prefer to call get_bytes_attribute rather than get_string_attribute? Is it only after an exception was caught when calling get_string_attribute, indicating that the attribute is not value UTF-8?

lgritz: Is there merit to declaring that all string attributes in the spec or a ParamValueList MUST be UTF-8, and therefore the strings should be sanitized prior to being added to the list/spec in the first place?

splidje: I'm python 2 there's already an if statement that defines `PY_STR` as `py::bytes` so `get_bytes_attribute` will behave the same as `get_string_attribute`

An example of how I'd use this in my python3 code, would be probably to always use `get_bytes_attribute` every time and deal with decoding myself (e.g. `str(b, encoding="utf-8", errors="ignore")`)

Perhaps this could be done instead by oiio under the hood, but I'm not sure if metadata values should always be forced to be utf-8. Maybe some values are intended to have some binary data.

With my example, I have an EXR with a metadata key `AudioInfo` and its value is `\xff\xff\xff...` repeated with `@` at the end. Clearly not particularly informative, and my ignoring decode errors leaves me with the string `"@"`. But should oiio be taking it upon itself to ditch all those `\xff`s and giving me just `"@"`?

lgritz: I think that it would be reasonable to require *string* metadata to be UTF-8. We already separately support metadata that's just an array of bytes, binary data (in fact, that's how we store ICCProfile metadata).

I'd need to look at the OpenEXR docs to see what they say about the contents of string metadata. I'm not sure they were thinking about UTF-8 at the time. But in general, a string isn't a good way to store arbitrary binary byte data, since a 0 character in the middle of a string will mess all sorts of things up.

lgritz: Hmmm, put another way: If you always retrieve strings as get_bytes_attribute and then do the str conversion... can you reliably distinguish between metadata that's already UTF-8, versus a binary blob that happens to be a sequence that is UTF-8 valid? Is there a danger of doing the conversion to UTF-8 one to many times, if you know what I mean?

splidje: Ah, ok. If there's already a type for binary data, maybe this is unnecessary and oiio should sanitise string metadata values.

So is OpenEXR returning string values, but not caring about Unicode?

With my one example, I don't think the value is important. Could quite easily accept it's invalid and needs sanitising.

i.e. I'm not using that particular key. I'm just trying to avoid it erroring.

splidje: Thinking about it a bit further, what's causing the error is `py::str` being used just in Python3 to convert the `char *`.

Therefore in Python2 the `char *`s just come out as is, there's no concept of converting from UTF-8.

I can see two options:

1. oiio in Python2 doesn't care about metadata string values being UTF-8 or not, so in Python3 it doesn't either, and all string metadata values are returned as `bytes`
2. oiio in Python3 now cares about metadata string values being UTF-8 and will sanitise them before returning as `str` - however, if you have some code relying on a metadata value that isn't UTF-8, you can still access the raw byte string using `get_bytes_attribute`

splidje: I can see a byte array in oiio is done with `TypeDesc(TypeDesc::UINT8, total_length)`

However, looking at the OpenEXR code, I only see the attribute types:

`EXR_ATTR_STRING` and `EXR_ATTR_STRING_VECTOR`

None of them seem to define an arbitrary byte array.

Seeing as I have at least this one example of an EXR metadata value not being valid UTF-8, that suggests OpenEXR's definition of a string is just an array of bytes.

Therefore, in Python3 land (and I guess Python2 to be consistent) all EXR string metadata values would need to be `TypeDesc(TypeDesc::UINT8, total_length)` to truly represent their type, but that clearly wouldn't be backwards compatible and seems silly.

Are any EXR string metadata values ever intended to store binary data? They're not zero terminated strings, their length is specified and then the data. You could store any byte array in an EXR metadata string - I think.

As we have no way of knowing whether an EXR string metadata value is intended to be text or raw data, I guess best thing to do in Python3 is assume text, and diligently decode as UTF-8. In which case, this `get_bytes_attribute` provides a useful way to avoid decoding errors for code expecting a particular metadata key to contain raw data.

e.g. Naively I want to eye-ball some `CameraName` metadata key, and there it is in text full of emojis perhaps. However, I'm also specifically looking for a key called `IconPNG` which I know will contain the PNG data for a small icon. I'm in Python3 using a new version of oiio, so I know I can pull out this data using `get_bytes_attribute`

lgritz: I would like strings to be strings, if possible.

OIIO has a standard way of handling non-text data that's just a raw byte array, as you found. But OpenEXR does not (as far as I know; I've asked about this on the openexr mail list to verify). 

(Aside: I think the best approach is for OpenEXR to add a "byte array" attribute type, and declare string attributes to be UTF-8. But that will probably take some time and can't be 100% relied upon until a new OpenEXR release that supports this is widely adopted.)

I think it's an unnecessary burden on all python programs for OIIO to return bytes everyplace it now returns python str's, and expect the python script to explicitly convert to string. Conceptually, these are strings, so they should just return strings.

So here is what I propose:

For the python bindings, before returning a string from getattribute(), we should verify that it's valid UTF-8, and if it's not, then return it as a byte array rather than a string.

We can also keep your get_bytes_attribute, maybe there times when it's helpful to ask for that directly.

BTW, IconPNG -- do you really need it to be a fully encoded bit pattern of a PNG file? Or do you just need a preview image? Because there's already an OpenEXR "PreviewImage" attribute! (Though I think OIIO neglects to support it at the moment, but we could.)

splidje:
Sounds good to me!

src/doc/pythonbindings.rst
splidje: Oh, that `IconPNG` was just me reaching for an invented example of a reason to have binary data in an EXR metadata key

PR #3397, opened 2022-04-22
ustring::from_hash()

lgritz:
If you know the hash of a ustring, this tells you the ustring.
(Caveat: in the very unlikely chance of two ustrings having the same
hash, this will return the first one it finds in the table.)


cmstein (approved):
LGTM!

PR #3403, opened 2022-04-30
New bitcast is closer to std::bit_cast

lgritz:
We have a template OIIO::bit_cast which has served us for a while, but
it turns out that C++20 has std::bit_cast which operates ALMOST the
same -- unfortunately, the tempalte arguments are reversed from the
way we did it!

So here we:

* Set up a new `bitcast<>` that works exactly like C++20 std::bit_cast.
* Implement our old bit_cast in terms of the new one, for now.
* Set up `#if` trick so that the confusing OIIO::bit_cast will be
  marked as deprecated beginning with OIIO 2.5, and will be removed
  entirely from OIIO 3.0. (Farther out than that, we may choose to
  re-introduce bit_cast as truly being identical semantics to
  std::bit_cast. But before that is possible, we need to completely
  ween downstream users off the old one.)
* Took the opportunity to expand the set of compilers for which
  certain bitcast overloads use Intel intrinsics. Previously, we only
  did this for the Intel compiler, but it seems to work for modern
  versions of clang and gcc as well. Do a little bitcast testing in
  fmath_test.cpp to verify that the use of these intrinsics doesn't
  break for any of the compilers we use for CI.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -757,9 +757,12 @@ inline OIIO_HOSTDEVICE float sign (float x)
 //
 // Type and range conversion helper functions and classes.
 
-
-template <typename IN_TYPE, typename OUT_TYPE>
-OIIO_FORCEINLINE OIIO_HOSTDEVICE OUT_TYPE bit_cast (const IN_TYPE& in) {
+/// Standards-compliant bit cast of two equally sized types. This is used
+/// equivalently to C++20 std::bit_cast, but it works prior to C++20 and
+/// it has the right decorators to work with Cuda.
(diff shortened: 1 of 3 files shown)


AlexMWells:
Would this be a breaking change to anyone using OIIO::bitcast<IN_TYPE, OUT_TYPE>?  
If it is, is there changes doc to add this breaking change to?

lgritz:
There was no bitcast before. We only had bit_cast, which still exists (and is now implemented in terms of the new bitcast).

lgritz:
The release notes always explain breaking changes. But this isn't yet a breaking change.

AlexMWells (approved):
LGTM

PR #3407, opened 2022-05-08
CI: adjust to upgrading Intel Clang compiler to 2022.1.0

lgritz:
The Intel OneAPI LLVM-based C++ compiler, as we download it from Intel
for our CI, bumped from 2022.0.0 to 2022.1.0.

New warning popped up -- newest Intel LLVM-based compiler warns about
overriding certain floating point math flags.

Also had to adjust thresholds on a couple texture tests that have a
few pixels change slightly due to numerical differences.


AlexMWells (commented):
LGTM

PR #3417, opened 2022-05-23
UDIM speedups by eliminating mutex

lgritz:
The udim_lookup_mutex_pool is not necessary. Just using an atomic
pointer for udiminfo.icfile is enough.

I'm seeing about a 5-8% overall speed improvement in testtex-based
intense udim benchmarks under heavy threading, a ~2% overall
improvement in a real rendered scene that uses a bunch of udim
textures (I think there are bigger bottlenecks elsewhere that I'll
move on to next, but every little bit helps).

Some testtex enhancements coming along for the ride:

* UDIM-oriented tests (automatically decuded if the test textures
  contain `<UDIM>` in their name.

* --maketests will construct random test textures and use them (handy
  if you need lots of textures for a big test, but don't want to store
  them). The tests are based on a file pattern using std::format-like
  replacement with an index. Also, if the pattern contains `<UDIM>` in
  the name, it's understood to be a udim pattern and many individual
  udim piece files will be created.

An example of these two items in action is:

```
testtex --maketests 200 "tmp/foo{:04}_<UDIM>.exr" --threadtimes 8 -iters 100000000 -runstats -v --handle -maxfiles 1000  -trials 5 -wedge
```

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -3474,15 +3474,6 @@ ImageCacheImpl::close_all()
 
 
 
-namespace {
-    // Mutex to protect all the UDIM table access.
-    static mutex_pool<spin_rw_mutex, ustring, ustringHash, 8>
-        udim_lookup_mutex_pool;
-    // static spin_rw_mutex udim_lookup_mutex;
-}  // namespace
-
-
-
 void
 ImageCacheFile::udim_setup()
 {
@@ -3658,22 +3649,9 @@ ImageCacheImpl::resolve_udim(ImageCacheFile* udimfile, Perthread* thread_info,
     if (udiminfo.filename.empty())
         return nullptr;
 
-    // Which is our mutex from the pool? Use a hash baseed on the filename.
-    spin_rw_mutex& udim_lookup_mutex(
-        udim_lookup_mutex_pool[udimfile->filename()]);
-
-    // First, try a read lock and see if we already have an entry
-    ImageCacheFile* realfile = nullptr;
-    {
-        spin_rw_mutex::read_lock_guard rlock(udim_lookup_mutex);
-        realfile = udiminfo.icfile;
-    }
-    // If that didn't work, get a write lock and we'll make the entry for
-    // the first time.
+    ImageCacheFile* realfile = udiminfo.icfile;
     if (!realfile) {
-        realfile = find_file(udiminfo.filename, thread_info);
-        // Grab the actual write lock to change the `ImageCacheFile*`.
-        spin_rw_mutex::write_lock_guard rlock(udim_lookup_mutex);
+        realfile        = find_file(udiminfo.filename, thread_info);
         udiminfo.icfile = realfile;
     }
     return realfile;
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -111,19 +111,31 @@ struct ImageCacheStatistics {
 
 struct UdimInfo {
     ustring filename;
-    ImageCacheFile* icfile = nullptr;
+    std::atomic<ImageCacheFile*> icfile { nullptr };
     int u, v;
 
-    UdimInfo()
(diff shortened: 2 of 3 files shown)


etheory:
Thanks @lgritz! I'll take a look now.
I was more hoping to COMPLETELY remove all locking around the UDIM access.
I have a feeling that the atomics will maintain a high performance penalty.
The way we handle this in glimpse is to pre-resolve the UDIMs associated with any open file, ensuring that any subsequent access to it's UDIMs can be done with no need for any locking at all.
The way I implemented this is to hold a vector of TextureHandles within a standard TextureHandle, that point to the UDIMs from within a UDIM file path.
The argument against this design is that you are opening potentially unused TextureHandles, and doing this whenever a TextureHandle to a UDIM file is accessed.
It also means you end up loading TextureHandles to files that might not be accessed, but the performance improvement of this change is absolutely spectacular.
I'll run this and compare it to the data I collected previously and let you know how effective it is.
THANK YOU SO MUCH!

lgritz:
@etheory I believe you are describing how it works now. The ImageCacheFile (that's what a TextureHandle actually points to, internally) contains a vector of UdimInfo, which if the ICF is of a "file" that's really a udim pattern, will contain one entry for each concrete file, indexed simply by the udim tile number. That UdimInfo itself is merely the filename and TextureHandle (pointer to ICT) of the concrete file that is that section of the udim.

Now, that UdimInfo.icfile (the handle to the real file) is (as of this PR) a `std::atomic<ImageCacheFile*>`. There is no lock, it's just an atomic retrieve. 

It's lazy -- meaning it does not open the files or even allocate the ICF for tile that are never accessed. So that should make us come out ahead if we don't end up touching all the files. 

Atomics that retrieve only should be pretty speedy, if everybody is only retrieving and nobody is simultaneously writing -- multiple simultaneous retrieves don't stall each other, but if anybody is writing to that cache line block (including to OTHER data in that block, "false sharing"), the write will invalidate that cache line in all the cores and cause a stall for them. But the write only happens ONCE, the first time the ICF is needed. There should be no false sharing, because the vector is an array of just the `ICF*`, a ustring for the filename, and the u,v indices of the tile, none of which are written by anybody else after the udim pattern is first written.

So DOES the atomic retrieve produce a delay in practice? I tested that!

Because on x86_64, a 64 bit read is atomic (you can't, for example, read a 64 bit value and what you get only has the upper or lower half correct because a write is still in progress -- which is a danger on some other architectures), I tried an experiment where I replaced

    ImageCacheFile* realfile = udiminfo.icfile; // <- this is really a std::atomic<ImageCacheFile*>::load()

with

    ImageCacheFile* realfile = *(ImageCacheFile**)&udiminfo.icfile; // <- this is a normal 64 bit pointer read

which bypasses the atomic and treats it as a raw pointer. That should be as fast as you can make it go, it's exactly the code you should get if there were no atomic at all, and you had pre-filled in all the pointers (except since we're still using an atomic write that one first time when we set it, it's "safe", though in a non-portable way).

It wasn't any faster.

etheory:
Thanks @lgritz that makes sense. We had a few cases recently where atomics caused insane slowdowns (surprisingly) but the pattern you show above does make sense.
I'm compiling now for a test, I'll keep you posted. Thanks!

lgritz:
If some cores are writing to the same cache line, it can drastically slow down atomic reads. That's what happens with something like an atomic counter. (Though still better than using a mutex.) But if everybody is always only reading, with just an occasional write (or one write ever), it should be pretty good performance.

etheory (approved):
@lgritz I can confirm that in my tests this definitely improves performance.
It doesn't improve the overall render time as much as I'd hoped, but it's a definite improvement, thank you!

lgritz:
good, will merge and I'm in hot pursuit of a big whale

etheory:
Aweosome news. I'll send you some more data tonight, but we are still 1.98x slower than we need to be ;-). So that'll be a nice little challenge.

PR #3436, opened 2022-06-11
ustringhash

lgritz:
ustring is a unique string (that lives in an internal hash table) whose
representation is the pointer to the characters that live in the table.

ustring hash is its alter ego -- it also refers to a ustring in the
internal hash table, but the representation is just the hash itself.

It presents an interface that is a subset of that of ustring, and in
many ways they can be used interchangeably.  For ustringhash,
accessing the characters requires using the hash to find the string in
the table -- which is not hard, but more expensive than accessing the
charcters with a ustring, where the representation is the pointer
directly to the characters.

For some uses where you don't need access to the characters in any
performance-critical paths, this may be a more convenient
representation.  In particular, it's well suited to a GPU that doesn't
have access to the character memory. Another interesting difference is
that from run to run, a ustring may have a different literal value,
since there's no reason to expect that the pointer to a string like
"foo" will refer to the same memory location every time the program
executes, but in contrast, the hash is guaranteed to be identical
from run to run.

In the process of adding unit tests for ustringhash, I ended up
overhauling all of ustring_test.cpp, as I saw that the ustring tests
were extremely cursory. Did a tiny bit of (non-breaking) touch-ups
on a few ustring methods as well.

### DIFF
--- a/src/include/OpenImageIO/ustring.h
+++ b/src/include/OpenImageIO/ustring.h
@@ -27,6 +27,10 @@
 
 OIIO_NAMESPACE_BEGIN
 
+class ustringhash;  // forward declaration
+
+
+
 /// A ustring is an alternative to char* or std::string for storing
 /// strings, in which the character sequence is unique (allowing many
 /// speed advantages for assignment, equality testing, and inequality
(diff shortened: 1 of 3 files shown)


lgritz:
@cmstein Do you think this will help you?

lgritz:
I have use for this; merging.

PR #3440, opened 2022-06-12
JPEG2000: better detect failure to open the file

### LINKED ISSUE #3427: [BUG] Exiv2 test jp2 image crashes in read_image
`        std::vector<unsigned char> pixels(width * height * channels);
        bool success = imageInput->read_image(0, 0, 0, -1, oiio::TypeDesc::UINT8, &pixels[0]);
[issue_ghsa_583f_w9pm_99r2_poc.zip](https://github.com/OpenImageIO/oiio/files/8864448/issue_ghsa_583f_w9pm_99r2_poc.zip)
`

image

lgritz:
We were missing cases where the underlying OpenJPEG library itself
had errors reading corrupted files, or errors didn't propagate around
correctly.

Fixes #3427


etheory:
@lgritz I'll try and test this out next week. Hopefully this will help the issue I was having before. Thanks.

lgritz:
@etheory, I'm confident it fixes the bug Doug found -- which was a case where a corrupted file wasn't catching the error and later crashed as a result. If you are dealing with valid jp2 files, this may be a separate issue entirely.

lgritz:
Fixes #3441

lgritz:
I've been trying to backport as many of these little fixes to 2.3 as possible, but this one is not a clean cherry-pick. Given my immense backlog of new work and that we are fairly close to branching for a 2.4 release due this summer, I'm tempted to let it go and just declare this a 2.4 fix unless @DougRogers specifically needs a 2.3 release with this patch in it.

DougRogers:
This will crash my app, but the chance of someone encountering this issue in the near future is pretty small.  I can disable jp2 support if I have to.  

2.4 is acceptable.  Do you have an ETA for that?

DougRogers:
If I ship with master, I won't have this issue, though, right?  I am fine with that.

DougRogers:
Confirm this fixes jp2 issue.  I do not need a release build.

Issue #2338, opened 2019-09-09
Support for DDS S3TC BC4 to BC7 compression formats


Silverlan:
It looks like oiio currently only supports DXT1 to DXT5, but not any of the newer ones (i.e. BC4, BC5, BC6 and BC7). While I'm not sure how commonly used BC4, BC5 and BC7 are, BC6 is the only one of the bunch which properly supports HDR colors. BC5 is great for normal maps, but DXT5n can be used as an alternative for now.

The same support would be nice for the KTX format as well, if that does get added.

The [GLI library](https://github.com/g-truc/gli) has support for all of these and could be handy as a reference.

lgritz:
Partial Update:

PR #2451 proposes a patch that lets us use an external libsquish for DDS compression, if found. Current libsquish  is v1.15, versus the embedded version we've had since 2010 which is v1.10 + some backported code from 1.13 (or so I'm told).

libsquish 1.14 added support for BC4 & BC5.

So a possible path forward here would include:
* Updating our embedded libsquish to 1.15, or alternately, remove the embedded version and rely entirely on the external one, as we do for several other formats.
* Then modifying our dds reading code to use the new BC4/5 decoding that is provided by newer editions of libsquish.

aras-p:
"Reading" and "Writing" perhaps should be considered as separate tasks. The former one is enormously easier than the latter -- code required to read/decompress BCn formats (including BC6H/BC7) is fairly small and can be made decently fast. Code required to write/compress BC6H & BC7 is _massive_ in comparison.

aras-p:
I'd like to take on this issue, but first would be good to have opinions whether my approach makes sense (cc @lgritz?).

DDS in OIIO today only supports *reading* the files, not *writing* them. So embedding/using whole of libsquish just to read the files sounds like an overkill. Plus, libsquish is probably a "dead" project by now, even upgrading it to latest version would not solve the BC4/BC5 issues -- latest libsquish (1.15) adds support for *compressing* BC4/BC5, but *decompression* (i.e. "reading") them does not work there.

My plan would be like:

1. Extend oiio dds test suite coverage. Today it only contains 1 image (dxt1/bc1 format), I'd add all the bc1..bc7 formats, plus several uncompressed ones, some with mipmaps as well. This can be done without any changes to oiio code itself; add new images into the test images repository, then extend the testsuite/dds/run.py to test the new images.
[...]

PR #3462, opened 2022-07-02
DDS: replace libsquish with bcdec, add BC4-BC7

aras-p:
## Description

Implements support for BC4, BC5, BC6H (unsigned and signed) and BC7 GPU texture compressed formats in DDS reader (#2338).

- BC4: loaded into 1 channel uint8,
- BC5: loaded into 2 channel uint8,
- BC6H: loaded into 3 channel half,
- BC7: loaded into 4 channel uint8.

This is implemented by replacing whole of libsquish (~4700 lines of code) with bcdec.h (https://github.com/iOrange/bcdec, ~1300 lines of code). bcdec.h has an advantage of being faster, smaller and supports all BCn formats for decoding.

bcdec.h license is MIT or Unlicense (aka public domain), updated all the relevant third party / license files that I could find.

## Tests

This uses just-added extended DDS test coverage (#3459) and enables previously disabled BC4..BC7 tests. The pixel value SHA1 hashes of BC1..BC3 formats do change - BC1..BC3 decoding is not specified bit-exactly anywhere, and out there in the wild there are 4 possible ways to decode it. bcdec.h decodes it slightly differently from libsquish (i.e. some color values might be 1 off in 0..255 range, sometimes). Visually I've compared the results and they are all fine.

## Checklist:


lgritz (approved):
Looks great!

Thanks for going as far as changing the license notices, etc.

lgritz:
@aras-p  Keeping in mind that certainly within the next month, we'll branch and have a beta for the first 2.4 release, which will be the primary supported release going forward (with 2.3 becoming the legacy release family that only gets critical fixes), do you think we should also backport this to 2.3? Or just let this overhaul be a 2.4 feature?

aras-p:
> @aras-p Keeping in mind that certainly within the next month, we'll branch and have a beta for the first 2.4 release, which will be the primary supported release going forward (with 2.3 becoming the legacy release family that only gets critical fixes), do you think we should also backport this to 2.3? Or just let this overhaul be a 2.4 feature?

No idea what's the expected backporting / multiple releases policy in OIIO :) In most places I've worked previously, a change like this would not get backported - it's technically not "a fix for something that's broken", but rather "implementing new functionality", and these by default don't get backported.

lgritz:
I will backport "features" on a case-by-case basis, judging risk versus user benefit, but only when it can be done without breaking backward ABI/link compatibility, and when it's unlikely to harm anybody who isn't using that particular feature. In this case, I would consider backport fairly low risk because it doesn't change any external APIs at all, and it can't possibly hurt anything other than reading DDS files specifically, which is a feature only required by a small percentage of users.

On the other hand, it also means that only a small percentage of users will benefit from the improvement, so maybe they can wait until 2.4 is released within a couple months.

Since you yourself don't need it in a release branch right now, I'll leave it for 2.4 unless somebody else requests the backport.

Thanks, Aras!

PR #3493, opened 2022-07-22
Protections against 0-length string attributes

lgritz:
Files that have string metadata with length 0 can cause trouble
if we aren't careful. Put in a number of protections.

### DIFF
--- a/src/jpeg.imageio/jpegoutput.cpp
+++ b/src/jpeg.imageio/jpegoutput.cpp
@@ -221,11 +221,10 @@ JpgOutput::open(const std::string& name, const ImageSpec& newspec,
     m_next_scanline = 0;  // next scanline we'll write
 
     // Write JPEG comment, if sent an 'ImageDescription'
-    ParamValue* comment = m_spec.find_attribute("ImageDescription",
-                                                TypeDesc::STRING);
-    if (comment && comment->data()) {
-        const char** c = (const char**)comment->data();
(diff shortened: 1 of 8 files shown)


lgritz:
@grdanny

johnhaddon:
Hi @lgritz, sorry to dredge this one up, but could you elaborate on the problem with zero-length strings? I'm specifically interested in the case of reading an empty string attribute from an EXR file, which was disabled in this PR, although we can still _write_ them, and `exrheader` seems to report them quite happily.

lgritz:
I'm not sure it was a problem, I think I just believed that empty metadata strings were being set accidentally and propagated around, cluttering "iinfo -v" output, etc. I think I assumed that setting it to "" was a shorthand for intending to clear it.

In many file formats, metadata (things like the description of an image) has fixed fields and having an empty string in the slot is truly the same thing as not having that metadata. I suppose in OpenEXR, where there is a list of metadata rather than slots for specific fields, there could be a semantic difference.

Should we preserve empty metadata strings in OpenEXR files? Is that a format in which people find it useful to denote metadata which is present, but empty?

johnhaddon:
> Should we preserve empty metadata strings in OpenEXR files?

I would think so - I can't see a good reason to not round-trip data that is round-trippable. Should I make a PR?

lgritz:
Yes, please.

PR #3508, opened 2022-08-03
Relative image compare

lgritz:
Extend ImageBufAlgo::compare() to take relative metrics, which if
nonzero check the symmetric mean relative error, in addition to the
current absolute error. To be formal, this means that a value
comparison will fail if

    abs(A-B) > failthresh && abs(A-B)/((abs(A)+abs(B))/2) > failrelative

Setting the failthresh to 0 is equivalent to only checking relative
error, and setting failrelative to 0 is equivalent to the (previous
behavior) of checking only absolute error.

The idiff utility is extended to take new `-failrelative` and
`-warnrelative` command line arguments.

The hope is that there are various kinds of image comparisons within
tolerances -- like we frequenty use for unit tests, shader
verification, etc. -- where this metric is more helpful than absolute
error.

### DIFF
--- a/src/doc/idiff.rst
+++ b/src/doc/idiff.rst
@@ -142,22 +142,27 @@ Thresholds and comparison options
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. describe:: -fail A
-              -failpercent B
-              -hardfail C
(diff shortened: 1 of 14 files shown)


fpsunflower (approved):
LGTM!

PR #3517, opened 2022-08-11
Fix vector-subscript-out-of-range on MSVC Debug builds

toxieainc:
## Description

Fix vector-subscript-out-of-range on MSVC Debug builds
Also fixes potential problems with >31bit pixelsizes

## Tests

Verified in our own testsuites. Happens on every .hdr file that is being read

## Checklist:

- [X ] I have read the [contribution guidelines](https://github.com/OpenImageIO/oiio/blob/master/CONTRIBUTING.md).
- [X ] If this is more extensive than a small change to existing code, I
- [X ] I have updated the documentation, if applicable.
- [X ] I have ensured that the change is tested somewhere in the testsuite
- [X ] My code follows the prevailing code style of this project.

### DIFF
--- a/src/hdr.imageio/hdrinput.cpp
+++ b/src/hdr.imageio/hdrinput.cpp
@@ -324,7 +324,7 @@ HdrInput::RGBE_ReadPixels_RLE(float* data, int y, uint64_t scanline_width,
         return RGBE_ReadPixels(data, y, scanline_width * num_scanlines);
 
     unsigned char rgbe[4], *ptr, *ptr_end;
-    int i, count;
+    int count;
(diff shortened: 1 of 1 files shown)


lgritz:
Is it easier to change that scanline_width parameter itself to be uint64_t, and then only have to adjust the one place where the function is called to make sure a uint64_t is passed?

toxieainc:
Done

lgritz:
Ah, yeah, much simpler, thanks. I figure any time you find yourself casting the same thing a bunch of times, maybe that means it should just be declared as the other type.

lgritz (approved):

PR #3527, opened 2022-08-18
Remove 'socket' file type

lgritz:
This was envisioned to be useful in the very earliest days of OIIO,
but we never really finished it or used it in any helpful way.
In addition to being essentially dead and untested code, it ties us
to boost.asio, which we use nowhere else.

So just get rid of it this vestigial piece.


lgritz:
Going once... going twice...

PR #3540, opened 2022-09-05
Fixes a warning from an unsuitable format specifier

johannesunterguggenberger-huawei:
## Description

`DASSERT_MSG` is was used with the format specifier `%zd` which is unsuitable for either of the options that `oiio_span_size_type` could evaluate to. This PR should fix it for both cases (`OIIO_SPAN_SIZE_IS_UNSIGNED` is defined and not defined) by using the appropriate format specifier in each case.

## Tests

Compiler warning is now gone.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -2419,7 +2419,13 @@ inline OIIO_HOSTDEVICE float
 interpolate_linear (float x, span_strided<const float> y)
 {
 #ifndef __CUDA_ARCH__
-    DASSERT_MSG (y.size() >= 2, "interpolate_linear needs at least 2 knot values (%zd)", y.size());
+#ifdef OIIO_SPAN_SIZE_IS_UNSIGNED
+    // using oiio_span_size_type = size_t; => %zu is suitable
+    DASSERT_MSG (y.size() >= 2, "interpolate_linear needs at least 2 knot values (%zu)", y.size());
+#else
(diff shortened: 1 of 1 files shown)


lgritz (commented):
Since this assertion is for the case where the size < 2 (so we're not worried about overflow and there are only two possible wrong values), I wonder if the simpler approach would be just to cast to int:

`OIIO_DASSERT_MSG (y.size() >= 2, "interpolate_linear needs at least 2 knot values (%d)", int(y.size()));`

Or even, since this is an assertion for debug builds only that's meant for developers, not a user error message, we could just get rid of the extra value:

`OIIO_DASSERT_MSG (y.size() >= 2, "interpolate_linear needs at least 2 knot values");`

A developer with a debug build who encounters this is not going to have any trouble figuring out whether it was 0 or 1 values that were passed instead of 2 or more.

johannesunterguggenberger-huawei:
Yeah, in this case, a more concise way of writing the assert is probably the best way to go. 

```
OIIO_DASSERT_MSG (y.size() >= 2, "interpolate_linear needs at least 2 knot values (%d)", int(y.size()));
``` 
is probably the best way since it does not lose any information and prevents the compiler warning (which is a problem if warnings are treated as errors).

lgritz:
You can just edit and push the changes onto your same branch, and it will amend this PR, then I can merge it.

johannesunterguggenberger-huawei:
Done!

lgritz (approved):
LGTM, thanks for the fix!

PR #3550, opened 2022-09-12
oiiotool --iccread and --iccwrite

lgritz:
'--iccread' reads a named file and directly adds its contents as the "ICCProfile" metadata of the top image (as a byte array).

'--iccwrite' takes the "ICCProfile" metadata of the top image and writes it to a named file.


lgritz:
@christopherdavies this is the first half of the ICC work

lgritz:
The failure in the "bleeding edge" test seems unrelated to this patch, so I'm going to merge.

PR #3556, opened 2022-09-17
Python internals: support int8[] metadata and ICCProfile, refactor

lgritz:
This patch allows ParamValue, ParamValueList, and ImageSpec to support attribute/getattribute of metadata consisting of uint8[] arrays, which is how we store certain binary blobs or raw byte arrays.  We translate this to Python as a numpy array of type uint8 (dtype='B').

This fixes a limitation where it was not possible to directly retrieve or set the "ICCProfile" metadata (which is stored as a uint8 array) from the Python APIs, as you could easily do from the C++ APIs.

In the process, I found similar bits of code in several different places, in various states of incompleteness. I refactored to make a single make_pyobject() utility and then used that call in place of several other redundant pieces of functionality.


lgritz:
@christopherdavies this is the fix necessary to allow get/set of ICCProfile metadata from python.

src/python/py_oiio.cpp
```diff
@@ -219,6 +219,47 @@ oiio_attribute_typed(const std::string& name, TypeDesc type,
...
+    if (type.basetype == TypeDesc::DOUBLE)
+        return C_to_val_or_tuple((const double*)data, type, nvalues);
+    if (type.basetype == TypeDesc::HALF)
+        return C_to_val_or_tuple((const half*)data, type, nvalues);
+    if (type.basetype == TypeDesc::UINT8 && type.arraylen > 0) {
+        // Array of uint8 bytes
+        // Have to make a copy of the data, because make_numpy_array will
+        // take possession of it.
+        uint8_t* ucdata(new uint8_t[type.arraylen * nvalues]);
+        memcpy(ucdata, data, type.arraylen * nvalues);
```
p12tic, line 251: Maybe `std::memcpy`? Recent C++ standards are starting to diverge from C, e.g. functions from <cmath> got constexpr specifiers.

lgritz: But these are dynamic memory and so can't possibly use a constexpr version here anyway.

I dunno, is it considered bad style to use the "non-std::" way to call a traditional C function that's also in C++? I've always just called the C function when they are identical.

p12tic: Yes, it is considered bad style. Prior to C++23 these headers were actually deprecated. C++23 clarifies the status of the headers as "For compatibility with the C standard library" and also [says](https://github.com/cplusplus/draft/pull/5025/files#diff-f46371f476893d323cfcb1383e9a70f07ff1c1f0c217e22294d0bf937aae6794R5893) that "Source files that are not intended to also be valid ISO C should not use any of the C headers."

If interesting, this [paper](http://www7.open-std.org/JTC1/SC22/WG21/docs/papers/2021/p2340r1.html) contains rationale.

lgritz: Interesting!

Ok, I will change this. I'm sure we have this kind of thing scattered all over the place. We'll clean up a bit here and there as we go.

lgritz: Updated, thanks for the suggestion and the education.

PR #3561, opened 2022-09-20
Add color metadata from libraw

demoulinv:
## Description

Add pre_mul, cam_mul, rgb_cam and cam_xyz arrays availbale in Libraw color info in OpenImageIO metadata.
These metadata are prefixed with "raw:". 2D arrays rgb_cam and cam_xyz are delivered as 1D arrays.
These metadata are fully documented in Libraw: https://www.libraw.org/docs/API-datastruct.html#libraw_colordata_t.
This PR is related to issue https://github.com/OpenImageIO/oiio/issues/3538.

## Checklist:


lgritz:
I think the changes are good, but it's failing tests now because the various reference outputs no longer match the current output. Would you like me to take a stab at fixing that for you?

demoulinv:
Yes, it would be great and much more efficient if you could sort this out for me. Thanks a lot.

lgritz:
I'm going to merge your patch, and then I've got a patch I've tested that I will push right on top of that, which fixes the reference output.

PR #3568, opened 2022-09-27
ustring improvements: ustringHasher and ustring ctr from ustringhash

lgritz:
I realized that the ustringhash class (recently added) is confusingly similr to the hashing functor ustringHash (been there so long I'd forgotten about it). To clear this up, make ustringHasher for clarty. Get on a path of eventually deprecating and removing ustringHash.

I also noticed that there's no directly ustringhash -> ustring conversion. Add it.


lgritz:
Withdrawing, I have a better way to do this.

PR #3571, opened 2022-09-28
Build: enable parallel compilation when using visual studio

aras-p:
## Description

Cmake-generated visual studio project files were using just one CPU core for all the compilation, which is not terribly fast. Add the `/MP` option so that it compiles separate .cpp files in parallel. On my (16 core, 32 thread) machine this makes a Release build of OpenImageIO.lib go from 136 sec down to 30 sec.

## Tests

N/A

## Checklist:

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -188,6 +188,7 @@ endif ()
 if (MSVC)
     # Microsoft specific options
     add_compile_options (/W1)
+    add_compile_options (/MP)
     add_definitions (-D_CRT_SECURE_NO_DEPRECATE)
     add_definitions (-D_CRT_SECURE_NO_WARNINGS)
     add_definitions (-D_CRT_NONSTDC_NO_WARNINGS)


lgritz:
Wow, I had no idea this was necessary. CMake seems pretty good about automatically using parallel builds (at least when CMAKE_BUILD_PARALLEL_LEVEL is set) for make or ninja. I just assumed it would do the right thing for MSVS as well.

Or... are you saying that this is forcing a parallel build because you don't use or don't know about CMAKE_BUILD_PARALLEL_LEVEL? Is there any downside to forcing it and not having a way for users to turn it off?

aras-p:
`CMAKE_BUILD_PARALLEL_LEVEL` seems to only affect the case when "make itself" does the build. When you generate a visual studio project and from the VS IDE, then all the build gets done through visual studio's MSBuild system. And that one needs `/MP` specified in the compiler settings to compile multiple .cpp files in parallel in the same project. It will still try to build multiple _projects_ in parallel by default (when they don't depend on each other), but in OIIO case there isn't all that many to begin with.

I just checked with passing `CMAKE_BUILD_PARALLEL_LEVEL` either as an env. var or a cmake option, and neither affects the visual studio project build throughput.

lgritz:
Oh, I see, that makes sense -- for make/ninja as well, it only affects the "cmake --build" step, not the configure step.

lgritz (approved):
LGTM, thanks!

PR #3572, opened 2022-09-28
Update one more ref image after recent raw change

lgritz:


lgritz:
I updated this slightly -- also eliminating 3 of the reference output samples for the "raw" test that don't appear to be used anymore (corresponding to platform + libraw version combinations that are no longer in our CI matrix).

Issue #3525, opened 2022-08-15
[HELP] Converting a DDS normal map to jpeg


yetigit:
How do I convert this kind of normal map here to jpeg using iconvert or oiiotool ?
[normaltext.zip](https://github.com/OpenImageIO/oiio/files/9337661/normaltext.zip)

expected output:
![Wood_Normal](https://user-images.githubusercontent.com/7016200/184651722-513c6ce8-cfe0-493c-821e-b0045a2a475e.jpg)

lgritz:
What have you tried so far? Did something not work as expected?

lgritz:
There definitely appears to be some problems with OIIO's interpretation and conversion of that file. I'm not sure what's up with that yet. The input seems to be (according to OIIO) a 2-channel file, but maybe that's not correct.

yetigit:
could manage to get my normal converted to png with the standalone nvidia tool https://developer.nvidia.com/nvidia-texture-tools-exporter it has some command line stuff in its install dir :
<img width="1200" alt="explorer_b2uWvNlFjT" src="https://user-images.githubusercontent.com/7016200/184839118-e833243b-4333-432a-924b-4209b3f7852c.png">

they also have nvddsinfo.exe 
so I don't know if it's of any help but it sorta describes the dds texture 

```
C:\Program Files\NVIDIA Corporation\NVIDIA Texture Tools>nvddsinfo.exe "C:\Users\baidh\Downloads\Bistro_v5_2\Textures\Wood_Normal.dds"
Flags: 0x000A1007
        DDSD_CAPS
        DDSD_PIXELFORMAT
        DDSD_WIDTH
        DDSD_HEIGHT
        DDSD_LINEARSIZE
        DDSD_MIPMAPCOUNT
Height: 2048
Width: 2048
Depth: 0
Linear size: 4194304
Mipmap count: 12
Pixel Format:
        Flags: 0x80000004
                DDPF_FOURCC
                DDPF_NORMAL
        FourCC: 'ATI2' (0x32495441)
        Swizzle: 'A2XY' (0x59583241)
        Red mask:   0x00000000
        Green mask: 0x00000000
        Blue mask:  0x00000000
        Alpha mask: 0x00000000
Caps:
        Caps 1: 0x00401008
                DDSCAPS_COMPLEX
                DDSCAPS_TEXTURE
                DDSCAPS_MIPMAP
        Caps 2: 0x00000000
        Caps 3: 0x00000000
[...]

PR #3573, opened 2022-09-28
DDS: various improvements

aras-p:
## Description

- If a DDS pixel format header contains "this is a normal map" flag set, then handle that properly (#3525):
  - For a DXT5, this means it's a "DXT5nm" format. Normal X & Y components are in green and alpha channel respectively. Compute Z component from those, and the result is a 3-channel tangent space normal map.
  - For ATI2/BC5 format that also has a "normal map" flag set, compute the Z component automatically too, and result is a 3-channel image. However a bunch of ATI2/BC5 normal maps out there in the wild do not have that bit set, so also have a global "dds:bc5normal" attribute to optionally turn this on.
- Fix support of R10G10B10A2 format - was producing garbage data due to code assumption that truncation down to 8 bits never needs to happen.
- Fix bit expansion ("too dark") of very low-bit formats, e.g. R3G3B2. Code was simply shifting the bits, instead of properly repeating the high bits into new low bit positions (i.e. a 3-bit "111" needs to become an 8-bit "11111111" and not a "11100000").
- Support "RXGB" (DXT5, with red & alpha swapped) as well as "BC4U" and "BC5U" formats (the latter two are just BC4 and BC5, produced by some older tools).
- Support alpha-only (e.g. A8) formats.

All this mostly prompted by doing DDS improvements in Blender, and wanting to get Cycles renderer (which uses OIIO) to the same feature parity. See https://developer.blender.org/T101405 and https://developer.blender.org/D16087.

## Tests

⚠️ Not yet! I should probably add a bunch of new images to oiio-images test suite, but first wanted to check if these suggested improvements even make sense.

## Checklist:

### DIFF
--- a/src/dds.imageio/ddsinput.cpp
+++ b/src/dds.imageio/ddsinput.cpp
@@ -60,13 +60,11 @@ private:
     std::vector<unsigned char> m_buf;  ///< Buffer the image pixels
     int m_subimage;
     int m_miplevel;
-    int m_nchans;            ///< Number of colour channels in image
-    int m_nfaces;            ///< Number of cube map sides in image
-    int m_Bpp;               ///< Number of bytes per pixel
-    int m_redL, m_redR;      ///< Bit shifts to extract red channel
-    int m_greenL, m_greenR;  ///< Bit shifts to extract green channel
-    int m_blueL, m_blueR;    ///< Bit shifts to extract blue channel
-    int m_alphaL, m_alphaR;  ///< Bit shifts to extract alpha channel
+    int m_nchans;               ///< Number of colour channels in image
+    int m_nfaces;               ///< Number of cube map sides in image
+    int m_Bpp;                  ///< Number of bytes per pixel
+    uint32_t m_BitCounts[4];    ///< Bit counts in r,g,b,a channels
+    uint32_t m_RightShifts[4];  ///< Shifts to extract r,g,b,a channels
     Compression m_compression = Compression::None;
     dds_header m_dds;  ///< DDS header
     dds_header_dx10 m_dx10;
@@ -91,7 +89,8 @@ private:
 
     /// Helper function: calculate bit shifts to properly extract channel data
     ///
-    inline void calc_shifts(int mask, int& left, int& right);
+    inline static void calc_shifts(uint32_t mask, uint32_t& count,
+                                   uint32_t& right);
 
     /// Helper function: performs the actual file seeking.
     ///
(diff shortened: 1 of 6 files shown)


src/dds.imageio/ddsinput.cpp
```diff
@@ -636,7 +717,18 @@ DDSInput::seek_subimage(int subimage, int miplevel)
     return true;
 }
 
-
+static uint32_t
+bit_convert(uint32_t value, uint32_t in_bits, uint32_t out_bits)
```
lgritz, line 721: In fmath.h, we have `bit_range_convert()` (both a templated version for when you know the in and out bit depths, and a function version for when you don't).

Can you look that over and be sure it does the right/same thing as this, and if so, let's use that version. (Or if you don't think the one in fmath does the right thing, please do let me know.)

lgritz (changes requested):
LGTM, other than my comment on how I think we already have an equivalent function to your bit scaling.

src/dds.imageio/ddsinput.cpp
aras-p: Ah cool! Yeah that does exactly what is needed here, updated the code to use it.

lgritz (changes requested):
Can I ask for one more thing? In the documentation file builtinplugins.rst, do you think you could add a brief note to the DDS section explaining what's up with normal maps in this format and advising use of the global "dds:bc5normal" attribute if they wish to change the default behavior of how these files are interpreted?

aras-p:
> Can I ask for one more thing? In the documentation file builtinplugins.rst, do you think you could add a brief note to the DDS section

Done! I also took a liberty of replacing "this is an awful format" sentence with something that says "this is for direclt-to-GPU usage". It's not awful; it's just very, very specific use case. For something you want to stuff directly into a GPU, DDS is an _excellent_ format; but indeed it's not much of use for use cases outside of that.

lgritz (approved):
Perfect, and thanks for the cleanup of the intro. I must have been extra cranky when I wrote that, and it's definitely not the style I'm going for in the docs.

aras-p:
@lgritz by the way would it be possible to get this into 2.4.x release? For Blender they are looking at updating to 2.4 "soon after it's out" and lack of feature parity wrt DDS was one of the things that would be nice to address.

lgritz:
I was definitely going to backport it. Ordinarily, I would have waited until the next tagged release (just had a release yesterday!), but since this is important for Blender, I can do a patch right away if people are waiting for it.

lgritz:
Released as v2.4.4.2

PR #3577, opened 2022-10-03
ustring hashing improvements

lgritz:
I realized that the recently added ustringhash class is confusingly similar to the (obscure, barely used, mostly forgotten) ustringHash functor. But also ustringHash doesn't need to exist, because it's equivalent to `std::hash<ustring>`, if we define it. So now we do define it, and also `std::hash<ustringhash>`, and deprecate ustringHash, to be removed entirely for 3.0.

Also, add a ustring ctr from ustringhash (explicit only for now), and `operator<` for ustringhash.

Finally, modernize a few places we still used `(void)` to just `()`.


AlexMWells (approved):
LGTM.
btw:  I see the ustringhash constructor can't be constexpr because it constructs a ustring which is surely cannot be constexpr.

PR #3581, opened 2022-10-04
DDS: alpha/luminance format fixes, add & enable more tests

aras-p:
## Description

Makes A8, L8, L8A8 formats actually work now. While developing #3573 at one time I had them working properly, but then
while fixing the address sanitizer failure for 3-channel formats I botched them again. Note to self: never again do a "yeah I'll add tests later", since if I had all of them running all the time this would not have happened. 🤦

## Tests

Extend test coverage for more formats & channel size cases. With more test images recently added (https://github.com/OpenImageIO/oiio-images/pull/5), start using them for DDS tests. This covers now:
- Alpha, Luminance and Alpha+Luminance formats,
- RGB formats with rarer channel sizes (rgb10 a2, r3g3b2),
- Several "broken" DDS file cases
- Removed usage of sample-DXT1 since it is well covered by others.

## Checklist:

### DIFF
--- a/src/dds.imageio/ddsinput.cpp
+++ b/src/dds.imageio/ddsinput.cpp
@@ -453,17 +453,29 @@ DDSInput::open(const std::string& name, ImageSpec& newspec)
         m_nchans = GetChannelCount(m_compression,
                                    m_dds.fmt.flags & DDS_PF_NORMAL);
     } else {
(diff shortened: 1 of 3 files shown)


lgritz (approved):
LGTM, thanks

PR #3582, opened 2022-10-04
oiiotool protect against OpenEXR thread deadlock on Windows

lgritz:
Suggested by SideFX: explicitly shut down the OpenEXR thread pool before exiting oiiotool, because sometimes on Windows it seems to hang otherwise. A better solution is probably possible on the OpenEXR side, but I don't know what it would be. (And in any case, we'd still need to do this be be well behaved when using OpenEXR versions earlier than where a fix might land.)

Solution tracked down by Edward Lam.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -19,6 +19,7 @@
 
 #include "oiiotool.h"
 
+#include <OpenEXR/IlmThreadPool.h>
 #include <OpenEXR/ImfTimeCode.h>
 #include <OpenImageIO/Imath.h>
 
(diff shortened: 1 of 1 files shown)


EvanAW:
Upon seeing this, my first thought was "Didn't @debaetsd commit some fixes that touched this about a year are so ago when working on Windows CI fixes?"

There were other fixes (and https://github.com/OpenImageIO/oiio/pull/2038 even earlier), but it turns out https://github.com/OpenImageIO/oiio/issues/2797 and https://github.com/AcademySoftwareFoundation/openexr/issues/890 are still open.

> A better solution is probably possible on the OpenEXR side, but I don't know what it would be. 

Perhaps its time to take a second at the proposed solution [in that OpenEXR thread?](https://github.com/debaetsd/openexr/commit/a2ea041eb92ff20f571d7c9aa4e3787f847e4480)

PR #3584, opened 2022-10-06
DDS: honor ImageInput thread policy

lgritz:


lgritz:
GH only lets you designate somebody as a reviewer if they are an "org member", which I think means a designated committer with write perms. That's crazy, it comes up all the time that you might want a review by the person who reported a problem, or an expert in some area who is not a frequent contributor, etc. I understand why you don't want to allow any repo to request a review from any user at all -- it would be a way for people to spam well-known developers -- but I don't see why it wouldn't be simple to allow anyone who has previously interacted with the project to have their review requested or accepted.

PR #3586, opened 2022-10-06
CI/build: build_libtiff.bash should be sure to build shared library

lgritz:
Something changed in libtiff master and it doesn't build shared libs by default, breaking our "bleeding edge" CI test. So when we build it, be sure to ask for it. It can be overridden in build_libtiff by invoking it with env variable `LIBTIFF_BUILD_SHARED_LIBS=OFF`.


lgritz:
@aras-p FYI, this is why that bleeding edge test was failing. My guess was right, something changed in the last couple days in libtiff trunk.

PR #3587, opened 2022-10-06
Subimage compression

Shootfast:
## Description

This PR enables oiiotool's --compression flag to work on a per subimage basis. As OpenEXR files allow multipart files to use different compression schemes, this seemed to be the best way to enable this functionality.

## Tests

    oiiotool -a -i multipart.exr --compression:subimages=1 zip -o out.exr

## Checklist:

### DIFF
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -1535,10 +1535,22 @@ Writing images
     Sets the compression method, and optionally a quality setting, for the
     output image.  Each ImageOutput plugin will have its own set of methods
     that it supports.
+
+    Optional appended modifiers include:
+
+      `:subimages=` *indices-or-names*
+        Include/exclude subimages (see :ref:`sec-oiiotool-subimage-modifier`).
+        Only included subimages will have the attribute changed. If subimages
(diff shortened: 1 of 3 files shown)


lgritz:
Looks like this is not quite working right, changes a bunch of behavior in CI. I suspect it's this, as you describe in the docs: 

> If subimages are not set, only the first subimage will be changed, or all subimages if the `-a` command line flag was used.

I believe that this might not be quite the fit you want. The `--compression` flag has always sets the default compression for all subsequent outputs, and all subimages. I think that should stay the default behavior.

I'm also not sure how confident you can be that "subimage 1", say, should always be a particular compression for *all* files that get output.

It's worth noting that (I *think*) you could already (as of relatively recently) change a requested compression of a particular subimage of a specific image by using

    --attrib:subimages=1 compression "zip"

and that would apply to just the one image at the stop of the stack.

Shootfast:
Hi Larry,

I guess it just felt as though the --compression flag was the better fit for specifying that information, rather than relying on compression _also_ being a string attribute. 
I can confirm your workaround also works on the latest master branch. We can just abandon this PR then, and I'll do it the other way. Thanks for the tip!

Issue #3585, opened 2022-10-06
[BUG] Regression in HDR read performance with 2.4 release (~10x)


jessey-git:
**Describe the bug**
There seems to be a clear regression in performance when moving to 2.4.x and reading HDR files. A small app, included below, shows a 10second read time in 2.4 vs. a 1second read in 2.3

**To Reproduce**
1. Create an ImageInput from a .hdr file
2. Measure how long it takes to read the image
3. Observe it takes a lot longer than it used to

**Expected behavior**
Performance closer to the 1second case should hopefully be possible. Certainly not the 10 seconds.

**Evidence**
See below for the sample program as well as where to download a free .hdr if you don't have one handy.

It will output a simple timing:
2.3
```
Read time: 992 ms
Read time: 994 ms
```

2.4
```
Read time: 10057 ms
Read time: 10383 ms
```

A quick profile just says that seeking and freading are the main bottlenecks. I'm guessing these are called more frequently/less efficiently since the big HDR refactor last year?
![first-pass](https://user-images.githubusercontent.com/7989986/194224262-cf9402b7-c68b-4edc-93de-7492758a2cc2.png)

**Platform information:**
 - OIIO branch/version: 2.4.4.2
 - OS: Windows 10
 - C++ compiler: VS 2019 and VS 2022 compiled in the Release configuration

----
Use the following free HDR: https://polyhaven.com/a/autoshop_01 (use the 8k HDR variant which should be 97mb)

```
#include <cstdio>
#include <memory>
#include <chrono>
#include <OpenImageIO/imageio.h>

int main()
{
    const char* filepath = "t:/autoshop_01_8k.hdr";
    std::unique_ptr<OIIO::ImageInput> in = OIIO::ImageInput::create(filepath);

    OIIO::ImageSpec spec;
    OIIO::ImageSpec config;
    config.attribute("oiio:UnassociatedAlpha", 1);

    if (!in->open(filepath, spec, config)) {
        printf("Failed to open\n");
        return -1;
    }

    int width = spec.width;
    int height = spec.height;
    int channels = spec.nchannels;

    int scanlinesize = width * channels * sizeof(unsigned char);
    unsigned char* pixels = (unsigned char*)malloc(width * height * 4);

    auto start = std::chrono::high_resolution_clock::now();
[...]

PR #3588, opened 2022-10-07
HDR: fix read performance regression (#3585)

aras-p:
## Description

HDR code overhaul to support IOProxy (#3218) changed all reads into a `pread()` with explicit file position tracking. Turns out, that is quite a larger overhead compared to simple sequential reads; even more so on Windows.

On my PC (Windows 10, VS2022, Ryzen 5950X) this change gets file read time for an 8x resolution .HDR image (with RLE compression) from 8.28s back to 1.10s just like it was in OIIO 2.3.

On Windows the extra cost when using pread() is a bit more extra mutex locks, but the real cost is in enventual ReadFile; looks like any seek operations make it take some sort of "way slower" path with various callbacks and whatnot inside the kernel.

Before the fix (the extra ~2 sec of file seeking cost are outside the shot):
read_cost_before

After the fix:
read_cost_after

(exercise for the future: get rid of that `ldexp` cost, and perhaps also stop reading the file mostly two bytes at a time -- but both are optimizations in a "nice to have" category; not this performance regression)

## Tests

No new tests needed.

## Checklist:


lgritz:
This is awesome, thanks!

Now for some discussion (which I think should not preclude accepting this for now).

Switching from IOProxy::pread to using IOProxy::seek + IOProxy::read makes the IOProxy stateful and not thread-safe without an external lock (which this HdrInput has, so it's not going to break anything), precluding any future removal of the locks from read_native_scanline to make the whole ImageInput able to read concurrently. That's probably not important for this file format, but for others (especially crucial those usable as texture), it would be desirable to rely on those stateless pread calls and remove the lock from the ImageInput.

I think this patch is fine. We don't expect a lot of concurrent reads from hdr like we do for tiff and openexr that we use extensively for highly threaded on-demand reading of texture. It fixes the recent performance regression, and it doesn't really cost us concurrency because we already had the lock in HdrInput. But in general, we want to prefer eschewing read/seek and using pread instead, so we can expect maximum concurrency from multiple threads using an ImageInput (as we do in ImageCache/TextureSystem).

So I believe that a revised and more complete diagnosis is that the problem isn't the pread vs read per se, but that for RLE-compressed images we were calling pread separately for every 4 bytes! The calls to read are faster simply because they are buffered (in FILE) and require fewer OS system calls. But the real sin is that we call it way too many times.

I think that the rle case could go back to pread if we want, but do just one pread per scanline (rather than one per pixel value run), asking for the maximum amount that will be needed to read the scanline for the worst case rle representation (it's ok if it's the end of the file, pread will just read up to the end and return the true number of bytes read). In other words, we'd be doing the buffering ourselves.

But anyway, the real misdesign of this HDR reader is ancient, and consists of doing many many tiny reads. Even the buffered fread would be sped up a lot by not doing it separately for every 2-4 bytes of the file. I don't know if it's worth fixing for this format. Depends on whether anybody really needs to maximize its performance or expects multiple threads reading the same hdr file to be fully concurrent.

aras-p:
Yeah I agree -- the actual cause of perf regression is like "pread has some extra overhead compared to just read", which is not an issue if you're reading in "sensible" chunk sizes. But does very much become an issue if you're reading 2 or 4 bytes at a time :) Maybe someday Someone :tm: will redo the reading code here to read in larger chunks, and it could get back to pread.

Issue #3295, opened 2022-01-23
[BUG] oiiotool crash/undefined behavior at exit


ThiagoIze:
**Describe the bug**
In Windows, oiiotool returns 0 from main, but the process returns 3 to the shell. This is because the c runtime abort function is terminating the process due to an error and it returns 3.

In Linux a few random times I got segfaults at exit. Valgrind reports errors of the form:
```
==56783== Command: oiiotool --info testrender.tif
==56783== 
testrender.tif       :  320 x  240, 4 channel, uint8 tiff
==56783== Invalid read of size 8
==56783==    at 0x78610E: OpenImageIO_v2_4_0::ImageBufImpl::clear() (src/libOpenImageIO/imagebuf.cpp:703)
==56783==    by 0x785D95: OpenImageIO_v2_4_0::ImageBufImpl::~ImageBufImpl() (src/libOpenImageIO/imagebuf.cpp:485)
==56783==    by 0x78357D: OpenImageIO_v2_4_0::ImageBuf::impl_deleter(OpenImageIO_v2_4_0::ImageBufImpl*) (src/libOpenImageIO/imagebuf.cpp:333)
==56783==    by 0x786F6E: ~unique_ptr (unique_ptr.h:292)
==56783==    by 0x786F6E: OpenImageIO_v2_4_0::ImageBuf::~ImageBuf() (src/libOpenImageIO/imagebuf.cpp:569)
==56783==    by 0x4288D1: std::_Sp_counted_ptr<OpenImageIO_v2_4_0::ImageBuf*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (shared_ptr_base.h:377)
==56783==    by 0x49CD2B: _M_release (shared_ptr_base.h:155)
==56783==    by 0x49CD2B: ~__shared_count (shared_ptr_base.h:730)
==56783==    by 0x49CD2B: ~__shared_ptr (shared_ptr_base.h:1169)
==56783==    by 0x49CD2B: _Destroy<std::shared_ptr<OpenImageIO_v2_4_0::ImageBuf> > (stl_construct.h:98)
==56783==    by 0x49CD2B: __destroy<std::shared_ptr<OpenImageIO_v2_4_0::ImageBuf> *> (stl_construct.h:108)
==56783==    by 0x49CD2B: _Destroy<std::shared_ptr<OpenImageIO_v2_4_0::ImageBuf> *> (stl_construct.h:136)
[...]

PR #3591, opened 2022-10-07
oiiotool: work around static destruction order issue (#3295)

aras-p:
## Description

In some cases (at least on Windows, with static linking) the oiiotool main returns zero, but the full process returns 3 error code due to abort() being called. And that is called because a global shared image cache is already destroyed, but the image references from another global (Oiiotool itself) try to use it, get a call into a "pure virtual function" stub that is put by MSVC C runtime into a destroyed object vtable, and that ends up in an abort().

Work around that by making sure various member variables holding ImageRecRef are cleared before main() finishes and global destructors are called.

## Tests

N/A -- but it does make running OIIO test suite locally much easier for me. Previously python runners that were launching oiiotool were getting exit code 3 when "everything was fine", so I had to resort to running their final command lines manually. Which is quite cumbersome!

## Checklist:

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -7018,6 +7018,14 @@ main(int argc, char* argv[])
         std::cout << "\n" << ot.imagecache->getstats(2) << "\n";
     }
 
+    // Release references of images that might hold onto a shared
+    // image cache. Otherwise they would get released at static destruction
+    // time, at which point due to undefined destruction order the shared
+    // cache might be already gone.
+    ot.curimg = nullptr;
+    ot.image_stack.clear();
+    ot.image_labels.clear();
+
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM. Definitely seems reasonable.

lgritz:
Hmmm, this also makes me think that it probably would have been better for ImageCache::create to have returned a ref-counted pointer. And perhaps for "ot" to be dynamically allocated rather than static, so we can explicitly destroy it before oiiotool exits.

lgritz:
Does this patch fully address #3295 ?

aras-p:
> Does this patch fully address #3295 ?

It does for me, but I dunno if it does for the original bug reporter. My guess would be on "yes", from their posted callstack.

PR #3592, opened 2022-10-07
BMP: fix reading 16bpp images (#2808)

### LINKED ISSUE #2808: [BUG] 565 bmp doesn't import correctly
**Describe the bug**
OIIO fails to import BMPs correctly when the bits per sample is less than 8. For example this attached image I think is 565 (zipped as github won't let me upload a bmp).

[bmp.zip](https://github.com/OpenImageIO/oiio/files/5764477/bmp.zip)

iinfo states that oiio:BitsPerSample is 4. The code [here](https://github.com/OpenImageIO/oiio/blob/8ac0f966e611120148abb9a5bfcfcaeff915af3b/src/bmp.imageio/bmpinput.cpp#L167) seems to mask for 555 but then shift as if it's 444. I notice that the DibInformationHeader class has red/green/blue_mask members, but they don't seem to be used.

Even assuming the correct masking and shifting was being done, the data should probably be rescaled to 8 bit anyway.

Loading the image in the zip I get this:

bad

wheras I expect it to look like this:

good

**Platform information:**
 - OIIO branch/version: 2.2.9
 - OS: windows 10
 - C++ compiler: msvc

aras-p:
## Description

16 bit per pixel (e.g. 555, 565, 444) formats probably have never worked. One part of the code was assuming 5 bits in the mask, another was shifting by 4 bits, and the 16-bit pixel fetch code was reading just one byte. Fixes #2808.

## Tests

The test suited added back in #2976 already had a `TODO: seems broken: g16bf555.bmp, g16bf565.bmp, g16def555.bmp` comment, which is exactly 16 bpp test cases. So I've uncommented them here, and made them work.

Also removed one `g32bf.bmp` from the tests lists; it was listed there twice (looked more like a copy-paste accident than on purpose).

## Checklist:

### DIFF
--- a/src/bmp.imageio/bmp_pvt.cpp
+++ b/src/bmp.imageio/bmp_pvt.cpp
@@ -101,7 +101,8 @@ DibInformationHeader::read_header(Filesystem::IOProxy* fd)
             return false;
         }
 
-        if (size == WINDOWS_V4 || size == WINDOWS_V5 || size == UNDOCHEADER52
+        if ((size == WINDOWS_V3 && bpp == 16 && compression == 3)
+            || size == WINDOWS_V4 || size == WINDOWS_V5 || size == UNDOCHEADER52
             || size == UNDOCHEADER56) {
             if (!fread(fd, &red_mask) || !fread(fd, &green_mask)
                 || !fread(fd, &blue_mask)) {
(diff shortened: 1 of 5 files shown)


lgritz (approved):
LGTM

PR #3597, opened 2022-10-11
Windows: stop including <windows.h> from public OIIO headers

aras-p:
## Description

Including windows.h leads to a lot of possibly common names being "taken", mostly by preprocessor defines. Some of these are possible to turn off (like "max" and "min" by defining "NOMINMAX" before including <windows.h>). But many others are not possible to turn off, e.g. "near" and "far" just get defined to nothing, preventing from using variables named like that.

Additionally, includes have a cost, and while <windows.h>, being a C header does not have expensive language constructs that would slow down compilation a lot, it's still something close to 100k lines of preprocessed code to wade through.

Now, only the OIIO implementation .cpp files that actually need <windows.h> include it (all of them under "libutil").

There were two exceptions:
1) thread.h `pause()` implementation uses `YieldProcessor()` macro when on
  a Windows system. Replaced it with manually "re-implemented"
  things that YieldProcessor expands to. Right now made the
  implementation for all 4 platforms that Windows normally supports
  (x64, x86, arm, arm64); the latter two OIIO does not support though
  when on Windows so it's untested.
2) timer.h `Timer::now()` inline function used `QueryPerformanceCounter`
  and `LARGE_INTEGER`. These are quite hard to somehow "forward-declare"
  or similar, without resulting in duplicate type definitions. So on
  Windows I moved the now() function to be non-inline. A though for another
  day: maybe all of that could be replaced by C++ `<chrono>` instead.

## Tests

Existing CI tests.

## Checklist:

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -20,7 +20,11 @@
 #include <OpenImageIO/ustring.h>
 
 #ifdef _WIN32
-// # include <windows.h>   // Already done by platform.h
+#    define WIN32_LEAN_AND_MEAN
+#    define VC_EXTRALEAN
+#    define NOMINMAX
+#    include <windows.h>
+
 #    include <direct.h>
 #    include <io.h>
 #    include <shellapi.h>
--- a/src/include/OpenImageIO/platform.h
+++ b/src/include/OpenImageIO/platform.h
@@ -38,22 +38,6 @@
 #    include <malloc.h>  // for alloca
 #endif
 
-#if defined(_WIN32)
-#    ifndef WIN32_LEAN_AND_MEAN
-#        define WIN32_LEAN_AND_MEAN
-#    endif
-#    ifndef VC_EXTRALEAN
-#        define VC_EXTRALEAN
-#    endif
-#    ifndef NOMINMAX
-#        define NOMINMAX
-#    endif
-#    ifndef NOGDI
-#        define NOGDI
-#    endif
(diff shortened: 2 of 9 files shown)


lgritz:
Yeah, the timer.h predates being able to depend on C++11 chrono. We'll want to benchmark to be sure that on all the platforms, chono is not higher overhead than when we're currently doing, but if it's good, I would have no objection to a future timer.h cleanup that bases it on chrono wherever possible.

src/libutil/filesystem.cpp
```diff
@@ -20,7 +20,12 @@
 #include <OpenImageIO/ustring.h>
 
 #ifdef _WIN32
-// # include <windows.h>   // Already done by platform.h
+#    define WIN32_LEAN_AND_MEAN
+#    define VC_EXTRALEAN
+#    define NOMINMAX
+// clang-format off
+#    include <windows.h>
+// clang-format on
```
lgritz, line 28:
```suggestion
#    include <windows.h>

```

I don't think you need the clang-format on/off, a blank line after the windows.h include will prevent it from being sorted with the subsequent lines.

lgritz (approved):
LGTM, I like getting windows.h out of the public headers.

src/libutil/filesystem.cpp
aras-p: Ah cool! Did not know about this trick. Amended the commit with the change.

lgritz (approved):

PR #3606, opened 2022-10-16
Ensure that ustring hashes are 64 bits

lgritz:
ustring::hash() and the representation of ustringhash were size_t, which might be 32 bits on some platforms. But we really would prefer ustrings hash to 64 bits, to greatly reduce the chance of hash collisions.

* Add strhash64() which is like strhash() but guaranteed to be a 64 bit value (returning uint64_t instead of size_t -- though note that these are the same on 64 bit CPUs).

* ustring::hash() and ustringhash::rep_t are now uint64_t.

* Note that std::hash(ustring) and std::hash(ustringhash) are still size_t, per the C++ definition of std::hash. So ustring::hash() may have more bits than std::hash(ustring) on 32 bit platforms.

This will NOT be backported to the release branch, since it changes ABI on 32 bit platforms.

### DIFF
--- a/src/include/OpenImageIO/ustring.h
+++ b/src/include/OpenImageIO/ustring.h
@@ -126,6 +126,7 @@ class OIIO_UTIL_API ustring {
 public:
     using rep_t      = const char*;  ///< The underlying representation type
     using value_type = char;
+    using hash_t     = uint64_t;  ///< The hash type
     using pointer    = value_type*;
     using reference  = value_type&;
     using const_reference        = const value_type&;
@@ -311,7 +312,7 @@ public:
     }
 
     /// Return a hashed version of the string
-    size_t hash() const noexcept
+    hash_t hash() const noexcept
     {
         if (!m_chars)
             return 0;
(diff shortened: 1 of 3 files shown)


fpsunflower (approved):
LGTM!

src/include/OpenImageIO/ustring.h
```diff
@@ -983,8 +985,8 @@ class OIIO_UTIL_API ustringhash {
 
 
 
-static_assert(sizeof(ustringhash) == sizeof(size_t),
-              "ustringhash should be the same size as a size_t");
+static_assert(sizeof(ustringhash) == sizeof(uint64_t),
```
AlexMWells, line 988: Should this compare against sizeof(hash_t)?

src/libutil/ustring.cpp
```diff
@@ -69,7 +69,7 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
     }
 #endif
 
-    const char* lookup(string_view str, size_t hash)
+    const char* lookup(string_view str, uint64_t hash)
```
AlexMWells, line 72: Should this (and others in ustring.cpp be hash_t instead of uint64_t?

PR #3640, opened 2022-10-31
CMake: also find jconfig.h from libturbojpeg from arch-specific include directory

illwieckz:
## Description

This header file is architecture-dependent and can be stored as:

- `/usr/include/i386-linux-gnu/jconfig.h`
- `/usr/include/x86_64-linux-gnu/jconfig.h`

Instead of:

- `/usr/include/jconfig.h`

That fixes build errors like this:

```
CMake Error at src/cmake/modules/FindJPEGTurbo.cmake:32 (file):
  file STRINGS file "/usr/include/jconfig.h" cannot be read.
  src/cmake/checked_find_package.cmake:127 (find_package)
  src/cmake/externalpackages.cmake:139 (checked_find_package)
  CMakeLists.txt:155 (include)
```
## Checklist:

- [x] ~~If this is more extensive than a small change to existing code, I
  [corporate](https://github.com/OpenImageIO/oiio/blob/master/src/doc/CLA-CORPORATE)).~~   
  This change is too small and not original enough to be subject to copyright.
- [x] ~~I have updated the documentation, if applicable.~~  
  Not applicable.
- [x] ~~I have ensured that the change is tested somewhere in the testsuite
  (adding new test cases if necessary).~~  
  Not applicable.

### DIFF
--- a/src/cmake/modules/FindJPEGTurbo.cmake
+++ b/src/cmake/modules/FindJPEGTurbo.cmake
@@ -29,7 +29,22 @@ endif ()
 if (JPEG_INCLUDE_DIR AND JPEG_LIBRARY)
   set (JPEG_INCLUDE_DIRS ${JPEG_INCLUDE_DIR} CACHE PATH "JPEG include dirs")
   set (JPEG_LIBRARIES ${JPEG_LIBRARY} CACHE STRING "JPEG libraries")
-  file(STRINGS "${JPEG_INCLUDE_DIR}/jconfig.h"
+
+  find_path(jconfig_header_file jconfig.h PATHS "${JPEG_INCLUDE_DIR}")
+
+  if (NOT jconfig_header_file)
+    file(GLOB_RECURSE jconfig_header_file_list "${JPEG_INCLUDE_DIR}/*/jconfig.h")
+
+    if (jconfig_header_file_list)
+      list(GET jconfig_header_file_list 0 jconfig_header_file)
+    endif()
+
+    if (NOT jconfig_header_file)
+      message(WARNING "Cannot find jconfig.h from libturbojpeg")
+    endif()
+  endif()
+
+  file(STRINGS "${jconfig_header_file}"
                 jpeg_lib_version REGEX "^#define[\t ]+JPEG_LIB_VERSION[\t ]+.*")
   if (jpeg_lib_version)
       string(REGEX REPLACE "^#define[\t ]+JPEG_LIB_VERSION[\t ]+([0-9]+).*"


src/cmake/modules/FindJPEGTurbo.cmake
```diff
@@ -29,7 +29,22 @@ endif ()
...
+
+  if (NOT jconfig_header_file)
+    file(GLOB_RECURSE jconfig_header_file_list "${JPEG_INCLUDE_DIR}/*/jconfig.h")
+
+    if (jconfig_header_file_list)
+      list(GET jconfig_header_file_list 0 jconfig_header_file)
+    endif()
+
+    if (NOT jconfig_header_file)
+      message(FATAL_ERROR "Cannot find jconfig.h from libturbojpeg")
```
lgritz, line 43: I'm not so sure about the FATAL_ERROR here -- if this header can't be found, I think we want only the find_package(JPEGTurbo) to not succeed, and then to fall back on regular find_package(JPEG). We don't want the entire cmake build to fail.

lgritz: @illwieckz Any comment or further action on this? In general, I think it's fine except for the FATAL_ERROR part.

illwieckz: Sorry, I was busy and totally forgot about it. I'll look at it again.

lgritz: @illwieckz Ping -- can you please take a look here at my previous comment and offer an opinion?

illwieckz: Changed to a `WARNING`.

src/cmake/modules/FindJPEGTurbo.cmake
```diff
@@ -29,7 +29,22 @@ endif ()
...
+      list(GET jconfig_header_file_list 0 jconfig_header_file)
+    endif()
+
+    if (NOT jconfig_header_file)
+      message(WARNING "Cannot find jconfig.h from libturbojpeg")
+    endif()
+  endif()
+
+  file(STRINGS "${jconfig_header_file}"
                 jpeg_lib_version REGEX "^#define[\t ]+JPEG_LIB_VERSION[\t ]+.*")
```
lgritz, line 48: This line `file(STRINGS ...)` needs to be inside an `if (jconfig_header_file)` conditional. If the header is not found, trying to extract strings from it results in a build-breaking cmake error, as you can see from the windows CI job.

lgritz (changes requested):

domin144:
Starting with version 2.1.0 libjpeg-turbo provides it's own cmake config files, but they are named different to what is used in oiio cmake files.
I think the line:
```
checked_find_package (JPEGTurbo
```
Should be changed to:
```
checked_find_package (libjpeg-turbo
```
so that distro provided library is detected if present.

lgritz:
That's interesting, @domin144. I'm tempted to advise abandoning this PR entirely and instead simply removing FindJPEGTurbo.cmake entirely, relying on jpeg-turbo's exported config, and calling it a day. That will mean that we are raising the floor of JPEG-turbo that we support to 2.1. But since that's still 2.5 years old and this is not a required dependency (we can automatically fall back to libjpeg), I think that would be acceptable.

Opinions?

lgritz:
I tried the other strategy here: https://github.com/OpenImageIO/oiio/pull/3987

What do people think of that proposal, which would obviate the need for this one here?

PR #3650, opened 2022-11-03
ociodisplay gets an inverse control

### LINKED ISSUE #3648: [FEATURE REQUEST] Add an inverse modifier to oiiotool's --ociodisplay option
**Problem Description**

The oiiotool --ociodisplay option is very useful for applying an OCIO DisplayViewTransform.  This takes an image in a "src" color space and then applies a "view" transform for a specific "display".  For example, one might take a source image in ACEScg space and apply the "ACES 1.0 - SDR Video" view for the "sRGB - Display", using either of the following:
 
```
% oiiotool acescg.exr --ociodisplay:from="ACEScg" "sRGB - Display" "ACES 1.0 - SDR Video" -o srgb.jpg
% oiiotool acescg.exr --iscolorspace "ACEScg" --ociodisplay "sRGB - Display" "ACES 1.0 - SDR Video" -o srgb.jpg
```
However, it is often useful to be able to apply the DisplayViewTransform in the inverse direction.  For example, to convert sRGB back to ACEScg, inverting a specific view transform.  The ocioconvert tool (one of the OCIO command-line tools) is already able to do this, as follows:

```
% ocioconvert --invertview srgb.exr "sRGB - Display" "ACES 1.0 - SDR Video" acescg.exr "ACEScg"
```
Inverting a DisplayViewTransform is easily possible via the OCIO v2 API.  Simply set the direction of the DisplayViewTransform to TRANSFORM_DIR_INVERSE.  

**Proposed solution**

The --ociofiletransform option in oiiotool already has an "inverse" modifier.  So the proposed solution is to add an "inverse" modifier to --ociodisplay.

lgritz:
Fixes #3648

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -1252,12 +1252,14 @@ ColorConfig::createLookTransform(ustring looks, ustring inputColorSpace,
 ColorProcessorHandle
 ColorConfig::createDisplayTransform(string_view display, string_view view,
                                     string_view inputColorSpace,
-                                    string_view looks, string_view context_key,
+                                    string_view looks, bool inverse,
+                                    string_view context_key,
                                     string_view context_value) const
 {
     return createDisplayTransform(ustring(display), ustring(view),
                                   ustring(inputColorSpace), ustring(looks),
-                                  ustring(context_key), ustring(context_value));
+                                  inverse, ustring(context_key),
+                                  ustring(context_value));
 }
 
 
@@ -1265,7 +1267,7 @@ ColorConfig::createDisplayTransform(string_view display, string_view view,
 ColorProcessorHandle
 ColorConfig::createDisplayTransform(ustring display, ustring view,
                                     ustring inputColorSpace, ustring looks,
-                                    ustring context_key,
+                                    bool inverse, ustring context_key,
                                     ustring context_value) const
 {
     if (display.empty())
@@ -1275,7 +1277,8 @@ ColorConfig::createDisplayTransform(ustring display, ustring view,
     // First, look up the requested processor in the cache. If it already
     // exists, just return it.
     ColorProcCacheKey prockey(inputColorSpace, ustring() /*outputColorSpace*/,
-                              context_key, context_value, looks, display, view);
+                              context_key, context_value, looks, display, view,
+                              ustring() /*file*/, inverse);
     ColorProcessorHandle handle = getImpl()->findproc(prockey);
     if (handle)
         return handle;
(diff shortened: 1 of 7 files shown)


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -1848,7 +1876,7 @@ bool
 ImageBufAlgo::ociodisplay(ImageBuf& dst, const ImageBuf& src,
                           string_view display, string_view view,
                           string_view from, string_view looks, bool unpremult,
-                          string_view key, string_view value,
+                          bool inverse, string_view key, string_view value,
                           ColorConfig* colorconfig, ROI roi, int nthreads)
 {
```
doug-walker, line 1881: Don't know how to select the line 1872, but don't you want to pass inverse there too?

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -1889,19 +1917,47 @@ ImageBufAlgo::ociodisplay(ImageBuf& dst, const ImageBuf& src,
...
-                          bool unpremult, string_view key, string_view value,
-                          ColorConfig* colorconfig, ROI roi, int nthreads)
+                          bool unpremult, bool inverse, string_view key,
+                          string_view value, ColorConfig* colorconfig, ROI roi,
+                          int nthreads)
 {
     ImageBuf result;
     bool ok = ociodisplay(result, src, display, view, from, looks, unpremult,
-                          key, value, colorconfig, roi, nthreads);
+                          unpremult, key, value, colorconfig, roi, nthreads);
```
doug-walker, line 1926: unpremult --> inverse

doug-walker (changes requested):
Thanks for adding this Larry!

src/libOpenImageIO/color_ocio.cpp
lgritz: yes, I think you're right

src/libOpenImageIO/color_ocio.cpp
lgritz: ouch, yes

lgritz:
Updated the PRs to address @doug-walker's suggestions.

doug-walker (approved):
LGTM!

lgritz:
Because this isn't *tested*, I'm going to hold off on a merge while I do a little tinkering...

I'm exploring more direct support for OpenColorIO 2.2 and its built-in configs, which will allow me to test display transforms (merely because it has some, unlike the sample config that's in the testsuite now).

If minimal support of this is easy to achieve, I'll add it to this PR (even if the full range of tests only work when we're building against OCIO 2.2). But if that looks like a bigger job that doesn't make sense to append here, I'll merge this PR first (despite not being tested) and the follow up with a separate one with 2.2 support and more testing.

lgritz:
I'm going to merge this now, because the follow-on has grown to be more complex.

PR #3668, opened 2022-11-14
PSD: Fix thumbnail extraction for PSD files

### LINKED ISSUE #3660: [BUG] PSD Thumbnail is black
```

    ImageSpec config;
    config["oiio:UnassociatedAlpha"] = 1;
    bool success                     = false;
    auto inp                         = ImageInput::open(filePath.toStdWString(), &config);
    if (inp)
    {
        const ImageSpec &spec = inp->spec();

        ImageBuf thumb;

        inp->get_thumbnail(thumb, 0);
        if (thumb.initialized())
        {
            // data is all zero
      }
   ...

```

https://www.dropbox.com/s/g5pde4djl2nfcuk/_DSC0725.zip

lgritz:
Fixes #3660

In investigating this issue, I also noticed that now that IOProxy support can be relied on, the old code that in the psd reader that directly fumbles with libjpeg is not needed and can be replaced by just a couple of lines to make an ImageBuf read from an IOProxy.

And it fixes the bug, too. So I'm not even going to try to figure out what was going wrong with the old code.


lgritz:
The sanitizer failure is addressed separately by #3666, discovered along the way. But they were technically unrelated, so I made two PRs. When combined, everything works.

PR #3669, opened 2022-11-14
[raw] update exif orientation with raw flip if user flip is set

servantftechnicolor:
If raw:user_flip is set, the flip field of the raw input is replaced with the value of raw:user_flip and the original flip value is lost.

If raw:user_flip is set to 1, we can consider the user wanted to disable the buffer rotation.

We may want to display it correctly anyway. This PR proposes to store the original flip value inside the exif orientation code.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -947,8 +947,7 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // If user flip is set to 1, it means we ignore the flip
     // Let's set the orientation exif flags to the original flip
     // value so that it is still displayed correctly
-    if (config.get_int_attribute("raw:user_flip", -1) == 1)
-    {
+    if (config.get_int_attribute("raw:user_flip", -1) == 1) {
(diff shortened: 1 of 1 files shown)


fabiencastan:
Fix https://github.com/OpenImageIO/oiio/issues/3574

lgritz (changes requested):
The changes LGTM. 

Please look at the results of the clang-format CI test and adjust formatting as it advises.

Don't worry about the other failed test -- it seems to have failed while setting up dependencies, and is unrelated to your changes. Probably a momentary error in availability of a server that is needed to download deps.

servantftechnicolor:
The Clang check fail but i don't see anything to do in the log ...

lgritz:
If you click on the "details" button for that test, at the bottom of the log in the build section, you'll see a diff of the clang-formatted version and what you submitted.

Looks like there is one line with some extra whitespace. Don't worry about it, I'll merge this and then fix it separately. I don't want to make more work for you than is necessary, I really appreciate your submitting this fix.

fabiencastan:
Thanks

PR #3703, opened 2022-12-09
bump to 3.1.5 for vfxreferenceplatform cy2022

mcoliver:
## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/OpenImageIO/oiio/blob/master/CONTRIBUTING.md).


lgritz (approved):
LGTM.

The way we use these `build_<package>.bash` scripts in CI, we always explicitly specify the `<PACKAGE>_VERSION` env variable. So it has probably been a long time since we went through them and updated the default versions to something modern for the sake of people who may run these scripts by hand to help set up dependencies.

Thanks for the fix.

PR #3722, opened 2022-12-22
ARM: Fix signed/unsigned simd mismatch in vbool4::load

### LINKED ISSUE #3721: [BUG] 2.4.6.1 build fails on ARM 64 (aarch64)
**Describe the bug**
Looks like a type conversion error specific to aarch64.

**Evidence**
There are many errors but here's one:
```
In file included from /builddir/build/BUILD/oiio-2.4.6.1/src/include/OpenImageIO/fmath.h:44,
                 from /builddir/build/BUILD/oiio-2.4.6.1/src/libutil/fmath.cpp:11:
/builddir/build/BUILD/oiio-2.4.6.1/src/include/OpenImageIO/simd.h: In member function 'void OpenImageIO_v2_4::simd::vbool4::load(bool, bool, bool, bool)':
/builddir/build/BUILD/oiio-2.4.6.1/src/include/OpenImageIO/simd.h:3264:31: note: use '-flax-vector-conversions' to permit conversions between vectors with differing element types or numbers of subparts
 3264 |     m_simd = vld1q_s32 (values);
      |                               ^
/builddir/build/BUILD/oiio-2.4.6.1/src/include/OpenImageIO/simd.h:3264:24: error: cannot convert 'int32x4_t' to 'OpenImageIO_v2_4::simd::vbool4::simd_t' {aka 'uint32x4_t'} in assignment
 3264 |     m_simd = vld1q_s32 (values);
      |              ~~~~~~~~~~^~~~~~~~
      |                        |
      |                        int32x4_t
```

Full build log here:
https://kojipkgs.fedoraproject.org//work/tasks/3501/95613501/build.log

**Platform information:**
 - OIIO branch/version: 2.4.6.1
 - OS: Fedora Rawhide
 - C++ compiler: gcc 12.2.1
 - Any non-default build flags when you build OIIO: (in build log)

lgritz:
Fixes #3721

Please read the comments in #3721. This is the "local" fix for the build break due to the type mismatch. It is a band-aid. And it may be the best solution for the 2.4 branch if we don't want to potentially break ABIs by changing the definition of any public types.

Still pending is to examine the issue of whether it was a mistake to define vbool4 storage for neon as uint32x4_t or if we should change it to int32x4_t to better match the non-simd reference implementation. After debating that (and identifying somebody with access to an ARM-based machine to test the solution for us), we may return to tackle this more fundamental change.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -3261,7 +3261,9 @@ OIIO_FORCEINLINE void vbool4::load (bool a, bool b, bool c, bool d) {
     m_simd = _mm_castsi128_ps(_mm_set_epi32(-int(d), -int(c), -int(b), -int(a)));
 #elif OIIO_SIMD_NEON
     int values[4] = { -int(a), -int(b), -int(c), -int(d) };
-    m_simd = vld1q_s32 (values);
+    m_simd = vld1q_u32((const uint32_t*)values);
+    // this if we were using int:
+    // m_simd = vld1q_s32(values);
 #else
     m_val[0] = -int(a);
     m_val[1] = -int(b);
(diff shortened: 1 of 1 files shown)


hobbes1069:
Looks like that got it:
https://koji.fedoraproject.org/koji/taskinfo?taskID=95618167

PR #3735, opened 2023-01-03
Move ARM_NEON definition and fix possible double define

lgritz:
It should not have been set in platform.h (inside the cpuid function!) in the first place. And apparently, though this was ok for Windows ARM, on Mac ARM it's an error because that symbol is already defined. So only redefine it if it's currently undefined. And move the whole check and definition into simd.h where all the other ARM_NEON stuff lives.


lgritz:
@mikaelsundell and @Gerdya does this look ok to both of you?

mikaelsundell:
Thanks for that, works on both my arm64 and x86_64 Mac builds.

Excellent support,

Mikael

Mikael Sundell
***@***.***

PR #3736, opened 2023-01-03
Add EXIF write support to PNG output.

joris-nijs:
## Description

Stores any EXIF data from the given ImageSpec into the PNG eXIf chunk.

## Checklist:

- [x] ~~If this is more extensive than a small change to existing code, I
  [corporate](https://github.com/OpenImageIO/oiio/blob/master/src/doc/CLA-CORPORATE)).~~ Change is small enough to not require a CLA
- [x] ~~I have updated the documentation, if applicable.~~ Not applicable I think

### DIFF
--- a/src/png.imageio/png_pvt.h
+++ b/src/png.imageio/png_pvt.h
@@ -684,6 +684,13 @@ write_info(png_structp& sp, png_infop& ip, int& color_type, ImageSpec& spec,
                      (png_uint_32)(yres * scale), unittype);
     }
 
+#ifdef PNG_eXIf_SUPPORTED
+    std::vector<char> exifBlob;
+    encode_exif(spec, exifBlob, endian::big);
+    png_set_eXIf_1(sp, ip, static_cast<png_uint_32>(exifBlob.size()),
+                   reinterpret_cast<png_bytep>(exifBlob.data()));
+#endif
+
     // Deal with all other params
     for (size_t p = 0; p < spec.extra_attribs.size(); ++p)
         put_parameter(sp, ip, spec.extra_attribs[p].name().string(),
--- a/testsuite/png/ref/out-libpng15.txt
+++ b/testsuite/png/ref/out-libpng15.txt
@@ -1,0 +1,24 @@
+Reading ../oiio-images/oiio-logo-no-alpha.png
+../oiio-images/oiio-logo-no-alpha.png :  135 x  135, 4 channel, uint8 png
+    SHA-1: 75ED9B183E083ECF0A4881EEE0B553457B698CF4
+    channel list: R, G, B, A
+    Comment: "Created with GIMP"
+    DateTime: "2009:03:26 17:19:47"
+    ResolutionUnit: "inch"
+    XResolution: 72.009
+    YResolution: 72.009
+    oiio:ColorSpace: "sRGB"
+Comparing "../oiio-images/oiio-logo-no-alpha.png" and "oiio-logo-no-alpha.png"
+PASS
+Reading ../oiio-images/oiio-logo-with-alpha.png
+../oiio-images/oiio-logo-with-alpha.png :  135 x  135, 4 channel, uint8 png
+    SHA-1: 8AED04DCCE8F83B068C537DC0982A42EFBE431B6
+    channel list: R, G, B, A
+    Comment: "Created with GIMP"
+    DateTime: "2009:03:26 18:44:26"
+    ResolutionUnit: "inch"
+    XResolution: 72.009
+    YResolution: 72.009
+    oiio:ColorSpace: "sRGB"
+Comparing "../oiio-images/oiio-logo-with-alpha.png" and "oiio-logo-with-alpha.png"
+PASS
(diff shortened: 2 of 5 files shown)


joris-nijs:
If you think this contribution is valuable, could you give some guidance on:
- whether a test should be added somewhere
- whether a CLA is required for a change this small (I did do this on company time)

src/png.imageio/png_pvt.h
```diff
@@ -684,6 +684,13 @@ write_info(png_structp& sp, png_infop& ip, int& color_type, ImageSpec& spec,
                      (png_uint_32)(yres * scale), unittype);
     }
 
+#ifdef PNG_eXIf_SUPPORTED
```
lgritz, line 687: What determines if PNG_eXIf_SUPPORTED is found? Is it always present in sufficiently new libpng? (If so, how new? Is it always present based on a minimum version that we should try enforcing?)

src/png.imageio/png_pvt.h
```diff
@@ -684,6 +684,13 @@ write_info(png_structp& sp, png_infop& ip, int& color_type, ImageSpec& spec,
...
+    std::vector<char> exifBlob;
+    encode_exif(spec, exifBlob, endian::big);
+    png_set_eXIf_1(sp, ip, static_cast<png_uint_32>(exifBlob.size()),
```
lgritz, line 690: Should we also be using png_get_eXIf functions for the reading, or is what we have sufficient?

lgritz:
**whether a test should be added somewhere** -- yes, it would be helpful to test this. I think the best approach would be to augment testsuite/png/run.py to just write a simple small png with some Exif metadata, then print the metadata for the file you wrote and make sure it matches what you think you set? I can help you with this if you aren't quite sure what to do in those scripts.

**whether a CLA is required for a change this small (I did do this on company time)** -- that mostly depends on the arrangement you have with your company. For a change this small, the OIIO project doesn't insist on a CLA. But you know more about the relationship with your employers. If you think there could be any dispute and you want to get it in writing that it was ok for you to contribute this code, then by all means have them fill out a CLA (more for your safety than ours). But if you are confident that company policy is that you are free to make contributions to open source projects like this one, then we'll take it at face value and won't require a CLA.

src/png.imageio/png_pvt.h
joris-nijs: Indeed libpng >= 1.6.32 added support for the eXIf chunk in PNG and defines PNG_EXIf_SUPPORTED. To be honest, I'm not entirely sure if there is a way to build libpng >= 1.6.32 without this support. I took the same approach as is used elsewhere in this file for reading eXIf data.

src/png.imageio/png_pvt.h
kmilos: Looks like this is already the case: https://github.com/OpenImageIO/oiio/blob/master/src/png.imageio/png_pvt.h#L326-L336

src/png.imageio/png_pvt.h
lgritz: Makes sense.

src/png.imageio/png_pvt.h
joris-nijs: Indeed, OIIO seems to read Exif from PNG from the eXIf chunk and from embedded text (unofficial convention from before the eXIf chunk existed).

joris-nijs:
I added a test to the png test suite as you suggested. Please let me know if you want this to be extended more.

lgritz (approved):
Well done, that's exactly the kind of test I had in mind.

lgritz:
It turns out that for many of our CI tests (which purposely span many versions of various dependencies), the ones running in CentOS7-based containers are using an older version of libpng that doesn't support this feature! So the new testsuite/png/ref/out.txt now no longer matches the output of those tests.

To fix it, restore the original out.txt and name it out-libpng15.txt (leaving out.txt being the correct output for the newer libpng varieties). The testsuite scripts will know that a test can pass if it matches *any* of the reference outputs.

lgritz:
CI is still failing, but I know why, it's simple. I'm going to merge this and then just immediately amend the fix myself.

Thanks for the patch!

testsuite/png/ref/out-libpng15.txt
```diff
@@ -0,0 +1,24 @@
...
+    SHA-1: 8AED04DCCE8F83B068C537DC0982A42EFBE431B6
+    channel list: R, G, B, A
+    Comment: "Created with GIMP"
+    DateTime: "2009:03:26 18:44:26"
+    ResolutionUnit: "inch"
+    XResolution: 72.009
+    YResolution: 72.009
+    oiio:ColorSpace: "sRGB"
+Comparing "../oiio-images/oiio-logo-with-alpha.png" and "oiio-logo-with-alpha.png"
+PASS
```
lgritz, line 24:
```suggestion
PASS
Reading exif.png
exif.png             :   64 x   64, 3 channel, uint8 png
    SHA-1: 7CB41FEA50720B48BE0C145E1473982B23E9AB77
    channel list: R, G, B
    oiio:ColorSpace: "sRGB"
```

PR #3737, opened 2023-01-03
Docs: for rtd, upgrade sphinx/breathe versions, use furo theme

lgritz:
I think this looks better.

Also a few minor tweaks to typos and misquotes.


lgritz:
Here is (temporarily) an example of the new look: https://openimageio.readthedocs.io/en/rtd/index.html

PR #3740, opened 2023-01-04
Revert 3728, that was ill advised

lgritz:
Note that these intrinsics don't seem to be available for gcc. They are available on clang, but are defined in terms of the memcpy trick. But in Intel, they seem to actually help.


AlexMWells (approved):
LGTM

PR #3744, opened 2023-01-07
CI: Lock down to older icc so it's not broken

lgritz:
The new 2023 one on the Intel site seems to require a glibc too new to be compatible with the one on the ASWF CentOS-7 based containers, so has been breaking our CI for a couple weeks.


AlexMWells (commented):
LGTM, but we will probably want to have different locked versions for icc vs. icx.  For instance, I think it would be safe to  move icx ahead to version 2023.0.0.

lgritz:
This patch only locks down icc. The icx is allowed to float and stay on the latest version, which did not break as icc did.

PR #3750, opened 2023-01-09
Fix wrong oiio:UnassociatedAlpha metadata for PSD files

brechtvl:
## Description

Refactoring in #1641 incorrectly caused this to be set on the last layer rather than the composite image as before. Now set it on the composite image and every layer, as all of them use unassociated alpha.

## Tests

Test was added.

## Checklist:


lgritz (approved):
LGTM, thanks for the fix!

PR #3761, opened 2023-01-31
Add color management to texture lookups

lgritz:
This isn't 100% final -- I'm sure there are some corner cases I missed, maybe things that will eventually need some performance work, but enough works and passes tests that I'm confident in the API and so I want to get this merged and we can iterate on internals later.

The use case is as follows: User or shader knows that a particular texture is in something other than the default working color space, this is not necessarily apparent from the file itself and so cannot be fully automatic, but it is presumed that it can be communicated in the texture call. We aim to be efficient, fully converting each tile upon input. Note that we are sweeping under the rug the fact that a non-default color space lookup is not mathematically correct with respect to the MIP-map downsizing math if that was not considered at texture creation time. But by converting texels as they are read, rather than the shader itself trying to do a color transformation on the final texture result, at least we are doing the right math for the filtering within and among MIP levels.

The new interface consists of communicating the presumed working color space, finding the colortransform ID of a given from/to (with the "to" defaulting to the working space, if not specified), and then on each texture call, supplying that colortransform ID as one of the settings in TextureOpt (or TextureOptBatch).

The outline of changes are as follows:

* BREAKING CHANGE: TextureOpt and TextureOptBatch now contain an additional int `colortransformid`. The default, 0, is the old (and, frankly, preferred) behavior. Nonzero values indicate that we want the texels to go through a color transformation as the tiles are read into the cache.

* BREAKING CHANGE: New public methods have been added to TextureSystem `get_colortransform_id(fromspace, tospace)` that return a transform ID given the from and to color spaces (ustring or ustringhash). This is the value that goes in the TextureOpt to request a color transformation on the pixels as they are read.

* BREAKING CHANGE: IC::get_image_handle / TS::get_texture_handle now take an additional optinal paramter that is a TextureOpt*. This is currently unused, but is reserved for possible future use and/or for alternate TextureSystem implementations that choose a different implementation strategy wherein an entirely different TextureHandle may correspond to each color transformation. That's not how I've done it here, but it's a valid approach (and maybe one we'll switch to in the future). We presume that it's ill-defined to ask for a handle with one explicit color transformation, then use that handle with a TextureOpt that implies a different color transformation. Using those consistently ensure that either TS strategy will work equally well.

* New TextureSystem attributes "colorspace" gives the working color space name (defaulting to "scene_linear"), and "colorconfig" can optionally point to a color config file (though this is not yet functional -- it just uses the default color config).

* Internally, the tile cache hash has been adjusted so that the same tile in the disk file will appear as a separate tile in the cache for each color space pair that is applied to it. Thus, there is no conflict if some different texture calls ask for different color space interpretations of the same texture file. (That said, it's not recommended to be throwing color spaces around willy nilly -- it still takes multiple cache entrie, plus extra color conversion math as the tiles are read in.)

* A new preprocessor symbol OIIO_TEXTURESYSTEM_SUPPORTS_COLORSPACE in texture.h allows client software can easily tell if this is a version new enough to support color management and has the additional fields in the texture option structures.

Because there are breaking changes here to ABI, this is destined only for master (future 2.5) and will not be backported to OIIO 2.4.


lgritz:
@MasterZap this is the OIIO side. It's getting pretty late, but in the morning I'll post a (smaller, simpler) PR for some changes you'll also need on the OSL side that interfaces with this.

src/include/OpenImageIO/texture.h
```diff
@@ -264,6 +266,8 @@ class OIIO_API TextureOpt {
     float rblur;   ///< Blur amount in the r direction
     float rwidth;  ///< Multiplier for derivatives in r direction
 
+    int colortransformid;       ///< Color space id of the texture
```
etheory, line 269: If colortransformid can never be negative, why not make it an unsigned int?

src/libtexture/imagecache.cpp
```diff
@@ -791,36 +794,41 @@ ImageCacheFile::init_from_spec()
 
 
 bool
-ImageCacheFile::read_tile(ImageCachePerThreadInfo* thread_info, int subimage,
-                          int miplevel, int x, int y, int z, int chbegin,
-                          int chend, TypeDesc format, void* data)
+ImageCacheFile::read_tile(ImageCachePerThreadInfo* thread_info,
+                          const TileID& id, void* data)
```
etheory, line 798: Wow, I MUCH prefer the parsing of the structure here, that's WAY nicer as an API.

src/libtexture/imagecache.cpp
```diff
@@ -852,6 +860,19 @@ ImageCacheFile::read_tile(ImageCachePerThreadInfo* thread_info, int subimage,
         thread_info->m_stats.bytes_read += b;
         m_bytesread += b;
         ++m_tilesread;
+        if (id.colortransformid() > 0) {
```
etheory, line 863: This is nice, but it feels a little scary that when a tile gets evicted, we have to do this all over again.
Sure, that's not a reason not to, but I'm just pointlessly sighing about that, even though it seems unavoidable.

etheory (commented):
LGTM so far, I need to digest this a little bit more, however, but it's a really great PR. Thanks.

src/include/OpenImageIO/texture.h
lgritz: Just general animus toward unsigned types? All the other fields in this struct were int even though they technical have no meaningful negative values.

src/libtexture/imagecache.cpp
lgritz: Not sure what the alternative could be. The I/O is probably as expensive than the computation. Also, remember the context: people who really care about performance should be ensuring that their textures are in the working space when they are created, not in the middle of a shader.

src/libtexture/imagecache.cpp
```diff
@@ -875,6 +894,15 @@ ImageCacheFile::read_unmipped(ImageCachePerThreadInfo* thread_info,
     // N.B. No need to lock the mutex, since this is only called
     // from read_tile, which already holds the lock.
 
+    int subimage = id.subimage();
+    int miplevel = id.miplevel();
+    int x        = id.x();
+    int y        = id.y();
+    // int z        = id.z();
```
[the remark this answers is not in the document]
lgritz: No, I'm just commenting out a variable that wasn't used. But 3d textures were never correctly supported by this special function that populates downsized MIP level tiles for unmipped files. MIP mapping is not used for volume textures currently, thus no mip level > 0 is read from for a texture3d lookup, so this function never is called for the 3d volume case.

lgritz:
The corresponding change on the OSL side is https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1641

lgritz:
I will merge this at end of day tomorrow unless somebody has concerns.

PR #3764, opened 2023-02-03
Initialize OpenEXROutput::m_levelmode() in init().

jreichel-nvidia:
Avoids conditional jump on uninitialised values in close(). Not sure whether ONE_LEVEL is the best value here ...

## Description

Without that fix, valgrind reports a conditional jump on uninitialised values on code like
```
auto out = OIIO::ImageOutput::create( "exr");
```

## Checklist:


lgritz (approved):
LGTM

PR #3767, opened 2023-02-09
improve thread safety for concurrent tiff loads

angusdavisadobe:
## Description
This fixes a thread-safety issue with concurrent parsing of multiple tiff files. The code uses several static std::unordered_maps and the indexing operator[] was used to perform the lookup. Use of this operator has the side effect of adding a default value to the map if the key was not found. This is not what should happen and isn't thread safe. The proper fix is to make the maps const and use the find() method of the map to perform the lookup and return a default value if not found.

## Tests

Tests were not performed. However the code changes were tested within the context of loading USD files with multiple tiff files.

## Checklist:

- [X ] I have read the [contribution guidelines](https://github.com/OpenImageIO/oiio/blob/master/CONTRIBUTING.md).
- [X ] My code follows the prevailing code style of this project.


src/libOpenImageIO/icc.cpp
```diff
@@ -131,20 +137,32 @@ icc_color_space_name(string_view color_space)
...
     };
-    return primary_platforms[platform];
-}
+    // std::unordered_map::operator[](const Key& key) will add the key to the map if
+    // it doesn't exist. This isn't what is intended and isn't thread safe.
+    // Instead, just do the lookup and return the value or a nullptr.
+    //
+    // return primary_platforms[platform];
+    auto it = primary_platforms.find(platform);
+    return (it != primary_platforms.end()) ? it->second : nullptr;
```
lgritz, line 165: I think you lost a `}` at this line...

src/libOpenImageIO/icc.cpp
```diff
@@ -131,20 +137,32 @@ icc_color_space_name(string_view color_space)
...
     };
-    return color_space_names[color_space];
+    // std::unordered_map::operator[](const Key& key) will add the key to the map if
+    // it doesn't exist. This isn't what is intended and isn't thread safe.
+    // Instead, just do the lookup and return the value or a nullptr.
+    //
+    // return color_space_names[color_space];
+    auto it = color_space_names.find(color_space);
+    return (it != color_space_names.end()) ? it->second : nullptr;
+
```
lgritz, line 147: Can you get rid of the extra blank line here, please?

lgritz (commented):
Wow, good catch. It did not occur to me that it might be queried with a key that wasn't already in the map, thus potentially altering it and breaking thread safety. I should have declared these const all along.

The whole thing LGTM except for the syntax error I pointed out with the missing brace.

src/libOpenImageIO/icc.cpp
angusdavisadobe: Sorry for the carelessness. I added back the missing `}`

src/libOpenImageIO/icc.cpp
angusdavisadobe: I removed the blank line.

lgritz (approved):
LGTM, thanks!

PR #3774, opened 2023-02-13
Don't use check_cxx_source_runs to make cross-compiling easier

### LINKED ISSUE #3773: [FEATURE REQUEST] make oiio easier to cross compile by not using check_cxx_source_runs

I'm cross-compiling oiio (specifically Linux x86_64 -> Linux aarch64), and I have to set `-DDCOMPILER_SUPPORTS_ATOMIC_WITHOUT_LIBATOMIC_EXITCODE=0` to be able to compile this library, which works fine but looking at the test that's being performed it seems like a `check_cxx_source_compiles` would be sufficient to see if libatomic is necessary? This would allow cmake to cross compile it without any extra flags passed. 

Use `check_cxx_source_compiles` instead of `check_cxx_source_runs` for libatomic.

Doing nothing

**Additional context**

If this seems reasonable, I'm happy to open a PR.

russelltg:
## Description

Fix #3773

## Tests

I did test it on my local Linux system, and it properly detects that it doesn't need to link against . I also tested compiling the test source file in a system that needs linking against latomic (32-bit ubuntu 14.04 using clang, there is probably some other system that is this way as well) and the compile itsself did indeed fail with undefined reference errors, as expected. So this check should be just as robust. 

## Checklist:

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -350,7 +350,7 @@ include (CheckLibraryExists)
 #
 if (NOT MSVC AND NOT APPLE AND NOT ANDROID)
     cmake_push_check_state ()
-    check_cxx_source_runs(
+    check_cxx_source_compiles(
        "#include <atomic>
         #include <cstdint>
         std::atomic<uint64_t> x {0};


lgritz (approved):
I see. I think I assumed that check_cxx_source_compiles only checked for compilation of the source file, and did not realize it would link the executable. You're right, that's enough, it doesn't need to be executed. LGTM.

Issue #3776, opened 2023-02-20
Please add an option to disable the stacktrace functionality


jreichel-nvidia:
[oiio_disable_boost_stacktrace.diff.txt](https://github.com/OpenImageIO/oiio/files/10784406/oiio_disable_boost_stacktrace.diff.txt)
**Is your feature request related to a problem? Please describe.**
The stacktrace functionality creates an additional dependency on helper libraries like dbgeng.dll, even if it is not enabled. Such a dependency might be undesired in larger integrations.

**Describe the solution you'd like**
A cmake option to disable that functionality, see attachment for proof of concept.

**Additional context**
Should OIIO_HAS_STACKTRACE be affected by the patch? It is not affected by the Boost version check either (which might create wrong resultsif  the header is used with a different boost version than the one used for building OIIO). Its documentation also says "has", not "working". Is the define basically just for convenience and serves the same purpose as a version check on OIIO_VERSION?

lgritz:
I think this is a very reasonable request, and your proposed patch looks fine. Can you please submit it as a PR?

Let's keep OIIO_HAS_STACKTRACE as it is -- it indicates that the API exists, even if internally we have it disabled at OIIO-build time.

mhmdanas:
#3777 has been merged, is there any reason for this to remain open?

lgritz:
Closed by #3777

PR #3777, opened 2023-02-21
Add cmake option to disable the stacktrace functionality.

jreichel-nvidia:
Setting the cmake option OIIO_DISABLE_BOOST_STACKTRACE to ON disables the use of Boost stacktrace at compile time. This avoids dependencies on helper libraries like dbgeng.dll.

See issue #3776.

## Checklist:


lgritz (approved):

lgritz:
I'm not sure what's up with the CI jobs that die. It doesn't seem related to your patch. They're not actually failing, the entire job never starts. I assume it's a temporary GHA outage.

Issue #2929, opened 2021-04-02
[BUG] OIIO::ImageBufAlgo::colorconvert returns black pixels when converting from sRGB to linear RGB (MSVC, SSE enabled)


andreasdeclercq:
I noticed that the OIIO::ImageBufAlgo::colorconvert method consistently returns black pixels when converting from sRGB to linear RGB, but only when:

- Compiler is MSVC (my version is 14.28.29910) and OS is Windows (10)
- Build type is Release or RelWithDebInfo (Debug works just fine)
- OIIO_NO_SSE is NOT defined (works fine if SSE is disabled)

Below is a simple gtest snippet that passes when I run it in debug mode but fails when I run it in release mode.
Note that the image size, number of channels or initial pixels values do not matter as long as they are not black or white (255 is ok, 254 is not ok).

```
#include "gtest/gtest.h"
#include <array>
#include "OpenImageIO/imagebufalgo.h"

TEST(ColorConversionTests, sRGBToLinearRGBConversionTest)
{
    // Create a uint8 Image Buffer with dummy data
    constexpr int imageWidth = 1;
    constexpr int imageHeight = 1;
    constexpr int channels = 3;
    constexpr int strideLength = imageWidth * channels + 1;
    OIIO::ImageSpec spec(imageWidth, imageHeight, channels, OIIO::TypeDesc::UINT8);
    spec["oiio:ColorSpace"] = "sRGB";
    auto image = OIIO::ImageBuf(spec);

    // Set all pixels to a uniform color (that is neither black or white)
    std::array<unsigned char, imageWidth* imageHeight* channels> imageData = {};
    imageData.fill(128);
    image.set_pixels(OIIO::ROI::All(), OIIO::TypeDesc::UINT8, imageData.data(), OIIO::AutoStride, strideLength);
    ASSERT_TRUE(image.pixels_valid());

    const bool conversionOk = OIIO::ImageBufAlgo::colorconvert(image, image, "sRGB", "Linear");
    EXPECT_TRUE(conversionOk);

    for (OIIO::ImageBuf::ConstIterator<unsigned char, unsigned char> it(image); !it.done(); ++it)
    {
        ASSERT_TRUE(it.exists() && it.valid());
        for (int c = 0; c < channels; ++c)
            EXPECT_EQ(it[c], 55);
    }
}
```

[...]

PR #3804, opened 2023-04-14
Fix `fast_exp2` with MSVC and sse4.2

ericmehl:
## Description

I ran into the problem described in #2929 and did a little digging to see if I could find the source of the bad color conversions. I tracked it down to `fast_exp2()` https://github.com/OpenImageIO/oiio/blob/a77e36250fadfb4c391688eb1df4b3dd88d5e367/src/include/OpenImageIO/fmath.h#L2050-L2080

and in particular the return statement on L2071. In the process of trying to figure out what part of that return statement was going awry, I found that separating the bitcasts to separate variables as in this PR, the problem went away. All testing was done with the `color_test` added in #3023.

It also succeeded when adding various `std::cout` lines to check what the values of each computation were, so it does sound like a compiler optimization that does or does not get triggered is the problem.

Unfortunately I don't know why it's not failing on the CI builds, I double checked to make sure I'm using the same settings and flags. I also tested on two different machines and get the same failures, and success with this patch, on both.

One strange wrinkle is that even without this patch, color conversions worked fine when using AVX2 for the SIMD mode.

Summary :
Previous results (without this patch)

- USE_SIMD=0 - PASS
- USE_SIMD=SSE4.2 - FAIL
- USE_SIMD=AVX - FAIL
- USE_SIMD=AVX2 - PASS

With this patch
- USE_SIMD=0 - PASS
- USE_SIMD=SSE4.2 - PASS
- USE_SIMD=AVX - PASS
- USE_SIMD=AVX2 - PASS

Visual Studio 2019 versions 16.11.32228.343, 16.11.33328.57 and 16.11.33423.256(same as current Github build runners) in RelWithDebInfo mode.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -2068,7 +2068,19 @@ OIIO_FORCEINLINE OIIO_HOSTDEVICE T fast_exp2 (const T& xval) {
     r = madd(x, r, kD);
     r = madd(x, r, kE);
     r = madd(x, r, one);
-    return bitcast_to_float (bitcast_to_int(r) + (m << 23));
+
+    // Original code was:
+    //     return bitcast_to_float (bitcast_to_int(r) + (m << 23));
+    // This was producing the wrong results when called via `sRGB_to_linear()`
+    // and `linear_to_sRGB()` on some Windows MSVC builds.
+    // We found that it was fixed by using a temporary intN, as below.
+    // Presumed to be an optimizer bug in MSVC versions 16.11.x
+    // and earlier.
+    // See PR #3804
+
+    intN i = bitcast_to_int(r);
+    T f = bitcast_to_float(i + (m << 23));
+    return f;
 #else
     T r;
     for (int i = 0; i < r.elements; ++i)


lgritz:
How strange!

Can I request that you please add a test to fmath_test.cpp (perhaps in `test_math_functions()`) using specific values that would fail without the fix, and succeed with it?

src/include/OpenImageIO/fmath.h
```diff
@@ -2068,7 +2068,10 @@ OIIO_FORCEINLINE OIIO_HOSTDEVICE T fast_exp2 (const T& xval) {
     r = madd(x, r, kD);
     r = madd(x, r, kE);
     r = madd(x, r, one);
-    return bitcast_to_float (bitcast_to_int(r) + (m << 23));
+
```
lgritz, line 2071: Suggestion to please add a comment here along the lines of:

```
  // Original code was:
  //     return bitcast_to_float (bitcast_to_int(r) + (m << 23));
  // This was producing the wrong results when <explain configuration>
  // and we found that it was fixed by using a temporary intN, as below.
  // Presumed to be an optimizer bug in <explain compiler>?
  // See PR #3804
```

lgritz:
This is quite an obscure problem, I appreciate your taking the time to track it down to a single line and produce a fix.

src/include/OpenImageIO/fmath.h
ericmehl: Done and squashed onto the original commit into daeda454412beaa3f305287acaad115e46f7d620

ericmehl:
> This is quite an obscure problem, I appreciate your taking the time to track it down to a single line and produce a fix.

My pleasure! It's been a strange journey.

> Can I request that you please add a test to fmath_test.cpp (perhaps in test_math_functions()) using specific values that would fail without the fix, and succeed with it?

I tried adding a test for `fast_exp2()` but I couldn't get it to fail on the unpatched code. Even when a test of `linear_to_sRGB()` and `sRGB_to_linear()` directly before it failed. I also tried pasting in the code for the return value of https://github.com/OpenImageIO/oiio/blob/3332c78f5eacd65c622f97941c2249318b395db9/src/include/OpenImageIO/color.h#L374-L379

and _that_ passed. So it seems to be unique to the case of being called / inlined into the `linear_to_sRGB()` and `sRGB_to_linear()` functions.

Does that suggest any new solutions or other avenues to investigate?

lgritz (approved):

johnhaddon:
Thanks for merging @lgritz. Do you think this could be backported for the next 2.4 release (it fixes a problem with black icons in Gaffer on Windows)?

lgritz:
Certainly. Do you need a tagged release right away, or is the scheduled May 1 patch good enough?

johnhaddon:
Thanks! May 1st will be perfectly sufficient - we have a workaround in the meantime.

lgritz:
I already had merged it into the head of dev-2.4, by the way. The only thing it's lacking is a release (and whatever else I may add between now and the end of the month.

PR #3806, opened 2023-04-15
fix(psd): Handle very wide images

lgritz:
When decoding RLE data, some arguments to decompress_packbits() were inexplicably declared uint16_t (they were passed uint32_t arguments, and I honestly don't know why we didn't get a warning all along). This prevented us from correctly reading PSD images whose width was more than 32767, which I suppose are rare enought that it's not hard to understand why this went unreported for so long.

But anyway, just changing those parameters to uint32_t does the trick.


etheory (approved):
LGTM

PR #3808, opened 2023-04-18
Fix building on OpenBSD

brad0:


lgritz (changes requested):
Thanks for the fix!

LGTM, though it seems to be failing the "clang-format" test, can you please fix the line wrap here? (Click on the failed test to see where it "breaks," which will include the diff that shows how to make it conform.)

brad0:
> though it seems to be failing the "clang-format" test

Sure thing, done.

lgritz (approved):
LGTM, thanks. Will merge as soon as the rest of the CI passes.

lgritz:
Oops, I thought the "merge queue" would wait until all the checks passed. Oh well. I think it's safe, I don't think this can affect any non-OpenBSD builds, which of course aren't included in our CI.

brad0:
Please backport to 2.5 and 2.4.

lgritz:
I backported to 2.4 (in the dev-2.4 branch) and it will be part of the next tagged release scheduled for 1 May.

I'm not sure what you mean by "backport to 2.5." There is no official 2.5 release family yet; the current "master" branch is what will become 2.5 later this year.

brad0:
> I'm not sure what you mean by "backport to 2.5." There is no official 2.5 release family yet; the current "master" branch is what will become 2.5 later this year.

I saw something that looked like tags towards a 2.5 release and wasn't sure so included it just in case.

lgritz:
Yeah, I occasionally add "developer preview" tags to mark places that seem stable in the master branch, for people who need things that are only in master. But there's no support or backporting for those, just the occasional tag dropped to mark "I've used it in production here, seems ok."

PR #3810, opened 2023-04-19
Refactor simd.h to be sure to disable Intel inrinsics when not on Intel

### LINKED ISSUE #3809: [HELP] Compiler error in simd.h with CUDA/__AVX__
Hi Larry and everyone,

It seems like you can run into a compiler error when compiling a CUDA file that defines `__AVX__` and includes `simd.h`. 
This happens because the SIMD intrinsics header [is not included when `__CUDA_ARCH__` is defined](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/simd.h#L85) but later the code [still attempts to define the AVX version](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/simd.h#L139
) and uses the types resulting in undefined symbols.
For the SSE case [this is prevented](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/simd.h#L107) by explicitly defining `OIIO_NO_SSE` and the idea for the fix we had in mind was to add the AVX equivalent.

```
#if defined(__CUDA_ARCH__) && !defined(OIIO_NO_AVX)
#define OIIO_NO_AVX 1
#endif
```

Thought I would check in with you before making a PR to see if you were aware of the issue and maybe had a different fix in mind.
For reference, we encountered this issue when building OpenShadingLanguage with Optix support.

Cheers,
Edoardo

lgritz:
Especially when compiling for Cuda.

Fixes #3809

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -81,30 +81,38 @@
 #  endif
 #endif
 
+// Disable Intel SIMD intrinsics on non-Intel architectures (including
+// building for Cuda on an Intel host).
+#if defined(_M_ARM64) || defined(__aarch64) || defined(__aarch64__) \
+    || defined(__CUDA_ARCH__)
+#  ifndef OIIO_NO_SSE
+#    define OIIO_NO_SSE 1
+#  endif
+#  ifndef OIIO_NO_AVX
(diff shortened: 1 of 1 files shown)


AlexMWells (approved):
LGTM

PR #3813, opened 2023-04-21
Support HIF extension

AdamMainsTL:
## Description

I've come across some unsupported files from Canon cameras that have a .hif extension. These are HEIC/HEIF files with a slightly changed extension. I have no idea why Canon decided to do this. Without any changes I am able to open the images, probably because it checks the header, but I am unable to write to paths with a hif extension.

Some random articles/forum posts mentioning the HIF format in relation to Canon cameras:
https://www.cameralabs.com/jpeg-vs-heif-canon-hif-comparison/
https://www.dpreview.com/forums/thread/4558661
https://www.reddit.com/r/canon/comments/kjc6xl/new_eos_r6_owner_here_photo_files_show_up_as_hif/

## Tests

I did not add tests to the project, but I did compile the library into our application on Windows and Mac environments and verified that the old version does not support writing to paths with a "hif" extension and the updated code does. Can provide more info if needed.

## Checklist:

### DIFF
--- a/src/heif.imageio/heifinput.cpp
+++ b/src/heif.imageio/heifinput.cpp
@@ -76,7 +76,7 @@ heif_input_imageio_create()
     return new HeifInput;
 }
 
-OIIO_EXPORT const char* heif_input_extensions[] = { "heic",  "heif", 
+OIIO_EXPORT const char* heif_input_extensions[] = { "heic",  "heif",
                                                     "heics", "hif",
 #if LIBHEIF_HAVE_VERSION(1, 7, 0)
                                                     "avif",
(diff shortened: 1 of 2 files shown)


src/heif.imageio/heifinput.cpp
```diff
@@ -76,7 +76,8 @@ heif_input_imageio_create()
     return new HeifInput;
 }
 
-OIIO_EXPORT const char* heif_input_extensions[] = { "heic", "heif", "heics",
+OIIO_EXPORT const char* heif_input_extensions[] = { "heic",  "heif",
```
lgritz, line 79: There seems to be an extra space here that's tickling the clang-format failure:

```suggestion
OIIO_EXPORT const char* heif_input_extensions[] = { "heic",  "heif",
```

lgritz:
I assume that the HIF is a concession to flash memory cards with FAT filesystem that require an 8.3 filename scheme.

Looks reasonable to me. Please accept the format fix suggestion and then I'll be happy to merge.

PR #3814, opened 2023-04-22
Fix incorrect branch name when cloning openexr-images for the tests

jessey-git:
## Description

The `openexr-images` repo was never updated after the default branch was renamed to `main`. This will fix the following silent errors in CI:
```
Cloning into '/__w/oiio/oiio/build/testsuite/openexr-images'...
warning: Could not find remote branch master to clone.
fatal: Remote branch master not found in upstream origin
```

Additionally, the cmake attempted to symbolically link local repos like `oiio-images` if they were found. Unfortunately this type of link will fail unless running as admin on Windows (which is rare) and the COPY_ON_ERROR directive did _not_ work here (cmake 3.24; weird; sigh...).  The output will now look something like the following:
```
-- Linking oiio-images from C:/Users/jesse/source/oiio-images
-- Link result failed to create symbolic link 'C:/Users/jesse/source/oiio/build/testsuite/oiio-images': A required privilege is not held by the client.
-- Copying oiio-images from C:/Users/jesse/source/oiio-images
```

## Tests

These were found while trying to get the test suite running on Windows here in preparation for some development. Sadly the suite is still not running correctly at all so I may have to rely on the CI.

## Checklist:

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -360,15 +360,18 @@ function (oiio_get_test_data name)
        # Arguments: <prefix> <options> <one_value_keywords> <multi_value_keywords> args...
     if (IS_DIRECTORY "${OIIO_LOCAL_TESTDATA_ROOT}/${name}"
         AND NOT EXISTS "${CMAKE_BINARY_DIR}/testsuite/${name}")
+        set (_ogtd_LINK_RESULT "")
         if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.14)
             # Just make a link if we can
-            message (STATUS "Linking ${name} from ${OIIO_LOCAL_TESTDATA_ROOT}/../${name}")
+            message (STATUS "Linking ${name} from ${OIIO_LOCAL_TESTDATA_ROOT}/${name}")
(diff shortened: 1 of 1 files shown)


lgritz:
Good catch. Fixing the URL is definitely good. In fact, it looks like maybe we have been skipping some of the openexr tests in CI because of this, and had not noticed! That's very bad, really embarrassing oversight. (I, frankly, never thought to check and if you had asked me, I would have said I assumed that the branch rename would forever after redirect properly so that any requests for master would get main automatically.) Some of the test failures you're seeing appear to be because these tests haven't run for a while for those testing combinations, and need some updates to the reference output for those tests. I won't hold up this PR because of that, I will just fix it after the fact by checking in new ref output.

I'm still confused about why the `COPY_ON_ERROR` is not sufficient here. Do you have an explanation? I wouldn't think that code change was necessary.

lgritz:
OK, then let's keep your logic change, but can you please put in a comment that explains that we thought that `COPY_ON_ERROR` would implement this behavior, it doesn't appear to in this circumstance and we don't know why, but that it's something we should check from time to time and maybe can return to the simpler construct if cmake eventually fixes the behavior.

lgritz (approved):
LGTM, thanks for the fix

PR #3823, opened 2023-04-27
Fix float128 compilation of NVPTX backend for LLVM

dedoardo:
## Description

The latest version of libfmt began using the `__float128` type but this is [not supported in the NVPTX backend for LLVM](https://discourse.llvm.org/t/compiling-cuda-code-fails/61240/13). A way to detect this particular platform seems to be to check both `__CUDA_ARCH__` and  `__clang__`.
I am not disabling it entirely on any CUDA target because NVCC 10+ understands `_float128` and generates instructions for it when tested on godbolt. 

Following up from https://github.com/OpenImageIO/oiio/issues/3809.
             

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/detail/fmt.h
+++ b/src/include/OpenImageIO/detail/fmt.h
@@ -30,8 +30,8 @@
 #    define FMT_DEPRECATED_OSTREAM 1
 #endif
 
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM, thanks for the fix!

PR #3826, opened 2023-05-03
Add an ImageInput::valid_file(IOProxy) overload

jessey-git:
## Description
Every format that has an existing `valid_file(std::string)` implementation, and which supports an
ioproxy, now also has a `valid_file(IOProxy)` implementation for completeness.

The existing std::string overloads have been implemented in terms of the IOProxy overload where
possible to cut down on the boilerplate. This is done at the base of the hierarchy in ImageInput.

The remaining formats that support ioproxy, but do not currently have a `valid_file` method, will be
submitted in a follow-up PR as the additional logic required for those will require closer review.

## Tests
I haven't had much time/luck to sort out running the tests locally on Windows, or setting up a parallel WSL
environment, so I admit this is coming in with no test coverage beyond it working locally for me in a
test app. The app invokes both `valid_file` methods on a directory filled with various files for the formats in question.

## Checklist:

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -1024,9 +1024,11 @@ public:
     /// @returns
     ///         `true` upon success, or `false` upon failure.
     virtual bool valid_file (const std::string& filename) const;
+
+    /// Check valid file using an IOProxy.
     virtual bool valid_file (Filesystem::IOProxy* io) const;
 
-    /// Check valid vile using a UTF-16 encoded wstring filename.
+    /// Check valid file using a UTF-16 encoded wstring filename.
     bool valid_file (const std::wstring& filename) const {
         return valid_file(Strutil::utf16_to_utf8(filename));
     }
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -1024,6 +1024,7 @@ public:
     /// @returns
     ///         `true` upon success, or `false` upon failure.
     virtual bool valid_file (const std::string& filename) const;
+    virtual bool valid_file (Filesystem::IOProxy* io) const;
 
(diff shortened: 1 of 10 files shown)


src/include/OpenImageIO/imageio.h
```diff
@@ -1025,7 +1025,10 @@ class OIIO_API ImageInput {
     ///         `true` upon success, or `false` upon failure.
     virtual bool valid_file (const std::string& filename) const;
 
-    /// Check valid vile using a UTF-16 encoded wstring filename.
+    /// Check valid file using an IOProxy.
+    virtual bool valid_file (Filesystem::IOProxy* io) const;
+
+    /// Check valid file using a UTF-16 encoded wstring filename.
```
lgritz, line 1031: Would you mind, here, (a) moving the new one after the next method (so the two that take filenames are next to each other), and (b) replicating the full documentation comment for the IOProxy one, of course adjusted to describe the new argument?

lgritz (changes requested):
I like this, thanks!

The only comment I had was very minor, just an organization/documentation tweak about the header.

Check the clang-format test in the CI run to see what's failing.

src/include/OpenImageIO/imageio.h
jessey-git: Sure thing! I also noticed that the public API here preferred to use `ioproxy` as the parameter name so I changed everything from `io` to `ioproxy` for consistency.

lgritz: perfect, thanks

lgritz (approved):
LGTM

lgritz:
The CI failure for icc doesn't appear to be related to your changes, it's (temporarily?) unable to download package containing the icc compiler. So I'm going to merge.

PR #3830, opened 2023-05-06
cleanup: root out deprecated simd type names; `OIIO_DISABLE_DEPRECATED`

lgritz:
We have some OIIO 1.x era type aliases float4, int4, etc., for the
simd types that have long since been renamed vfloat4, vint4, etc.
Hide those aliases when buiding OIIO itself, to be sure we aren't
inadvertently using them. And indeed, in one spot, we were! (Fix it.)

Introuce a new preprocessor symbol, `OIIO_DISABLE_DEPRECATED`, which
if set to an encoded version number (such as 20100 for version 2.1) by
a downstream package, indicates that we are permitted to hide
declarations that were considered deprecated from that version
on. This is an aid to downstream projects who wish to enforce that
they are not inadvertently using any deprecated OIIO calls which may
disappear in a future release.

So back to the deprecated simd names: Also hide them if
`OIIO_DISABLE_DEPRECATED` is defined to exclude things that were
deprecated as of OIIO 1.9+, which apparently is when we changed this
definition.

### DIFF
--- a/src/include/OpenImageIO/oiioversion.h.in
+++ b/src/include/OpenImageIO/oiioversion.h.in
@@ -88,6 +88,21 @@
                              OIIO_VERSION_MINOR, OIIO_VERSION_PATCH, \
                              OIIO_VERSION_TWEAK, "")
 
+// OIIO_DISABLE_DEPRECATED encodes the version for which any declarations that
+// were deprecated as of that version may be hidden from view of software
+// including the OIIO headers.  For example, if a downstream project says
+//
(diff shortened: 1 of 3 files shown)


lgritz:
Updated the PR, primarily to make OIIO_DISABLE_DEPRECATED not be simply defined/not defined, but to encode an OIIO version number, so the signal is "is's ok to hide declarations that were considered deprecated for OIIO equal to or later than this version."

PR #3836, opened 2023-05-13
build(deps): fmt 10.0 support

### LINKED ISSUE #3835: [BUG] Unable to compile with fmt 10.0.0
**Describe the bug**

Using the latest release of fmt, oiio fails to compile due to breaking changes from upstream.

The errors are most likely related to the following change:
```
Removed deprecated implicit conversions for enums and conversions to primitive types for compatibility with std::format and to prevent potential ODR violations. Use format_as instead.
```

**To Reproduce**
1. Install fmt 10.0.0 or set `BUILD_FMT_VERSION` to `"10.0.0"`
2. Try to build the project

**Expected behavior**

Build succeeds.

**Evidence**

Compilation terminates very early with errors such as:
```
[12/206] Building CXX object src/libut...es/strutil_test.dir/strutil_test.cpp.o
FAILED: src/libutil/CMakeFiles/strutil_test.dir/strutil_test.cpp.o 
ccache /usr/bin/clang++ -DOIIO_INTERNAL=1 -DUSE_BOOST_FILESYSTEM -DUSE_FREETYPE=1 -DUSE_JPEG_TURBO=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/oiio/build/include/OpenImageIO -I/oiio/build/include -I/oiio/build/src/include -I/oiio/src/include -isystem /vcpkg/installed/arm64-osx/include -isystem /vcpkg/installed/arm64-osx/include/Imath -g -std=c++14 -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.3.sdk -Wall -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -fno-math-errno -MD -MT src/libutil/CMakeFiles/strutil_test.dir/strutil_test.cpp.o -MF src/libutil/CMakeFiles/strutil_test.dir/strutil_test.cpp.o.d -o src/libutil/CMakeFiles/strutil_test.dir/strutil_test.cpp.o -c /oiio/src/libutil/strutil_test.cpp
In file included from /oiio/src/libutil/strutil_test.cpp:8:
In file included from /oiio/build/include/OpenImageIO/Imath.h:10:
In file included from /oiio/src/include/OpenImageIO/detail/fmt.h:41:
In file included from /oiio/build/include/OpenImageIO/detail/fmt/format.h:49:
/oiio/build/include/OpenImageIO/detail/fmt/core.h:1690:3: error: static_assert failed due to requirement 'formattable' "Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt"
  static_assert(
  ^
/oiio/build/include/OpenImageIO/detail/fmt/core.h:1711:10: note: in instantiation of function template specialization 'fmt::detail::make_value<fmt::basic_printf_context<fmt::appender, char>, const OpenImageIO_v2_4::basic_string_view<char> &>' requested here
  return make_value<Context>(val);
         ^
/oiio/build/include/OpenImageIO/detail/fmt/core.h:1825:23: note: in instantiation of function template specialization 'fmt::detail::make_arg<true, fmt::basic_printf_context<fmt::appender, char>, fmt::detail::type::custom_type, const OpenImageIO_v2_4::basic_string_view<char> &, 0>' requested here
        data_{detail::make_arg<
                      ^
/oiio/build/include/OpenImageIO/detail/fmt/core.h:1844:10: note: in instantiation of function template specialization 'fmt::format_arg_store<fmt::basic_printf_context<fmt::appender, char>, OpenImageIO_v2_4::basic_string_view<char>, OpenImageIO_v2_4::basic_string_view<char>>::format_arg_store<const OpenImageIO_v2_4::basic_string_view<char> &, const OpenImageIO_v2_4::basic_string_view<char> &>' requested here
  return {FMT_FORWARD(args)...};
         ^
/oiio/build/include/OpenImageIO/detail/fmt/printf.h:620:24: note: in instantiation of function template specialization 'fmt::make_format_args<fmt::basic_printf_context<fmt::appender, char>, const OpenImageIO_v2_4::basic_string_view<char> &, const OpenImageIO_v2_4::basic_string_view<char> &>' requested here
                  fmt::make_format_args<context>(args...));
                       ^
/oiio/src/include/OpenImageIO/strutil.h:180:35: note: in instantiation of function template specialization 'fmt::sprintf<const char *, OpenImageIO_v2_4::basic_string_view<char>, OpenImageIO_v2_4::basic_string_view<char>, char>' requested here
    sync_output (stdout, Strutil::sprintf(fmt, args...));
                                  ^
/oiio/src/libutil/strutil_test.cpp:310:14: note: in instantiation of function template specialization 'OpenImageIO_v2_4::Strutil::printf<OpenImageIO_v2_4::basic_string_view<char>, OpenImageIO_v2_4::basic_string_view<char>>' requested here
    Strutil::printf ("escape '%s' <-> '%s'\n", raw, escaped);
             ^
1 error generated.
```

**Platform information:**
 - OIIO branch/version: v2.4.9.0
 - OS: Windows/macOS/Linux
 - C++ compiler: MSVC/Apple Clang/GCC
 - Any non-default build flags when you build OIIO: N/A

lgritz:
fmt 10.0 was recently released, and it breaks a lot for us because they have fully removed legacy support for automatically being able to format anything that has an iostream `<<` operator. This facility was deprecatd in 8.1 but could be re-enabled with FMT_DEPRECATED_OSTREAM, but even that was removed in 10.0.

So this means that we had to write several explicit format handlers for some types for use with format() and print().

Also, they are even more strict with sprintf/printf which don't honor the custom formatters as far as I can tell, so in several cases I also switched those to the std::format/print style.

Fixes #3835

### DIFF
--- a/src/include/OpenImageIO/detail/fmt.h
+++ b/src/include/OpenImageIO/detail/fmt.h
@@ -26,6 +26,7 @@
 
 // fmt 8.1 stopped automatically enabling formatting of anything that supports
 // ostream output. This breaks a lot! Re-enable this old behavior.
+// NOTE: fmt 10.0 removed this support entirely.
 #ifndef FMT_DEPRECATED_OSTREAM
 #    define FMT_DEPRECATED_OSTREAM 1
 #endif
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,7 +257,7 @@ jobs:
             cc_compiler: gcc-12
             cxx_compiler: g++-12
             cxx_std: 20
-            fmt_ver: 9.1.0
+            fmt_ver: 10.0.0
             openexr_ver: main
             pybind11_ver: master
             python_ver: "3.10"
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,7 +25,7 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * libjpeg >= 8, or libjpeg-turbo >= 1.1 (tested through jpeg9d and jpeg-turbo
    2.1)
  * Boost >= 1.53 (recommended: at least 1.66; tested through 1.81)
- * [fmtlib](https://github.com/fmtlib/fmt) >= 6.1.2 (tested through 9.1). If
+ * [fmtlib](https://github.com/fmtlib/fmt) >= 6.1.2 (tested through 10.0). If
    not found at build time, this will be automatically downloaded unless the
    build sets `-DBUILD_MISSING_FMT=OFF`.
 
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -319,7 +319,7 @@ endmacro()
 option (BUILD_FMT_FORCE "Force local download/build of fmt even if installed" OFF)
 option (BUILD_MISSING_FMT "Local download/build of fmt if not installed" ON)
 option (INTERNALIZE_FMT "Copy fmt headers into <install>/include/OpenImageIO/detail/fmt" ON)
-set (BUILD_FMT_VERSION "9.1.0" CACHE STRING "Preferred fmtlib/fmt version, when downloading/building our own")
+set (BUILD_FMT_VERSION "9.0.0" CACHE STRING "Preferred fmtlib/fmt version, when downloading/building our own")
 
 macro (find_or_download_fmt)
     # If we weren't told to force our own download/build of fmt, look
@@ -356,3 +356,8 @@ macro (find_or_download_fmt)
 endmacro()
 
 find_or_download_fmt()
+
+if (FMT_VERSION VERSION_EQUAL 90100
+        AND GCC_VERSION VERSION_GREATER 0.0 AND NOT GCC_VERSION VERSION_GREATER 12.2)
+    message (WARNING "${ColorRed}fmt 9.1 is known to not work with gcc <= 7.2${ColorReset}")
+endif ()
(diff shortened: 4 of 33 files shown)


src/include/OpenImageIO/detail/fmt.h
```diff
@@ -26,6 +26,7 @@
 
 // fmt 8.1 stopped automatically enabling formatting of anything that supports
 // ostream output. This breaks a lot! Re-enable this old behavior.
+// NOTE: fmt 10.0 removed this support entirely.
```
fpsunflower, line 29: How old of a version of fmt do we need to support?

Would it be possible to drop the "deprecated ostream" macro here now that you've implemented the missing items?

If I've understood correctly, the formatter functionality is not actually new, its just that we didn't implement the formatters before, right ?

lgritz: I don't know how far back we "need" to support. I think we can drop some old releases (in master / 2.5-to-be), but I'm not sure how far to push forward.

I'd need to do some archeology on fmt itself to see how far back it fully supports the format extensions (maybe all the way back, I'm just not sure), and also make a judgment call about whether dropping that macro might break anything people depend on -- for example, if people are using any of the utilities in strutil.h such as OIIO::print, format, or debug (which all wrap the fmt functionality), might they be subtly depending on this macro when dealing with types with ostream adapters but no formatter templates?

I want this PR to be fully backportable to solve the immediate problem of fmt 10.0 breaking our build. They we can separately tackle whether there is simplifications we can make in master but with some acceptable risk of breaking users of format/print if they haven't kept up with writing formatters.

fpsunflower: Makes sense. From my reading of the fmt release notes, the ostream_formatter which you are using here was only added in 9.0 (same time they deprecated the "automatic" `<<` handling). So I guess this is the oldest version we can support.

fpsunflower (approved):
LGTM!

PR #3850, opened 2023-05-25
admin: Add draft ASWF documents to the repo

lgritz:
Establish an ASWF subdirectory.

Add draft or placeholder CLAs, CoC, governance (including TSC member list), and project intake checklist.

Suggestions, amendments, and additional TSC members should be communicated via review comments on this PR (prior to merge) or further PRs against these documents (after the merge, while they are still in draft / unvoted form),

The CLAs and CoC are presumed final, since they are merely copies of the standard ones for ASWF. Project intake checklist replicated from OSL, will be replaced when a new one is given to us (but is presumably going to be very similar, so this will help us get started). Governance document is copied from OSL, and is for the OIIO TSC to review and vote on to arrive at a final form appropriate for this project. The technical charter is currently an empty placeholder, as we are still resolving questions about the proposed draft with LF.

Note that none of the new documents (including the CLAs) will become effective until they are approved, the project is officially and finally signed over to the ASWF, and the repo is moved to the ASWF GitHub account.


mdavies-laika:
Hey Larry,

You asked me to leave a comment to add me as a voting TSC member.

Martin

ASWF/CONTRIBUTING.md
```diff
@@ -0,0 +1,333 @@
...
+community, then it may be prudent to email oiio-dev pointing everybody to
+the pull request URL and discussing any issues you think are important.
+
+1. All pull requests automatically launch CI jobs on GitHub Actions to
+ensure that the build completes and that the tests suite runs correctly, for
+a variety of platform, compiler, library, and flag combinations. The status
+of the CI tests for your PR will be displayed on the GitHub PR page. We will
+not accept PRs that don't build cleanly or pass the existing testsuite.
+
+1. The reviewer will look over the code and critique on the "comments" area.
```
ThiagoIze, line 168: I think we should clarify if the submitter can be their own reviewer and if they can merge their own code into master. Are there some devs that should have this ability? Maybe the Chief Architect?

ASWF/CONTRIBUTING.md
```diff
@@ -0,0 +1,333 @@
...
+
+Mail List and Slack
+-------------------
+
+Contributors should be reading the oiio-dev mail list:
+
+* [oiio-dev](https://lists.aswf.io/g/oiio-dev)
+For developers of the OpenImageIO code itself, or users who are really
+interested in the OIIO internals. This is where we mostly discuss the code
+(including bug reports), but are also happy to answer user questions about
```
ThiagoIze, line 29: It says we will mostly discuss the code, including bug reports, on the email list. But then later under the Bug Reports section it says we report bugs in github. So this implies we report bugs in github and then do not discuss them there but must discuss on the email list? Why not keep it all in the github issue where it was reported? 

I have the same question for enhancements.

ASWF/CONTRIBUTING.md
```diff
@@ -0,0 +1,333 @@
...
+
+with enough detail that others can easily reproduce the problem just by
+following your instructions. Please quote the exact error message you
+received. If you are having trouble building, please post the full cmake
+output of a fresh VERBOSE=1 build.
+
+Suspected security vulnerabilities should be reported by the same process.
+
+If confidentiality precludes a public question or issue for any reason, you
+may contact us privately at [security@openimageio.org](security@openimageio.org).
```
ThiagoIze, line 78: Who is subscribed to security@openimageio.org? Should we mention who this is? Can anyone add themselves? Is it everyone on the TSC? A subset of the TSC based on whoever volunteers to subscribe to this?

ASWF/CONTRIBUTING.md
```diff
@@ -0,0 +1,333 @@
...
+
+and it will automatically reformat your code according to the configuration
+file found in the `.clang-format` file at the root directory of the OIIO
+source code checkout.
+
+One of the CI test matrix entries runs clang-format and fails if any
+diffs were generated (that is, if any of your code did not 100% conform to
+the `.clang-format` formatting configuration). If it fails, clicking on that
+test log will show you the diffs generated, so that you can easily correct
+it on your end and update the PR with the formatting fixes.
```
ThiagoIze, line 219: Different versions of clang-format can produce different output. Is it worth pointing this out in case folks hit this and don't understand why it's failing?

ASWF/CONTRIBUTING.md
```diff
@@ -0,0 +1,333 @@
...
+    void foo (const T& newfoo) { m_foo = newfoo; }
+    const T& foo () const { return m_foo; }
+
+Avoid multiple inheritance.
+
+Namespaces: yes, use them!
+
+### Third-party libraries
+
+Prefer C++11 `std` rather than Boost, where both can do the same task.
```
ThiagoIze, line 293: What about C++14, 17, etc.? If it's a newer C++, do we use boost for the older C++ and the `std` version for the newer (with #ifdefs)? Having that split would make it easier to drop the boost version once we drop support for the older C++.

lgritz:
Thiago has helpfully reread the CONTRIBUTING document more carefully than I had and calls out several areas that should be revisited either because they are ambiguous (and have been all along) or should be examined anew in light of new events such as ASWF adoption, the existence of a TSC, and the explicit desire to becomes a true community project.

I feel like those points deserve a full discussion and consensus, and in some cases maybe some brainstorming about exactly how to proceed and what policies we, as a group, would like to adopt for this project. There is no deadline for this -- the contribution and review process can be revised at any time, and I'm not aware of any other tasks that are gated on this being finalized (as they are for adoption of the charter or licensing). 

We're still waiting to receive votes from two TSC members (Lewis and Robin), but technically we have already received YES votes from a majority or members on all the questions put forward. I don't want to hold up the rest of the items, so here is what I propose:

* Accept the current set of documents, including the original CONTRIBUTING edits (which really just carry forward the document we've had all along for OIIO, with minor corrections that are essential, like explaining the DCO requirement).
* Separately continue to refine the CONTRIBUTING guidelines individually, as we will repeatedly do over the life of the project.

ASWF/CONTRIBUTING.md
lgritz: I'm going to merge what we have when we get the rest of the votes in (since technically it already has the majority approval). Then I'll make a separate PR that takes a stab at what I think I can do to tighten it up, but we should discuss to get consensus (in the next meeting, if we can't do it easily on the PR). 

But I will add some of my initial thoughts here to get the ball rolling. In all cases, this is not set in stone, these are just my ideas based on other projects and what I think will make things efficient and careful.

The reviewer should definitely be somebody other than the patch author. I think once reviewed, either the reviewer or the patch author can be the one to push the "merge" button. Ideally it should be the patch author in cases where that person has commit privileges, but when that's not the case, any onlooker with commit privileges can do the actual merge if they are satisfied that the review is sufficient.

I would also like to be stricter about every patch being reviewed, versus my current begrudgingly accepted, but necessary practice of merging my own patches without review if they sit for a few days without any comments. I don't want to do that anymore, but what I need to make that work is for the TSC members and committers to make a serious effort to step in with the reviews and not allow any PRs to fester without review.

A few exceptions should be noted: 
* Some kinds of low-risk changes that can't break anything (like changes to documentation or administrative files) or high-priority fixes to critical issues that provably fix something very broken (such as a change to the CI scripts that change us from failing to passing CI) are ok for a senior developer to merge before getting a review if integration is time sensitive as long as a PR was submitted and CI fully passes.
* Some changes are so extensive or fundamental (especially backward-incompatible changes to APIs or behavior) that it is prudent to get reviews from multiple people, or representatives of multiple companies. There are no formal guidelines for this, we should just watch out for situations where multiple people or more time to review would be helpful for big changes.
* Backports of a reviewed/approved/CI-passing patch from master to a release branch can be done by the chief architect or release manager without a separate PR/review cycle just for the backport.

ASWF/CONTRIBUTING.md
lgritz: The security email is specifically for people to report important security vulnerabilities so that we can fix them before they are discussed in public.

It's not a subscription mail list. Right now it only forwards to me. After the domain is transferred to ASWF, we will designate whoever we want (or several people) to receive it. It could be me, or Danny, or somebody else who volunteers to be the security front line responder. It should be somebody on the TSC because it's could involve sensitive information. In reality, it only gets email once every few months, usually from people who are doing fuzz testing and want CVEs entered for the problems they find.

Just recalling how we handle it for other projects, I think it's not necessary to document who specifically receives it, but we should clarify that it is only seen by one or two people on the TSC who are the minimal set of people who need to have access so they can fix critical vulnerabilities.

ASWF/CONTRIBUTING.md
lgritz: I'll edit this to spell it out, but you are correct -- at any given time, there is a *specific* version of clang-format that is the ultimate source of truth in case other clang-format versions contradict each other, and it is the version we use for the "clang-format" CI test. I tend to keep the version steady for long periods, and should only change by PR in which it is deliberately bumping the clang-format version being used and discussing the implications of that (and usually that would require checking in an assortment of other changes across the codebase for the small number of things that might change their formatting).

ASWF/CONTRIBUTING.md
lgritz: Haha, I think you found a paragraph that probably dates from when we first were able to use C++11 as a minimum. It should surely be updated. You are right, my preference is to use the boost features only when needed (including proper `#if` guards to select `std` things on a version-by-version basis -- as we do for std::filesystem). We should also mention that a long-term goal is to replace all the current boost use and rid ourselves of that dependency entirely, though there is no specific deadline for doing so.

lgritz:
We have received votes from all but one TSC member, and all documents and measures were approved (not just by majority, but unanimously except for the missing vote). I'll aim to merge this PR on July 1 -- nice round date of 1/2 way through the year -- which will officially put the project charter, new CLAs, CoC, and Governance documents into effect.

We will strive to set up the EasyCLA system as soon as possible so everybody's companies (where applicable) to get started signing the new CLAs, since at that point the old ones will no longer apply. We'll give some time, probably at least a few weeks and likely corresponding to when we actually make the final move of the repo itself to the AcademySoftwareFoundation GitHub account, before we turn on the mandatory EasyCLA enforcement, so there is some grace period before we actually are unable to merge approved PRs whose authors aren't registered in the CLA system.

lgritz:
The missing vote was received from Robin, he's been incommunicado because he's travelling but let me know by DM that he approves all the measures that were up for vote.

PR #3855, opened 2023-05-29
ci: Get icc tests working again

lgritz:
For a couple years, we installed the
intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic package from Intel's yum repo https://yum.repos.intel.com/oneapi to get a copy of icc and icx to test against.

This broke in January 2023 when Intel bumped the version of the package and the new package's icc (but curiously not icx) wouldn't run on the CentOS images and complained about

    /opt/intel/oneapi/compiler/2023.1.0/linux/bin/intel64/icpc: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /opt/intel/oneapi/compiler/2023.1.0/linux/bin/intel64/icpc)

In https://github.com/OpenImageIO/oiio/pull/3744, we tried to combat this by locking down to a specific version of icc 2022.1 that was known to work and that we used to use until the default version was bumped.

But in April 2023, the 2022.1.0-3768 package seemed to disappear from the repository entirely, so we're back to being unable to install icc and test against it from any of the ASWF VFX Platform compliant containers. So in this patch, we disable that part of the CI testing.

However, with some tinkering, I was able to use that icc package from the plain Ubuntu 22.04 environment (just not from the VFX Platform compliant container). So set that up so we're still testing icc somehow.

icx has been fine all along, this doesn't change anything about how we test icx, which is still done in the VFX Platform 2022 container.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,20 +94,22 @@ jobs:
             simd: "avx2,f16c"
             fmt_ver: 8.1.1
             pybind11_ver: v2.8.1
-          - desc: icc/C++17 py3.9 boost1.76 exr3.1 ocio2.1 qt5.15
-            nametag: linux-vfx2022-icc
-            os: ubuntu-latest
-            vfxyear: 2022
-            cxx_std: 17
-            python_ver: 3.9
-            # simd: "avx2,f16c"
-            fmt_ver: 7.1.3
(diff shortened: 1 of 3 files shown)


lgritz:
Abandoning this PR for now, I have a simpler solution that I will submit separately.

PR #3858, opened 2023-05-30
raw input: fix user_flip usage

mugulmd:
## Description

When reading a raw image, setting the `user_flip` to 1 meant ignoring the `flip` value (to avoid buffer rotation from LibRaw).

However the value 1 for the `user_flip` is **not neutral**: values lie between 0 and 7 (see https://www.libraw.org/docs/API-datastruct-eng.html#libraw_image_sizes_t), and 1 corresponds to a horizontal mirror (in my understanding of the source code).

The proposed fix is to use 0 as `user_flip` value for ignoring the `flip` value.

This modification invalidates the changes made in this PR: https://github.com/OpenImageIO/oiio/pull/3847.
Therefore, flip to orientation conversion has been adapted to match exactly LibRaw's documentation, which handles only rotations (+90, +180, -90 degrees).

## Tests

The previous PR was only tested on images of landscape and human faces, hence the mirroring issue remained unnoticed.
When reading images containing text however the problem is clearly visible.
This PR fixes this mirroring issue on the following 4 orientations: 0, +90, +180 and -90 degrees rotation.
Other orientations could not be tested as they don't seem to ever appear on raw images datasets.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -436,20 +436,13 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // Store flip before it is potentially overwritten
     // LibRaw's convention for flip values differs from Exif orientation tags
     // so we need to convert it
-    int original_flip                 = 1;
-    int libraw_flip                   = m_processor->imgdata.sizes.flip;
+    int original_flip = 1;
+    int libraw_flip   = m_processor->imgdata.sizes.flip;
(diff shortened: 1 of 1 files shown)


mugulmd:
> LGTM, and thanks for the fix. Before I merge, can you please fix the formatting as indicated by examining the output of the failed "clang-format" test in CI? Thanks.

Done 😄

lgritz (approved):
LGTM, thanks

PR #3862, opened 2023-06-02
If OpenEXR is found by cmake config file, OpenEXR_VERSION is set instead of OPENEXR_VERSION

bebuch:
## Description

I use CMake with [`CMAKE_FIND_PACKAGE_PREFER_CONFIG`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html). OpenEXR installs its own config files which set `OpenEXR_VERSION` instead of `OPENEXR_VERSION`.

## Tests

It would be usefull to add a test pipline where CMake is called with `CMAKE_FIND_PACKAGE_PREFER_CONFIG`. However this is out of scope to this PR and should be done by a separate patch. Unfortunately, I don't currently have the time to familiarize myself enough to do this myself.

## Checklist:

- [ ] (not applicable) If this is more extensive than a small change to existing code, I
- [ ] (not applicable) I have updated the documentation, if applicable.
- [ ] (not applicable) I have ensured that the change is tested somewhere in the testsuite

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -108,7 +108,7 @@ include_directories(BEFORE ${IMATH_INCLUDES} ${OPENEXR_INCLUDES})
 if (MSVC AND NOT LINKSTATIC)
     add_definitions (-DOPENEXR_DLL) # Is this needed for new versions?
 endif ()
-if (OPENEXR_VERSION VERSION_GREATER_EQUAL 3.0)
+if (OpenEXR_VERSION VERSION_GREATER_EQUAL 3.0)
     set (OIIO_USING_IMATH 3)
 else ()
     set (OIIO_USING_IMATH 2)


bebuch:
@lgritz I just want to let you know that I'm still going to take care of this PR and #3863. Just didn't get to it this week. But it should happen durring the next 2 weeks.

bebuch:
I updated the patch. The Find Moduke seams to provide both `OpenEXR_VERSION` and `OPENEXR_VERSION`, so we can just use `OpenEXR_VERSION` in the check.

lgritz:
@bebuch In the time since you first submitted this, we have switched to accepting all new code under Apache-2.0 license (changing from our historical use of BSD-3-Clause) and also enabled use of DCO sign-off.

Would you mind please acknowledging here (a comment in this PR is sufficient) that you are ok submitting under Apache-2.0?

And also, could you please do the sign-off so it passes the DCO check? The easiest way to do that is:

    git commit --amend -s             # the -s will add the sign-off line
    git push origin patch-2 --force

bebuch:
I'm fine with submitting under Apache-2.0 license.

lgritz (approved):

lgritz (approved):

PR #3864, opened 2023-06-06
fix(exr): try to improve exr thread pool weirdness

lgritz:
There's some code I put in not long ago to automatically call exr's thread pool setNumThreads(0) upon exit, as an attempt to work around some exr thread pool deadlock conditions. But... we've seen some weirdness from this workaround as well, so try to limit its application by not bothering when using OpenEXR >= 3.1.8 (at which point we believe the exr side bug was fixed) and also to do so only on Windows.


lgritz:
@olegul Can you try this patch on your end (combined with building against OpenEXR 3.1.8) and see if your crash goes away?

PR #3866, opened 2023-06-07
ImageInput and ImageOutput docs updated to Python 3.

### LINKED ISSUE #3846: [DOC] ImageOutput: Writing Images wrong Python example code
**Describe the bug**
In the **Image Output Made Simple** section: https://openimageio.readthedocs.io/en/latest/imageoutput.html#image-output-made-simple

The example Python code simply doesn't work.

**To Reproduce**

```python
# main.py

import OpenImageIO as oiio
import numpy as np

filename = "foo.jpg"
xres = 640
yres = 480
channels = 3  # RGB
pixels = np.zeros((yres, xres, channels), dtype=np.uint8)

out = oiio.ImageOutput.create (filename)
if out is None:
    return
spec = ImageSpec(xres, yres, channels, 'uint8')
out.open (filename, spec)
out.write_image (pixels)
out.close ()
```

```shell
$ python main.py

 File "<cwd>\main.py", line 13
    return
    ^^^^^^
SyntaxError: 'return' outside function
```

Also `ImageSpec` is not imported from `oiio` so it doesn't work.

ziggycross:
## Description

Fixes #3846 by updating Python examples.

Current code examples for [ImageInput](https://openimageio.readthedocs.io/en/latest/imageinput.html) and [ImageOutput](https://openimageio.readthedocs.io/en/latest/imageoutput.html) raise a SyntaxError in Python 3:
```python
File "<cwd>\main.py", line 13
    return
    ^^^^^^
SyntaxError: 'return' outside function
```

It seems these code examples might be for Python 2.7. Perhaps it would be worth creating an extra tab instead of replacing the old code?

## Tests

Tested on Python 3.11, no issues.

## Checklist:

### DIFF
--- a/src/doc/imageinput.rst
+++ b/src/doc/imageinput.rst
@@ -38,14 +38,13 @@ memory, even if that's not the way they're stored in the file):
         import numpy as np
 
         inp = ImageInput.open(filename)
-        if inp is None :
-            return
-        spec = inp.spec()
-        xres = spec.width
-        yres = spec.height
-        channels = spec.nchannels
-        pixels = inp.read_image("uint8")
-        inp.close()
+        if inp :
+            spec = inp.spec()
+            xres = spec.width
+            yres = spec.height
+            channels = spec.nchannels
+            pixels = inp.read_image("uint8")
+            inp.close()
 
 Here is a breakdown of what work this code is doing:
 
--- a/src/doc/imageoutput.rst
+++ b/src/doc/imageoutput.rst
@@ -44,12 +44,11 @@ to a file:
        pixels = np.zeros((yres, xres, channels), dtype=np.uint8)
 
        out = oiio.ImageOutput.create (filename)
-       if out is None:
-           return
-       spec = ImageSpec(xres, yres, channels, 'uint8')
-       out.open (filename, spec)
-       out.write_image (pixels)
-       out.close ()
+       if out:
+           spec = oiio.ImageSpec(xres, yres, channels, 'uint8')
+           out.open (filename, spec)
+           out.write_image (pixels)
+           out.close ()
 
 This little bit of code does a surprising amount of useful work:
 


lgritz (approved):
Thanks for the fix!

lgritz:
I think that it's not so much a "python 2 vs 3" thing, as it is:

(a) The examples in the docs are program *fragments*, not literal full working programs. So 'return' would be fine if this code were inside a function, but doesn't make sense in the main script body.

(b) In many cases, I started out with C++ (also fragments, not compilable complete C++ programs) and hand-translated them to make the python equivalents. But I didn't always test them, so probably made plenty of little mistakes, compounded by the fact that I don't use Python every day like I do C++, so my Python can be a little un-fluent, especially when I'm not testing it.

In the longer run, I think it would be interesting to set up some kind of fully automated scheme in which most of the sample code in the docs gets run as testsuite items. Or alternately, the other way around -- that they live entirely in the testsuite and get "included" into the docs when code examples are needed.

ziggycross:
You're totally right - I don't have much experience with Python 2 and suspected it might be a compatibility issue. But after looking into it, it doesn't seem the return keyword was much different in v2, so that makes sense if it was just un-tested code.

I don't know much about testsuite but would be happy to see what needs to be done to link it up with the docs.

lgritz:
I have no doubt that there are *also* lots of examples in the docs that I wrote in the python2 days and are not valid python3.

I bet if I try out what I'm talking about with the testsuite/docs connection on ONE example, then with that existing as a template, it should be fairly easy to convert a lot of the rest of the code examples in the docs to be like that.

lgritz:
I meant, if I do one, it should be easy for OTHER PEOPLE to do the same thing to all the other examples.

ziggycross:
> I have no doubt that there are _also_ lots of examples in the docs that I wrote in the python2 days and are not valid python3.
> 
> I bet if I try out what I'm talking about with the testsuite/docs connection on ONE example, then with that existing as a template, it should be fairly easy to convert a lot of the rest of the code examples in the docs to be like that.

If you do this, I'd be happy to copy the rest over, at least for the Python examples.

PR #3870, opened 2023-06-08
fix(simd.h): Fix sense of hiding deprecated type names

lgritz:
I got the test backwards, and it was incorrectly hiding the definition for the default case of OIIO_DISABLE_DEPRECATED == 0.


cmstein:
I haven't tested it locally, but LGTM!

PR #3877, opened 2023-06-12
fix(psd): Prevent simultaneous psd thumbnail reads from clashing

### LINKED ISSUE #3824: [BUG] ASAN heap-use-after-free while concurrently opening PSD files
**Describe the bug**
See below for the full ASAN output but it seems like the ImageCache might not be threadsafe in this particular instance maybe?

**To Reproduce**
1. Concurrently .open() the 8 .PSD files from the `oiio-images` repo
2. ASAN will complain while processing the resources in the `PSDInput::load_resource_thumbnail` code path (relating to the `ImageCache` and copying of `ImageSpec`s)

**Expected behavior**
No crash

**Evidence**
[oiio-asan.txt](https://github.com/OpenImageIO/oiio/files/11342734/oiio-asan.txt)

If I disable threading within the application, so it only loads 1 .PSD at a time, things seem ok.

I haven't been able to nail down the cause. Any ideas? I can attempt a stand alone repro sometime tomorrow maybe and will try to help out as best I can.

**Platform information:**
 - OIIO branch/version: 2.4.9
 - OS: Linux
 - C++ compiler: 11.3

lgritz:
When reading a PSD file, if it contains a thumbnail, we read it from the in-memory blob with a combination of an IOMemReader and an ImageBuf.  But... we always named it "thumbnail.jpg" without considering that multiple simultaneous PSD files reading identically named images (but which are in fact different) could cause a weird kind of clashing in any underlying use of ImageCache!

The simple fix is to use ordinary ImageInput approach to reading the thumbnail from the IOMemReader, but just store it in the ImageBuf.

Fixes #3824

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -1344,9 +1344,27 @@ PSDInput::load_resource_thumbnail(uint32_t length, bool isBGR)
     if (!ioread(&jpeg_data[0], jpeg_length))
         return false;
 
+    // Create an IOMemReader that references the thumbnail JPEG blob and read
+    // it with an ImageInput, into the memory owned by an ImageBuf.
     Filesystem::IOMemReader thumbblob(jpeg_data.data(), jpeg_length);
-    m_thumbnail = ImageBuf("thumbnail.jpg", 0, 0, nullptr, nullptr, &thumbblob);
-    m_thumbnail.read(0, 0, true);
+    m_thumbnail.clear();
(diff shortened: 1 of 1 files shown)


lgritz:
@jessey-git are you able to try this proposed fix on your end and let us know if it appears to solve the problem?

jessey-git:
@lgritz Yes, I can confirm my small repro app works fine after the change here.  I can't quite verify inside Blender itself as building the libraries there is giving me some issues.  That said the fix seems sound by itself and I don't anticipate further problems as the stack traces I was seeing in user reports matches what the test app was hitting.

Issue #3851, opened 2023-05-26
[BUG] thread_pool shutdown dead lock


LazyDodo:
**Describe the bug**
 
Occasionally idiff.exe refuses to shutdown on the blender CI environment, previously openexr was to blame here but with recent developments there that specific problem appears to be solved, however a new one has sprung up in its place. This one will be a bit more annoying since i've been unable to reproduce it on my local environment sofar, best I have is a process dump from the CI environment. which shows the following stack with only the main thread still being alive. 

```
ntdll.dll!NtYieldExecution()
KERNELBASE.dll!SwitchToThread()
[Inline Frame] OpenImageIO_Util.dll!std::this_thread::yield() Line 179
[Inline Frame] OpenImageIO_Util.dll!OpenImageIO_v2_4::atomic_backoff::operator()() Line 146
OpenImageIO_Util.dll!OpenImageIO_v2_4::spin_mutex::lock() Line 219
[Inline Frame] OpenImageIO_Util.dll!std::unique_lock<OpenImageIO_v2_4::spin_mutex>::{ctor}(OpenImageIO_v2_4::spin_mutex &) Line 137
[Inline Frame] OpenImageIO_Util.dll!OpenImageIO_v2_4::pvt::ThreadsafeQueue<std::function<void __cdecl(int)> *>::pop(std::function<void __cdecl(int)> * &) Line 63
OpenImageIO_Util.dll!OpenImageIO_v2_4::thread_pool::Impl::clear_queue() Line 178
OpenImageIO_Util.dll!OpenImageIO_v2_4::thread_pool::Impl::stop(bool isWait) Line 249
[Inline Frame] OpenImageIO_Util.dll!OpenImageIO_v2_4::thread_pool::Impl::{dtor}() Line 122
OpenImageIO_Util.dll!OpenImageIO_v2_4::thread_pool::Impl::`scalar deleting destructor'(unsigned int)
OpenImageIO_Util.dll!`OpenImageIO_v2_4::default_thread_pool'::`2'::`dynamic atexit destructor for 'shared_pool''()
ucrtbase.dll!<lambda>(void)()
ucrtbase.dll!__crt_seh_guarded_call<int>::operator()<<lambda_7777bce6b2f8c936911f934f8298dc43>,<lambda>(void) &,<lambda_3883c3dff614d5e0c5f61bb1ac94921c>>()
ucrtbase.dll!_execute_onexit_table()
[...]

PR #3882, opened 2023-06-13
API: Add shutdown method

LazyDodo:
## Description

This change adds a `shutdown` method to the API that can be used to perform shutdown of the thread-pool(s) (or other clean-up activities). Applications that utilize OpenImageIO must call it before exiting their main function to prevent possible deadlocks during application shutdown (see #3851)

This change also updates the documentation and all oiio-tools to call the new `shutdown` method.

This is the final change for fixing #3851

## Tests

No tests added / unsure how to test for it.

## Checklist:

  [corporate](https://github.com/OpenImageIO/oiio/blob/master/src/doc/CLA-CORPORATE)). - Minimal change, no CLA required


LazyDodo:
An update to #3805 to make it not use `atexit` is still forthcoming but not required to mark #3851 as fixed, hence not included.

lgritz (changes requested):
Code LGTM, but let's move the docs to the existing API chapter as I noted inline.

lgritz:
Should we do anything about Python? Expose OIIO::shutdown() through the python bindings? Or should we ensure that when python module unloads, it will call shutdown?

Is there harm in calling shutdown twice()? In calling anything that uses the thread pool after shutdown()? (Does it automatically set it up again?) 

What are the instructions we should give to a library author who uses OIIO and isn't sure if, at runtime, the app itself uses OIIO, or knows the library uses OIIO, so the app would call shutdown? Or what if the app uses other libraries that use OIIO but each one isn't sure if they are the only one?

LazyDodo:
I'll have to admit : not a python guy. 

If i had to guess I'm guessing when python shuts down it unloads all modules before existing causing the regular dtor for the threadpool to run long before the process exists so all threads still be alive? so likely not a problem? But I admit wildly guessing here, would have to check with the debugger or sprinkle some printf's to validate this theory.

lgritz:
Oh, yes, you're probably right. I guess this is only a problem when the process ends naturally but terminates so quickly that the threads are killed before the pool has finished releasing?

Let's move forward on this and we can always amend later if there's another edge case we haven't considered.

lgritz (approved):
LGTM

PR #3887, opened 2023-06-16
fix(IBA): Improve error message for IBA::ocio functions

lgritz:
If colorconvert, ociolook, ociodisplay, ociofiletransform fail
because OCIO support is not enabled, make sure the error message
makes that clear.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -1989,8 +1989,15 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
             if (colorconfig->error())
                 dst.errorfmt("{}", colorconfig->geterror());
             else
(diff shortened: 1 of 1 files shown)


jessey-git:
Just double-checking: There doesn't seem to be any other use of the `#ifdef OCIO` define right now.  Did you mean `#ifdef USE_OCIO` which is used in some other source files?

lgritz:
Oh my, you are correct. Thanks for spotting that.

jessey-git:
Is it worth changing the `ociolook` and `ociodisplay` functions in the same file wrt this error message too?

Beyond that, the change works for me and looks good.

lgritz:
Oh, yeah, you're probably right about that.

lgritz:
This is why we have code review! Thanks for making me make it better, @jessey-git

jessey-git:
LGTM!

PR #3892, opened 2023-06-22
ci: Simplify build_llvm.bash script

### LINKED ISSUE #3891: [BUG] openImageIO/2.4.12.0/src/build-scripts/build_llvm.bash has bad pull URL
**Describe the bug**
Would appear there's been things moved on GitHub that the llvm build script needs and now it fails to download thus making extraction fail.

**To Reproduce**
1. Run `src/build-scripts/build_llvm.bash`
2. Script executes:
```bash
curl --location https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-error.tar.xz -o clang+llvm-13.0.0-x86_64-linux-gnu-error.tar.xz
```
**Expected behavior**
Need this to build as expected.

**Evidence**
Visiting the URL provided in bash script shows 404 on GitHub.

Error log:
```bash
tar: This does not look like a tar archive
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```
**Platform information:**
 - OIIO 2.4.12.0
 - OS: Ubuntu
 - C++ compiler: GCC

lgritz:
* Remove leftover logic from when we used TravisCI.
* Pick a default for LLVM_DISTRO_NAME so it works when the script is run standalone when not on GitHub Actions CI. (But it's still really only meant to run on x86 Ubuntu, and not designed for anything useful outside of our CI.)
* Eliminate all the complexity with old LLVM versions -- in practice, our CI only uses this to download fairly modern clang versions.

Fixes #3891

### DIFF
--- a/src/build-scripts/build_llvm.bash
+++ b/src/build-scripts/build_llvm.bash
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
 # Utility script to download and build LLVM & clang
+#
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: BSD-3-Clause
+# https://github.com/OpenImageIO/oiio
 
 # Exit the whole script if any command fails.
 set -ex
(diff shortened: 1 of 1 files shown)


lgritz:
N.B.: The broken tests are entirely unrelated to this change and have been occurring on the nightly CI builds for a couple days, apparently linked to a dependency repo change. I have reported the problem and hope it is fixed soon.

PR #3895, opened 2023-06-25
ci: Get rid of long-unused install_test_images.bash

lgritz:


lgritz:
Does this seem ok to everybody? I'm trying to avoid doing merges without a single review.

PR #3896, opened 2023-06-26
testing: Fix missing simd test due to copy paste typo

lgritz:


lgritz:
Does this seem ok to everybody? I'm trying to avoid doing merges without a single review.

ThiagoIze (approved):
This is replacing a duplicate test run with a test that wasn't previously being run. Looks good to me.

PR #3897, opened 2023-06-26
docs: Document use of the DCO which is now required for all PRs

lgritz:


lgritz:
Merging without review because it was already approved by TSC and enabled, and this is merely documenting what contributors need to do to affix the DCOs to their commits.

PR #3898, opened 2023-06-26
Remove redundant instructions from tile hash

curtisblack:
## Description

Optimises the code in the tile hash function. During a render this function is called billions of times so this small improvement adds up, in some test scenes it reduces total render time by an average of about 2%.

See https://godbolt.org/z/jMzz4PTce for the exact details. A summary would be:

- Reduced instruction count from 75 to 61, this is due to the redundant add and shift operations that can be implicitely handled by taking advantage of the member layout in the class.
- Halves the number of memory fetch instructions because all accesses are now 64-bit.
- Removes branching instructions and extra memory indirection inside `ustring::hash()`.

There is a slight change in behaviour, which is that the hash is now computed from the `m_file` pointer rather than from `m_file->filename().hash()`, so runtime reproducibility is lost. However the performance gain seems worth it.

## Checklist:

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -560,13 +560,17 @@ struct TileID {
     /// Digest the TileID into a size_t to use as a hash key.
     size_t hash() const
     {
-        const uint64_t a = (uint64_t(m_x) << 32) + uint64_t(m_y);
-        const uint64_t b = (uint64_t(m_z) << 32) + uint64_t(m_subimage);
-        const uint64_t c = (uint64_t(m_miplevel) << 32)
-                           + (uint64_t(m_chbegin) << 16) + uint64_t(m_chend);
-        const uint64_t d = m_file->filename().hash()
-                           + uint64_t(m_colortransformid);
-        return fasthash::fasthash64({ a, b, c, d });
+        static constexpr size_t member_size
+            = sizeof(m_x) + sizeof(m_y) + sizeof(m_z) + sizeof(m_subimage)
+              + sizeof(m_miplevel) + sizeof(m_chbegin) + sizeof(m_chend)
+              + sizeof(m_colortransformid) + sizeof(m_padding) + sizeof(m_file);
+        static_assert(
+            sizeof(*this) == member_size,
+            "All TileID members must be accounted for so we can hash the entire class.");
+        static_assert(
+            sizeof(*this) % sizeof(uint64_t) == 0,
+            "FastHash uses the fewest instructions when data size is a multiple of 8 bytes.");
+        return fasthash::fasthash64(this, sizeof(*this));
     }
 
     /// Functor that hashes a TileID
@@ -590,6 +594,7 @@ private:
     int m_miplevel;            ///< MIP-map level
     short m_chbegin, m_chend;  ///< Channel range
     int m_colortransformid;    ///< Colorspace id (0 == default)
+    int m_padding = 0;         ///< Unused
     ImageCacheFile* m_file;    ///< Which ImageCacheFile we refer to
 };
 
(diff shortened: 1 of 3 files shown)


src/libtexture/imagecache_pvt.h
```diff
@@ -590,6 +594,7 @@ struct TileID {
     int m_miplevel;            ///< MIP-map level
     short m_chbegin, m_chend;  ///< Channel range
     int m_colortransformid;    ///< Colorspace id (0 == default)
+    int m_padding = 0;         ///< Unused
```
fpsunflower, line 597: This should probably be evaluated separately, but maybe instead of adding 32-bits of padding, `m_miplevel` and `m_colortransformid` could maybe be changed to 16-bit to shave off an extra 8 bytes?

I'd be at least curious to know if that saves any measurable amount of time.

curtisblack: Yeah I didn't necessarily wish to bundle that type of change into this PR but I also thought something similar. 

Unfortunately I don't have a build with all the recent color space additions to test performance. The performance was measured on v2.4.6, prior to the addition of `m_colortransformid`.

lgritz: We can save this for another pass, but it seems like a good idea to me.

lgritz (approved):
LGTM, great idea.
I will merge when CI finishes.

lgritz:
@curtisblack Please look at the failed CI test, which implicates this code. The one test that fails is when using gcc12, presumably triggering a new warning that didn't exist in older gcc versions. It's behaving as if there is an uninitialized field in the TileID struct, but looking at your code I don't see why that would be the case.  Can you please take a close look and see if you can convince yourself whether the warning is correct (and fix) or if it's some kind of false positive (in which case you should probably use some kind of pragma to locally disable it where it is causing trouble, so that we can continue to pass CI).

curtisblack:
Thanks for checking @lgritz, as far as I can tell it does appear to be a false positive that only seems to affect GCC. Clang doesn't see this as a warning or error. I will look into locally disabling this specific warning for that part of the code.

See https://godbolt.org/z/5q7Y7ndfb for a comparison with clang.

It might also be showing in that CI build because it enables `-Werror`, The bug does also seem to affect earlier versions of GCC but without the `-Werror` flag they pass CI.

I did some additional testing, using the `testtex --testhash` tool (and increasing the test iterations to 1e9 for a sturdy comparison) this PR increases the hash rate from 361 to 495 Mhash/sec on my machine.

src/libtexture/imagecache_pvt.h
```diff
@@ -570,7 +570,12 @@ struct TileID {
         static_assert(
             sizeof(*this) % sizeof(uint64_t) == 0,
             "FastHash uses the fewest instructions when data size is a multiple of 8 bytes.");
+        // This appears to be a false positive which only affects GCC.
+        // https://godbolt.org/z/5q7Y7ndfb
+        OIIO_PRAGMA_WARNING_PUSH
+        OIIO_GCC_ONLY_PRAGMA(GCC diagnostic ignored "-Wuninitialized")
```
lgritz, line 576: Still failing. Based on the error message, I think this might need to be `-Wmaybe-uninitialized`?

lgritz:
I think I have the right fix, @curtisblack -- the pragma needs to be in hash.h, immediately surrounding the line where it thinks there is a problem (and the other spots were not needed, and I think also were turning off the wrong warning). I've taken the liberty of pushing my fix on top of yours. We'll see, if it passes CI then I'll merge the combined patch. Cross fingers...

lgritz:
Relevant tests passed, warning squashed, so I merged!

PR #3901, opened 2023-06-30
fix(strutil.h): ensure proper constexpr of string hashing

lgritz:
The underlying farmhash algorithm (which is the basis for Strutil::strhash, and thus ustring's hash representation) turned out to not be properly constexpr under certain circumstances and needed to be expressed a different way to be fully constexpr on all platforms and for all strings.

New implementation suggested by Alex Wells.

Also, make sure string_view ctr from char* can be constexpr, even for
C++14. This is related for the sake of the strhash calls that, when
taking string literals, actually choose the variety that accepts a
string_view.

### DIFF
--- a/src/include/OpenImageIO/detail/farmhash.h
+++ b/src/include/OpenImageIO/detail/farmhash.h
@@ -239,21 +239,77 @@ STATIC_INLINE uint32_t Fetch32(const char *p) {
 
 #else
 
-template <typename T>
-STATIC_INLINE T FetchType(const char *p) {
-  T result = 0;
-  for (size_t i = 0; i < sizeof(T); i++)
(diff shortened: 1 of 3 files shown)


lgritz:
@AlexMWells, can you confirm if this totally fixes your problem?

AlexMWells:
Yes it does fix the issue and allow global constexpr variables and consteval expressions.  Thanks for adding a test by the way.

AlexMWells (approved):
LGTM

lgritz:
Updated: Also, make sure string_view ctr from char* can be constexpr, even for
C++14. This is related for the sake of the strhash calls that, when
taking string literals, actually choose the variety that accepts a
string_view.

PR #3903, opened 2023-07-01
admin: update licensing

lgritz:
Since LG has re-licensed all his old code under Apache-2.0, we can change the license notice from BSD to Apache-2.0 for all files in which LG is the only extant author.

Some files that LG authored (mostly test scripts) had no license notice, so in those cases, we added one.


lgritz:
Amended: Also change all files that LG co-authored with other people to specify "BSD-3-Clause and Apache-2.0".

lgritz:
Does this look ok, @grdanny ?

PR #3904, opened 2023-07-06
admin: Copyright and license notice updates

lgritz:
* Add copyright/license notices to some mixed-authorship files that lacked them before (all related to tests)
* Get rid of the pointless year in many copyright notices (change to the current LF recommended wording)


grdanny (approved):
looks good!

PR #3907, opened 2023-07-11
Update ImageInput docs to not use deprecated APIs

jessey-git:
## Description

Update the ImageInput documentation to not use the deprecated `read_image` APIs.

Based on the discussion in #3902

## Checklist:


jessey-git:
I tried to keep each sample relatively similar to each other and changed a few uses of non-standard dynamic arrays to just use vector.  

I've never used DCO before. Do I just edit the above message with the special text or do I need to do something else before I use the special text?

lgritz:
The easiest way to correct this is:

    git commit --amend -s
    git push --force origin deprecateddocs

The `-s` (which you could have used on any commit) appends the "Signed-off-by" line to your commit message.

lgritz:
https://github.com/OpenImageIO/oiio/blob/master/CONTRIBUTING.md

Read the section about DCO signoff. There's also a tool you can use (per repo) that will automatically sign all your commits, if you are like me and will forget otherwise.

lgritz:
This in particular: https://github.com/coderanger/dco

src/doc/imageinput.rst
```diff
@@ -360,10 +362,11 @@ flexible functionality.  A few representative examples follow:
 
 * Flip an image vertically upon reading, by using *negative* ``y`` stride::
 
-    unsigned char pixels[spec.width * spec.height * spec.nchannels];
+    std::vector<unsigned char> pixels(spec.width * spec.height * spec.nchannels);
```
lgritz, line 365: Just as an FYI, if I were writing this for real, I think the best idiom would be:

    auto pixels = std::unique_ptr<unsigned char[]>(new unsigned char [spec.width * spec.height * spec.nchannels]);

That would be freed upon scope exit (like std::vector), but it's much more efficient -- new[] doesn't initialize the memory, but std::vector will fill it with zeroes unnecessarily!

jessey-git: How about I change them to the following. Seem ok?
`auto pixels = std::make_unique<unsigned char[]>(spec.width * spec.height * spec.nchannels);`

lgritz: Will that really work without the `new`?

jessey-git: Yes, make_unique does work like that, however it looks like the semantics of make_unique include value initialization. After trying it out on compiler explorer the make_unique calls will include some form of memset basically (rep stosq)... c++ why are you the way you are...  https://godbolt.org/z/9PEYKbY6f

I'll replace all of them with your suggestion then.

lgritz (approved):

PR #3911, opened 2023-07-14
build: Fix broken OIIO_NO_NEON definition

### LINKED ISSUE #3909: [BUG] neon support is disabled on macOS arm
Commit 22a8196c8f5a77a767cac36de595bcd07afb885d  does `#define OIIO_NO_NEON 1` whenever `__aarch64__` is set so a macOS arm build ends up with no neon support. By the commit message and comments, I don't think that was the intent.

Stuff still works. I only found out because I have an assertion in my code that some buffers provided by OSL should be 16 byte aligned and they weren't.

lgritz:
Clean up the `#if` tests that disable Intel SIMD when not on Intel architectures, which incorrectly disabled ARM NEON when on ARM.

Fixes #3909

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -95,6 +95,9 @@
 #  ifndef OIIO_NO_AVX2
 #    define OIIO_NO_AVX2 1
 #  endif
+#endif
+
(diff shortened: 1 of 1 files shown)


lgritz:
@1-Olivier reports that this fixes the reported problem

PR #3918, opened 2023-07-21
CMYK PSD files should copy alpha

jasonbaumeister:
## Description

During the CMYK to RGB conversion in PSDInput, the alpha channel was never copied to the RGBA output. Converting a PSD to JPEG worked because it ignored all the 0 alpha values, but converting to WebP would result in a fully transparent image.

## Tests

I added a cmyk-with-alpha.psd test to the psd-colormodes test to check this

## Checklist:


lgritz:
Need the DCO sign-off on these commits as well.

jasonbaumeister:
> Need the DCO sign-off on these commits as well.

Should all be squashed and signed now.

lgritz (approved):
LGTM, thanks!

PR #3919, opened 2023-07-21
Adding IOProxy support to ICOInput

jasonbaumeister:
## Description

This updates ICOInput to use the IOProxy model so it can read from files or the IOMemReader

## Tests

The current ICO tests still pass exercising the new code. I didn't see anything in the other formats with IOProxy support to explicitly test the in memory options

## Checklist:

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -816,6 +816,24 @@ Windows.  ICO files use the file extension :file:`.ico`.
      - int
      - if nonzero, will cause the ICO to be written out using PNG format.
 
+**Configuration settings for ICO input**
+
+When opening an ICO ImageInput with a *configuration* (see
+Section :ref:`sec-input-with-config`), the following special configuration
+attributes are supported:
+
+.. list-table::
+   :widths: 30 10 65
+   :header-rows: 1
+
+   * - Input Configuration Attribute
+     - Type
+     - Meaning
+   * - ``oiio:ioproxy``
+     - ptr
+     - Pointer to a ``Filesystem::IOProxy`` that will handle the I/O, for
+       example by reading from memory rather than the file system.    
+
 **Configuration settings for ICO output**
 
 When opening an ICO ImageOutput, the following special metadata tokens
--- a/src/ico.imageio/icoinput.cpp
+++ b/src/ico.imageio/icoinput.cpp
@@ -25,6 +25,10 @@ public:
     ICOInput() { init(); }
     ~ICOInput() override { close(); }
     const char* format_name(void) const override { return "ico"; }
+    int supports(string_view feature) const override
+    {
+        return feature == "ioproxy";
+    }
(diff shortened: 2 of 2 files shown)


lgritz:
Can you please add the DCO sign-off to your patch? (`git commit --amend -s`, then re-push ought to do it)

jasonbaumeister:
> Can you please add the DCO sign-off to your patch? (`git commit --amend -s`, then re-push ought to do it)

Done.

lgritz:
Sorry, it needs to be on each commit. You can squash them all down and then make sure it's on the one, or you can edit each commit message to have the "Signed-off-by: " field.

jasonbaumeister:
> Sorry, it needs to be on each commit. You can squash them all down and then make sure it's on the one, or you can edit each commit message to have the "Signed-off-by: " field.

Sorry about that, should all be fixed now.

lgritz (approved):
LGTM, thanks for these fixes!

I took the liberty of pushing an additional change that adds the docs for this. After it passes CI on that, I'll merge it.

lgritz (changes requested):
Wait, I just realized, you forgot something -- both ICOInput and ICOOutput need implementations of `supports()` that will respond correctly to a query about IOProxy. Look in, say, dpxinput.cpp for an example.

lgritz:
As for your comment about testing -- the existing imageinout_test will try IOProxy and MemReader for every format that says it supports IOProxy.  (In fact, starting to type this reply to you is how I remembered to check that you had done that, and saw that you didn't! Once you declare that it supports IOProxy, the test should catch anything you've done wrong in the implementation.)

jasonbaumeister:
supports() should be implemented for ICOInput now.

I took a quick look at supporting IOProxy for ICOOutput as well and it looks like the only real catch is that it needs to be able to open a file in 'r+b' instead of 'wb' that ImageOutput::ioproxy_use_or_open() always uses.

I can add a new value to the Filesystem::IOProxy::Mode enum for ReadWrite and add a mode param to ImageOutput::ioproxy_use_or_open, defaulted to Filesystem::IOProxy::Mode::Write and have that working, if that sounds like the correct approach to you.

lgritz:
How about this: Let's merge the ICO input proxy support on its own, and tackle the output proxy separately. We need to take a hard look at whether append is really needed here or it's just a quirk of the implementation from 13 years ago (supporting evidence: I don't think any of the other format writers did this), or if is truly needed, make sure that the operation we do on IOProxy is not introducing any bugs or weird semantics.

lgritz:
Uh oh, it looks like now that you marked it as supports("ioproxy"), the imageinout_test is failing.

jasonbaumeister:
imageinout_test is passing now. I took a closer look at how BMPInput worked and the ioseek(0) before the first read was required.

lgritz (approved):
LGTM, thanks!

PR #3920, opened 2023-07-22
build: bump build_openexr and build_opencolorio defaults to latest versions

lgritz:


ThiagoIze (approved):
CI passes, and these are bigger numbers, so looks good to me.

PR #3924, opened 2023-07-23
build(deps): Raise minimum CMake dependency from 3.12 to 3.15

lgritz:
This bump brings access to the following interesting features (not a comprehensive list):

- 3.13  Nov 21, 2018 

    - -B, -S  let you specify source and build directories without the "cd" dance 
    - target_link_options() 
    - ctest --progress / CTEST_PROGRESS_OUTPUT

- 3.14  Mar 14, 2019 

    - cmake --build now supports -v/--verbose 
    - file(CREATE_LINK ) 
    - FindGIF now has imported targets

- 3.15  Jul 17, 2019 

    - env CMAKE_GENERATOR can be used instead of -G 
    - when using --build, the --target can now list multiple targets (and also can be `-t`) 
    - --install can be used separately without running through the whole build 
    - message() now supports NOTICE, VERBOSE, DEBUG and TRACE. 
    - CMAKE_FIND_PACKAGE_PREFER_CONFIG variable 
    - Enable C++20 for Apple Clang

Make sure all CI cases have a new enough cmake

Use GIF targets, now enabled with CMake 3.14

Use message(VERBOSE ...) feature to simplify lots of `if` blocks.

Clean up code for older cmake versions

Note that this will NOT be backported to the current release branch,
but it will be part of the upcoming 2.5 family of reseases.


lgritz:
We discussed this a couple TSC meetings ago and everybody thought it was fine, so I'm going to merge it. (2.5 only, won't be backported)

PR #3925, opened 2023-07-24
refactor: Convert iconvert.cpp stream io and sprintf to modern

lgritz:
Also fix a typo

### DIFF
--- a/src/iconvert/iconvert.cpp
+++ b/src/iconvert/iconvert.cpp
@@ -20,7 +20,7 @@
 
 
 using namespace OIIO;
-
+using OIIO::Strutil::print;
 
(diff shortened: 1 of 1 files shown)


src/iconvert/iconvert.cpp
```diff
@@ -20,7 +20,7 @@
 
 
 using namespace OIIO;
-
+using OIIO::Strutil::print;
```
jessey-git, line 23: Probably should add in a `#include <OpenImageIO/strutil.h>` and remove the now unnecessary `<iostream>` include.

jessey-git (commented):
One minor comment

jessey-git (approved):
LGTM

PR #3905, opened 2023-07-07
admin: Relicense code under Apache 2.0

lgritz:
To help the effort to re-license this project under Apache 2.0, we are asking all historical contributors to OpenImageIO to please re-license their prior work to us under the new license. You still retain full ownership, we are not asking for a copyright assignment. Only that the prior code you licensed under BSD, that now you let us know it's fine to use Apache 2.0. I have already done this myself for all the code that I own.

If this is acceptable to you,

* Add a **COMMENT** to this PR that says "all of my contributions originally submitted to this project under the 3-Clause BSD License are hereby relicensed to the Apache License, Version 2.0, and are submitted pursuant to the Developer Certificate of Origin, version 1.1."

If you were working at the time for an employer who owned your IP output or you submitted the code under the auspices of your company's CLA to the project, we'll also need a similar grant from somebody who can re-submit on behalf of the company. That can be you, if you are still at that company, as long as you check that relicensing under Apache 2.0 is ok with the company and they authorize you to do so. But if you're not authorized to re-license any commits owned by the company, please alert us to the situation so that we -- or better, you -- can contact somebody there to provide that piece.

Thank you so much, we would really appreciate it if you could do us this very quick favor.

PR #3926, opened 2023-07-24
admin: Relicense code under Apache 2.0

lgritz:
Per PR #3905 (https://github.com/OpenImageIO/oiio/pull/3905), update the
RELICENSING document with the names of people and companies who have
signed on to the relicensing thus far, and update the SPDX license
notices on source files whose extant authorship (per 'git blame')
comprises only authors who have relicensed.

This is part 2, the original PR is ongoing and we will continue to
update as additional people relicense.


ThiagoIze (approved):
I have not double checked that every file has the correct license attributed with it, but the rest of the changes do look good to me.

lgritz:
I don't expect anybody to easily double check it. But the way this works is that Danny wrote some scripts that a few weeks back produced the list of the extant authors of every file. As each person relicenses, I get rid of their lines from the list. When there are no more unlicensed names associated with a file, I change its license to just Adobe 2.0. I also have a spreadsheet where I'm keeping track of all the authors and how many total lines they each have, and which people have relicensed. That's how I'm computing the percentages.

PR #3928, opened 2023-07-24
build(deps): Raise OpenEXR minimum from 2.3 to 2.4

lgritz:
This will not be backported to the release branch, it's only for OIIO 2.5 and later.


ThiagoIze (approved):
Arnold has been using 2.4 for a long time, so this is fine for us. It's also not hard to upgrade from 2.3 to 2.4. I haven't reviewed the patch itself, so this approval is only for the idea behind this PR.

PR #3929, opened 2023-07-24
ci: Lock down icx version

lgritz:
Intel's servers have again updated the default icx that you get when you install from their yum repository, and the new one seems incompatible with our libstdc++ toolchain. So lock down on 2023.1.0 for now so that our CI can pass.


lgritz:
Merging because it fixes broken CI.

PR #3931, opened 2023-07-28
Fix arithmetic overflow in oiio_bufinfo (Python interop)

jessey-git:
## Description

When an incoming 1-dimensional, buffer-like, object was used in Python, the corresponding check inside OIIO would overflow if its length exceeded a signed 32bit value.

An error similar to `Pixel data array error: Python array shape is [2304000000] but expecting h=24000, w=24000, ch=4` would be raised.

I modified a few similar calculations and tried to follow the predominant pattern already used.

## Tests
Tested using the following snippet which would previously fail:
```
import OpenImageIO as oiio
import numpy as np

filename = "t:/foo.png"
xres = 24000
yres = 24000
channels = 4  # RGB
pixels = np.zeros(xres * yres * channels, dtype=np.uint8)

spec = oiio.ImageSpec(xres, yres, channels, "uint8")
image = oiio.ImageOutput.create(filename)
image.open(filename, spec)
if not image.write_image(pixels):
    print("error = ", image.geterror())
image.close()
```

## Checklist:

### DIFF
--- a/src/python/py_oiio.cpp
+++ b/src/python/py_oiio.cpp
@@ -148,7 +148,8 @@ oiio_bufinfo::oiio_bufinfo(const py::buffer_info& pybuf, int nchans, int width,
             }
         } else if (pybuf.ndim == 1
                    && pybuf.shape[0]
-                       == int64_t(width) * int64_t(height) * int64_t(nchans)) {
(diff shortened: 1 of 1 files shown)


lgritz (approved):
Good catch, thanks

Issue #3932, opened 2023-07-31
Text with shadow rendering 


progweb:
As render text with shadow with channel alpha, I don't get the same result that without channel alpha.

Without channel alpha, shadow option permit to draw a dilated text in black, whereas with channel alpha, OIIO makes a gap...

Working with dev-2.2 branch, I can see the same issue on master branch.

[Note]: OpenImageIO is used in the [gpx2video project](https://github.com/progweb/gpx2video).

progweb:
```
From bec7340a06844588d5364f444033624bfe8d8559 Mon Sep 17 00:00:00 2001
From: Nicolas VIVIEN <nicolas@vivien.fr>
Date: Mon, 31 Jul 2023 13:44:22 +0200
Subject: [PATCH] Fix shadow text rendering on alpha buffer

---
 src/libOpenImageIO/imagebufalgo_draw.cpp | 13 +++++++++++--
 1 file changed, 11 insertions(+), 2 deletions(-)

diff --git a/src/libOpenImageIO/imagebufalgo_draw.cpp b/src/libOpenImageIO/imagebufalgo_draw.cpp
index 376f66451..18dc7c527 100644
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -930,7 +930,10 @@ ImageBufAlgo::render_text(ImageBuf& R, int x, int y, string_view text,
         // If the buffer doesn't have an alpha, but the text color passed
         // has 4 values, assume the last value is supposed to be alpha.
         textalpha = textcolor[3];
+		alpha_channel = 3;
     }
+    } else
+		alpha_channel = -1;
 
     // Convert the UTF to 32 bit unicode
     std::vector<uint32_t> utext;
@@ -1009,8 +1012,14 @@ ImageBufAlgo::render_text(ImageBuf& R, int x, int y, string_view text,
         float val   = t[0];
         float alpha = a[0] * textalpha;
         R.getpixel(r.x(), r.y(), pixelcolor);
-        for (int c = 0; c < nchannels; ++c)
-            pixelcolor[c] = val * textcolor[c] + (1.0f - alpha) * pixelcolor[c];
+		if (alpha == 0.0)
+			continue;
+        for (int c = 0; c < nchannels; ++c) {
+			if (c == alpha_channel)
[...]

PR #3934, opened 2023-07-31
fix: text rendering did not work correctly with alpha

progweb:
As render text with shadow with channel alpha, I don't get the same result that without channel alpha.

Without channel alpha, shadow option permit to draw a dilated text in black, whereas with channel alpha, OIIO makes a gap...

Working with dev-2.2 branch, I can see the same issue on master branch.

In using this patch, shadow option works as expected on image with alpha buffer.

Ref: #3932 

[Note]: OpenImageIO is used in the [gpx2video project](https://github.com/progweb/gpx2video).

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -1078,7 +1078,9 @@ ImageBufAlgo::render_text(ImageBuf& R, int x, int y, string_view text,
         // If the buffer doesn't have an alpha, but the text color passed
         // has 4 values, assume the last value is supposed to be alpha.
         textalpha = textcolor[3];
-    }
+        alpha_channel = 3;
+    } else
+        alpha_channel = -1;
 
     // Convert the UTF to 32 bit unicode
     std::vector<uint32_t> utext;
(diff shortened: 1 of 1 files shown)


progweb:
orig

progweb:
After:

patch

src/libOpenImageIO/imagebufalgo_draw.cpp
```diff
@@ -1162,8 +1164,14 @@ ImageBufAlgo::render_text(ImageBuf& R, int x, int y, string_view text,
...
         R.getpixel(r.x(), r.y(), pixelcolor);
-        for (int c = 0; c < nchannels; ++c)
-            pixelcolor[c] = val * textcolor[c] + (1.0f - alpha) * pixelcolor[c];
+        if (alpha == 0.0)
+            continue;
+        for (int c = 0; c < nchannels; ++c) {
+            if (c == alpha_channel)
+                pixelcolor[c] = alpha + (1.0f - alpha) * pixelcolor[c];
+            else 
+                pixelcolor[c] = (val * alpha * textcolor[c]) + (1.0f - alpha) * pixelcolor[c];
```
lgritz, line 1173: You may have noticed CI failures. Basically, every test with text in it is failing with this PR.

I think this line is the part that's responsible. Compared to the old code, you have an extra multiplication of the foreground color by alpha. I believe (but admit that I am not 100% sure) that it was correct before because freetype is supplying color results that are already scaled by the glyph's alpha coverage, so now you are *doubly* accounting for alpha coverage in the foreground.

progweb:
Not available, I’ll fix this in few days for the next week.

lgritz:
This PR has gone for years with review comments unaddressed. Closing. Feel free to resubmit if it's still a problem and somebody has a solution.

PR #3939, opened 2023-08-03
dev(ustring.h): string literal operator for ustring and ustringhash

lgritz:
* Add string literal operators for ustring and ustringhash so you can write: ``` "foo"_us "bar"_ush ``` Note that on Cuda (but not on the host), this ustringhash construction is constexpr.

* Add a new `ustringhash(const char*, size_t)` constructor, which is also constexpr on the device side.

* Mark the existing `ustringhash(string_view)` constructor as constexpr on the device side.

We can mark those things as constexpr on the device side because on the Cuda side, the implementation only takes the hash (which is constexpr), and does not actually add a ustring to the table (which cannot be constexpr). To make such declarations easy, add a new macro to platform.h, `OIIO_DEVICE_CONSTEXPR` which is constexpr for the device, merely inline for the host.

### DIFF
--- a/src/include/OpenImageIO/ustring.h
+++ b/src/include/OpenImageIO/ustring.h
@@ -818,18 +818,24 @@ public:
     OIIO_DEVICE_CONSTEXPR explicit ustringhash(const char* str)
     {
 #ifdef __CUDA_ARCH__
-        m_hash = Strutil::strhash(str);  // GPU: just compute the hash
+        // GPU: just compute the hash. This can be constexpr!
+        m_hash = Strutil::strhash(str);
 #else
-        m_hash = ustring(str).hash();       // CPU: make ustring, get its hash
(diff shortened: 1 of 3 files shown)


jessey-git (approved):
Seems straightforward and LGTM.

src/include/OpenImageIO/ustring.h
```diff
@@ -815,23 +815,32 @@ class OIIO_UTIL_API ustringhash {
     }
 
     /// Construct a ustringhash from a null-terminated C string (char *).
-    OIIO_HOSTDEVICE explicit ustringhash(const char* str)
+    OIIO_DEVICE_CONSTEXPR explicit ustringhash(const char* str)
```
AlexMWells, line 818: Maybe add a comment on why the CPU side can't be constexpr.

lgritz: done

PR #3940, opened 2023-08-08
build: suppress Xcode warnings

antond-weta:
## Description

When building the code with Xcode ("cmake -G Xcode"), there are 1000+ warnings about implicit 64 to 32 bit conversion.
This change adds a compile option to suppress such warnings when generating an Xcode project.

## Tests

None required

## Checklist:

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -148,6 +148,9 @@ if (CMAKE_COMPILER_IS_CLANG OR CMAKE_COMPILER_IS_APPLECLANG)
         # Don't warn about using unknown preprocessor symbols in `#if`
         add_compile_options ("-Wno-expansion-to-defined")
     endif ()
+    if (CMAKE_GENERATOR MATCHES "Xcode")
+        add_compile_options ("-Wno-shorten-64-to-32")
+    endif ()
 endif ()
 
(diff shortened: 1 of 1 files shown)


lgritz (approved):

lgritz:
I know I already approved and merged this, but...

Do you have any explanation for why this only was a warning for xcode, and not any of the other compilers (including clang on Apple, when not having CMake generate xcode project files)?

antond-weta:
That particular warning is enabled by default in Xcode ("Implicit Conversion to 32 Bit Type" in Xcode project settings), so it has to be explicitly disabled in the project file.

lgritz:
Are we confident that it's not a warning we should be addressing -- i.e., disabling the warning could lead to bugs and we should really aim to be fixing the places where the warning occurs?

antond-weta:
> Are we confident that it's not a warning we should be addressing -- i.e., disabling the warning could lead to bugs and we should really aim to be fixing the places where the warning occurs?

That would be good. Actually, I might just give it a go and try fixing as many as I can.

lgritz:
I guess the question is -- are there really thousands of separate places that are just too painful to fix? Or is it in a header and fixing it there will simultaneously fix a whole lot of the warnings?

PR #3942, opened 2023-08-13
exr: Re-enable openexr core default when recent enough

lgritz:
We had disabled the use of the "openexr core" library by default, even for OpenEXR versions that nominally support it, when it was discovered that it stil didn't support the DWA compression types.

That has finally been added to the OpenEXR trunk as well as the OpenEXR 3.1.10 release. When we are building against a sufficiently new OpenEXR to fully support all the compression types for exr core, then re-enable it by default.

Bump the "latest" CI tests to OpenEXR 3.1.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,7 +256,7 @@ jobs:
             cxx_compiler: g++-11
             cxx_std: 17
             fmt_ver: 9.1.0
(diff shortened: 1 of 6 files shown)


lgritz:
Does this seem ok to everybody?

grdanny (approved):
lgtm

PR #3944, opened 2023-08-17
Silence gcc new/delete warnings for texturesys

Shootfast:
## Description

Disable GCC warnings around mismatched new/delete for shared_texturesys creation

## Checklist:

### DIFF
--- a/src/libtexture/texturesys.cpp
+++ b/src/libtexture/texturesys.cpp
@@ -82,9 +82,9 @@ static const OIIO_SIMD4_ALIGN vbool4 channel_masks[5] = {
 TextureSystem*
 TextureSystem::create(bool shared, ImageCache* imagecache)
 {
-// Because the shared_texturesys is never deleted (by design)
-// we silence the otherwise useful compiler warning on newer GCC versions
-OIIO_PRAGMA_WARNING_PUSH
+    // Because the shared_texturesys is never deleted (by design)
+    // we silence the otherwise useful compiler warning on newer GCC versions
+    OIIO_PRAGMA_WARNING_PUSH
 #if OIIO_GNUC_VERSION > 100000
     OIIO_GCC_ONLY_PRAGMA(GCC diagnostic ignored "-Wmismatched-new-delete")
 #endif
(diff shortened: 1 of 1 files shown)


src/libtexture/texturesys.cpp
```diff
@@ -82,6 +82,12 @@ static const OIIO_SIMD4_ALIGN vbool4 channel_masks[5] = {
 TextureSystem*
 TextureSystem::create(bool shared, ImageCache* imagecache)
 {
+// Because the shared_texturesys is never deleted (by design)
+// we silence the otherwise useful compiler warning on newer GCC / Clang versions
+#pragma GCC diagnostic push
+#if __GNUC__ > 10
+#    pragma GCC diagnostic ignored "-Wmismatched-new-delete"
+#endif
```
lgritz, line 90: We have macros defined in platform.h for these things. In the parlance we use through the rest of the code base, this would be

```c++
OIIO_PRAGMA_WARNING_PUSH
#if OIIO_GNUC_VERSION > 100000
    OIIO_GCC_ONLY_PRAGMA(diagnostic ignored "-Wmismatched-new-delete")
#endif
...
OIIO_PRAGMA_WARNING_POP
```

I admit that for this small example, this is not especially any more clear than the raw pragmas you used. But the point is that it'll always work properly, no future non-gcc compiler will complain about those GCC pragma (which I am kinda superstitious about, even though CI shows it appears to be fine on the compilers we test).

Shootfast: Awesome, I've updated the pull request to use the existing macros.
I also edited the comment as this now only applies to GCC (as your macros explicitly single out that compiler). Usually clang respects gcc pragmas, but I don't have a clang toolchain easily available, and I guess the existing CI builders haven't complained so far

lgritz:
I took the liberty of pushing a fix to the clang-format CI complaints. If this passes CI, I will merge.

Shootfast:
One day I will remember to run the clang-format pass first :sweat_smile:

lgritz:
No prob. It was tricky in this case because I guess clang-format is ok with the `#if` and `#pragma` being in the first column, so it passed clang-format your first time through. But after changing to the macros as I suggested, clang-format thought it should be indented like regular code, and started complaining. 

One day I (or hopefully somebody else?) will figure out how to do a special GitHub action that will take a failed CI clang-format test and automatically turn it into an appended comment on the PR with the suggested diff of changes that can be merged in a single click. Actually, it appears there are a number of such implementations in the GitHub Action marketplace, the main task is probably to figure out which one suits our needs.

PR #3954, opened 2023-08-22
fix: various protections against corrupted files

### LINKED ISSUE #3952: AddressSanitizer: allocator is out of memory
version: 2.4 12.0
./bin/iconvert crash_000122 1.png
```
=================================================================
==121851==ERROR: AddressSanitizer: allocator is out of memory trying to allocate 0x3b7f84804 bytes
    #0 0x4c9e6d in operator new[](unsigned long) (/root/oiio/build/bin/iconvert+0x4c9e6d)
    #1 0x7f22ca2b4b9d in OpenImageIO_v2_4::ImageOutput::copy_image(OpenImageIO_v2_4::ImageInput*) /root/oiio/src/libOpenImageIO/imageoutput.cpp:587:36
    #2 0x4d169b in convert_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /root/oiio/src/iconvert/iconvert.cpp:449:27
    #3 0x4cd218 in main /root/oiio/src/iconvert/iconvert.cpp:523:14
    #4 0x7f22c87f2082 in __libc_start_main /build/glibc-SzIz7B/glibc-2.31/csu/../csu/libc-start.c:308:16

==121851==HINT: if you don't care about these errors you may set allocator_may_return_null=1
SUMMARY: AddressSanitizer: out-of-memory (/root/oiio/build/bin/iconvert+0x4c9e6d) in operator new[](unsigned long)
==121851==ABORTING
```
poc:
[crash_000122.zip](https://github.com/OpenImageIO/oiio/files/12390418/crash_000122.zip)

lgritz:
Fixes #3952

Several interrelated fixes to guard against corrupted files:

* Extend the `limits:imagesize_MB` sanity check to Targa files to address #3952 specifically, and check in that failing example file into the testsuite. (N.B.: Only a couple readers use this, TIFF and now Targa. it should be extended to the others. Mackerel!)

* Improve the error message used for the above, both for Targa and TIFF.

* For low-memory situations, perhaps the default of 32GB is too large. Make the default be the minimum of 32GB and 1/2 the size of physical memory on the machine. (An app that legit expects humongous files is expected to raise the limit appropriately.)

* The generic "Read error: hit end of file" we sometimes issued from `ImageInput::ioread()` was hard to discern which reader it came from, in cases where the file extension was absent or misleading and a succession of files were tried. By changing this error message to say which format reader triggered the error, it might make these errors easier to track down in the future. This also changed the reference output of a few test involving truncated or corrupted files.


ThiagoIze (commented):
I like the more informative error messages and extra checking. As I mentioned in https://github.com/OpenImageIO/oiio/issues/3952#issuecomment-1687374721, I'm not sold on half the physical memory limitation since some users might be running with older machines (laptops?) that don't have that much memory.

lgritz:
Yeah, this is just a proposal. If we don't like the physical_memory trick, we don't have to do it. But I want to find a way to address the fuzz testing with the sanitizer enabled, which when it crashes will be reported and we'll need to investigate with high priority under threat of CVEs being filed.

lgritz:
Would it be bad manners to use the min(physical_mem) trick only if the sanitizer is running? Can we even detect that? Feels kinda dirty, I don't want to mask legit bugs in a way that they won't be detected properly by the sanitizer. But I also don't want the sanitizer catching false positives that aren't actually problems.

jessey-git:
For the targa code check, and thinking ahead to the other formats, is there a way to extract the check+error message to some place common? Instead of several lines of boilerplate added to each format, only a one line call to the common method would be needed.

Tough call on the default size really. I can't come up with a reasonable value to satisfy both the fuzzing and small-device scenarios.

lgritz:
@jessey-git Yeah, for ImageOutput, there's a `ImageOutput::check_open()` method that provides a central place to do validity checks on a number of things. We could do something similar for ImageInput to centralize these checks. I'd like to do that in a separate PR, though, it's likely to be more intrusive.

ThiagoIze (approved):
OK, you have me convinced that most users with small amounts of RAM are unlikely to be operating on giant images.

ThiagoIze (approved):

Issue #3916, opened 2023-07-17
[FEATURE REQUEST] Print build-time warnings for LGPL gotchas


lgritz:
Background:

All the code in the OIIO repository itself (even parts borrowed from other projects) are compatible permissive licenses like BSD, Apache-2.0, MIT, or public domain. Those primarily obligate downstream recipients to include their copyright notices and properly attribute them, which of course we do.

Among the dependencies -- that is, not things we distribute in the OIIO repo, but things that a downstream builder of OIIO might link against -- are two LGPL-licensed libraries, ffmpeg and libheif.

It is our understanding that dynamically linking (i.e., .dll or .so) against these is also compatible, in the sense that those libraries themselves would be LGPL, but all the parts on the OIIO side or any other app that dynamically links against them, can still be redistributed with their original licenses (including closed source or commercial licenses in the case of an app). However, if you *statically* link against either ffmpeg or libheif libraries and redistribute the results, then we believe you only have the right to do so using LGPL on *all* components that are statically linked together, including OIIO and any surrounding app.

## Feature request 1

Modify the OIIO CMake build system so that if the ffmpeg or libheif libraries are found (and enabled), and the *static* library versions are being used, at build time we print an extra warning (`message(STATUS "...")`) explaining that because that particular library is linked statically, redistributing the resulting app is only permitted if the whole thing is under LGPL terms. This warning does not need to be printed if it's the dynamic versions of the library that are found and used, or, of course, if those dependencies are not found or used at all.

## Feature request 2

Also, ffmpeg may *optionally*, in a way that is beyond our control, have been built to either include or exclude certain GPL components. I'm not sure if there's even a way for us to tell this at OIIO build time, but when ffmpeg is found, it would probably be helpful if our build system printed a message to remind users to be sure they know how their copy of ffmpeg was built and make sure that if they redistribute OIIO linked against it, to be aware of what it takes to comply with ffmpeg's license terms.

---

I think that OIIO is already fully in compliance with all these obligations, but these two build-time messages (when the conditions are met) will help downstream users and builders ensure that they, too, are fully in compliance with the licenses of those dependencies that they are combining OIIO with when they build it. We can't really patrol what happens (and it only matters if the downstream users *redistribute* those executable artifacts), but I think doing this means we have truly given them all the information they need to do it by the book.

---

[...]

PR #3958, opened 2023-08-24
Print build-time warnings for LGPL gotchas

grdanny:
Trying out my first Mackerel from:

https://github.com/OpenImageIO/oiio/issues/3916

### DIFF
--- a/src/ffmpeg.imageio/CMakeLists.txt
+++ b/src/ffmpeg.imageio/CMakeLists.txt
@@ -3,6 +3,28 @@
 # https://github.com/OpenImageIO/oiio
 
 if (FFmpeg_FOUND)
+    if (LINKSTATIC)
+        set (_static_suffixes .lib .a)
+        set (_static_libraries_found 0)
+
+        foreach (_ffmpeg_library IN LISTS FFMPEG_LIBRARIES)
+            get_filename_component (_ext ${_ffmpeg_library} LAST_EXT)
+            list (FIND _static_suffixes ${_ext} _index)
+            if (${_index} GREATER -1)
+                MATH (EXPR _static_libraries_found "${_static_libraries_found}+1")
+            endif()
+        endforeach()
+
+        if (${_static_libraries_found} GREATER 0)
+            message (STATUS "${ColorYellow}")
+            message (STATUS "You are linking OpenImageIO against a static version of FFmpeg, which may have")
+            message (STATUS "LGPL and possibly GPL licensed components (depending on exactly how your copy")
+            message (STATUS "of FFmpeg was built). If you intend to redistribute this build of OpenImageIO,")
+            message (STATUS "we recommend that you review the FFmpeg build flags and licensing terms.")        
+            message ("${ColorReset}")
+        endif()
+    endif()
+
     add_oiio_plugin (ffmpeginput.cpp
                      INCLUDE_DIRS ${FFMPEG_INCLUDES}
                      LINK_LIBRARIES ${FFMPEG_LIBRARIES}
--- a/src/heif.imageio/CMakeLists.txt
+++ b/src/heif.imageio/CMakeLists.txt
@@ -3,6 +3,28 @@
 # https://github.com/OpenImageIO/oiio
 
 if (Libheif_FOUND)
+    if (LINKSTATIC)
+        set (_static_suffixes .lib .a)    
+        set (_static_libraries_found 0)
+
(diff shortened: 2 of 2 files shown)


src/ffmpeg.imageio/CMakeLists.txt
```diff
@@ -3,6 +3,34 @@
...
+
+    if (${_static_libraries_found} GREATER 0)
+     message("${ColorRed}")
+     message(STATUS "Looks like you are linking against some static ffmpeg libraries: ${FFMPEG_LIBRARIES}")
+     message(STATUS "ffmpeg is a LGPL-licensed library")
+     message(STATUS "When linking statically against a LGPL-licensed library your entire project needs to be under LGPL")
+     message("${ColorReset}")
+    endif()
+
+   endif()
```
lgritz, line 30: I like the approach. Danny, can you please clean up the indentation -- they don't always align, and some blocks intend 1 space, some 2, some 4? Let's use a consistent 4 like the rest of the cmake files.

src/heif.imageio/CMakeLists.txt
```diff
@@ -3,6 +3,33 @@
 # https://github.com/OpenImageIO/oiio
 
 if (Libheif_FOUND)
+   if (LINKSTATIC)
+    if (WIN32)
+        set (_static_suffixes .lib .a)
+    else ()
+        set (_static_suffixes .a)
+    endif ()
```
lgritz, line 11: Can this be simplified? If you just do .lib and .a, and not need the `if`, so what? It won't find .lib in the libraries on non-Windows, so it doesn't hurt to search for it.

src/heif.imageio/CMakeLists.txt
```diff
@@ -3,6 +3,33 @@
...
+       if (${_index} GREATER -1)
+         MATH(EXPR _static_libraries_found "${static_libraries_found}+1")
+       endif()
+    endforeach()
+
+    if (${_static_libraries_found} GREATER 0)
+     message("${ColorRed}")
+     message(STATUS "Looks like you are linking against some static libheif libraries")
+     message(STATUS "libheif is a LGPL-licensed library")
+     message(STATUS "When linking statically against a LGPL-licensed library your entire project needs to be under LGPL")
```
lgritz, line 26: Let's omit the prescriptive legal advice and go with a simpler informational message that leaves the problem on their doorstep:

```
You are linking OpenImageIO against a static version of libheif, which is LGPL
licensed. If you intend to redistribute this build of OpenImageIO, we recommend
that you review the libheif license terms, or you may wish to switch to using a
dynamically-linked libheif.
```

I think this gives them the information they need to point them in the right direction, and avoids letting them stumble unaware into possible trouble, but without trying to tell them exactly what they can and can't do (which, as non-lawyers, we should not be doing and could be wrong about).

A slightly different message more accurately covers the ffmpeg situation:

```
You are linking OpenImageIO against a static version of FFmpeg, which may have
LGPL and possibly GPL licensed components (depending on exactly how your copy
of FFmpeg was built). If you intend to redistribute this build of OpenImageIO,
we recommend that you review the FFmpeg build flags and licensing terms.
```

lgritz (changes requested):
I think you're on the right track to fully solve this issue. I recommended a few minor changes.

src/ffmpeg.imageio/CMakeLists.txt
```diff
@@ -3,6 +3,34 @@
...
+    foreach(_ffmpeg_library IN LISTS FFMPEG_LIBRARIES)
+       get_filename_component(_ext ${_ffmpeg_library} LAST_EXT)
+       list (FIND _static_suffixes ${_ext} _index)
+       if (${_index} GREATER -1)
+         MATH(EXPR _static_libraries_found "${static_libraries_found}+1")
+       endif()
+    endforeach()
+
+    if (${_static_libraries_found} GREATER 0)
+     message("${ColorRed}")
```
lgritz, line 23: Let's use ColorYellow here. It's not a build break, just a warning.

grdanny:
This should be ready for review.
The only additional thought I had was maybe to turn this into a macro and share across the two cases (and maybe then it's useful in some other similar future case?)
But not sure if that is more helpful or confusing, right now it should be pretty clear why we added this.

lgritz (approved):

lgritz:
I think it's fine as it is, Danny. These are the only two instances of this that we know about. If another pops up, we can think about whether to refactor into some common infrastructure.

lgritz:
Fixes #3916

PR #3963, opened 2023-08-26
fix: normalize cleanup

lgritz:
* Add documentation for oiiotool --normalize

* Add tests for oiiotool --normalize

* Revamp existing normalize tests in testsuite/python-imagebofalgo: Get rid of old test images and refs, because their values were already normalized, so was probably not testing the math.  Make new tests that are much smaller, 3x3 images that contain specific (non-normalized) values to verify they all get normalized corectly. They'll run a lot faster, too.


lgritz:
@ssh4net How does this look to you?

ssh4net (approved):
For tests that you understand better than me, i trust your solution.
For rest everything looks fine.,

PR #3965, opened 2023-08-27
build(win): Fix windows.h pre-definitions

### LINKED ISSUE #3964: Redefinition of NOMINMAX causes warning in Windows
In platform.h for Windows, there is a definition of NOMINMAX that causes warnings when it is already defined.

```
//Avoid min and max being defined
#if defined(_WIN32)
#define NOMINMAX
#endif
```
Furthermore, right above it is another definition of NOMINMAX correctly wrapped by an ifdef

```
#if defined(_WIN32)
#    ifndef WIN32_LEAN_AND_MEAN
#        define WIN32_LEAN_AND_MEAN
#    endif
#    ifndef VC_EXTRALEAN
#        define VC_EXTRALEAN
#    endif
#    ifndef NOMINMAX
#        define NOMINMAX
#    endif
#    ifndef NOGDI
#        define NOGDI
#    endif
#    include <windows.h>
#endif


```

lgritz:
* Avoid possible double define of NOMINMAX in platform.h
* Move other Windows symbol definitions into the cmake file instead of
  repeating them in certain .cpp files.

Fixes #3964

### DIFF
--- a/src/include/OpenImageIO/platform.h
+++ b/src/include/OpenImageIO/platform.h
@@ -42,9 +42,11 @@
 #    include <intrin.h>
 #endif
 
-//Avoid min and max being defined
-#if defined(_WIN32)
-#define NOMINMAX
+// Avoid min and max being defined for any subsequent include of windows.h
+#ifdef _WIN32
+#    ifndef NOMINMAX
+#        define NOMINMAX
+#    endif
 #endif
 
 #include <OpenImageIO/oiioversion.h>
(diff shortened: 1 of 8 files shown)


jessey-git:
The changes to `platform.h` look good.

For the others inside the `libutil` project, you _could_ consider doing the following in cmake, rather than pollute the source files with any of the #defines:
https://github.com/OpenImageIO/oiio/blob/master/src/libutil/CMakeLists.txt#L37
`add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX -DNOGDI -DVC_EXTRALEAN)`

Up to you though. The current changes are also ok.

lgritz:
Oh, good suggestion. Yes, I will do that.

jessey-git (approved):

PR #3967, opened 2023-08-28
feat(api): ImageInput::open_check

lgritz:
Not unlike the ImageOutput::open_check we added a few months back, this creates an ImageInput utility function that centralizes a number of validation checks. Currently, this is primarily a check on allowed resolution and total size, aimed at making a guess at files that are corrupted or malicious.

Also adds a new `supports("noimage")` query.

### DIFF
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -1286,4 +1286,58 @@ ImageInput::iotell() const
 }
 
 
+
+bool
+ImageInput::check_open(const ImageSpec& spec, ROI range, uint64_t /*flags*/)
+{
+    // Check for sensible resolutions, etc.
+    if ((m_spec.width <= 0 || m_spec.height <= 0 || m_spec.depth <= 0
+         || m_spec.nchannels <= 0)
+        && !supports("noimage")) {
+        errorfmt(
+            "{} image resolution must be at least 1x1, but the file specified {}x{}. Possible corrupt input?",
+            format_name(), m_spec.width, m_spec.height);
+        return false;
+    }
+    if (m_spec.width > range.width() || m_spec.height > range.height()) {
+        errorfmt(
+            "{} image resolution may not exceed {}x{}, but the file appears to be {}x{}. Possible corrupt input?",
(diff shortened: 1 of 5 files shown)


src/libOpenImageIO/imageinput.cpp
```diff
@@ -1286,4 +1286,58 @@ ImageInput::iotell() const
...
+            m_spec.height);
+        return false;
+    }
+    if (spec.depth >= range.depth()) {
+        errorfmt(
+            "{} volumetric slices may not exceed {}, but the file appears to be {}x{}x{}. Possible corrupt input?",
+            format_name(), range.depth(), spec.width, spec.height, spec.depth);
+        return false;
+    }
+    if (spec.nchannels < range.chbegin || spec.nchannels >= range.chend) {
```
ThiagoIze, line 1315: I see in the docs:
```c++
    /// And the channel extent:
    ///     [chbegin,chend)]
```
So assuming 0-based indexing, that means if `nchannels==1` is specified, `chbegin` of 1 is an error? And `chend==1` is valid? If so maybe the code should be
```suggestion
    if (spec.nchannels <= range.chbegin || spec.nchannels > range.chend) {
```

ThiagoIze (changes requested):
Looks good except for the possible logic error I commented on.

src/libOpenImageIO/imageinput.cpp
ThiagoIze: Also, is it ok to have 3 channels and specify a range [0,1)? That fails too since `3>1` (or `3>=1`). Maybe you want:
```suggestion
    if (spec.nchannels <= range.chbegin || spec.nchannels < range.chend || range.chbegin >= range.chend) {
```

lgritz: Oh, I see, perhaps I've done this inconsistently with how it works for ImageOutput::check_image. Let me edit a bit and I'll update.

ThiagoIze: The surprising thing for me is that all the CI checks passed. I would have assumed there would be tons of fails if what I said was true?

ThiagoIze (approved):
looks good

PR #3970, opened 2023-08-31
build: Fix broken libheif < 1.13

lgritz:
Typo introduced in #3894 broke our ability to build against libheif < 1.13. A conditional compilation that should have kicked in for >=1.16 was mistakenly written as 1.6.


jessey-git (approved):
LGTM

PR #3976, opened 2023-09-08
Remove unnecessary headers from strutil.cpp causing build trouble

jessey-git:
## Description

While testing out the 2.5 beta for use in Blender, I hit an odd build error[1] apparently related to include order. Apparently `windows.h` should be included before `shellapi.h`[2]. Swapping the order does fix things, however, it turns out neither is used in this particular file so just remove them both.  Unsure why we didn't see it until now.

[1]
```
20>[9/146] Building CXX object src\libutil\CMakeFiles\OpenImageIO_Util.dir\strutil.cpp.obj
20>FAILED: src/libutil/CMakeFiles/OpenImageIO_Util.dir/strutil.cpp.obj

20>cl : Command line warning D9002: ignoring unknown option '/arch:sse2'
20>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\shellapi.h(68): error C2146: syntax error: missing ';' before identifier 'DECLSPEC_IMPORT'
20>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\shellapi.h(79): error C2065: 'HDROP': undeclared identifier
20>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\shellapi.h(82): error C2086: 'int EXTERN_C': redefinition
20>C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\shellapi.h(68): note: see declaration of 'EXTERN_C'
```
[2] 
https://stackoverflow.com/questions/2953704/vs2010-lots-of-errors-when-including-standard-libraries
https://social.msdn.microsoft.com/forums/vstudio/en-us/ed93ae48-969a-4280-8fd5-ef57a332022b/problems-with-shellexecute-and-shellapih?forum=vcgeneral

## Checklist:

  nonconforming.

### DIFF
--- a/src/libutil/strutil.cpp
+++ b/src/libutil/strutil.cpp
@@ -31,11 +31,6 @@
 #include <OpenImageIO/thread.h>
 #include <OpenImageIO/ustring.h>
 
-#ifdef _WIN32
-#    include <shellapi.h>
-#    include <windows.h>
-#endif
-
 
 // We use the public domain stb implementation of vsnprintf because
 // not all platforms support a locale-independent version of vsnprintf.


lgritz:
Why didn't this fail on our Windows CI?

jessey-git:
I'm really not sure. `windows.h` was added to this file as part of 83c88530 but as to why CI didn't catch it, I'm guessing some difference in environment as I'm using vs2022 Preview while CI is still on the vs2019 toolsets.

lgritz (approved):
LGTM, thanks

PR #3978, opened 2023-09-08
fix(ustring): fix Cuda warnings

lgritz:
To my eyes, these two constructors look like they should be treated as doing exactly the same thing:

    MyType()
        : mymember(...)
    { }

and

    MyType()
    {
       mymember = ...;
    }

But nvcc in Cuda mode says no, if the constructor is constexpr, the former is okey dokey, but latter gives a warning. Fine, Cuda, whatver you say.

### DIFF
--- a/src/include/OpenImageIO/ustring.h
+++ b/src/include/OpenImageIO/ustring.h
@@ -816,41 +816,41 @@ public:
 
     /// Construct a ustringhash from a null-terminated C string (char *).
     OIIO_DEVICE_CONSTEXPR explicit ustringhash(const char* str)
-    {
 #ifdef __CUDA_ARCH__
         // GPU: just compute the hash. This can be constexpr!
-        m_hash = Strutil::strhash(str);
+        : m_hash(Strutil::strhash(str))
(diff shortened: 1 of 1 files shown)


ThiagoIze (approved):
I think this was an issue in C++ as well until c++20, as we can see in these c++ examples: https://gcc.godbolt.org/z/a9xvPsYso with clang and gcc.

I don't know if this is a good idea, but you could create a macro (and then undef it later) that does the `Strutil::strhash(str)` / `ustring(str).hash()` and then you won't need all these ifdefs and comments in the various ustringhash constructors. This might not be worth it for just three places, so I'm approving it as is.

lgritz:
Yes. The warning specifically says "this is not allowed until C++20". Still surprising to me that even before C++20, any compiler or the language would treat unconditional assignment of the sole member variable *inside* the braces as being any different than the `: foo(value)` member initialization. I mean, sure, it's a different syntactic construct, but in my brain it has to result in the same code, so it should be the same rules about whether it's allowed, right?

PR #3979, opened 2023-09-09
perf(oiiotool): oiiotool --mosaic improve type conversion

lgritz:
`oiiotool --mosaic` was always allocating a `float` buffer for the result.  Instead, make it use the "widest" data type of the input images. So, for example, if you are making a mosaic of uint8 images, the result buffer should be uint8, not have to allocate 4x bigger nor do uint8->float conversions as it pastes the images.


lgritz:
Last chance for objections...

grdanny (approved):
looks reasonable to me.

PR #3980, opened 2023-09-10
png: a variety of minor optimizations to the PNG writer

lgritz:
* Implement PNGOutput::write_scanlines, previously we only had write_scanline. Sadly, being able to write multiple scanlines at once reduces overhead by much less than I had hoped. But there's a bit of savings, so I'll keep it.

* Rely on libpng's ability to do endianness byte swapping (for new enough libpng). This means that when we do need byte swapping and it's one of the cases where libpng can do it, we can avoid an extra copy.

* New compression recognized: "pngfast", which translates to the Z_BEST_SPEED setting. I don't particularly recommend this, but it's useful for benchmarking.

* New compression recognized: "none", which turns off compression. Also not recommended, as it makes the files much larger. Mostly done for benchmarking and other comparisons.

* Don't unconditionally copy user's data buffer to m_scratch. It's only necessary if there's a data type conversion, stride shuffling, alpha deassociation, or endian swapping. When none of those are needed, we can avoid the extra allocatin and copy, and so now we do. (Sigh, in real use, we're almost always doing the alpha deassociation, so this rarely is a savings.)

* Better documentation on the custom output attribute hint `"png:filter"` noting that a non-default value can make it write PNG files dramatically faster, but with the tradeoff of having much larger files.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1603,10 +1603,10 @@ control aspects of the writing itself:
 
        **Important**: We have noticed that 8 (PNG_FILTER_NONE) is much
        faster than the default of NO_FILTERS (sometimes 3x or more faster),
-       but also makes larger resulting files that are quite a bit larger
-       (sometimes 2x larger). If you need to optimize PNG write speed and
-       are willing to have larger PNG files on disk, you may want to use
-       that value for this attribute.
+       but it also makes the resulting files quite a bit larger (sometimes
+       2x larger). If you need to optimize PNG write speed and are willing
+       to have larger PNG files on disk, you may want to use that value for
+       this attribute.
 
 **Custom I/O Overrides**
 
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1602,11 +1602,11 @@ control aspects of the writing itself:
        (``PNG_FILTER_PAETH``).
 
        **Important**: We have noticed that 8 (PNG_FILTER_NONE) is much
-       faster than the default of NO_FILTERS (someimes 3x or more faster),
-       but also makes larger resutling files are quite a bit larger (sometimes
-       2x larger). If you are need to optimize PNG write speed and are willing
-       to have larger PNG files on disk, you may want to use that value for
-       this attribute.
+       faster than the default of NO_FILTERS (sometimes 3x or more faster),
+       but also makes larger resulting files that are quite a bit larger
+       (sometimes 2x larger). If you need to optimize PNG write speed and
+       are willing to have larger PNG files on disk, you may want to use
+       that value for this attribute.
 
 **Custom I/O Overrides**
 
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1601,6 +1601,13 @@ control aspects of the writing itself:
        (``PNG_FILTER_UP``), 64 (``PNG_FILTER_AVG``), or 128
        (``PNG_FILTER_PAETH``).
 
+       **Important**: We have noticed that 8 (PNG_FILTER_NONE) is much
+       faster than the default of NO_FILTERS (someimes 3x or more faster),
+       but also makes larger resutling files are quite a bit larger (sometimes
+       2x larger). If you are need to optimize PNG write speed and are willing
+       to have larger PNG files on disk, you may want to use that value for
+       this attribute.
+
 **Custom I/O Overrides**
 
 PNG input and output both support the "custom I/O" feature via the special
--- a/src/png.imageio/pngoutput.cpp
+++ b/src/png.imageio/pngoutput.cpp
@@ -181,13 +181,6 @@ PNGOutput::open(const std::string& name, const ImageSpec& userspec,
     }
 
     m_need_swap = (m_spec.format == TypeDesc::UINT16 && littleendian());
-#if defined(PNG_WRITE_SWAP_SUPPORTED)
-    if (m_need_swap) {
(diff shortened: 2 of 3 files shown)


src/doc/builtinplugins.rst
```diff
@@ -1601,6 +1601,13 @@ control aspects of the writing itself:
        (``PNG_FILTER_UP``), 64 (``PNG_FILTER_AVG``), or 128
        (``PNG_FILTER_PAETH``).
 
+       **Important**: We have noticed that 8 (PNG_FILTER_NONE) is much
+       faster than the default of NO_FILTERS (someimes 3x or more faster),
+       but also makes larger resutling files are quite a bit larger (sometimes
+       2x larger). If you are need to optimize PNG write speed and are willing
+       to have larger PNG files on disk, you may want to use that value for
+       this attribute.
+
```
jessey-git, line 1610: Some spelling mistakes above. How about:
```
**Important**: We have noticed that 8 (PNG_FILTER_NONE) is much
faster than the default of NO_FILTERS (sometimes 3x or more faster),
but it also makes the resulting files quite a bit larger (sometimes
2x larger). If you need to optimize PNG write speed and are willing
to have larger PNG files on disk, you may want to use that value for
this attribute.
```

src/png.imageio/pngoutput.cpp
```diff
@@ -289,11 +310,12 @@ bool
 PNGOutput::write_scanline(int y, int z, TypeDesc format, const void* data,
                           stride_t xstride)
 {
+    bool need_swap = (m_spec.format == TypeDesc::UINT16 && littleendian());
```
jessey-git, line 313: Why not use `m_need_swap` here in this method, and below in `PNGOutput::write_scanlines`?

jessey-git (changes requested):
Some comments/questions.

src/png.imageio/pngoutput.cpp
```diff
@@ -323,6 +344,62 @@ PNGOutput::write_scanline(int y, int z, TypeDesc format, const void* data,
...
+        if (m_spec.format == TypeDesc::UINT16)
+            deassociateAlpha((unsigned short*)data, npixels, m_spec.nchannels,
+                             m_spec.alpha_channel, m_gamma);
+        else
+            deassociateAlpha((unsigned char*)data, npixels, m_spec.nchannels,
+                             m_spec.alpha_channel, m_gamma);
+    }
+
+    // PNG is always big endian
+    if (need_swap)
```
jessey-git, line 387: Forgotten `m_need_swap` replacement.

jessey-git (changes requested):

src/doc/builtinplugins.rst
lgritz: Agreed. Replacing my text with yours.

lgritz:
Updated based on review comments.

src/png.imageio/pngoutput.cpp
```diff
@@ -165,10 +170,26 @@ PNGOutput::open(const std::string& name, const ImageSpec& userspec,
         png_set_compression_strategy(m_png, Z_RLE);
     } else if (Strutil::iequals(compression, "fixed")) {
         png_set_compression_strategy(m_png, Z_FIXED);
+    } else if (Strutil::iequals(compression, "pngfast")) {
+        png_set_compression_strategy(m_png, Z_DEFAULT_STRATEGY);
+        png_set_compression_level(m_png, Z_BEST_SPEED);
+    } else if (Strutil::iequals(compression, "none")) {
+        png_set_compression_strategy(m_png, Z_NO_COMPRESSION);
+        png_set_compression_level(m_png, 0);
+        print("compression zero\n");
```
jessey-git, line 179: Unnecessary `print` I think.

jessey-git (commented):
One inlined comment more.  Also, I'm experiencing failures in the Blender test suite after applying this patch but I won't have time to investigate until much later tonight.

src/png.imageio/pngoutput.cpp
lgritz: oops, you are right, I was using that to debug

lgritz:
seem ok now?

jessey-git:
Actually, no, when I built earlier I got test failures in Blender's test suite when saving out 16 bit PNGs. I won't be able to investigate further until later tonight.  Some treachery is afoot somewhere...

lgritz:
I see the bug!

lgritz:
never mind, no I don't
false alarm

lgritz:
@jessey-git One thing I'm suspicious of is at line 184:

```
#if defined(PNG_WRITE_SWAP_SUPPORTED)
    if (m_need_swap) {
        // libpng can do the swap for us, we don't need to
        png_set_swap(m_png);
        m_need_swap = false;
    }
#endif
```

If something is going wrong with 16 bit files, try comment out the above fragment and see if that fixes it. Maybe trusting libpng to do that for us was misguided.

jessey-git:
@lgritz Seems there's 2 problems. Repro attached below if you want to try this.
- Inside the new `write_scanlines` method, you'll need to divide `nbytes` by `sizeof(unsigned short)` in the call to `swap_endian((unsigned short*)data, nbytes);`
- Yes, disabling the `png_set_swap` and letting `m_need_swap` remain `true` seems to fix things.

Repro:
`iconvert -d uint16 template-rgba32.exr out.png`
[template-rgba32.zip](https://github.com/OpenImageIO/oiio/files/12592884/template-rgba32.zip)

lgritz:
Thanks, that example helped a lot. Updated.

jessey-git (approved):
Seems fine now!

PR #3982, opened 2023-09-11
feat(python): Implement ImageCache.get_imagespec()

lgritz:
This C++ method of ImageCache was never exposed previously.


jessey-git (approved):
LGTM

PR #3983, opened 2023-09-12
build: Test ABI compliance

lgritz:
This patch adds a new test to the CI matrix: verifying that we do not break the ABI during releases.

Basically, in addition to building the current checkout, it also builds a checkout of the designated "ABI standard" (generally, the first non-beta release of a major version), then generates ABI dumps of both versions and runs the ABI compliance checker using the tools from https://github.com/lvc

Making two builds sounds expensive, but both builds are rigged to be minimal, since only the main two libraries are needed for ABI checks (we can turn off the binary tools, unit tests, python bindings, and support for many formats), and of course the test suite need not run. So actually, it's less expensive than most of the test matrix entries, and it helps ensure that we don't accidentally break ABI when we don't mean to.

I'm not sure how many false positives we'll have. We might need to adjust the methodology slightly as we gain experience with this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,6 +186,27 @@ jobs:
             setenvs: export BUILDTARGET=clang-format
                             OIIO_CMAKE_FLAGS=-DUSE_PYTHON=0
 
+          # Test ABI stability. `abi_check` is the version or commit that we
+          # believe is the current standard against which we don't want to
+          # break the ABI. Basically, we will build that version as well as
+          # the current one, and compare the resulting libraries.
+          - desc: abi check
+            nametag: linux-vfx2023
+            os: ubuntu-latest
+            container: aswftesting/ci-osl:2023-clang15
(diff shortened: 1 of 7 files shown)


lgritz:
Any opinions about this methodology?

hobbes1069:
Looks good to me! Interested in the output. I'm assuming added functions should not cause a failure, correct? Only removed or modified ones.

dekekincaid:
We always use https://abi-laboratory.pro when checking why something broke with open source libraries so 👍 big fan here.  

I've been meaning to bring this up for the aswf because it would be great to have all the libraries on their graphs and up to date.  Openimageio and EXR hasn't been updated since 2020.

https://abi-laboratory.pro/index.php?view=timeline&l=oiio
https://abi-laboratory.pro/index.php?view=timeline&l=openexr

lgritz:
Yes, the intent is not to fail for *additions*.

PR #3984, opened 2023-09-12
Write out proper tiff header version in png EXIF blobs

jessey-git:
## Description

When writing out EXIF headers in PNG files, the `tiff_version` field wasn't handling endianness properly.

This lead to some tools like `exiv2` refusing to process the file at all[1]

[1] https://github.com/Exiv2/exiv2/issues/2745

## Checklist:

  nonconforming.

### DIFF
--- a/src/libOpenImageIO/exif.cpp
+++ b/src/libOpenImageIO/exif.cpp
@@ -1302,6 +1302,8 @@ encode_exif(const ImageSpec& spec, std::vector<char>& blob,
     TIFFHeader head;
     head.tiff_magic   = (endianreq == endian::little) ? 0x4949 : 0x4d4d;
     head.tiff_version = 42;
+    if (endianreq != endian::native)
+         swap_endian(&head.tiff_version);
     // N.B. need to swap_endian head.tiff_diroff  below, once we know the sizes
     append(blob, head);
 
--- a/src/libOpenImageIO/exif.cpp
+++ b/src/libOpenImageIO/exif.cpp
@@ -1303,7 +1303,7 @@ encode_exif(const ImageSpec& spec, std::vector<char>& blob,
     head.tiff_magic   = (endianreq == endian::little) ? 0x4949 : 0x4d4d;
     head.tiff_version = 42;
     if (endianreq != endian::native)
-         swap_endian(&head.tiff_version);
+        swap_endian(&head.tiff_version);
     // N.B. need to swap_endian head.tiff_diroff  below, once we know the sizes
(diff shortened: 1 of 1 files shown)


jessey-git:
There is one additional quirk which I have not been able to find the cause of.  `exiftool -v3` will complain about what looks like a duplicate set of eXIf headers in the png:
```
From exiftool: Warning = IFD0 pointer references previous IFD0 directory

From exiv2:
STRUCTURE OF PNG FILE: T:\png_exif.png
 address | chunk |  length | data                           | checksum
       8 | IHDR  |      13 | ............                   | 0x3cafe9a7
      33 | eXIf  |      54 | MM.*.......,.......H.......... | 0x37db1a45    <-- XResolution\YResolution TIFF exif data
      99 | oFFs  |       9 | ........                       | 0xda2ab6ce
     120 | pHYs  |       9 | ...,...H                       | 0x73db2dc6
     141 | IDAT  |      29 | ..cdhhh`@..`.....4.......%...  | 0xd095bfc8
     182 | eXIf  |      54 | MM.*.......,.......H.......... | 0x37db1a45    <-- Why has it been duplicated??
     248 | IEND  |       0 |                                | 0xae426082

```

jessey-git:
I believe this is ready for review.  The "duplication" problem noted is not critical and can be solved separately if no one happens to have a good solution to checkin along side the primary fix.

brechtvl:
I'm not sure that duplication is an actual problem. That header follows the calls to `png_set_IHDR`, `png_set_oFFs`, `png_set_pHYs` in `png_pvt.h`. It seems like the image resolution and physical resolution are equal which seems valid?

jessey-git:
Wow, I might be getting extremely unlucky.  We're using libpng 1.6.39 (Nov 2022) and there's an interesting fix in 1.6.40 which might explain the duplication at least. https://sourceforge.net/p/libpng/code/ci/e6c5bf46c4fa3e3738a2ed0896ce33fc270e0cac/
```
Ensure that only one eXIf chunk is written in the entire datastream

EXIF data can be stored in an eXIf chunk before IDAT, or after IDAT,
but the entire PNG datastream may contain one eXIf chunk at most. ...
```

I haven't found good information about the `IFD0 pointer references previous IFD0 directory` warning though - just that the exiv2 authors said it's a real issue in the first report. I could see if bumping the libpng version helps, it'll just take a bit of time to rebuild the world.

@lgritz Does the primary fix here seem ok though?

lgritz (approved):
LGTM

PR #3987, opened 2023-09-17
build: Use exported targets for libjpeg-turbo and bump min to 2.1

lgritz:
libjpeg-turbo has exported cmake configs since its version 2.1. Relying on this simplifies the build and allows us to completely remove the troublesome Findlibjpeg-turbo.cmake. This implies raising the floor for libjpeg-turbo support to 2.1 (the current release is 3.0) which was released in April 2021 (beta in Nov 2020).

That's cutting it a little close (we usually support dependencies going back at least 3 years at a minimum, and often significantly longer).  But since this is an optional dependency anyway and simplifies the build, I'm tempted to do so.


domin144:
The solution fits my needs, but I am probably not a representative user ;-) 

Alternatively, the find file could be ranamed to Findlibjpeg-turbo.cmake and provide the libjpeg-turbo::jpeg instead of defining the variables.
This way it would be transparent if the find file or the config was used.
I think, the preferred one should be the config file, but I don't know how to give it priority. Maybe the find file could start with checking for config file.

lgritz:
We definitely prefer config files. The Find modules only exist because dependencies didn't export cmake config files (or don't have them as far back as the oldest version we try to support).

I hadn't realized that libjpeg-turbo had finally added them (I guess a couple years ago now, I didn't think to check, it certainly didn't in 2016 when I first added support for libjpeg-turbo). I would rather switch to using it. The only controversial part is that this requires us to bump our minimum to 2.1, which is a hair under 3 years old, which is a little more recent than we usually like to support. But since it's an optional dependency anyway, I'm tempted to do it anyway.

PR #3991, opened 2023-09-21
docs: Spruce up the main README and add "Building_the_docs"

lgritz:
For the upcoming DevDays, I wanted to spruce up our main README to make it easier to find info, and I wrote a new guide for how to locally build the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+<!-- Copyright Contributors to the OpenImageIO Project. -->
 <p align="center">
   <img src="ASWF/logos/openimageio-horizontal-gradient.png">
 </p>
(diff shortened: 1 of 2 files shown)


lgritz:
@grdanny Would love your feedback on the "building the docs" docs, since you had to do that recently.

lgritz:
You can see a preview of the new README here: https://github.com/lgritz/oiio/tree/lg-devdays

lgritz:
Does this look like an improvement to people?

grdanny:
It looks good to me.
Was able to follow it all on our work machines (using spk instead of pip).
I'm trying on my personal laptop too, but that is a bit more challenging because it's all windows so the main build has some issues.

I think you can go ahead and merge. I'll keep playing with it just so we are ready to answer some questions on dev day.
Your advice of trying to get all to build and manage your dependencies ahead of time is a good one as it does take a while to get all the different packages you need install and built.

PR #3994, opened 2023-09-21
testing: Set up placeholders for docs-examples-are-tests

lgritz:
I had previously prototyped an instance of this for the ImageOutput chapter.

This patch makes empty placeholders for all of the other API chapters. They don't contain any actual tests or generate any output now, but they do build and run, so people can drop new examples into them without having to worry about how to set up new files for each chapter.

### DIFF
--- a/testsuite/docs-examples-python/src/docs-examples-imagebuf.py
+++ b/testsuite/docs-examples-python/src/docs-examples-imagebuf.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/OpenImageIO/oiio
+
+from __future__ import print_function
+from __future__ import absolute_import
+
+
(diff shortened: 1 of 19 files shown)


testsuite/docs-examples-python/src/docs-examples-imagebuf.py
```diff
@@ -0,0 +1,46 @@
...
+
+############################################################################
+# This file contains code examples from the ImageInput chapter of the
+# main OpenImageIO documentation.
+#
+# To add an additional test, replicate the section below. Change
+# "example1" to a helpful short name that identifies the example.
+
+
+# BEGIN-imageinput-example1
```
jessey-git, line 19: Should be `BEGIN-imagebuf-example1` (along with the END below)

jessey-git (approved):
Left 1 minor comment but the rest is fine!

lgritz:
Thanks. I tried to fix all my copy pasta, but it doesn't surprise me that I missed one.

PR #3995, opened 2023-09-25
fix: Overhaul of ColorConfig internals to solve perf issues

lgritz:
First, some background: In the course of working toward OIO 2.5, we added some heuristics that used OpenColorIO 2.2's new isColorSpaceLinear, and also using its new "built-in config" to figure out if an arbitrarily named color spaces in a user's config were equivalent to some built-in ones that are commonly used.

All seemed well until... we recently discovered that it was moderately expensive (~0.1s per color space!) on the OCIO side, which wasn't particularly noticeable when I was testing, but can add up to several seconds of needless startup time when inventorying large configs with dozens and dozens of color spaces. And even worse, it was exceptionally expensive for color spaces with LUT3Ds because of a bug on the OCIO side that would invert the 3D LUT needlessly (taking >10s per LUT3D inversion), leaving to MINUTES of overhead for some configs with many such spaces.

So this PR attempts to work around all of this with several big changes (which are amplified by the fact that I've really shuffled around the code between methods and within the file, sorry):

1. Per suggestion from Doug Walker, there's a new heuristic to try to detect if a color space from or to the reference space appears to contain a 3D LUT, and if so, skip some of the more expensive heuristics that appear to cause the LUT to be inverted. It's not perfect, but we're making a bet that no LUT3D-involving color spaces are going to be one of the few canonical spaces we're trying to detect. This fixes the "10s of seconds of delays" problem for those particular color spaces.

2. When using OCIO 2.3+, use its newly introduced utility IdentifyBuiltinColorSpace to have it tell us which color space in the user's config matches some of the canonical ones we care about in the built-in config.

3. Whew, switch to a kind of lazy evaluation of the propererties of each color space. Whereas before, we fully inventoried the config and for each cs we found in it, do all the heuristics, now we only do the inexpensive tests up front (silly rules of thumb where we guess based on the *names* of the color spaces). We save the potentially more expensive but thorough ones (like transforming some sentinel colors around to see if the behavior of the cs matches a color space we care about, and asking OCIO what it thinks about isColorSpaceLinear) only if we are asked to do certain things with *that* color space. In other words, a cs in the config that OIIO doesn't need to *use* in some way will now not undergo the expense of the trickier investigation about what it might be equivalent to.

These add up to making the overhead for ordinary use be significantly sub-second, no longer a problem. Nature is healing.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -164,7 +164,8 @@ struct CSInfo {
         is_Rec709          = 32,  // Rec709 primaries and transfer function
         is_known           = is_srgb | is_lin_srgb | is_ACEScg | is_Rec709
     };
-    int flags = 0;
+    int m_flags   = 0;
+    bool examined = false;
     std::string canonical;  // Canonical name for this color space
 #ifdef USE_OCIO
     OCIO::ConstColorSpaceRcPtr ocio_cs;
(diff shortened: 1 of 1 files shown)


lgritz:
This is not the kind of change I would normally have no desire to include in an imminent release, but the perf problem when using OCIO 2.2 is really severe without the fix, so I'm tempted to squeeze it in and just consider it a top priority to address any problems that are reported.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -162,9 +162,11 @@ struct CSInfo {
         is_lin_srgb        = 8,   // sRGB/Rec709 primaries, linear response
         is_ACEScg          = 16,  // ACEScg
         is_Rec709          = 32,  // Rec709 primaries and transfer function
-        is_known           = is_srgb | is_lin_srgb | is_ACEScg | is_Rec709
+        is_known           = is_srgb | is_lin_srgb | is_ACEScg | is_Rec709,
+        has_lut3d          = 64,  // has a LUT3D transform
```
doug-walker, line 166: The has_lut3d flag seems to get set below but never checked?  I think you should be able to remove the Lut3D check once the OCIO bug is fixed, so hopefully there is no future need that would make it useful as part of the CSInfo.

lgritz: Fair enough. Will clean up a bit.

lgritz: Updated

lgritz:
I know all of this is kind of abusing OCIO's intent in a number of ways, at least as envisioned to work within a single consistent color managed environment like a studio. 
But as a practical matter of trying to make this library behave with reasonable defaults by a wide range of users who may build and use it outside any controlled environment (or certainly outside my ability to predict exactly what will be in their configs, if they even have one), I've had to make quite a few compromises of ideals in the name of obeying the principle of least surprise for the widest range of users.

doug-walker (approved):

PR #3996, opened 2023-09-26
Prevent infinite loop in bit_range_convert

### LINKED ISSUE #3993: [BUG] OIIO library freezes when loading R16G16 uncompressed DDS files
**Describe the bug**
Since Blender switched to OIIO from hardcoded image loaders (since version 3.6.0 I think) I'm having issues opening certain DDS files. I tried debugging Blender on Windows 11 (visual studio 19) and when I step into ImageInput::read_image(), the function never returns. As this theoretically could be an issue with the incorrect use of OIIO API in Blender, I also tried opening the problematic file in the "iv" image viewer on Fedora 38 - turns out it freezes as well. The process shows 1400 %CPU time so I guess there is some infinite loop. The problematic file is an uncompressed R16G16 normal map DDS with 8 mipmaps.

**To Reproduce**
1. Open the IV image viewer (or Blender 3.6.0)
2. Load the testimage.dds file
3. the application freezes

**Expected behavior**
Could at least throw an error that this format is not supported ;). Ideally, it would just load the texture.

**Evidence**
Here's the problematic file https://drive.google.com/file/d/1gN8VpdVD9qb-wzNrq-NX4wTX4Tvloufm/view?usp=sharing
Can be opened with NVIDIA Texture Tools with no issues. nvddsinfo.exe results:
`Flags: 0x00021007
        DDSD_CAPS
        DDSD_PIXELFORMAT
        DDSD_WIDTH
        DDSD_HEIGHT
        DDSD_MIPMAPCOUNT
Height: 128
Width: 128
Depth: 0
Mipmap count: 8
Pixel Format:
        Flags: 0x00000040
                DDPF_RGB
        Bit count: 32
        Red mask:   0x0000FFFF
        Green mask: 0xFFFF0000
        Blue mask:  0x00000000
        Alpha mask: 0x00000000
Caps:
        Caps 1: 0x00401008
                DDSCAPS_COMPLEX
                DDSCAPS_TEXTURE
                DDSCAPS_MIPMAP
        Caps 2: 0x00000000
        Caps 3: 0x00000000
        Caps 4: 0x00000000`

**Platform information:**
 - OIIO branch/version: 2.4.15.0 on Fedora 38, not sure which version they use in Blender 3.6.0
 - OS: Windows 11 and Fedora 38
 - C++ compiler:

jessey-git:
## Description

It was possible for `bit_range_convert` to be called with a 0 for the `FROM_BITS` parameter leading to an infinite loop in its processing.

Fixes #3993 

## Tests

The test image from the above bug is now able to be read correctly.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -1229,6 +1229,7 @@ convert_type (const S &src)
 /// shifted fully to the right.
 template<unsigned int FROM_BITS, unsigned int TO_BITS>
 inline OIIO_HOSTDEVICE unsigned int bit_range_convert(unsigned int in) {
+    static_assert(FROM_BITS > 0, "FROM_BITS cannot be 0");
     unsigned int out = 0;
     int shift = TO_BITS - FROM_BITS;
     for (; shift > 0; shift -= FROM_BITS)
@@ -1244,10 +1245,12 @@ inline OIIO_HOSTDEVICE unsigned int
 bit_range_convert(unsigned int in, unsigned int FROM_BITS, unsigned int TO_BITS)
 {
     unsigned int out = 0;
-    int shift = TO_BITS - FROM_BITS;
-    for (; shift > 0; shift -= FROM_BITS)
-        out |= in << shift;
-    out |= in >> -shift;
+    if (FROM_BITS) {
+        int shift = TO_BITS - FROM_BITS;
+        for (; shift > 0; shift -= FROM_BITS)
+            out |= in << shift;
+        out |= in >> -shift;
+    }
     return out;
 }
 


src/include/OpenImageIO/fmath.h
```diff
@@ -1229,6 +1229,7 @@ convert_type (const S &src)
 /// shifted fully to the right.
 template<unsigned int FROM_BITS, unsigned int TO_BITS>
 inline OIIO_HOSTDEVICE unsigned int bit_range_convert(unsigned int in) {
+    static_assert(FROM_BITS > 0, "FROM_BITS cannot be 0");
```
lgritz, line 1232: The static_assert makes it impossible to successfullyinstantiate the template with FROM_BITS == 0. So why is the `if (FROM_BITS)` necessary later on?

jessey-git: There's 2 variations of bit_range_convert (1 template and 1 normal) and I suspect you're looking at the wrong one from the diff?

lgritz: Oh, right you are. My brain thought I was just looking later in the same template. My bad.

lgritz (approved):

PR #3998, opened 2023-09-29
admin: Repo rename -- fix all URL references

lgritz:


ThiagoIze (approved):
Aside from project_intake.md (which looks good to me), this looks like standard search and replace. I verified a few of the links are correct, so I'll assume the rest are too.

lgritz:
Yeah, almost all just search/replace. They're either all right, or all wrong, I guess.

PR #3999, opened 2023-09-30
admin: Adjust repo location

lgritz:
Missed one spot.


lgritz:
Preemptively merging because it fixes broken CI.

PR #4000, opened 2023-10-02
Use modern FindPython from CMake 3.12

hmaarrfk:
This really helps with cross compilation and virtual environments

The minimum CMake Version is already 3.15

xref: https://github.com/conda-forge/openimageio-feedstock/pull/87

## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/CONTRIBUTING.md).
  nonconforming.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -156,8 +156,7 @@ else ()
     message (STATUS "Using internal PugiXML")
 endif()
 
-# From pythonutils.cmake
-find_python()
+find_package(Python COMPONENTS Interpreter Development)
 
 
 ###########################################################################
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Sigh, it hasn't had the effect i wanted...

lgritz:
Can you explain what you're trying to do here?

The scripts already did use the modern FindPython. Our find_python() code you deleted called checked_find_package, which in turn is just a fancy wrapper around find_package (with lots of extra bells and whistles). The find_python sets some options/variables beforehand, so maybe you just wanted to alter or delete some of those?

How can we help you in Conda land?

hmaarrfk:
You are correct. I've been trying to get cross compilation working and for some reason the "wrong" python is sometimes found.

PR #4005, opened 2023-10-06
feat: print unretrieved global error messages

lgritz:
In the ongoing effort to help newcomers to OIIO diagnose their own problems, this uses a trick to print any global error messages that never got retrieved by OIIO::geterror() at the time that the program exits.  (This is all similar to a change we made recently that adds an analogous behavior to ImageBuf, but in this case it's for those error messages that don't have an object like an ImageBuf to attach to.)

This should have no effect on anybody who diligently checks OIIO::has_error() or OIIO::geterror() after any failed calls that would send error messages to that global store. Also, it is possible to disable the behavior by setting the new global attribute "oiio:print_uncaught_errors" to 0.

But for users who don't know to check errors and are experiencing baffling behavior with OIIO, maybe seing those error messages will prompt them to understand what's going wrong before they feel the need to file an issue against OIIO or need to ask us for help.

Doing this helped me to find two places where our own code in OIIO failed to clear the global error at times that it should, and these are also fixed as part of this PR.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -2980,6 +2980,21 @@ OIIO_API std::string geterror(bool clear = true);
 ///    the log information. When the `log_times` attribute is disabled,
 ///    there is no additional performance cost.
 ///
+/// - `oiio:print_uncaught_errors` (1)
+///
+///   If nonzero, upon program exit, any error messages that would have been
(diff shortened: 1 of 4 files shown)


ThiagoIze (approved):
This looks good to me. I was concerned about any slowdowns this might introduce, but the per-thread error string was already there, so all this adds is a check on that string when the thread is destroyed, which seems pretty harmless. Forcing by default error messages to print out seems like a good idea to me.

PR #4007, opened 2023-10-08
testing: imagebuf_test add benchmarks for iterator traversal

lgritz:
To help with some other investigation I'm doing, amend imagebuf_test with benchmarks to compare the relative speeds of traversing a big ImageBuf with Iterator, ConstIterator, or just using pointers on an equivalently large array.

### DIFF
--- a/src/libOpenImageIO/imagebuf_test.cpp
+++ b/src/libOpenImageIO/imagebuf_test.cpp
@@ -540,6 +540,62 @@ test_mutable_iterator_with_imagecache()
 
 
 
+void
+time_iterators()
+{
+    print("Timing iterator operations:\n");
+    const int rez = 4096, nchans = 4;
+    ImageSpec spec(rez, rez, nchans, TypeFloat);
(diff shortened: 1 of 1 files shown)


lgritz:
No objections, right?

grdanny (approved):
lgtm 
for my own curiosity - why is this test under /src and not /testsuite ?

lgritz:
> for my own curiosity - why is this test under /src and not /testsuite ?

Historical precedent?

When dinosaurs roamed the earth, major classes got unit compiled unit tests that live in the same directory as the library containing the class (imagebuf.cpp also has imagebuf_test.cpp, etc.). They are *run* as part of the testsuite, but they are built along with the library (but they are not "installed", just an artifact of the build that's used for testing).

It would have been an equally reasonable choice, I guess, to make that source code itself live in the testsuite and only build when the testsuite is run. But that's not the way I did it at the time.

grdanny:
got it, thanks for explaining.

Issue #2058, opened 2018-11-10
Great Big `iv` Wish List


lgritz:
# Master list of `iv` feature requests

Rather than proliferate scattered issues, let's coalesce all the requested iv features or changes here. Feel free to add comments with new suggestions, and I will occasionally merge them into this list.

They are in no particular order.

-----

- [ ] **Full support of OCIO color spaces and displays**
- [ ] Dark mode (was #311)  (Suggestion: does this help? https://github.com/ColinDuquesnoy/QDarkStyleSheet
- [ ] Difference mode: shows you the absdiff between the current image and the last image. (originally #291)
- [ ] Wipe mode: There should be a mouse+modifier that splits the screen horizontally and/or vertically (depending on the mouse movement direction) of the current image and the previous image. (originally #292)
- [ ] Have iv able display partially-written exr files (originally #1191).
- [ ] Area probe: Allow a way to sweep out a rectangle and report the min/max/avg values within the rectangle. (originally #310)
- [ ] Flip and rotate should be added to the edit menu (or view?). (originally #290)
- [ ] "Save window" doesn't work -- it saves the whole image. (originally #286)
- [ ] iv Help menu should have a choice that tells you all the available plugins or formats supported (similar to `oiiotool --help`) (originally #279)

Silverlan:
Support for the KTX-format would be great!
https://www.khronos.org/opengles/sdk/tools/KTX/file_format_spec/

lgritz:
@silverlan That's a great idea. Can you file a new issue about supporting this file format? This is not really an `iv` related issue, as this ticket is dedicated to, since iv doesn't contain any format-specific code at all (it just relies on OIIO's support for many formats).

lgritz:
I added the KTX request on a different ticket.

lgritz:
I marked this as "good first issue", but just to clarify -- this is a "super-issue" that lists many previously suggested improvements to iv. Any *one* of these may be a good first issue, not trying to do all of them at once, obviously.

CheeksTheGeek:
Hi, I want to take up the task for 
"Load a directory: iv dirname ought to be the same as specifying all the image files in that directory"
which image file formats is the iv tool supposed to work with?

CheeksTheGeek:
also, is it supposed to also work with `iv filename1 dirname` and so on?

lgritz:
[...]

PR #4010, opened 2023-10-10
[2058] Implement Directory Argument Loading for iv

CheeksTheGeek:
Solves the issue "Load a directory: iv dirname ought to be the same as specifying all the image files in that directory" from the Great Big iv Wish List #2058 

## Description

This PR addresses the request from the Master list of iv feature requests to allow loading a directory as an argument, which in turn loads all images within the specified directory. The code modifications include checks to validate the file or directory existence, and if a directory is provided, iterates through its files to load all valid images into iv. This enhancement simplifies the process of viewing multiple images collectively located in a single directory.

## Tests

N/A

## Checklist:

  nonconforming.


src/iv/ivmain.cpp
```diff
@@ -108,8 +108,38 @@ main(int argc, char* argv[])
...
+            Filesystem::get_directory_entries(f, files);
+
+            std::vector<std::string> validImages;  // Vector to hold valid images
+            for (auto& file : files) {
+                auto in = ImageInput::open(file);
+                if (in) {
+                    validImages.push_back(file);
+                    in->close();  // Close the ImageInput object to free resources
+                }
+            }
```
lgritz, line 130: I think that this is technically correct (except for the fact that if the open fails, you are not retrieving any error messages to throw them out).

But I think this could be unnecessarily expensive. What you need to know is that what ImageInput::open does is first try the image reader corresponding to the file's extension (if there is one), but if that fails, then it will try every other file reader, just in case the file is mis-named. So if the directory is filled with lots of non-image files, it might try to open it with each of a dozen or more reader types, all to no avail.

So one thing I would suggest is that you only open() the files that match one of the known extensions corresponding to a supported image file. (See this](https://openimageio.readthedocs.io/en/v2.5.4.0/imageioapi.html#_CPPv4N4OIIO12getattributeE11string_view8TypeDescPv) for the getattribute query that will give you all known extensions).

Also, since iv is going to do everything through the ImageCache, it may be good to use it here, too. Assuming that you have set up

    ImageCache* cache = ImageCache::create();
    ustring uexists("exists");

outside the loop, then you can check if a file is a valid image in one step:

    int exists = 0;
    bool ok = cache->get_image_info(ustring(file), 0, 0, uexists, OIIO::TypeInt, &exists);
    if (ok && exists)
        validImages.push_back(file);

rather than the open call. I'm not certain that this is significantly better, but it may be, because it will prevent you from having to open each file twice (remember, you're probably gonna want to display each image, too).

CheeksTheGeek:
I've added a processing area for the extensions since they're formatted like `openexr:exr,sxr,mxr;tiff:tif,tiff,tx,env,sm,vsm;jpeg:jpg,jpe,jpeg.......`, and then inside the looping for the directory members,  it matches it via a vector. 
The first loop across the files handles filling in the validImages vector, and the second one which only runs if validImages is not empty.

lgritz:
@CheeksTheGeek ALL the commits need to be signed. (Silly, yes, but that's what the bot is looking for.)

The easiest way to do this is 

    git rebase -i HEAD~5     # (replace 5 with whatever number of commits you have)

This will bring up an editor window with the list of commits. Change "pick" to "r" on every line and save & exit. It will then bring up an editor on each commit in sequence, allowing you to edit the commit messages individually. For each one, just add this line:

    Signed-off-by: Chaitanya Sharma <39400946+CheeksTheGeek@users.noreply.github.com>

and save & exit. When this is all done, `git push origin feature/iv-directory-as-arg --force`

CheeksTheGeek:
Done!

src/iv/ivmain.cpp
```diff
@@ -107,9 +107,60 @@ main(int argc, char* argv[])
...
+    while (std::getline(ss, item, ';')) {
+        size_t pos = item.find(':');
+        if (pos != std::string::npos) {
+            std::string exts = item.substr(pos + 1);
+            std::stringstream inner_ss(exts);
+            while (std::getline(inner_ss, item, ',')) {
+                extensionsVector.push_back(item);
+            }
+        }
+    }
```
lgritz, line 125: I would like to suggest here

```suggestion
    std::vector<std::string> extensionsVector;
    auto all_extensions = OIIO::get_string_attribute("extension_list");
    for (auto oneformat : OIIO::Strutil::splitsv(all_extensions, ";")) {
        auto format_exts = OIIO::Strutil::splitsv(oneformat, ":", 2);
        for (auto ext : OIIO::Strutil::splitsv(format_exts[1], ","))
            extensionsVector.emplace_back(ext);
    }

```

lgritz:
I build and tested this on my end -- works as advertised, very nice!

src/iv/ivmain.cpp
CheeksTheGeek: Oh that's much more elegant, thank you!!

CheeksTheGeek:
Done!

lgritz:
Works very well!

There is one small behavior that is surprising. I guess when you enumerate directory entries, it comes out in any random order (I suppose in the order the files within the directory appears to the OS), and that's a little surprising because if you'd said `iv dir/*.exr` you would have gotten them in alphabetical order. This will be especially crucial if, for example, your directory has all the frames of a shot (`dir/foo.0001.exr`, `dir/foo.0002.exr`, etc.).

I wonder if you would mind putting in an extra step so that the files within each directory are sorted lexicographically? (But only within each directory, not across all files and directories, if you know what I mean.)

src/iv/ivmain.cpp
```diff
@@ -107,9 +107,53 @@ main(int argc, char* argv[])
...
+    ustring uexists("exists");
+    std::vector<std::string> extensionsVector; // Vector to hold all extensions
+    auto all_extensions = OIIO::get_string_attribute("extension_list");
+    for (auto oneformat : OIIO::Strutil::splitsv(all_extensions, ";")) { // Split the extensions by semicolon
+        auto format_exts = OIIO::Strutil::splitsv(oneformat, ":", 2);
+        for (auto ext : OIIO::Strutil::splitsv(format_exts[1], ","))
+            extensionsVector.emplace_back(ext); 
+    }
+
+
```
lgritz, line 119: I know this is absurdly nit-picky, but would you mind eliminating the extra blanks here? We often use one blank line to separate conceptual steps (like here, between the gathering of all extensions, and the start of handling the images), but I prefer than there not be more than two blank lines in a row *within a function*. I think that extra blanks inside a function doesn't really add to the readability, but can make the function spread out too much vertically and may not all fit in a screenfull. (Though I do, as it turns out, tend to prefer multiple blank lines to separate standalone functions, to visually separate them.)

lgritz:
Thank you, the sorting definitely is better behavior.

The only other request I had is about those blank line, but otherwise I think this is basically complete!

src/iv/ivmain.cpp
CheeksTheGeek: Completely valid, though, does OIIO have a preferred formatter and/or linter for ensuring pre PR consistency?

lgritz: We use clang-format, though I don't think we are enforcing the blank line issue, because our preference is for at most 1 blank line in a row within a function, exactly 3 blank lines between freestanding (top level) functions/methods, and 1 blank line between method declarations within a class declaration, is not a granularity that clang-format appears to support.

CheeksTheGeek: Okayy, I'll keep this in mind for other PRs(stuff in docs examples and tests one and possibly other ones if I get to complete thr first one!).

lgritz (approved):
LGTM!

Issue #3249, opened 2021-12-29
[FEATURE REQUEST] pip install OpenImageIO


c0ffymachyne:
Hi,

Please automate and unify python builds and make it available via `pip install OpenImageIO`
Exemplary project providing build for all possible python versions under releases: https://github.com/isl-org/Open3D/releases/tag/v0.14.1

lgritz:
This would be a great thing. But it's well outside my expertise, so what I really need is for somebody who has more Python ecosystem knowledge, preferably somebody who has done this exact thing on another project, to step forward and take the lead on this task. I think that this has recently been done on both OpenColorIO and OpenTimelineIO, so maybe there is somebody who both works on those projects and OIIO and would like to give this a try and can do it efficiently since they already know how it works?

ddesmond:
upvote on this one!

studiowildcard:
Please! It's sooo complicated to get oiio working with Python in Windows, especially for casual Python users. Just spent an entire day figuring it out (including little idiosyncrasies like making sure it builds with ocio) via Microsoft vcpkg - and even them I'm stuck with Python 3.10 and have no idea how to build it for 3.9. Anyone??

JeanChristopheMorinPerso:
I'm currently working on that. I have fully static builds working on Linux and Windows. I haven't started macOS yet.

I'm aiming for having a library that comes with most plugins. The only excluded plugins are OpenVDB, R3D, Nuke, DCMTK, OpenCV and the `iv` command.

JeanChristopheMorinPerso:
Oh, and I'll need testers eventually. If you (or someone you know) are interested, please let me know and also tell which platform and architecture you will test on.

Thanks!

ddesmond:
I can test on mac, windows and centos 7.

howieyoo:
willing to test on ubuntu 20.04

stereorobin:
This would be a really big help, especially for pure python developers (or tech-artists).
I've spent a good part of my day trying to get the python bindings to work on windows through vcpkg, and I seem to come across a lot of old threads where others face the same issue.

ddesmond:
@JeanChristopheMorinPerso pinging - any updates on this? can we help you test? thanks!

JeanChristopheMorinPerso:
I'll update when I have more news. The work I'm doing is not sponsored by any company, which means I do that in my own free time.

Rusching:
[...]

PR #4011, opened 2023-10-11
build: Add support for `pip install OpenImageIO`

aclark4life:
## Description

Adds support for`pip install OpenImageIO` requested in #3249

## Checklist:

  nonconforming.


pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+name = "OpenImageIO"
+version = "2.4.0.dev1"
+description = "Reading, writing, and processing images in a wide variety of file formats, using a format-agnostic API, aimed at VFX applications."
+authors = [
+    {name = "Larry Gritz", email = "lg@larrygritz.com"},
+]
+dependencies = [
+    "setuptools>=68.2.2",
+]
+requires-python = ">2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, !=3.6.*, !=3.9.0"
```
jessey-git, line 11: Is there a particular reason to exclude 3.9.0 exactly?  Also, would attempting to use python 2.7.18 pass this check? If so, I have a meta question/comment around if we want to continue to support and encourage such nonsense :)

jessey-git (commented):
To fix the failing "DCO" check, the last time this happened to me I had to amend my commit and force-push my branch again (the PR will auto-update):
```
git commit --amend -s
git push --force origin <your branch>  # or similar...
```
The `-s` (which you could have used on any commit) appends the "Signed-off-by" line to your commit message.

I feel this could maybe have a better description too, since this typically ends up in git history in some form.

What is `setup.py`? That file doesn't exist in the repo at the moment.  Who is Jean-Christophe? They don't seem to be any former contributor here; is that information important? I had to do a bit of internet searching to realize who this was :)  Some of the results are, of course, interesting.

pyproject.toml
JeanChristopheMorinPerso: 3.9 is known to be buggy in when used with Pybind11.

JeanChristopheMorinPerso: And no, I don't think we should support 2.7. We'll need at least Pybind11 2.10, which doesn't support python 2.7.

lgritz: VFX Platform 2022 uses Python 3.9, so it's definitely in the mix of things we support, and I don't recall any pybind11 problems (for OIIO anyway).

pyproject.toml
```diff
@@ -0,0 +1,52 @@
+[project]
+name = "OpenImageIO"
+version = "2.4.0.dev1"
+description = "Reading, writing, and processing images in a wide variety of file formats, using a format-agnostic API, aimed at VFX applications."
+authors = [
+    {name = "Larry Gritz", email = "lg@larrygritz.com"},
+]
+dependencies = [
+    "setuptools>=68.2.2",
```
JeanChristopheMorinPerso, line 9: Setup tools should not be needed if we use scikit-build-core. And anyway it would go into the build dependencies, not the runtime dependencies.

JeanChristopheMorinPerso:
Hi @jessey-git 👋, I'm the Jean-Christophe mentioned in the description. I volunteered to help with https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3249, mainly to review PRs, but I'll be happy to help in other things ways too if there is a need. Obviously, that is if you want me to review PRs.

jessey-git:
@JeanChristopheMorinPerso Hi!  I did eventually find you within the ASWF umbrella after searching and you're of course more than welcome to help out in whatever capacity you'd like :)  And ah! The description was missing a link back to the issue which would have made things a lot more clear where this was coming from.

aclark4life:
@jessey-git Sorry, this is a WIP I created to get the CLA signed before tomorrow!

aclark4life:
@JeanChristopheMorinPerso Can you iplease dentify which of the commits listed here are most important to cherry-pick? https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3249#issuecomment-1739756119 For example, we don't need to build with setup.py so not going to cherry-pick that one. Thanks

pyproject.toml
```diff
@@ -0,0 +1,52 @@
+[project]
+name = "OpenImageIO"
+version = "2.4.0.dev1"
```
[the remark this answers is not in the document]
aclark4life: Yes, I updated with 2.6.0.2 because that looked reasonably close. It doesn't really matter yet because even after this is merged, and we can see GitHub actions building wheels, we'll need to [integrate GitHub Actions with Trusted Publishers](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/). Or in the short term, if you wanted to publish to PyPI before that happens, we could make sure the version number is correct in the release branch then [manually upload wheels with twine](https://twine.readthedocs.io/en/stable/).

pyproject.toml
aclark4life: Removed all Python version restrictions. Definitely not going to support 2.7 or anything older than 3.9 at this point, but I'm not sure yet if we need to be explicit about that here.

pyproject.toml
aclark4life: Removed setuptools

pyproject.toml
lgritz: 2.7 is definitely fine to remove, we'll surely remove all 2.7 support by the time OIIO 2.6 is fully baked next summer.

"Nothing older than 3.9" means "people can't pip install a version that will work with any VFX platform year prior to 2022. Which may be fine? But it's a choice we should make with eyes open.

JeanChristopheMorinPerso: I would still support 3.7 and 3.8. 3.7 is still widely used by the industry and all ASWF projects ship wheels for 3.7, 3.8, 3.9, 3.10 and 3.11. By experience, supporting 3.7 doesn't add any extras work or burden when using Pybind11.

aclark4life: Also 3.9.0 was released Oct. 5, 2020. 3.9.18 is the [latest point release](https://www.python.org/downloads/), released Aug 24, 2023. So it's probably not practical to exclude a single point release when there are 17 releases between it and the latest point release. If anything, we'd probably exclude `<3.9.18,<3.10.13,<3.11.6,<3.12.0` and so on.

lgritz: We don't want to cut off users who need particular older versions, but at the same time, if the only people who care about those versions are big studios or products that build from source anyway, then maybe those are indeed cases we should exclude by design.

I think we need to start with clear first principles statement here: which group of users are we aiming to cater to with a pip install solution, and what versions of python does that pool of users care about?

JeanChristopheMorinPerso: Usually, the target group for wheels is anyone not part of a big studio, students, scientists, etc. All people that don't necessarily either have control over the python versions they have, or don't necessarily know that python 3.7 is EOL. Usually, when users pip install, they expect things to just work. By experience, if we don't support 3.7 or 3.11, we'll get support requests all the time. Users will ask for them. 3.7 because DCCs use it, and 3.11 because users want it. Requests for 3.12 will also start to arrive soon (we already had one in OTIO).

aclark4life: This is the roadmap, right? https://vfxplatform.com/

In that case, you could go all the way back to VFX 2017 and add support for whatever Python was in use then, but for the purposes of the sprint I would say since CY2024 remains on VFX 2022 that we support 3.8 and higher in the next OIIO release because that's what's supported by Python at the time of the release.

The further back you go with Python versions, the more work you may end up doing. But you are correct you need to know what version of Python is in use and you may want to include that information on the VFX roadmap. I can picture a "worst case" scenario where everyone uses Python 3.7 with VFX 2022. I wouldn't like it, but we could probably focus on that 3.7 only in that case :smile: .

Lastly, this may be slightly premature as it will be nice to get just some useful wheels built in any version over the next 48 hours or so!

lgritz: This may help: https://docs.google.com/spreadsheets/d/1EwRlz5ZYObEOdBfIk8iTX5thlpTyEAfp3bxOgAfFOiU/edit#gid=1592527777

That shows (near the bottom) what VFX Platform years the different studios are using. Prior to 2020 is definitely not needed, and I think even 2020 (maybe 2021) is on the way to being phased out. This also matches with VFX Platform's own recommendation of trying to support up to 3 years back from the present.

lgritz: I'll put it in even more stark terms: my own studio will definitely be transitioned away from Python 2.7 by the end of this year, I hope. (And onto an assortment of 3.7, 3.9, 3.10, depending on the specific one of the several DCCs we are depending on.)

lgritz: In terms of priority, it's likely that the users who need a simple "pip install" approach most desperately (i.e., want to avoid building anything from source) are very likely on a very recent python. So that's probably top priority. But I'm playing devil's advocate here to make sure we don't back ourselves into a corner, design wise, that will make it difficult to eventually support python versions as far back as the VFX years we claim to support.

lgritz: I do agree that we should definitely not support python older than 3.7. Whether 3.7 or 3.9 is the minimum we want to support for "pip install" is reasonable for debate, and I don't have a firm opinion on it, other than that we should choose among them for a principled reason.

JeanChristopheMorinPerso: The code base already supports 3.7 and up thanks to Pybind11, so building for 3.7 isn't more difficult than building for 3.11. 3.12 is a different story though because it's still pretty recent (it's only one week old). In other words, once you get a build running and working wheels for one version, you get the other versions for free without any additional work.

aclark4life:
Happy sprint day 1 folks!

Please correct me if any of these assumptions are wrong:

- VFX industry uses OpenImageIO as part of the VFX Platform and ASWF supports development of the VFX Platform
- There's currently no formal binary distribution of OpenImageIO from ASWF, thus leaving the community with ad hoc attempts to provide them e.g.:
  - https://github.com/MethanePowered/OpenImageIOBinary
  - https://www.patreon.com/posts/openimageio-oiio-53939451
- DCCs may include binary distributions of OpenImageIO thus lowering the burden of installation for VFX industry
  - Binary distributions may include the `oiiotool` command line tool as well as "much of its functionality" available via Python bindings
  - Python wheels would further reduce OpenImageIO installation burden by making its Python API available via `pip install OpenImageIO`

Most importantly, if I understand correctly, the `oiiotool` and related runtime libraries are not needed for Python API usage when OpenImageIO is installed via `pip install OpenImageIO` because all of the dependencies will be included in the wheel.

lgritz:
>    VFX industry uses OpenImageIO as part of the VFX Platform and ASWF supports development of the VFX Platform

OIIO is not in the VFX Platform. ASWF is not the originator of the VFX Platform. But the VFX Platform largely specifies an important subset of the permutations of versions that we (OIIO and the other ASWF projects) definitely need to be able to build against. 

>    There's currently no formal binary distribution of OpenImageIO from ASWF, thus leaving the community with ad hoc attempts to provide them e.g.:

Correct.

>    DCCs may include binary distributions of OpenImageIO thus lowering the burden of installation for VFX industry

Many DCCs and other applications use OIIO internally. The OIIO components are thus generally compiled in statically, or put in libraries with custom symbol namespaces so as to not interfere with any other OIIO used in the environment. I'm not aware of any DCC that ships OIIO in a form that's exposed to and useful to the users of the DCC.

> Python wheels would further reduce OpenImageIO installation burden

Yes.

> Most importantly, if I understand correctly, the oiiotool and related runtime libraries are not needed for Python API usage when OpenImageIO is installed via pip install OpenImageIO because all of the dependencies will be included in the wheel.

I would disagree with this. I think there is one set of users who are only after the Python APIs and thus view OIIO like any other Python library that they want to import. But a second set of users merely want "pip install openimageio" as the simplest and least painful way to get a full binary install of the entire package and its dependencies. Those people want python, C++, oiiotool, the whole schmear. Also, I think that somebody who starts out only wanting/needing the python interface could easily find themselves in a position of later needing oiiotool, and I wouldn't want them to suddenly be stuck without any way to get it without slogging through a full set of builds from source.

lgritz:
A crucial thing to understand about VFX Platform it is not a list of recommendations, nor is it a list of all the things needed in a VFX studio. It's a list of packages that have historically been a such a compatibility versionitis nightmare (often because of deficiencies in their design or packaging) that we needed to get the major DCC vendors to agree to all use the same version in the same year, and this is the list of which versions they negotiated. Nothing more and nothing less.

aclark4life:
Copy that, thanks!

JeanChristopheMorinPerso:
>I would disagree with this. I think there is one set of users who are only after the Python APIs and thus view OIIO like any other Python library that they want to import. But a second set of users merely want "pip install openimageio" as the simplest and least painful way to get a full binary install of the entire package and its dependencies. Those people want python, C++, oiiotool, the whole schmear. Also, I think that somebody who starts out only wanting/needing the python interface could easily find themselves in a position of later needing oiiotool, and I wouldn't want them to suddenly be stuck without any way to get it without slogging through a full set of builds from source.

Indeed @lgritz. The work I did last year included adding the tools into the wheels for this exact reason (because people need th tools or the API or both).

aclark4life:
OK so in Python lingo, we have `console_scripts = oiiotool` so that folks have access to the same command line they'd have available if they compiled the C++?

JeanChristopheMorinPerso:
Yep, and it's already there, see https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4011/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R37-R44

lgritz:
We should probably talk at some point about which are important enough to include in the python wheel. Definitely oiiotool and maketx. Probably not testshade, which I think is only used in the testsuite for unit tests but isn't used by users (am I right about that?). Technically, iinfo, iconvert, and idiff don't do anything that oiiotool can't do (they exist mostly because they predate oiiotool), though some people find them useful because they each do one simple thing well without having to look up the oiiotool commands needed to do the same thing. I honestly don't know if anybody uses igrep (it seemed cool on the day I wrote it, but I also can't recall the last time I needed it). And iv... maybe? (Though also maybe it should be renamed, it's so short and generic it does scare me a little with it being confused with or hiding other things.)

src/build-scripts/gh-installdeps.bash
```diff
@@ -83,7 +84,10 @@ else
         locales wget \
         libopencolorio-dev \
         libopencv-dev \
-        libhdf5-dev
+        libhdf5-dev \
```
lgritz, line 87: As an aside, I think libhdf5-dev can be removed. I believe that is a holdover from when we used field3d (which we no longer support).

src/build-scripts/ci-startup.bash
```diff
@@ -73,4 +73,4 @@ elif [[ "${RUNNER_OS}" == "macOS" ]] ; then
 fi
 
 # Save the env for use by other stages
-src/build-scripts/save-env.bash
+# src/build-scripts/save-env.bash
```
lgritz, line 76: I think this line is going to turn out to be important.

src/build-scripts/gh-installdeps.bash
```diff
@@ -213,4 +237,4 @@ if [[ "$USE_ICC" != "" ]] ; then
 fi
 
 # Save the env for use by other stages
-src/build-scripts/save-env.bash
+# src/build-scripts/save-env.bash
```
lgritz, line 240: Pretty sure that was necessary

src/build-scripts/gh-installdeps.bash
aclark4life: Removed

src/build-scripts/ci-startup.bash
aclark4life: Fixed

src/build-scripts/gh-installdeps.bash
aclark4life: Fixed

aclark4life:
OK rebased and fixed issues raised by @lgritz and it looks like @JeanChristopheMorinPerso signed DCO :rocket: so today I will continue with cherry-picking remainder of applicable @JeanChristopheMorinPerso commits.

lgritz:
@aclark4life Can you please mark this as a draft PR, and then remove the draft status when you think it's ready to go? I'm afraid that my knowledge about what you're trying to do is so thin that I'm not sure it will be obvious to me when this is ready to review or complete.

Alternately, if you know how to break it into tasks or milestones, you could amend the description to have that checklist and tick them off as they get done, then it'll be easy to see which pieces are done and which are still in progress.

aclark4life:
@lgritz Almost done with something that can get merged! 

In #3249 you'll see a list of commits from @JeanChristopheMorinPerso that I've cherry-picked and fixed conflicts. A few more to go, then I'll add the trusted publisher configuration and we'll call it a sprint.

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+igrep = "OpenImageIO._commands:igrep"
+iinfo = "OpenImageIO._commands:iinfo"
+maketx = "OpenImageIO._commands:maketx"
+oiiotool = "OpenImageIO._commands:oiiotool"
+testtex = "OpenImageIO._commands:testtex"
+
+[build-system]
+requires = [
+    "scikit-build-core",
+    "wheel",
```
JeanChristopheMorinPerso, line 49:
```suggestion
```

src/build-scripts/build_pugixml.bash
```diff
@@ -21,6 +21,12 @@ PUGIXML_BUILD_DIR=${PUGIXML_BUILD_DIR:=${PUGIXML_SRC_DIR}/build}
 LOCAL_DEPS_DIR=${LOCAL_DEPS_DIR:=${PWD}/ext}
 PUGIXML_INSTALL_DIR=${PUGIXML_INSTALL_DIR:=${LOCAL_DEPS_DIR}/dist}
 #PUGIXML_BUILD_OPTS=${PUGIXML_BUILD_OPTS:=}
+PUGIXML_BUILD_SHARED_LIBS=${PUGIXML_BUILD_SHARED_LIBS:="ON"}
+
+PUGIXML_CXX_FLAGS=${PUGIXML_CXX_FLAGS:=}
+if [[ "$OSTYPE" == 'msys' ]] && [[ "${PUGIXML_BUILD_SHARED_LIBS}" != "ON" ]]; then
+    PUGIXML_CXX_FLAGS="${PUGIXML_CXX_FLAGS} //MT"
```
JeanChristopheMorinPerso, line 28: `/MT` shouldn't be needed if we are not building for Python 2.7.

src/build-scripts/docker-centos.bash
```diff
@@ -0,0 +1,123 @@
+# docker run -ti --rm -v $(pwd):/jc quay.io/pypa/manylinux2014_x86_64:latest bash
+set -ex
+
+if [[ "$OSTYPE" != "msys" ]]; then
+    yum install ninja-build wget less ccache -y
```
JeanChristopheMorinPerso, line 5: If we are going through `scikit-build-core`, we shouldn't have to install ninja.

src/build-scripts/docker-centos.bash
```diff
@@ -0,0 +1,123 @@
...
+fi
+
+python -m pip install conan numpy --user
```
JeanChristopheMorinPerso, line 8: `numpy` should be put as a build dependency in `pyproject.toml`. We'll actually need to use `oldest-supported-numpy`.

src/cmake/compiler.cmake
```diff
@@ -580,7 +579,11 @@ if (LINKSTATIC)
     if (WIN32)
         set (CMAKE_FIND_LIBRARY_SUFFIXES .lib .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
     else ()
-        set (CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+        set (CMAKE_FIND_LIBRARY_SUFFIXES .a)
+    endif ()
+
+    if (MSVC)
+        add_compile_options (/MT)
```
JeanChristopheMorinPerso, line 586: Shouldn't be needed if we don't build for Python 2.7.

src/cmake/externalpackages.cmake
```diff
@@ -186,6 +186,10 @@ if (OpenColorIO_FOUND)
     endif ()
 else ()
     set (OpenColorIO_FOUND 0)
+else ()
+    checked_find_package (yaml-cpp)
```
JeanChristopheMorinPerso, line 190: I thin we might be able to skip these if we use the latest OCIO available...

src/cmake/externalpackages.cmake
```diff
@@ -198,25 +202,29 @@ checked_find_package (TBB 2017
                       PREFER_CONFIG)
 
 checked_find_package (DCMTK VERSION_MIN 3.6.1)  # For DICOM images
-checked_find_package (FFmpeg VERSION_MIN 3.0)
+checked_find_package (ffmpeg
```
JeanChristopheMorinPerso, line 205: I'm not 100% sure, but I think CMake is case sensitive and I think the diff here is due to conan generating CMake configs with a package name that differs from upstream. So we might have to figure out a way to support both `FFmpeg` and `ffmpeg`...

src/libOpenImageIO/CMakeLists.txt
```diff
@@ -149,6 +153,9 @@ target_link_libraries (OpenImageIO
             ${format_plugin_libs} # Add all the target link libraries from the plugins
             $<TARGET_NAME_IF_EXISTS:OpenColorIO::OpenColorIO>
             $<TARGET_NAME_IF_EXISTS:OpenColorIO::OpenColorIOHeaders>
+            $<TARGET_NAME_IF_EXISTS:yaml-cpp::yaml-cpp>
+            $<TARGET_NAME_IF_EXISTS:pystring::pystring>
+            $<TARGET_NAME_IF_EXISTS:EXPAT::EXPAT>
```
JeanChristopheMorinPerso, line 158: I think the most recent version of OCIO might fixed the issue that lef me to add this.

JeanChristopheMorinPerso (commented):
Great start @aclark4life!

For others following, I think the next steps would be to finish that and get a working build, and then make sure that we can build using `python -m build -w` (ie make sure that scikit-build-core can be used). Then I guess we'll need a GitHub Action workflow that does the whole end to end build with cibuildwheel. After that, we'll know if the theory works.

CMakeLists.txt
```diff
@@ -144,6 +133,7 @@ message(STATUS "Setting Namespace to: ${PROJ_NAMESPACE_V}")
 add_definitions (-DOIIO_INTERNAL=1)
 
 list (APPEND CMAKE_MODULE_PATH
+      "${PROJECT_SOURCE_DIR}/build"
```
lgritz, line 136: Why? There aren't going to be any cmake modules in this location, which won't yet exist when this cmake file is processed and the modules are needed.

INSTALL.md
```diff
@@ -239,7 +239,7 @@ Additionally, a few helpful modifiers alter some build-time options:
 | make STOP_ON_WARNING=0    |  Do not stop building if compiler warns        |
 | make EMBEDPLUGINS=0 ...   |  Don't compile the plugins into libOpenImageIO |
 | make USE_OPENGL=0 ...     |  Skip anything that needs OpenGL               |
-| make USE_QT=0 ...         |  Skip anything that needs Qt                   |
+| make USE_QT5=0 ...         |  Skip anything that needs Qt                   |
```
lgritz, line 242: Extremely minor nit: can you fix the algnment?

src/build-scripts/build_opencolorio.bash
```diff
@@ -20,13 +20,28 @@ OPENCOLORIO_BUILD_DIR=${OPENCOLORIO_BUILD_DIR:=${LOCAL_DEPS_DIR}/OpenColorIO-bui
...
 OPENCOLORIO_BUILDOPTS="-DOCIO_BUILD_APPS=OFF -DOCIO_BUILD_NUKE=OFF \
                        -DOCIO_BUILD_DOCS=OFF -DOCIO_BUILD_TESTS=OFF \
                        -DOCIO_BUILD_GPU_TESTS=OFF \
                        -DOCIO_BUILD_PYTHON=OFF -DOCIO_BUILD_PYGLUE=OFF \
                        -DOCIO_BUILD_JAVA=OFF \
                        -DBUILD_SHARED_LIBS=${OPENCOLORIO_BUILD_SHARED_LIBS:=ON}"
+
+if [ "${OPENCOLORIO_BUILD_SHARED_LIBS}" != "ON" ]; then
+    OPENCOLORIO_BUILDOPTS="${OPENCOLORIO_BUILDOPTS} -DCMAKE_POSITION_INDEPENDENT_CODE=ON"
+fi
```
lgritz, line 43: Why not just add this unconditionally?

src/build-scripts/build_opencolorio.bash
```diff
@@ -46,9 +61,10 @@ git checkout ${OPENCOLORIO_VERSION} --force
 time cmake -S . -B ${OPENCOLORIO_BUILD_DIR} \
            -DCMAKE_BUILD_TYPE=${OPENCOLORIO_BUILD_TYPE} \
            -DCMAKE_INSTALL_PREFIX=${OPENCOLORIO_INSTALL_DIR} \
-           -DCMAKE_CXX_FLAGS="${OPENCOLORIO_CXX_FLAGS}" \
-           ${OPENCOLORIO_BUILDOPTS}
-time cmake --build ${OPENCOLORIO_BUILD_DIR} --config Release --target install
+           -DCMAKE_CXX_FLAGS_RELEASE="${OPENCOLORIO_CXX_FLAGS}" \
+           -Dyaml-cpp_CXX_FLAGS="${OPENCOLORIO_YAML_CXX_FLAGS}" \
+           ${OPENCOLORIO_BUILDOPTS} ${OPENCOLORIO_SOURCE_DIR}
+time cmake --build . --config Release --target install
```
lgritz, line 67: Is there a reason why someone using this utility script should no longer be able to override the build directory name and only allow the build to be in ./build ?

src/build-scripts/build_openexr.bash
```diff
@@ -39,21 +47,43 @@ mkdir -p ${OPENEXR_INSTALL_DIR} && true
...
+            -DCMAKE_INSTALL_PREFIX="${OPENEXR_INSTALL_DIR}" \
+            -DCMAKE_PREFIX_PATH="${CMAKE_PREFIX_PATH}" \
+            -DILMBASE_PACKAGE_PREFIX="${OPENEXR_INSTALL_DIR}" \
+            -DOPENEXR_BUILD_UTILS=0 \
+            -DOPENEXR_BUILD_TESTS=0 \
+            -DOPENEXR_BUILD_PYTHON_LIBS=0 \
+            -DCMAKE_CXX_FLAGS="${OPENEXR_CXX_FLAGS}" \
+            ${OPENEXR_CMAKE_FLAGS} ${OPENEXR_SOURCE_DIR}
+    time cmake --build . --target install --config ${OPENEXR_BUILD_TYPE}
+else
```
lgritz, line 64: I feel like maybe you forward-ported something from a while back. We recently simplified this code to no longer need the "v2.3.0" clause because OpenEXR 2.4 is now the oldest version we support.

src/build-scripts/build_pybind11.bash
```diff
@@ -46,8 +46,9 @@ if [[ -z $DEP_DOWNLOAD_ONLY ]]; then
     time cmake -S . -B ${PYBIND11_BUILD_DIR} -DCMAKE_BUILD_TYPE=Release \
                -DCMAKE_INSTALL_PREFIX=${PYBIND11_INSTALL_DIR} \
                -DPYBIND11_TEST=OFF \
-               ${PYBIND11_BUILD_OPTS}
-    time cmake --build ${PYBIND11_BUILD_DIR} --config Release --target install
+               -DCMAKE_MODULE_PATH=${CONAN_CMAKE_FILES} \
+               ${PYBIND11_BUILD_OPTS} ..
+    time cmake --build . --config Release --target install
```
lgritz, line 51: Again here, this script seems to have lost the ability to put the build dir anywhere we want.

src/build-scripts/ci-build.bash
```diff
@@ -27,7 +27,8 @@ if [[ ${GITHUB_ACTIONS} == true ]] ; then
     OIIO_CMAKE_FLAGS+=" -DCMAKE_UNITY_BUILD=${CMAKE_UNITY_BUILD:=ON} -DCMAKE_UNITY_BUILD_MODE=${CMAKE_UNITY_BUILD_MODE:=BATCH}"
 fi
 
-cmake -S . -B build -G "$CMAKE_GENERATOR" \
+pushd build
+cmake .. -G "$CMAKE_GENERATOR" \
```
lgritz, line 31: I think we regressed here. Now that cmake 3.15 is the minimum we support, we can use -B instead of having to cd into a build directory.

src/build-scripts/gh-installdeps.bash
```diff
@@ -106,11 +110,12 @@ else
...
     if [[ "$USE_LIBHEIF" != "0" ]] ; then
-       sudo add-apt-repository ppa:strukturag/libde265 || true
-       sudo add-apt-repository ppa:strukturag/libheif || true
+       sudo add-apt-repository ppa:strukturag/libde265 -y
+       sudo add-apt-repository ppa:strukturag/libheif -y
+       time sudo apt-get update
        time sudo apt-get -q install -y libheif-plugin-aomdec \
             libheif-plugin-aomenc libheif-plugin-libde265 \
-            libheif-plugin-x265 libheif-dev || true
+            libheif-plugin-x265 libheif-dev -y
```
lgritz, line 118: The `|| true` was to keep CI from failing if ppa:strukturag was down, which I suspect must have been happening at some point.

I think there is room for debate about whether it's better to fail CI entirely if a repo is temporarily down, or whether it's fine to press on with the rest of the CI run but take the chance that if the repo stays down, maybe we wouldn't notice that we were not properly testing the heic support.

src/cmake/externalpackages.cmake
```diff
@@ -198,25 +202,29 @@ checked_find_package (TBB 2017
...
+                      COMPONENTS avcodec avformat avutil swscale
+                      VERSION_MIN 3.0)
 checked_find_package (GIF
                       VERSION_MIN 4
                       RECOMMEND_MIN 5.0
                       RECOMMEND_MIN_REASON "for stability and thread safety")
 
 # For HEIF/HEIC/AVIF formats
-checked_find_package (Libheif VERSION_MIN 1.3
+checked_find_package (libheif VERSION_MIN 1.3
```
lgritz, line 214: Is the effect of this that we're relying on libheif's exported cmake config?

That might be fine, but if so, does that alter the minimum version of libheif that we support? (I.e., the fact that we have our own FindLibheif.cmake likely means that when we added libheif support, that package didn't have a reliable exported cmake config.)

src/cmake/externalpackages.cmake
```diff
@@ -198,25 +202,29 @@ checked_find_package (TBB 2017
...
                       RECOMMEND_MIN 1.7
                       RECOMMEND_MIN_REASON "for AVIF support")
 if (APPLE AND LIBHEIF_VERSION VERSION_GREATER_EQUAL 1.10 AND LIBHEIF_VERSION VERSION_LESS 1.11)
     message (WARNING "Libheif 1.10 on Apple is known to be broken, disabling libheif support")
-    set (Libheif_FOUND 0)
+    set (libheif_FOUND 0)
 endif ()
 
-checked_find_package (LibRaw
+checked_find_package (libraw
```
lgritz, line 222: This change is gonna make it not use our FindLibRaw.cmake module. Does libraw have an exported cmake config? If so, when did it appear and does that force us to raise the minimum libraw release we support?

src/cmake/externalpackages.cmake
```diff
@@ -236,12 +244,12 @@ checked_find_package (OpenVDB
                       DEPS         TBB
                       DEFINITIONS  -DUSE_OPENVDB=1)
 
-checked_find_package (Ptex PREFER_CONFIG)
-if (NOT Ptex_FOUND OR NOT Ptex_VERSION)
+checked_find_package (ptex PREFER_CONFIG)
+if (NOT ptex_FOUND OR NOT ptex_VERSION)
```
lgritz, line 248: Same. Ptex does seem to have an exported cmake these days, and I don't mind using that and getting rid of our FindPtex.cmake module. But let's understand clearly what that does to the minimum Ptex that we're able to support.

lgritz:
I think a lot of the CI failures are from having broken the build_foo.bash scripts, no longer able to do the build dependencies in the designated directories but instead forcing the builds to be in ./build.

src/cmake/externalpackages.cmake
JeanChristopheMorinPerso: These come from Conan. So they are Conan generated configs.

CMakeLists.txt
JeanChristopheMorinPerso: Conan will create cmake files there.

src/build-scripts/gh-installdeps.bash
```diff
@@ -66,9 +66,10 @@ if [[ "$ASWF_ORG" != ""  ]] ; then
         icpx --version
     fi
 
-else
+elif [[ "$USE_CONAN" != "1" ]] ; then
     # Using native Ubuntu runner
 
+    sudo add-apt-repository ppa:ubuntu-toolchain-r/test
```
JeanChristopheMorinPerso, line 72: I don't remember why I did these changes. We could probably remove them I guess...

pyproject.toml
aclark4life: Fixed

src/build-scripts/docker-centos.bash
aclark4life: What is the oldest supported numpy? I don't see it here https://docs.google.com/spreadsheets/d/1EwRlz5ZYObEOdBfIk8iTX5thlpTyEAfp3bxOgAfFOiU/edit#gid=1592527777

src/cmake/compiler.cmake
aclark4life: Removed

src/cmake/externalpackages.cmake
aclark4life: Confirming CMake variable names are case sensitive https://cmake.org/cmake/help/book/mastering-cmake/chapter/Writing%20CMakeLists%20Files.html#variables

src/libOpenImageIO/CMakeLists.txt
aclark4life: Gone

INSTALL.md
aclark4life: Done!

src/build-scripts/build_opencolorio.bash
aclark4life: Done

aclark4life:
@lgritz @JeanChristopheMorinPerso Can either of you summarize the build directory regression? I just fixed a bunch of issues, but not that one because I'm hoping there's a consistent fix that can be applied everywhere we have that issue.

src/build-scripts/gh-installdeps.bash
aclark4life: Gone

src/build-scripts/docker-centos.bash
lgritz: I'm honestly not sure. Our numpy needs are so shallow that I'm not sure there really is a minimum. Whatever range works with pybind11, I guess?

src/cmake/externalpackages.cmake
lgritz: Yes, I think this is not necessary for modern OCIO with good exported cmake configs.

src/cmake/externalpackages.cmake
lgritz: I think in this case, I think ffmpeg doesn't generate a cmake config, we have our own src/cmake/modules/FindFFmpeg.cmake, which we could rename to Findffmpeg.cmake (and probably change the variable names therein to match) and have this work? But we're in real trouble if the conan versions ever conflict with the naming of the exported cmake configs from any of the dependencies.

CMakeLists.txt
lgritz: But `${PROJECT_SOURCE_DIR}/build` is the wrong place. If they invoke cmake (`-B`, for example) with any other directory, it could be any other location. I think you want `${CMAKE_BINARY_DIR}` at the very least.

src/cmake/externalpackages.cmake
lgritz: It looks like recent libheif exports cmake configs, and calls it libheif-config.cmake, so we should either change our FindLibheif.cmake module (which I believe I wrote before libheif exported cmake configs) to the lower case convention, or possibly eliminate it entirely if we are willing to bump our libheif minimum to whenever they started exporting cmake configs reliably.

lgritz: This kind of thing has happened many times over the years -- we needed a dependency that has no exported cmake config, so we made our own FindFoo module, and then years later, the dependency project starts exporting cmake config but uses a different capitalization than the one we originally guessed.

lgritz:
> @lgritz @JeanChristopheMorinPerso Can either of you summarize the build directory regression? I just fixed a bunch of issues, but not that one because I'm hoping there's a consistent fix that can be applied everywhere we have that issue.

Just compare it directly to the current master. I understand that you added a few of these build scripts for dependencies where we lacked them entirely. But for the ones that we already had, nothing should change except the absolute minimum additions you need to get the python wheel build, right?

src/build-scripts/docker-centos.bash
JeanChristopheMorinPerso: The oldest needed at compile time will be https://github.com/scipy/oldest-supported-numpy. We will want to use the `oldest-supported-numpy` package at build time for the reasons hilighted in https://github.com/scipy/oldest-supported-numpy.

src/build-scripts/build_opencolorio.bash
```diff
@@ -20,13 +20,26 @@ OPENCOLORIO_BUILD_DIR=${OPENCOLORIO_BUILD_DIR:=${LOCAL_DEPS_DIR}/OpenColorIO-bui
 OPENCOLORIO_INSTALL_DIR=${OPENCOLORIO_INSTALL_DIR:=${LOCAL_DEPS_DIR}/dist}
 OPENCOLORIO_BUILD_TYPE=${OPENCOLORIO_BUILD_TYPE:=Release}
 OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:="-Wno-unused-function -Wno-deprecated-declarations -Wno-cast-qual -Wno-write-strings"}
+OPENCOLORIO_YAML_CXX_FLAGS=${OPENCOLORIO_YAML_CXX_FLAGS:=''}
+if [[ "$OSTYPE" != "msys" ]]; then
+    OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:="-Wno-unused-function -Wno-deprecated-declarations -Wno-cast-qual -Wno-write-strings"}
+    OPENCOLORIO_YAML_CXX_FLAGS=${OPENCOLORIO_YAML_CXX_FLAGS:=''}
+else
+    if [[ "${OPENCOLORIO_BUILD_SHARED_LIBS}" != "ON" ]]; then
+        OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:="//MT"}
```
JeanChristopheMorinPerso, line 29: `/MT` won't be needed if we don't build wheels for Python 2.7.

lgritz: There's no reason to support Python 2.7. Or any particularly old dependencies. I view the set of users who want python wheels as distinct from the studio users who are stuck on old versions of things. If the "pip install openimageio" only supports fairly modern versions of everything, I have no objections.

lgritz:
Wondering what the progress on this is, or what roadblocks exist that we can help with. It might be nice to have this solved once and for all by the time of the 3.0 release this fall.

aclark4life:
@lgritz Sorry, got sidetracked but still interested in completing, and fall is super reasonable. I'm [pitching Pillow project to TAC today](https://github.com/AcademySoftwareFoundation/tac/issues/631) during which time I'll mention my status:

- Pillow has no C++ and does not use CMake, but we do have some C code and we do have an elaborate GHA infrastructure already in place for deploying wheels to PyPI via Trusted Publisher: https://github.com/python-pillow/Pillow/blob/main/.github/workflows/wheels.yml
- Other ASWF projects seem to be slightly ahead in publishing wheels e.g. OpenEXR https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/python-wheels-publish.yml
- All ASWF projects should share a common infrastructure for publishing wheels via Trusted Publisher and I'd like to continue to help build that infrastructure.

lgritz:
Yes, I agree that the only sensible path is to use the same mechanisms as other ASWF projects that are publishing python wheels, unless there are serious deficiencies in their methodologies (in which case we should fix that, then copy what they've done).

CMakeLists.txt
aclark4life: "Fixed" in 803b05c05cae45f79d6f809dff2aeb52ac846231

src/build-scripts/build_opencolorio.bash
aclark4life: Fixed in f2520fa081c68b948da952bb4d19d33eae895f83

src/build-scripts/build_pybind11.bash
aclark4life: Fixed

src/build-scripts/gh-installdeps.bash
aclark4life: Re-add ``|| true``

aclark4life:
> Just compare it directly to the current master. I understand that you added a few of these build scripts for dependencies where we lacked them entirely. But for the ones that we already had, nothing should change except the absolute minimum additions you need to get the python wheel build, right?

I think I fixed these

src/cmake/externalpackages.cmake
aclark4life: Looks like we're back to ``Libheif`` and if I'm reading this correctly we need to "guess right" the first time?

src/cmake/externalpackages.cmake
aclark4life: Gone

src/cmake/externalpackages.cmake
aclark4life: Not sure but I fixed the case at least in 6e0ff75f1fde987a3839ec3598684168e959cd2b to match what's in master

src/cmake/externalpackages.cmake
aclark4life: Fixed case

aclark4life:
Not sure how much actual progress I've made but I just rebased and fixed a bunch of issues raised in comments (or at least tried to!)

lgritz:
What's the status of this PR?

Is this waiting for me specifically?

It's still marked as a draft, but also it seems out of date with many merge conflicts if it were to be merged now.

And most importantly, I should not be the reviewer here. I don't know anything about python wheels. Ideally, this should be reviewed by (a) somebody who has made python wheels before and will be consuming them; and also (b) somebody who set up the python wheels for another ASWF project and therefore knows if this is conforming to a common methodology.

aclark4life:
> What's the status of this PR?

Just lazy-rebased by resolving conflicts with master using master branch files. That means we've lost changes made to those files for this PR because they are now identical to the master branch files. We could just close this for now, however there are still some changes that could be reviewed by anyone interested helping with this issue. Not to mention this is an awesome learning process for me! So I'm going to leave it open for the time being if that's OK with you.

> 
> Is this waiting for me specifically?

No, not especially, unless you are interested in helping with this issue.

> 
> It's still marked as a draft, but also it seems out of date with many merge conflicts if it were to be merged now.

Right this is a fork of JC's branch in which I've tried to finish what JC started but have not managed to do so yet. Since then, the issue of "how do we do this consistently across all projects" was raised.

> 
> And most importantly, I should not be the reviewer here. I don't know anything about python wheels. 

I would not feel obligated to review, but also note wheels are just zip files and they are installed by unzipping the contents to Python's `site-packages` directory. When Python is run, `>>> import OpenImageIO` "just works" via Python's import mechanism. In the case of libraries compiled against C++, apparently that mechanism includes loading `so` files e.g.

Screenshot 2024-06-02 at 9 57 54 AM


> Ideally, this should be reviewed by (a) somebody who has made python wheels before and will be consuming them; and also (b) somebody who set up the python wheels for another ASWF project and therefore knows if this is conforming to a common methodology.

The general consensus from various discussions I've had and observed IIUC is "big studios don't need this" and "many small studios do". So, we probably need more input from "many small studios" on #3249 particularly when it comes time to testing. You mentioned timing this with the fall release and I'm still tentatively planning to do that. However I'm mostly preoccupied with job search now, hoping the need to search will subside by beginning of summer.🤞

JGoldstone:
I'm not qualified to review this PR either but … just wanted to note that OpenVPCal installation instructions say that they'd be able to make that project available on PyPi as soon as OIIO is on PyPi.

https://github.com/Netflix/OpenVPCal/tree/main?tab=readme-ov-file#why-not-pypi

JeanChristopheMorinPerso:
>Right this is a fork of JC's branch in which I've tried to finish what JC started but have not managed to do so yet. Since then, the issue of "how do we do this consistently across all projects" was raised.

I'm not sure to get what you mean by "how do we do this consistently across all projects". Build backends (setuptools, scikit-build-core, etc) are an implementation detail that users and the tooling will never see and building wheels for bindings nowadays is a well defined process and standardized process. All ASWF projects that have bindings do approximately the same thing which is to use cibuildwheel. So I don't know what consistency means here since our projects are already consistent in how they build their wheels (minus build backends, but again that doesn't matter).

>I would not feel obligated to review, but also note wheels are just zip files and they are installed by unzipping the contents to Python's site-packages directory. When Python is run, >>> import OpenImageIO "just works" via Python's import mechanism. In the case of libraries compiled against C++, apparently that mechanism includes loading so files e.g.

Reviewing wheels is more involved than just looking at if importing the library works. A reviewer has to check what files are included, licenses bundled, the linking of libraries (static vs shared), the size of the wheels, etc. The method of how the wheels are uploaded, the validity and accuracy of the metadata also has to be reviewed. And then there are per project things to check that will differ from project to project.

As I said in the past, I'll be happy to review once this PR is complete and ready to review. Right now what's missing is the glue. There is a disconnect between the actual build and where dependencies are fetched from. There is also no CI to test the wheels and test that the builds actually work.

>The general consensus from various discussions I've had and observed IIUC is "big studios don't need this" and "many small studios do". So, we probably need more input from "many small studios" on https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3249 particularly when it comes time to testing.

The question is not really big studios vs small studios. The main argument of creating wheels and more importantly making wheels available on PyPI is to make projects easily accessible for **all** users, including other projects (like @JGoldstone highlighted above).

lgritz:
This needs DCOs on all commits

lgritz:
> "how do we do this consistently across all projects".

The goal is not a blind consistency across all ASWF projects for its own sake, but rather, we observe that some projects have already worked this task out and have been making wheels long enough to have shaken out most of the rookie mistakes. Therefore, using their methodology as a starting point seems like a way to greatly reduce the risk of unforced errors on our part, and also increases the chances that process improvements in one project can be ported to the others, and that anyone who understands the wheel creation in one project is already mostly up to speed on the others.


> There is also no CI to test the wheels and test that the builds actually work.

It seems vital to have a comprehensive and fully automated test of the wheel as a gating factor for publishing it to PyPI.

aclark4life:
Thanks @lgritz and @JeanChristopheMorinPerso ! OK let's pick this up by [answering the question that @JeanChristopheMorinPerso asks here.](https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3249#issuecomment-1580867603). See: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3249#issuecomment-2153070607

PR #4012, opened 2023-10-12
docs: Make an example of doc-to-test in the imagebufalgo chapter

lgritz:
Slightly different set of issues, so I wanted to provide a full example for anyone who wants to work on this chapter.


lgritz:
I'm going to preemptively merge this because it only affects the docs (can't break any code), passes CI, and I want it in place to help the Dev Days people.

PR #4013, opened 2023-10-12
fix: to_OpenCV fails for ImageCache-backed images

### LINKED ISSUE #3800: [BUG] Access violation using IBA::to_OpenCV() with an ImageCache-backed ImageBuf
**Describe the bug**
I get an `Exception 0xc0000005 encountered at address ***: Access violation reading location 0x00000000` while trying to convert a HEIC image to OpenCV format. Here is the PoC I'm working with:

``` C++
#include <algorithm>
#include <filesystem>
#include <format>
#include <iostream>
#include <ranges>
#include <string>
#include <vector>

#include <OpenImageIO/imagebufalgo.h>
#include <OpenImageIO/imageio.h>
#include <opencv2/core.hpp>
#include <opencv2/highgui.hpp>

#include "lib.hpp"

namespace fs = std::filesystem;
namespace rng = std::ranges;
namespace vws = std::ranges::views;

auto main() -> int
{
  // Get image paths
  auto images = std::vector<fs::path> {};
  std::transform(fs::directory_iterator("img/"),
                 fs::directory_iterator(),
                 std::back_inserter(images),
                 [](auto const& entry) { return entry.path(); });

  for (auto const& image : images) {
    auto loaded_image = OIIO::ImageBuf(image.string());
    if (not loaded_image.initialized()) {
      std::cout << loaded_image.geterror() << '\n';
      continue;
    }

    auto cv_image = cv::Mat {};
    try {
      OIIO::ImageBufAlgo::to_OpenCV(cv_image, loaded_image, {}, 1);
      cv::imshow(image.filename().string(), cv_image);
    }catch(const std::exception& e){
      std::cout << "Error when converting: " << e.what() << "\n";
      return -1;
    }
  }

  cv::waitKey();
  return 0;
}
```

**To Reproduce**
1. I'm using vcpkg on Windows with `openimageio[libheif,opencv]` as dependencies
2. In the folder I have .heic images taken with an iPhone

**Expected behavior**
I would expect to see the images shown via OpenCV highUI.

**Evidence**
When debugging I get a `Exception 0xc0000005 encountered at address ***: Access violation reading location 0x00000000` exception.

**Platform information:**
 - OIIO branch/version: 2.4.9.0
 - OS: Windows 11
 - C++ compiler: Microsoft (R) C/C++ Optimizing Compiler Version 19.35.32215 for x64

lgritz:
We were using parallel_convert_image but not realizing that asking for the IB::localpixels() would give us nullptr for an IC-backed image.

Instead, "wrap" the cv::Mat with an IB and then use IBA::copy().  This handles both in-memory and IC-backed ImageBuf's, and also has many other advantages (such as format conversion) we may wish to take advantage of in the future.

Fixes #3800

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_opencv.cpp
+++ b/src/libOpenImageIO/imagebufalgo_opencv.cpp
@@ -343,12 +343,10 @@ ImageBufAlgo::to_OpenCV(cv::Mat& dst, const ImageBuf& src, ROI roi,
 
     size_t pixelsize = dstSpecFormat.size() * chans;
     size_t linestep  = pixelsize * roi.width();
-    bool converted   = parallel_convert_image(
-        chans, roi.width(), roi.height(), 1,
-        src.pixeladdr(roi.xbegin, roi.ybegin, roi.zbegin, roi.chbegin),
-        spec.format, spec.pixel_bytes(), spec.scanline_bytes(), 0, dst.ptr(),
-        dstSpecFormat, pixelsize, linestep, 0, -1, -1, nthreads);
-
+    // Make an IB that wraps the OpenCV buffer, then IBA:: copy to it
+    ImageBuf cvib(ImageSpec(roi.width(), roi.height(), chans, dstSpecFormat),
+                  dst.ptr(), pixelsize, linestep, AutoStride);
+    bool converted = ImageBufAlgo::copy(cvib, src);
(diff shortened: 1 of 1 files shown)


lgritz:
On #3800, the original reporter of the issue confirms that this patch fixes the problem. Merging.

PR #4015, opened 2023-10-12
idiff: allow users to specify a directory as the 2nd argument

### LINKED ISSUE #4009: [FEATURE REQUEST] idiff "directory mode"
On Unix/etc command line, the "diff" utility lets you say

    diff path/to/file1.txt anotherpath/toanother/file2.txt

but also if you 

    diff path/to/file1.txt pathto/dir

where `pathto/dir` is the name of a directory, this is shorthand for

    diff path/to/file1.txt pathto/dir/file1.txt

That is, you can just specify a directory as the second argument, and it knows that you mean to diff against a file of the same filename, in that other directory.

I would like our `idiff` utility to work the same way. You currently have to give both full paths, but many times I've wished you could give just a directory for the second name, and it would know what you meant and behave like `diff` users expect.

davvid:
## Description

Teach `idiff` to accept a directory as the 2nd argument.

Fixes #4009

## Tests

Functional tests for `idiff` don't currently exist. Manual testing verifies it works as advertised.

## Checklist:

- [x] I have updated the documentation, if applicable. (Usage strings updated)
  (adding new test cases if necessary). (N/A)
  exposed the new functionality as oiiotool options). (N/A)
  nonconforming.

### DIFF
--- a/src/idiff/idiff.cpp
+++ b/src/idiff/idiff.cpp
@@ -43,7 +43,7 @@ getargs(int argc, char* argv[])
     ArgParse ap;
     ap.intro("idiff -- compare two images\n"
              OIIO_INTRO_STRING)
-      .usage("idiff [options] image1 image2")
+      .usage("idiff [options] <image1> <image2 | directory>")
       .add_version(OIIO_VERSION_STRING)
       .print_defaults(true);
 
--- a/src/idiff/idiff.cpp
+++ b/src/idiff/idiff.cpp
@@ -169,6 +169,24 @@ print_subimage(ImageBuf& img0, int subimage, int miplevel)
 }
 
 
+// Append the filename from "first" when "second" is a directory.
+// "second" is an output variable and modified in-place.
+inline void
+add_filename_to_directory(const std::string& first, std::string& second)
+{
+    if (Filesystem::is_directory(second)) {
+        char last_byte = second.at(second.size() - 1);
+        if (last_byte != '/' && last_byte != '\\') {
+#if defined(_MSC_VER)
+            second += '\\';
+#else
+            second += '/';
+#endif
+        }
+        second += Filesystem::filename(first);
+    }
+}
+
 
 int
 main(int argc, char* argv[])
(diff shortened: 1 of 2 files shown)


davvid:
@CheeksTheGeek here's my take on the idiff feature.

I didn't originally update the usage/help string, so I took inspiration from https://github.com/CheeksTheGeek/oiio/commit/c7522b70f85b41971eb5ad669d850acd059666ea and adjusted this PR to do something similar. I mentioned you in the commit message for that tweak.

One notable difference between this commit and the one linked above is that this is checking for `size() == 2` before attempting to access the 2nd argument and it doesn't do any particular error handling itself because the code that follows already handles that.

src/idiff/idiff.cpp
```diff
@@ -169,6 +169,24 @@ print_subimage(ImageBuf& img0, int subimage, int miplevel)
...
+#if defined(_MSC_VER)
+            second += '\\';
+#else
+            second += '/';
+#endif
+        }
+        second += Filesystem::filename(first);
+    }
+}
+
```
davvid, line 189: The `Filesystem` class didn't already provide this functionality so adding this to idiff seemed like the right place to keep this for now. If other code ends up needing to build paths then moving this to the Filesystem class might be worth it at that point in time. It didn't seem worth extending its public API just for this one small tweak, though.

lgritz:
Don't worry about the icc/icx CI failures. That's a problem on the Intel side, they updated their signature key but didn't re-sign some of the older versions they had on that repo. It's on them to fix, I think. We will accept PRs that pass all the other CI tests.

lgritz (approved):
LGTM!

This is fine as it is, though may I be greedy and ask for more? Would you be so kind as to touch up src/doc/idiff.rst to document the new feature?

lgritz:
Fixes #4009

lgritz:
Hmmm, that didn't work, probably because I'm not the author. @davvid, would you mind adding the magic line "Fixes #4009" to the description of this PR? That will link the issue and make it close automatically when this PR is merged.

lgritz:
BTW, I built this on my end and tested, works! This is cool, now my "diff" muscle memory will work with "idiff" as well.

davvid:
I think the commits themselves have `Closes: #4009` in them so that should handle it. I'll update the description as well.

I'll also take a stab at updating `idiff.rst` as well -- thanks for the pointer to that file. I didn't notice it until you pointed it out.

davvid:
I pushed in a new commit that updates `src/doc/idiff.rst` as well. Thanks for reviewing!

lgritz (approved):
LGTM, thanks for adding it to the docs

Issue #3992, opened 2023-09-21
Convert documentation examples into testsuite tests


lgritz:
The [documentation](https://openimageio.readthedocs.io/) contains lots of code examples embedded in the text. Some of them have bugs or are incorrect in various ways, either because they were never correct (having been typed in for the docs themselves and never directly tested), or were once correct but now the APIs have changed and the example was never updated to match.

To combat this, we would like to go through all the documentation, and to the greatest extent possible, convert every significant code example in the docs into a test -- that is, into actual code that will compile and run every time we run the testsuite -- and then the documentation will merely include the test code by reference, so they can never get out of sync again.

Explanation: https://github.com/OpenImageIO/oiio/wiki/Converting-documentation-examples-to-tests

There are dozens and dozens of code examples in the docs that we could do this with. Small steps -- do this for one or two at a time, that's fine.

aryangavit:
Hey I would like to work on this, can you assign me this issue.

lgritz:
I'm not going to formally assign the issue to anyone, because no one's piece of it is going to close the issue (until there are literally no more examples left to convert in the docs). This will take probably dozens of PRs (ad people, maybe) before it's truly complete.

Please feel free to work on it. Maybe leave a comment here saying which particular example(s) you want to do, so that others who also want to contribute to the task won't pick the same specific bit that you did.

aryangavit:
> I'm not going to formally assign the issue to anyone, because no one's piece of it is going to close the issue (until there are literally no more examples left to convert in the docs). This will take probably dozens of PRs (ad people, maybe) before it's truly complete.
> 
> Please feel free to work on it. Maybe leave a comment here saying which particular example(s) you want to do, so that others who also want to contribute to the task won't pick the same specific bit that you did.

Hi, I'm a beginner in open source and excited to learn about testsuits, I'll learn more about them and hoping to give it my all in fixing the documentation. If you guys don't mind I'll start with `Image Output: Writing Images` section code snippets.

lgritz:
I would recommend that anybody wanting to leave a note here declaring that they want to work on an example pick ONE specific code example to start with, and drive that to a complete accepted PR, before declaring a whole section or chapter off-limits to others. Get the hang of the process with something small, figure out how rapidly you can do the conversions, and then for round 2, you can reserve a larger section that might be a few days of work for you.

buddly27:
[...]

PR #4016, opened 2023-10-12
docs: Convert examples within the imagebufalgo chapter.

buddly27:
## Description
Convert C++ and Python examples from the "imagebufalgo" chapter into tests within the "docs-examples" testsuites (#3992).

## Checklist:

  nonconforming.

### DIFF
--- a/src/doc/imagebufalgo.rst
+++ b/src/doc/imagebufalgo.rst
@@ -37,24 +37,22 @@ Most ImageBufAlgo functions that produce image data come in two forms:
    case, an entirely new image will be created to hold the result. In case of
    error, the result image returned can have any error conditions checked with
    `has_error()` and `geterror()`.
-   
-   .. tabs::
-   
-      .. code-tab:: c++
-   
-          // Method 1: Return an image result
-          ImageBuf dst = ImageBufAlgo::over (fg, bg);
-          if (dst.has_error())
-              std::cout << "error: " << dst.geterror() << "\n";
 
-      .. code-tab:: py
-   
-          # Method 1: Return an image result
-          fg = ImageBuf("fg.exr")
-          bg = ImageBuf("bg.exr")
-          dst = ImageBufAlgo.over (fg, bg)
-          if dst.has_error() :
-              print("error:", dst.geterror())
+   .. tabs::
+
+      .. tab:: C++
+            .. literalinclude:: ../../testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp
+             :language: c++
+             :start-after: BEGIN-imagebufalgo-output-error1
+             :end-before: END-imagebufalgo-output-error1
+             :dedent: 4
+
+      .. tab:: Python
+         .. literalinclude:: ../../testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
+             :language: py
+             :start-after: BEGIN-imagebufalgo-output-error1
+             :end-before: END-imagebufalgo-output-error1
+             :dedent: 4
 
 
 2. Pass a destination ImageBuf reference as the first parameter.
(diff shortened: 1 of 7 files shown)


src/doc/imagebufalgo.rst
```diff
@@ -1075,7 +1071,7 @@ Shuffling channels
 
     .. tabs::
   
-       .. code-tab:: c++
+       .. code-tab:: c++`
```
MichaelRCarroll-Intel, line 1074: Is this backtick a typo?... or perhaps there is something I'm not seeing something about needed syntax in .rst?

Thanks for the diff... Seeing everything should make mine quicker.

buddly27: Definitely a typo, thanks for checking, I'll remove it on the next pass!

testsuite/docs-examples-cpp/ref/out.txt
```diff
@@ -1,3 +1,15 @@
+error: Uninitialized input image
+error: Uninitialized input image
```
lgritz, line 2: I'm a little concerned about these lines... is something in an example not working?

testsuite/docs-examples-cpp/run.py
```diff
@@ -11,7 +11,7 @@
     prefix = "./build/"
 
 # Prep:
-command += run_app("cmake -E copy ../common/grid-small.exr grid.exr")
+command += run_app("cmake -E copy ${OIIO_TESTSUITE_ROOT}/common/grid-small.exr grid.exr")
```
lgritz, line 14: Why this change?

testsuite/docs-examples-cpp/ref/out.txt
lgritz: Oh, never mind, those are from the ones where you are purposely making an error and showing how to check it, right?

testsuite/docs-examples-python/run.py
```diff
@@ -6,7 +6,7 @@
 
 
 # Prep:
-command += run_app("cmake -E copy ../common/grid-small.exr grid.exr")
+command += run_app("cmake -E copy ${OIIO_TESTSUITE_ROOT}/common/grid-small.exr grid.exr")
```
lgritz, line 9: Did something not work before?

I'm not sure this is what you want here. The test will be run from within `build/testsuite/<testname>`, and it really is ../common where the grid-small.exr is located.

I believe OIIO_TESTSUITE_ROOT is the *source* testsuite, not the copy in build, if you know what I mean.

lgritz:
Other than the comment about OIIO_TESTSUITE_ROOT, this is all looking really good! You didn't get tripped by anything I might have expected. Excellent work!

testsuite/docs-examples-cpp/ref/out.txt
buddly27: This is the result of testing the error outputs:
https://openimageio.readthedocs.io/en/v2.5.4.0/imagebufalgo.html#return-values-and-error-messages

I reproduced the tests, which print out "error: ..." for the two first examples:
https://github.com/buddly27/OpenImageIO/blob/pr3992/testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp#L46
https://github.com/buddly27/OpenImageIO/blob/pr3992/testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp#L59

This could potentially be tested differently:
```cpp
// BEGIN-imagebufalgo-output-error2
// Method 2: Write into an existing image
ImageBuf dst;   // will be the output image
bool ok = ImageBufAlgo::over (dst, fg, bg);
std::string error;
if (! ok)
    error = dst.geterror();
// END-imagebufalgo-output-error2

static_assert(!error.empty(), "An error should have been set");
```

Would you prefer that?

testsuite/docs-examples-python/run.py
buddly27: It was because the relative path resolution was failing when running "ctest" as follows:
```console
ctest --test-dir ./build 
```

I just now realized you provided a way to run single tests with `make test TEST=...`, I thought using an absolute path would offer more flexibility. Also, being in "source" is not necessarily a blocker, as the copied image will be there:
https://github.com/buddly27/OpenImageIO/blob/pr3992/testsuite/common/grid-small.exr

We could also use a CMAKE variable (e.g. [CMAKE_CURRENT_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html)) but I'm not entirely sure if it would be resolved as expected within the python scripts.

testsuite/docs-examples-cpp/ref/out.txt
lgritz: No, I think it's fine as it is. I just saw the error message in the captured output and momentarily forgot that this was expected.

testsuite/docs-examples-python/run.py
lgritz: The case I'm actually concerned about is that sometimes (willing to entertain the notion that it's nonstandard and not necessarily a good idea) I do this to run a test by hand, completely outside of ctest/cmake/make, usually when I'm tracking down some kind of broken behavior:

    cd build/testsuite/foo
    python3 ../runtest.py

But I'm not overly worried about this. I can probably find another way to do the kind of debugging I do this way.

lgritz: But my point is, if you execute the test script by hand outside of ctest, the OIIO_TESTSUITE_ROOT environment variable will not be set!

buddly27: Oh, I see, this will be a problem indeed. Then what about using the "test_source_dir" variable that is defined within `runtest.py`?
```python
command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr grid.exr")
```

lgritz: sure, I think that's fine

buddly27: Updated!

testsuite/docs-examples-cpp/run.py
buddly27: Related to other [note](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4016#discussion_r1357463066)

lgritz (approved):
LGTM, this is great!

Issue #287, opened 2012-04-10
iv - Multiple windows


lgritz:
iv should allow the current image to be split off into a separate screen window in the same iv process.

lgritz:
Coalescing.

PR #4017, opened 2023-10-12
feat: split off the current image in iv into a separate window (#2058)

antond-weta:
## Description

This change adds a menu item "Split off" under the menu "File" which creates a new window and initialises it with the currently displayed image. 

## Tests

I have tested the UI changes on Mac and linux, I don't have a Windows machine currently. 
No other functionality should be affected by this change.

## Checklist:

  nonconforming.

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -141,7 +141,7 @@ ImageViewer::ImageViewer()
     setWindowTitle(tr("Image Viewer"));
     resize(m_default_width, m_default_height);
     //    setSizePolicy (QSizePolicy::Ignored, QSizePolicy::Ignored);
-    
+
     setAttribute(Qt::WA_DeleteOnClose);
 }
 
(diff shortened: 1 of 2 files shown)


antond-weta:
The new window gets initialised with the default settings. I'm not sure how much of the current window's viewer state we want to retain.

lgritz:
Nice! Instead of "split off", I would suggest calling the menu item "Move to new window", that seems to be what the equivalent operation is called in some other apps I just randomly sampled (Firefox, Chrome, Obsidian).

lgritz (changes requested):
I suggested a different name for the command, but other than that, this LGTM.

antond-weta:
> I suggested a different name for the command, but other than that, this LGTM.

Thanks Larry, I have renamed the command.
There are currently some EasyCLA issues happening on our side, they are being chased with LFX.

lgritz (approved):
LGTM! Just awaiting CLA to merge.

Note that #2058 was an aggregation of many scattered issues. This suggestion for this feature originally came from #287, filed in 2012!

lgritz:
Aha, your CLA is working now!

@antond-weta, there appear to be a couple very minor clang-format fixes needed. As soon as that's ready, we're definitely ready to merge this.

PR #4019, opened 2023-10-13
Added tests for Simple Image input and updated rst

CalvinSch:
## Description

Added the simple input tests from the ImageInput chapter of the docs as C++ and Python examples, a part of the larger issue #3992 . Followed prototype (https://github.com/AcademySoftwareFoundation/OpenImageIO/wiki/Converting-documentation-examples-to-tests) approach.

## Checklist:

  nonconforming.

### DIFF
--- a/testsuite/docs-examples-cpp/ref/out.txt
+++ b/testsuite/docs-examples-cpp/ref/out.txt
@@ -29,9 +29,9 @@ lines.exr            :  640 x  480, 4 channel, float openexr
 box.exr              :  640 x  480, 4 channel, float openexr
     SHA-1: 6CBE5E98B145C5975CB930A91BAEE981C30E6B32
 text1.exr            :  640 x  480, 3 channel, float openexr
-    SHA-1: 014ECFC5EBF07F77DF24A592F43BC7CB101534AE
+    SHA-1: 72E989716A14F0399886CE39B04B92C20BCB3242
 text2.exr            :  640 x  480, 3 channel, float openexr
-    SHA-1: 53359E96A286F909A89ACC99A67A9ED3BADC4A7A
+    SHA-1: 7210D03C177F8DD854B2BA70FB298BB6287AC2E7
 cshift.exr           :  256 x  256, 4 channel, half openexr
     SHA-1: 000F95FDC44D4DBDA8B4041C2506149C7AE28ACA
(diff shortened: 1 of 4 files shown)


testsuite/docs-examples-cpp/ref/out.txt
```diff
@@ -29,9 +29,9 @@ lines.exr            :  640 x  480, 4 channel, float openexr
 box.exr              :  640 x  480, 4 channel, float openexr
     SHA-1: 6CBE5E98B145C5975CB930A91BAEE981C30E6B32
 text1.exr            :  640 x  480, 3 channel, float openexr
-    SHA-1: 014ECFC5EBF07F77DF24A592F43BC7CB101534AE
+    SHA-1: 72E989716A14F0399886CE39B04B92C20BCB3242
 text2.exr            :  640 x  480, 3 channel, float openexr
-    SHA-1: 53359E96A286F909A89ACC99A67A9ED3BADC4A7A
+    SHA-1: 7210D03C177F8DD854B2BA70FB298BB6287AC2E7
 cshift.exr           :  256 x  256, 4 channel, half openexr
     SHA-1: 000F95FDC44D4DBDA8B4041C2506149C7AE28ACA
```
lgritz, line 36: I don't think these should have changed, since you didn't change any of the code that output those files.

lgritz: I believe it's this change that is causing all the CI failures.

lgritz (changes requested):
It all looks correct to me (and I built the docs on my end, looks good!) except for the CI failures that are due to those unnecessary checksum changes. I assume maybe you were fooling around with something as you experimented, backed out of it, but accidentally checked in the changed out.txt file that was meant to be temporary?

testsuite/docs-examples-cpp/ref/out.txt
CalvinSch: Thanks for pointing this out! My mistake. I think these out.txt values were being changed after I picked up other docs test suite changes and didn't realize my code wasn't modifying these outputs. After running the test suite on my end I noticed these values changed and copied over the differences without giving it much thought.

I reverted these back in a new commit!

CalvinSch (commented):
Reverted ref/out.txt values to address CI issues

lgritz (approved):
LGTM!

PR #4022, opened 2023-10-14
build: Make C++17 be the default (C++14 is still the minimum for now)

lgritz:
This is for master only! Will not be backported to any release branch.

We are assuming the 2.5 release family will be the last to support C++14, and the next year's 2.6 family will raise the minimum to C++17. We're not ready to throw that switch yet, but as an intermediate step, let's raise the *default* level to C++17. It's still possible to compile for C++14, but to do so, one must proactively `-DCMAKE_CXX_STANDARD=14`.

This has two benefits:

1. The majority of people who are already using a C++17-capable compiler will take advantage of those features by default, instead of needlessly restricting to C++14.

2. The people using a compiler that tops out at C++14 (e.g., gcc 6.3) are going to have to set `-DCMAKE_CXX_STANDARD=14`, thus making them aware that (in master anyway) they're swimming against the current and have a limited time in which that will continue to work.

I hope that some time around the new year, we'll be able to fully raise the minimum to 17? But it's going to depend on my place and other major users finally retiring the DCC versions that require a gcc6/C++14 ecosystem, and we're not quite ready to do that at this moment. Soon, soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -369,7 +369,7 @@ help:
 	@echo "      STOP_ON_WARNING=0        Do not stop building if compiler warns"
 	@echo "      OPENIMAGEIO_SITE=xx      Use custom site build mods"
 	@echo "      MYCC=xx MYCXX=yy         Use custom compilers"
-	@echo "      CMAKE_CXX_STANDARD=14    C++ standard to build with (default is 14)"
+	@echo "      CMAKE_CXX_STANDARD=14    C++ standard to build with (default is 17)"
 	@echo "      USE_LIBCPLUSPLUS=1       For clang, use libc++"
 	@echo "      GLIBCXX_USE_CXX11_ABI=1  For gcc, use the new string ABI"
 	@echo "      EXTRA_CPP_ARGS=          Additional args to the C++ command"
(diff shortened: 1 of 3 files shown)


ThiagoIze (approved):
This seems reasonable to me.

jessey-git:
I think this is ok as well.  There's 2 additional locations though. Should those change as well?
```
testsuite\cmake-consumer\CMakeLists.txt:16:set (CMAKE_CXX_STANDARD 14 CACHE STRING "C++ standard to prefer (14, 17, etc.)")

testsuite\docs-examples-cpp\CMakeLists.txt:16:set (CMAKE_CXX_STANDARD 14 CACHE STRING "C++ standard to prefer (14, 17, etc.)")
```

lgritz:
@jessey-git I don't think so? These are testsuite entries that are intentionally independent of the OIIO build system itself. So there's a real sense that it's helping to test that from a user program perspective, software can still be built against OIIO at the lowest supported C++ level (even if OIIO itself was built at a higher level, if you know what I mean).

jessey-git (approved):

PR #4023, opened 2023-10-15
build: Protect against mismatch of OpenVDB vs C++

lgritz:
Looks like OpenVDB >= 10.0 requires a minimum of C++17. Warn at build time if trying to use an OpenVDB too new for the C++ compiler being used.


ThiagoIze (approved):
https://github.com/AcademySoftwareFoundation/openvdb/commit/bfe4452b23acc916c96e41eba69131c772eed019 is where doc/build.txt was updated to say that C++17 is required (their docs probably should be updated to say this is a minimum requirement or what range of C++ are allowed?), and that page says this change existed in v10.0.0. So disabling OpenVDB for 10.0 and newer seems good to me.

PR #4024, opened 2023-10-15
testing: Add opencv regression test

lgritz:
Testing to make sure we can convert an IC-backed IB to a cv::Mat.

Related to issue #3800 and PR #4013


lgritz:
Seem ok?

grdanny (approved):
lgtm

PR #4026, opened 2023-10-16
ci: CI tests on MacOS ARM, and fixes found consequently

lgritz:
There's a new GitHub runner in town!

Set up CI to run on Mac ARM. But not all the time -- it's too expensive!

This runner is only available for paid accounts, and it's 20x more expensive than paying for a Linux runner. The ASWF can't afford to run this job on every iteration of every PR, and again upon merge.

We split it into a separate workflow so we can direct it to run only for tagged releases, branches with names containng "-alpha", "-beta", "release", or contain the substring "macarm" (this last one is so we can force a particular PR that we think is directly relevant to ARM to test, even though it's not one of those other release-related cases). We also run it once a month, on the 27th -- a few days before each monthly release -- to try to catch any ARM problems we've accidentally introduced by PRs that didn't test.

It's also rigged to run only on the ASWF repo, so you don't need to worry that you will get charged for pushes to your own fork. (On the other hand, it also means you aren't testing against ARM when you push to your fork prior to making a PR.)

In the process of testing this, of course I found some minor things that needed to be fixed on ARM -- a couple compile issues, and several testsuite reference outputs that needed to be updated because they differ ever so slightly in some math calculations on the different CPU architecture.


grdanny:
Does it matter that one of the checks is failing:
https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/6528102307/job/17723826211?pr=4026

lgritz:
No, that's the Intel icx compiler test. It's failing because two weeks ago, Intel changed the signing keys on their yum repo, but didn't re-sign the packages for all the older versions of the compiler. They're working on it, is my understanding. The icc one was also broken until a few days ago. So I'm hoping they fix the icx one soon, too.

grdanny (approved):
lgtm

PR #4027, opened 2023-10-16
make_texture doc2test

grdanny:
## Description

Takes the make_texture section of the imagebufalgo doc and turns it into a unit test in both python and c++.

## Tests

This adds a make_texture section to the `docs-examples-cpp` test and the `docs-examples-python` test.

## Checklist:

- [x]  I have ensured that the change is tested somewhere in the testsuite
  nonconforming.

### DIFF
--- a/src/doc/imagebufalgo.rst
+++ b/src/doc/imagebufalgo.rst
@@ -2925,41 +2925,31 @@ Import / export
 
   Examples:
 
-    .. tabs::
-  
-       .. code-tab:: c++
+   .. tabs::
 
-          ImageBuf Input ("in.exr");
-          ImageSpec config;
-          config["maketx:highlightcomp"] = 1;
-          config["maketx:filtername"] = "lanczos3";
-          config["maketx:opaque_detect"] = 1;
-      
-          bool ok = ImageBufAlgo::make_texture (ImageBufAlgo::MakeTxTexture,
-                                                Input, "texture.exr", config);
-          if (! ok)
-              std::cout << "make_texture error: " << OIIO::geterror() << "\n";
+      .. tab:: C++
+         .. literalinclude:: ../../testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp
+              :language: c++
+              :start-after: BEGIN-imagebufalgo-make-texture
+              :end-before: END-imagebufalgo-make-texture
+              :dedent: 4
 
-       .. code-tab:: py
+      .. tab:: Python
+         .. literalinclude:: ../../testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
+              :language: py
+              :start-after: BEGIN-imagebufalgo-make-texture
+              :end-before: END-imagebufalgo-make-texture
+              :dedent: 4
 
-          Input = ImageBuf("in.exr")
-          config = ImageSpec()
-          config["maketx:highlightcomp"] = 1
-          config["maketx:filtername"] = "lanczos3"
-          config["maketx:opaque_detect"] = 1
-      
-          ok = ImageBufAlgo.make_texture (OpenImageIO.MakeTxTexture,
-                                          Input, "texture.exr", config)
-          if not ok :
-              print("make_texture error:", OpenImageIO.geterror())
+      .. tab:: oiiotool
+         .. sourcecode:: bash
+         
+            oiiotool in.exr -otex:hilightcomp=1:filtername=lanczos3:opaque_detect=1 texture.exr
 
-       .. code-tab:: bash oiiotool
+      .. tab:: maketx
+         .. sourcecode:: bash
 
-          oiiotool in.exr -otex:hilightcomp=1:filtername=lanczos3:opaque_detect=1 texture.exr
-
-       .. code-tab:: bash maketx
-
-          maketx in.exr --hicomp --filter lanczos3 --opaque-detect -o texture.exr
+            maketx in.exr --hicomp --filter lanczos3 --opaque-detect -o texture.exr
       
 |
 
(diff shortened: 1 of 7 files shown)


testsuite/docs-examples-cpp/run.py
```diff
@@ -13,6 +13,7 @@
 # Prep:
 command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr grid.exr")
 command += run_app("cmake -E copy " + test_source_dir + "/../common/tahoe-small.tif tahoe.tif")
+command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr texture_input.exr")
```
lgritz, line 16: There's already a copy of grid-small.exr, named grid.exr. Why not just use that as the input?

testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp
```diff
@@ -267,6 +267,23 @@ void example_circular_shift()
 // Section: Import / export
 
 
+void example_make_texture()
+{
+    // BEGIN-imagebufalgo-make-texture
+    ImageBuf Input ("texture_input.exr");
```
lgritz, line 273: Just change this to grid.exr, and you won't need the extra copy.

It's ok to alter the examples if you think there's a way to clarify or simplify them.

testsuite/docs-examples-python/run.py
```diff
@@ -8,6 +8,7 @@
 # Prep:
 command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr grid.exr")
 command += run_app("cmake -E copy " + test_source_dir + "/../common/tahoe-small.tif tahoe.tif")
+command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr texture_input.exr")
```
lgritz, line 11: same here

src/doc/imagebufalgo.rst
```diff
@@ -2926,32 +2926,20 @@ Import / export
...
-          if (! ok)
-              std::cout << "make_texture error: " << OIIO::geterror() << "\n";
 
-       .. code-tab:: py
+         .. tab:: C++
+             .. literalinclude:: ../../testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp
+              :language: c++
+              :start-after: BEGIN-imagebufalgo-make-texture
+              :end-before: END-imagebufalgo-make-texture
+              :dedent: 4
```
lgritz, line 2935: It turns out that you can't have a mix of `tab` and `code-tab`. It looks ok at first, but things go wrong when you switch among the tabs. I think that in order to look right, this is how you want this section (minor edits):

```
  .. tabs::

     .. tab:: C++
           .. literalinclude:: ../../testsuite/docs-examples-cpp/src/docs-examples-imagebufalgo.cpp
            :language: c++
            :start-after: BEGIN-imagebufalgo-make-texture
            :end-before: END-imagebufalgo-make-texture
            :dedent: 4

     .. tab:: Python
        .. literalinclude:: ../../testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
            :language: py
            :start-after: BEGIN-imagebufalgo-make-texture
            :end-before: END-imagebufalgo-make-texture
            :dedent: 4

     .. tab:: oiiotool
        .. sourcecode::

          oiiotool in.exr -otex:hilightcomp=1:filtername=lanczos3:opaque_detect=1 texture.exr

     .. tab:: maketx
        .. sourcecode::

          maketx in.exr --hicomp --filter lanczos3 --opaque-detect -o texture.exr
```

lgritz (changes requested):
Minor improvements suggested

src/doc/imagebufalgo.rst
grdanny: Ah - that makes sense, I was actually building the docs too and testing that these tabs works and sometimes some would disappear on me.
Let me test with this more consistent way and see if that sticks more robustly.

For a second there I was thinking of maybe trying to add yet another test that tests the bash (shell) commands, but it was tricky to get to a clean "oiiotool" command that'll work across all platforms in a bash script.
It would be neat if we could do that because that would make the doc consistent across all the tabs (and provide some test coverage for the maketx/oiio flags).

lgritz: Yeah, there are some kinks still to work out with the best way to do this kind of think for the oiiotool examples, but we'll come back to that later.

testsuite/docs-examples-cpp/run.py
grdanny: No real reason, grid is just as good.
It was just to get the doc to look a certain way, but grids make good textures too.

src/doc/imagebufalgo.rst
grdanny: Tried your suggestion out, it does look more stable now when I test the docs, I see we have a few examples in there already that mix tab/codetab (e.g imagebufalgo-zero).
So there is a bit more cleanup to do, but we can do that in a separate PR

testsuite/docs-examples-cpp/run.py
lgritz: I'm not sure I understand this comment, Danny. Both before and after, you were using the grid-small image. But in one case, you copied it to be named "texture_input" rather than already using the identical file that was named "grid.exr".

lgritz:
I think you still don't have a DCO sign-off on one of the commits, can you fix that, please?

grdanny:
Rolled everything into a single signed commit. Thanks!

lgritz (approved):
LGTM!

testsuite/docs-examples-cpp/run.py
grdanny: Yeah - I was just taking small-grid and renaming it.
It was just gymnastics so that the doc would say:
```
Input = ImageBuf("texture_input.exr")
```

But you are right, the doc is clear enough with:
```
Input = ImageBuf("grid.exr")
```

PR #4028, opened 2023-10-16
LibRaw wavelet denoise options

ssh4net:
``raw:threshold``
     - float
     - Libraw parameter for noise reduction through wavelet denoising..
       The best threshold should be somewhere between 100 and 1000.
- ``raw:fbdd_noiserd`` - int - Controls FBDD noise reduction before demosaic. 0 - do not use FBDD noise reduction, 1 - light FBDD reduction, 2 (and more) - full FBDD reduction

## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/CONTRIBUTING.md).
  nonconforming.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1926,8 +1926,15 @@ options are supported:
      - int
      - Set libraw user flip value : -1 ignored, other values are between [0; 8] with the same 
        definition than the Exif orientation code.
-
-
+   * - ``raw:threshold``
+     - float
+     - Libraw parameter for noise reduction through wavelet denoising.. 
+       The best threshold should be somewhere between 100 and 1000.
+   * - ``raw:fbdd_noiserd``
+     - int
+     - Controls FBDD noise reduction before demosaic.
+       0 - do not use FBDD noise reduction, 1 - light FBDD reduction,
+       2 (and more) - full FBDD reduction
 
 
 |
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1936,7 +1936,7 @@ options are supported:
      - Controls FBDD noise reduction before demosaic.
        0 - do not use FBDD noise reduction, 1 - light FBDD reduction,
        2 (and more) - full FBDD reduction
-       (Default: 0.0)
+       (Default: 0)
 
 
 |
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1928,13 +1928,15 @@ options are supported:
        definition than the Exif orientation code.
    * - ``raw:threshold``
      - float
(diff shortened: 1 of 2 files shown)


ssh4net:
@lgritz nothing special just two options:
raw:threshold and raw:fbdd_noiserd
If the first one clearly visible results. Recommended to use values from 100.00 to 1000.00 

but fbdd_noiserd I did not get how it works in libraw so just added it as it is also about denoising (1 or 2 should enable different modes of this pre-demosaic denoiser).

My main concern, that compared to direct libraw demosaic that not affect processing time a lot. OpenImageIO has 2-3 times slowdown on wavelet demosaic,
But i don't see anything suspicious except the debugger 3 or 4 times passing this libraw options. First time on metadata read, and set nothing but last three (or more?) times it already has actual settings set.
But worth to mention, looks like demosaic did only once. So i can't get where this slowdown in oiio happen.

src/doc/builtinplugins.rst
```diff
@@ -1926,8 +1926,15 @@ options are supported:
      - int
      - Set libraw user flip value : -1 ignored, other values are between [0; 8] with the same 
        definition than the Exif orientation code.
-
-
+   * - ``raw:threshold``
+     - float
+     - Libraw parameter for noise reduction through wavelet denoising..
```
lgritz, line 1931: Minor correction: `..` -> `.`

lgritz:
Check the CI run to see that clang-format found a couple of very minor things, but other than that, it looks good to me.

ssh4net:
Fixed.
I also added info about default values.

Do you have any ideas why enabling this pre-demosaic denoising can have such a significant impact on plugin speed when libraw directly has only 5-10% slowdown?

src/doc/builtinplugins.rst
```diff
@@ -1926,8 +1926,17 @@ options are supported:
...
+     - float
+     - Libraw parameter for noise reduction through wavelet denoising.
+       The best threshold should be somewhere between 100 and 1000.
+       (Default: 0.0)
+   * - ``raw:fbdd_noiserd``
+     - int
+     - Controls FBDD noise reduction before demosaic.
+       0 - do not use FBDD noise reduction, 1 - light FBDD reduction,
+       2 (and more) - full FBDD reduction
+       (Default: 0.0)
```
lgritz, line 1939: I'm sorry, last thing, I promise! Since this one is an integer parameter, let's document the default as 0 instead of 0.0.

lgritz, line 1939:
```suggestion
       (Default: 0)
```

lgritz (approved):
LGTM!

(I fixed the 0.0 -> 0 suggestion for you)

PR #4032, opened 2023-10-19
docs: Fix RTD configuration for v2

lgritz:
RTD recently cut off old configuration file style and these additions were needed.


lgritz:
Preemptively merging because it's only docs and it verifiably fixes something broken.

PR #4034, opened 2023-10-24
Deprecate Sysutil::physical_concurrency()

lgritz:
It is a implemented by simply wrapping boost::physical_concurrency(). So this is one more step in eliminating boost from our requirements.

Until it's fully removed, just make it be a synonym for hardware_concurrency().

We don't seem to use physical_concurrency anywhere, and I'm not sure what it's good for. The underlying boost impementation just falls back to hardware_concurrency on many platforms or circumstances.

Note for the record that Sysutil::hardware_concurrency() is also redundant, and merely wraps the std::thread function of the same name.

### DIFF
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -86,15 +86,24 @@ OIIO_API bool
 put_in_background(int argc, char* argv[]);
 
 /// Number of virtual cores available on this platform (including
-/// hyperthreads).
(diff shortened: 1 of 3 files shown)


ThiagoIze (approved):
Looks good to me.

lgritz:
I went all the way down the road of fully implementing it (inspired by, but not copying, the boost implementation -- turns out there wasn't much to it after all). But then realized that we don't use it in the code base, and I'm not sure how or why I'd use it anywhere, so I decided to just deprecate it instead.

PR #4036, opened 2023-10-27
ci: Nomenclature change 'os' to 'runner' for clarity

lgritz:


lgritz:
No objections -> merging

PR #4037, opened 2023-10-29
build: Better cmake verbose behavior

lgritz:
When we bumped to cmake 3.15 minimum for 2.5, we changed lots of constructs like this:

    if (VERBOSE)
        message (STATUS "blah")
    endif ()

to this:

    message (VERBOSE "blah")

but I didn't realize that this was not enough, and to do it right requires CMAKE_MESSAGE_LOG_LEVEL. So even when VERBOSE=ON, we weren't actually getting all the verbose messages. Here, we fix it.

Also, when verbose, print the list of C++ compile features that cmake discovers.


lgritz:
No objections -> merging

PR #4048, opened 2023-11-02
fix: always seek to the beginning of the ioproxy during open for DDS and PSD files

### LINKED ISSUE #4043: [BUG] `ImageBuf` not working with in-memory DDS files
I am trying to create `ImageBuf` using an `IOMemReader` pointing to an in-memory DDS file. However, `ImageBuf::init_spec` fails with `Invalid DDS header, possibly corrupt file`. `ImageInput::open` works fine with `IOMemReader`. The problem seems to be specific to DDS format, I could not reproduce with PNG.

Code for reproduction:
```c++
#include <OpenImageIO/filesystem.h>

#include <fstream>
#include <iostream>

namespace OIIO = OpenImageIO_v2_4;

void fails(const std::string& ddsData)
{
	OIIO::Filesystem::IOMemReader memReader(ddsData.data(), ddsData.size());
	OIIO::ImageBuf buf("not_existing_on_disk.dds", 0, 0, nullptr, nullptr, &memReader);

	if (!buf.init_spec(buf.name(), 0, 0))
	{
		std::cout << "init_spec failed: " << buf.geterror() << std::endl;
		return;
	}

	std::cout << "init_spec succeeded; size: " << buf.spec().width << " x " << buf.spec().height << std::endl;
}

void works(const std::string& ddsData)
{
	OIIO::Filesystem::IOMemReader imageReader(ddsData.data(), ddsData.size());
	auto in = OIIO::ImageInput::open("not_existing_on_disk.dds", nullptr, &imageReader);
	if (!in)
	{
		std::cout << "ImageInput::open failed" << std::endl;
		return;
	}

	std::cout << "ImageInput::open succeeded; size: " << in->spec().width << " x " << in->spec().height << std::endl;
}

int main()
{
	std::fstream fs("example.dds");
	const std::string ddsData((std::istreambuf_iterator<char>(fs)), std::istreambuf_iterator<char>());

	fails(ddsData);
	works(ddsData);

	return 0;
}
```

The result I get:
```
init_spec failed: Invalid image file "not_existing_on_disk.dds": Invalid DDS header, possibly corrupt file
ImageInput::open succeeded; size: 512 x 256
```

The result I _expect_:
```
init_spec succeeded; size: 512 x 256
ImageInput::open succeeded; size: 512 x 256
```

I am on Debian Testing, using OIIO version 2.4.14 (distro package). Here's the whole project along with `example.dds`:
[openimageio_test.zip](https://github.com/AcademySoftwareFoundation/OpenImageIO/files/13228237/openimageio_test.zip)

jessey-git:
## Description

During certain creation tasks, open may be called twice on the ImageInput. If the first open call moves the ioproxy forward, then the second call to open may fail.

DDS and PSD were affected so they've been fixed to mirror what other formats (png, hdr, tga, etc.) do by always explicitly seeking to position 0.

Fixes #4043 

## Tests

The example code from the bug now works.

## Checklist:

  nonconforming.

### DIFF
--- a/src/dds.imageio/ddsinput.cpp
+++ b/src/dds.imageio/ddsinput.cpp
@@ -357,6 +357,7 @@ DDSInput::open(const std::string& name, ImageSpec& newspec)
 
     if (!ioproxy_use_or_open(name))
         return false;
+    ioseek(0);
 
     static_assert(sizeof(dds_header) == 128, "dds header size does not match");
     if (!ioread(&m_dds, sizeof(m_dds), 1))
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -546,6 +546,7 @@ PSDInput::open(const std::string& name, ImageSpec& newspec)
     m_filename = name;
     if (!ioproxy_use_or_open(name))
         return false;
+    ioseek(0);
 
     // File Header
     if (!load_header()) {


lgritz (approved):
LGTM!

lgritz:
The CI failures look unrelated. Merging.

PR #4052, opened 2023-11-16
ci: Add tiff-misc reference for slightly changed error messages

lgritz:
Everything works, no code changes needed. But with some libtiff builds, there is some variation in exactly which error messages bubble up when reading certain corrupted files. This fixes some recently broken CI.


lgritz:
Merging, because this only affects CI and un-breaks most of the tests.

PR #4053, opened 2023-11-17
ci: Remove MacOS-11 test

lgritz:
It's started failing on GHA and is old enough that it's not worth trying to track down what's broken. We still have MacOS 12 and 13 tests in the matrix.


grdanny (approved):
lgtm

PR #4055, opened 2023-11-18
Initial JPEG XL support for image input/output

1div0:
## Description

This draft PR is intended for adding JPEG XL support.

## Checklist:

  nonconforming.


src/cmake/modules/FindJXL.cmake
```diff
@@ -0,0 +1,27 @@
+# Find libjxl
```
lgritz, line 1: Please put our usual 3-line copyright boilerplate here

src/cmake/modules/FindJXL.cmake
```diff
@@ -0,0 +1,27 @@
...
+# Will define:
+# - JXL_FOUND
+# - JXL_INCLUDE_DIRS directory to include for libjxl headers
```
lgritz, line 4: I believe the more standard name for this is JXL_INCLUDES

src/cmake/modules/FindJXL.cmake
```diff
@@ -0,0 +1,27 @@
...
+  NAMES jxl_threads)
+mark_as_advanced(JXL_THREADS_LIBRARY)
+
+find_package_handle_standard_args(JXL
+  REQUIRED_VARS JXL_LIBRARY JXL_THREADS_LIBRARY JXL_INCLUDE_DIR)
+
+if(JXL_FOUND)
+  set(JXL_LIBRARIES ${JXL_LIBRARY} ${JXL_THREADS_LIBRARY})
+  set(JXL_INCLUDE_DIRS ${JXL_INCLUDE_DIR})
+endif(JXL_FOUND)
```
lgritz, line 27: If this is all found, in addition to those variables, please also set up targets in the modern CMake style. You can find an example of this [in our FindLibheif.cmake module](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/src/cmake/modules/FindLibheif.cmake#L43). 

Also, it is important for this module to set a JXL_VERSION so we can check version range. A good example of how to extract this from header files can be found [in our FindOpenVDB.cmake](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/src/cmake/modules/FindOpenVDB.cmake#L48)

src/jxl.imageio/jxloutput.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+
+bool
+JxlOutput::open(const std::string& name, const ImageSpec& newspec,
+                OpenMode mode)
+{
+    // Save name and spec for later use
+    m_filename = name;
+
+    return true;
```
lgritz, line 88: If these are not functional, they should all return false, since they are not succeeding at writing the file.

I think you may also want to consider just not having a JxlOutput or a jxloutput.cpp file at all unless/until you plan to implement JXL writing. It's totally reasonable to get the reader working and merged first, then come back to the writer. (Also, there are some formats that we want to be able to read, but care a lot less about being able to write, so not all readers have corresponding writers. Whether JXL falls into this category is a choice I'll leave to you.)

src/jxl.imageio/jxl_pvt.h
```diff
@@ -0,0 +1,60 @@
...
+
+class JxlInput final : public ImageInput {
+public:
+    JxlInput() { init(); }
+    ~JxlInput() override { close(); }
+    const char* format_name(void) const override { return "jxl"; }
+    int supports(string_view feature) const override
+    {
+        return (feature == "exif" || feature == "ioproxy");
+    }
```
lgritz, line 24: I would say that by default, the definition for JxlInput class should be in jxlinput.cpp if that's the only place that reads it.

Some of our format plugins have a format_pvt.h that contains definitions that will be needed by both the input and output classes. You may want to wait until there is an implemented output (when you will know what needs to be shared) before making an extra header for the shared definitions.

lgritz (commented):
You have this marked as "draft", so I know you aren't seeking a final review, but I thought I'd add some comments at this stage. Mostly just some organizational tips.

I'm not commenting yet on the code in jxlinput.cpp, because I figure it's all still in-progress. I also assume that at some point, you will add something to the testsuite that exercises this code so we have confirmation that it correctly reads jxl files.

src/cmake/modules/FindJXL.cmake
lgritz: This comment still applies

src/jxl.imageio/README.md
```diff
@@ -0,0 +1,8 @@
+# JPEG XL
+
+* https://jpegxl.info
+* https://jpegxl.info/test-page
+* https://people.csail.mit.edu/ericchan/hdr/hdr-jxl.php
+* https://saklistudio.com/jxltests
+* https://thorium.rocks
+* https://bugs.chromium.org/p/chromium/issues/detail?id=1451807
```
lgritz, line 8: Let's put these in a comment in jxlinput.cpp, please

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,358 @@
...
+    switch (signature) {
+    case JXL_SIG_CODESTREAM:
+    case JXL_SIG_CONTAINER: break;
+    default: return false;
+    }
+
+    DBG std::cout << "JxlInput::valid_file() return true\n";
+    return true;
+}
+
```
lgritz, line 110: I know it's picky, but would you mind using our standard of 3 blank lines between top-level functions? It's not critical, but I think it makes the code easier to visually parse.

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,358 @@
...
+        return false;
+    }
+
+    std::vector<uint8_t> jxl;
+
+    DBG std::cout << "proxytype = " << proxytype << "\n";
+    if (proxytype == "file") {
+        size_t size = m_io->size();
+        DBG std::cout << "size = " << size << "\n";
+        jxl.resize(size);
```
lgritz, line 177: Turns out that std::vector.resize() will initialize the memory, which is a lot if this is a large image. I would recommend that instead you use:

    std::unique_ptr<uint8_t[]> jxl;
    ...
    jxl.reset(new uint8_t[size]);

If you only need that fixed size and aren't otherwise growing or shrinking the vector, it will all work just like the vector did, including auto-freeing when it leaves scope. But it won't needlessly touch the memory an extra time by zeroing it out.

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,358 @@
...
+        ImageSpec configsave;
+        if (m_config)
+            configsave = *m_config;
+        ImageSpec dummyspec;
+        // int subimage = current_subimage();
+        // if (!close() || !open(m_filename, dummyspec, configsave)
+        //     || !seek_subimage(subimage, 0))
+        //     return false;  // Somehow, the re-open failed
+        // OIIO_DASSERT(m_next_scanline == 0 && current_subimage() == subimage);
+    }
```
lgritz, line 338: I think this entire `if` clause can go away. It's not doing anything now.

src/jxl.imageio/jxloutput.cpp
```diff
@@ -0,0 +1,338 @@
+// Copyright Contributors to the OpenImageIO project.
+// SPDX-License-Identifier: BSD-3-Clause and Apache-2.0
```
lgritz, line 2: This shouldn't say BSD-3-Clause, just Apache-2.0 if it's new code

src/jxl.imageio/jxloutput.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+bool
+JxlOutput::copy_image(ImageInput* in)
+{
+    if (in && !strcmp(in->format_name(), "jxl")) {
+        return true;
+    }
+
+    return ImageOutput::copy_image(in);
+}
```
lgritz, line 336: I don't think this is correct at all. If I understand, what you're saying is that if it's copying from another jxl file, it DOES NOTHING but returns true indicating success? If you don't have a special way to copy from jxl to jxl efficiently without decompressing and recompressing, then you don't need to provide a copy_image method at all.

lgritz (changes requested):
This looks like it's getting very close. Only minor comments about the code itself, as you can see.

Big missing items:
* Documentation -- need a section in builtplugins.rst
* Tests -- the testsuite needs to exercise this in some way, like we do for the other formats.

lgritz:
The CI is failing on Mac. I bet that this is because the Mac CI tests are the only ones with libjxl installed at all, so they are the only ones building this code. That's probably something else to address, would be nice for at least one Linux build to try jxl. But maybe the reason it's failing on Mac is that it's using a different version of libjxl than you are, and something in their API changed?

1div0:
> The CI is failing on Mac. I bet that this is because the Mac CI tests are the only ones with libjxl installed at all, so they are the only ones building this code. That's probably something else to address, would be nice for at least one Linux build to try jxl. But maybe the reason it's failing on Mac is that it's using a different version of libjxl than you are, and something in their API changed?

Found JXL 0.9.1, while the latest is 0.10.1, which I am using now. I may add conditional compilation in order to have som backward compatibilty. Or wait until the newset release will be available everywhere? That takes ages.

src/jxl.imageio/jxlinput.cpp
1div0: Sure, good point. Will try.

lgritz:
Your call on how far back you want to support libjxl. You could set the MIN_VERSION in the checked_find_package call to reject a version that's too old. 

If you want 0.10 to be the minimum version, that is less work now in the sense that you don't need any `#if` now, though you may later when 0.11 or whatever comes out. We'll probably never again be in a position where we can dictate that the only version we support is the very latest.

On the minus side, 0.10 minimum would mean our CI wouldn't currently build and test JPEGXL at all. In that case, I think we would need a build_jxl.bash like the other files in src/build-scripts, which we can run during the CI setup to ensure it's available in the versions you want to test, and that also makes it handy for end users who want a simple way to install it so that OIIO can find it.

1div0:
Tried that already.

https://ffmpeg.org/pipermail/ffmpeg-devel/2023-July/311369.html

1div0:
> Your call on how far back you want to support libjxl. You could set the MIN_VERSION in the checked_find_package call to reject a version that's too old.
> 
> If you want 0.10 to be the minimum version, that is less work now in the sense that you don't need any `#if` now, though you may later when 0.11 or whatever comes out. We'll probably never again be in a position where we can dictate that the only version we support is the very latest.
> 
> On the minus side, 0.10 minimum would mean our CI wouldn't currently build and test JPEGXL at all. In that case, I think we would need a build_jxl.bash like the other files in src/build-scripts, which we can run during the CI setup to ensure it's available in the versions you want to test, and that also makes it handy for end users who want a simple way to install it so that OIIO can find it.

0.10 as the minimum required version seems to be the good choice, especially with the latest improvements:
https://cloudinary.com/blog/jpeg-xl-and-the-pareto-front

lgritz:
OK, I'm with with making jxl have a supported minimum of 0.10.

But that means that we should write a build_libjxl.bash script, and use it for at least one of our CI tests so that we are for sure building and testing this code during CI.

lgritz:
Guess what? Homebrew JUST upgraded their jpeg-xl package to 0.10.1. Let me force these CI tests to run again, it may run on the right version on the Mac CI, which would let us directly test you new coe.

lgritz:
Weird. I tried myself, and even with libjxl 0.10.1, I'm seeing errors related to JxlDecoderGetICCProfileSize and JxlDecoderGetColorAsICCProfile being passed wrong arguments -- in jxl/decoder.h, neither of those functions seem to take the `&format` argument you're passing. How are you getting it to work on your end?

jessey-git:
I have a few general questions about the plans for this area.

For output:
- It looks like this will always write out using a 32bit float format. Are there plans to allow the caller to decide what bit-depth and format to write with?
- There's no control over compression settings of any kind. That seems less than ideal.

For input:
- Apparently JXL allows both associated and unassociated alpha. How does this work in practice and does this mean it's even more important to at least respect the `oiio:UnassociatedAlpha` attribute?

I'll review the actual code a bit tomorrow.

1div0:
> Guess what? Homebrew JUST upgraded their jpeg-xl package to 0.10.1. Let me force these CI tests to run again, it may run on the right version on the Mac CI, which would let us directly test you new coe.



> Weird. I tried myself, and even with libjxl 0.10.1, I'm seeing errors related to JxlDecoderGetICCProfileSize and JxlDecoderGetColorAsICCProfile being passed wrong arguments -- in jxl/decoder.h, neither of those functions seem to take the `&format` argument you're passing. How are you getting it to work on your end?

It was unfortunate impedance mismatch between three versions of the libjxl I had installed. So cleaned up and checked against the latest and greatest official release.

1div0:
> Guess what? Homebrew JUST upgraded their jpeg-xl package to 0.10.1. Let me force these CI tests to run again, it may run on the right version on the Mac CI, which would let us directly test you new coe.

Just poured jpeg-xl--0.10.1.x86_64_linux.bottle.tar.gz and going to test against this version as well.

Silly idea: what is preventing us to use Homebrew installation on the Linux CI runs similarly to macOS?

src/jxl.imageio/jxlinput.cpp
1div0: Corrected in the latest iteration.

1div0:
> This looks like it's getting very close. Only minor comments about the code itself, as you can see.
> 
> Big missing items:
> 
>     * Documentation -- need a section in builtplugins.rst
> 
>     * Tests -- the testsuite needs to exercise this in some way, like we do for the other formats.

Color management is unfinished. Nor proper metadata handling. Still so much to do and so little free(time_t).

src/jxl.imageio/jxlinput.cpp
```diff
@@ -320,29 +344,17 @@ JxlInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
     lock_guard lock(*this);
     if (!seek_subimage(subimage, miplevel))
         return false;
-    if (y < 0 /* || y >= (int)m_cinfo.output_height*/)  // out of range scanline
+    if (y < 0)  // out of range scanline
```
lgritz, line 347: Should check for y too big here, too.

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,370 @@
...
+            }
+            // if (buffer_size != info.xsize * info.ysize * 16) {
+            if (buffer_size
+                != info.xsize * info.ysize * m_channels * sizeof(float)) {
+                fprintf(stderr,
+                        "Invalid out buffer size %" PRIu64 " %" PRIu64 "\n",
+                        static_cast<uint64_t>(buffer_size),
+                        // static_cast<uint64_t>(info.xsize * info.ysize * 16));
+                        static_cast<uint64_t>(info.xsize * info.ysize
+                                              * m_channels * sizeof(float)));
```
lgritz, line 293:
```suggestion
                print(stderr, "Invalid out buffer size {} {}\n", buffer_size,
                      info.xsize * info.ysize* m_channels * sizeof(float));
```

src/jxl.imageio/jxlinput.cpp
1div0: Is that correct now? Or shoud be greater or equal to height?

lgritz:
> Silly idea: what is preventing us to use Homebrew installation on the Linux CI runs similarly to macOS?

No particular reason, it just never came up. For CI, at least, we tend to try to use the "native" packager. So adding rpm/yum packages in gh-installdeps.bash is what we usually do for Linux. I'd be slightly afraid that a dependency from homebrew might not be representative of what they'd get installing packages the way most Linux users do -- different version, installed in a different place, who knows.

src/jxl.imageio/jxlinput.cpp
lgritz: `>=`.  If height is 10, the scanlines range from 0 to 9, so 10 would result in the subsequent memcpy misbehaving.

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+OIIO_PLUGIN_NAMESPACE_BEGIN
+
+#define DBG if (0)
+
+class JxlInput final : public ImageInput {
+public:
+    JxlInput() { init(); }
+    ~JxlInput() override { close(); }
+    const char* format_name(void) const override { return "jxl"; }
```
lgritz, line 36: Should this be jpeg-xl or jpegxl? We tend to make the "format name" be the name of the format, not necessarily the short file extension. For example, we use "jpeg", "openexr", "tiff", "targa".

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,370 @@
...
+    int supports(string_view feature) const override
+    {
+        return (feature == "exif" || feature == "ioproxy");
```
lgritz, line 39: Check the list of features in imageio.h's documentation of the `supports()` method. Are there others that you already support? Do you really already support exif now?

src/jxl.imageio/jxlinput.cpp
1div0: It should be JPEG XL as written here with non breaking space. No dash.

src/jxl.imageio/jxlinput.cpp
1div0: I intend to add Exif support based upon my experience with Exiv2 development. It is little bit complicated, but doable sans the Brotli compression of metadata.

lgritz:
Looks like we're building ok on Mac with jpeg-xl installed. So I think the minimum requirement for getting this initial PR merged into master are:

* Enforce the version limit in externalpackages.cmake.
* Fix the clang-format CI failure.
* Disable the format (temporarily) in imageinout_test -- we'll come back to this
* Resolve what you want to use for the format_name() method.

Then with this as a baseline, the next set of necessary improvements are (not necessarily in this order):

* Documentation
* Handling alpha correctly (associated vs unassociated)
* Track down what's going wrong with imageinout_test
* Adding a testsuite/jpegxl that tests basic read and write of jxl files, including a few combinations of supported features.
* Full support of all the pixel data types the jxl supports.
* Compression options.
* Right now, I think you are reading in the entire image at once into a (large) buffer and doling out pieces via memcpy as requested from the caller. I haven't read the libjxl docs -- does it support asking or passing ranges of scanlines or tiles? I don't know if the jxl format allows such things easily.
* Getting the `supports()` method reporting the right set of things.

I'm sure we'll think of other things, but it's ok to get the big hunk of initial experimental support in now (as long as it builds for everyone correctly and passes the tests we already have), and then can have a series of smaller PRs to complete support and bring it all up to parity with the other formats.

1div0:
> Looks like we build successfully on Mac now with jpeg-xl installed via Homebrew!
> 
> But we are failing the testsuite, specifically the imageinout_test. That may be tricky, and I'm ok with your modifying imageinout_test.cpp, circa line 324, to simply skip the jxl format for now. We can restore it in subsequent iterations over this code and addition of the tests.

I will try now the macOS Sonoma build in the QEMU KVM based virtual machine.

1div0:
Adding support for streaming mode comes to top of my wish list. Also, it would be cool to handle stereo and multiview images in this age.

lgritz:
Stereo and multiview should probably look like "subimages" in OIIO speak. Like how OpenEXR handles "parts" which are also used for stereo.

src/jxl.imageio/jxlinput.cpp
lgritz: I think I would prefer that it be lower case and not to use a space, since there are circumstances in which the format name is incorporated into filenames.

For that matter, I think all the others make the format_name match the "plugin name", we'll want to make sure this matches the references in imageioplugin.cpp.

I think "jpegxl" is the one you should use. You can't go wrong with being "C identifier safe."

src/libOpenImageIO/imageioplugin.cpp
```diff
@@ -280,6 +280,7 @@ PLUGENTRY(ico);
 PLUGENTRY(iff);
 PLUGENTRY(jpeg);
 PLUGENTRY(jpeg2000);
+PLUGENTRY(jxl);
```
lgritz, line 283: This should match what format_name() returns.

src/libOpenImageIO/imageioplugin.cpp
```diff
@@ -382,6 +383,9 @@ catalog_builtin_plugins()
 #if defined(USE_OPENJPEG) && !defined(DISABLE_JPEG2000)
     DECLAREPLUG (jpeg2000);
 #endif
+#if defined(USE_JXL)
+    DECLAREPLUG (jxl);
```
lgritz, line 387: This should match what format_name() returns.

src/jxl.imageio/jxlinput.cpp
lgritz: We should probably `git mv` the whole directory to be called src/jpegxl.imageio. Then plugin name, short format name, and format_name() will all match, as they do for the rest of the formats.

1div0:
It took bit longer to set up all dependencies in the virtual machine. However, I encountered this:

[ 17%] Built target ustring_test
[ 17%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo.cpp.o
[ 18%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_pixelmath.cpp.o
[ 18%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_channels.cpp.o
[ 19%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_compare.cpp.o
[ 19%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_copy.cpp.o
[ 20%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_deep.cpp.o
[ 20%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_draw.cpp.o
/Users/peter.kovar/1div0/OpenImageIO/src/libOpenImageIO/imagebufalgo_draw.cpp:743:20: error: unused variable 'default_font_name' [-Werror,-Wunused-variable]
static const char* default_font_name[] = { "DroidSans", "cour", "Courier New",
                   ^
1 error generated.
make[2]: *** [src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_draw.cpp.o] Error 1
make[1]: *** [src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/all] Error 2
make: *** [all] Error 2
nice make  593,42s user 102,51s system 99% cpu 11:38,20 total
peter.kovar@Peters-iMac-Pro x86-64 % uname -a
Darwin Peters-iMac-Pro.local 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:10 PST 2023; root:xnu-10002.61.3~2/RELEASE_X86_64 x86_64

lgritz:
I see. I think this is a problem where the static variable is declared unconditionally, but it's only used if Freetype dependency is found and enabled. It's unrelated to your code. 

I will submit a separate PR that fixes this case. But in the mean time, you can either install freetype or disable that warning.

1div0:
In the meantime I am analyzing https://github.com/libjxl/libjxl/issues/3326 in order to implement streaming encoding and decoding in the OIIO.

lgritz:
Can you clarify what "streaming" means in this context?

src/jxl.imageio/jxlinput.cpp
lgritz: OK, that's fine, but let's remove it from the reported feature list until it's done.

src/jxl.imageio/jxlinput.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+OIIO_PLUGIN_NAMESPACE_BEGIN
+
+#define DBG if (0)
+
+class JxlInput final : public ImageInput {
+public:
+    JxlInput() { init(); }
+    ~JxlInput() override { close(); }
+    const char* format_name(void) const override { return "jpeg xl"; }
```
lgritz, line 36: No space, please. People might use the format_name to construct filenames or other uses where spaces will not be good. Let's call it "jpegxl".

1div0: Understood. I mentioned non breaking space instead. However I am OK with jpegxl name. Will modify ASAP.

lgritz:
I strongly recommend

    git mv src/jxl.imageio src/jpegxl.imageio

and doing other fixups necessary (CMakeLists.txt, the stuff in imgeioplugins.cpp, etc.). We want

* The plugin-in name (which is taken from the directory name)
* The canonical name returned by the format_name() functions

to match exactly. The reason I want you to do this in time for the initial merge is that if we do the "git mv" later, I think it won't transfer the history and attribution properly, it will look like a new file is created and a git log or git blame won't necessarily show your contributions properly (I think).

You also need to fix imageinout_test to exclude this format from the test, for now.

1div0:
I am getting this weird error in the test.

  jpegxl (jxl):
    Writing imageinout_test-jpegxl.jxl ... OK
PtexReader error: read failed (EOF) PtexFile: imageinout_test-jpegxl.jxl
/2TB/usr/src/github.com/1div0/OpenImageIO/src/libOpenImageIO/imageinout_test.cpp:360:
FAILED: in && "Could not create reader"
    Writing Proxy jpegxl ... OK
    Writing bad to bad/bad.jxl ... OK (Could not open file "bad/bad.jxl")

lgritz:
Yeah, that's the failure we've been seeing in CI with the imageinout_test. Like I said, I think it's acceptable for right now to disable this new format from that test, so we can get an initial merge with everything you've done so far, and then come back with all the little fixes including whatever is necessary to restore this test.

In effect, I want to checkpoint what you've got so far, so that every additional change can be viewed individually as a small patch to consider, rather than continuing to pile things onto this one, which means we need to look at your whole patch again every time there's a modification. Let's accept all the parts that seem good so far and iterate.

1div0:
With the debug build of JPEG XL library latest modification revealed error in the writer.

m_filename = imageinout_test-jpegxl.jxl
m_basic_info 64×64×4
lib/jxl/encode.cc:1264: Invalid number of color channels
JxlOutput::write_scanlines(ybegin = 0, yend = 64, ...)
data = 0x3605ac0 nvals = 16384
JxlOutput::close()
JxlOutput::save_image()
data = 0x364d920 size = 65536
lib/jxl/encode.cc:2218: Basic info or color encoding not set yet
status = 1
JxlEncoderAddImageFrame failed with error 129
JxlOutput::close()

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -263,7 +263,8 @@ JxlOutput::save_image()
                                      size);
     DBG std::cout << "status = " << status << "\n";
     if (status != JXL_ENC_SUCCESS) {
-        DBG std::cout << "JxlEncoderAddImageFrame failed.\n";
+        JxlEncoderError err = JxlEncoderGetError(m_encoder.get());
+        DBG std::cout << "JxlEncoderAddImageFrame failed with error " << err << "\n";
```
lgritz, line 267: Don't forget a call to errorfmt() here.

1div0: Thanks. I guess 4 channel image, i.e. RGB+α is not properly handled yet. Will fix ASAP and try w/ Blender.

goodusername123:
Apologies if this isn't exactly the right time or place to ask this, But are there any plans to support depth/z channels and *maybe* even thermal channels? (could possibly be in a later pull request since this one is just "Initial" support?). This would be very useful for projects like Blender to have depth/z maps/data stored/saved in rendered images for example.

libjxl does have support for handling these channel types and there's even a JXL file with such channels here as an example: https://saklistudio.com/jxltests/suz-multichannel-800px.jxl (RGB + Alpha (empty) + Depth + Thermal)


I'm really not too familiar with OpenImageIO as a whole so maybe this would be out of scope? although it seems like the EXR support/code has depth/z channel handling in some form via `z_channel`? unsure, though I do see a few references to `z_channel` in the docs.

lgritz:
OpenImageIO can deal with any number of channels having any names. So yes, totally within scope and nothing special at all.

z_channel and alpha_channel are just a special designations in the ImageSpec to mark it for ease of clients of the library so you don't need to go searching for them every time. But it's in no way implies that those are the only channels you can have.

1div0:
JPEG XL export to RGB with α channel tested by Blender.

src/libOpenImageIO/imageinout_test.cpp
```diff
@@ -368,6 +368,18 @@ test_all_formats()
...
+                size_t n = pixels.size() / sizeof(float);
+                float sum_of_absolute_differences = 0.0;
+
+                for (size_t i = 0; i < n; i++) {
+                    float difference = read_pixels[i] - orig_pixels[i];
+                    sum_of_absolute_differences += fabs(difference);
+                    std::cout << "["<< i << "] " << orig_pixels[i] << " " << read_pixels[i] << " δ " << difference << "\n";
+                }
+                std::cout << "sum of absolute differences " << sum_of_absolute_differences << "\n";
+            }
```
lgritz, line 382: This is the section that's failing the clang-format check.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -0,0 +1,374 @@
...
+class JxlOutput final : public ImageOutput {
+public:
+    JxlOutput() { init(); }
+    ~JxlOutput() override { close(); }
+    const char* format_name(void) const override { return "jpegxl"; }
+    int supports(string_view feature) const override
+    {
+        return (feature == "alpha" || feature == "nchannels"
+                || feature == "exif" || feature == "ioproxy"
+                || feature == "tiles");
```
lgritz, line 31: I think that since this is buffering all output until the end, it can support "random_access" also because it doesn't care what order you send the scanlines.

lgritz:
Now that we pass the imageinout_test, I believe that we are ready for a merge if you fix the clang format issue! Then we can come back to beef up this format with other improvements.

lgritz:
I pushed a few other minor changes. Tests pass now. Mergins.

lgritz:
@1div0 Don't forget we have some important outstanding issues, including


- [ ] Document in builtinformats.rst
- [ ] Handle different pixel types -- right now, always reads/writes float.
- [ ] Compression options
- [ ] EXIF
- [ ] Test cases in the testsuite

ssh4net:
@1div0 I can probably add support for UINT8, UINT16, and FLOAT16 output. But I want to be sure we are not referring to each other.

1div0:
@ssh4net OK, I can add EXIF.

PR #4060, opened 2023-11-23
fix(oiiotool): --autocc bugfix and color config inventory cleanup

lgritz:
The important bug fix is in oiiotool image input when autocc is used, where we reversed the sense of a test and did the autocc-upon-input if the file's color space was equivalent to the scene_linear space (pointlessly), and skipped the conversion if they were different (oh no, big bug!).

Along the way:

* Add missing try/catch around OCIO call that should have had it.

* Some very ugly SPI-specific recognize-by-name color space heuristics. I hate this, sorry, but it solves a really important problem for me.

* A bunch of additional debugging messages during color space inventory, enabled only when debugging, so nobody should see these but me.

* A couple places where we canonicalize the spelling of "oiio:ColorSpace".

Note that there is still an issue with the color space classification using OCIO 2.3+'s identification of built-in equivalents by name. These are shockingly expensive. I will return to this later.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -49,7 +49,7 @@ static const Imath::C3f test_colors[n_test_colors]
 }  // namespace
 
 
-#if 0 /* allow color configuration debugging */
+#if 0 || !defined(NDEBUG) /* allow color configuration debugging */
 static bool colordebug = Strutil::stoi(Sysutil::getenv("OIIO_COLOR_DEBUG"));
 #    define DBG(...)    \
         if (colordebug) \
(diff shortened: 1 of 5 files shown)


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -698,26 +726,36 @@ void
 ColorConfig::Impl::identify_builtin_equivalents()
 {
 #if OCIO_VERSION_HEX >= MAKE_OCIO_VERSION_HEX(2, 3, 0)
+    Timer timer;
```
jessey-git, line 729: `OIIO_MAYBE_UNUSED Timer timer;`

jessey-git (commented):
Just noticed one minor item.

src/libOpenImageIO/color_ocio.cpp
lgritz: Is that needed? I'm not seeing any build errors on CI. I think that the constructor of the timer itself "does something", so it's not unused.

jessey-git: Unsure if it's really needed but another method a little below this used that pattern.

PR #4061, opened 2023-11-24
fix(diff): Fix issue when computing perceptual diff

kohakukun:
## Description

- Fix perceptual test false positive by removing the initialization of `cr` to avoid losing the information when leaving block scope

## Tests

Included additional test with parameter `-fail 1`, which was giving a false positive pass for a perceptual difference. 
This was due to the fact that the result variable `cr` was getting out of scope for the perceptual calculation and the first `if (cr.nfail <= imagesize_t(allowfailures)) {` was testing for the result of the absolute difference instead of the perceptual difference.

Before:
```
Comparing "img1.exr" and "img2.exr"
PASS
```
After:
```
Comparing "img1.exr" and "img2.exr"
64 x 64, 3 channels
  Mean error = 0
  RMS error = 0
  Peak SNR = 0
  Max error  = 1 @ (5, 17, R)  values are 0.1, 0.1, 0.1 vs 0.1, 0.6, 0.1
  0 pixels (0%) over 0.008
  121 pixels (2.95%) over 1
  121 pixels (2.9541%) failed the perceptual test
FAILURE
```

### DIFF
--- a/src/idiff/idiff.cpp
+++ b/src/idiff/idiff.cpp
@@ -301,7 +301,7 @@ main(int argc, char* argv[])
 
             int yee_failures = 0;
             if (perceptual && !img0.deep()) {
-                ImageBufAlgo::CompareResults cr;
+                cr           = {};
                 yee_failures = ImageBufAlgo::compare_Yee(img0, img1, cr);
             }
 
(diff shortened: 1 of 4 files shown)


lgritz:
Hi, @kohakukun. Your fix looks correct to me, thanks so much! There are just two minor CI problems I need you to fix before we can merge:

(1) A very minor clang-format violation (see the "clang-format" CI test, the build output at the end will show you the problem).

(2) It looks like testsuite/diff/ref/out-fmt6.txt is still just a little bit off, as some of the formatting is slightly different for very old fmt versions. If you look in one of those failed CI tests, you should be able to see the problem and correct it.

kohakukun:
Thanks for the quick review 👍

lgritz (approved):
LGTM. Good catch, thanks!

I will also be sure to backport this to the release branch in time for the next patch release circa Dec 1.

PR #4069, opened 2023-12-03
Remove Findfmt.cmake

domin144:
## Description

fmt provides cmake config files since at least version 7.0.0, which is the minimum version supported.
Fixed some variable names, so the provided config can be used.
The previous variable names were missing, if the library was installed in a custom location,
like in case of manual build or provided by conan.

## Checklist:

  nonconforming.


domin144:
Conan still does not like this solution:
```
CMake Error at src/include/CMakeLists.txt:77 (file):
  file COPY cannot find
  "/home/domin144/.conan2/p/b/openi2b837708c2554/b/src/src/include/$<$<CONFIG:Release>:/home/domin144/.conan2/p/b/fmt8d14cec70a06c/p/include>/fmt/core.h":
  No such file or directory.
```

domin144:
Not really proud of the changes, but it should work now.

lgritz:
Sorry for the delay, I've been travelling (still am, but had a little time today to catch up here).

Conceptually, it's fine with me to get rid of Findfmt.cmake since our minimum fmt no longer requires it, and then cleaning up what remains.

Looks like the changes as they stand are failing most of the CI tests, though. 
```
CMake Error at src/cmake/externalpackages.cmake:354 (add_library):
  add_library called with IMPORTED argument but no library type.
```

I'll merge once that gets cleaned up.

domin144:
Another attempt. If this does not work, I will have to get back to this next week.

domin144:
It took some tries, but it finally passes.

lgritz (approved):
LGTM, thanks for the fixes

Issue #4051, opened 2023-11-10
[OpenEXR] Support for luminance chroma images


jreichel-nvidia:
I'd like to request support for luminance chroma images, see https://openexr.com/en/latest/TechnicalIntroduction.html#luminance-chroma-images . An example texture can be found at https://github.com/AcademySoftwareFoundation/openexr-images/blob/main/LuminanceChroma/MtTamNorth.exr .

Right now the check for xSampling/ySampling at the end of OpenEXRInput::PartInfo::query_channels() fails with that example texture. Not sure whether generic sampling support is necessary. Maybe simply using the RGBA interface of OpenEXR in case of Y/RY/BY(/A) images is good enough? (That's what FreeImage is doing.)

dazcjones:
This support would be great as luminance/chroma images can often be smaller in size for similar visual quality.

lgritz:
Does anybody actually use subsampled Luminance/Chroma exr images, aside from testing conformance with the openexr-images examples? Do they exist in the wild?

They have a number of disadvantages -- in addition to being lossy, they only work for scanline images, and you have to read and buffer the whole image at once (to be able to interpolate, you can't process single scanlines at once). The RGBA interface only supports subsamples 1,1 or 2,2 (no other combinations) and doesn't support extra channels (including "layers").

Nuke doesn't read the RY/BY channels correctly, and doesn't support subsampled channels. Apple's Preview.app and Quickview also don't work with it. I've never seen an exr file with lum/chroma and/or subsampling used in real life, and in 15 years of OIIO history, I don't recall anybody complaining about lack of support.

It's definitely possible to add support for this, but before doing so, I want us to think carefully about whether the additional complexity is worth it, if it's not a file type anybody will ever encounter outside of trying to conformance test against the openexr-images samples.

For people who want to take advantage of chroma subsampling, I think the openexr team believes you should be using DWA compression. "I think anyone tempted to try chroma subsampling for efficiency might be better looking at DWA lossy compression of RGB, as that also does YCbCr conversion and is smarter than just blindly throwing away a bunch of chroma data in the hope that nobody will notice," is how Peter Hilman put it.

lgritz:
Just for clarification, when you see in-the-wild images that are Luma/Chroma, are they also subsampled?

PR #4070, opened 2023-12-08
Add support for luminance-chroma OpenEXR images.

jreichel-nvidia:
## Description

A suggestion how to implement support for luminance-chroma OpenEXR images (issue #4051).

Open questions:
- Is there some optimized routine that could be used for the actual pixel copying?
- Should the new code set some attribute such that these converted images can be distinguished by the caller from true RGB(A) images?
- Does this need any documentation (apart from the possible attribute above)?

## Tests

The test images from the OpenEXR testsuite have been added and work with the C++ variant of the test.

## Checklist:

  nonconforming.


src/openexr.imageio/exrinput.cpp
```diff
@@ -1320,6 +1437,12 @@ OpenEXRInput::read_native_tiles(int subimage, int miplevel, int xbegin,
     if (!seek_subimage(subimage, miplevel))
         return false;
     chend = clamp(chend, chbegin + 1, m_spec.nchannels);
+    const PartInfo& part(m_parts[m_subimage]);
+    if (part.luminance_chroma) {
+        errorf(
+            "OpenEXRInput::read_native_deep_scanlines is not supported for luminance-chroma images");
```
lgritz, line 1443: Wrong error message here? should it say read_native_tiles?

lgritz (commented):
Additional comments coming.

lgritz:
I think this is a totally reasonable approach.

One part I don't quite understand is, so far you've only changed exrinput.cpp. What happens if we're using the "core API" and so we're in exrinput_c.cpp? I think we'll want to make the equivalent changes in exrinput_c.cpp so this approaches works regardless of whether we're using the old C++ or the newer core API.


> Is there some optimized routine that could be used for the actual pixel copying?

It depends on how common we think these images are in the performance case. The fact that in 15 years, you're the first to come along and need support for luminance/chroma image makes me think it's not very common, so I think its reasonable at first to just be able to read them, and then we can assess if more needs to be done to make it better performance. You're probably the main user of it, so really it's about whether you think it's worth more work for extra perf, of you're ok for now with whatever perf the RGBA reading gives you.

>    Should the new code set some attribute such that these converted images can be distinguished by the caller from true RGB(A) images?

Probably a good idea? Maybe in that case set a "openexr:luminancechroma" attribute (=1) in the ImageSpec that open returns? Primarily so that `iinfo -v` or equivalent will at least say something that we can use to tell what kind of file it really was.

>    Does this need any documentation (apart from the possible attribute above)?

It would be great if you could touch up the openexr section in builtinplugins.rst to (a) document the attribute above, if you add it; (b) somehow explain, in an appropriate place, that lum/chroma images are supported and will appear as RGBA images; (c) in the "limitations" area, document any remaining limitations, such as if you still don't support subsampled images (if that's true). Apologies that the limitations section did not, previously, properly document that luminance/chroma images weren't supported.

lgritz:
Oh, you're probably right, the exr core vs C++ API is probably not explained anywhere. I will remedy that.

Nomenclature and background notes: OpenEXR has three APIs: (a) a higher level but complete C++ class-based API, and (b) a very restricted, RGBA-only C++ API (which is pretty simplistic but which does some handy conversions such as you're using here), both of which OpenEXR has had forever, and (c) newer OpenEXR versions also have a lower level purely function-based API that happens also to be callable from C. The function-based API has better performance (for now), especially for multithreading, because the old C++ class-based API has a few unfortunate design choices that make certain calls stateful, and therefore not re-entrant. Right now they are entirely separate, but the goal is eventually to re-implement the C++ classes to be implemented with the C functions underneath (and clean up the C++ API to fix the stateful parts), at which point OIIO can switch back to just using the C++ classes and get the better multithreaded performance.

The thing I was concerned about is that the current default is, if OIIO is built against a sufficiently modern OpenEXR, to use the core API when reading, and so if users/builders don't know any better, they won't be using your newly added ability to read these files. The default to use the C++ API can be selected at built time, and also there is a runtime attribute that can be set ("openexr:core", 1), but you have to know to do that. 

There are performance reasons to favor the core API, so that gets you into a chicken and egg problem. We would really need to bend over backwards to be in the core API, realize we're dealing with a subsampled YCbCr file and switch to using the class API for the ReadRGBA classes. This might be more trouble than you think it's worth.

Ok, so in the short term, I think what we have is:

* The vast majority of people who don't expect to encounter YCbCr can stick to the defaults.
* If you suspect you may encounter these files and are not especially concerned about that last bit of multithreaded performance, build or set the option to use the C++ API and then you'll be in the code path that uses this PR's ability to read those chroma files.

In the longer term, hopefully OpenEXR will shift the C++ class API to be implemented on top of the core API (and make the minimal changes needed to allow the better threading performance) and then we'll switch OIIO back to the class API entirely moving forward, and all this mess will go away.

lgritz:
@jreichel-nvidia: Most of the performance gain of the core API is related to multiple threads being able to simultaneously read different tiles or scanlines from the same open file handle. This is crucial for renderers / texture systems that are doing cached, on-demand loading of sections of images. But for applications that read entire exr files in one shot, they aren't going to experience this. So, @jreichel-nvidia, maybe that case is unimportant to you and so it's no big deal for you to use the option to use the C++ API as your default, without experiencing any performance loss in the process.

lgritz:
> Do you have an opinion on restricting the new code paths to subsampled luminance-chroma images (see my question above)?

I don't have an opinion. I think either way, this is a step forward from what we had before, and it's unclear who, if anyone, would need the omitted combinations. I'm happy to let this ride and we can revisit later to enhance it if needed.

src/cmake/testing.cmake
```diff
@@ -282,7 +282,7 @@ macro (oiio_add_all_tests)
         list (APPEND all_openexr_tests openexr-compression)
     endif ()
     # Run all OpenEXR tests without core library
-    oiio_add_tests (${all_openexr_tests}
+    oiio_add_tests (${all_openexr_tests} openexr-luminance-chroma
```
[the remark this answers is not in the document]
lgritz: Makes sense, thanks for the explanation.

src/doc/builtinplugins.rst
```diff
@@ -1446,6 +1451,11 @@ by :file:`libIlmImf`.
   data.  OpenImageIO's OpenEXR writer will silently convert data in formats
   (including the common UINT8 and UINT16 cases) to HALF data for output.
 
+* Subsampled channels are not supported with the exception of reading
+  luminance-chroma images with vertical and horizontal sampling rates of 2.
+  This limited support does work when OpenEXR's C Core API in used, only when
```
lgritz, line 1456: I think you meant to say "does *not* work..."

lgritz (changes requested):
LGTM, only thing I could spot is that one sentence in the docs where I think you left out a "not", and so explain it backwards. Fix that and I'm ready to merge if you are.

lgritz (approved):
LGTM

I will backport this to the release branch. A patch released is scheduled for Feb 1.

PR #4072, opened 2023-12-09
ptex: support for static library

domin144:
## Description
In some environments, only one kind of the library could be available.
Choose proper fallback in such cases. Prefer static library, if
LINKSTATIC is ON and dynamic otherwise.

## Checklist:

  nonconforming.


lgritz (approved):
LGTM

PR #4073, opened 2023-12-10
fix(jpeg2000): include the headers we need to discern version

lgritz:
Without getting the version properly, we weren't disabling use of a deprecated field for OpenJpeg >= 2.5, as we thought we were.

This fixes recently broken CI when building against the current openjpeg master.


ThiagoIze (approved):
CI passed and this is just including headers, so pretty safe assuming these headers exist in all versions of jpeg2000, so I guess we're good.

PR #4075, opened 2023-12-13
admin: Account for duplicate emails in the .mailmap

lgritz:
Please look this over. If you are one of the duplicated names and I haven't chosen your preferred canonical email (I made my best guess), please comment here with corrections.


lgritz:
Nobody has objected or offered corrections --> merging

PR #4076, opened 2023-12-14
fix(simd.h): Make all-architecture matrix44::inverse()

lgritz:
We had an awkward situation of the simd matrix44::inverse() only being defined for Intel SSE, and in all other cases falling back on Imath::M44f::inverse. The problem with that is that it made simd.h depend either directly on ImathMatrix.h, or assume that it would be included prior to inclusion of simd.h. This was not good.

This patch rewrites inverse() to eliminate all of the direct SSE intrinsics in terms of other simd wrapper functions, which means that they will be defined one way or another on all architectures.

Along the way I added a new vfloat4 shuffle that combines two vectors (corresponding to `_mm_shuffle_ps` in SSE land), and a vfloat4 constructor from 2 float*', where the first two elements come from the first ptr and the second two elements come from the second ptr. Ultimately, the final implementation I settled on for the new inverse didn't use these (an earlier version did), but I want to keep them around anyway in case they are useful in the future. I did add a test for them.


lgritz:
Does anybody think this is a bad idea?

ThiagoIze (approved):
The new code is nicer to read. Since this can be an expensive and important function, has this been benchmarked (or the assembly examined) to see if it introduces a slowdown on x86_64 or ARM? Assuming it's not slower, I'm all for merging in this PR.

lgritz:
simd_test benchmarks these, so I know there is no slowdown (though I haven't tested on ARM). It's not especially as much faster than the Imath one as I would have hoped, so to the extent that it's ever performance critical, I'm sure it could use some improvement. But for now, I just wanted to get some implementation so we could remove the Imath header inclusion from simd.h.

ThiagoIze (approved):
Glad to hear at least the x86_64 was benchmarked. I also like that ARM and x86_64 both (mostly) run the same code, which means they'll have roughly the same numerical precision behavior. I think this is safe to merge.

ThiagoIze:
@lgritz do you know why https://github.com/lgritz/OpenImageIO/commit/4cb956b0a0a123376e468a5b5e6e188178ec077e did not remove the `#include <OpenImageIO/Imath.h>` while the 2.6 branch does have it removed? It's a shame 2.5 still depends on Imath.h because of that.

lgritz:
I don't remember now. We could try removing it to see if anything breaks?

PR #4077, opened 2023-12-16
ci: restrict Mac ARM running

lgritz:
I didn't have the condition right -- the `if` isn't respected in the `pull_request` section, so this job was running on every PR instead of just the ones with "macarm" in the branch name. But I think `branches` works with `pull_request`.


lgritz:
Well that didn't work -- the PR didn't trigger the Mac ARM CI workflow. Will need to experiment a bit. Please excuse a few openings and closings of PRs until I get this right.

lgritz:
I think it did work this time. Let's go for this.

PR #4078, opened 2023-12-16
ci: restrict Mac ARM running

lgritz:
This is not the real PR. This is a fake one with a name that doesn't contain 'macarm' to make sure it doesn't start the ARM run.


lgritz:
Experiment worked. Closing this test PR

PR #4086, opened 2023-12-21
build: Add a way to cram in a custom extra library for iv

lgritz:
It's a little involved to explain why, but at work a combination of libraries and some trickery with static linkage led me to really need a way to inject another library into the mix, just for iv linkage.


lgritz:
No objections + harmless change that makes no behavior change by default = merge

PR #4087, opened 2023-12-21
fix(iv): avoid crash with OpenGL + multi-channel images

lgritz:
For image with more than 4 channels, we were misallocating the OpenGL buffer by using the total channels instead of the 4 we wanted to create the texture out of. Also, make sure the calculation usees wide types to avoid possible integer overflow.


lgritz:
No objections + fixes a crash = merge

PR #4088, opened 2023-12-23
perf: Improve perf of IBA::channels in-place operation

lgritz:
For the in-place case (where src and dst refer to the same IB), we were doing an unnecessary full-image copy.


ThiagoIze (approved):
Before we had a copy and the writing done by channels(). Now we just have the writing by channels() and the internal data then gets moved instead of copied, so this looks correct and yes should be faster.

PR #4090, opened 2023-12-23
test: Fix docs test, used wrong namespace

lgritz:
We did `import OpenImageIO as oiio` then instead of using `oiio`, we used `OpenImageIO`.

This was failing sometimes. Frankly, I'm not sure why it ever worked.
Maybe it works on some versions of Python but not others?
I dunno. But this fixes it everywhere and is clearly correct.

### DIFF
--- a/testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
+++ b/testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
@@ -263,10 +263,10 @@ def example_make_texture():
     config["maketx:filtername"] = "lanczos3"
     config["maketx:opaque_detect"] = 1
 
-    ok = ImageBufAlgo.make_texture (OpenImageIO.MakeTxTexture,
+    ok = ImageBufAlgo.make_texture (oiio.MakeTxTexture,
                                     Input, "texture.exr", config)
     if not ok :
-        print("make_texture error:", OpenImageIO.geterror())
+        print("make_texture error:", oiio.geterror())
 
     # END-imagebufalgo-make-texture
 
--- a/testsuite/docs-examples-python/run.py
+++ b/testsuite/docs-examples-python/run.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
+redirect = " >> out.txt 2>&1 "
 
 # Prep:
 command += run_app("cmake -E copy " + test_source_dir + "/../common/grid-small.exr grid.exr")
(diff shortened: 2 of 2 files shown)


testsuite/docs-examples-python/src/docs-examples-imagebufalgo.py
```diff
@@ -263,10 +263,10 @@ def example_make_texture():
     config["maketx:filtername"] = "lanczos3"
     config["maketx:opaque_detect"] = 1
 
-    ok = ImageBufAlgo.make_texture (OpenImageIO.MakeTxTexture,
+    ok = ImageBufAlgo.make_texture (oiio.MakeTxTexture,
```
JeanChristopheMorinPerso, line 266: For consistency with other examples, should this use `MakeTxTexture` instead of `oiio.MakeTxTexture`? All the other examples use stuff imported from `from OpenImageIO import *`.

lgritz: Sure.

I'm not very immersed in Python culture. Remembering that this is not just a test, but its snippets will be directly used in the documentation as examples for the API calls, what's customary when documenting a library and providing code examples? Is it typical to omit any namespacing at all? Or to use a namespace in the examples? To always use the full namespace?

lgritz: Numpy example: https://numpy.org/doc/1.26/user/absolute_beginners.html#how-to-import-numpy

They show how to import, use a short namespace alias, and then they use the namespace everywhere in the rest of the examples. I think this is wise, as for beginners it more easily distinguishes between the Numpy API and things that might be features of Python itself.

I think we should do the same. So I want to keep the use of the oiio namespace here. And let's make a mental note that, separately, we should find anyplace in the examples that does a full `from OpenImageIO import *` and change that to use a namespace to be more clear.

JeanChristopheMorinPerso: If possible, it's recommended to avoid using start imports.

lgritz: Agreed. In general, in both C++ and Python, I think it's better style to always use the namespacing explicitly, for clarity. At least as a default, and take exceptions one by one in cases where removing the namespacing clearly the code more readable.

lgritz: I'm not sure where the `*` import came from. Git blame says me, of course, but it's contrary to my usual style, so I think I just wasn't considering it very carefully at that moment.

lgritz:
Look ok other than that quibble?

PR #4091, opened 2023-12-23
test: Shuffle some tests between directories

lgritz:
Move a bunch of specific tests from testsuite/oiiotool to testsuite/oiiotool-copy to group similar tests together, and also to load balance better between the tests.


lgritz:
No objections + passes CI + doesn't change user-facing code = merge

PR #4093, opened 2023-12-26
feat(oiiotool): oiiotool --cryptomatte-colors

lgritz:
This patch adds a new oiiotool command: `--cryptomatte-colors`, which takes the name of a cryptomatte set of channels, and produces a color-coded matte in which each ID gets a distinct color in the image. This can be useful for visualizing the matte, among other things.

Maybe this is the first cryptomatte related functionality of many?


lgritz:
No objections + need this = merge

Can always revise this feature if people have concerns later. I don't think it can break anything that doesn't try to use the new command.

PR #4095, opened 2023-12-26
OIIO::bitcast adjustments

lgritz:
Use std::bit_cast when available.

Use gcc __builtin_bit_cast when available.


lgritz:
Withdrawing. I'm still mulling this over.

PR #4097, opened 2023-12-30
ci: Repair Sonar scanner analysis

lgritz:
This broke some time back, probably partly related to our change in repository home. This seems to get it mostly working? The reports are still not showing up, but I think the solution might be to make sure this is checked into master, so we need to merge this before assessing whether there is more work to be done in this area.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -43,30 +43,28 @@ jobs:
           - desc: sonar gcc9/C++14 py39 boost1.76 exr3.1 ocio2.2
             nametag: static-analysis-sonar
             os: ubuntu-latest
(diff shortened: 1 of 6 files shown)


lgritz:
I'm going to preemptively merge this without review because (a) it doesn't change any user-facing code, so risk is virtually zero, (b) it takes the static analysis part of CI from being completely broken to seeming to run again (even if it's not yet generating the report web pages). I think pushing it to master may be the piece that's missing to get the reports. But in any case, we'll be revisiting this but I need to get this pushed to build from there.

lgritz:
Addendum: Merging and having an analysis run on master DID make the report show up properly, as I had hoped.

PR #4100, opened 2023-12-31
fix(raw): Avoid buffer overrun for flip direction cases

lgritz:
If you are looping i over range [0,width), and want index to go the other direction, you need `index = width - 1 - i`, not `index = width - i`.

Identified by Sonar static analysis.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -1493,7 +1493,7 @@ RawInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         // For none or 180 degree rotation, the scanlines are still contiguous in memory
         if (sizes.flip == 0 /*no rotation*/ || sizes.flip == 3 /*180 degrees*/) {
             if (sizes.flip == 3) {
-                scanline_start = sizes.raw_width * (m_spec.height - y)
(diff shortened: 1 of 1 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -1505,7 +1505,7 @@ RawInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         // to the array direction so we must copy the pixels into a temporary contiguous buffer
         else if (sizes.flip == 5 /*90 degrees CCW*/
                  || sizes.flip == 6 /*90 degrees CW*/) {
-            scanline_start = m_spec.height - y + sizes.left_margin;
+            scanline_start = m_spec.height - 1 - y + sizes.left_margin;
```
jessey-git, line 1508: Does the `scanline_start` expression on Line 1496 also need the `-1` adjustment?

lgritz: I believe you are correct, I will fix there as well

lgritz:
@Shootfast I would love your opinion on this.

Shootfast:
LGTM!

Shootfast (approved):

PR #4105, opened 2024-01-08
refactor(simd.h): Simplify vbool16 casting

lgritz:
Some of the ways we were using unions for type punning in the vbool16 class were unnecessary, as both union members were basically the same type. Simplify by getting rid of the unnecessary m_bits union member.

This also makes Sonar static analysis happier about these dubious uses of unions for type punning.


lgritz:
Any objections?

lgritz:
No objections -> merging

PR #4106, opened 2024-01-09
refactor(fmath.h): Move bitcast, byteswap, and rotl/rotr to new bit.h

lgritz:
These things were scattered around fmath.h. This grouping corresponds to things in C++20 <bit> (but of course works for older C++ and for Cuda) and lets modules use those without needing the full fmath.h.

Switch our swap_endian implementation to the "byteswap" name and semantics to match C++20. The old swap_endian just calls those. Also add a version of byteswap (in fmath) that takes an OIIO::span, since we generally are leaning toward preferring spans rather than raw pointers and lengths as separate parameters.

None of this should break source or binary compatibility for applications.


lgritz:
How do people feel about this shuffling of things between header files and some minor renaming? (All of which is 100% API & ABI back compatible.)

lgritz:

PR #4108, opened 2024-01-12
fix: Improve over-blurring of certain oiiotool --fit situations

lgritz:
There are two different IBA::fit() entry points -- one that takes a Filter2D* and one that takes a name of the filter. In both cases, if the supplied filter is empty, fit will make its best guess (which, roughly speaking, is lanczos3 when shrinking and blackman-harris when zooming.

You might expect that both, for the "pick my filter case", would end up with the same filters, but they did not! And the "wrong" one was -- ick -- using an awful triangle filter. So when they were unified to both use the same logic, it was easy to see that the one being used by oiiotool was using the triangle and definitely over-blurring! So I think the images look much nicer now that it's using the lanczos3 in appropriate circumstances.

Sometimes lanczos3 can over-sharpen and you see some ringing. That was easy to see at the edges of images that were being letterboxed, so I also put a little extra guard in filtered_sample() so that when it's asked to do edge clamping, it doesn't try to apply the filter too far off the edge of the image boundary.

You *still* may see a little over-ringing in some cases, and if you do, that's a sign that maybe you should be selecting a different, slightly less harsh, filter rather than use its default. Or, does anybody want to argue that we should switch the default resizing filter to be blackman-harris, which is a lot less likely to have unsightly ringing, and if you want something even sharper, you should have to ask for lanczos3?


lgritz:
I realize that this is pretty esoteric, but does anyone have thoughts? Merge and see if anyone's images change? (They do, but I claim it will be better because it was very clearly over-blurring before.)

lgritz:
Going to merge this because nobody has objected and I have another addition to apply on top of it. I'm reasonably sure that this fix is only going to make people's image quality improve, but please let me know if you found a counter-example where the results are unacceptable.

PR #4109, opened 2024-01-12
build(deps): Account for header changes in fmt project trunk

lgritz:
Fixes a CI breakage


lgritz:
Pre-emptively merging because it fixes broken CI for the bleeding edge test.
We can revisit if anybody thinks there's a mistake here.

PR #4110, opened 2024-01-14
fix(dev): Make sure C++ knows ustring & ustringhash are trivially copyable

lgritz:


ThiagoIze (approved):
`ustring` is just doing a straight copy of its data which is just a pointer to a `char*`, and pointers are trivially copyable, so this should be fine. 
The hash is a 64-bit int that is also just being copied as is, so that too is indeed trivially copyable.

src/include/OpenImageIO/ustring.h
```diff
@@ -1154,6 +1150,13 @@ template<> struct hash<OIIO::ustringhash> {
         return static_cast<std::size_t>(u.hash());
     }
 };
+
+
+// Force C++ to understand that ustring and ustringhash are trivially copyable
+// (i.e. memcpy from one ustring to another is fine).
```
jessey-git, line 1156: It would be good to mention the aspects that prevent c++ from just figuring this out. It seems that to make it trivially copyable, we'd have to `= default` the copy assignment operator as well as the destructor here. After that is_trivially_copyable will do the right thing by itself.

Is there any issue with just doing the following:
```
    /// ustring destructor.
    ~ustring() noexcept = default;

    /// Assign a ustring to another ustring.
    constexpr ustring& operator=(const ustring& str) = default;
```

It seems to work locally, and on a reduced repro in c++14 mode on godbolt, but it would mean changing the copy-assignment operator for both types to not return a `const` reference. That's not a typical copy-assignment setup and `=default` doesn't like that return value.  Would that break ABI maybe?

lgritz: Typical copy-assignment returns a mutable reference? Did I just miss this all along?

I don't mind doing this instead.

lgritz: I tink it shouldn't break ABI because these are all inline functions.

lgritz: But in any case, any of these changes that would break ABI will not be backported to the current release branch.

lgritz: I haven't been able to make this work. Making these two default (and changing the return value of = and assign to be non-const, which I guess is the usual way) still doesn't make this trivially copyable, on any of our CI platforms. I'll noodle with it a little more and see if I can figure out why. @jessey-git can you please share your godbolt link so I can try to figure how what I'm doing is different?

jessey-git: This is what I used to test: https://godbolt.org/z/sYn6jdh3G

lgritz: Aha, my mistake.

lgritz:
Take 2:

* Use default implementations of copy ctr, destructor, and assignment
  from ustring. This allows ustring & ustringhash to be recognized as
  being a variety of flavors of trivially constructible, copyable, and
  destructible.

* It seems that the assignment operators were returning `const &`
  whereas it's more typical to return `&`, so now we do.

jessey-git (approved):

PR #4112, opened 2024-01-16
Fixes to reduce problems identified by static analysis

lgritz:
* pngoutput.cpp: Catch possible exceptions.
* printinfo.cpp: assertions to assure we don't dereference a null pointer.
* testtex.cpp: make sure allocated chunk is initialized
* WriterInternal.h: Catch condition that could lead to buffer underflow.
* imagebufalgo_mad.cpp: Simplify pointless clause. The prior test was redundant. Since a few lines above, if B was an image but A was not, we swapped to ensure that A was always an image. So here, we can just test A to know for sure that at least one of them is an image. That makes the test here simpler, but it also makes it more clear to static analysis that from this point forward, A can't be nullptr.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_mad.cpp
+++ b/src/libOpenImageIO/imagebufalgo_mad.cpp
@@ -147,7 +147,7 @@ ImageBufAlgo::mad(ImageBuf& dst, Image_or_Const A_, Image_or_Const B_,
         A_.swap(B_);
     // Get pointers to any image. At least one of A or B must be an image.
     const ImageBuf *A = A_.imgptr(), *B = B_.imgptr(), *C = C_.imgptr();
-    if (!A && !B) {
+    if (!A) {
         dst.errorfmt(
             "ImageBufAlgo::mad(): at least one of the first two arguments must be an image");
         return false;


src/dpx.imageio/libdpx/WriterInternal.h
```diff
@@ -203,7 +203,11 @@ namespace dpx
 	// this routine expects a type of U16
 	template <typename IB, Packing METHOD>
 	void WritePackedMethodAB_10bit(IB *src, IB *dst, const int len, const bool reverse, BufferAccess &access)
-	{	
+    {
+        OIIO_DASSERT(len >= 1);
+        if (len < 1)
+            return;
+
```
jessey-git, line 210: Looks like `libdpx` is formatted with tabs and you're space here.  Is clang-format excluded for libs like this?

jessey-git (approved):
LGTM with 1 curiosity noted above.

PS. Disappointed that sonar is detecting simd's `shuffle` as an unsafe pRNG lol :)

src/dpx.imageio/libdpx/WriterInternal.h
lgritz: Yes, this is "imported" code that all along I excluded from clang-format so that it remains possible to diff it against the original and incorporate any fixes, should that original library ever be updated. (Though to the best of my knowledge, it's been unmaintained for a long time, so maybe we should give up the pretense?)

But yeah, in vscode it looks like it matches the surrounding text, but I think it looks odd in the github review diff because of space vs tabs differences.

lgritz:
Yeah, Jesse, Sonar is kind of a mixed bag of false positives (sometimes bordering on the inexplicably absurd, such as mistaking the SIMD shuffle<> template for a substandard PRNG, presumably because of the name alone?) but also some legit latent bugs that would be very hard to find any other way. So for now, I'm tolerating the silly false positives (marking them as such where possible) and fixing the real bugs as it identifies them.

lgritz:
I really botched this PR merge, will have to open a new one and merge it, sorry for the confusion.

lgritz:
reopening

PR #4113, opened 2024-01-17
Fixes to reduce problems identified by static analysis

lgritz:
* pngoutput.cpp: Catch possible exceptions.
* printinfo.cpp: assertions to assure we don't dereference a null pointer.
* testtex.cpp: make sure allocated chunk is initialized
* WriterInternal.h: Catch condition that could lead to buffer underflow.
* imagebufalgo_mad.cpp: Simplify pointless clause. The prior test was redundant. Since a few lines above, if B was an image but A was not, we swapped to ensure that A was always an image. So here, we can just test A to know for sure that at least one of them is an image. That makes the test here simpler, but it also makes it more clear to static analysis that from this point forward, A can't be nullptr.


lgritz:
This is a reposting of botched PR #4112. Please see there for the review discussion and approval.

PR #4114, opened 2024-01-18
build(deps): deal with changes in fmt's trunk

lgritz:
Bleeding edge CI test failures due to more changes in fmt's trunk. Solutions:

* Accommodate some recent header refactoring in fmt.

* Straighten up some const issues that worked before, but break under the latest fmt. Making the custom formatter's 'format' method be const fixes the latest, but breaks some of the oldest versions of fmt we support, so we add the const only from fmt 9.0 on.


lgritz:
Pre-emptively merging without review since this fixes broken CI.

PR #4115, opened 2024-01-18
ci: improve parallel builds by basing on number of cores

lgritz:
With GitHub runners with more cores becoming available, adjust our CI scripts to do parallel builds with nprocs+2 instead of hard-coding. We use slightly more than the number of available cores because these builds and tests tend to be I/O bound, so they can be a little over-threaded to get maximum utilization.


lgritz:
No objections for a week + only affects CI == merge

PR #4118, opened 2024-01-19
iv: Assume iv display gamma 2.2

### LINKED ISSUE #318: openimageio-tools: handling of gamma for JPEG files in "iv" is weird
A Debian user filed the following bug report:

http://bugs.debian.org/668476

It's not related to packaging, so fowarding it here.
Thanks for your efforts in fixing this.

Cheers.

lgritz:
A better default for users is probably to assume a 2.2 value if the GAMMA env variable is not set. If you want something different, set GAMMA proactively or use a real color managed workflow.

Will this break anybody in important ways?

Fixes #318

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -118,7 +118,7 @@ ImageViewer::ImageViewer(bool use_ocio, const std::string& image_color_space,
 {
     readSettings(false);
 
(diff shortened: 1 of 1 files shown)


lgritz:
If there are no objections to this by tomorrow EOD, I will merge it.

lgritz:
No objections for a week -> merging

PR #4119, opened 2024-01-19
Fix AVX-512 round function

AngryLoki:
## Description
This PR fixes vfloat16 round function. Intrinsic `_mm512_roundscale_ps` was
used incorrectly, and caused failure on Zen4 CPU.

```
/var/tmp/portage/media-libs/openimageio-2.5.5.0-r1/work/OpenImageIO-2.5.5.0/src/libutil/simd_test.cpp:1579:
FAILED: round(F) == mkvec<VEC>(std::round(F[0]), std::round(F[1]), std::round(F[2]), std::round(F[3]))
	values were '-1.5 0 1.5 4 -1.5 0 1.5 4 -1.5 0 1.5 4 -1.5 0 1.5 4' and '-2 0 2 4 -2 0 2 4 -2 0 2 4 -2 0 2 4'
``` 

In old code `_mm512_roundscale_ps (a, (1<<4) | 3)` meant the following:
```
[0001] - Number of fixed points to preserve
[0] - Use MSCSR exception mask
[0] - Select mode from imm
[11] - Truncate mode
```
Effectively enabling rounding to nearest 0.5, not to integer.

References:
* https://www.felixcloutier.com/x86/vrndscalepd#fig-5-29
* https://stackoverflow.com/questions/50854991/instrinsic-mm512-round-ps-is-missing-for-avx512

## Tests
* This fixes test_simd

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -10180,7 +10180,7 @@ OIIO_FORCEINLINE vfloat16 floor (const vfloat16& a)
 OIIO_FORCEINLINE vfloat16 round (const vfloat16& a)
 {
 #if OIIO_SIMD_AVX >= 512
-    return _mm512_roundscale_ps (a, (1<<4) | 3); // scale=1, round to nearest smaller mag int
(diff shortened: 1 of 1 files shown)


lgritz (approved):
This looks right to me, thanks for the fix!

PR #4120, opened 2024-01-20
build: correctly disable OpenVDB when it's incompatible

lgritz:
OpenVDB >= 10.1 is incompatible with building OIIO using C++14. We had tried to disable this disallowed combination, but a typo prevented it from working properly.


lgritz:
no objections -> merging

PR #4121, opened 2024-01-21
cleanup: Coalesce redundant STRINGIZE macros -> OIIO_STRINGIZE

lgritz:
We did this separately each place we needed it. Do it just once, in a public header, and give it a name that won't clash with other packages.


lgritz:
no objections -> merge

PR #4124, opened 2024-01-21
feat: query/print build information

lgritz:
New OIIO global attributes (read only):

  - "build:platform" retrieves the OS and CPU type that this build of OIIO is fof (e.g., "Linux/x86_64").
  - "build:compiler" retrieves the compiler name and version used to build OIIO itself (e.g. "gcc 9.3").
  - "build:dependencies" retrieves a semicolon-separated list of library dependencies -- both for image format support (like libtiff) but also general (boost, TBB, Python, fmt, OCIO).
  - "build:simd" is a new (preferred) synonym for the SIMD and other hardware capabilities selected at build time. The old name, "oiio:simd" is hereby softly deprecated as confusing.

`oiiotool --buildinfo` is a new command that prints this information.

Example output:

```
$ oiiotool --buildinfo

OIIO 2.6.0.2spi | MacOS/x86_64
Dependencies: OpenEXR 3.2.1, LIBTIFF Version 4.6.0, jpeg-turbo
    3.0.1/jp80, dcmtk 3.6.8, FFMpeg 6.0 (Lavf60.16.100), gif_lib 5.2.1,
    libheif 1.17.6, OpenJpeg 2.5.0, null 1.0, OpenVDB 11.0.0abi11, libpng
    1.6.40, Ptex 2.4, libraw 0.21.2-Release, Webp 1.3.2, Boost 1.83,
    OpenColorIO 2.3.1, NO Python!, TBB 2021.11.0, fmt 10.2.1
```

This is all a convenience utility to make it easy to ask users to run a simple command that will tell us a lot of information relevant to resolving their issues in cases where it's related to some of their build-time compiler or dependencies.

Also cleaned up the "bug report issue template" to clarify the instructions and specifically ask people to run this command and include the information with any bug report.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,29 +8,40 @@ assignees: ''
 ---
 
 **Describe the bug**
+
 A clear and concise description of what the bug is. What happened, and
 what did you expect to happen instead.
 
+
+**OpenImageIO version and dependencies**
+
+Please run `oiiotool --buildinfo` and paste the output here.
+
+Also please tell us if there was anything unusual about your environment or
+nonstandard build options you used.
+
+
 **To Reproduce**
+
 Steps to reproduce the behavior:
 1. Do this...
 2. Then this...
 3. Then THIS happens (reproduce the exact error message if you can)
+4. Whereas I expected this other specific thing to happen instead.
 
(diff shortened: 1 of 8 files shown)


lgritz:
Does this seem good to everybody? Is there additional information I've forgotten that you think we should be sure to print that will help us to diagnose problems reported by users?

PR #4125, opened 2024-01-23
span and range checking enhancements

lgritz:
* New header span_util.h with span-related helpers: 
  - make_span/make_cspan helps for converting vectors & arrays to spans when passing to templated functions. 
  - spancpy, spanset, spanzero are memory-safe, range-checking substitutions for memcpy, memset when working with spans.
* Add range check DASSERT to span access in debug mode. There were a couple spots where we were taking the address of an out-of-range value that needed to be adjusted to use a better idiom.
* ParamValue: extra OIIO_DASSERT range check for get() method.
* Simplify constructors by replacing `&x[0]` with using `data()`, which is always well-defined.
* Fix typos in span.h comments

The range checking imposes no extra cost for release/optimized builds. But it might help us for CI analysis and any time we're debugging.

### DIFF
--- a/src/include/OpenImageIO/paramlist.h
+++ b/src/include/OpenImageIO/paramlist.h
@@ -226,6 +226,8 @@ public:
     // a std::string.
     template<typename T> const T& get(int i = 0) const noexcept
     {
+        OIIO_DASSERT(i >= 0 && i < int(m_nvalues * m_type.basevalues())
+                     && "OIIO::ParamValue::get() range check");
         return (reinterpret_cast<const T*>(data()))[i];
     }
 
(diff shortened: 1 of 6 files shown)


lgritz:
Does anybody have objections to this, or comments about it?

There are primarily two sets of changes in this PR:

1. Adding debug-mode-only range checking to the indexing operators for our `span<>` classes. Because it's debug only, there should be no release-build performance penalty for this, but it should make it easier to identify addressing range errors when debugging and when running the testsuite in CI, especially on the sanitizer and analysis runs.

2. Adding some new utility functions that do buffer copies with spans. We don't use these (yet), and are in a new header, so they can't possibly break anything that currently works. But I'd like to start using them where applicable, because it should give some additional protection for address range bugs.

lgritz:
Discussed in TSC, people like spans.

PR #4127, opened 2024-01-26
fix(oiiotool): Overhaul and fix bugs in mixed-channel propogation

lgritz:
Remember that ImageBuf only stores one pixel data type per buffer, but of course they may have differing-per-channel types in image files. So how does oiiotool know what to write? There is a series of heuristics: -d is for explicit user control, if not, then try to deduce it from the input files.

A bedrock precept is that `oiiotool in.exr -o out.exr` ought to write the same channel types that it read in.

The logic to implement this with the various image operations and possible overrides can be hard to get right. Indeed, I did not get it right. It made mistakes even in simple cases like the above. And the code was complicated and hard to understand, as well.

This PR rewrites a couple functions that implement this chunk of this logic, both correcting several errors as well as (I hope) being clearer and better commented so that we can more easily reason about it and modify it in the future.

I also added some additional tests to testsuite/perchannel to verify some of the cases that were previously failing.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -443,7 +443,7 @@ Oiiotool::remember_input_channelformats(ImageRecRef img)
                 if (input_channelformats[subchname] == "")
                     input_channelformats[subchname] = chtypename;
             } else {
-                if (input_channelformats[chname] != "")
+                if (input_channelformats[chname] == "")
                     input_channelformats[chname] = chtypename;
             }
         }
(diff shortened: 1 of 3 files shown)


lgritz:
I would appreciate some eyeballs on this PR review. This fully passes the testsuite, and it passes new tests we added where the old behavior was clearly broken and the new behavior is clearly correct. But it feels much riskier than most PRs because it really affects the heart of how oiiotool's `-d` command works and thus a mistake here could break all sorts of workflows.

lgritz:
Last chance: this has lingered for a while, does anybody have concerns about it?

sh-coursework:
I was the one who originally reported the issue, and I've tested 2.5.8 with the patch, and I believe it's working.
It now preserves the channel data types, and the file doesn't grow nearly as much as before.
With a small deep file, it used to grow about 6% when I reported the bug, and now is 0.8%. We're guessing this is just some additional metadata. That 0.8% was roughly 900 bytes.

(I haven't actually read the code, but running it from the outside, it looks like it works.)

PR #4128, opened 2024-01-27
Internals: various fixes for memory safey and reduce static analysis complaints

lgritz:
* jpeg output: Extra memory check
* TS: batch texture call fixed to handle any number of channels
* TS: assert to guard against possible null pointer dereference
* oiiotool printinfo: clarify assertion for the sake of static analysis
* jpeg2000 reader: assertion to ensure we don't dereference a null pointer

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -126,7 +126,7 @@ jobs:
             echo "BUILD_OUTPUT_DIR is " "${{ env.BUILD_WRAPPER_OUT_DIR }}"
             find . -name "*.gcov" -print
             # sonar-scanner --define sonar.cfamily.build-wrapper-output="${{ env.BUILD_WRAPPER_OUT_DIR }}"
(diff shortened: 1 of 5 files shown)


lgritz:
Some explanation:

I think the place in texturesys.cpp where we switch from allocating 4 components of those variables to using OIIO_ALLOCA ws a "real bug", though one that nobody will likely ever have encounterd since you'd need to make a texture call asking for >4 texture channels to be returned. (OSL, which is how most people end up with the TS in their lives, won't ever ask for more than 4 texture channels.)

The other changes here are, as far as I know, "not real bugs". The added assertions merely (a) confirm the guarantees we believe the code already makes, though it's nice to have the extra verification, and (b) will cause the static analysis to stop marking bugs there. What I mean by that is:

    int func(Foo* bar) {
        return bar->baz;
    }

might flag in static analysis that we could be dereferencing a null pointer, because it can't tell what pointer might be passed to func, which just blithely dereferences it. But if we change it to

    int func(Foo* bar) {
        OIIO_DASSERT(bar != nullptr);
        return bar->baz;
    }

then there is no more static analysis error because it sees (in the macro expansion of OIIO_DASSERT) that if bar is nullptr, it will take a different code path to error and abort, and so there are no code paths that get to bar->baz that allow bar to be null. We run the static analysis on a debug build, so those assertions are fully active. In release/optimized builds, the OIIO_DASSERT goes away, so there is no runtime penalty for the extra checks in production code, but there is in CI, analysis, and debug builds.

lgritz:
No objections -> merge

PR #4129, opened 2024-01-27
ci: Update all github actions to their latest versions that's compatible

lgritz:


lgritz:
CI only + works + no risk to users -> merge

PR #4130, opened 2024-01-28
ci: Bump 'latest' test to newer dep versions, document

lgritz:


lgritz:
CI only + works + no risk to users -> merge

PR #4132, opened 2024-01-30
fix: certain int->float type conversions in TypeDesc/ParamValueList

lgritz:
We had all along omitted an important 'int' conversion case for an internal to_floats() helper function that's used by the TypeDesc::convert_type utility. Which is in turn used by ParamValue::get_float() and ParamValueList::get_float().

All of these would fail to return the proper float if the underlying data was an int, despite these functions saying that they would have no trouble returning the float equivalent of int data when asked.

I added a test that verifies that it's wrking.


ThiagoIze (approved):
This matches the UINT logic above and comes with tests, so looks good to me.

PR #4134, opened 2024-01-31
ci: Start using macos-14 ARM runners, bump latest OCIO

lgritz:
No more need for the paid macos13-arm-py311 runner and only use it on the ASWF repo selectively. An ARM-based MacOS-14 runner is now available for open source projects, so can run free on every PR and on user forks.


lgritz:
CI only, can't break anything -> merge

PR #4140, opened 2024-02-03
testing/api: Add test for filter values and 'filter_list' query

lgritz:
Add testsuite/filters that verifies that the filter shapes are correct and don't change if we make future changes to the filter code.

Along the way, add a new global getattribute query 'filter_list', that returns a semicolon-separated list of all the 2D filters it knows about.


lgritz:
No comments or objections for a week + only affects tests and can't break existing code = merge

PR #4142, opened 2024-02-06
fix(heic): Don't auto-transform camera-rotated images

### LINKED ISSUE #4123: provide Orientation metadatum in HEIF files
oiio extracts some metadata from HEIF files, but the camera orientation does not seem to be among them. I think the orientation value may be hidden 'somewhere deeper' - exiftool does extract it, together with huge amounts of information - much more than the set of values iinfo produces - like this:

```
  ...
  | Item 51) 'Exif' (2384 bytes)
  | + [TIFF directory]
  | | ExifByteOrder = MM
  | | + [IFD0 directory with 13 entries]
  | | | 0)  Make = Apple
  | | | 1)  Model = iPhone 12
  | | | 2)  Orientation = 6
  | | | 3)  XResolution = 72 (72/1)
  ...
```

I tried this with a test image in portrait orientation (exif code 6) from and iphone 12 and iterated over the extra_attribs (this was done in C++ with a v.2.4.7.1 oiio install), and I also tried a build of iinfo from oiio master (oiio 2.6.0.2dev with libheif 1.15.1-1) which did not yield an Orientation value either.

Since this is such a crucial bit of information - especially for a format used on smartphones, which are often used in portrait orientation, gaining access via oiio would be greatly appreciated!

lgritz:
It seems that unlike most image formats where we merely note the camera rotation but don't rotate the pixels, the heic reader automatically rotated.

The trick is that using the C++ API wrapper of libheif, there is no way to pass the decoder options struct that is where you need to say not to auto-transform.  Needed to drop down to the lower level C API for this one spot.

We want this format to behave like the others -- the orientation is advisory, and it's up to the app how to deal with it. But we have back compatibility to deal with.

If a rotated image is aecountered, OIIO 2.5 and earlier will by default still auto-rotate to preserve compatibility, but OIIO 2.6+ will not auto-rotate.

When not rotating, the Orientation metadata will reflect the desired display orientation of the image, according to what was in the file.

When auto-rotating, Orientation will be 1 (canonical display orientation), and a new attribute "heif:Orientation" will reflect what was originally in the file.

When opening a file for input, the special configuration metadata hint "heif:reorient" can express a preference for auto-rotation, overriding the default. Thus, setting this hint to 0 for OIIO 2.5 will turn off auto-rotation, and setting it to 1 for OIIO 2.6+ will make it auto-rotate like it used to do.

I also added some additional tests for heif files.

Along the way I fixed a rather embarrassing documentation problem as well, where all the orientation codes were "off by 1", oh boy.

Fixes #4123

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -260,8 +260,10 @@ macro (oiio_add_all_tests)
                     IMAGEDIR openexr-images
                     URL http://github.com/AcademySoftwareFoundation/openexr-images)
     oiio_add_tests (heif
-                    FOUNDVAR Libheif_FOUND ENABLEVAR ENABLE_Libheif
-                    URL https://github.com/nokiatech/heif/tree/gh-pages/content)
+                    FOUNDVAR Libheif_FOUND
+                    ENABLEVAR ENABLE_Libheif
+                    IMAGEDIR oiio-images/heif
+                    URL http://github.com/AcademySoftwareFoundation/openexr-images)
     oiio_add_tests (ico
                     ENABLEVAR ENABLE_ICO
                     IMAGEDIR oiio-images URL "Recent checkout of oiio-images")
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -685,6 +685,11 @@ preferred except when legacy file access is required.
    * - ``oiio:Gamma``
      - float
      - the gamma correction specified in the RGBE header (if it's gamma corrected).
+   * - ``heif:Orientation``
+     - int
+     - If the configuration option ``heif:reorient`` is nonzero and
+       reorientation was performed, this will be set to the original
+       orientation in the file.
 
 
 **Configuration settings for HDR input**
@@ -704,6 +709,14 @@ options are supported:
      - ptr
      - Pointer to a ``Filesystem::IOProxy`` that will handle the I/O, for
        example by reading from memory rather than the file system.
+   * - ``oiio:reorient``
+     - int
+     - The default of 1 means to let libheif auto-reorient the image to
+       undo the camera's orientation (this will set a "heif:Orientation"
+       metadata to the Exif orientation code indicating the original
+       orientation of the image). If this hint is set to 0, the pixels will be
+       left in their orientation as stored in the file, and the "Orientation"
+       metadata will reflect that.
 
 **Configuration settings for HDR output**
 
@@ -750,6 +763,15 @@ currently supported for reading, but not yet writing. All pixel data is
 uint8, though we hope to add support for HDR (more than 8 bits) in the
 future.
 
+The default behavior of the HEIF reader is to reorient the image to the
+orientation indicated by the camera, and to report the "Orientation" metadata
+as 1 (indicating that the image should be displayed as returned) and set the
+"oiio:OriginalOrientation" metadata to what was originally stored in the file.
+If you want to read the image without automatic reorientation, you can set the
+configuration option "oiio:reorient" to 0, in which case the pixels will be
+left in their orientation as stored in the file, and the "Orientation"
+metadata will reflect that.
+
 **Configuration settings for HEIF input**
 
 When opening an HEIF ImageInput with a *configuration* (see
@@ -769,6 +791,12 @@ attributes are supported:
        cause the reader to leave alpha unassociated (versus the default of
        premultiplying color channels by alpha if the alpha channel is
        unassociated).
+   * - ``oiio:reorient``
+     - int
+     - If nonzero, asks libheif to reorient any images (and report them as
+       having Orientation 1). If zero, then libheif will not reorient the
+       image and the Orientation metadata will be set to reflect the camera
+       orientation.
 
 **Configuration settings for HEIF output**
 
--- a/src/doc/imageinput.rst
+++ b/src/doc/imageinput.rst
@@ -923,6 +923,10 @@ hints are supported by each reader) are:
        cause the reader to leave alpha unassociated (versus the default of
        premultiplying color channels by alpha if the alpha channel is
        unassociated).
+   * - ``oiio:reorient``
+     - int
+     - If zero, disables any automatic reorientation that the reader may
+       ordinarily do to present te pixels in the preferred display orientation.
 
 Examples:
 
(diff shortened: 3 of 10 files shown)


kfjahnke:
Loking at your PR, I think some of the code looks dodgy (343ff.):
```
            switch (rot) {
            case 270: orientation = cw[orientation];
            case 180: orientation = cw[orientation];
            case 90: orientation = cw[orientation];
            default: break;
```
A case switch is okay, but you code it with fallthrough, so e.g. if control flow arrives the first case label, it will re-set orientation thrice in a rather intransparent manner. Are you sure that is intended?
Further on,
```
            if (mirror == heif_transform_mirror_direction_vertical) {
                orientation = mirrorv[orientation];
            } else if (mirror = heif_transform_mirror_direction_horizontal) {
                orientation = mirrorh[orientation];
```
This looks like it might work, but I think it's intransparent - the orientation value may change with every execution of the loop body, and I find it hard to figure out what the final outcome might be.
I think it would be better first to extract the properties into variables (int is_rotated, bool is_flipped) - the image can be rotated in four ways, and possibly flipped, so there are eight cases to deal with, which could be handled e.g. in a nested conditional statement in a second step after the properties have all been seen (after the loop over xpcount is done). This would yield 'orientation' and you could fix it so that the value is EXIF-compatible, and the code could put a comment alongside each case, making it all more readable.

lgritz:
> A case switch is okay, but you code it with fallthrough, so e.g. if control flow arrives the first case label, it will re-set orientation thrice in a rather intransparent manner. Are you sure that is intended?

No, it's correct. It isn't resetting it each time, it is looking up an additional 90 degree rotation in the `cw` array.

But I agree that it's unclear, and I will fix it with an alternate formulation. This one may be a shade too "clever."

lgritz:
Here is a more clear way to express it, maybe you can see that the case statement as I wrote it is equivalent:

            // cw[] maps to one additional clockwise 90 degree turn
            static const int cw[] = { 0, 6, 7, 8, 5, 2, 3, 4, 1 };
            for (int i = 0; i < rot / 90; ++i)
                orientation = cw[orientation];

kfjahnke:
Okay, I see it now. Still I'm not happy with it, and I think it's hard to comprehend. To make it palatable, it would require a lengthy comment, so instead you could just code it without the artistics. Just my fifty cents. Of course I appreciate a good bit of 'juggling', but I've come off that kind of stuff and write KISS code now ;-)

lgritz:
Yeah, I'm changing it to the loop, which I think is more clear than the switch.

lgritz:
Pushed an update based on the discussion in #4123.  New description of what this PR really does:

---

fix(heic): Make auto-transform of camera-rotated images optional

Policy change about ImageInput automatic transformation of images with
camera orientation data, as follows:

* Some ImageInput format readers may have the capability to
  automatically reorient images to the preferred display orientation
  upon input, based on the camera Orientation metadata. Whether a
  reader does so will tend to depend on the capabilities of the
  underlying format library. If it can support reorientation easily,
  it is expected to do the reorientation automatically by default.

* If the pixel data are re-oriented, the "Orientation" metadata should
  be set to 1 to indicate that the image returned to the caller is in
  the preferred display orientation, and "oiio:OriginalOrientation"
  metadata should be set to indicate the original camera orientation
  (prior to the automatic re-orientation).

* Open-with-configuration hint "oiio:reorient" (default: 1), if set to
  0, is a request to NOT do the reorientation. In that case,
  "Orientation" metadata should be set correctly to the actual
  orientation.

The heif reader was previously always re-orientating images as the
default behavior of the underlying libheif. That remains the default,
but this patch makes the following changes to libheif behavior to
conform to the policy above: (a) changes to code to allow it to
disable the reorienting behavior if hint "oiio:reorient" is 0; (b)
correct setting of Orientation and oiio:OriginalOrientation metadata;
(c) make this more clean in the documentation.

Tech details: The trick is that using the C++ API wrapper of libheif,
there is no way to pass the decoder options struct that is where you
need to say not to auto-transform.  Needed to drop down to the lower
level C API for this one spot.

Fixes #4123

A subsequent PR will revisit the RAW reader to make it also conform to
the new policy.

kfjahnke:
Excellent! Just what I had hoped for.

One thing I am missing is concerning the uniform use of the EXIF standard's rotation/flip values. Prescribing that OIIO should present the Orientation property using this standard would make it clear that other ways of expressing rotation or flip remain a detail of the plugin - e.g. libRAW's use of other numerical codes. In a way you silently accept this standard at least partly by proposing the value '1' as the 'correctly oriented for display' metadatum, which is arbitrary, but conforms to the EXIF standard.

The standard is nicely documented in various places, e.g. [here](https://sirv.com/help/articles/rotate-photos-to-be-upright/) and there are test images to be had, e.g. [here](https://github.com/recurser/exif-orientation-examples), where one can find a complete set of JPEGs to test whether one's processing pipeline does the right thing to display the images correctly - I used this set to test my latest addition to lux, the handling of 'flipped' images, which I had ignored previously.

You might want to state that if the user wants to affect pre-rotation using a plugin's capabilities (e.g. by passing raw:user_flip) they would have to adhere to that plugin's semantics and could not expect to rely on orientation values to be automagically translated.

lgritz:
I consider any deviation from the Exif (and, long before that, TIFF) encoding conventions of orientation to be a bug in the raw reader. The OIIO docs clearly say what the codes of the "Orientation" metadata matches TIFF/Exif conventions -- or at least, it is supposed to, though there was a very unfortunate typo that was fixed as part of this PR.

lgritz:
Is this, as it currently stands, adequate for heif? Should we merge what we have so far?

kfjahnke:
I did a test build from your fork's feature branch. Switching off the autorotation works and returns the data in 'native' orientation, but then I don't seem to always get the Orientation tag which I'd need to perform the rotation myself. Is this deliberate or am I maybe doing something wrong? I have the feeling that when the orientation is set in the EXIF metadata (as it is in my 'fudged' test image) it comes through, but otherwise it doesn't (e.g with the portrait HEIF test image we both discovered).

kfjahnke:
Oops.. I think it's my mistake, the oiio:reorient did not come through in time

lgritz:
Not all files contain the Orientation metadata. When it's not present, you should assume the default orientation. In OIIO speak, I'd recommend

    int orientation = spec.get_int_attribute("Orientation", 1);

and then you'll get the 1 if the metadata is not present.

kfjahnke:
Point taken, thanks for the hint!

lgritz:
I think we're good on this. Merging. We can always revise more later.

PR #4071, opened 2023-12-08
SIMD: faster vint4 load/store with unsigned char conversion

aras-p:
## Description

vint4::load from unsigned char pointer got pre-SSE4 code path. Testing on Ryzen 5950X / VS2022 (with only SSE2 enabled in the build):
- vint4 load from unsigned char[]: 946.1 -> 4232.8 Mvals/sec

vint4::store to unsigned char pointer got simpler/faster SSE code path, and a NEON code path. Additionally, it got test correctness coverage, including what happens to values outside of unsigned char range (current behavior just masks lowest byte, i.e. does not clamp the integer lanes).

- vint4 store to unsigned char[]: 3489.8 -> 3979.3 Mvals/sec
- vint8 store to unsigned char[]: 5516.9 -> 7325.3 Mvals/sec

NEON code path as tested on Mac M1 Max (clang 15):
- vint4 store to unsigned char[]: 4137.2 -> 6074.8 Mvals/sec

## Tests

vint4 store to unsigned char pointer got actual correctness checking test, which seemingly was lacking before (only had a benchmark test).

## Checklist:

- [x] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/CONTRIBUTING.md).
- [ ] I have updated the documentation, if applicable.
- [x] I have ensured that the change is tested somewhere in the testsuite
  (adding new test cases if necessary).
- [ ] If I added or modified a C++ API call, I have also amended the
  corresponding Python bindings (and if altering ImageBufAlgo functions, also
  exposed the new functionality as oiiotool options).
- [x] My code follows the prevailing code style of this project. If I haven't
  already run clang-format before submitting, I definitely will look at the CI
  test that runs clang-format and fix anything that it highlights as being
  nonconforming.

PR #4143, opened 2024-02-07
fix(simd.h): Address NEON issues

### LINKED ISSUE #4111: [BUG] OIIO 2.5.7.0 fails to build on Fedora only on aarch64
**Describe the bug**
```
/builddir/build/BUILD/OpenImageIO-2.5.7.0/src/include/OpenImageIO/simd.h:4796:32: error: cannot convert ‘int16x8_t’ to ‘OpenImageIO_v2_5::simd::vint4::simd_t’ {aka ‘int32x4_t’} in initialization
 4796 |     simd_t val16 = vcombine_s16(vqmovn_s32(clamped), vdup_n_s16(0));
      |                    ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                |
      |                                int16x8_t
```
Full log:
https://kojipkgs.fedoraproject.org//work/tasks/1777/111791777/build.log

Perhaps this is do to the new SIMD updates? I never built 2.5.6.0 in the official repos but all test builds completed.

**Platform information:**
 - OIIO branch/version: 2.5.7.0
 - OS: Fedora Rawhide
 - C++ compiler: gcc 13.2.1
 - Any non-default build flags when you build OIIO: See full logs

lgritz:
Primarily, recent changes (PR #4071) to vint4::store for the NEON case appear to have some type mismatches, which apple clang on ARM-based Mac (including our CI) seems ok with, but which is generating type errors on other ARM Linux platforms.

I think the types were weird here, so I tightened it up to get the types right for temporary variables in that function. That's the primary fix here.

Secondarily, I modified simd.h and the CMake setup so that build option USE_SIMD=0 will disable NEON in the same way that it disables SSE. (I realized that USE_SIMD=0 was not disabling NEON, so there was no way for a NEON platform to completely disable SIMD if they needed to.)

Fixes #4111


lgritz:
Fixes #4111

lgritz:
@aras-p Could you take a look and see what you think? This is addressing code you recently added. I think that functionally your code was doing the right operations (and I believe it is tested by simd_test in the testsuite), but it seems that some compilers accepted your odd typing, whereas other compilers seem distressed by the implied bitcasts and wanted the right neon intrinsics to make that explicit.

aras-p:
Right, sorry about that! Clang is indeed more relaxed wrt NEON intrinsics compared to Gcc or Msvc. I should have checked on those. (but also, I guess, would be useful to have those compilers for Arm in OIIO CI?)

PR #4144, opened 2024-02-08
safety: refactor iffoutput.cpp for memory safety

lgritz:
Static analysis has been yelling about code in iffoutput.cpp possibly overrunning buffers, and frankly the code in this module is so confusing that I can't tell if it's correct or not. (It's 13 years old, hasn't been touched for a long time.)

In an ideal world, this would all be rewritten to use spans so the buffer lengths are known and checked. But I don't really have the time or inclination to rewrite it -- after all, iff is not a very important file format, though I do think it's used enough that we can't drop it entirely.  The payoff from a full rewrite is marginal. So I came up with the following compromise, embodied by this PR:

I'm making spans of the regions we're ultimately reading from and writing to, passing those down the chain of function calls, and so even though the actual operations are total pointer arithmetic spaghetti, we can use those spans to verify that the pointers are still within the span bounds any time we read or write through them. We do it with assertions that are only active for debug builds, so there shouldn't be any measurable performance hit. But the assertions will be active in CI for both the static and dynamic analysis tests, so should catch any latent bugs.

### DIFF
--- a/src/iff.imageio/iffoutput.cpp
+++ b/src/iff.imageio/iffoutput.cpp
@@ -83,13 +83,16 @@ private:
     }
 
     // helper to compress verbatim
-    void compress_verbatim(const uint8_t*& in, uint8_t*& out, int size);
+    void compress_verbatim(const uint8_t*& in, uint8_t*& out, int size,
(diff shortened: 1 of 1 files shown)


lgritz:
no objections -> merging

PR #4149, opened 2024-02-12
api: prototype using optional keyword/value params for ImageBufAlgo

lgritz:
Many IBA functions take lots of optional parameters, which are from time to time augmented, leaving us with a clutter of semi-deprecated versions.

This PR prototypes a new approach, which is to make IBA functions have explicit parameters for their core/necessary controls, and the minor optional behavioral control parameters are passed as keyword/value argument lists. I think this has a few really big advantages over explicit parameter lists with ever-growing optional parameters:

1. It makes the call sites very readable, since the keywords document what the parameters do.

2. It allows us to add (or deprecate) optional parameters without ABI breaking changes, and without proliferation of many versions of each function.

A new class is added: ParamValueSpan, which as the name implies, is simply a `span` of ParamValue structs (actually, it's a subclass that inherits from `span<const ParamValue>`, and adds a bunch of helper methods).  Within the IBA namespace, this is aliased to KWArgs, to make a short and self-documenting name to pass this collection of optional parameters. This leads to IBA declarations that look like this:

    ImageBuf OIIO_API resize(const ImageBuf &src, KWArgs options = {},
                             ROI roi = {}, int nthreads = 0);

and calls that looks like this:

    resize(dst, src, { { "filtername", "lanczos3" },
                       { "filterwidth", 6.0f } });

In this first stab, I have converted IBA::resize(), fit(), and warp(). That seems like enough to evaluate whether we like the direction this is going. If so, we'll tackle more IBA functions over time. I believe I've done this all in a way that preserves API and ABI compatibility for now

A few other odds and ends that come along for the ride:

* Filter1D and Filter2D have a new create_shared static method that returns a shared_ptr instead of raw pointer.

* Make all filter ctrs ok with 0 default filter width value, which means to use the natural default width for that filter.

* IBA::make_kernel is extended so that if you pass 0 for width or hight, it selects the a resolution using the default size for the named filter kernel. This way, the caller doesn't need to know the correct extents of the kernel, just the name.

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -669,8 +669,8 @@ bool OIIO_API rotate (ImageBuf &dst, const ImageBuf &src,
 ///
 ///     Advanced use: It is also possible to pass a custom reconstruction
 ///     filter as a `Filter2D*`, overriding any filtername and filterwidth
-///     that may also be passed. It needs to be be passed as
-///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
+///     that may also be passed. The easiest way to pass it is as:
+///     `make_pv("filterptr", raw_filter_ptr)`.
 ///     Use with caution!
 ///
 /// The caller may either (a) explicitly pass a reconstruction `filter`, or
@@ -745,8 +745,8 @@ bool OIIO_API resample (ImageBuf &dst, const ImageBuf &src,
 ///
 ///     Advanced use: It is also possible to pass a custom reconstruction
 ///     filter as a `Filter2D*`, overriding any filtername and filterwidth
-///     that may also be passed. It needs to be be passed as
-///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
+///     that may also be passed. The easiest way to pass it is as:
+///     `make_pv("filterptr", raw_filter_ptr)`.
 ///     Use with caution!
 ///
 ///   - "fillmode" : string (default: "letterbox")
@@ -848,8 +848,8 @@ bool OIIO_API fit(ImageBuf &dst, const ImageBuf &src, KWArgs options = {},
 ///
 ///     Advanced use: It is also possible to pass a custom reconstruction
 ///     filter as a `Filter2D*`, overriding any filtername and filterwidth
-///     that may also be passed. It needs to be be passed as
-///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
+///     that may also be passed. The easiest way to pass it is as:
+///     `make_pv("filterptr", raw_filter_ptr)`.
 ///     Use with caution!
 ///
 
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -191,8 +191,13 @@ private:
 
 
 
-
 namespace ImageBufAlgo {
+
+
+/// IBA::KWArgs is a span of ParamValue, used to pass keyword/value optional
+/// arguments to IBA functions.
+using KWArgs = ParamValueSpan;
+
 
 // old name (DEPRECATED 1.9)
 OIIO_DEPRECATED("use parallel_options (1.9)")
@@ -643,6 +648,31 @@ bool OIIO_API rotate (ImageBuf &dst, const ImageBuf &src,
 /// window of each correspond to each other, regardless of resolution).
 /// If `dst` is not yet initialized, it will be sized according to `roi`.
 ///
+/// The `options` list contains optional ParamValue's that control the
+/// resizing behavior.  The following options are recognized:
+///
+///   - "filtername" : string (default: "")
+///
+///     The type of reconstruction filter used to weight the `src` pixels
+///     falling underneath it for each `dst` pixel.  If the value is the
+///     empty string or not supplied, a reasonable high-quality filter will
+///     be chosen automatically (blackman-harris when upsizing, lanczos3
+///     when downsizing).
+///
+///   - "filterwidth" : float (default: 0)
+///
+///     The width of the reconstruction filter, expressed in pixel units of
+///     the `dst` image. If 0 or not supplied, the default width of the
+///     named filter will be used.
+///
+///   - "filterptr" : pointer to a Filter2D (default: nullptr)
+///
+///     Advanced use: It is also possible to pass a custom reconstruction
+///     filter as a `Filter2D*`, overriding any filtername and filterwidth
+///     that may also be passed. It needs to be be passed as
+///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
+///     Use with caution!
+///
 /// The caller may either (a) explicitly pass a reconstruction `filter`, or
 /// (b) specify one by `filtername` and `filterwidth`. If `filter` is
 /// `nullptr` or if `filtername` is the empty string `resize()` will choose
@@ -651,16 +681,11 @@ bool OIIO_API rotate (ImageBuf &dst, const ImageBuf &src,
 /// pixels falling underneath it for each `dst` pixel; the filter's size is
 /// expressed in pixel units of the `dst` image.
 
-ImageBuf OIIO_API resize (const ImageBuf &src,
-                          string_view filtername = "", float filterwidth=0.0f,
-                          ROI roi={}, int nthreads=0);
-ImageBuf OIIO_API resize (const ImageBuf &src, Filter2D *filter,
-                          ROI roi={}, int nthreads=0);
-bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src,
-                      string_view filtername = "", float filterwidth=0.0f,
-                      ROI roi={}, int nthreads=0);
-bool OIIO_API resize (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
-                      ROI roi={}, int nthreads=0);
+ImageBuf OIIO_API resize(const ImageBuf &src, KWArgs options = {},
+                         ROI roi = {}, int nthreads = 0);
+bool OIIO_API resize(ImageBuf &dst, const ImageBuf &src, KWArgs options = {},
+                     ROI roi = {}, int nthreads=0);
+
 /// @}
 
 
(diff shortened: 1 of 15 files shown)


nrusch:
My impression is that the `ParamValueSpan` is really nice for the reasons you mentioned: easy to add/deprecate features (even behind build flags), better self-documentation at call sites, etc.

However, I do want to point out what I see as the main tradeoffs, just to put them out there:

**1. Reduced discoverability/increased reliance on documentation**

Because of the loss of argument visibility, there's some increased pressure on rigorous comment and documentation maintenance. If that breaks down, the only option is to find and read the OIIO source itself to understand how a function consumes options from the `ParamValueSpan`.

There are some echoes here of Python's `**kwargs` syntax for key-value argument passing and all of the readability/maintenance headaches that come along with it. However, this is obviously a very different scenario, and there are some very real benefits that would come along with this, so I don't want to lean too heavily on that comparison.

**2. Loss of compile-time argument validation**

This one is fairly self-evident.

As a related follow-up question, how do you imagine these changes affecting the Python bindings? I suspect it would be a much bigger deal to lose the individual arguments there, since the binding wrappers have no embedded docstrings to refer to...

lgritz:
@nrusch To take your last question first: I don't see this changing the Python bindings at all, since Python already has the ability to specify optional arguments out-of-order with keyword identifiers:

    // C++ proposed:
    dst = resize(src, { { "filtername", "lanczos3" }, { "filterwidth", 6.0f } });

    # Python, already:
    dst = resize(src, filtername="lanczos3", filterwidth=6.0)

I'm not sure I see a need to alter the Python side. If anything, this moves the C++ side to appear a bit more like the Python in some nice ways (modulo some limitations of C++ syntax).

I do agree with your warning about how we're losing some compile-time checking. On the other hand, if there were previously several different optional params that take a float, it was already easy to screw up and pass the wrong value for the wrong one, without it being at all apparent at the call site and without any help from the compiler. So we're losing some compile-time checking, but at the same time we are gaining a different kind of checking and self-documentation and on the whole perhaps we are coming out ahead?

As for the discoverability, it's not discoverable from the raw function declaration, but I'm trying to be extra attentive to making the comments at each declaration be a really detailed description of the possible parameters (see the code in this PR for an example), and those get automatically incorporated into the web docs as well. So anybody who looks at the header or the docs, I'm confident, will get plenty of instruction about the options available.

What you didn't mention, and I am mildly concerned about, is simply that this is an unusual idiom to use in C++, so NEW people to OIIO might find it takes a little getting used to. It's not hard to get the hang of, but it's different. That isn't ideal, but again, it solves some other problems, especially with ABI stability, that I think make it worth it in the long run.

nrusch:
> To take your last question first: I don't see this changing the Python bindings at all [...]

Great! I just wasn't sure if you had additional ideas around refactoring those as well.

> So we're losing some compile-time checking, but at the same time we are gaining a different kind of checking and self-documentation and on the whole perhaps we are coming out ahead?

Makes sense to me. I think that sounds like a reasonable tradeoff given your point about the potential to mix up multiple similarly-typed arguments.

lgritz:
The reaction at last week's TSC meeting to this was positive.

Would anybody (especially from the TSC) care to give this PR an unequivocal approve, or express any remaining doubts, or make suggestions to improve the idioms I've chosen?

If approved, I'd be in favor of merging this and looking for other IBA functions to simplify in an analogous way.

lgritz:
@sfriedmapixar on the PR for the earlier stab at this, you expressed preference for the cspan approach. Does this look roughly like you were assuming?

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -307,13 +408,92 @@ warp_impl(ImageBuf& dst, const ImageBuf& src, const Imath::M33f& M,
...
+    IBA_check_optional(options, recognized);
+
+    Filter2D::ref filterref;
+    if (options.contains("filterptr", TypePointer)) {
+        auto f = options.find("filterptr", TypePointer);
+        OIIO_DASSERT(f != options.cend());
+        filterref = Filter2D::ref(f->get<const Filter2D*>(),
+                                  Filter2D::no_destroy);
+    } else {
+        filterref = get_warp_filter(options.get_string("filtername"),
```
ThiagoIze, line 460: This spooks me that `filtername` and `filterwidth` might have a typo here or in `recognized` and it'll be hard to catch unless there's a test for each of these. Same goes for all other places this kind of usage occurs. Could we create a single const variable somewhere, such as `const ustring OIIO_Strings::filtername("filtername")` that we can instead reference? This way if there's a typo when using the variable we'd get an immediate compile time error.

Whether this variable is exposed to users as well is another question. Exposing it would lend more of the same safety to users.

Another small benefit of this approach is that you wouldn't have the overhead of having to create these ustrings each time the function is called.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -307,13 +408,92 @@ warp_impl(ImageBuf& dst, const ImageBuf& src, const Imath::M33f& M,
...
+                   KWArgs options, ROI roi, int nthreads)
+{
+    static ustring recognized[] = { "filtername"_us,    "filterwidth"_us,
+                                    "wrap"_us,          "edgeclamp"_us,
+                                    "recompute_roi"_us, "filterptr"_us };
+    IBA_check_optional(options, recognized);
+
+    Filter2D::ref filterref;
+    if (options.contains("filterptr", TypePointer)) {
+        auto f = options.find("filterptr", TypePointer);
```
ThiagoIze, line 455: Why `contains` (which under the hood I think calls `find`) followed by `find`? Would it be simpler and faster to just do the `find` and then the `OIIO_DASSERT` code is what is used for the actual condition logic?

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -307,13 +408,92 @@ warp_impl(ImageBuf& dst, const ImageBuf& src, const Imath::M33f& M,
...
+
 bool
 ImageBufAlgo::warp(ImageBuf& dst, const ImageBuf& src, M33fParam M,
                    const Filter2D* filter, bool recompute_roi,
                    ImageBuf::WrapMode wrap, ROI roi, int nthreads)
 {
-    return warp_impl(dst, src, M, filter, recompute_roi, wrap,
-                     false /*edgeclamp*/, roi, nthreads);
+    return warp(dst, src, M,
+                { { "filterptr", TypePointer, 1, &filter },
```
ThiagoIze, line 493: What is the `1`? I can go look for the corresponding code to find out, but this is harder to follow than the old direct passing of a pointer. Is there something we can do to make this easier to read? If this is a count, could we add some logic so not including the count it defaults to 1? 

In OIIO internals, it's fine if we require the reader to be an advanced user of OIIO, but what about when users of OIIO need to pass optional pointers? We should try to make that not too painful.

ThiagoIze (commented):
Here are a few comments for the small amount of the code that I have looked at. Some of my suggestions I'm not too invested in, so please feel free to ignore them if you think it better to do so.

src/include/OpenImageIO/paramlist.h
```diff
@@ -503,4 +503,138 @@ class OIIO_UTIL_API ParamValueList : public std::vector<ParamValue> {
...
+    const_iterator find(string_view name, TypeDesc type = TypeUnknown,
+                        bool casesensitive = true) const;
+    const_iterator find(ustring name, TypeDesc type = TypeUnknown,
+                        bool casesensitive = true) const;
+
+    /// Case insensitive search for an integer, with default if not found.
+    /// Automatically will return an int even if the data is really
+    /// unsigned, short, or byte, but not float. It will retrieve from a
+    /// string, but only if the string is entirely a valid int format.
+    int get_int(string_view name, int defaultval = 0, bool casesensitive = true,
```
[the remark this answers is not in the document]
lgritz: Which way do you think is better in this case? I think for ImageSpec metadata, we search case-insensitively. I don't know if that's the "right" choice. But since it would be so confusing and unwise to have two separate parameters that differed only in case, I don't see an advantage to encouraging it. On the other hand, case-sensitive string compares are a lot faster. Opinions?

src/include/OpenImageIO/paramlist.h
```diff
@@ -503,4 +503,138 @@ class OIIO_UTIL_API ParamValueList : public std::vector<ParamValue> {
...
+        return cspan<ParamValue>::operator[](idx);
+    }
+
+    const_iterator find(string_view name, TypeDesc type = TypeUnknown,
+                        bool casesensitive = true) const;
+    const_iterator find(ustring name, TypeDesc type = TypeUnknown,
+                        bool casesensitive = true) const;
+
+    /// Case insensitive search for an integer, with default if not found.
+    /// Automatically will return an int even if the data is really
```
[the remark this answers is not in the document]
lgritz: I think we don't need to worry until there are any uses of 64 bit parameters in these keyword lists? We can always add a get_int64/get_uint64 as needed later.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -307,13 +408,92 @@ warp_impl(ImageBuf& dst, const ImageBuf& src, const Imath::M33f& M,
...
+// Return true if all are fine, false if not.
+static bool
+IBA_check_optional(ImageBufAlgo::KWArgs options, cspan<ustring> recognized,
+                   cspan<ustring> obsolete = {})
+{
+    bool ok = true;
+    for (auto&& pv : options) {
+        ok &= IBA_find_optional(pv, recognized, obsolete);
+    }
+    return ok;
```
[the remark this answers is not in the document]
lgritz: I just think that the parameter lists are very short and the usual case is that there are no mistakes, so an early out for unknown parameters seems unlikely to be a performance issue. As for simpler, I guess it's a matter of taste. I don't have a specific objection to taking the extra out, but it is (inconsequentially so) an extra branch and adding multiple exit points from the function? Either way I guess.

src/libOpenImageIO/imagebufalgo_xform.cpp
```diff
@@ -307,13 +408,92 @@ warp_impl(ImageBuf& dst, const ImageBuf& src, const Imath::M33f& M,
...
+}
+
+
+
+bool
+ImageBufAlgo::warp(ImageBuf& dst, const ImageBuf& src, M33fParam M,
+                   KWArgs options, ROI roi, int nthreads)
+{
+    static ustring recognized[] = { "filtername"_us,    "filterwidth"_us,
```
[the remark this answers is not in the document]
lgritz: I'm not sure it's really different, but sure, it conveys that these are not lists that can be written to.

src/libOpenImageIO/imagebufalgo_xform.cpp
lgritz: Oh, I see your point. Sure, I can do that.

src/libOpenImageIO/imagebufalgo_xform.cpp
lgritz: I've rearranged this better.

src/libOpenImageIO/imagebufalgo_xform.cpp
lgritz: Ah, this is related to the actual constructor for a ParmValue, which takes 
* name
* the type T
* the number of T items
* a `T*` pointing to the values to copy

This is the general constructor, but it's a little awkward for just one item of common types, so we have helper constructors for the extremely common cases of needing one int, float, or string. There is currently no direct simple constructor for pointer types, but maybe I should add one. I had noticed when I wrote this that it was falling into the awkward general case. Let me see what I can throw together.

lgritz: The update I pushed helps this, I think.

lgritz:
A variety of improvements based on this feedback have been incorporated into an update to this PR.

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -762,58 +787,77 @@ bool OIIO_API fit (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
...
+///     If nonzero, will enable special edge clamp behavior to reduce
+///     artifacts at the image edges (experimental).
+///
+///   - "recompute_roi" : int (default: 0)
+///
+///     If nonzero and `dst` is not yet initialized, the result image
+///     will be sized to be large enough to hold the warped image. If
+///     zero (the default), the `dst` image will have the same ROI as
+///     `src`. If the `dst` image already is initialized, its size will
+///     not be chaned and this option will be ignored.
```
ThiagoIze, line 845:
```suggestion
///     not be changed and this option will be ignored.
```

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -762,58 +787,77 @@ bool OIIO_API fit (ImageBuf &dst, const ImageBuf &src, Filter2D *filter,
...
+///
+///   - "filterptr" : pointer to a Filter2D (default: nullptr)
+///
+///     Advanced use: It is also possible to pass a custom reconstruction
+///     filter as a `Filter2D*`, overriding any filtername and filterwidth
+///     that may also be passed. It needs to be be passed as
+///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
```
ThiagoIze, line 852: give simpler way for single item? `make_pv`?

src/include/OpenImageIO/paramlist.h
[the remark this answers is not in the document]
ThiagoIze: I agree with your points, but I don't have a strong preference for what to default to. My main concern is that the comment and the code are currently opposite.
If I do have to pick something, I would probably default to case insensitive so it gives the results users likely expect. If anyone really needs the speed, they can still request the case sensitive fast path.

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -643,6 +648,31 @@ bool OIIO_API rotate (ImageBuf &dst, const ImageBuf &src,
...
+///     The width of the reconstruction filter, expressed in pixel units of
+///     the `dst` image. If 0 or not supplied, the default width of the
+///     named filter will be used.
+///
+///   - "filterptr" : pointer to a Filter2D (default: nullptr)
+///
+///     Advanced use: It is also possible to pass a custom reconstruction
+///     filter as a `Filter2D*`, overriding any filtername and filterwidth
+///     that may also be passed. It needs to be be passed as
+///     `ParamValue("filterptr", OIIO::TypePointer), 1, &raw_filter_ptr)`.
```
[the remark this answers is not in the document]
lgritz: yes, fixing

src/include/OpenImageIO/imagebufalgo.h
lgritz: yes, done and pushed

ThiagoIze (approved):
This looks good to me.

PR #4150, opened 2024-02-12
fix: Fixes to buildinfo queries

lgritz:
* Make sure clang variants are before clang and things that might emulate gcc are before gcc.
* Use the right symbol for detecting ICC
* Capitalize `__linux__` correctly, oops.


lgritz:
I believe this can't break anything and fixes what is definitely an error. 

No objections -> merging.

PR #4162, opened 2024-02-20
Allow TypeDesc to be used directly in a C API by forcing it to be POD.

scott-wilson:
Signed-off-by: Scott Wilson <scott@propersquid.com>

## Description

This PR is for allowing the TypeDesc type to be used directly in a C API without indirection, and therefore also in a Rust API without indirection.

## Tests

I added static asserts for the TypeDesc so it should validate if the struct is okay at compile time.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/typedesc.h
+++ b/src/include/OpenImageIO/typedesc.h
@@ -164,10 +164,7 @@ struct OIIO_UTIL_API TypeDesc {
     TypeDesc (string_view typestring);
 
     /// Copy constructor.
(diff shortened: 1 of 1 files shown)


lgritz:
I think there are a couple CI failures from compilers that don't accept static_assert in C++14 without the explanation string.

(Don't worry about the "bleeding edge" test, that is failing because something is broken in the openexr trunk, it's not your fault.)

scott-wilson:
No worries, will fix shortly.

lgritz (approved):
LGTM, thanks!

PR #4166, opened 2024-02-29
fix: Add some color space aliases

### LINKED ISSUE #4165: Feature request: improve sRGB parsing
Howdy all, a new beta version of Gaffer is out and it uses a newer version of OIIO. A number of us ran into a problem when the Gaffer launcher couldn't find an sRGB color space.

In my case I traced the problem to using a custom OCIO config and OIIO being unable to find the keywords it's looking for.  My config's sRGB texture section originally read as follows:
```
name: sRGB texture
aliases: [srgb_tx, Utility - sRGB texture, srgb_texture, Input - Generic - sRGB texture]
```
OIIO couldn't parse this. I was able to fix the problem by either changing the `name` to `sRGB - Texture` or adding a new `srgb` alias. I should note that my config is based on the official ASWF CG config which also does not have an `srgb` alias.

It seems like one solution might be to add some more aliases from the ASWF config to OIIO's search: maybe `srgb_tx` and `srgb_texture`. Thank you!

lgritz:
Automatically recognize some additional color space name synonyms:

"srgb_texture", "lin_rec709" and "lin_ap1" are names that MaterialX and USD use as canonical names.

Also add common permutation "srgb_tx" and "srgb texture" as additional aliases for srgb.

Fixes #4165

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -592,15 +592,20 @@ ColorConfig::Impl::classify_by_name(CSInfo& cs)
     // believe them! Woe be unto the poor soul who names a color space "sRGB"
     // or "ACEScg" and it's really something entirely different.
     if (Strutil::iequals(cs.name, "sRGB")
+        || Strutil::iequals(cs.name, "srgb_tx")
+        || Strutil::iequals(cs.name, "srgb_texture")
+        || Strutil::iequals(cs.name, "srgb texture")
         || Strutil::iequals(cs.name, "sRGB - Texture")) {
         cs.setflag(CSInfo::is_srgb, srgb_alias);
     } else if (Strutil::iequals(cs.name, "Rec709")) {
         cs.setflag(CSInfo::is_Rec709, Rec709_alias);
     } else if (Strutil::iequals(cs.name, "lin_srgb")
+               || Strutil::iequals(cs.name, "lin_rec709")
                || Strutil::iequals(cs.name, "Linear Rec.709 (sRGB)")) {
         cs.setflag(CSInfo::is_lin_srgb | CSInfo::is_linear_response,
                    lin_srgb_alias);
-    } else if (Strutil::iequals(cs.name, "ACEScg")) {
+    } else if (Strutil::iequals(cs.name, "ACEScg")
+               || Strutil::iequals(cs.name, "lin_ap1")) {
         cs.setflag(CSInfo::is_ACEScg | CSInfo::is_linear_response,
                    ACEScg_alias);
     }
@@ -728,6 +733,8 @@ ColorConfig::Impl::reclassify_heuristics(CSInfo& cs)
 void
 ColorConfig::Impl::identify_builtin_equivalents()
 {
+    if (disable_builtin_configs)
+        return;
 #if OCIO_VERSION_HEX >= MAKE_OCIO_VERSION_HEX(2, 3, 0)
     Timer timer;
     if (auto n = IdentifyBuiltinColorSpace("srgb_tx")) {


johnhaddon:
Thanks for getting on to this so swiftly Larry. We've had reports of the same problem from several folks now, and we don't have access to the relevant OCIO configs in all cases. What I'm trying to figure out is if this PR is likely to fix all of those, or if it potentially only works for Brian's.

What's bugging me is that I don't fully understand why everything worked for everyone without these additional synonyms in OIIO 2.4. My suspicion is that it might be because the fall-through to OIIO's own conversions was removed here :

https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/src/libOpenImageIO/color_ocio.cpp#L1813-L1816

From the comment, the assumption seems to be that we won't get here for OCIO >= 2.2 because we'll be using the built-in config. But if we're not using the built-in config, and instead are using some arbitrary user config, perhaps keeping the fall-through would be appropriate? I _think_ that means we'd be back to the OIIO 2.4 behaviour for these problematic cases.

Bit more background on this : we do "proper" full OCIO-based color management in all Gaffer's viewers, but are still relying on the basic sRGB->linear conversions in a couple of places in the UI where we're loading icons and suchlike. If dealing with this in OIIO is problematic, it might be that we can do something on our side for those few special cases.

lgritz:
I see. So we might be worried about cases where OIIO is built with OCIO support (but less than 2.2 where it will have the built-in configs to rely on), *and* a config is found, but that config does not seem to have any color space whose name matches any of the things it's looking for (as of this PR: "sRGB", "srgb_tx", "srgb_texture", "srgb texture", "sRGB - Texture").

I think we should start with this PR, if it seems to you like it's a step forward (just accept it in the review dialog). Then we can tackle the second part, which I think can be addressed by, instead of making the "use config" and "use fallbacks" mutually exclusive, we can rig it to use our fallbacks even if there is a config, if that config doesn't seem to have the requested color spaces.

lgritz:
By the way, since I have your attention, @johnhaddon, could I trouble you to please sign onto #3905 if you are able? (And excuse me if my recollection is incorrect that you have not done so yet.)

johnhaddon (approved):
> So we might be worried about cases where OIIO is built with OCIO support (but less than 2.2 where it will have the built-in configs to rely on), and a config is found, but that config does not seem to have any color space whose name matches any of the things it's looking for

Almost, but we're actually seeing this with OCIO **2.2.1** (and OIIO 2.5.8.0). I think the existence of a user-supplied config (without the recognised names) is somehow preventing OIIO from using the built-in OCIO config. In the past (OIIO 1.4) that was OK, because we then fell through and used the OIIO-supplied `ColorProcessor_sRGB_to_linear`. But that is now gated behind [`#if OCIO_VERSION_HEX < MAKE_OCIO_VERSION_HEX(2, 2, 0)`](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/src/libOpenImageIO/color_ocio.cpp#L1813-L1816). In my previous reply I was suggesting that perhaps that `#ifdef` could be removed, so we always have the final backstop in place.

> I think we should start with this PR, if it seems to you like it's a step forward

Given the situation, adding more aliases does seem reasonable. I feel like maybe what we all really need though is a step _backwards_ - with OCIO enforcing canonical names for well-defined spaces, and then with the creative/opinionated stuff built on top. Applications are currently at the mercy of config authors if they want to be able to rely on any "standard" spaces.

I now realise though that I don't think we even want the full OCIO transforms in this limited use case - we just want the sRGB transfer function so we can composite icons in linear space, but with no change in primaries. So some way of always getting a `ColorProcessor_sRGB_to_linear` regardless of OCIO configs. I dunno, maybe we're asking too much - if you want, we can certainly just deal with this directly on our side.

Many thanks for taking the time to look at this...
Cheers...
John

johnhaddon:
> could I trouble you to please sign onto https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/3905 if you are able?

Done, and sorry for the delay.

johnhaddon:
> Given the situation, adding more aliases does seem reasonable.

Although I don't think this very specific corner of Gaffer warrants a proliferation of aliases if you think they're detrimental to the  ecosystem as a whole. If it fits with OIIO's philosophy then that's great, but we can definitely just deal with this on the Gaffer side if you'd prefer.

zachlewis:
> I now realise though that I don't think we even want the full OCIO transforms in this limited use case - we just want the sRGB transfer function so we can composite icons in linear space, but with no change in primaries. So some way of always getting a `ColorProcessor_sRGB_to_linear` regardless of OCIO configs. I dunno, maybe we're asking too much - if you want, we can certainly just deal with this directly on our side.

Sounds like a great use case for OCIO [NamedTransforms](https://opencolorio.readthedocs.io/en/latest/api/namedtransform.html#named-transform).

The ASWF CG and Studio configs contain a named transform called "sRGB - Curve", containing just the transfer function.

``` 
named_transforms:
...
 - !<NamedTransform>
    name: sRGB - Curve
    aliases: [srgb_crv, Utility - Curve - sRGB, crv_srgb]
    transform: - !<ExponentWithLinearTransform> {gamma: 2.4, offset: 0.055}
```

One would apply a NamedTransform with OCIO's overloaded Config.getProcessor method, e.g:

```
import PyOpenColorIO as ocio
cfg = ocio.Config.CreateFromBuiltinConfig("default")
proc = cfg.getProcessor("sRGB - Curve", ocio.TRANSFORM_DIR_FORWARD)
cpu = proc.getDefaultCPUProcessor()
cpu.applyRGB(...)
```

nrusch:
I don't want to add too much extra noise to this discussion, but it seems like there might be room to support OIIO-specific role names that could be leveraged by OCIO config authors. This would add an extra axis of flexibility by enabling authors to point OIIO at existing colorspaces that may not adhere to the expected set of possible naming conventions.

lgritz:
I think that the main thing we want to do is support [this list of standard names](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#color-spaces-and-color-management-systems) that is evolving to be the set of canonical color space names that ACES, MaterialX, USD, and OCIO built-in configs are converging upon. We want to

(a) support those all the time, regardless of any other OCIO availability, version, or config.
(b) expect/demand that people using those color spaces will use those canonical names (or, in some cases, accept a few very obvious or common aliases).
(c) if OCIO support is present and a config is available, additionally support any other color spaces in that config.

---

A detail of part (a) is the degree to which we can and should rely on OCIO itself to provide that support, or if we need to write our own fallbacks for any that are missing. I think that OCIO 2.2 has both the built-in configs with these spaces available, and 2.3 additionally has some utilities to tell us if two color spaces are equivalent. One possible strategy is to just say "you need OCIO 2.2/2.3+, or all color management bets are off." Another would be to require OCIO as a mandatory dependency. Another is to make OIIO's build system automatically build a self-contained static OCIO if one is not found or not a sufficiently new version. Any are potentially on the table.

lgritz:
I would perhaps recommend that the most synergistic approach is for Gaffer to also use and support the very same list of canonical color space names. The more apps in our ecosystem all subscribe to the same canonical set, the more valuable it is for all the other apps.

lgritz:
I'm going to merge this, but we're going to keep iterating on approaches to make the color management more reliable.

doug-walker:
I'd like to mention that coming up with a standard set of color space names is a top priority of the new ASWF Color Interop Forum, which will be a monthly meeting starting March 25. The OCIO, OpenUSD, and MaterialX projects have agreed that this will be the working group that recommends a set of standard core color space names (along with an implementation for each).

Please note that the list of color spaces you linked to above from MaterialX, while certainly an important input to this group, is different from what is currently used by OCIO and ACES. The final recommendation will probably not exactly match what is used by any of the projects.

zachlewis:
> A detail of part (a) is the degree to which we can and should rely on OCIO itself to provide that support, or if we need to write our own fallbacks for any that are missing. I think that OCIO 2.2 has both the built-in configs with these spaces available, and 2.3 additionally has some utilities to tell us if two color spaces are equivalent. One possible strategy is to just say "you need OCIO 2.2/2.3+, or all color management bets are off." Another would be to require OCIO as a mandatory dependency. Another is to make OIIO's build system automatically build a self-contained static OCIO if one is not found or not a sufficiently new version


I'd love to see OCIO turn into a mandatory dependency, built statically by the build system if missing; and OIIO-3.0 seems like a pretty good opportunity for that... 

Currently, OIIO supports building with OCIO-2.1+, right? I think that's probably sufficient, presuming OIIO ships with an "internal" OCIO config defining the desired built-in color spaces, I believe OCIO-2.1's `Config::GetProcessorFromConfigs()` should be able to create Processors for ColorSpaceTransforms and DisplayViewTransforms spanning OIIO's internal OCIO config and user OCIO configs, provided the internal and user configs both define either an `aces_interchange` role or a `cie_xyz_d65_interchange` role. 

(OCIO-2.2 is able to infer a connection space for user configs that don't define those interchange roles; but apart from that, It's not clear to me that any of the color space equivalency checking stuff is necessary or useful for image processing. Maybe useful for metadata + logging, in case a user specifies an OIIO name for a color space that also happens to exist in a user-provided config)

johnhaddon:
> Sounds like a great use case for OCIO [NamedTransforms](https://opencolorio.readthedocs.io/en/latest/api/namedtransform.html#named-transform).

@zachlewis, thanks, this suggestion led me to the realisation that I can construct my own `OIIO::ColorConfig` with the standard ACES config, insulating myself from whatever is in `getenv( "OCIO" )`. I hadn't spotted the `ImageBufAlgo::colorconvert()` overload that took a `ColorConfig`, but now I have, I think I have all I need independently of any future OIIO improvements.

> I would perhaps recommend that the most synergistic approach is for Gaffer to also use and support the very same list of canonical color space names.

@lgritz, absolutely, we have no intention of straying from the herd. We're already using OCIO for all user-facing color configuration, so from what I understand, we'll be supporting the canonical names as soon they're available there.

> A detail of part (a) is the degree to which we can and should rely on OCIO itself to provide that support, or if we need to write our own fallbacks for any that are missing.

Requiring OCIO would be absolutely fine for us, although I wonder if it might be a barrier for usage outside of VFX. My only suggestion would be that if OIIO _does_ continue to provide its own transforms, it would be nice if they were available explicitly via something like `ColorConfig( "oiio" )` - no fallbacks, no heuristics, just a straightforward guarantee that those are the ones you're getting.

PR #4167, opened 2024-03-01
ci: Fix Windows CI, need to build newer openexr and adjust boost search

lgritz:


lgritz:
This fixes broken CI, merging.

PR #4169, opened 2024-03-02
docs: Fix some typos and add missing oiiotool expression explanations

lgritz:


lgritz:
Preemptively merging because it's only docs and comment fixes, can't break anything

PR #4173, opened 2024-03-03
int(simd.h): Make the simd types trivially_copyable

lgritz:
Basically, this is a way of saying memcpy of any of these simd types is fine. It doesn't matter a lot for OIIO itself, but it helps quiet some static analysis warnings for OSL.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -10360,6 +10360,24 @@ template<> struct formatter<OIIO::simd::matrix44>
     : OIIO::pvt::array_formatter<OIIO::simd::matrix44, float, 16> {};
 } // namespace fmt
 
+
+// Allow C++ metaprogramming to understand that the simd types are trivially
+// copyable (i.e. memcpy to copy simd types is fine).
+namespace std {  // not necessary in C++17, just say std::is_trivially_copyable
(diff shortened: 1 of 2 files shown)


ThiagoIze:
Could the approach from https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4110#discussion_r1451873540 be used?

lgritz:
> Could the approach from [#4110 (comment)](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4110#discussion_r1451873540) be used?

I think that can't work in this case (I tried) because these types contain a union. I've tried a bunch of different things, and I can't seem to coax C++ into figuring it out on its own for these oddball types, even though we humans know that it's just memory and are indeed trivially copyable.

ThiagoIze:
In my quick test it seems to work: https://gcc.godbolt.org/z/bdxYbY7dv (toggle the #if to see the assert fail and then pass with the `default`). Could it be maybe that there's only some of the classes that won't work?

lgritz:
Just as an update, looking into this sent me down a rathole that might end up being a much larger overhaul to the simd internals. Will post an update in a few days, I'm working on it in dribs and drabs.

lgritz:
I'm going to withdraw this PR and start a new one, I got this working in a better way.

PR #4174, opened 2024-03-03
dpx: use OIIO functions for byte swapping to make Sonar happy

lgritz:


ThiagoIze (approved):
This looks fine. What do you think about getting rid of the old code now that it's not used?

lgritz:
This is imported code from another project, so I have tried over the years to make the absolute minimal textual changes possible, because I'm superstitious about trying to preserve our ability to easily diff against the original and cherry-pick any future updates or fixes to the original project (we don't clang-format these particular files, either, for this reason).

I think the original project is no longer active, so maybe as possibly the only maintained copy, we should just consider it irrevocably forked and not worry about it? But so far I have resisted that.

So yes, we *can* remove the old code (and reformat, and get rid of other parts that just aren't used by OIIO), if there's a benefit to doing so that outweighs preserving the ability to directly compare against the originals.

ThiagoIze:
I see, in that case keeping the old code would indeed make it easier to handle merge conflicts. 

Would it maybe help future readers to add a comment in the code explaining why there's two versions? Something about sonar and keeping merge conflicts simple?

lgritz:
https://github.com/PatrickPalmer/dpx
Hasn't been touched in 15 years! So perhaps we should end the charade. But maybe not as part of this PR. I'll add a comment and call it a day.

ThiagoIze (approved):

PR #4176, opened 2024-03-05
build: Cleanup - get rid of "site" files

lgritz:
Remove some SPI-specific make and cmake customizations from long ago, we haven't used those makefiles to build for a long time. Also remove the detectplatform.mk file, we haven't used the results of what it does for a while. This just declutters.


grdanny (approved):
getting rid of all the app config stuff/rez/spcomp sound like a good thing to me.

PR #4177, opened 2024-03-05
fix(build): Fix warning when Freetype is disabled

lgritz:
A variable was declared unconditionally but only used when Freetype is enabled. Move its declaration to be within the right conditional, on order to avoid compiler warnings about unused static variables.


dgovil:
This works for me if I patch out the lines in my copy of the 2.5.9 tag.

PR #4182, opened 2024-03-11
fix: Mismatched pragma push/pop, oops

lgritz:


ThiagoIze (approved):
Looks good to me: the corresponding push is three lines above.

PR #4189, opened 2024-03-16
chore: more convertion to new string formatting

lgritz:
In this PR, we convert nearly all `sprintf` uses to `format`.

It should not change any behavior or compatibility. Nonetheless, I think this is destined for master only and won't be backported, to reduce risk.

Along the way, we add format_to and format_to_n to Strutil.


lgritz:
I'm going to merge this by tomorrow if I don't hear any reservations.

PR #4190, opened 2024-03-17
build: make an OpenImageIO_Util_static library and target

lgritz:
Add a `OpenImageIO_BUILD_STATIC_UTIL_LIBRARY` CMake variable that, if enabled, will build a separately named static version of the OpenImageIO_Util library, in addition to the dynamic one. It has a separate name, libOpenImageIO_Util_static, and a separate exported cmake target, `OpenImgaeIO::OpenImageIO_Util_static.`

This can be a help to applications that only need OpenImageIO_Util, and not the main OpenImageIO library, and furthermore would prefer a static library so they don't need a runtime dependency on our dynamic libraries.

Cavat emptor: Beware that there are singletons inside ustring, the oiio thread pool, and possibly others I'm not remembering at the moment. It could be a bad thing to have an application that contains both the static and dynamic versions of libOpenImageIO_Util, or to have multiple copies of the static library inside the same executable (including transitively through dependency libraries that were built independently). You have been warned!

Note also that this is swimming upstream in cmake land. For reasons I never quite understood, cmake assumes that a project wants to build either static or dynamic libraries, but provides little direct support for doing both in the same build. So we're taking the matter into our own hands, perhaps. If somebody thinks the way I'm proposing is very wrong, please do speak up.

### DIFF
--- a/src/libutil/CMakeLists.txt
+++ b/src/libutil/CMakeLists.txt
@@ -9,57 +9,6 @@ set (libOpenImageIO_Util_srcs argparse.cpp benchmark.cpp
                   strutil.cpp sysutil.cpp thread.cpp timer.cpp
                   typedesc.cpp ustring.cpp xxhash.cpp)
 
-add_library (OpenImageIO_Util ${libOpenImageIO_Util_srcs})
-target_include_directories (OpenImageIO_Util
-        PUBLIC
-            $<INSTALL_INTERFACE:include>
-        )
-target_link_libraries (OpenImageIO_Util
-        PUBLIC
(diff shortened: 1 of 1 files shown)


lgritz:
Any opinions on this? Bad idea? Good idea? Or "ok idea, but let's make the default be OFF for now?"

lgritz:
Objections? I'd like to merge this today if there's nothing standing in the way. We can always revise if anybody thinks the default should be to not build a static library.

lgritz:
No objections -> merging

PR #4193, opened 2024-03-20
build: switch to target-based definitions

lgritz:
We had lots of stray usage of add_defintions, add_compile_options,
etc.  In modern cmake, this is considered bad form, since it makes
subprojects pollute the global option lists. A better way is to add
all these to specific targets, so now that's what we do in this PR.
This should make all the definitions and options we want to set for
compilng OIIO's code not have any interference with subprojects, or
with higher-level surrounding projects.


lgritz:
Opinions on this approach?

lgritz:
Forgot to mention this in TSC today. Any thoughts? Does this approach look ok?

lgritz:
Nobody has taken the chance to object or comment. I think I will merge. I believe we can't go wrong in cmake land by ensuring that all our compiler directives are applied directly to our targets and don't pollute any surrounding project, which is really all this is doing.

lgritz:
Apologies: I think this whole PR was unnecessary, as the built-in functions I'm replacing here already are "scoped" to the current directory and below, and don't pollute any surrounding project. I will propose a new PR that reverts this, for simplicity.

PR #4195, opened 2024-03-21
ci: Enable Windows 2022 CI tests

lgritz:
Needed some fixups with its python version to get it fully working.


lgritz:
Merging without review, since all it does is add another CI  test matrix case.

PR #4198, opened 2024-03-23
refactor: Make span default ctr and assignment be `= default`

lgritz:
Minor simplification.

### DIFF
--- a/src/include/OpenImageIO/span.h
+++ b/src/include/OpenImageIO/span.h
@@ -113,7 +113,7 @@ public:
     /// Construct from a fixed-length C array.  Template magic automatically
     /// finds the length from the declared type of the array.
     template<size_t N>
-    constexpr span (T (&data)[N]) : m_data(data), m_size(N) { }
+    constexpr span (T (&data)[N]) noexcept : m_data(data), m_size(N) { }
 
     /// Construct from std::vector<T>.
     template<class Allocator>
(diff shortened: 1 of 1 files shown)


src/include/OpenImageIO/span.h
```diff
@@ -143,11 +143,7 @@ class span {
         : span (il.begin(), il.size()) { }
 
     /// Assignment copies the pointer and length, not the data.
-    span& operator= (const span &copy) {
-        m_data = copy.data();
-        m_size = copy.size();
-        return *this;
-    }
+    constexpr span& operator= (const span &copy) = default;
```
ThiagoIze, line 146: This change is pretty close to what [std::span does](https://en.cppreference.com/w/cpp/container/span/operator%3D). The only bit missing, if you want it to fully match, is marking it noexcept:
```c++
constexpr span& operator=( const span& other ) noexcept = default;
```

ThiagoIze (approved):
Looks fine. You could mark it `noexcept` if it's important to you that you perfectly match std::span.

src/include/OpenImageIO/span.h
lgritz: You're right. And I noticed a few other spots where we can be noexcept. Testing now and will amend the PR if it works.

lgritz: updated

ThiagoIze (approved):

PR #4199, opened 2024-03-23
build!: Move to C++17 minimum

lgritz:
These are build-breaking changes that impose a variety of new minimum dependencies for the toolchain.

C++17 is the new minimum. All hail C++17.

Bump minimum compiler versions to the respective releases of each one that is requred to fully support C++17 for us. This means new minimum changes of gcc 6.1 -> 9.3, clang 3.4 -> 5.0, icc 17 -> 19, and MSVS v19.14 or higher now.

Remove CI tests for too-old platforms, and adjust some others that needed particular components upgraded. We now will no longer attempt to test on "VFX Platform 2019" or 2020. The 2021 platforms are the earliest we claim to support (reflecting VFXP's guidance about trying to support up to 3 years back).

By removing the old test cases, we're also removing the last platforms that test Python 2.7. So while this PR isn't the one that officially removes the last vestiges of Python 2, it does mean that once this is merged, we will no longer be testing less than Python 3.7. (A separate PR removing all Python 2 support will follow and presumably be merged around the same time.)

With C++17 min, remove traces of boost::filesystem. We can rely on std::filesystem now.

Clean up pre-C++17 macros in platform.h

Needless to say, this will not be backported to any release branches, which guarantee that they will never raise the minimum supported versions of any dependencies. This is just for future OIIO. I won't merge this until after April 1, at which point I will also tag the last spot in master that is known to work with C++14/gcc6.


ThiagoIze (approved):

src/libutil/filesystem.cpp
```diff
@@ -494,22 +456,12 @@ Filesystem::temp_directory_path()
...
-#    if defined(_WIN32)
-    // boost internally doesn't use MultiByteToWideChar (CP_UTF8,...
-    // to convert char* to wchar_t* because they do not know the encoding
-    // See boost/filesystem/path.hpp
-    // The only correct way to do this is to do the conversion ourselves
+#if defined(_WIN32)
     std::wstring modelStr = Strutil::utf8_to_utf16wstring(model);
-#    else
+#else
     std::string modelStr = model.str();
```
jessey-git, line 463: This `modelStr` variable is not used (neither before nor after these changes). Is that a bug?

jessey-git (approved):
Looks good. Just have 1 comment from the existing code that I noticed.

src/libutil/filesystem.cpp
lgritz: I think that's an entirely separate bug! I'll make a separate PR for it since it's not C++17 specific.

lgritz: This topic moving to #4203

PR #4200, opened 2024-03-24
build!: Move to Python 3.7 minimum

lgritz:
These are build-breaking changes that impose a variety of new minimum dependencies for the toolchain.

The new minimum Python version supported is now 3.7.

Get rid of 'from future' lines in python scripts in testsuite.

Remove python 2.7 specific ref images from testsuite.

Needless to say, this will not be backported to any release branches, which guarantee that they will never raise the minimum supported versions of any dependencies. This is just for future OIIO. I won't merge this until after April 1, at which point I will also tag the last spot in master that is known to work with Python 2.7.


grdanny (approved):
looks great to me. nice to see those future import go away.

PR #4201, opened 2024-03-24
ci: Update scrorecard workflow to fix breakage

lgritz:
Also rename file so it matches what's listed in the file to trigger it.


lgritz:
Works, merging.

PR #4203, opened 2024-03-25
fix: Filesystem::unique_path wasn't using the unicode rectified string

lgritz:


src/libutil/filesystem.cpp
```diff
@@ -517,8 +517,9 @@ Filesystem::unique_path(string_view model)
     std::lock_guard<std::mutex> lock(mutex);
     std::string name;
     while (true) {
-        name = model.str();
-        for (size_t i = 0, e = model.size(); i < e; ++i)
+        name = modelStr;
```
jessey-git, line 520: For the Windows case, `name` is `std::string` here but `modelStr` is a `std::wstring` so I don't think this will work. CI is probably using boost filesystem for the windows job so didn't surface the issue.

lgritz: But what I don't quite understand is, why did this pass CI on Windows? Is there an implicit string=wstring allowed, and so it's not a compile error but would possibly have been a runtime misbehavior?

jessey-git: The windows CI uses boost. It traces out `-- Using Boost::filesystem` during the Build step at least, which means the main `#ifdef USE_BOOST_FILESYSTEM` takes over just above this.

jessey-git (approved):
Looks good now I think.

PR #4206, opened 2024-03-28
Add Metadata for OpenEXR

TodicaIonut:
## Description

Add Metadata for OpenEXR

## Checklist:

  nonconforming.


lgritz:
Hi, thanks for this effort, but I think you might misunderstand what this table is.

OIIO is meant to make a common API for reading and writing image files, regardless of format, so we strive to make it so that for most ordinary operations, the app doesn't need to know or care what type of image file it is. But one problem from that is that different image formats have different conventions about what to call certain metadata. We don't want every app to need to understand all these details -- for example, that the same information may be called "ImageDescription" for TIFF, "comments" for OpenEXR, or "caption" for IPTC, etc. So for certain common metadata, we picked a single canonical name for users of the OIIO APIs, and the readers/writers will automatically rename it coming in or going out to the files.

You can see the list of these canonical names in the documentation: https://openimageio.readthedocs.io/en/v2.5.9.0/stdmetadata.html

So what that table is in exroutput.cpp (and there are corresponding ones for input in exrinput.cpp and exrinput_c.cpp that you didn't modify) is the translation key -- **the list of the canonical "OIIO" names for [certain common metadata](https://openimageio.readthedocs.io/en/v2.5.9.0/stdmetadata.html), and the names that OpenEXR says they should use when saving to OpenEXR files.**

Any metadata that doesn't appear in [this list](https://openimageio.readthedocs.io/en/v2.5.9.0/stdmetadata.html) doesn't need an entry in the table, because it doesn't have a conventional OIIO name that must be automatically kept the same across all file formats.

If a metadata name is not in the table, it doesn't mean that it can't be saved in an OpenEXR file. It just means that it'll keep whatever name you give it.

Some of the entries you put in the table, like "CameraTint", are definitely not listed in that metadata chapter of the docs and appear nowhere in the code base, so no translation of the names is needed, thus it doesn't need to be in the table. Other items you put in the table, such as "CameraMake" already **IS** in that chapter, so should be using the canonical name we already chose, i.e., I think that should read

    ExrMeta("Make", "cameraMake", TypeString),

which means "metadata having the conventional OIIO name of 'Make' should be written to exr files with the name 'cameraMake', and when reading exr, rename 'cameraMake' to 'Make' so that OIIO applications will know to look for it in the OIIO-standard place."

Is this making things a little more clear now?

So here's what we really need to do:

* If there are "EXR standard" metadata names that correspond to OIIO canonical metadata and those names are different, there needs to be entries in the tables of exrinput.cpp, exrinput_c.cpp, and exroutput.cpp. We should definitely add any that are missing (lots have been added to the exr docs recently, so it's easy for me to believe that this needs some adjustment on the OIIO side).
* If we think that some of the EXR standard metadata names (that are not currently documented by OIIO) should be elevated to be "OIIO standard metadata", we should add them to the OIIO documentation in src/doc/stdmetadata.rst. If we use the EXR/SMPTE names for these "new" metadata, then they don't need to be in the exr tables, since no name translation is necessary. But we may need to do translation in other format readers/writers and to make sure it gets saved correctly if in Exif or IPTC if it seems to correspond to things that those give specific names to!
* Everything else can just stay as-is, because the exr writer will already save all named metdata with its original name by default.

src/doc/stdmetadata.rst
```diff
@@ -284,14 +284,94 @@ The following metadata items are specific to photos or captured images.
 
     For captured or scanned image, the model of the camera or scanner.
 
+.. option:: "Focus" : float
+
+    The camera's focus distance of the captured image.
+
```
lgritz, line 290: I really believe that before we add to this list of "standard OIIO metadata", we should have an open discussion about what to add, and why. And I'd rather not do it item by item, but instead, propose a general rule or philosophy about the naming conventions. For example, maybe we should just decide that all of the SMPTE standard names be adopted. Or maybe we should vow to always follow OpenEXR's names and set of standard attributes. Or maybe something else. And regardless, we need to think about exactly how, or if, we should transition from any of the contradictory names we've been using and how to handle it going forward.

You've waded into a longstanding open question about policy and design philosophy, and I don't think it can be solved by a PR that makes some specific changes, not others, and doesn't explain itself at all as to what design its trying to implement. Start with that -- a proposal for how we should approach the metadata problem, and then the stakeholders can collectively decide what we want to do about it, and after that, an implementation that embodies the decision should be straightforward.

src/doc/stdmetadata.rst
```diff
@@ -284,14 +284,94 @@ The following metadata items are specific to photos or captured images.
...
     The f/stop of the camera when it captured the image.
 
+.. option:: "openexr:CameraTemperature" : float
+
+    Color temperature in Kelvin of the camera when it captured the image.
+
+.. option:: "openexr:CameraTint" : float
+
+    Green/magenta tint of the camera when it captured the image.
+
```
lgritz, line 314: Why are these, alone, prefixed as "openexr:"?

Usually, we use the format name as a prefix for one of three reasons:

* Things that are "pseudo-metadata" -- they look like metadata, but actually they are *controls* or *hints* for the writer, information about how the file is written. An example of that is "openexr:roundingmode" that controls how MIP map resolution levels are computed, but isn't really considered the usual token/value style of arbitrary metadata. 
* Items that are so incredibly specific to one format that it makes no sense to try to allow the metadata to copy from one file type to another. Each writer tends to completely ignore metadata that is prefixed by the name of any other format, so that those particular items are never propagated when converting file formats.
* Items where the format-specific way to handle that item is radically different than all the other formats, and we just don't want them mixing.

src/openexr.imageio/exrinput.cpp
```diff
@@ -113,9 +113,31 @@ static std::map<std::string, std::string> exr_tag_to_oiio_std {
...
+    { "aperture", "FNumber" },
+    { "isoSpeed", "Iso", },
+    { "focus", "Focus", },
+    { "cameraCCTSetting", "openexr:CameraTemperature", },
+    { "cameraTintSetting", "openexr:CameraTint", },
+    { "nominalFocalLength", "NominalFocalLength", },
+    { "pinholeFocalLength", "PinholeFocalLength", },
+    { "effectiveFocalLength", "EffectiveFocalLength", },
+    { "entrancePupilOffset", "EntrancePupilOffset", },
+    { "sensorPhotositePitch", "SensorPhotositePitch", },
```
lgritz, line 125: Why are these, and many like it, in the table at all?

What is the point in declaring a new OIIO standard metadata called "effectiveFocalLength" that is automatically translated to an OpenEXR attribute called "EffectiveFocalLength"?

Remember what I explained earlier:

The table is for items that are "standard" in the sense that OIIO-using applications already have a name for them that is intended to be the same across all image file format types, *AND* OpenEXR conventions expect the very same item to have a different name. We resolve this discrepancy by automatically translating the names between the OIIO conventional name and the OpenEXR conventional name, using these tables.

* Any item that isn't an "OIIO standard metadata name" should not be in the table.
* Any item that is an OIIO standard metadata name, and the EXR name is the same spelling, also doesn't need to be in the table.

Remember that OpenEXR already has the ability to store arbitrary name+value metadata without anything special being done, and without needing to explicitly list the set of names. None of this is necessary for OIIO simply to read and write these metadata in OpenEXR files. It already does that. If you tell OIIO (today, before your patch) to write a metadata called "EffectiveFocalLength" to an OpenEXR file, it will do that with no problem whatsoever.

src/doc/stdmetadata.rst
TodicaIonut: done.

src/openexr.imageio/exrinput.cpp
TodicaIonut: done.

src/doc/stdmetadata.rst
JGoldstone: Adopting metadata is essentially buying in to a model of the camera, and of image processing, that the metadata describes. To take the example of camera correlated color temperature (CCT) and tint, please see the discussion of how cameraCCTSetting, cameraTintSetting and cameraColorBalance relate to each other in ImfStandardAttributes.h, and the distinction between forensic and prescriptive metadata.

As the author of the new definitions that were added in 2023 I am keenly interested in seeing these used, but used effectively, and that means doing the philosophical review Larry mentions. I am biased, but I think OpenEXR has done the best job so far in precisely defining metadata semantics, types, and value constraints; but to reconcile their use with OpenImageIO's pan-image-format model, we need to not only look at other formats' definitions, but also to the elephant in the room: Nuke 15's really awful switch to strings for numeric attributes, carrying both the numeric value but also, appended as characters, the units. Eventually I hope to bring them back into the fold, but for those files they've already generated, what should we do? Reject entirely? Convert to canonical form? And what about an image that is read, flopped (say), and rewritten? Do we write the mangled values or canonicalized ones?

This thread feels somehow frantically urgent to the PR submitter; or is that just my take on it?

lgritz: An additional unsettled item for me is what the best thing to do when we see that two widely adopted metadata conventions have different names for things. For example, when reading from formats that support Exif, we name the metadata "Exif:NAME" where the NAME is taken straight from the word the Exif spec uses to describe the item. What do we do when a particular piece of Exif data seems to *exactly* correspond to an item described by OpenEXR/SMPTE (which at the very least won't have the "Exif:" namespace prefix, and may additionally not have the same name stem).

src/doc/stdmetadata.rst
```diff
@@ -288,10 +288,98 @@ The following metadata items are specific to photos or captured images.
...
+GPS
+===
+
+The following GPS-related metadata items are specific.
+
+.. option:: "Longitude" : float
+
+    The degrees, minutes, and seconds of longitude.
+
```
lgritz, line 373: No, no, there already is a "GPS Exif data" section in this document. It already describes longitude, latitude, and altitude. Unfortunately, they have different names -- AND TYPES! We should definitely not have two different "OIIO standard" names for the same thing.

It's totally valid to propose "let's switch to the OpenEXR convention and auto-translate when we read and write Exif", but you have never proposed that, nor it is not even what this PR is implementing.

src/openexr.imageio/exrinput.cpp
```diff
@@ -113,9 +113,31 @@ static std::map<std::string, std::string> exr_tag_to_oiio_std {
...
+    { "cameraFirmwareVersion", "FirmwareVersion", },
+    { "cameraUuid", "Uuid", },
+    { "cameraLabel", "Label", },
+    { "lensMake", "LensMake", },
+    { "lensModel", "LensModel", },
+    { "lensSerialNumber", "LensSerialNumber", },
+    { "lensFirmwareVersion", "LensFirmwareVersion", },
+    { "longitude", "Longitude", },
+    { "latitude", "Latitude", },
+    { "altitude", "Altitude", },
```
lgritz, line 138: What exactly are you trying to do here? What is the reason for making a new standard OIIO name "LensModel" but then saving it in EXR files as "lensModel"? 

If you want to propose adding a new "standard OIIO metadata" that corresponds to the OpenEXR one, why not just use the same name as OpenEXR and not have to do any name translation at all?

I'll say it again:

> This table is for translating OpenEXR standard metadata names to and from OTHER NAMES that we chose for OIIO to use across all formats, and which therefore we cannot easily change.

For any NEW names that we wish OIIO to use across all formats, we can pick any name we want, so why wouldn't we pick the name OpenEXR is already using?

lgritz:
Please stop making small changes to this PR.

Let's step back, take a deep breath, and start with a (non-code) discussion about what you are trying to accomplish here.

If we're going to change any of the naming of OIIO standard metadata names, let's do it for real, starting with a principled proposal for how we want to move forward with the various competing standards and potentially doing full overhaul. The fact that late this summer we are bumping to "OIIO 3.0" implies that we can make the kind of compatibility breaks that we ordinarily do not allow except for once every 5+ years. This is the ideal time to change or fix our metadata conventions, but we're not going to do it without a real discussion and consensus on the way forward.

src/openexr.imageio/exrinput.cpp
TodicaIonut: done.

lgritz:
This PR is not ready to review because:

* It contains NO discussion of what the goal is, so it's hard to evaluate whether the code achieves it. What is the problem that this tries to solve? How does the code meet that goal?
* It seems to be implementing a policy change about how we should handle standard metadata names, but we have not received any proposal for such a new metadata strategy that can be used to achieve consensus of the project leadership.
* Parts of the PR are inexplicable and without any apparent justification, despite multiple review comments, revisions, and many pleas for more information. Why make entirely new standard metadata that are named or capitalized differently than the way we want them stored in OpenEXR files?

PR #4209, opened 2024-03-29
build: have checked_find_package accept CONFIG, fix DCMTK

lgritz:
Add another optional flag to our checked_find_package cmake function -- CONFIG means only accept a config for the package, never use a FindBlah.cmake module.

(Reminder: this is in contrast to the existing PREFER_CONFIG option, which tries a config first, then if that fails, will try the module, and also different from the default if neither is used, which is to prefer the module over the exported config.)

And rig our search for DCMTK to use the new CONFIG option. There is a FindDCMTK.cmake modules that comes with cmake, and for at least some versions of cmake, it seems subtly broken in such a way that actually wrecks our finding of other packages properly.

### DIFF
--- a/src/cmake/checked_find_package.cmake
+++ b/src/cmake/checked_find_package.cmake
@@ -64,6 +64,8 @@ endfunction ()
 #     version, accepting but warning if it is below this number (even
 #     if above the true minimum version accepted). The warning message
 #     can give an optional explanation, passed as RECOMMEND_MIN_REASON.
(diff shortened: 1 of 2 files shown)


lgritz:
Merging without review because
- I need this at work
- I don't think it can break anything that doesn't use the new CONFIG token I added

If this turns out to cause problems for anybody, pipe up and I'll get a fix right away.

PR #4214, opened 2024-04-02
ci: Fix broken Windows CI by building our own libtiff

lgritz:
Windows CI was failing because it couldn't get a proper libtiff from VcPkg, which in turn is because lzma is a dependency of the libtiff build in VcPkg, and for the time being, it looks like VcPgk is failing to get lzma which is part of the big xz debacle of 2024.

So work around for now by just building our own libtiff rather than relying on VcPkg. Seems to work fine.


lgritz:
Pre-emptively merging because it demonstrably fixes the broken CI.

PR #4215, opened 2024-04-03
Fix out-of-bounds reads when using OpenEXR decreasingY lineOrder.

acolwell:
## Description

This change fixes out-of-bounds reads and incorrect scanline ordering when OpenEXR's decreasingY lineOrder mode is used.

OpenEXR expects to process scanlines in decreasing order when the decreasingY lineOrder option is enabled. This change fixes the code that provides the chunks of scanlines to OpenEXR so that the proper scanlines are available when it accesses the FrameBuffer OIIO created. The old code was providing incorrect scanlines AND setting up the FrameBuffer with incorrect pointers so out-of-bounds reads were occurring.

The key things to keep in mind are:
- Chunks of scanlines need to be sent to OpenEXROutput::write_scanlines() from the bottom of the image to the top when decreasingY is enabled. If the chunk's scanline range does not contain the scanlines OpenEXR is expecting to fetch, then it can cause out-of-bound reads because the wrong information is used to construct the pointers for the Slices created when constructing a FrameBuffer.
- OpenEXROutput::write_scanlines() does its own chunking of scanlines and this logic must also properly start from the bottom of the chunk data passed into this function when decreasingY is enabled.
- The to_native_rectangle() call in OpenEXROutput::write_scanlines() may return a pointer to m_scratch if a format conversion is needed. This means the pointers passed to OpenEXR, in this situation, _WILL NOT_ be a full frame buffer. This caused OpenEXR to do out-of-bounds reads because the computations used to create the Slices were using the wrong scanline number to adjust the pointer.
- The progress callback computation needed to be modified to handle traversing scanlines in decreasing order.

## Tests

I added an openexr-decreasingY test to verify that the code no longer crashes because of the out-of-bound reads and it also verifies that the decreasingY images match the increasingY images.

## Checklist:

  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -1419,19 +1419,22 @@ ImageBuf::write(ImageOutput* out, ProgressCallback progress_callback,
                               1024);
             std::unique_ptr<char[]> tmp(new char[chunk * slsize]);
 
+            // Special handling for flipped vertical scanline order. Right now, OpenEXR
+            // is the only format that allows it, so we special case it by name. For
+            // just one format, trying to be more general just seems even more awkward.
             const bool isDecreasingY = !strcmp(out->format_name(), "openexr")
                                        && outspec.get_string_attribute(
                                               "openexr:lineOrder")
                                               == "decreasingY";
-            const int yStart = isDecreasingY
-                                   ? ((outspec.height - 1) / chunk) * chunk
-                                   : 0;
-            const int yDelta = isDecreasingY ? -chunk : chunk;
+            const int numChunks  = outspec.height > 0
+                                       ? 1 + ((outspec.height - 1) / chunk)
+                                       : 0;
+            const int yLoopStart = isDecreasingY ? (numChunks - 1) * chunk : 0;
+            const int yDelta     = isDecreasingY ? -chunk : chunk;
+            const int yLoopEnd   = yLoopStart + numChunks * yDelta;
+
             for (int z = 0; z < outspec.depth; ++z) {
-                for (int y = yStart; ((isDecreasingY && y >= 0)
-                                      || (!isDecreasingY && y < outspec.height))
-                                     && ok;
-                     y += yDelta) {
+                for (int y = yLoopStart; y != yLoopEnd && ok; y += yDelta) {
                     int yend = std::min(y + outspec.y + chunk,
                                         outspec.y + outspec.height);
                     ok &= get_pixels(ROI(outspec.x, outspec.x + outspec.width,
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -1448,7 +1448,8 @@ ImageBuf::write(ImageOutput* out, ProgressCallback progress_callback,
                         && progress_callback(
                             progress_callback_data,
                             (float)(z * outspec.height
-                                    + (isDecreasingY ? (outspec.height - y) : y))
+                                    + (isDecreasingY ? (outspec.height - 1 - y)
+                                                     : y))
                                 / (outspec.height * outspec.depth)))
                         return ok;
                 }
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -1418,8 +1418,20 @@ ImageBuf::write(ImageOutput* out, ProgressCallback progress_callback,
             int chunk = clamp(round_to_multiple(int(budget / slsize), 64), 1,
                               1024);
             std::unique_ptr<char[]> tmp(new char[chunk * slsize]);
+
+            const bool isDecreasingY = !strcmp(out->format_name(), "openexr")
+                                       && outspec.get_string_attribute(
+                                              "openexr:lineOrder")
+                                              == "decreasingY";
+            const int yStart = isDecreasingY
+                                   ? ((outspec.height - 1) / chunk) * chunk
+                                   : 0;
+            const int yDelta = isDecreasingY ? -chunk : chunk;
             for (int z = 0; z < outspec.depth; ++z) {
-                for (int y = 0; y < outspec.height && ok; y += chunk) {
+                for (int y = yStart; ((isDecreasingY && y >= 0)
+                                      || (!isDecreasingY && y < outspec.height))
+                                     && ok;
+                     y += yDelta) {
                     int yend = std::min(y + outspec.y + chunk,
                                         outspec.y + outspec.height);
                     ok &= get_pixels(ROI(outspec.x, outspec.x + outspec.width,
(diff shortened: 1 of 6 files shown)


src/libOpenImageIO/imagebuf.cpp
```diff
@@ -1418,8 +1418,20 @@ ImageBuf::write(ImageOutput* out, ProgressCallback progress_callback,
             int chunk = clamp(round_to_multiple(int(budget / slsize), 64), 1,
                               1024);
             std::unique_ptr<char[]> tmp(new char[chunk * slsize]);
+
+            const bool isDecreasingY = !strcmp(out->format_name(), "openexr")
```
acolwell, line 1422: I fully admit this is not great. I am totally open to suggestions on the preferred way to hide this behind the ImageOutput abstraction. I wasn't sure if a new virtual method was in order or if this should be a new "feature name" passed to supports().

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -1418,8 +1418,20 @@ ImageBuf::write(ImageOutput* out, ProgressCallback progress_callback,
...
+            const int yStart = isDecreasingY
+                                   ? ((outspec.height - 1) / chunk) * chunk
+                                   : 0;
+            const int yDelta = isDecreasingY ? -chunk : chunk;
             for (int z = 0; z < outspec.depth; ++z) {
-                for (int y = 0; y < outspec.height && ok; y += chunk) {
+                for (int y = yStart; ((isDecreasingY && y >= 0)
+                                      || (!isDecreasingY && y < outspec.height))
+                                     && ok;
+                     y += yDelta) {
```
lgritz, line 1434: This is a little awkward. Why not 

```
            const int yStart = isDecreasingY
                                   ? ((outspec.height - 1) / chunk) * chunk
                                   : 0;
            const int yDelta = isDecreasingY ? -chunk : chunk;
            const int yEnd = yStart + yDelta;
            for (int z = 0; z < outspec.depth; ++z) {
                for (int y = yStart; y != yEnd && ok; y += yDelta) {
```

src/libOpenImageIO/imageoutput.cpp
```diff
@@ -518,17 +518,31 @@ ImageOutput::write_image(TypeDesc format, const void* data, stride_t xstride,
         int rps   = m_spec.get_int_attribute("tiff:RowsPerStrip", 64);
         int chunk = std::max(1, (1 << 26) / int(m_spec.scanline_bytes(true)));
         chunk     = round_to_multiple(chunk, rps);
+
+        const bool isDecreasingY = !strcmp(format_name(), "openexr")
```
lgritz, line 522: I think this is fine, but why let's put a comment here that says something like

```
// Special handling for flipped vertical scanline order. Right now, OpenEXR
// is the only format that allows it, so we special case it by name. For
// just one format, trying to be more general just seems even more awkward.
```

src/libOpenImageIO/imageoutput.cpp
```diff
@@ -518,17 +518,31 @@ ImageOutput::write_image(TypeDesc format, const void* data, stride_t xstride,
...
+                                          == "decreasingY";
+        const int yStart = isDecreasingY ? ((m_spec.height - 1) / chunk) * chunk
+                                         : 0;
+        const int yDelta = isDecreasingY ? -chunk : chunk;
+
         for (int z = 0; z < m_spec.depth; ++z)
-            for (int y = 0; y < m_spec.height && ok; y += chunk) {
+            for (int y = yStart; ((isDecreasingY && y >= 0)
+                                  || (!isDecreasingY && y < m_spec.height))
+                                 && ok;
```
lgritz, line 533: same comment as above regarding `...; y != yEnd; ...` would be simpler, with `yEnd = yStart + yDelta`

src/openexr.imageio/exroutput.cpp
```diff
@@ -1485,21 +1485,34 @@ OpenEXROutput::write_scanlines(int ybegin, int yend, int z, TypeDesc format,
...
+    const bool isDecreasingY = m_spec.get_string_attribute("openexr:lineOrder")
+                               == "decreasingY";
+
+    int yChunkStart = isDecreasingY ? ybegin + ((yend - ybegin) / chunk) * chunk
+                                    : ybegin;
+    const int yDelta = isDecreasingY ? -chunk : chunk;
+    for (; ok
+           && ((isDecreasingY && yChunkStart >= ybegin)
+               || (!isDecreasingY && yChunkStart < yend));
+         yChunkStart += yDelta) {
```
lgritz, line 1498: and here

lgritz (changes requested):
LGTM pending the few comments I had about how to make the `for` loop iteration maybe a little less confusing. Other than that, looks good. Thanks for adding a careful test as well.

src/libOpenImageIO/imagebuf.cpp
acolwell: Agreed. I updated the code to something similar yEnd = yStart + numChunks * yDelta which is I believe what you were getting at.

I also change yStart and yEnd to yLoopStart and yLoopEnd because there were already yend variables near this code and having folks try to keep the difference between yend and yEnd straight just seems like asking for bugs.

src/libOpenImageIO/imageoutput.cpp
acolwell: Done

src/libOpenImageIO/imageoutput.cpp
acolwell: added similar fix here.

src/openexr.imageio/exroutput.cpp
acolwell: ditto

acolwell (commented):
Addressed all comments. Thanks for the review.

src/libOpenImageIO/imagebuf.cpp
lgritz: Yeah, sorry, I think I was on an airplane when I wrote that, don't take what I say tooooo literally.

src/libOpenImageIO/imageoutput.cpp
```diff
@@ -518,17 +518,34 @@ ImageOutput::write_image(TypeDesc format, const void* data, stride_t xstride,
...
                 ok &= write_scanlines(y + m_spec.y, yend, z + m_spec.z, format,
                                       d, xstride, ystride);
                 if (progress_callback
-                    && progress_callback(progress_callback_data,
-                                         (float)(z * m_spec.height + y)
-                                             / (m_spec.height * m_spec.depth)))
+                    && progress_callback(
+                        progress_callback_data,
+                        (float)(z * m_spec.height
+                                + (isDecreasingY ? (m_spec.height - y) : y))
```
lgritz, line 547: should this be `m_spec.height - 1 - y` ?

acolwell: Yep. Sorry about that. Fixed here and in imagebuf.cpp.

acolwell:
I'm not really sure what to do about the 1 failed builder. It doesn't appear to be related to my change. Unfortunately, I'm not familiar with fmt or the related cmake magic that appears to be including it so I don't have a suggested fix.

lgritz:
That's not related to this PR. There is one build we do that uses the current top-of-tree of several major dependencies. It often fails because one of those dependencies has checked a bug into their own tree.

lgritz (approved):
LGTM, thanks for the patch!

PR #4216, opened 2024-04-05
R3D

1div0:
This is an initial code for the RED Digital Cinema support.

CPU only so far.


src/iv/imageviewer.cpp
```diff
@@ -64,7 +64,7 @@ IsSpecSrgb(const ImageSpec& spec)
 static const char *s_file_filters = ""
     "Image Files (*.bmp *.cin *.dcm *.dds *.dpx *.fits *.gif *.hdr *.ico *.iff "
     "*.jpg *.jpe *.jpeg *.jif *.jfif *.jfi *.jp2 *.j2k *.jxl *.exr *.png *.pbm *.pgm "
-    "*.ppm *.psd *.ptex *.rla *.sgi *.rgb *.rgba *.bw *.int *.inta *.pic *.tga "
+    "*.ppm *.psd *.ptex *.R3D *.r3d *.rla *.sgi *.rgb *.rgba *.bw *.int *.inta *.pic *.tga "
```
lgritz, line 67: I don't think you need the capitals -- I believe it's a case-insensitive compare? Otherwise, all of the above would need them.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+#include <list>
+#include <map>
+#include <mutex>
+#include <thread>
+#include <time.h>
+#include <unistd.h>
+#include <vector>
+
+unsigned char*
+AlignedMalloc(size_t& sizeNeeded)
```
lgritz, line 43: I believe we already have an aligned_malloc and aligned_new in platform.h

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+// Export version number and create function symbols
+OIIO_PLUGIN_EXPORTS_BEGIN
+
+OIIO_EXPORT int r3d_imageio_version = OIIO_PLUGIN_VERSION;
+
+OIIO_EXPORT const char*
+r3d_imageio_library_version()
+{
+    return "R3D";
```
lgritz, line 154: This should say a version number, if it's at all possible to know.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+void
+R3dInput::initialize()
+{
+    DBG std::cout << "R3dInput::initialize()\n";
+
+    // initialize SDK
+    // R3DSDK::InitializeStatus init_status = R3DSDK::InitializeSdk(".", OPTION_RED_CUDA);
+    R3DSDK::InitializeStatus init_status
+        = R3DSDK::InitializeSdk("/opt/R3DSDKv8_5_1/Redistributable/linux",
```
lgritz, line 177: Should we be worried that this is an absolute path? It is really true that this is the only place, on any system, where it could be installed? What about when not on Linux? Does the SDK exist for other platforms?

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+    r = (uint16_t*)((uint8_t*)m_image_buffer + y * scanline_size);
+    g = (uint16_t*)((uint8_t*)m_image_buffer + y * scanline_size + plane_size);
+    b = (uint16_t*)((uint8_t*)m_image_buffer + y * scanline_size
+                    + 2 * plane_size);
+
+    for (int x = 0; x < m_spec.width; x++) {
+        *d++ = r[x];
+        *d++ = g[x];
+        *d++ = b[x];
+    }
```
lgritz, line 380: I think you can use OIIO::copy_image for this. It's declared in imageio.h, and in tiffinput.cpp you can find examples of how it's used to copy "planar" to "contiguous" striding.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+        = R3DSDK::InitializeSdk("/opt/R3DSDKv8_5_1/Redistributable/linux",
+                                OPTION_RED_NONE);
+    if (init_status != R3DSDK::ISInitializeOK) {
+        R3DSDK::FinalizeSdk();
+        DBG std::cout << "Failed to load R3DSDK Lib: " << init_status << "\n";
+        return;
+    }
+
+    DBG std::cout << "SDK VERSION: " << R3DSDK::GetSdkVersion() << "\n";
+#ifdef GPU
```
lgritz, line 186: Where is GPU defined? Or is this just a placeholder for later?

Let's definitely get the non-GPU one finalized and merged first, then move on to GPU.

I have some long-incubating Cuda code that I've been sitting on for a while, that I should revive and maybe it's time to have some general Cuda setup utilities for the various parts of OIIO that might be inclined to want/need them.

lgritz (commented):
This is looking good. I have no complaints other that pointing out a few places that you can simplify by using utilities we already have.

src/iv/imageviewer.cpp
1div0: I added capitals after failure in test run of the image viewer. Will dig deeper and check.

src/r3d.imageio/r3dinput.cpp
1div0: I have observed this output during run, so it can be obtained and used during build phase as well.

R3dInput::initialize()
SDK VERSION: R3DAPI 8.5.1-53601 (20240308 Lx64S) R3DSDK 8.5.1-53601 (20240308 Lx64D C3B1) [/opt/R3DSDKv8_5_1/Redistributable/linux/]

src/r3d.imageio/r3dinput.cpp
1div0: peter.kovar@Pascal /opt/R3DSDKv8_5_1/Redistributable
€ LANG=en_US ls -lR
.:
total 4
drwxrwxr-x 2 root root  119 Mar  8 15:15 linux
drwxrwxr-x 2 root root   95 Mar  8 15:15 mac
drwxrwxr-x 2 root root 4096 Mar  8 15:15 win

./linux:
total 43300
-rwxr-xr-x 1 root root 11970637 Mar  8 15:01 REDCuda-x64.so
-rwxr-xr-x 1 root root 16306349 Mar  8 15:01 REDDecoder-x64.so
-rwxr-xr-x 1 root root  6068308 Mar  8 14:53 REDOpenCL-x64.so
-rwxrwxr-x 1 root root  4963603 Mar  8 14:01 REDR3D-x64.so
-rwxrwxr-x 1 root root  5008183 Mar  8 14:01 REDR3D-x86.so

./mac:
total 32552
-rwxr-xr-x 1 root root 11544448 Mar  8 14:13 REDDecoder.dylib
-rwxr-xr-x 1 root root  3459216 Mar  8 14:13 REDMetal.dylib
-rwxr-xr-x 1 root root 11112384 Mar  8 14:13 REDOpenCL.dylib
-rwxr-xr-x 1 root root  7198960 Mar  8 14:01 REDR3D.dylib

./win:
total 55052
-rw-r--r-- 1 root root 12414976 Mar  8 15:14 REDCuda-x64.dll
-rw-r--r-- 1 root root  6083584 Mar  8 14:27 REDCuda-x86.dll
-rw-r--r-- 1 root root 15573504 Mar  8 15:14 REDDecoder-x64.dll
-rw-r--r-- 1 root root  4557312 Mar  8 14:30 REDDecoder-x86.dll
-rw-r--r-- 1 root root  5178880 Mar  8 14:35 REDOpenCL-x64.dll
-rw-r--r-- 1 root root  4408320 Mar  8 14:30 REDOpenCL-x86.dll
-rw-r--r-- 1 root root  4258816 Mar  8 14:04 REDR3D-x64.dll
-rw-r--r-- 1 root root  3871744 Mar  8 14:03 REDR3D-x86.dll

1div0: So in total 5 variants, but maybe they will add macOS AArch64 variant too in not so distant future. I wanted to have something working on Linux x86-64 at first.

src/r3d.imageio/r3dinput.cpp
1div0: Good 2 know. Will modify the code accordingly.

src/r3d.imageio/r3dinput.cpp
1div0: It is a leftover from the previous standalone proof of concept code.

src/r3d.imageio/r3dinput.cpp
1div0: Thanks. Will try and use that.

src/iv/imageviewer.cpp
lgritz: https://doc.qt.io/qt-6/qfiledialog.html#getSaveFileName doesn't say one way or the other if the filters file file extensions are case sensitive or not, but the examples in the docs are all lower case only. I would assume that if they were case sensitive, their examples would include the upper case versions, too, or they would get endless complaints from Windows users?

src/r3d.imageio/r3dinput.cpp
lgritz: I think we can certainly figure out programamtically whether to end with linux, mac, win. I'm more concerned about the `/opt/R3DSDKv8_5_1/Redistributable/` part -- not only is the version baked in, but /opt, really, is that the one and only place it could ever be?

This may be ok as a first stab at it, but I wonder if maybe what we eventually want is some kind of search procedure, like we have for finding fonts in imagebufalgo_draw.cpp, where the location could be different for different systems, so we check several likely places, and also have an environment variable that can be used to explicitly give a nonstandard location?

src/iv/imageviewer.cpp
lgritz: OK, leave both capitalizations, then.

src/r3d.imageio/r3dinput.cpp
1div0: What is the best practice in this domain? Sync with R3D so we will not reinvent the wheel.

I would propose following search order:

1. Check of the environment variable and insert its value in the search path.
2. Check and read the runtime configuration file in user home directory.
3. Check the system wide configuration file common to all users.
4. At last try the value obtained at build time.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+    // calculate how many bytes we need
+    sizeNeeded = 16U - (ptr % 16U);
+
+    return buffer + sizeNeeded;
+}
+
+OIIO_PLUGIN_NAMESPACE_BEGIN
+
+#define DBG if (1)
```
lgritz, line 68: I think we want to turn debug messages off before merging this.

1div0: ACK

1div0: C001

1div0: Will try it with the environment variable controling the debug level verbosity.

lgritz (changes requested):
Some very minor comments. Please fix the clang-format so we don't fail CI. The rest are mostly requests for clarification, but nothing we can't fix in subsequent PRs that improve the RED support.

I'm taking for granted that this works -- I don't have the SDK installed and don't have any RED frames to test. But I assume you do on your end.

I wonder also if you could, either as part of this PR or in the near future, add a section in the builtinplugins.rst for RED? The most important part is probably just clarifying there what the RED reading currently can and cannot do, what limitations users should be aware of in its current state, etc.

src/cmake/modules/FindR3DSDK.cmake
```diff
@@ -46,23 +53,36 @@ else()
...
 
 set(R3DSDK_INCLUDE_DIR ${R3DSDK_ROOT}/Include)
-find_library(R3DSDK_LIBRARY NAMES ${R3DSDK_LIB_NAME}
-             PATHS ${R3DSDK_ROOT}/Lib/linux64/
-                   ${R3DSDK_ROOT}/Lib/win64/
-             NO_DEFAULT_PATH
-            )
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    find_library(R3DSDK_LIBRARY NAMES ${R3DSDK_LIB_NAME}
+            PATHS ${R3DSDK_ROOT}/Lib/linux64/
```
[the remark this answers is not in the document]
1div0: Yes, really.

€ LANG=en_US.UTF-8 ls -l /opt/R3DSDKv8_5_1/Lib/
total 8
drwxrwxr-x 2 root root  100 Mar  8 14:04 linux32
drwxrwxr-x 2 root root  100 Mar  8 14:04 linux64
drwxrwxr-x 2 root root   32 Mar  8 14:04 mac64
drwxrwxr-x 2 root root 4096 Mar  8 14:04 win32
drwxrwxr-x 2 root root 4096 Mar  8 14:04 win64

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,473 @@
...
+#include <unistd.h>
+#include <vector>
+
+OIIO_PLUGIN_NAMESPACE_BEGIN
+
+#if 0 || !defined(NDEBUG) /* allow color configuration debugging */
+static bool colordebug = Strutil::stoi(Sysutil::getenv("OIIO_COLOR_DEBUG"));
+#    define DBG(...)    \
+        if (colordebug) \
+        Strutil::print(__VA_ARGS__)
```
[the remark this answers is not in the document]
1div0: Thanks for notice. Will fix that in the next round of changes.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,473 @@
...
+    m_job.Mode = R3DSDK::DECODE_FULL_RES_PREMIUM;
+
+    // store the image here
+    m_job.OutputBuffer = m_image_buffer;
+
+    // store the image in a 16-bit planar RGB format
+#ifdef PLANAR
+    m_job.PixelType   = R3DSDK::PixelType_16Bit_RGB_Planar;
+    m_job.BytesPerRow = width * sizeof(uint16_t);
+#else
```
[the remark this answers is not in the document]
1div0: I intend to let the planar go away. It is left for performance comparison only.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+
+
+bool
+R3dInput::seek(int frame)
+{
+    DBG("R3dInput::seek({})\n", frame);
+    return true;
+}
```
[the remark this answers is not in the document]
1div0: Again, something to do next week.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,473 @@
...
+
+
+
+int64_t
+R3dInput::time_stamp(int frame) const
+{
+    DBG("R3dInput::time_stamp({})\n", frame);
+    return 0;
```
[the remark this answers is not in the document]
1div0: Yes, will implement that ASAP.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -148,7 +148,8 @@ OIIO_EXPORT int r3d_imageio_version = OIIO_PLUGIN_VERSION;
 OIIO_EXPORT const char*
 r3d_imageio_library_version()
 {
-    return "R3D";
+    // Note: SDK version can differ from the actual library loaded
+    return "R3D 8.5.1";
```
lgritz, line 152: Can you get this from the library or the r3dsdk headers? This will obviously be wrong as soon as a new release is out.

src/r3d.imageio/r3dinput.cpp
```diff
@@ -0,0 +1,429 @@
...
+#include <thread>
+#include <time.h>
+#include <unistd.h>
+#include <vector>
+
+OIIO_PLUGIN_NAMESPACE_BEGIN
+
+#if 0 || !defined(NDEBUG) /* allow R3D configuration debugging */
+static bool r3d_debug = Strutil::stoi(Sysutil::getenv("OIIO_R3D_DEBUG"));
+#    define DBG(...)    \
```
lgritz, line 63:
```suggestion
#    define DBG(...)   \
```

1div0: Just modified. Wating for the build results adn leaving for today.

src/r3d.imageio/r3dinput.cpp
1div0: I can get this from the CMake script. Indeed, hardcoding the version string is not proper way to handle this. The actual library version may be different at runtime, as I've tested so far. But nothing exploded. This version is reported just after initialization from R3DSDK::GetSdkVersion().

lgritz (approved):
LGTM. I'm satisfied enough to merge this to checkpoint it, and you can base the next set of changes on top of this.

PR #4217, opened 2024-04-05
build!(deps): Raise minimum LibRaw to 0.20 (from 0.18)

lgritz:
LibRaw 0.20 dates from mid-2020 and in any case, this version or newer is necesary for use with C++17 which is our new minimum.

Bump the version built by build_scripts/build_LibRaw.bash to the latest.

Remove some old libraw testsuite output references that were only applicable to the older versions.

This change will not be backported to release branches, the bump will only be for OIIO 3.0 and later.


lgritz:
Comments or concerns?

lgritz:
Merging since there are no objections, and in any case, older versions of LibRaw will no longer work now that C++17 is our minimum.

PR #4218, opened 2024-04-07
build(deps):  Raise OpenVDB minimum to 9.0

lgritz:
Soon we'll raise the minimum OpenEXR to 3.1, and OpenVDB 9.0 is the first version to support that. And because OpenVDB has Imath types in its public APIs, it's really hard to mix and match our using an OpenEXR/Imath 3 internal to OIIO with an OpenVDB that uses an OpenEXR/IlmBase 2.


lgritz:
Any objections to this?

lgritz:
Does anybody think there is merit to continuing to support OpenVDB older than 9.0? And, really, we simply cannot, given our decision to bump the minimum OpenEXR version to 3.x, which was not supported until OpenVDB 9. So asking for OpenVDB < 9 is asking to keep supporting OpenEXR 2.4.

I think there is no question that we should merge this, but thought I'd give one last chance...

PR #4221, opened 2024-04-10
Remove boost thread_specific_ptr

fpsunflower:
## Description

Remove the dependency on boost's thread_specific_ptr by using hashmap's of static duration between the owning object and the target datatype.

After digging a bit into how boost's pointer works behind the scenes, it is actually a very similar mechanism, with the lookup being into an std::map. I don't have an easy way to measure performance here, but none of these are on a hot path (with the possible exception of the per_thread_info for the image cache, but this one can be managed externally if you want the best performance.

## Tests

Going to rely on the CI to run the testsuite here.

## Checklist:

  nonconforming.


fpsunflower:
I split this up into three commits to help with the review.

There is a bit of a pattern with the error messages that could maybe be unified - but I opted to stay as close to the original code as possible for now.


I only tested that this works properly on my local machine - so will be monitoring the CI to ensure its working properly everywhere else.

src/libOpenImageIO/imageoutput.cpp
```diff
@@ -261,17 +262,13 @@ ImageOutput::append_error(string_view message) const
 {
     if (message.size() && message.back() == '\n')
         message.remove_suffix(1);
-    std::string* errptr = m_impl->m_errormessage.get();
-    if (!errptr) {
-        errptr = new std::string;
-        m_impl->m_errormessage.reset(errptr);
-    }
+    std::string& err_str = error_messages[this];
```
lgritz, line 265: So there is a per-thread map of ImageOutput -> string. This line adds a slot in this thread's map for this ImageOutput. When the thread terminates, I assume the whole map (for that thread) will delete. But as long as the thread is alive, the strings just sit there even when the ImageOutput they correspond to is long gone, no? Don't we want something in `ImageOutput::Impl::~Impl` that does an `erase` for its slot in the map?

lgritz (commented):
I really like this!

I made a comment about ImageOutput, but I think it applies across the board. Are you not missing a provision in the object destructors to remove that object's doodad from the per-thread map?

src/libOpenImageIO/imageoutput.cpp
fpsunflower: The problem is the destructor only runs on one thread. So it can't reach into other thread's storage without additional bookkeeping.

It probably is solveable, though I worry about introducing complicated threaded bookkeeping just to release the memory a few error messages.

lgritz: Oooooh, yes of course.

OK, so for the error things, it might pseudo-leak a few strings -- and only if there are actual errors. So in practice, NBD.

And for ImageCache, the "object" is the cache itself, of which there is generally exactly one, so we don't really worry about a *succession* of IC's being created from scratch, then being destroyed without freeing their per-thread cruft.

Sounds ok to me, then.

lgritz: Chris, I still think that maybe the object destructors should have a

`if (threadlocal.contains(key)) threadlocal.erase(key);`

It's true that it will only free the thing if the same thread that's destroying the ImageOutput (say) is the same one that called error(). But that's a very common case!  So we could still leak a few strings, but usually it will cause us to fully clean up.

src/libOpenImageIO/imageinput.cpp
```diff
@@ -1120,11 +1117,11 @@ std::string
...
     std::string e;
-    std::string* errptr = m_impl->m_errormessage.get();
-    if (errptr) {
-        e = *errptr;
+    auto iter = input_error_messages.find(this);
+    if (iter != input_error_messages.end()) {
+        e = iter.value();
         if (clear)
-            errptr->clear();
+            iter.value().clear();
```
lgritz, line 1124: If this was `input_error_messages.erase(iter)`, instead of `clear()`, then any time an error was retrieved/clear by the right thread, the entry in the map would fully be cleaned up, instead of "leaking", even though the thing it leaks is usually an empty string.

src/libOpenImageIO/imageinput.cpp
```diff
@@ -1120,11 +1118,13 @@ std::string
...
-    if (errptr) {
-        e = *errptr;
-        if (clear)
-            errptr->clear();
+    auto iter = input_error_messages.find(this);
+    if (iter != input_error_messages.end()) {
+        e = iter.value();
+        if (clear) {
+            iter.value().clear();
+            iter.value().shrink_to_fit();
```
lgritz, line 1126: Why shrink_to_fit instead of erase? You're freeing the characters, but not getting rid of the node in the hash table that contains the string itself. You could imagine a pattern of repeated ImageInput constructions (with errors) that would lead to a fuller and fuller hash table containing unusabl empty strings? Versus actually removing the hash table entry when the ImageInput is destroyed by the right thread (which is mostly).

I don't think it's a problem in practice -- to be a real issue, you'd need tons of created and then destroyed ImageInput objects that encountered errors along the way. I'm happy to make that a "cross that very unlikely bridge when we get to it." But I'm curious if you did this because you spotted some big flaw in my proposal that I hadn't considered.

fpsunflower: Oh I see what you mean now. Yes of course erase makes more sense.

src/libOpenImageIO/imageoutput.cpp
fpsunflower: I can definitely add this to all the destructors. It shouldn't cost much in the common case and keeps the map for the main thread small(er).

I also just realized that technically, these error strings have the same bug I fixed in the image cache. It's much less likely to cause issues, but the bug would be:

1) create an `ImageInput` - gets allocated at address 0x1234 (for example)
2) do some operations (potentially across threads) which lead to errors, but you don't check or clear them
3) close the `ImageInput`
4) create a new `ImageInput` which happens to get allocated at 0x1234 again
5) now all threads still have the error message from the first one which never got cleared :(


The way I solved it for the per thread infos was to make the map use a strictly incrementing counter instead of their address for the lookup (in practice most apps only ever create a single image cache so its not a big diference).  I think I should do the same for the error messages (even though its asymptomatic in the testsuite, the potentialy for confusion if it ever happens would be quite high). The only drawback is an extra 64 bit ID per `ImageInput`,`ImageOutput`.

I'll ponder this a bit more to see if I think of any other clever way to solve this ....

lgritz:
The same thing you fixed with ImageCache and the could happen with the errors, right? Like, you make an ImageInput, it has an error (that the caller never retrieves), destroy it, and then the next ImageInput you make MIGHT have the same address and be born with an error message already there!

It's really a user bug to destroy an object that accumulates errors without ever checking and clearing those errors. But I still wonder if we should clear the error in the destructor, even though that will only work in cases where the errors were encountered by the destroying thread. Or maybe that's worry for nothing.

src/libOpenImageIO/imageoutput.cpp
lgritz: haha, I think our messages crossed, I had typed mine a couple hours ago, but only hit the button now without seeing you commented on the same thing.

lgritz: In short, I don't think it's worth doing anything elaborate. Let's consider it a user error to destroy an ImageInput that you used, accumulated errors, but never retrieved the error messages. But if you make the destructor clear the error for that thread (which will usually be the only thread), I think the stars will rarely align to give a problem, even when the user makes that mistake.

lgritz:
Something is not happy

fpsunflower:
Indeed ... and I can't repro on my mac. It seems to only happen on linux platforms ...

I'll rewind a bit and see if I can get the tests green again.

fpsunflower:
It looks like those erase calls in the destructor were to blame. I am not sure why though.

@lgritz are you able to take this branch for a spin on your work machines to see how it does on the renderer testsuite for example?

Its definitely worth testing this one carefully.

lgritz:
Yeah, I will try it and see what happens.

lgritz (approved):
LGTM!

There is a very subtle issue that Chris and I have been talking about -- the "static destruction order fiasco" keeps biting us when we try to make sure that we delete these error strings. We're going to merge this with some possible string leaking to avoid it, and return to solve the problem in a more robust way.

ThiagoIze:
Something to keep in mind with having a thread_local with a destructor, such as the `thread_local tsl::robin_map` in this PR, is that this destructor won't run until the thread is destroyed. If OIIO is dlclosed before the thread is closed, [this could cause a crash or hang if the OS doesn't handle this specifically](https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables) because the destructor is run at thread termination, after the OIIO code that has been unloaded. We see that it was [fixed in glibc 2.22](https://sourceware.org/glibc/wiki/Release/2.22) for linux. This means centos 7.7 still suffers from that (it's on 2.17) unless they backported that fix (I haven't checked).

From https://github.com/rust-lang/rust/issues/28794#issuecomment-368693049 it looks like maybe this was fixed on macos (dlclose won't unload the dylib if you have thread_local). I haven't checked what the status is on Windows.

PR #4222, opened 2024-04-10
build: Remove boost::stacktrace

lgritz:
To shed the boost dependency, we are disabling stacktrace for now. It's not vital and probably not worth keeping boost just for that.

C++23 has std::stacktrace, so as compilers add support for this, we will phase it back in.

We can also add it back if somebody has an alternate suggestion for a 3rd party solution that is a simply drop-in replacement, portable, lightweight dependency, with compatible license. A good compromise might be this one: https://github.com/jeremy-rifkin/cpptrace But I think what I'm proposing is that we accept this PR first, get rid of boost, then we can come back to looking at cpptrace as potentially a way to restore the functionality prior to C++23.


fpsunflower (approved):
LGTM!

PR #4223, opened 2024-04-10
build!: Move OpenEXR/Imath minimum to 3.1

### LINKED ISSUE #4156: Raise OpenEXR/Imath minimum to 3.1 for OIIO 3.0
A non-comprehensive list of sub-tasks, not necessarily in order:

- [ ] INSTALL.md -- document the new minimums
- [ ] externalpackages.cmake -- enforce the new minimums
- [ ] FindOpenEXR.cmake -- is this no longer needed because we can rely on OpenEXR's and Imath's exported configs?
- [ ] Search for OPENEXR_VERSION through the whole codebase, look for places where we are conditionally selecting code.
- [ ] ci.yml -- remove testing cases of the versions, and/or change those test cases to newer versions.
- [ ] `openexr.imageio/*.{cpp,h}` -- clean up everything where we special-case pre-3.1

We should not merge these fixes into the main branch until after April 1.

lgritz:
These are build-breaking changes that impose a new minimum OpenEXR/Imath version supported of 3.1 or higher.

Needless to say, this will not be backported to any release branches, which guarantee that they will never raise the minimum supported versions of any dependencies. This is just for future OIIO. There is already a "developer preview" branch dev-2.6.1 marking the last spot in master in which the old dependencies are supported.

Fixes #4156

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,18 +34,19 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: gcc9/C++17 py3.7 exr2.5 ocio2.0
+          - desc: gcc9/C++17 py3.7 exr3.1 ocio2.0
             nametag: linux-vfx2021
             runner: ubuntu-latest
             container: aswf/ci-osl:2021
             vfxyear: 2021
             cxx_std: 17
+            openexr_ver: v3.1.13
             python_ver: 3.7
             simd: "avx2,f16c"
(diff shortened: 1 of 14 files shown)


lgritz:
Thoughts? I have a really interesting one that will come after this!

lgritz:
Fixes #4156

jessey-git (approved):
Looks all correct from what I understand about cmake. And CI is all green so it _must_ be fine :)

PR #4226, opened 2024-04-13
fix(TS): fix missing initialization in TextureOptBatch

### LINKED ISSUE #4219: black lanes in batched texture lookup
**Describe the bug**

This is an odd bug. I can't reproduce it reliably, but it occurs quite regularly. When I use OIIO's batched texture lookup, every now and then, all batches of results come back with a specific set of lanes which contain black pixels. The pattern of lanes remains the same through the entire program, but it varies - sometimes it's just one lane, sometimes several.

I see this behaviour both with OIIO 2.4.17 from my package management and from a build from master (2.6.something).

Please run `oiiotool --buildinfo` and paste the output here.

I suppose this will be enough:

```
Input formats supported: bmp, cineon, dds, dicom, dpx, ffmpeg, fits, gif, hdr, heif, ico, iff, jpeg, jpeg2000, null, openexr, openvdb, png, pnm, psd, raw, rla, sgi, softimage, targa, tiff, webp, zfile
Output formats supported: bmp, dpx, fits, gif, hdr, heif, ico, iff, jpeg, jpeg2000, null, openexr, png, pnm, rla, sgi, targa, term, tiff, webp, zfile
OpenColorIO 2.1.3, color config: built-in
Known color spaces: "linear", "default", "rgb", "RGB", "sRGB", "Rec709"
Filters available: box, triangle, gaussian, sharp-gaussian, catmull-rom, blackman-harris, sinc, lanczos3, radial-lanczos3, nuke-lanczos6, mitchell, bspline, disk, cubic, keys, simon, rifman
Dependent libraries: OpenEXR 3.1.5, LIBTIFF Version 4.5.1, jpeg-turbo 2.1.5/jp62, dcmtk 3.6.7, FFMpeg 6.0 (Lavf60.16.100), gif_lib 5.2.2, libheif 1.17.6, OpenJpeg 2.5.0, null 1.0, OpenVDB 10.0.1abi10, libpng 1.6.43, libraw 0.21.2-Release, Webp
    1.3.2
OIIO 2.4.17.0 built for C++17/201703 sse2
Running on 4 cores 15.3GB sse2,sse3,ssse3,sse41,sse42,avx,avx2,fma,f16c,popcnt,rdrand

```

Also please tell us if there was anything unusual about your environment or
nonstandard build options you used.

I'm running this on debian testing, nothing unusual. I also tried on a different machine running ubuntu 22.04 LTS with OIIO 2.2.18, same thing there.

**To Reproduce**

1. compile this program:
```
#include <OpenImageIO/texture.h>
#include <random>

#define LANES 16

using namespace OIIO ;
int main ( int argc , char * argv[] )
{
  float s[LANES] ;
  float t[LANES] ;
  float dsdx[LANES], dtdx[LANES], dsdy[LANES], dtdy[LANES] ;
  float px[3*LANES] ;

  for ( int i = 0 ; i < LANES ; i++ )
    dsdx[i] = dtdx[i] = dsdy[i] = dtdy[i] = 0 ;

  
  std::mt19937 gen(42); // level playing field
  std::uniform_real_distribution<> dis(0.0, 1.0);
  TextureSystem * ts = TextureSystem::create() ;
  ustring ufilename ( argv[1] ) ;
  auto th = ts->get_texture_handle ( ufilename ) ;
  TextureOptBatch batch_options ;
  for ( int i = 0 ; i < LANES ; i++ )
    batch_options.swidth[i] = batch_options.twidth[i] = 0 ;

  for ( int k = 0 ; k < 100000 ; k++ )
  {
    for ( int i = 0 ; i < LANES ; i++ )
    {
      s[i] = dis(gen) ;
      t[i] = dis(gen) ;
    }
    bool result =
    ts->texture ( th , nullptr , batch_options , Tex::RunMaskOn ,
                  s , t ,
                  dsdx, dtdx, dsdy, dtdy ,
                  3 , px ) ;

    assert ( result ) ;
    for ( int i = 0 ; i < LANES ; i++ )
    {
      if ( px[i] == 0.0f && px[LANES+i] == 0.0f && px[2*LANES+i] == 0.0f )
        std::cout << "bingo " << k << " " << i << std::endl ;
    }
  }
  return 0 ;
}

```
let's say the program is called texture_fuzz.cc, compile like this (clang++ does the same):
```
g++ -O3 -std=c++17 -otexture_fuzz texture_fuzz.cc -lOpenImageIO -lOpenImageIO_Util
```
Then run it, passing an image file as the sole argument.

3. Then THIS happens (reproduce the exact error message if you can)
... this is the hard bit. The expected behaviour is something like this:
```
bingo 1063 15
bingo 1174 13
bingo 1982 9
bingo 2080 8
bingo 4386 4
bingo 5574 10
bingo 8361 0
bingo 8744 10
bingo 10935 11
bingo 12076 1
bingo 18688 4
bingo 19130 9
```
so, the program detects the odd output pixel which is entirely black - if there are no black pixels in the input, there may be no output.
But every now and then, the bug happens and the output becomes something like this:
```
bingo 0 1
bingo 0 3
bingo 0 9
bingo 0 11
bingo 1 1
bingo 1 3
bingo 1 9
bingo 1 11
and so on.
```
This is driving me nuts. Maybe I'm missing something, but I can't find any hint in the docu, e.g. more stuff I have to pass to OIIO to make it peform correctly - I tried all sorts of batch options, all to no avail. It looks like data being accessed through a pointer to memory which may or may not contain the correct content (like, it was there but may have been altered in the meantime) - or loading from a dangling pointer - just guessing. Like the mask gone random and then stuck, but passing in Tex::RunMaskOn should set all lanes. So I'm at a loss. I'd appreciate if someone could at least reproduce the behaviour - or tell me what I'm missing.

lgritz:
The doc comment said it should initialize, and the analogous structure for scalar texturing, TextureOpt, initialized. Missed this one.

Fixes #4219

### DIFF
--- a/src/include/OpenImageIO/texture.h
+++ b/src/include/OpenImageIO/texture.h
@@ -308,11 +308,18 @@ private:
 /// Texture options for a batch of Tex::BatchWidth points and run mask.
 class OIIO_API TextureOptBatch {
 public:
+    using simd_t = simd::VecType<float, Tex::BatchWidth>::type;
+
     /// Create a TextureOptBatch with all fields initialized to reasonable
     /// defaults.
     TextureOptBatch () {
-        for (int i = 0; i < Tex::BatchWidth; ++i)
-            rnd[i] = -1.0f;
+        *((simd_t*)&sblur) = simd_t::Zero();
+        *((simd_t*)&tblur) = simd_t::Zero();
+        *((simd_t*)&rblur) = simd_t::Zero();
+        *((simd_t*)&swidth) = simd_t::One();
+        *((simd_t*)&twidth) = simd_t::One();
+        *((simd_t*)&rwidth) = simd_t::One();
+        *((simd_t*)&rnd) = simd_t(-1.0f);
     }
 
     // Options that may be different for each point we're texturing


kfjahnke:
I tried this fix here and it works for me. I think you can merge it.

PR #4227, opened 2024-04-13
fix(fmt.h): fix build break from recent fmt change

lgritz:
We disable fmt execptions and force it to do one of our ASSERT's instead.  But fmt in their main branch recently added `[[noreturn]]` annotations in a way that cause this to be compiler errors for our overrides.  Needed an extra std::terminate() call to guarantee the noreturn promise.


lgritz:
Pre-emptively merging because this demonstrably fixes our broken CI. If anybody looks at this later and thinks it's a bad idea, please speak up and we'll find another solution.

PR #4228, opened 2024-04-14
fix: Improve IB::nsubimages and other related fixes

### LINKED ISSUE #4225: [BUG] OIIO::ImageBuf::nsubimages returns zero for existing image
**Describe the bug**

OIIO::ImageBuf::nsubimages returns zero for existing image:

```
OIIO::ImageBuf buf("existing_file.jpg"); // some file (tested JPG and TIFF)
int nsubimages = buf.nsubimages(); 
// -> 0 (incorrect) for OIIO 2.5.10.1
// -> 1 (correct) for OIIO 2.4.17.0, 2.3.15.0
```

Incorrect:

OIIO 2.5.10.1 | unknown arch?
    Build compiler: MSVS 1939 | C++14/199711
Dependencies:  , Boost 1.84.0, BZip2 1.0.8, FFmpeg NONE, Freetype 2.13.2, GIF 5.2.2, JPEG 62, Libheif NONE,
    libjpeg-turbo 3.0.2, LibRaw NONE, OpenColorIO NONE, OpenCV NONE, OpenEXR 3.2.3, OpenJPEG 2.5, PNG 1.6.43, Ptex
    NONE, Ptex NONE, Robinmap, TBB NONE, TIFF 4.6.0, WebP NONE, ZLIB 1.3.1

Correct:

OIIO 2.4.17.0, same dependencies

lgritz:
Background: Some formats that support multi-image files know as soon as they are opened and read the header what the total subimage count is.  Others (I'm looking at you, TIFF) can't know without reading each subimage in succession until you hit the end of the file. Format readers who can easily figure it out open first opening are supposed to set attribute "oiio:subimages" to the number of subimages, or 0 / unset if it can't be cheaply or immediately determined.

ImageBuf has a `nsubimages()` method that you'd think would give you the right count, and it did when all IBs were read through the ImageCache (which fully inventories the subimages of any file it reads). But when OIIO 2.5 changed IB behavior to only be backed by IC when explicitly requested, we effectively lost this functionality most of the time, and instead returned 0 ("don't know yet"). This was confusing. And also silly, because it reported 0 even for files whose formats don't permit multiple images (so you know the count is 1).

This patch does the following:

* Changes the behavior and documentation of `ImageBuf::nsubimages()` to the following return values:
  - 1, for formats that are always a single subimage. (Or for ImageBufs that are not direct reads of a file.) 
  - n>=1 for files that permit multiple subimages and the count is known. 
  - 0 for files that permit multiple subimages, but the count is not known because it's one of the formats where it can't easily be determined upon opening the file.
* Add a new ImageInput::supports("multiimage") that says if the input file format permits multiple images, and ensure that it returns true for all formats where this is the case.
* Ensure that formats permitting multiimage and can easily know the count up front set the "oiio:subimages" attribute (not all did).
* Document in the built-in format section of the docs the formats that are expected to set this attribute.

Fixes #4225

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -290,6 +290,9 @@ options are supported:
      - ptr
      - Pointer to a ``Filesystem::IOProxy`` that will handle the I/O, for
        example by reading from memory rather than the file system.
+   * - ``oiio:subimages``
+     - int
+     - The number of "image elements" (subimages) in the file.
 
 
 **Configuration settings for DPX output**
@@ -547,6 +550,9 @@ storage.  Currently, OpenImageIO only supports 2D FITS data (images), not 3D
    * - ``Hierarch``
      - string
      - FITS "HIERARCH" (*)
+   * - ``oiio:subimages``
+     - int
+     - The number of subimages in the file.
    * - *other*
      - 
      - all other FITS keywords will be added to the ImageSpec as arbitrary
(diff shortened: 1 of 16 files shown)


lgritz:
Any comments about this one?

lgritz:
Will merge this tomorrow if nobody objects.

PR #4232, opened 2024-04-15
style: Fix clang-format problem caused by 4221

lgritz:
Oops, we let that slip by.


fpsunflower:
Woops! sorry about that, must have gotten mixed up in the various refactors.

PR #4233, opened 2024-04-15
build(deps): Remove the very last references to boost

### LINKED ISSUE #4158: Dump Boost from dependencies
Boost used to be much more important to us, but over the years, most of the things we needed from boost have migrated to the C++11, 14, or 17 standard libraries. In more modern times, Boost is a very heavyweight dependency which causes a disproportionate amount of build pain, considering how very little of it we use. We don't use any boost classes in our public APIs, so removing legacy boost use is purely about internals and won't affect users at all. It's primarily a concern to people building OIIO, who have to make sure they have acceptable boost versions built as well. I think we can easily replace the very few remaining uses with alternatives and remove the Boost dependency entirely. They remaining uses are:

### boost::filesystem

This is only used in filesystem.cpp, and only for C++14 builds (when using a C++17-capable compiler, we already automatically switch to using std::filesystem. So there is no real work to be done here, it will be part of the bump o C++17 minimum. I'm just listing it to be complete.

### boost::thread_specific_ptr

This is used in several places. I wish there were a robust, single-header, dependency-free implementation of this were around and with a compatible license, but I haven't been able to find it. Maybe it's out there and somebody else can find it? Maybe just the parts we need can be extracted from Boost or some other large library that has the equivalent? Maybe since we use it in such a narrow way, we can just write our own minimal implementation?

Note that although this looks similar in spirit to C++ `thread_local`, it's different because thread_local must be of static scope, and we need something that can be a class member.

### boost::container::flat_map and flat_set

This will eventually be in C++23, but we're nowhere near being able to assume that's available. But I believe there are several acceptable single-header implementations that we could use which exactly emulate what will be in C++23 but work with older C++ versions. Or, possibly, just write it ourselves since we probably only use a tiny fraction of the features that will be in the std version.

### boost::algorithm functions: ifind_first, ifind_last, to_upper, to_lower

These are not really difficult to rewrite from scratch ourselves. Somebody just needs to do it.

### boost::stacktrace

This is only used in sysutil.cpp. This will be eventually replaced by C++23 std::stacktrace, but that's too far away for us to rely on now. However, I bet that (like all other C++ future library classes) there are multiple adequate implementations out there that we can use with older C++. Maybe somebody reading this already has their favourite to recommend.

That's it! Those are the only boost artifacts we have left in the code base. It feels totally doable.

```[tasklist]
### Tasks
- [x] filesystem
- [x] thread_specific_ptr
- [x] flat_map & flat_set
- [x] algorithms used in strutil.cpp
- [x] stacktrace
```

lgritz:
As Marie Kondo would say, Boost no longer sparks joy for us, so we take one last look, thank it for its many years of service to our project, and then send it on its way.

Fixes #4158

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: sonar gcc9/C++14 py39 boost1.76 exr3.1 ocio2.2
+          - desc: sonar gcc9/C++14 py39 exr3.1 ocio2.2
             nametag: static-analysis-sonar
             os: ubuntu-latest
             container: aswf/ci-osl:2023-clang15
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: gcc9/C++17 py3.7 boost1.73 exr2.5 ocio2.0
+          - desc: gcc9/C++17 py3.7 exr2.5 ocio2.0
             nametag: linux-vfx2021
             runner: ubuntu-latest
             container: aswf/ci-osl:2021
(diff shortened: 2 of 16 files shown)


ThiagoIze (approved):
If CI passes, this looks awesome to me :)

fpsunflower (approved):
🍾 🎉

PR #4234, opened 2024-04-15
cleanup: Remove some long-deprecated things

lgritz:
This is just the first step in many for removing deprecated things along the way to a 3.0 release. I'm doing a little at a time.

I believe that here I'm only removing things that (a) have been deprecated since 2.0 or earlier, (b) most or maybe all were marked with OIIO_DEPRECATED and so should have been triggering warnings for a long time if used downstream (thus, almost certainly nobody is using it), and (c) are inline definitions, so this change should not appear to break link ABIs.


ThiagoIze (approved):
Seems good to me.

PR #4235, opened 2024-04-16
Fix building failed from source on Windows

vic4key:
The instruction for building from source on Windows is incorrect and causes building to fail. I fixed it.


lgritz:
Thanks, this is very helpful.

Can you be sure to affix the DCO by signing the commit? You should be able to do it by:

    git commit --amend -s
    git push origin <yourbranch> --force

The -s affixes the "Signed-off-by" that confirms that you agree to the DCO.

vic4key:
Hi @lgritz The DCO check did not accept my name "Vic P.", it's fine with "Vic P" (without dot).
I updated the commit. You can merge it now. Thank you.

lgritz (approved):

PR #4236, opened 2024-04-16
JpegXL plugin

ssh4net:
## Description
JpegXL Output plugin
• support for uint8/uint16/half/float
• compression settings
  jpegxl:distance or compression jpegxl:##
  jpegxl:effort

JpegXL Input plugin
• support for uint8/uint16/half/float

**Encode:**
```
oiiotools -v ^
-i DISP_16bit.tif -o DISP_16bit_tif.jxl ^
-i  DISP_8bit.tif -o DISP_8bit_tif.jxl ^
-i DISP_float.exr -o DISP_float_exr.jxl ^
-i DISP_half.exr -o DISP_half_exr.jxl
```

jxlinfo.exe -v output
```
JPEG XL image, 1440x1440, (possibly) lossless, 16-bit RGB
JPEG XL image, 1440x1440, (possibly) lossless, 8-bit RGB
JPEG XL image, 1440x1440, (possibly) lossless, 32-bit float (8 exponent bits) RGB
JPEG XL image, 1440x1440, (possibly) lossless, 16-bit float (5 exponent bits) RGB
```

**Decode**
```
oiiotool.exe -v ^
-i w:\DISP_16bit_tif.jxl -o w:\DISP_16bit_tif_jxl.tif ^
-i w:\DISP_8bit_tif.jxl -o w:\DISP_8bit_tif_jxl.tif ^
-i w:\DISP_float_exr.jxl -o w:\DISP_float_exr_jxl.exr ^
-i w:\DISP_half_exr.jxl -o w:\DISP_half_exr_jxl.exr
```

iinfo.exe output
```
DISP_16bit_tif_jxl.tif : 1440 x 1440, 3 channel, uint16 tiff
DISP_8bit_tif_jxl.tif : 1440 x 1440, 3 channel, uint8 tiff
DISP_float_exr_jxl.exr : 1440 x 1440, 3 channel, float openexr
DISP_half_exr_jxl.exr : 1440 x 1440, 3 channel, half openexr
```

TODO:
- [ ] EXIF and metadata
- [ ] Color spaces support
- [x] Encoding settings
- [ ] Decoding settings

## OIIO Tests
not yet

## Checklist:

  nonconforming.


src/doc/builtinplugins.rst
1div0, line 1: Format official name is JPEG XL. Not JPEG-XL.

src/doc/builtinplugins.rst
```diff
@@ -1193,6 +1193,94 @@ special ``"oiio:ioproxy"`` attributes (see Sections
...
+.. _sec-bundledplugins-jpegxl:
+
+JPEG XL
+===============================================
+
+JPEG XL is a new image format that is designed to be a successor to JPEG
+and to provide better compression and quality. JPEG XL files use the file
+extension :file:`.jxl`. The official JPEG XL format specification and other
+helpful info may be found at: https://jpeg.org/jpegxl/
+
```
lgritz, line 1207: I just wanted to say: thanks so much for writing the documentation section, this is really helpful.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -272,27 +347,59 @@ JxlOutput::write_tile(int x, int y, int z, TypeDesc format, const void* data,
 
 
 bool
-JxlOutput::save_image()
+JxlOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
+                       int zend, TypeDesc format, const void* data,
+                       stride_t xstride, stride_t ystride, stride_t zstride)
+{
+    // placeholder
+
```
lgritz, line 355: I think you should call

    ImageOutput::write_tiles(...)

here, so that it's functional. The base class implements write_tiles by calling write_tile for each tile. But as this currently stands above, it will override that and write nothing at all, while still returning true to indicate success, meaning that any place we (or apps) call write_tiles will do the wrong thing.

lgritz (changes requested):
This looks great. My only concern for merging what you have so far is write_tiles.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -165,23 +192,72 @@ JxlOutput::open(const std::string& name, const ImageSpec& newspec,
...
+                lossless = true;
+            } else {
+                m_basic_info.uses_original_profile = JXL_FALSE;
+                JxlEncoderSetFrameDistance(
+                    m_frame_settings,
+                    1.0f / static_cast<float>(compqual.second));
+                JxlEncoderSetFrameLossless(m_frame_settings, JXL_FALSE);
+            }
+        } else {  // default to lossless
```
lgritz, line 230: Is this the right default? (I don't know, really am asking.) For no compression request, you use "jpegxl:100", and revert to lossless for any unrecognized compression name.

Imagine `oiiotool foo.xyz -o foo.jxl` so you may write with some compression name from another format without the user really being aware that this is making them lapse into lossless compression for the jxl file.

A possible alternative would be to use lossless if the compression was "none" or "lossless", but any other *unrecognized* name would default to a reasonable level of jpegxl compression (as you did if no "compression" request was made at all).

You be the judge -- I don't know what JXL user expectations would be, nor do I know how big a compression or quality different there will be between the lossless mode and the jpegxl:100 default you use when no compression is specified.

ssh4net: Personally i love Jpeg2000 and JpegXL because they have lossless compression that in same time doing files smaller than any other lossless file formats.
But if you think for a users better to set same settings as used in cjxl.exe from libjxl that use probably 70 or 95 (don't remember exactly)

src/doc/builtinplugins.rst
ssh4net: You welcome :)

src/jpegxl.imageio/jxloutput.cpp
lgritz: No, I trust your judgment. The information I didn't have is whether the lossless compression was any good. I'm ok with the rule of "you get lossless unless you specifically ask for lossy by name" as long as the lossless compression is pretty good and will not seem like a strange choice to JXL users.

jonnyawsom3: The reference implementation's cjxl losslessly transcodes jpeg inputs and uses normal lossless on GIF inputs, with a lossy quality of 90 by default for anything else

src/jpegxl.imageio/jxloutput.cpp
ssh4net: Sure.
I just found out how that was made in Tiff Writer that give me better idea of how that works.

src/jpegxl.imageio/jxloutput.cpp
ssh4net: @jonnyawsom3, Maybe that makes sense for a cxl tool that is an image re-encoding tool, and encodings is a "goal"?
When OpenImageIO is a library that is an image IO interface and data are the "hero"?

@lgritz uint16:
- tiff (zip) 7.93Mb
- png (9) 5.48Mb
- jxl (lossless) 2.86Mb

half:
- exr (piz) 3.71mb
- jxl (lossless) 2.28mb

And in same time jxl support hdr, multichannels, layers, etc. and sound like it can be read on any refrigerator soon :)

ssh4net:
Sorry, without PBM and PFM, it is inconvenient to work with any image compression libraries, so I also tried to add outputs for them in the PNM plugin.

PR #4237, opened 2024-04-19
security: change command line embedding for oiiotool & maketx output

lgritz:
It was pointed out that oiiotool and maketx's habit of embedding their exact command line arguments in the Software and ImageHistory metadata of output images was potentially problematic in that it might inadvertently leak important internal data (for example, file paths may reveal secret projects, etc.).

This patch changes the behavior of oiiotool and maketx as follows:

* The default behavior is now to use `"OpenImageIO <version> <hash>"` where `<hash>` is the SHA-1 hash of the command line arguments (which allows you to tell if two files use the same command line args, but doesn't let you know what they were).

* A new `--history` argument (or setting OIIOTOOL_METADATA_HISTORY env variable to a nonzero integer) instead uses the full command line arguments with no obfuscation. (This used to be the default behavior but is now deemed unsafe.) When the env variable is in effect, a `--no-history` argument explicitly disables it.

* For oiiotool, the existing `--nosoftwareattrib`, as before, causes the Software and ImageHistory metadata to not be modified at all.

Here is an example of the behavior:

```
$ oiiotool tmp.tif -o tmp.exr && iinfo -v tmp.exr
tmp.exr :  640 x  480, 3 channel, half openexr
    channel list: R, G, B
    Software: "OpenImageIO 2.6.2.0dev : B897973B86EC13E3F1858AAE5E057EFCCF99889A"
    ...

$ oiiotool --history tmp.tif -o tmp.exr && iinfo -v tmp.exr
tmp.exr :  640 x  480, 3 channel, half openexr
    channel list: R, G, B
    Software: "OpenImageIO 2.6.2.0dev : oiiotool --history tmp.tif -o tmp.exr"
    Exif:ImageHistory: "oiiotool --history tmp.tif -o tmp.exr"
    ...
```

### DIFF
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -365,4 +365,4 @@ inside the source code.
     images it outputs. The default if this is not set is to only write the
     name and version of the software and an indecipherable hash of the command
     line, but not the full human-readable command line. (This was added in
-    OpenImageIO 2.5.0.)
+    OpenImageIO 2.5.11.)
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -357,3 +357,12 @@ inside the source code.
     use for its thread pool. If both are set, ``OPENIMAGEIO_THREADS`` will
     take precedence. If neither is set, the default will be 0, which means
     to use as many threads as there are physical cores on the machine.
+
+``OIIOTOOL_METADATA_HISTORY``
+
+    If set to a nonzero integer value, `oiiotool` will by default write the
+    command line into the ImageHistory and Software metadata fields of any
+    images it outputs. The default if this is not set is to only write the
+    name and version of the software and an indecipherable hash of the command
+    line, but not the full human-readable command line. (This was added in
+    OpenImageIO 2.5.0.)
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -1975,10 +1975,10 @@ current top image.
 
     Prior to OpenImageIO 2.5.11, the full information was always written, but
     could be overridden with `--nosoftwareattrib`. Beginning with 2.5.11, the
-    default changed only say the software name and version (unless the
+    default changed to only write the software name and version (unless the
     `OIIOTOOL_METADATA_HISTORY` environment variable is set), and require the
-    new `--history` option to cause the command line arguments to be saved as
-    meetadata.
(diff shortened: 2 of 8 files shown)


src/doc/imageioapi.rst
```diff
@@ -357,3 +357,12 @@ inside the source code.
...
     to use as many threads as there are physical cores on the machine.
+
+``OIIOTOOL_METADATA_HISTORY``
+
+    If set to a nonzero integer value, `oiiotool` will by default write the
+    command line into the ImageHistory and Software metadata fields of any
+    images it outputs. The default if this is not set is to only write the
+    name and version of the software and an indecipherable hash of the command
+    line, but not the full human-readable command line. (This was added in
+    OpenImageIO 2.5.0.)
```
ThiagoIze, line 368: Should it be this?
```suggestion
    OpenImageIO 2.6.0.)
```

src/doc/oiiotool.rst
```diff
@@ -1952,10 +1952,33 @@ current top image.
...
+    nonzero integer value, the `--history` option will be enabled by default,
+    but can be disabled on the command line with `--no-history`.
+
+    The `--nosoftwareattrib` option prevents even the minimal default information
+    from being written, so that no information about the software is written
+    to any metadata field.
+
+    Prior to OpenImageIO 2.5.11, the full information was always written, but
+    could be overridden with `--nosoftwareattrib`. Beginning with 2.5.11, the
+    default changed only say the software name and version (unless the
```
ThiagoIze, line 1978:
```suggestion
    default changed only save the software name and version (unless the
```

ThiagoIze: or since "saved" is used later in the sentence in a different way and so could be confusing:
```suggestion
    default changed only except for the software name and version (unless the
```

src/doc/imageioapi.rst
lgritz: I was assuming people would want this backported to the next monthly patch of the release branch.

ThiagoIze: So then should it be 2.5.11?

lgritz: Sorry, I answered that too fast. It shouldn't read 2.5.0 either. It will end up being 2.5.11.

ThiagoIze (approved):

PR #4239, opened 2024-04-21
fix: Error retrieval safeguards for recycled objects

lgritz:
The recent switch from boost thread_specific_ptr left a bug: if an ImageInput (say) is deleted without the caller clearing its error state, and then later another ImageInput is allocated and happens to get the same address as the first one, it can "inherit" the unretrieved error state left behind by the other.

Fix by clearing the error state at construction/initialization.

This still doesn't 100% fix all possible cases. If the calling thread that left behind an unretrieved error in the first object is a different thread than the one that creates the next recycled object, there is still some cruft left behind.

I feel like we still don't have a foolproof replacement for the really nice lifetime management that thread_specific_ptr was doing for us. One could argue that unretrieved errors in an object are a "bug" in the user's program, but it's an exceptionally easy mistake to make and shouldn't result in heisenbehavor that differs based on whether allocations happen to use the same addresses for other objects.


lgritz:
Revision: unique counter in objects with per-thread error state

It seems too troublesome to index per-thread error state by object
address, since those objects may be deleted and the same address
reused later for a different object of the same type.

Instead, now using a unique ID (initialized by an incrementing atomic
counter) to index the object in the per-thread error store.

src/libOpenImageIO/imageoutput.cpp
```diff
@@ -80,6 +86,13 @@ ImageOutput::operator delete(void* ptr)
 ImageOutput::ImageOutput()
     : m_impl(new Impl, impl_deleter)
 {
+    // Ensure that if this ImageOutput has the same address as a previous one
```
fpsunflower, line 89: Did you mean to leave this in? The comment refers to the address, but is using the id.

lgritz: I did not. Fixing.

src/libtexture/imagecache.cpp
```diff
@@ -2763,7 +2761,7 @@ ImageCacheImpl::get_image_info(ImageCacheFile* file,
     if (file->broken()) {
         if (file->errors_should_issue()) {
             // eat any earlier messages
-            imcache_error_messages.erase(this);
+            imcache_error_messages.erase(imagecache_id_atomic);
```
fpsunflower, line 2764: I think you meant to use `imagecache_id` here?

Maybe rename `imagecache_id_atomic` with `imagecache_next_id` to avoid confusion?

lgritz: Quite right. Fix pushed.

fpsunflower (approved):
LGTM!

PR #4243, opened 2024-04-24
Add attribute to ignore system environment variables for plugin path

brechtvl:
## Description

Sometimes it's useful to run an application more isolated, without picking up plugins from `LD_LIBRARY_PATH`. These may have been installed for another purpose, or be incompatible in some way.

## Tests

Not easy to test, but hopefully not needed.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -2854,6 +2854,11 @@ OIIO_API std::string geterror(bool clear = true);
 ///    Colon-separated (or semicolon-separated) list of directories to search
 ///    for dynamically-loaded format plugins.
 ///
+/// - `int plugin_search_system_paths`
+///
+///    Search for plugins using LD_LIBRARY_PATH and DYLD_LIBRARY_PATH
+///    environment variables. Enabled by default.
+///
 /// - `int try_all_readers`
 ///
 ///    When nonzero (the default), a call to `ImageInput::create()` or
--- a/src/libOpenImageIO/imageio.cpp
+++ b/src/libOpenImageIO/imageio.cpp
@@ -55,6 +55,7 @@ int limit_imagesize_MB(std::min(32 * 1024,
                                 int(Sysutil::physical_memory() >> 20)));
 ustring font_searchpath(Sysutil::getenv("OPENIMAGEIO_FONTS"));
 ustring plugin_searchpath(OIIO_DEFAULT_PLUGIN_SEARCHPATH);
+int plugin_search_system_paths(1);
 std::string format_list;         // comma-separated list of all formats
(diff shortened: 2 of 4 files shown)


brechtvl:
For context, I've been debugging an issue where Blender launched through Steam is crashing for some users, due to some incompatible plugin in `/usr/lib`. I have not actually been able to reproduce it or figure out exactly why it happens.

However in general, I would prefer for OIIO not to search such system paths. We disable similar behavior in e.g. Python by default, and e.g. USD only searches in its own environment variable for plugins. It's just rather risky for a DCC that bundles a bunch of its own libraries to mix with system libraries.

lgritz:
This is fine, but if you this it would be better to get rid of the system paths entirely and ONLY search OIIO_LIBRARY_PATH, I would have no objections.

src/include/OpenImageIO/imageio.h
```diff
@@ -2854,6 +2854,11 @@ OIIO_API std::string geterror(bool clear = true);
 ///    Colon-separated (or semicolon-separated) list of directories to search
 ///    for dynamically-loaded format plugins.
 ///
+/// - `string plugin_search_system_paths`
```
sergeyvfx, line 2857: It is an integer, would be nice to correct the comment if we are to stick to this implementation.

brechtvl:
I'd be happy getting rid of `LD_LIBRARY_PATH` entirely. I just don't have any insight in how many people actually rely on this.

Docs only seem to mention `OIIO_LIBRARY_PATH`. And `LD_LIBRARY_PATH` isn't set by default on most Linux distributions I think, they rely more on `/etc/ld.so.conf`. So maybe it's fine.

I created #4245 to remove them entirely, so you can choose.

lgritz:
> I'd be happy getting rid of `LD_LIBRARY_PATH` entirely. I just don't have any insight in how many people actually rely on this.

The original OIIO design was that all formats were dynamic modules, literally no real difference in standing between the ones in the OIIO project and others people would add. So there was a premium to make sure no stone was unturned when looking for them.

But this was hard to get right, it seemed more foolproof to default to building them all as part of the library itself so there was nothing to "find", and the dynamic search is only needed if a site has a completely custom format plugin that for some reason they don't wish to just add to the open source project.

I think that in practice, there are few of these, they are written very purposefully by sophisticated users, for whom putting them in a canonical place and setting `OIIO_LIBRARY_PATH` to find it is probably not a big ask.

So I'm inclined to say that it's not going to hurt anybody relying on this feature. Really all they need to do is append $LD_LIBRARY_PATH's contents onto OIIO_LIBRARY_PATH. Low bar.

PR #4245, opened 2024-04-25
Don't use (DY)LD_LIBRARY_PATH as plugin search paths

brechtvl:
## Description

Only use the OIIO_LIBRARY_PATH environment variable by default.

Searching through such system paths has some risk loading plugins installed for another purpose or that may be incompatible.

If the previous behavior is desired, applications can add these environment variables to the plugin search path manually.

## Tests

N/A

## Checklist:

  nonconforming.


lgritz (approved):
Let's go for it!

This seems more in line with our strengthening security goals as well.

PR #4246, opened 2024-04-27
build: Disable clang18 warnings about deprecated unicode conversion

lgritz:
clang/llvm 18 (and probably other compilers soon) has made noisy warnings about the C++17-deprecated unicode conversions. Actually replacing them enirely with something else is rather hairy, we will tackle that later.

In the mean time, this patch seems to be the minimal change to disable the warnings that should be totally easy to backport.

These seem to be the only changes necessary to make a clean build with clang 18.


meshula (commented):
seems good as a temp workaround

PR #4247, opened 2024-04-27
cleanup: more old style format deprecation

lgritz:
Remove some long-ago deprecated printf-like error functions. Mark as deprecated some other ones that we will remove when it gets closer to 3.0 release. I think these are the very last uses internally of the old style that we want to remove.


jessey-git (approved):
Looks good.

PR #4250, opened 2024-04-29
dev: Add Strutil::eval_as_bool

lgritz:


src/oiiotool/oiiotool.cpp
```diff
@@ -1137,7 +1137,7 @@ control_if(Oiiotool& ot, cspan<const char*> argv)
     if (ot.running()) {
         // string_view command = ot.express(argv[0]);
         string_view value = ot.express(argv[1]);
-        cond              = eval_as_bool(value);
+        cond              = Strutil::eval_as_bool(value);
```
jessey-git, line 1140: There's a few remaining uses of the file-local `eval_as_bool` along with the old implementation itself. Were there reasons to leave those behind?

lgritz: There's an unrelated PR coming on the heels of that where they will get fixed, but I couldn't cleanly separate those out of the other changes.

jessey-git (approved):

PR #4253, opened 2024-05-01
pnm-plugin

ssh4net:
## Description

Fixes added:

- import/export support for 8/16 uint 32-float
- pnm:bigendian
- pnm:pfmflip for float pfm files (GIMP flip float PFM files by default, Photoshop do not)
- update in documentation

## Checklist:

  nonconforming.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1732,8 +1732,7 @@ attributes are supported:
      - If nonzero, the PNM file is big-endian (the default is little-endian).  
    * - ``pnm:pfmflip``
       - int
-      - If nonzero, the PFM (float) file is flipped upside-down (the default
-      is flipped).
+      - If this configuration hint is present and is zero, the automatic vertical flipping of PFM image will be disabled (i.e., scanline 0 will really be the first one stored in the file). If nonzero (the default), float PFM files will store scanline 0 as the last scanline in the file (i.e. the visual "top" of the image).
 
 **Configuration settings for PNM output**
 
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1766,8 +1766,7 @@ control aspects of the writing itself:
        Float PFM files are always written in binary format.
    * - ``pnm:pfmflip``
       - int
-      - If nonzero, the PFM (float) file is flipped upside-down (the default
-      is flipped).
+      - If this configuration hint is present and is zero, for PFM files, scanline 0 will really be stored first in the file, thus disabling the usual automatically flipping that accounts for PFM files conventionally being stored in bottom-to-top order. If nonzero (the default), float PFM files will store scanline 0 as the last scanline in the file (i.e. the visual "top" of the image).
 
 **Custom I/O Overrides**
 
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1688,7 +1688,7 @@ Without loss of generality, we will refer to these all collectively as
 for those who reject the nonsense about naming the files depending on the
 number of channels and bitdepth.
 
-PNM files widely used in the Unix world as simple ASCII or binary image 
+PNM files are widely used in the Unix world as simple ASCII or binary image 
 files that are easy to read and write. They are not compressed, and are
 not particularly efficient for large images. They are not widely used in
 the professional graphics world, but because of their historical
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1712,6 +1712,11 @@ Both the reader and writer accept configuration hints "pnm:pfmflip" (default: 1)
      - int
      - The true bits per sample of the file (1 for true PBM files, even
        though OIIO will report the ``format`` as UINT8).
+   * - ``pnm:binary``
+     - int
+     - nonzero if the file itself used the PNM binary format, 0 if it used
+       ASCII.  The PNM writer honors this attribute in the ImageSpec to
+       determine whether to write an ASCII or binary file.
 
 **Configuration settings for PNM input**
 
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1696,6 +1696,9 @@ significance and extreme simplicity, OpenImageIO supports them.
 PNM files do not support anything other than 1 or 3 channels, no tiles,
 no multi-image, no MIPmapping.
 
+The pbm, pgm, and ppm varieties are stored with scanlines ordered in the file as top-to-bottom (the same as the usual OIIO convention), but the float-based pfm files are conventionally ordered in the file as bottom-to-top. Therefore, by default, reading and writing of the pfm variety will automatically flip the image so that an application calling the OpenImageIO API can, as usual, assume that scanline 0 is the visual "top" (even though it is actually the last scanline stored in the file).
+
+Both the reader and writer accept configuration hints "pnm:pfmflip" (default: 1), which if set to 0 will disable this flipping and ensure that scanline 0 is written as the first in the file (therefore representing what PFM assumes is the visual "bottom" of the image). This hint only affects PFM files and has no effect on the pbm, pgm, or ppm varieties.
 **Attributes**
 
 .. list-table::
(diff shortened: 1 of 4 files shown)


src/pnm.imageio/pnminput.cpp
```diff
@@ -339,7 +348,13 @@ PNMInput::open(const std::string& name, ImageSpec& newspec,
                const ImageSpec& config)
 {
     ioproxy_retrieve_from_config(config);
-    return open(name, newspec);
+    if (!open(name, newspec)) {
+        errorfmt("Could parse spec for file \"%s\"", name);
```
lgritz, line 352:
```suggestion
        errorfmt("Could not parse spec for file \"%s\"", name);
```

src/pnm.imageio/pnminput.cpp
```diff
@@ -339,7 +348,13 @@ PNMInput::open(const std::string& name, ImageSpec& newspec,
...
+        return false;
+    }
+
+    m_spec["pnm:pfmflip"] = config.get_int_attribute("pnm:pfmflip", 1);
```
lgritz, line 356: `m_spec` is for communicating what's in the file to the caller.

`config` is for the caller to give us instructions about how to read.

So it's unusual to copy something from the config into m_spec. I suspect you are merely trying to communicate that config hint to the later reading functions, but usually what we do in a case like this is just have a `bool` as a private member of the PNMInput, set it here, and use it as needed in the read_scanline function.

src/doc/builtinplugins.rst
```diff
@@ -1731,6 +1727,13 @@ attributes are supported:
      - ptr
      - Pointer to a ``Filesystem::IOProxy`` that will handle the I/O, for
        example by reading from memory rather than the file system.
+   * - ``pnm:bigendian``
+     - int
+     - If nonzero, the PNM file is big-endian (the default is little-endian).  
+   * - ``pnm:pfmflip``
+      - int
+      - If nonzero, the PFM (float) file is flipped upside-down (the default
+      is flipped).
```
lgritz, line 1736: This is a little confusing. Is the default to flip or not to flip?

ssh4net: Flipped. Because of GIMP and NetPBM that have this by default flipped. As result libjxl tools also export flipped pfm.

When I personally prefer non flipped

lgritz:
I'm a little confused, so let me say what I think is going on, and you can confirm or let me know where I'm getting it wrong:

The pbm, pgm, and ppm varieties are stored with scanlines ordered in the file as top-to-bottom (the same as the usual OIIO convention), but the float-based pfm files are conventionally ordered in the file as bottom-to-top. Therefore, by default, reading and writing of the pfm variety will automatically flip the image so that an application calling the OpenImageIO API can, as usual, assume that scanline 0 is the visual "top" (even though it is actually the last scanline stored in the file). 

Both the reader and writer accept configuration hints "pnm:pfmflip" (default: 1), which if set to 0 will disable this flipping and ensure that scanline 0 is written as the first in the file (therefore representing what PFM assumes is the visual "bottom" of the image). This hint only affects PFM files and has no effect on the pbm, pgm, or ppm varieties.

ssh4net:
Correct.
pnm:pfmflip is only affecting PFM (float NetPBM) and only because of historical reasons someone from GIMP project decided to have it flipped upside down: https://netpbm.sourceforge.net/doc/pfm.html

If you think it worth to disable flipping to be on the same page with USC ITC (https://pauldebevec.com/Research/HDR/PFM/ ) and Adobe I'm in!

lgritz:
Is GIMP the only major faction that assumes y=0 is the bottom? Adobe thinks it's the top?

Is there an official "PFM spec" and does it address the issue?

When you encounter pfm files in the wild, do they tend to mostly follow the GIMP convention?

ssh4net:
Gimp, tools from libjxl, NetPBM package... looks like all that use bottom line as y=0.

And there is no official spec for this extension of NetPBM format 🤷‍♂️

Float NetPBM is rarely used, as most of image sources are uint.

src/doc/builtinplugins.rst
lgritz, line 1736:
```suggestion
      - If this configuration hint is present and is zero, the automatic vertical flipping of PFM image will be disabled (i.e., scanline 0 will really be the first one stored in the file). If nonzero (the default), float PFM files will store scanline 0 as the last scanline in the file (i.e. the visual "top" of the image).
```

src/doc/builtinplugins.rst
```diff
@@ -1753,6 +1756,19 @@ control aspects of the writing itself:
...
+   * - ``pnm:binary``
+     - int
+     - nonzero if the file itself used the PNM binary format, 0 if it used
+       ASCII.  The PNM writer honors this attribute in the ImageSpec to
+       determine whether to write an ASCII or binary file.
+       Float PFM files are always written in binary format.
+   * - ``pnm:pfmflip``
+      - int
+      - If nonzero, the PFM (float) file is flipped upside-down (the default
+      is flipped).
```
lgritz, line 1771:
```suggestion
      - If this configuration hint is present and is zero, for PFM files, scanline 0 will really be stored first in the file, thus disabling the usual automatically flipping that accounts for PFM files conventionally being stored in bottom-to-top order. If nonzero (the default), float PFM files will store scanline 0 as the last scanline in the file (i.e. the visual "top" of the image).
```

testsuite/pnm/ref/out.txt
```diff
@@ -46,6 +51,7 @@ src/rgb-binary.ppm   :   32 x   32, 3 channel, uint8 pnm
     channel list: R, G, B
     oiio:ColorSpace: "Rec709"
     pnm:binary: 0
+    pnm:pfmflip: 1
```
lgritz, line 54: I don't think we should add the pnm:pfmflip attribute reporting to files we read that are not pfm files.

src/pnm.imageio/pnminput.cpp
lgritz: I think at the very least, we should only set this in m_spec if it really is a pfm file, not for the non-float varieties.

lgritz (changes requested):
Code looks fine other than the reporting of the flip attribute upon read for non-pfm files.

I made some suggestions for clarifying the docs regarding flipping.

The rest all looks good to me.

src/doc/builtinplugins.rst
```diff
@@ -1708,11 +1709,6 @@ It's not a smart choice unless you are sending your images back to the
      - int
      - The true bits per sample of the file (1 for true PBM files, even
        though OIIO will report the ``format`` as UINT8).
-   * - ``pnm:binary``
-     - int
-     - nonzero if the file itself used the PNM binary format, 0 if it used
-       ASCII.  The PNM writer honors this attribute in the ImageSpec to
-       determine whether to write an ASCII or binary file.
```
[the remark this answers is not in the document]
ssh4net: probably deleted by mistake.

testsuite/pnm/ref/out.txt
ssh4net: fixed

ssh4net:
@lgritz hope fixed now

src/doc/builtinplugins.rst
```diff
@@ -1678,22 +1678,38 @@ and :ref:`sec-imageinput-ioproxy`) as well as the `set_ioproxy()` methods.
...
 number of channels and bitdepth.
 
-PNM files are not much good for anything, but because of their
-historical significance and extreme simplicity (that causes many
-"amateur" programs to write images in these formats), OpenImageIO
-supports them.  PNM files do not support floating point images, anything
-other than 1 or 3 channels, no tiles, no multi-image, no MIPmapping.
-It's not a smart choice unless you are sending your images back to the
-1980's via a time machine.
+PNM files widely used in the Unix world as simple ASCII or binary image
```
lgritz, line 1691:
```suggestion
PNM files are widely used in the Unix world as simple ASCII or binary image 
```

lgritz (approved):
I just fixed one last typo myself rather than do another edit cycle for you.

LGTM!

PR #4257, opened 2024-05-04
build: More warning elimination for clang18

lgritz:
This is a continuation of PR 4246, had one more spot where we needed to disable warnings for unity builds involving clang18 and deprecated unicode conversion functions.

There's some changes included to ci.yml and install_homebrew_deps.bash that are part of work I did to make it easier for me to force it to use clang-18 for certain tests while working on this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -430,7 +430,7 @@ jobs:
             !build/testsuite/j2kp4files_v1_5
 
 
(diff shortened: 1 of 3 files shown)


lgritz:
@fpsunflower This should fix that OSL Mac CI failure.

lgritz:
Merging, since it fixes an OSL CI failure.

PR #4258, opened 2024-05-04
cleanup: change raw cerr stream output to print

lgritz:
A small piece of the ongoing switch to std::format/print style text output. This part is for error and debugging output when reading/writing exr images.

I also switched conditional compilation to the DBG macro idiom that I've used elsewhere, where it takes a combination of DEBUG compile and setting of environment variable OIIO_DEBUG_OPENEXR to trigger the printf style debugging (the first use case of this was in our color management code, which I also modify slightly here to conform).

Also opportunistically simplify gcd declaration now that we're C++17 minimum.


lgritz:

PR #4260, opened 2024-05-09
nVIDIA CUDA R3D

1div0:
## Description

This is intended to add support of R3D CUDA based decoder, which is much faster than R3D CPU only decoder.

RED One clips are handled by the CPU.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -992,7 +992,7 @@ inline void convert_type (const S *src, D *dst, size_t n)
 template<typename S, typename D>
 void convert_type (cspan<S> src, span<D> dst,
                    D min = std::numeric_limits<D>::min(),
-                   D max = std::numeric_limits<D>::min())
+                   D max = std::numeric_limits<D>::max())
 {
     OIIO_DASSERT(src.size() == dst.size());
     convert_type(src.data(), dst.data(), std::min(src.size(), dst.size()),


1div0:
//Flags used by the linker during the creation of shared libraries
// during all build types.
CMAKE_SHARED_LINKER_FLAGS:STRING=-L/usr/local/cuda -lcudart

1div0:
There is still needed to add:

`CMAKE_SHARED_LINKER_FLAGS        -lcudart`

src/r3d.imageio/r3dinput.cpp
```diff
@@ -5,12 +5,17 @@
...
 //
-// The code has been tested with the version 8.5.1 installed in
-// /opt/R3DSDKv8_5_1 directory and setting up the variable
-// export R3DSDK_ROOT="/opt/R3DSDKv8_5_1"
+// The code has been tested with the version 9.0.1 installed in
+// /opt/R3DSDKv9_0_1 directory and setting up the shell variable
+// export R3DSDK_ROOT="/opt/R3DSDKv9_0_1"
 
+#define GPU
+#define CUDA
```
lgritz: I'm concerned about setting these unconditionally.
Does every system that would have R3DSDK always have the CUDA SDK available?

We [already have](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/cmake/cuda_macros.cmake) a CMake variable called OIIO_USE_CUDA to enable CUDA at build time, and when that is enabled and the CUDA SDK is found, it will define the C++ preprocessor symbol `OIIO_USE_CUDA`, so you should use that rather than making new symbols.

1div0: Yes, indeed. These definitions do belong to the build configuration settings. Will modify ASAP.

Within the R3DSDK there are four options: CPU only decoding, nVIDIA CUDA, OpenCL and 🍏Metal.

Issue #4263, opened 2024-05-14
[FEATURE REQUEST] Proxy support for EXR multipart output


jreichel-nvidia:
It looks as if EXR multipart output does not support I/O proxies. In my experiments all methods seem to succeed, but I never receive any data from my proxy. There is a comment in exroutput.cpp:
```
    // m_output_multipart.reset (new Imf::MultiPartOutputFile (*m_output_stream,
    //                                          &m_headers[0], subimages)();
    // FIXME: Oops, looks like OpenEXR 2.0 currently lacks a
    // MultiPartOutputFile ctr that takes an OStream, so we can't
    // do this quite yet.
```
Maybe I'm missing something but it looks as if that constructor is available since at least OpenEXR 2.0.0:
```
    MultiPartOutputFile (
        OStream&      os,
        const Header* headers,
        int           parts,
        bool          overrideSharedAttributes = false,
        int           numThreads               = globalThreadCount ());
```
Isn't that the one that is needed?

lgritz:
Was closed by #4263

lgritz:
> Does this fix meet the requirements for a merge to the 2.5 branch?

@jreichel-nvidia Yes, certainly, it will be in the June 1 patch release.

PR #4264, opened 2024-05-15
Add proxy support for EXR multipart output (#4263)

jreichel-nvidia:
## Description

Add proxy support for EXR multipart output by taking the setup code from the other open() overload and calling the constructor accepting a stream instead of a filename.

## Tests

I did not add any new tests. Is proxy support tested by the testsuite?

## Checklist:

  nonconforming.

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -573,13 +573,21 @@ OpenEXROutput::open(const std::string& name, int subimages,
 
     // Create an ImfMultiPartOutputFile
     try {
-        // m_output_stream.reset (new OpenEXROutputStream (name.c_str())();
-        // m_output_multipart.reset (new Imf::MultiPartOutputFile (*m_output_stream,
-        //                                          &m_headers[0], subimages)();
-        // FIXME: Oops, looks like OpenEXR 2.0 currently lacks a
-        // MultiPartOutputFile ctr that takes an OStream, so we can't
-        // do this quite yet.
-        m_output_multipart.reset(new Imf::MultiPartOutputFile(name.c_str(),
+        if (!m_io) {
+            m_io = new Filesystem::IOFile(name, Filesystem::IOProxy::Write);
+            m_local_io.reset(m_io);
+        }
+        OIIO_ASSERT(m_io);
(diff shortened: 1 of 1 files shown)


lgritz:
To make the issue close when the PR is merged, you need to use specific words:

`Closes #1234`

or

`Fixes #1234`

src/openexr.imageio/exroutput.cpp
```diff
@@ -573,13 +573,21 @@ OpenEXROutput::open(const std::string& name, int subimages,
...
+            m_io = new Filesystem::IOFile(name, Filesystem::IOProxy::Write);
+            m_local_io.reset(m_io);
+        }
+        OIIO_ASSERT(m_io);
+        if (m_io->mode() != Filesystem::IOProxy::Write) {
+            // If the proxy couldn't be opened in write mode, try to
+            // return an error.
+            std::string e = m_io->error();
+            errorf("Could not open \"%s\" (%s)", name,
+                   e.size() ? e : std::string("unknown error"));
```
lgritz, line 586: We're trying to deprecate the printf style in favor of the std::format style:

```suggestion
            errorfmt("Could not open \"{}\" ({})", name,
                     e.size() ? e : std::string("unknown error"));
```

src/openexr.imageio/exroutput.cpp
```diff
@@ -573,13 +573,21 @@ OpenEXROutput::open(const std::string& name, int subimages,
...
-        //                                          &m_headers[0], subimages)();
-        // FIXME: Oops, looks like OpenEXR 2.0 currently lacks a
-        // MultiPartOutputFile ctr that takes an OStream, so we can't
-        // do this quite yet.
-        m_output_multipart.reset(new Imf::MultiPartOutputFile(name.c_str(),
+        if (!m_io) {
+            m_io = new Filesystem::IOFile(name, Filesystem::IOProxy::Write);
+            m_local_io.reset(m_io);
+        }
+        OIIO_ASSERT(m_io);
```
lgritz, line 580: I think this assertion can be dropped, since there's no possible way it will ever trigger? Either m_io starts as non-null, or you unconditionally assign it a new object.

lgritz (commented):
It LGTM, and thanks!

I did make two very minor suggestions, if you are inclined.

PR #4270, opened 2024-05-20
build: gcc-14 support, testing, CI

lgritz:
* Switch runner type for bleeding edge, use gcc14 and python 3.12

* Various minor gcc-14 + C++20 + Python 3.12 fixes

Notable: gcc14 warns about std::atomic_load for share_ptr as deprecated, since C++20 provides a true atomic shared pointer.


lgritz:
Does this seem reasonable?

lgritz:
Merging, as this fixes CI and allows builds with gcc 14.

PR #4272, opened 2024-05-21
build: Fix warning on windows

lgritz:


ThiagoIze (approved):
There's various ways to fix this and this particular way seems fine to me and I'm happy to accept it. If you want to be super [pedantic](https://stackoverflow.com/a/1464194), I think you want `intptr_t` instead of `size_t`.

lgritz:
Yeah, I opted for the pedantic suggestion, thanks.

ThiagoIze (approved):

PR #4273, opened 2024-05-23
build: Partially revert #4193, the avoidance of add_blah wasn't necessary

lgritz:
I misunderstood all along -- the add_compile_definitions, add_compile_options, etc., aren't truly global, they are restricted to targets in the same directory or below. They aren't really prone to polluting "surrounding" projects when OIIO is used as a subproject.

So for simplicity, reverting that part and using the usual built-ins.


lgritz:
Any opinions? Sorry for jerking everybody around by having made a change and now backing out of it. I like it better the old simpler way, and the changes I made were mostly unnecessary (I realize now).

PR #4274, opened 2024-05-25
build: new set_utils.cmake for various handy "set()" wrappers

lgritz:
New set_utils.cmake has some handy new utilities: set_cache() is much like the built-in set() when making a cache variable, and set_option() is much like build-in option(). The biggest difference is that both allow an environment variable of the same name, if it exists, to supply the default value. This is something that cmake does with many of its own controls, like CMAKE_BUILD_TYPE, but does not make any provision for built-in set() or option() let users do it.


lgritz:
You may recognize this as a piece of #4242 that I've broken off -- just the new "set" wrapper utilities -- because I would like to review those in isolation and merge them right away so that I can start using them elsewhere, even before we're reached consensus on the rest of #4242. Ever since I wrote them some weeks back, I find myself reaching to use them in other instances, and keep getting frustrated that they're not yet part of the code base, but only part of a PR that's been slowly chugging along for weeks.

lgritz:
I'm planning to merge this today if nobody objects. This PR merely adds new utilities (which I find helpful and want to use), but doesn't use them anywhere yet, so I don't think that this merge can possibly break anything.

PR #4275, opened 2024-05-25
fix(fmath.h): One more fast_exp fix

lgritz:
In PR #4268, I accidentally changed the clause addressing what is seemingly a bug with older MSVS. I'm just putting things back the way they were. It's not a new change, just restoring this line closer to the way it was about 3 commits back.

### DIFF
--- a/src/include/OpenImageIO/fmath.h
+++ b/src/include/OpenImageIO/fmath.h
@@ -1852,7 +1852,7 @@ template<typename T>
 OIIO_FORCEINLINE OIIO_HOSTDEVICE T fast_exp2 (const T& xval) {
     using namespace simd;
     typedef typename T::vint_t intN;
-#if OIIO_SIMD_SSE || OIIO_SIMD_NEON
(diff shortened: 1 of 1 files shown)


ThiagoIze (approved):
Shoot, when I saw this removed I was hoping this was because it was fixed on older msvc on account of us no longer relying on undefined behavior. Oh well, if we aren't sure it's fixed, let's play it safe.

lgritz:
I think I was hoping that my last fix took care of this, too, but apparently not, and it's a *different* symptom than what got fixed with the undefined behvaior fix. It's hard to reproduce, and only on Windows while using the older MSVS (whereas the other easily reproduced in gcc as well), so I think this one really is a compiler bug, but I admit that my confidence in that claim is lower than it was before, only because the other thing did also affect fast_exp. But I can't find any identifiable problems, so the only plan I have at the moment is disabling this one thing on the one older compiler that is problematic, unless/until it reproduces anywhere else.

PR #4277, opened 2024-05-31
refactor: move most of imageio_pvt.h.in to just a regular .h

lgritz:
This file imageio_pvt.h.in was used as the input template for CMake to create imageio_pvt.h, doing a couple build variable substitutions along the way. But this was inconvenient (because it doesn't look like a real .h file to editors or clang-format) and it turns out that there were only 2 lines that needed the substitution.

So in this patch, we split the file, making 99% of it just plain old imageio_pvt.h that needs no rewriting, and the couple lines that need variable substitions becomes buildopts.h.in and is included only from the one source file that actually cares, imageio.cpp.

Other than cleaving those two definitions off to another header and some file renaming, I haven't changed any of the other content of the headers, so there should be no change in code the compiler sees or other behavior.

The reformatting you see is because now that it's a legit .h file, clang-format wants its way with it.

### DIFF
--- a/src/include/imageio_pvt.h
+++ b/src/include/imageio_pvt.h
@@ -50,31 +50,40 @@ extern atomic_ll IB_local_mem_current;
 extern atomic_ll IB_local_mem_peak;
 extern std::atomic<float> IB_total_open_time;
 extern std::atomic<float> IB_total_image_read_time;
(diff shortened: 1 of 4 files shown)


ThiagoIze (approved):
this is much better!

PR #4283, opened 2024-06-03
build: Raise freetype minimum to 2.10.0

lgritz:
freetype 2.10.0 dates from 2019, so we're still supporting 5 years back.

Previously, we enforced no minimum, so we're not sure how far back it would really work, but the earliest version we tested in CI was 2.8, which dated from 2018.

Add a build_Freetype.bash for easy setup for CI (and other users).

CI many cases need to build freetype, because the old CentOS 7 era freetype is now too old.

Remove some refrence output that was only needed for older freetype.


lgritz:
Amended this PR:

* Rebased on top of the recently merged infrastructure that allows missing dependencies to build themselves locally.
* Added such a builder for freetype.

lgritz:
Any objections?

PR #4288, opened 2024-06-11
build: Upgrade to more modern python3 finding

lgritz:
Latest versions of cmake are warning about FindPython being a deprecated module, they recommend FindPython2 or FindPython3 instead. In master, we're already python3 minimum, so go ahead and switch this to the modern method.


lgritz:
This also seems to fully fix our recently broken CI on one of the Windows tests!

PR #4291, opened 2024-06-12
fix: Address ignored annotation nvcc warnings on explicitly-defaulted functions

chellmuth:
## Description

We're seeing nvcc compilation warnings like the following:
```
ustring.h(869): warning #20012-D: __device__ annotation is ignored on a function("operator=") that is explicitly defaulted on its first declaration
     __attribute__((host)) __attribute__((device)) 
```

According to nvidia's documentation (https://docs.nvidia.com/cuda/archive/10.1/cuda-c-programming-guide/index.html#compiler-generated-functions), functions that are explicitly-defaulted should not have execution space specifiers like \_\_device\_\_ or \_\_host\_\_. Instead the compiler will determine the annotations from the functions that invoke it.

This patch goes through the ustring and typedesc headers and removes OIIO_HOSTDEVICE from all defaulted functions.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/typedesc.h
+++ b/src/include/OpenImageIO/typedesc.h
@@ -164,8 +164,7 @@ struct OIIO_UTIL_API TypeDesc {
     TypeDesc (string_view typestring);
 
     /// Copy constructor.
-    OIIO_HOSTDEVICE constexpr TypeDesc (const TypeDesc &t) noexcept = default;
-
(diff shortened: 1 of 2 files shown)


lgritz (approved):
LGTM. Will merge when CI finishes.

lgritz:
> Instead the compiler will determine the annotations from the functions that invoke it.

WhY cAn'T aLl FuNcTiOnS dO tHiS, NVIDIA? Aaaaaah!

PR #4292, opened 2024-06-14
Simplify tile cache clearing.

curtisblack:
## Description

Simplifies the clearing of the tile cache so it no longer allocates a vector of tile IDs prior to erasing them all.

## Checklist:

  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -461,6 +461,18 @@ public:
             bin.unlock();
     }
 
+    /// Removes all items from the map.
+    void clear()
+    {
+        for (size_t b = 0; b < BINS; b++) {
+            m_bins[b].lock();
+            BinMap_t map;
+            std::swap(m_bins[b].map, map);
+            m_size -= map.size();
+            m_bins[b].unlock();
+        }
+    }
+
     /// Return true if the entire map is empty.
     bool empty() { return m_size == 0; }
 
--- a/src/include/OpenImageIO/unordered_map_concurrent.h
+++ b/src/include/OpenImageIO/unordered_map_concurrent.h
@@ -464,12 +464,17 @@ public:
     /// Removes all items from the map.
     void clear()
     {
+        if (empty())
+            return;
         for (size_t b = 0; b < BINS; b++) {
-            m_bins[b].lock();
             BinMap_t map;
-            std::swap(m_bins[b].map, map);
-            m_size -= map.size();
-            m_bins[b].unlock();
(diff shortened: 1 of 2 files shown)


lgritz:
What was the motivation, Curtis? Is there a performance issue you addressed with this?

curtisblack:
Hi Larry, admittedly the improvement here is quite small, but the motivation was to reduce peak memory in the case we wanted to clear a full image cache.

lgritz:
I asked because I wanted to make sure you had timed it and there's a real gain and not a loss.

The way it's structured the way it was is because of the locking. (It's up for debate whether I chose a strategy well, as I will explain below.)

The old way, the tile cache iterator is locking the shard it's iterating over. So my way is holding that lock only long enough to copy references to the tiles it will need to delete, then it moves on to the next shard. Of course, it also ends up calling erase_tile on each of those afterwards, which will lock those shards again -- that's a lot of locks, but each one is very brief. (As I think about it, I think I didn't realize how many locks that might entail.)

The new way, you are holding the lock on the shard for the time it takes you to clear that whole shard. That seems like a much longer time to hold a lock. But you only lock each shard once.

It's not obvious to me which is better in terms of the threading. I had thought my way would be better because any lock hold is very short, and you have longer holds. But now that I write the above explanation, I see that I have a lot more locks. So I'm inclined to think your way is better. But intuition about these things can get pretty tricky.

Or maybe it doesn't matter, because this is only about calling invalidate_all, which probably isn't going to happen in the midst of a lot of threads contending for the cache anyway (it's the kind of thing you probably do between frames).

Anyway, I think I do like your method better. I just wanted to prod you to spot-check the performance to make sure you don't notice a slowdown in practice. I think I was hoping you'd say "yeah, this really speeds it up!" But even if it's just neutral, I like what you did here better than my original code.

curtisblack:
Thanks for the explanation Larry, indeed it did skip my mind to test performance as I was more focused on memory usage.

The approach here should only need to hold the lock for the duration of swapping the internals of the map to the temporary map. This looks to be just a bunch of pointer swaps inside the robin_map, without actually moving items. Then after that the lock is unlocked, then the destructor of the temporary map runs, which is presumably where we want to be careful of long running item destructors.

Just to be sure, I did a quick test of performance. In a render which fills the cache with 2588 tiles, it used to take (2.9±0.3)ms, now it takes (2.5±0.1)ms. These are both the average of 16 runs.

lgritz:
Great, I feel better with the double-check.

lgritz (approved):
LGTM!

I think you need to get on the CLA for this project, Curtis. It's still flagging you as not having one.

PR #4293, opened 2024-06-14
Rudimentary CUDA support infrastructure

lgritz:
Build-time check for Cuda toolkit as optional dependency.  This can be explicitly disabled with OIIO_USE_CUDA=OFF (though currently, it defaults to off, you have to turn it on at build time because this is all still experimental).  When this is enabled at build time, C++ preprocessor symbol OIIO_USE_CUDA will be defined within the OIIO codebase.

OIIO global `attribute("gpu:device", "CUDA")` can be used to enable CUDA functionality, which at this point doesn't do anything other than initialize the CUDA device and find out information about it.  Several global `getattribute("cuda:*")` queries let you find out some information about any device found.

I'm trying to keep a lot of the approaches fairly generic, so that the interface won't need to change much to allow OpenCL or Metal or whatever else comes along, though for now I'm trying to flesh out the implementation mostly with CUDA.

oiiotool adds a new `--cuda` option, reserved for enabling CUDA functionality. Currently the only way this is used is to print information about the CUDA device in the `oiiotool --help` messages. An example of that is:

```
$ oiiotool --cuda --help
oiiotool -- simple image processing operations
...

OIIO 2.6.2.0spi | Linux/x86_64
    Build compiler: gcc 9.3 | C++17/201703
    CUDA 11.6.0 support enabled at build time
...
Running on 32 cores 187.3GB sse2,sse3,ssse3,sse41,sse42,avx,avx2,avx512f,
    avx512dq,avx512cd,avx512bw,avx512vl,fma,f16c,popcnt,rdrand
Compute hardware available: CUDA on Quadro RTX 6000, driver 12020,
    runtime 11060, compat 705, memory 23.6 GB
```

Note that at this point, we aren't actually *using* CUDA for anything. This is just setting up some basics for future expansion. It's all hidden in the pvt namespace for internal use only, and any of the functions or nomenclature may change as we continue to add functionality. I want to iterate on it a bit before it's in any way exposed in public headers or interfaces.

I'm a big fan of introducing even big strategic changes in the form of a series of MVP easy to review discrete steps. This step's only goals are: (a) allowing the build system to find and link against the CUDA Toolkit; (b) allowing runtime initialization and querying of a few basic facts about the device found; (c) start noodling around just a bit with function and attribute interfaces to gain a little experience that will guide what we eventually want.

LONG TERM goals of this initiative include:

* A full TextureSystem work-alike that can be used from CUDA (and possibly other compute device APIs) based renderers and with OSL's OptiX back and, having an analogous (ideally identical) interface and as close as possible to feature parity with CPU TextureSystem.

* Augment ImageBuf with the ability to store its images in GPU (or unified) memory and for IBA functions to be able to operate on those GPU-side buffers with GPU compute kernels, somewhat analogously to how libraries like PyTorch provides optional GPU support for tensor operations. (The hope is that this improves impage processing performance, though it's possible that it'll turn out that typical OIIO image processing workflows are so I/O dominated that there's not a lot of potential upside.  I think we'll have to try before we really find out if it's going to be worth it.)

* Allow image reading/writing plugins (or other parts of OIIO functionality) to leverage a GPU if available, and if there is something productive they can do with it.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -6160,8 +6160,8 @@ Oiiotool::getargs(int argc, char* argv[])
     ap.arg("--threads %d:N")
       .help("Number of threads (default 0 == #cores)")
       .OTACTION(set_threads);
-    ap.arg("--cuda")
-      .help("Enable CUDA if available")
+    ap.arg("--gpu")
+      .help("[EXPERIMENTAL] Use GPU if available (options: device=...)")
       .action([&](cspan<const char*>){
                   OIIO::attribute("gpu:device", "CUDA");
               });
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5931,17 +5931,24 @@ print_build_info(Oiiotool& ot, std::ostream& out)
 
     auto platform = format("OIIO {} | {}", OIIO_VERSION_STRING,
                            OIIO::get_string_attribute("build:platform"));
-    print("{}\n", Strutil::wordwrap(platform, columns, 4));
+    print(out, "{}\n", Strutil::wordwrap(platform, columns, 4));
 
     auto buildinfo = format("    Build compiler: {} | C++{}/{}",
                             OIIO::get_string_attribute("build:compiler"),
                             OIIO_CPLUSPLUS_VERSION, __cplusplus);
-    print("{}\n", Strutil::wordwrap(buildinfo, columns, 4));
+    print(out, "{}\n", Strutil::wordwrap(buildinfo, columns, 4));
 
     auto hwbuildfeats
         = format("    HW features enabled at build: {}",
                  OIIO::get_string_attribute("build:simd", "no SIMD"));
-    print("{}\n", Strutil::wordwrap(hwbuildfeats, columns, 4));
+    print(out, "{}\n", Strutil::wordwrap(hwbuildfeats, columns, 4));
+#ifdef OIIO_USE_CUDA
+    int cudaver = OIIO::get_int_attribute("cuda:build_version");
+    print(out, "    CUDA {}.{}.{} support enabled at build time\n",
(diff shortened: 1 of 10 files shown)


src/oiiotool/oiiotool.cpp
```diff
@@ -6137,6 +6160,11 @@ Oiiotool::getargs(int argc, char* argv[])
     ap.arg("--threads %d:N")
       .help("Number of threads (default 0 == #cores)")
       .OTACTION(set_threads);
+    ap.arg("--cuda")
```
ld-kerley, line 6163: perhaps more generically forward looking `--gpu=cuda`? allowing for `--gpu=metal` in the future without an explosion of arguments.

lgritz: Yes, you're right.

I also had a question in my mind about the "gpu" terminology. What I really am trying to express is "compute device", but I'm using gpu as a very short stand-in unless somebody can suggest something better.

lgritz: Updated, with change to --gpu.
For now, it does the same thing as before, but in the future we'll expand it to have a way to choose what kind of device to use if there is more than supported and one available.

lgritz: We'll build this all up step by step.

jessey-git:
Do any of the existing CI runners have CUDA installed?

lgritz:
Yes, all of the ones using ASWF containers have CUDA, so they're building against it.
They're not finding GPUs at runtime, though, so by the time we're executing CUDA code, we'll want to get set up using the paid runners with GPUs. I'm long overdue getting that set up for OSL, as well, so will probably kill two birds with one stone there.

lgritz:
Does this seem ok to everybody as a first step, and then we'll keep building on top of it?

jessey-git (approved):
Looks alright I think.  I'm not entirely setup to test this out myself locally but seems good enough to build off of.

PR #4294, opened 2024-06-16
build: Enhacements to the dependency report

lgritz:
* List external dependencies found, as well as those not found.
* Sort and deduplicate the lists before printing.
* Print versions of everything


lgritz:
Here's what the dependency report looks like now with this PR:

Screenshot 2024-06-19 at 2 23 53 PM

PR #4296, opened 2024-06-16
build: raise libtiff min to 4.0 + self-build for libtiff (and libdeflate)

lgritz:


lgritz:
No objections -> merging

PR #4297, opened 2024-06-16
build: pybind11 raise min to 2.7 and make self-builder

lgritz:


lgritz:
Any concerns? Is anybody under constraints to use pybind11 older than 2.7, which was released back in 2021?

And apologies, the original headline on this PR said "2.12", but it was wrong -- the minimum is being raised to 2.7, but 2.12 is the version we'll build automatically if requested and it's not found externally.

PR #4302, opened 2024-06-19
Fix 4298: various PSD files fail to load correctly

### LINKED ISSUE #4298: [BUG] 16 and 32 bit PSD files fail to load correctly
**Describe the bug**

It seems as though the recent PSD changes are causing loading issues for 16 and 32bit files.

version: `2.5.11.0`
compiler: `MSVS 1929`
simd: `sse2`
dependencies: `Boost 1.82.0;fmt 9.1.0;JPEG 80;libjpeg-turbo NONE;OpenEXR 3.2.4;OpenJPEG 2.5;PNG 1.6.37;pugixml 1.10;pybind11 2.10.1;Python 3.11.7;Robinmap;TBB 2020.3;TIFF 4.6.0;WebP 1.3.2;ZLIB 1.2.13`

**To Reproduce**

1. Try loading https://github.com/AcademySoftwareFoundation/OpenImageIO-images/blob/master/psd_rgb_16.psd or https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/testsuite/psd/src/Layers_32bit_RGB.psd

**Evidence**

Result from https://github.com/AcademySoftwareFoundation/OpenImageIO-images/blob/master/psd_rgb_16.psd
bad16

Result from https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/testsuite/psd/src/Layers_32bit_RGB.psd
bad32

jessey-git:
## Description

The primary issue seemed to be that the ChannelInfo width and height were not filled in for some of the layers, notably the merged image data layer in most(all?) cases.

A second issue was that there was incorrect nesting of the switch statement in `read_channel_row`. The last 3 case statements were all nested under the `if (!bigendian()) {` conditional from the 1st case.

Fixes #4298

## Tests

Use `iconvert` on all PSD files contained in the attached .zip.  Converting to both .exr and .png is a good test to ensure the data is flowing out to those formats correctly.
[psd.zip](https://github.com/user-attachments/files/15905814/psd.zip)

## Checklist:
  nonconforming.


jessey-git:
Having issues figuring out what's going on with the `psd-colormodes` test. The CI output is pretty confusing and the tests do not work natively in Windows so it's been difficult to repro locally.

CC @EmilDohne in case you can take a look soon.

EmilDohne:
Hey thanks for making a pr for this. I am sorry for not having taken a look myself yet as Im quite busy but this should speed things up :)

EmilDohne:
Hey @jessey-git, these changes look good to me. I was able to confirm that adding the width and height to the `load_image_data()` function made the reads correct. Although I am not sure why exactly this would lead to the weird result.

The changes to the incorrect nesting with `if(!bigendian())` also appear to have been an oversight on my side but doesnt appear to be the cause of any issues in the test suite.

As for the failing test cases I am unsure if I can be of much help since I also struggled with this last time and had @lgritz help me out which I fear will have to be the case here too. 

Thanks a lot for the contribution :)

lgritz:
Let me take a stab at the reference image issues. Stay tuned...

lgritz:
I think the problem was that the reference images were wrong -- checked in from before the 16 bit reading was done correctly. I double checked against the original PSDs and updated the references appropriately. Let's see if this CI run succeeds.

EmilDohne:
Thanks a lot for looking into that!

lgritz (approved):
LGTM

PR #4304, opened 2024-06-21
int: Simplify openexr includes

lgritz:
In master, since we have raised the minimum Imath/OpenEXR to 3.x, no more 2.x, we no longer need generate our Imath.h and half.h from templates, they can be turned into regular header files.

Additionally, define IMATH_HALF_NO_LOOKUP_TABLE in half.h before including Imath/half.h, which ensures that the Imath code we include converts float <-> half using actual inline code and not the lookup table. This can reduce or sometimes eliminate the need to link agains Imath only for the sake of accessing that lookup table.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,8 +20,8 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * Compilers: **gcc 9.3** - 14.1, **clang 5** - 18, MSVS 2017 - 2019 (**v19.14
    and up**), **Intel icc 19+**, Intel OneAPI C++ compiler 2022+.
  * **CMake >= 3.15** (tested through 3.29)
- * **OpenEXR/Imath >= 3.1** (tested through 3.2
-   and main) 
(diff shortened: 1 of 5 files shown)


fpsunflower:
A side question that I guess is more about Imath itself -- why is `IMATH_HALF_NO_LOOKUP_TABLE` even necessary? Wouldn't most people prefer to have the inline implementation that doesn't require linking?


What happens if someone includes the half.h header without that define before they include the OIIO headers?

lgritz:
> A side question that I guess is more about Imath itself -- why is `IMATH_HALF_NO_LOOKUP_TABLE` even necessary?

It may have been an escape valve when the inline implementation was added, in case it was found to be broken or had some case where it was not performant enough. It's possible that it's no longer needed or should be the default.


> What happens if someone includes the half.h header without that define before they include the OIIO headers?

They would get the table. But if they're using half.h in their code for some reason, they are already linking (or should be) against libImath. That's on them. What this is designed to address is the case where they don't even know that the half conversions are used by OIIO internally, and thus are unaware that they might need to link libImath at all.

PR #4307, opened 2024-06-22
ci: As a favor to OpenEXR, test their experimental core library changes

lgritz:


lgritz:
Unfortunately, this reveals that the exr changes are broken! Which I guess is the point. But now I'm conflicted about whether I should commit this so we know at every stage if the impending changes to openexr internals are working (but at the inconvenience of our CI failing that test until it's fixed), or if I should hold this PR until OpenEXR is fixed (keeps our CI passing, but harder to verify that it's been fixed without launching those CI jobs by hand periodically).

PR #4308, opened 2024-06-23
cleanup: simd.h deprecations

lgritz:
Fully remove some things that have had deprecation warnings for a long, long time (mostly since 1.x), so we think it can't possibly break anyone.

Add deprecation warnings to some things that have been nominally deprecated for some time. The build warnings will serve as final notice that they will fully disappear by OIIO 3.0 release.

This is part of a series of small steps to fully deprecate a whole lot of cruft between now and the 3.0 release.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -4980,7 +4980,7 @@ OIIO_FORCEINLINE vint4 rotl(const vint4& x, int s) {
 }
 
 // DEPRECATED (2.1)
-OIIO_DEPRECATED("Use rotl32 instead")
+OIIO_DEPRECATED("Use rotl instead")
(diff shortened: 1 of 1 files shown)


src/include/OpenImageIO/simd.h
```diff
@@ -5003,6 +4980,7 @@ OIIO_FORCEINLINE vint4 rotl(const vint4& x, int s) {
 }
 
 // DEPRECATED (2.1)
+OIIO_DEPRECATED("Use rotl32 instead")
```
ThiagoIze, line 4983: Typo? Should this be `rotl` instead of `rotl32`?
```suggestion
OIIO_DEPRECATED("Use rotl instead")
```

lgritz: quite right, thanks

ThiagoIze (approved):

PR #4310, opened 2024-06-23
Initial Exif and XML metadata support for JPEG XL

ssh4net:
## Description

Initial Exif and XML metadata support for JPEG XL.
EXIF - works and enabled by default
XML - works and enabled by default
IPTC - that part need some help. JPEG XL support XML based IPTC, but I do not get what IPTC create an OIIO
JBIG - not sure

new options:
`jpegxl:use_boxes` - export EXIF, XMP, IPTC, JBIG
`jpegxl:compress_boxes` - compress meta boxes (disabled for this moment by default)

jpegxl:exif_box = 1 (enabled)
jpegxl:xmp_box = 1 (enabled)
jpegxl:jumb_box = 0 (disabled)
jpegxl:iptc_box = 0 (disabled)

## Tests

Not yet, but if we have I can add.

## Checklist:

  nonconforming.


src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -365,6 +367,110 @@ JxlOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
...
+        if (!xmp.empty()) {
+            xmp_data.insert(xmp_data.end(), xmp.c_str(),
+                            xmp.c_str() + xmp.length());
+        }
+
+        // IPTC
+        std::vector<uint8_t>* iptc_data = nullptr;
+        if (!iptc.empty()) {
+            static char photoshop[] = "Photoshop 3.0";
+            std::vector<char> head(photoshop,
```
1div0, line 416: Wouldn't be better to use std::vector<uint8_t> ?

ssh4net: I can't.
everything from OpenImageIO like encode_iptc_iim(m_spec, iptc); use std::vector<char>

1div0: I was not precise enough.
I meant to use std::vector<uint8_t> instead of std::vector<char>

1div0: Understood. Yes, especially for the UTF-8 metadata handling it would be better choice.

lgritz:
You're getting build failures on Mac because of an unused variable warning.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -365,6 +367,111 @@ JxlOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
...
+
+    // Write IPTC IIM metadata tags, if we have anything
+    std::vector<char> iptc;
+    encode_iptc_iim(m_spec, iptc);
+
+    bool use_boxes = m_spec.get_int_attribute("jpegxl:use_boxes", 1) == 1;
+    int compress_boxes
+        = m_spec.get_int_attribute("jpegxl:compress_boxes", 0) == 0
+              ? 0
+              : 1;  // TODO: change to 1 for compression by default
```
lgritz, line 390: This seems a little harder to follow then necessary (if it equals zero the condition is true, but then in that case set to 0).  Maybe I would suggest the simpler


```suggestion
    int compress_boxes = m_spec.get_int_attribute("jpegxl:compress_boxes") ? 1 : 0;
    // TODO: change to 1 for compression by default
```

Or even change it to bool like the one above?

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -365,6 +367,111 @@ JxlOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
...
+        }
+
+        // Jumbf
+        std::vector<uint8_t>* jumbf_data = nullptr;
+
+        const struct BoxInfo {
+            const char* type;
+            const std::vector<uint8_t>* bytes;
+            const bool enable;
+        } boxes[] = { { "Exif", exif_data,
```
lgritz, line 440:
```suggestion
        struct BoxInfo {
            const char* type;
            const std::vector<uint8_t>* bytes;
            const bool enable;
        };
        
        const BoxInfo boxes[] = { { "Exif", exif_data,
```

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -412,6 +519,9 @@ JxlOutput::save_image(const void* data)
 
     DBG std::cout << "data = " << data << " size = " << size << "\n";
 
+    // Write EXIF info
+    bool metadata_success = save_metadata(m_spec, m_encoder);
```
lgritz, line 523: Clang gives a warning about the fact that you don't ever use `metadata_success`. Should you actually handle any errors preparing the metadata?

lgritz (changes requested):
I made a couple very minor style suggestions for clarity.

Also, there is a warning failure for clang related to not using the return value of save_metadata.

But other than those, LGTM!

Can you incorporate someting into the jxl tests so that the testsuite verifies writing and reading of metadata?

src/jpegxl.imageio/jxloutput.cpp
ssh4net: Hm. I'll add something.

ssh4net:
@lgritz I'll check how can add something to tests. But at this moment only writing was added.

I also need change a docs a bit

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -365,6 +367,111 @@ JxlOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
...
+            const std::vector<uint8_t>* bytes;
+            const bool enable;
+        } boxes[] = { { "Exif", exif_data,
+                        m_spec.get_int_attribute("jpegxl:exif_box", 1) == 1 },
+                      { "xml ", &xmp_data,
+                        m_spec.get_int_attribute("jpegxl:xmp_box", 1) == 1 },
+                      { "jumb", jumbf_data,
+                        m_spec.get_int_attribute("jpegxl:jumb_box", 0) == 1 },
+                      { "xml ", iptc_data,
+                        m_spec.get_int_attribute("jpegxl:iptc_box", 0) == 1 } };
```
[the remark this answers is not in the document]
ssh4net: As i understand OIIO generating XML based iptc. But not sure. Box encoded correctly but other tools do not see it.

ssh4net:
@lgritz don't know how to do a tests without implementing exif for import.
I'll probably add jpegxl import exif support, but later. Need to work on other project at this moment.

lgritz:
> @lgritz don't know how to do a tests without implementing exif for import.

No prob, we can just remember to add tests when the input support is added.

lgritz:
@ssh4net This LGTM except for the CI failure that appears to not like 'const' in the struct definition for BoxInfo.

Also, would you mind rebasing on top of the current main just to be sure there aren't any conflicts?

After those two items, I'm good to merge this.

src/jpegxl.imageio/jxloutput.cpp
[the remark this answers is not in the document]
ssh4net: Well, this code with boxes stright from JPEG XL cjxl. And as far as I understand there are two IPTC formats, one obsolete (probably binary) and a newer XML based.

Just as most standardised formats documentation is quite complicated to understand.

lgritz (approved):
LGTM, thanks!

PR #4317, opened 2024-07-01
api: add memsize() methods to ParamValueList and ImageSpec

lgritz:
To aid some memory accounting, add a method to ParamValueList and
ImageSpec that return the memory cost (the size of the struct and all
the allocated things hanging off of it). We're trying to be more
careful about understanding how much memory is used by having many
copies of ImageSpec's lying around.

Have ImageCache stats estimate how much mem is tied up in ImageSpecs.


lgritz:
Withdrawing in favor of #4322

PR #4320, opened 2024-07-03
make_texture: only write full metadata to the first mip level

lgritz:
It's silly to copy the full metadata from the source (which could include big things like ICC profiles) to EVERY level of a MIPmap.

This mostly applies to TIFF textures, since OpenEXR MIPmaps don't store attributes separately for each MIP level anyway.

This can be disabled (that is, make it copy full metadata to all levels of the MIPmap) by setting a new make_texture configuration hint, "maketx:mipmap_metadata" to nonzero.

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -758,7 +758,9 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
                 mipimages.erase(mipimages.begin());
             } else {
                 // Resize a factor of two smaller
-                smallspec        = outspec;
+                smallspec = outspec;
+                if (!configspec.get_int_attribute("maketx:mipmap_metadata"))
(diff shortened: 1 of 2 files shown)


src/libOpenImageIO/maketexture.cpp
```diff
@@ -758,7 +758,9 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
                 mipimages.erase(mipimages.begin());
             } else {
                 // Resize a factor of two smaller
-                smallspec        = outspec;
+                smallspec = outspec;
+                if (!configspec.get_int_attribute("maketx:allow_pixel_shift"))
```
curtisblack, line 762: Looks like a copy/paste typo?
```suggestion
                if (!configspec.get_int_attribute("maketx:mipmap_metadata"))
```

lgritz: I'm not quite sure how I did that, sure does look like c/p problem. Thanks for spotting it.

PR #4322, opened 2024-07-04
api: add pvt::heapsize() and pvt::footprint() methods and image cache memory tracking

bfraboni:
## Description

First PR of two, following @lgritz and I discussions on memory tracking in the OIIO::ImageCache.
- Add two template methods and their specializations for types various to help memory tracking:
```c++
// return the total heap allocated memory held by the object and its members
template<typename T> inline size_t heapsize(const T& t);
// return the total memory footprint including the size of the structure itself
template<typename T> inline size_t footprint(const T& t);
```
- Specialized for: ParamValue, ParamValueList, ImageSpec, ImageInput, ImageOutput, std::vector, std::shared_ptr, std::unique_ptr, oiio::intrusive_ptr, ImageCacheImpl related objects.
- New files: 
   - `include/memory.h` : adds base and few types specialization of heapsize and footprint.
   - `libtexture/imagecache_memory_pvt.h` :  adds specilializations for ImageCacheImpl and related objects.
   - `libtexture/imagecache_memory_print.h` : adds a helper function that print a breakdown of the ImageCacheImpl total memory usage as well, as well as per image format. Note: this is slow, but gives accurate memory estimation.  

Related PR from Larry : https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4317

## Tests
- [ ] todo: fix CI fails

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -36,6 +36,7 @@
 #include <OpenImageIO/strutil.h>
 #include <OpenImageIO/thread.h>
 #include <OpenImageIO/typedesc.h>
+#include <OpenImageIO/memory.h>
 
 OIIO_NAMESPACE_BEGIN
 
@@ -1884,6 +1885,9 @@ private:
     std::unique_ptr<Impl, decltype(&impl_deleter)> m_impl;
 
     void append_error(string_view message) const; // add to error message
+
+    /// declare a friend heapsize definition
+    template <typename T> friend size_t pvt::heapsize(const T&);
 };
 
 
@@ -2788,7 +2792,23 @@ private:
     std::unique_ptr<Impl, decltype(&impl_deleter)> m_impl;
 
     void append_error(string_view message) const; // add to m_errmessage
+
+    /// declare a friend heapsize definition
+    template <typename T> friend size_t pvt::heapsize(const T&);
 };
+
+
+
+/// Memory tracking. Specializes the base memory tracking functions from memory.h.
+
+// heapsize specialization for `ImageSpec`
+template <> OIIO_API size_t pvt::heapsize<ImageSpec>(const ImageSpec&);
+
+// heapsize specialization for `ImageInput`
+template <> OIIO_API size_t pvt::heapsize<ImageInput>(const ImageInput&);
+
+// heapsize specialization for `ImageOutput`
+template <> OIIO_API size_t pvt::heapsize<ImageOutput>(const ImageOutput&);
 
 
 
--- a/src/include/OpenImageIO/memory.h
+++ b/src/include/OpenImageIO/memory.h
@@ -1,0 +1,117 @@
+// Copyright Contributors to the OpenImageIO project.
+// SPDX-License-Identifier: Apache-2.0
+// https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+
+/////////////////////////////////////////////////////////////////////////
+/// @file   memory.h
+///
+/// @brief  Utilities for memory tracking.
+/////////////////////////////////////////////////////////////////////////
+
+
+#pragma once
+
+#define OPENIMAGEIO_MEMORY_H
+
+#include <cstring>
+#include <memory>
+#include <vector>
+
+OIIO_NAMESPACE_BEGIN
+
+namespace pvt {
+
+/// Return the total heap memory allocated by `object`.
+/// The template specialization can be used to give improved results for non trivial types
+/// that perform heap allocation, and to include members allocations recursively.
+template<typename T>
+inline size_t
+heapsize(const T& t)
+{
+    return 0;
+}
+
+/// Return the total memory footprint of `object`. If possible, including any heap
+/// allocations done by any constituent parts. The default implementation just reduces
+/// to sizeof(object), given that heapsize(object) would return 0.
+/// The template specialization can be used to give improved results for non trivial types
+/// that perform heap allocation.
+template<typename T>
+inline size_t
+footprint(const T& t)
+{
+    return sizeof(T) + heapsize(t);
+}
+
+template<typename T>
+inline size_t
+footprint(const T* t)
+{
+    return sizeof(T) + (t ? footprint(*t) : 0);
+}
+
+/// Specializations for common STL types
+
+
+// heapsize specialization for std::string
+template<>
+inline size_t
+heapsize<std::string>(const std::string& s)
+{
+    // accounts for small string optimization that does not
+    // use any heap allocations
+    const char* const sbegin = (const char*)&s;
+    const char* const send   = sbegin + sizeof(std::string);
+    const char* const sdata  = s.data();
+    const bool is_small      = sdata >= sbegin && sdata < send;
+    return is_small ? 0 : s.capacity();
+}
+
+// heapsize specialization for std::vector
+template<typename T>
+inline size_t
+heapsize(const std::vector<T>& vec)
+{
+    size_t size = 0;
+    for (const T& elem : vec)
+        size += footprint(elem);
+    return size;
+}
+
+// heapsize specialization for std::shared_ptr
+template<typename T>
+inline size_t
+heapsize(const std::shared_ptr<T>& ref)
+{
+    return ref ? footprint(*ref.get()) : 0;
+}
+
+// footprint specialization for std::shared_ptr
+template<typename T>
+inline size_t
+footprint(const std::shared_ptr<T>& ref)
+{
+    return sizeof(std::shared_ptr<T>) + heapsize(ref);
+}
+
+// heapsize specialization for std::unique_ptr
+template<typename T>
+inline size_t
+heapsize(const std::unique_ptr<T>& ref)
+{
+    return ref ? footprint(*ref.get()) : 0;
+}
+
+// footprint specialization for std::unique_ptr
+template<typename T>
+inline size_t
+footprint(const std::unique_ptr<T>& ref)
+{
+    return sizeof(std::unique_ptr<T>) + heapsize(ref);
+}
+
+}  // namespace pvt
+
+
+OIIO_NAMESPACE_END
--- a/src/include/OpenImageIO/paramlist.h
+++ b/src/include/OpenImageIO/paramlist.h
@@ -16,6 +16,7 @@
 
 #include <OpenImageIO/attrdelegate.h>
 #include <OpenImageIO/export.h>
+#include <OpenImageIO/memory.h>
 #include <OpenImageIO/strongparam.h>
 #include <OpenImageIO/typedesc.h>
 #include <OpenImageIO/ustring.h>
@@ -276,9 +277,15 @@ private:
                       Copy _copy                = Copy(true),
                       FromUstring _from_ustring = FromUstring(false)) noexcept;
     void clear_value() noexcept;
+
+    /// declare a friend heapsize definition
+    template<typename T> friend size_t pvt::heapsize(const T&);
 };
 
-
+/// heapsize specialization for `ParamValue`
+template<>
+OIIO_API size_t
+pvt::heapsize<ParamValue>(const ParamValue&);
 
 /// Factory for a ParamValue that holds a single value of any type supported
 /// by a corresponding ParamValue constructor (such as int, float, string).
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -16,6 +16,7 @@
 #include <OpenImageIO/export.h>
 #include <OpenImageIO/hash.h>
 #include <OpenImageIO/imagebuf.h>
+#include <OpenImageIO/memory.h>
 #include <OpenImageIO/refcnt.h>
 #include <OpenImageIO/texture.h>
 #include <OpenImageIO/timer.h>
@@ -42,6 +43,7 @@ namespace pvt {
 struct TileID;
 class ImageCacheImpl;
 class ImageCachePerThreadInfo;
+struct ImageCacheFootprint;
 
 const char*
 texture_format_name(TexFormat f);
@@ -466,6 +468,12 @@ private:
     friend class ImageCacheImpl;
     friend class TextureSystemImpl;
     friend struct SubimageInfo;
+
+    /// Memory tracking. Specializes the base memory tracking functions from memory.h.
+    /// declare a friend heapsize definition
+    template<typename T> friend inline size_t heapsize(const T&);
+    /// declare a friend image cache footprint definition
+    friend inline size_t footprint(const ImageCacheImpl&, ImageCacheFootprint&);
 };
 
 
@@ -1197,7 +1205,9 @@ private:
     spin_mutex m_fingerprints_mutex;  ///< Protect m_fingerprints
     FingerprintMap m_fingerprints;    ///< Map fingerprints to files
 
-    TileCache m_tilecache;          ///< Our in-memory tile cache
+    /// FIXME: if unordered_map_concurrent had const iterators,
+    /// m_tilecache wouldn't need to be mutable
+    mutable TileCache m_tilecache;  ///< Our in-memory tile cache
     TileID m_tile_sweep_id;         ///< Sweeper for "clock" paging algorithm
     spin_mutex m_tile_sweep_mutex;  ///< Ensure only one in check_max_mem
 
@@ -1236,6 +1246,11 @@ private:
             // done it with nobody else interfering.
         } while (llstat->compare_exchange_strong(*llnewval, *lloldval));
     }
+
+    /// declare a friend heapsize definition
+    template<typename T> friend inline size_t heapsize(const T&);
+    /// declare a friend image cache footprint definition
+    friend inline size_t footprint(const ImageCacheImpl&, ImageCacheFootprint&);
 };
 
 
--- a/src/libutil/paramlist.cpp
+++ b/src/libutil/paramlist.cpp
@@ -383,6 +383,17 @@ ParamValue::clear_value() noexcept
 
 
 
+template<>
+size_t
+pvt::heapsize<ParamValue>(const ParamValue& pv)
+{
+    return (pv.m_nonlocal && pv.m_copy)
+               ? pv.m_nvalues * static_cast<int>(pv.m_type.size())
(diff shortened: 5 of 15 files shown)


lgritz:
I have an alternate design I'd like to suggest, that I think might have some benefits.

We start with the fully general template declaration that from a caller's point of view we use for everything:

```
/// Return the total memory footprint of `object`. If possible, including any heap
/// allocations done by any constituent parts. The default implementation just reducees
/// to sizeof(object), but template specialization can be used to give improved results
/// for tricky types of those that perform heap allocation.
template<typename T>
size_t footprint(const T& object) noexcept
{
    // The base case / fallback is equivalent to sizeof(T)
    return sizeof(object);
}
```

Everything else is defined by specialization:

```
// Specialization that does std::string correctly
template<>
size_t footprint(const std::string& str) noexcept
{
    size_t size = sizeof(str);
    if (!is_short_string(str) /* implementation TBD */)
        size += str.capacity();
    return size;
}

// Specialization that does vectors correctly
template<typename T>
size_t footprint(const std::vector<T>& vec) noexcept
{
    size_t size = sizeof(vec);
    for (const auto& e : vec)
        size += footprint(e);
    size += (vec.capacity() - vec.size()) * sizeof(T);
    return size;
}

// Specialization that does arrays correctly even for nontrivial types
template<typename T, size_t N>
size_t footprint(const T (&arr)[N]) noexcept
{
    size_t size = 0;
    for (const auto& e : arr)
        size += footprint(e);
    return size;
}
```

we will think of more specializations and add as needed. 

Then we add for our own types:

```
// in paramlist.h:
template<> size_t footprint(const ParamValueList& pl);

// in paramlist.cpp:
template<> size_t footprint(const ParamValueList& pl)
{
    ... implementation ...
}
```

We don't really need ParamValueList::footprint(), though if we really want, we could add the method and either define it as a call to the template, or define the template to call the method, whatever we prefer.

But this also would allow us in the future to easily add footprint ability to classes that maybe we couldn't add it as a method without breaking ABIs, because it would actually be outside the class. And then we could add the convenience method later, when it's a point that we are comfortable breaking interfaces.

By defining it to look like a single uniform template, that means that anywhere in the code, we could just ask for `footprint(thing)` and it would just work -- you don't need to pick one of several names depending on the type of data structure that thing is. You don't even really need to know if there's a footprint method on the object, because it will always fall back to at least giving you the size of the base object itself, even if it doesn't fully account for any dynamic allocations hanging off of it (and you can add the specialization later).

src/include/OpenImageIO/memory.h
```diff
@@ -0,0 +1,80 @@
...
+template<class VecType>
+inline size_t
+vector_string_datasize(const VecType& v)
+{
+    using T = typename VecType::value_type;
+    // account for the size of elements
+    size_t size = vector_bytes(v);
+    // account for the size of allocated data
+    for (const T& ref : v)
+        size += ref.capacity();
```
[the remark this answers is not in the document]
bfraboni: Yeah I've seen the comment in your PR afterwards, but is there anyway to test whether it is large / small at runtime ? 
Something like: `size += ref.capacity() > sizeof(std::string-2) ? 0 : ref.capacity();` ?

I have still a lot of string allocations reported by my heaptrack reports from OIIO and OSL and because I'm currently not sure if they matter or not I'd like to consider the actual. Also this is generic memory tracking for vectors of strings and it is not only used for channel names, but also for vectors of paths in the second PR.

src/include/OpenImageIO/memory.h
```diff
@@ -0,0 +1,80 @@
...
+/// containers.
+///
+/////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#define OPENIMAGEIO_MEMORY_H
+
+OIIO_NAMESPACE_BEGIN
+
```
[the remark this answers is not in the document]
bfraboni: Good suggestion, I didn't know where to put that !
I've first implemented a templated version but removed it because of the "if constexpr is a c++17 extension" warnings.

src/include/OpenImageIO/paramlist.h
```diff
@@ -306,6 +302,12 @@ class OIIO_UTIL_API ParamValue {
     ustring get_ustring(int maxsize = 64) const;
     ustring get_ustring_indexed(int index) const;
 
+    /// Return the total heap memory allocated by the ParamValue.
+    size_t datasize() const noexcept;
+    /// Return the total memory used by the ParamValue, including this
+    /// object itself and any heap memory allocated data.
+    size_t footprint() const noexcept;
+
```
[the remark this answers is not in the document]
bfraboni: Yes, I still feel that decoupling datasize() and footprint is worth it for readbility.
- footprint() will always be `sizeof(Object) + datasize()`, 
- datasize() will only accumulate members datasize() and allocations.

Whenever Object A is part of the footprint of Object B, you only want to account for A.datasize() in the footprint() of Object B to not double count. This is more visible in PR #4323 !

bfraboni:
I like that template suggestion @lgritz , I can do that but keep datasize() as follows:
```c++
template<typename T>
size_t datasize(const T& object) noexcept
{
    // The base datasize() is always 0
    return 0;
}

/// 
/// specialize datasize() for every required object
/// 

// Specialization that does objects correctly
template<typename T>
size_t footprint(const T& object) noexcept
{
    // The base footprint() is always sizeof(T)
    return sizeof(T) + datasize<T>(object);
}

// Specialization that does pointers correctly
template<typename T>
size_t footprint(const T* ptr) noexcept
{
    // The base footprint() is always sizeof(T*)
    return sizeof(T*) + (ptr ? footprint<T>(*ptr) : 0);
}
```

I use the vector_datasize variants for std::vector<T> and std::vector<T*> a lot, and never used a "footprint" counterpart, especially in the other PR.
And what you suggest `size_t size = sizeof(vec);` would double count the sizeof vector if it is a member variable of an object, so I still feel it is worth having datasize() for every objects and  vector / array types, populate the datasize() correctly using the members datasize() and having a trivial footprint() function.

Would that make sense ?

src/include/OpenImageIO/memory.h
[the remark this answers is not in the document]
lgritz: I know how to do it. Will post a PR that adds it to strutil.h.

src/include/OpenImageIO/memory.h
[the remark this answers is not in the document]
lgritz: We have C++17 minimum.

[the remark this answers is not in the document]
lgritz: In master, I mean.

I think if there's part of this that simply must be backported to 2.5, then we can use enable_if tricks rather than `if constexpr` for those.

src/libutil/paramlist.cpp
```diff
@@ -383,6 +384,22 @@ ParamValue::clear_value() noexcept
 
 
 
+size_t
+ParamValue::datasize() const noexcept
+{
+    return m_nvalues * static_cast<int>(m_type.size());
+}
```
lgritz, line 391: Note that this doesn't actually match the description you gave in the declaration -- this isn't necessarily heap memory.

src/include/OpenImageIO/paramlist.h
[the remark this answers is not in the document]
lgritz: I'm not against datasize per se, I was just thinking that if it doesn't serve a distinct purpose, maybe we don't need it?

I think we all know what footprint is -- it's the total memory cost of an object and everything it owns, as near as we can calculate it.

I'm not sure we have a rigorous definition of what datasize really is, though. What does it mean for something like a string (and also ParamValue) that sometimes stores its data within the object itself (so is counted in sizeof(obj)) and sometimes has heap allocations? And what does it mean if the valid data takes up less than the capacity?

There is already a ParamValue::datasize() but it means "how big is the data that this represents", so you could do
```
    memcpy (mybuffer, pv.data(), pv.datasize());
```
It doesn't measure the heap allocation (if it did, it would return 0 for the cases where no heap allocation is needed), nor does it include the unused capacity.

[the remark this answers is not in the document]
bfraboni: Sorry about the mislead, we'd want datasize() to be anything not part of sizeof(T), if that only contains heap allocations we can rename that heapsize() if that makes more sense, but it because it recursively accumulate members datasize() it is sort of indirect allocations.

lgritz:
> I like that template suggestion @lgritz , I can do that but keep datasize() as follows:

Yeah, I'm ok with that. But let's change it to heapsize() or allocsize() or something more clear so that it (a) doesn't conflict with the existing `datasize()` name, which we already use in ParamValue and that represents the... data size... and not the allocation size; and (b) is more self-documenting about what it actually means, and has a rigorous definition to go along with it. That definition should be clear about how it should behave in "short string" or "short vector" (which we don't use yet, but I'd like to in the future) situations. 

May I suggest the following definitions?

foo::datasize() -- how big the valid data is, having nothing to do with size or location of allocation. That is, it's how big you'd need to malloc or memcpy to get the relevant data representation. Only some types have/need this.

footprint(obj) -- as best as we can determine, the TOTAL memory footprint of object and everything obj owns. Should be usable with any object, ours or std.

heapallocsize(obj) -- total size of any heap allocation within obj. (Is it recursive, does it include heap allocs of things contained in things that obj allocated?) I think that this in general should be equal to footprint(obj) - sizeof(obj), but I'm not certain I understand a consistent definition for the recursive parts.

lgritz:
I'm still not sure I know what datasize() (or heapsize or whatever) should return in all cases. If obj contains a std::string, should datasize(obj) count the heap allocations of the string, also? What about recursive objects?  Let me make an example:

```
struct A {  // pretend sizeof(A) == 32+32+4
    std::string s;    // might have a heap alloc, if the string is big
    std::vector<char> vc;  // has a heap alloc if not empty
    int i;
};

struct B {   // pretend sizeof(B) = 4+32+sizeof(A)
    float f;
    std::string s;  // might have a heap alloc if big
    A a;
};

B b;
```

I think we agree that 

footprint(b) = sizeof(B) + b.s.heapsize + b.a.s.heapsize + b.a.vc+heapsize

but what is heapallocsize(b) (or, in your terms, datasize(b))?

bfraboni:
I'd write the duo footprint / heapallocsize the following way:
```c++
struct A {  // pretend sizeof(A) == 32+32+4
    std::string s;    // might have a heap alloc, if the string is big
    std::vector<char> vc;  // has a heap alloc if not empty
    int i;
};

struct B {   // pretend sizeof(B) = 4+32+sizeof(A)
    float f; // trivial
    std::string s;  // might have a heap alloc if big
    A a; // might have heap alloc
};

// default heapallocsize
template <typename T> size_t heapallocsize(const T& ref) { return 0; }

// heapallocsize specialization for std::string
template <> size_t heapallocsize<std::string>(const std::string& s) { return s.is_small ?  0 : s.capacity(); }

// default footprint for object or ptr to object
template <typename T> size_t footprint(const T& t)
{ 
    return std::is_pointer(T)::value ? 
        sizeof(T) + heapallocsize(t) : 
        sizeof(T) + footprint(*t); 
}

// heapallocsize specialization for std::vector
template <typename T> size_t heapallocisze(const std::vector<T>& v)
{ 
    size_t size = 0;
    for (const auto& e : vec)
        size += footprint(e);
    return size;
}

// heapallocsize specialization for A
// - counts the heapallocsize of all member that potentially alloc on the heap, so the string and the vector
// - the rest is automatically accounted for in "footprint" 
template <> size_t heapallocsize<A>(const A& a) { return heapallocsize(a.s) + heapallocsize(a.vc); }

// heapallocsize specialization for B
// - counts the heapallocsize of all member that potentially alloc on the heap, so the string and the structure A
// - the rest is automatically accounted for in "footprint" 
template <> size_t heapallocsize<B>(const B& b) { return heapallocsize(b.s) +heapallocsize(b.a) }; 
```

The only work to do is for every structure we want to be able to track, list the internal members that have potential heap allocations and specialize heapallocsize !

I agree with all definitions, and sorry about the confusion, I didn't count the allocation of ParamValues correctly and I misused datasize(); will fix that too..

src/libutil/paramlist.cpp
bfraboni: Fixed :+1:

src/include/OpenImageIO/imageio.h
```diff
@@ -2787,10 +2791,45 @@ class OIIO_API ImageOutput {
...
+/// Specializes the base memory tracking functions from memory.h.
+namespace pvt {
+// heapsize specialization for `ImageSpec`
+template <> inline size_t heapsize<ImageSpec>(const ImageSpec& is)
+{
+    size_t size = heapsize(is.channelformats);
+    size += heapsize(is.channelnames);
+    size += heapsize(is.extra_attribs);
+    return size;
+}
```
lgritz, line 2812: What I think I would recommend for these type-specialized cases is that in the public header file, we should only declare the specialization:

```
// pvt::heapsize specialization for `ImageSpec`
template <> OIIO_API size_t pvt::heapsize<ImageSpec>(const ImageSpec& is);
```

Then in just one module, formatspec.cpp in this case, define the implementation:

```
template <> size_t heapsize<ImageSpec>(const ImageSpec& is);
{
    size_t size = heapsize(is.channelformats);
    size += heapsize(is.channelnames);
    size += heapsize(is.extra_attribs);
    return size;
}
```

I think this will (a) keep the clutter out of the header files; (b) make it so that the implementation is only compiled ONCE, not in every module that sees the header file; (c) hide the implementation details and allow us to change them without forcing a recompile of the whole codebase.

I'm only speaking of the OIIO-type-specialized ones, like `heapsize<ImageSpec>`; the broadly generic ones like `heapsize<vector<T>>` of course just stay in the memory.h header.

src/include/OpenImageIO/memory.h
```diff
@@ -0,0 +1,140 @@
...
+
+
+// heapsize specialization for `intrusive_ptr`
+template<typename T>
+size_t
+heapsize(intrusive_ptr<T> ref)
+{
+    return ref ? footprint(*ref.get()) : 0;
+}
```
lgritz, line 127: Should this go in refcnt.h, so it doesn't need to be seen (and doesn't force inclusion of refcnt.h) absolutely everywhere? Only modules already including refcnt.h will ever need this specialization.

src/include/OpenImageIO/paramlist.h
```diff
@@ -326,6 +327,9 @@ class OIIO_UTIL_API ParamValue {
                       Copy _copy                = Copy(true),
                       FromUstring _from_ustring = FromUstring(false)) noexcept;
     void clear_value() noexcept;
+
+    /// declare a friend heapsize definition
+    template<typename T> friend inline size_t heapsize(const T&);
```
lgritz, line 332: Does it not work to declare only the specialized one as a friend here? Like

    template<> friend size_t pvt::heapsize(const ParamValue&);

src/include/OpenImageIO/paramlist.h
```diff
@@ -711,4 +715,19 @@ class OIIO_UTIL_API ParamValueSpan : public cspan<ParamValue> {
...
+namespace pvt {
+template<>
+inline size_t
+heapsize<ParamValue>(const ParamValue& pv)
+{
+    return (pv.m_nonlocal && pv.m_copy)
+               ? pv.m_nvalues * static_cast<int>(pv.m_type.size())
+               : 0;
+}
+}  // namespace pvt
```
lgritz, line 730: Again, I would move the implementation into paramlist.cpp, only the declaration (not inline) needs to be in the header.

lgritz (changes requested):
The approach looks fine to me, we're very close, though as I noted, I think you can keep the minimal declarations in the headers and move most of the specialized implementations into the modules where the rest of those class methods are implemented. I think that's the only big suggestion I have at this point.

src/include/OpenImageIO/paramlist.h
bfraboni: I think when I've tried it didn't, but I can give it another go , just in case !

src/include/OpenImageIO/imageio.h
bfraboni: Thanks @lgritz , will implement all your remarks right now !

bfraboni:
I implemented your suggestions in the latest commit @lgritz , it should be closer now, all these make a lot of sense :+1:

lgritz:
Something seems unhappy in the CI.

bfraboni:
> Something seems unhappy in the CI.

Sorry about that, I haven't properly tested the latest version.
Should be fine now @lgritz, I fixed the compilation issue and tests pass locally !

src/libtexture/imagecache_pvt.h
```diff
@@ -1197,7 +1205,7 @@ class ImageCacheImpl final : public ImageCache {
     spin_mutex m_fingerprints_mutex;  ///< Protect m_fingerprints
     FingerprintMap m_fingerprints;    ///< Map fingerprints to files
 
-    TileCache m_tilecache;          ///< Our in-memory tile cache
+    mutable TileCache m_tilecache;  ///< Our in-memory tile cache
```
lgritz, line 1208: Why this change?

lgritz (approved):
This all looks great!

I did have one question about why you made one member mutable. I'll wait to merge until you have a chance to respond, just in case that was inadvertent.

bfraboni:
@lgritz I did that change to be able to loop through TileCache from a const reference to ImageCacheImpl, because unordered_map_concurrent doesn't implement const iterators. Maybe not the wiser choice, though ?

src/libtexture/imagecache_pvt.h
bfraboni: @lgritz I did that change to be able to loop through TileCache from a const reference to ImageCacheImpl, because unordered_map_concurrent doesn't implement const iterators. Maybe not the best choice. But I think that's already why the FilenameMap is mutable.

lgritz:
It's all a private type, not part of public APIs, so I think it's fine. Let's leave it, but perhaps you could put a comment next to that declaration that just says "FIXME: if unordered_map_concurrent had const iterators, this wouldn't need to be mutable"?

bfraboni:
Just added the comment!

lgritz (approved):
LGTM!

PR #4323, opened 2024-07-04
api: add virtual heapsize() to ImageInput and ImageOutput

bfraboni:
Second PR of two, following @lgritz and I discussions on memory tracking in the OIIO::ImageCache.

- the memory tracking system from https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4322 is not sufficient to track OIIO public objects that can be overriden. 
- add virtual `heapsize()` method to ImageInput and ImageOutput that return the total heap allocated memory held by the structure and its members.
- [ ] **TODO**: override for every internal OIIO type (bmp, tiff, etc). 

Related PR from Larry : https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4317
First PR: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4322

## Tests
- [ ] todo: fix CI fails

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -1778,6 +1778,16 @@ public:
     /// `ImageInput*`.
     typedef ImageInput* (*Creator)();
 
+    /// Memory tracking method.
+    /// Return the total heap memory allocated by `ImageInput`.
+    /// Overridable version of heapsize defined in memory.h.
+    virtual size_t heapsize() const;
+
+    /// Memory tracking method.
+    /// Return the total memory footprint of `ImageInput`.
+    /// Overridable version of footprint defined in memory.h.
+    virtual size_t footprint() const;
+
 protected:
     ImageSpec m_spec;  // format spec of the current open subimage/MIPlevel
                        // BEWARE using m_spec directly -- not thread-safe
@@ -1886,7 +1896,7 @@ private:
 
     void append_error(string_view message) const; // add to error message
 
-    /// declare a friend heapsize definition
+    /// declare friend heapsize and footprint definitions
     template <typename T> friend size_t pvt::heapsize(const T&);
 };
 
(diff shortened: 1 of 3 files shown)


lgritz:
I rebased onto master for you. I think this is fine in its current state. Shall I merge? You can follow later with individual heapsize implementations for particular formats if you wish.

What happens if an ImageInput implementation stores things in its object instead of strictly being heap based? Or should we just sweep that under the rug and account for that within heapsize just to keep things simple? Or do you feel like we also need to add footprint to these classes?

lgritz:
@bfraboni From your perspective, is this ready to merge for the sake of making the ABI break and then letting things settle down, and we should add more non-breaking implementation details later? Or was there more that you were set on adding before the initial merge?

bfraboni:
Hey @lgritz ,
Just saw your comments , thank you for the rebase!

> What happens if an ImageInput implementation stores things in its object

I actually haven't thought about that point, sizeof(ImageInput) won't account for the sizeof the derived classes, so we might want add a footprint() method too, and similarly top pvt::heapsize, pvt::footprint would simply call ImageInput::footprint, how does that sound ?

No that's perfect if we do the API breaking changes now, and add variants later, I have no additional changes to do except the point above ☝️

lgritz:
OK, let's add the footprint stubs now, and then merge the breaking changes, and then we can come back and flesh out the implementations that correctly account for what's going on in individual ImageInputs/ImageOutputs.

Note that there is still one as-yet-unaccounted aspect, which is that for inputs/outputs that depend on 3rd party libraries, it may be very difficult to know how much memory they are hanging onto in their internals that correspond to the open file.

bfraboni:
I can do the changes later today. 

Right, that part is the unknown, I don't know if some libraries have that info available through some api or visible to the user. Otherwise we can just assume that it is negligible for now.

bfraboni:
@lgritz I added footprint() the same way as heapsize() to the ImageInput/ImageOutput API and specialized pvt::footprint to call this one. I think once tests pass (locally it does) this version is good to be merged .

lgritz (approved):
LGTM

PR #4324, opened 2024-07-07
Fix GHA CI after they upgraded nodejs

lgritz:
They upgraded to a version that requires a glibc newer than is in the pre-2023 ASWF containers.


lgritz:
Merging immediately because this does fix the CI.

PR #4325, opened 2024-07-07
ci: deal with CentOS 7 EOL and disappearance of yum mirrors

lgritz:
This was breaking CI for us. Hard break for icc/icx tests, since those REQUIRED a yum install of the intel compilers. Softer undetected break for all the ASWF <= 2022 containers based on CentOS 7, which were not failing outright but were failing to install certain optional packages.

The solution (for now) is to configure yum to exclude the missing repo.


lgritz:
Whew, this gets us back to passing all CI tests in main!

PR #4326, opened 2024-07-08
fix: iv should enable the ImageCache

jessey-git:
## Description

Enable the ImageCache for the `iv` tool. It more or less expects it to be present and certain features, like loading the Image's mip-levels, requires the ImageCache to be used.

## Checklist:

  nonconforming.


lgritz (approved):
LGTM

Thanks. I guess when I recently changed the default behavior of ImageBuf to not be backed by IC unless this is set (or an IC ptr is explicitly passed to the IB constructor), I didn't fully consider the implications for iv.

PR #4330, opened 2024-07-09
api: consistent naming of OPENIMAGEIO_PLUGIN_PATH env variable

lgritz:
The old name, OIIO_LIBRARY_PATH, was inconsistently named in two ways:

First, it was not clear that it was specific to searching for plugins, and since it is related to the to the global attribute "plugin_searchpath", I thought making it mention plugins specifically would be more clear.

Second, I noticed that all of the publicly documented environment variables that are "for users" started with OPENIMAGEIO_, except for this one. There are a bunch of environment variables we check that begin with OIIO_, but all but this one are intended for developers when testing or debugging.

So I'm using a new name that is consistent with both principles.

Henceforth, the naming rule about environment variables is:

* `OPENIMAGEIO_FOO` for public interfaces, are fully documented and intended for users/sites to use to customize behavior.

* `OIIO_FOO` for developers to use when testing and debugging, are undocumented (except in the code itself), and are not considered part of the public API/behavior of OpenImageIO -- their meaning may change, they may disappear without warning, or they may only be functional for debug builds.

### DIFF
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -340,6 +340,16 @@ inside the source code.
         imagecache->attribute ("options", value);
 
 
+``OPENIMAGEIO_PLUGIN_PATH``
+
+    A colon-separated list of directories to search for OpenImageIO plugins
+    (dynamicaly loadable libraries that implement image format readers
+    and writers).
+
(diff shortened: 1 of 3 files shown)


jessey-git (approved):
+1 for consistency

PR #4332, opened 2024-07-09
api: Make span::size() return size_t

lgritz:
With 3.0 coming soon, it's time to break compatibility by switching OIIO::span::size() to return an unsigned value, matching C++20 std::span.

The history is that when we first added OIIO::span, at that time, the C++ draft proposal for std::span included a signed size() result, but by the time it got finalized into C++20, it was changed to unsigned to match all the other std containers, but it was too late for us to switch without breaking API compatibility rather seriously.


lgritz:
Please voice any reservations about making this change now, I would like to do a merge this weekend of several PRs that will collectively break some back compatibility.

PR #4335, opened 2024-07-11
api: various deprecations in array_view.h, benchmark.h, bit.h, color.h, errorhandler.h

lgritz:
* array_view.h: Remove this long-deprecated header.
* benchmark.h: Attach DEPRECATED warning to external use of benchmark.h's time_trial().
* bit.h: Remove bit_cast that's had deprecation warnings for several releases, and add deprecation warnings to the old rotl32 and rotl64
* color.h: add deprecation warnings and make the deprecated things inline so they can be removed later without an ABI break.
* errorhandler.h: remove deprecated old printf style methods


jessey-git (approved):
Seems alright.

lgritz:
I updated this slightly by adding docs/Deprecations-3.0.md where I'll catalogue the full set of deprecations for 3.0.

PR #4337, opened 2024-07-12
Reduce LevelInfo memory usage.

curtisblack:
## Description

Reduces the memory footprint of the `LevelInfo` struct. Three changes are here, in order of most to least impactful:

- Makes one `ImageSpec` optional, most of the time `spec` and `nativespec` are the same, it appears they are only different when autotile is enabled, this saves about 160 bytes.
- Uses `std::unique_ptr<float[]>` (8 bytes) instead of `std::vector<float>` (24 bytes).
- Reorders members to remove excess padding.

We noticed this memory usage on scenes with per-face textures with multiple mips per face. For a scene with 12.5 million mips, this saves about 2GB.

Note, it would probably also be possible to remove the `polecolorcomputed` member, instead relying on the `polecolor` pointer. But since this bool occupies existing padding in the struct it wouldn't save any memory currently.

## Checklist:

  nonconforming.

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -164,11 +164,11 @@ public:
     }
     const ImageSpec& spec(int subimage, int miplevel) const
     {
-        return levelinfo(subimage, miplevel).spec;
+        return levelinfo(subimage, miplevel).spec();
     }
     ImageSpec& spec(int subimage, int miplevel)
     {
-        return levelinfo(subimage, miplevel).spec;
+        return levelinfo(subimage, miplevel).spec();
     }
     const ImageSpec& nativespec(int subimage, int miplevel) const
     {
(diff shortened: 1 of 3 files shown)


lgritz:
I think the other way that the spec and nativespec can be different is if the pixel data type is not one of the four (float, half, uint16, uint8) that is supported by ImageCache. In that case, spec will reflect the *cache* while nativespec describes the original file.

Incidentally, I have another related patch in-progress that fully eliminates the second ImageSpec per level. But I still had a little work left to tidy it up and fix some minor regressions it causes. However, I'm inclined to merge this first and then re-evaluate my work on top of it.

src/libtexture/imagecache_pvt.h
```diff
@@ -278,8 +287,8 @@ class OIIO_API ImageCacheFile final : public RefCnt {
         bool full_pixel_range    = false;  ///< data window matches image window
         bool is_constant_image   = false;  ///< Is the image a constant color?
         bool has_average_color   = false;  ///< We have an average color
-        std::vector<float> average_color;  ///< Average color
         spin_mutex average_color_mutex;    ///< protect average_color
+        std::vector<float> average_color;  ///< Average color
```
lgritz, line 291: Aside: I'd like to find or create a good `small_vector<T, size>` implementation -- a template that acts like a vector, but can store up to size elements inside itself and only dynamically allocates if more than this expected/typical amount. This would be an ideal place to use a `small_vector<float,4>` where for the usual case of 1-4 channels no heap would be needed.

src/libtexture/imagecache_pvt.h
```diff
@@ -250,18 +250,27 @@ class OIIO_API ImageCacheFile final : public RefCnt {
     /// Info for each MIP level that isn't in the ImageSpec, or that we
     /// precompute.
     struct LevelInfo {
-        ImageSpec spec;         ///< ImageSpec for the mip level
-        ImageSpec nativespec;   ///< Native ImageSpec for the mip level
+        std::unique_ptr<ImageSpec> m_spec;  ///< ImageSpec for the mip level,
+            // only specified if different from nativespec
+        ImageSpec nativespec;  ///< Native ImageSpec for the mip level
+        mutable std::unique_ptr<float[]> polecolor;  ///< Pole colors
```
lgritz, line 256: This is fine for now, but I think that as a further optimization, we can move polecolor to the SubimageInfo struct, where for most files it will be empty/null, but for latlong env maps will point to a single allocation of float[2*nchannels*nmiplevels]. Then we don't even need the per-level pointer.

lgritz (approved):
LGTM, I will merge now since I know you will benefit immediately from this.

I think there's even more we can do, and we should aim eventually to only store the resolution related fields per-level. No metadata should be varying per MIP level, so in theory we should not need an ImageSpec per level, only per subimage. But standing in our way is the public APIs that return a pointer or reference to the IS for each level -- we need to refactor that, and it takes an API change. I'm thinking about the bet way to do that as a next step.

PR #4338, opened 2024-07-12
api: Deprecations: missing_math.h, parmalist.h, parallel.h, strutil.h, sysutil.h, thread.h, tiffutils.h, ustring.h, type_traits.h

lgritz:
* Remove long deprecated and empty header missing_math.h

* paramlist.h: Deprecate the old form of ParamValue constructors.

  People should be using the versions with strongly typed params, not
  raw bools.

* parallel.h: Remove deprecated parallel_for variants.

* strutil.h: Add deprecation warnings to old-style format.

* sysutil.h: Deprecate Sysutil::physical_concurrency.

  Also, I noticed that all the external declarations in this header were
  incorrectly OIIO_API but should have been OIIO_UTIL_API to reflect that
  they are in the OpenImageIO_Util library.

* thread.h: mark deprecated yield() and this_thread_is_in_pool().

* tiffutils.h: removed deprecated redundant functions.

* ustring.h: Remove deprecated ustringHash.

* type_traits.h: Remove deprecated OIIO::void_t, use std::void_t instead.


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2348,7 +2348,7 @@ colorconvert_impl(ImageBuf& R, const ImageBuf& A,
         unpremult = false;
     // clang-format off
     parallel_image(
-        roi, parallel_options(nthreads),
+        roi, paropt(nthreads),
```
jessey-git, line 2351: Looks like there's a, final, reference to `parallel_options` on line 2437 of this file still?  That said, the regression in class name to something so terse is a bit unfortunate :-/

lgritz: Good catch, thanks. Fix pushed.

I hear what you're saying about the name, but the parallel_options -> paropt is something that happened years ago, I'm just finally clearing away the old structure definition. The good thing is that the paropt parameters are almost always either defaulted, or just pass the thread count (paropt has a ctr for just that), so almost nobody external ever actually types "paropt."

jessey-git: I see.  This recent commit also removes the last usage of the `parallel_options` class altogether. Should it be removed too?

lgritz: Unfortunately, it's referenced in the public API of an OSL header. I need to do this incredibly minor API transition in OSL before I can remove it here, or it just creates a different headache.

jessey-git (approved):
LGTM

PR #4339, opened 2024-07-12
api(texture.h): get rid of texture deprecations

lgritz:
* Old version of TextureOptions we haven't used since OIIO 1.x.
* Texture methods that were based on the VaryingRef helper. These are actually remnants of early OSL needs when it was more Reyes-like, BEFORE we used LLVM!
* Get rid of the varyingref.h header, no longer needed by anything.


jessey-git (approved):
Looks good from here.

fpsunflower (approved):

PR #4349, opened 2024-07-20
build(deps): Raise giflib minimum to 5.0

lgritz:
Even that is quite old -- released in 2012! But it's the first version that guaranteed thread safety for error reporting. So we're finally making that our minimum and taking out the locking we did on our side.


jessey-git (commented):
There one other block inside `GIFInput::seek_subimage` that could be removed I think now.

jessey-git (approved):
Looks all good now.

PR #4350, opened 2024-07-21
feat(ustring): ustring hash collision protection

lgritz:
The gist is that the ustring::strhash(str) function is modified to strip out the MSB from Strutil::strhash.  The rep entry is filed in the ustring table based on this hash.  So effectively, the computed hash is 63 bits, not 64.

But rep->hashed field consists of the lower 63 bits being the computed hash, and the MSB indicates whether this is the 2nd (or more) entry in the table that had the same 63 bit hash.

ustring::hash() then is modified as follows: If the MSB is 0, the computed hash is the hash. If the MSB is 1, though, we DON'T use that hash, and instead we use the pointer to the unique characters, but with the MSB set (that's an invalid address by itself). Note that the computed hashes never have MSB set, and the char*+MSB always have MSB set, so therefore ustring::hash() will never have the same value for two different ustrings.

But -- please note! -- that ustring::strhash(str) and ustring(str).hash() will only match (and also be the same value on every execution) if the ustring is the first to receive that hash, which should be approximately always. Probably always, in practice.

But in the very improbable case of a hash collision, one of them (the second to be turned into a ustring) will be using the alternate hash based on the character address, which is both not the same as ustring::strhash(chars), nor is it expected to be the same constant on every program execution.

### DIFF
--- a/src/include/OpenImageIO/ustring.h
+++ b/src/include/OpenImageIO/ustring.h
@@ -30,6 +30,7 @@
 #define OIIO_USTRING_HAS_CTR_FROM_USTRINGHASH 1
 #define OIIO_USTRING_HAS_STDHASH 1
 #define OIIO_HAS_USTRINGHASH_FORMATTER 1
+#define OIIO_USTRING_SAFE_HASH 1
 
 
 OIIO_NAMESPACE_3_1_BEGIN
@@ -120,6 +121,16 @@ OIIO_NAMESPACE_3_1_BEGIN
 ///   - if you don't need to do a lot of string assignment or equality
 ///     testing, but lots of more complex string manipulation.
 ///
+/// The ustring implementation guarantees that no two ustrings return the same
+/// value for hash() (despite the slim probability that two strings could
+/// numerically hash to the same value).  For the first ustring added with a
+/// given hash, u.hash() will be the same value as ustring::strhash(chars),
+/// and will deterministically be the same on every execution. In the very
+/// improbable case of a hash collision, subsequent ustrings with the same
+/// numeric hash will use an alternate hash based on the character address,
+/// which is both not the same as ustring::strhash(chars), nor is it expected
+/// to be the same constant on every program execution.
+
 class OIIO_UTIL_API ustring {
 public:
     using rep_t      = const char*;  ///< The underlying representation type
@@ -285,11 +296,7 @@ public:
     /// Return a C++ std::string representation of a ustring.
     const std::string& string() const noexcept
     {
-        if (m_chars) {
-            const TableRep* rep = (const TableRep*)m_chars - 1;
-            return rep->str;
-        } else
-            return empty_std_string;
+        return m_chars ? rep()->str : empty_std_string;
     }
 
     /// Reset to an empty string.
@@ -300,17 +307,27 @@ public:
     {
         if (!m_chars)
             return 0;
-        const TableRep* rep = ((const TableRep*)m_chars) - 1;
-        return rep->length;
+        return rep()->length;
     }
 
-    /// Return a hashed version of the string
+    /// ustring::strhash() uses Strutil::strhash but clears the MSB.
+    static OIIO_HOSTDEVICE constexpr hash_t strhash(string_view str)
+    {
+        return Strutil::strhash(str) & hash_mask;
+    }
+
+    /// Return a hashed version of the string. To guarantee unique hashes,
+    /// we check if the "duplicate bit" of the hash is set. If not, then
+    /// we just return the hash which we know is unique. But if that bit
+    /// is set, we utilize the unique character address.
     hash_t hash() const noexcept
     {
         if (!m_chars)
             return 0;
-        const TableRep* rep = ((const TableRep*)m_chars) - 1;
-        return rep->hashed;
+        hash_t h = rep()->hashed;
+        return OIIO_LIKELY((h & duplicate_bit) == 0)
+                   ? h
+                   : hash_t(m_chars) | duplicate_bit;
     }
 
     /// Return a hashed version of the string
@@ -736,6 +753,8 @@ public:
     // if you know the rep, the chars are at (char *)(rep+1), and if you
     // know the chars, the rep is at ((TableRep *)chars - 1).
     struct TableRep {
+        // hashed has the MSB set if and only if this is the second or
+        // greater ustring to have the same hash.
         hash_t hashed;    // precomputed Hash value
         std::string str;  // String representation
         size_t length;    // Length of the string; must be right before cap
@@ -744,10 +763,29 @@ public:
         TableRep(string_view strref, hash_t hash);
         ~TableRep();
         const char* c_str() const noexcept { return (const char*)(this + 1); }
+        constexpr bool unique_hash() const
+        {
+            return (hashed & duplicate_bit) == 0;
+        }
     };
+
+    // duplicate_bit is a 1 in the MSB, which if set indicates a hash that
+    // is a duplicate.
+    static constexpr hash_t duplicate_bit = hash_t(1) << 63;
+    // hash_mask is what you `&` with hashed to get the real hash (clearing
+    // the duplicate bit).
+#if 1
+    static constexpr hash_t hash_mask = ~duplicate_bit;
+#else
+    // Alternate to force lots of hash collisions for testing purposes:
+    static constexpr hash_t hash_mask = ~duplicate_bit & 0xffff;
+#endif
+    bool has_unique_hash() const { return rep()->unique_hash(); }
 
 private:
     static std::string empty_std_string;
+
+    const TableRep* rep() const { return ((const TableRep*)m_chars) - 1; }
 };
 
 
@@ -798,7 +836,7 @@ public:
     OIIO_DEVICE_CONSTEXPR explicit ustringhash(const char* str)
 #ifdef __CUDA_ARCH__
         // GPU: just compute the hash. This can be constexpr!
-        : m_hash(Strutil::strhash(str))
+        : m_hash(ustring::strhash(str))
 #else
         // CPU: make ustring, get its hash. Note that ustring ctr can't be
         // constexpr because it has to modify the internal ustring table.
@@ -810,7 +848,7 @@ public:
     OIIO_DEVICE_CONSTEXPR explicit ustringhash(const char* str, size_t len)
 #ifdef __CUDA_ARCH__
         // GPU: just compute the hash. This can be constexpr!
-        : m_hash(Strutil::strhash(len, str))
+        : m_hash(ustring::strhash(len, str))
 #else
         // CPU: make ustring, get its hash. Note that ustring ctr can't be
         // constexpr because it has to modify the internal ustring table.
@@ -824,7 +862,7 @@ public:
     OIIO_DEVICE_CONSTEXPR explicit ustringhash(string_view str)
 #ifdef __CUDA_ARCH__
         // GPU: just compute the hash. This can be constexpr!
-        : m_hash(Strutil::strhash(str))
+        : m_hash(ustring::strhash(str))
 #else
         // CPU: make ustring, get its hash. Note that ustring ctr can't be
         // constexpr because it has to modify the internal ustring table.
@@ -918,13 +956,13 @@ public:
     /// Test for equality with a char*.
     constexpr bool operator==(const char* str) const noexcept
     {
-        return m_hash == Strutil::strhash(str);
+        return m_hash == ustring::strhash(str);
     }
 
     /// Test for inequality with a char*.
     constexpr bool operator!=(const char* str) const noexcept
     {
-        return m_hash != Strutil::strhash(str);
+        return m_hash != ustring::strhash(str);
     }
 
 #ifndef __CUDA_ARCH__
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -2,9 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/AcademySoftwareFoundation/OpenImageIO
 
+#include <cstring>
 #include <string>
 #include <unordered_map>
 
+#include <OpenImageIO/bit.h>
 #include <OpenImageIO/dassert.h>
 #include <OpenImageIO/export.h>
 #include <OpenImageIO/strutil.h>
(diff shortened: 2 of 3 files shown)


lgritz:
Comments @cmstein @sfriedmapixar @etheory ?

lgritz:
Maybe @chellmuth too?

src/include/OpenImageIO/ustring.h
```diff
@@ -303,17 +310,27 @@ class OIIO_UTIL_API ustring {
...
-        return rep->length;
+        return rep()->length;
     }
 
-    /// Return a hashed version of the string
+    /// ustring::strhash() uses Strutil::strhash but clears the MSB.
+    static OIIO_HOSTDEVICE constexpr hash_t strhash(string_view str)
+    {
+        return Strutil::strhash(str) & hash_mask;
+    }
```
sfriedmapixar, line 320: In our implementation, we do an OR instead of an and, so that "pointer hashes" look like pointers in a debugger, and it saves clearing the bit when we want to use it like a pointer, and "hash-hashes" are always negative when looking at them in a debugger.

src/include/OpenImageIO/ustring.h
```diff
@@ -303,17 +310,27 @@ class OIIO_UTIL_API ustring {
...
     hash_t hash() const noexcept
     {
         if (!m_chars)
             return 0;
-        const TableRep* rep = ((const TableRep*)m_chars) - 1;
-        return rep->hashed;
+        hash_t h = rep()->hashed;
+        return OIIO_LIKELY((h & duplicate_bit) == 0)
+                   ? h
+                   : hash_t(m_chars) | duplicate_bit;
```
sfriedmapixar, line 333: With a "pointer hash", you can actually just store that "hash" the "hashed" member on TableRep creation, then you don't need to do any of this hashing or checking at all, you just return the value like before.

src/include/OpenImageIO/ustring.h
```diff
@@ -757,10 +776,29 @@ class OIIO_UTIL_API ustring {
         TableRep(string_view strref, hash_t hash);
         ~TableRep();
         const char* c_str() const noexcept { return (const char*)(this + 1); }
+        constexpr bool unique_hash() const
+        {
+            return (hashed & duplicate_bit) == 0;
```
sfriedmapixar, line 781: If you flip the polarity of the indicator bit, this needs to flip.

src/libutil/ustring.cpp
```diff
@@ -71,6 +105,10 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
 
     const char* lookup(string_view str, uint64_t hash)
     {
+        if (OIIO_UNLIKELY(hash & ustring::duplicate_bit)) {
+            // duplicate bit is set -- the hash is related to the chars!
+            return OIIO::bitcast<const char*>(hash & ustring::hash_mask);
```
sfriedmapixar, line 110: If you flip the polarity of the indicator bit, the extra & is uneeded.

src/libutil/ustring.cpp
```diff
@@ -98,6 +135,10 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
     // the hash.
     const char* lookup(uint64_t hash)
     {
+        if (OIIO_UNLIKELY(hash & ustring::duplicate_bit)) {
+            // duplicate bit is set -- the hash is related to the chars!
+            return OIIO::bitcast<const char*>(hash & ustring::hash_mask);
```
sfriedmapixar, line 140: If you flip the polarity of the indicator bit, the extra & is uneeded.

src/libutil/ustring.cpp
```diff
@@ -108,38 +149,64 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
+            if (e->hashed == hash)
+                return e->c_str();
             ++dist;
             pos = (pos + dist) & mask;  // quadratic probing
         }
     }
 
     const char* insert(string_view str, uint64_t hash)
     {
+        OIIO_ASSERT((hash & ustring::duplicate_bit) == 0);  // can't happen?
```
sfriedmapixar, line 164: If you flip the polarity of the indicator bit, flip this assert.

src/libutil/ustring.cpp
```diff
@@ -108,38 +149,64 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
             pos = (pos + dist) & mask;  // quadratic probing
         }
 
         ustring::TableRep* rep = make_rep(str, hash);
-        entries[pos]           = rep;
+
+        // If we encountered another ustring with the same hash (if one
+        // exists, it would have hashed to the same address so we would have
+        // seen it), set the duplicate bit in the rep's hashed field.
+        if (duplicate_hash) {
```
sfriedmapixar, line 190: If you change the polarity of the indicator bit, you can pass "duplicate_hash" into make_rep, and then it can just store the pointer from the pool alloc in the "hashed" member, and all you have to do here is inc the collisions counter (or the inc and dbug print could be moved into make_rep as well. -- if you do that, you need to do another pobe for a valid "pos" here based on the pointer bits rather than the original hash bits.

src/libutil/ustring.cpp
```diff
@@ -108,38 +149,64 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
+
+        entries[pos] = rep;
         ++num_entries;
         if (2 * num_entries > mask)
-            grow();           // maintain 0.5 load factor
+            grow();  // maintain 0.5 load factor
+        // ensure low bit clear
+        OIIO_DASSERT((size_t(rep->c_str()) & 1) == 0);
+        // ensure low bit clear
+        OIIO_DASSERT((size_t(rep->c_str()) & ustring::duplicate_bit) == 0);
```
sfriedmapixar, line 209: If you change the polarity of the indicator bit, you need to flip this assert.

src/libutil/ustring.cpp
```diff
@@ -459,11 +510,8 @@ ustring::make_unique(string_view strref)
     if (!strref.data())
         strref = string_view("", 0);
 
-    hash_t hash = Strutil::strhash64(strref);
-    // This line, if uncommented, lets you force lots of hash collisions:
-    // hash &= ~hash_t(0xffffff);
+    hash_t hash = ustring::strhash(strref);
```
sfriedmapixar, line 514: You can use a similar "force a lot of collisions" to to locally test that the collision code is all correct, now that it's much more complicated with the ptr vs hash perfect hashing.

src/libutil/ustring.cpp
```diff
@@ -637,11 +596,12 @@ ustring::getstats(bool verbose)
 size_t
 ustring::hash_collisions(std::vector<ustring>* collisions)
 {
-    std::lock_guard<std::mutex> lock(collision_mutex);
-    if (collisions)
-        for (const auto& c : all_hash_collisions)
-            collisions->emplace_back(ustring::from_unique(c.first));
-    return all_hash_collisions.size();
+    if (collisions) {
+        // Disabled for now
```
sfriedmapixar, line 600: You don't need to keep a collection of all hash collisions if you put the pointer into the "hashed" member on collision...then you can just iterate through all the strings in the system, and print out those who's hash is actually a pointer.

sfriedmapixar (changes requested):
I implemented this locally with the MSB polarity flipped from what you've presented.  I did this it's easier to see (in a debugger) pointer-based perfect hashes are pointers, it doesn't require bit manip to deref them, and so and if it happens, you can just dive into those in a debugger without having to mask off the high bit.  Semantically I think it's also cleaner to actually use that pointer value as the hashing value into the map, and store it in the tablerep.  Really, just think of it as a way to create a perfect hash out of an imperfect one -- with the caveat that it assumes you're on a system with user-space pointers that can never have the MSB set.   You could do the same thing with the LSB instead since the tablerep always comes out of a pool, by guaranteeing the alignment of those allocations wouldn't have the low bit set for pointers and that would work everywhere.   If you just think of the collision resolution as part of the "hashing" function, then it makes sense to store that hash in the tablerep for it in the "hashed" member.

etheory:
I'm OK with any of these options as long as collisions are handled.
I do agree with @sfriedmapixar about the simplicity of debugging if the top bit is 1 for a hash and 0 for a string though.

Also remember that on every current system, the bottom few bits are zero due to alignment, and the top few are zero due to memory allocation guarantees: https://stackoverflow.com/questions/46509152/why-in-x86-64-the-virtual-address-are-4-bits-shorter-than-physical-48-bits-vs which means you can often rely on the top 12 to 16bits also being zero. Keep that in mind when you do any of these tricks. We use these bits in glimpse for storage of data, so it's quite useful.

src/include/OpenImageIO/ustring.h
lgritz: Interesting. I think I was trying to preserve the "0 means empty string" property without needing extra conditionals, but maybe it's easier than I thought. I will noodle with this and see if I can make it work just as simply with the reversed polarity, because I do see your point about how it's nice for the pointers to be the unchanged values.

lgritz:
Something that has been nagging at my mind... I know that the MSB will always be 0 for a real address on x86, but do we feel confident that this will also be true for ARM, RISC V, or others that may come along?

I also considered using the LSB for this, also safe because the allocations are always aligned, the characters will never start on an odd address, so if we "`|1`" the hashes to set the low bit, then that's also a simple way to distinguish pointers from hashes and will be true of any future architecture.

Thoughts?

lgritz:
The reason I really like empty string to be all 0 bits (and thus thought that must mean that it's pointers, not hashes, with the extra bit set) is so we can memset to 0 a region and know that we've made valid (empty) ustrings or ustringhashes.

ThiagoIze:
> I also considered using the LSB for this, also safe because the allocations are always aligned, the characters will never start on an odd address, so if we "|1" the hashes to set the low bit, then that's also a simple way to distinguish pointers from hashes and will be true of any future architecture.

Setting the LSB hash will reduce the hash quality in a risky way. Suppose for instance someone did: `my_array[hash%my_array.length()]` where the length is 8 (it could be any size, but this is simple to show). The `hash%8` will result in just the three lowest signed bits being used. But note that your proposal has the LSB at 1, so now we just have 2 bits in play and so only half the array will be referenceable.

lgritz:
Right, gotcha. That's a good reason to stick with MSB. 

It's ok, we'll have asserts that no real address has the high bit set. We'll have faith that we'll catch it right away if anybody ports to a hardware platform that might use that part of the address space.

lgritz:
OK, in my tree (not yet pushed here) I recoded it all to make the MSB 1 for "original" hashes, and 0 for rehashes (which are replaced by the char ptr as the return value for `ustring.hash()`). I believe it's all working, but there is a tradeoff involved that I'm not entirely happy with. Let me review the issues:

In the table (invisible to the user), each entry has a hashed field, and the chars themselves. The MSB of the hashed field reveals if this is the first entry with that hash (used to be 0, now is 1 in my new draft), or if it is a subsequently encountered string that had the same 31 bit hash. The public hash() function returns rep->hashed if it's a first string entry having that hash, or the char address if it's a duplicate hash.

A property we've always had that I think is very important to retain is that both ustring and ustringhash are represented by a true 0 value for empty strings. In other words, it's special -- the ptr is null and the hash should be 0. This is what lets us zero out a block of memory and know that it's initializing any ustring or ustringhash therein to be valid, and represent an empty string.

Now then, the way I did it first in this PR trivially preserved that property, because MSB was 1 only for duplicate hashes, so the empty string is canonical and hashes to 0. The alternate way, where it would "want" to be a MSB 1 for canonical hashes, requires special handling of the empty string case in a variety of places. I think this is a little confusing and error-prone, and certainly has a runtime cost for the extra testing and branching.

@ssteinbach and others pointed out that it's nice for the pointers to have the MSB=0 rather than =1, because then you can just cast it to a pointer to get to the characters, especially helpful when debugging.

But here's where I want to push back a little: Is it, really? The only case where you'd get a hash that's a pointer is for a 2nd string that had the same hash. This should be exceptionally rare, so it's not something that would come up... almost never... while in the debugger looking at a ustringhash. Usually you'd just see the hash, so there's not a direct way to get to the chars, so it doesn't matter if MSB is 0 or 1 for that case. (And if we set MSB on the pointer, that just means to get to the chars, you'd need to mask it... awkward, but how often does that really come up? You could also just CALL -- yes, in the debugger -- ustringhash.c_str() to get to the chars.)

So here's the crux of it:

MSB=1 for duplicate hash:
* PRO: no special handling of empty string case to make the hash 0
* CON: "pointer substituted for duplicate hash" needs to have MSB masked off to get to the chars, which is painful in the debugger but almost never happens.

MSB=0 for duplicate hash:
* PRO: duplicate hashes trivially castble to pointers, easy in the debugger, but this almost never happens.
* CON: extra operations and potentially a branch to ensure ustringhash("") == 0, in several places that actually execute frequently.

Thoughts? Are people willing to pay (slightly) in runtime perf to have this property of trivial cast of duplicate hashes to pointers in the debuger, that you'll almost never encounter in real life?

etheory:
You know what, add a static function to do the cast in the debugger and you solve all the issues. Should be straight forward enough. Then you can just call the correct function in the debugger directly. Just document it clearly and that should resolve the issue no?

lgritz:
> You know what, add a static function to do the cast in the debugger and you solve all the issues. 

Isn't that just the existing `ustringhash.c_str()`?

etheory:
> > You know what, add a static function to do the cast in the debugger and you solve all the issues.
> 
> Isn't that just the existing `ustringhash.c_str()`?

If you only have the pointer for some reason, you cannot call the function on it (though I guess you could cast it....).
My point is that if you provide a free function to convert the bits for display/conversion, then you can just call it at any time in a debugger in any situation. Whether you have the object reference or not. If you can just cast the address and call the function, then sure, it's fine as is.

lgritz:
If you have the pointer, you already have the pointer.

If you have a hash value `h` as a ustringhash, you get to the chars with `h.c_str()`.

If you have a hash value `v` as a uint64, you get to the chars with `ustringhash(v).c_str()`.

lgritz:
Does my last answer assuage concerns, or do we need an additional utility function?

etheory:
Yep all good thanks @lgritz - sorry for the slow turn around, I was on leave.

lgritz:
@sfriedmapixar you had asked about the stab I took at swapping the bit to the other way as you originally suggested. If you're curious, it's here: https://github.com/lgritz/OpenImageIO/tree/lg-ustringhash2 (just one additional commit on top of what I have here in this PR)

I like this PR here better, as it avoids some extra logic (both potentially perf impacting, but maybe more importantly, error prone and easy to forget) that is necessary to special case empty string handing.

Let me know if you feel really strongly about doing it the other way. If you're ok with this, though, I'd like to merge it.

PR #4354, opened 2024-07-22
deps: Raise WebP minimum version to 1.1

lgritz:
This new version dates to 2020. Previously, we had supported (purportedly) WebP 0.6.1, but we didn't test that configuration.  By bumping to 1.1 minimum, we can use WebP's own exported cmake config, rather than maintaining a FindWebP.cmake mmodule on our side.


jessey-git (approved):
Glad we don't have to carry that find module any longer and seems fine overall.

Jamim:
_Hi there,_

Due to these changes [we can't build](https://bugs.gentoo.org/937031#c2) **OpenImageIO** v3+ on [Gentoo Linux](https://www.gentoo.org/) with **WebP** support at the moment 😢

lgritz:
Here's the reply I posted on the Gentoo bug thread, which I think accurately summarizes the situation:

---

Hi, I'm the project lead for OpenImageIO.

Modern cmake guidance is is to get rid of the per-project FindFoo.cmake modules (which are idiomatic to the CMake 2.x era) for any dependencies which themselves have modern (3.x era) cmake build systems that export cmake configs for downstream dependencies to consume (those are the WebPConfig.cmake style files). 

The general philosophy currently is that a package should be responsible for exporting all the right things for downstream consumption because they know how to do it right, rather than EVERY downstream consumer project separately having the burden of hacking up their own FindFoo module and hoping they get it right and don't fall out of sync with the upstream project.

So once we raised our minimum supported WebP version high enough that we could count on WebP providing its exported config, we removed our own FindWebP.cmake substitute.

That WebPConfig.cmake is built by webp itself and should be part of the package installation in any distro that provides webp. Is there a reason why it's somehow being dropped on the floor by Gentoo?

OpenImageIO has lots of other dependencies for which we rely on the exported configs, so if webp is the only dependency having this problem, it can't be that your policy is to never install them. It must be something particular to your webp build and packaging, right?

Jamim:
_Hello @lgritz,_

Thank you for the quick response!

> Is there a reason why it's somehow being dropped on the floor by Gentoo?

**libwebp** [supports](https://chromium.googlesource.com/webm/libwebp#building) [multiple build systems](https://chromium.googlesource.com/webm/libwebp/+/HEAD/doc/building.md). Gentoo [uses](https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libwebp/libwebp-1.5.0.ebuild#n6) [autotools](https://devmanual.gentoo.org/eclass-reference/autotools.eclass/index.html) to build it.
It looks like **libwebp** [generates](https://chromium.googlesource.com/webm/libwebp/+/refs/heads/main/CMakeLists.txt#815) CMake configuration only if you use CMake to build it.

lgritz:
> **libwebp** [supports](https://chromium.googlesource.com/webm/libwebp#building) [multiple build systems](https://chromium.googlesource.com/webm/libwebp/+/HEAD/doc/building.md). Gentoo [uses](https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libwebp/libwebp-1.5.0.ebuild#n6) [autotools](https://devmanual.gentoo.org/eclass-reference/autotools.eclass/index.html) to build it. It looks like **libwebp** [generates](https://chromium.googlesource.com/webm/libwebp/+/refs/heads/main/CMakeLists.txt#815) CMake configuration only if you use CMake to build it.

Ah, yes, that seems to be the crux of the issue. Not all the components that downstream consumers rely on are built by the old autotools based system.

PR #4355, opened 2024-07-22
deps: Raise libPNG minimum to 1.6.0

lgritz:
Raised from 1.5.13 (maybe? we never carefully checked) to 1.6.0 minimum, which was released in 2013 (so absolutely nobody should be bothered by this).


jessey-git (approved):
Looks good.

Issue #4356, opened 2024-07-23
[BUG] build failed with libheif-1.17.6 on gentoo


vowstar:
**Describe the bug**

When I upgraded the system and compiled and installed OpenImageIO, I got a build error.

**OpenImageIO version and dependencies**

```bash
OIIO 2.5.13.1 | Linux/x86_64
    Build compiler: gcc 13.3 | C++17/201703
    HW features enabled at build:
    sse2,sse3,ssse3,sse41,sse42,avx,avx2,avx512f,avx512dq,avx512ifma,avx512cd,
    avx512bw,avx512vl,fma,f16c
Dependencies: Boost 1.85.0, BZip2 1.0.8, FFmpeg 6.0, fmt 9.1.0, Freetype
    2.13.0, GIF 5.2.2, Libheif 1.17.6, libjpeg-turbo 3.0.3, OpenColorIO
    2.3.2, OpenEXR 3.2.4, OpenGL, OpenJPEG 2.5, PNG 1.6.43, pugixml 1.14,
    pybind11 2.13.1, Python 3.11.9, Qt5 5.15.14, Qt6 NONE, Robinmap, TIFF
    4.6.0, WebP, ZLIB 1.3.1
```

**To Reproduce**

1. Install ``media-libs/libheif-1.17.6`` source code from ``::gentoo``
2. Install ``media-libs/openimageio-2.5.12.0`` source code from ``::gentoo``
3. Build failed.

If the problem occurs in your C++ or Python code that uses the OIIO APIs, can
you also reproduce the problem using oiiotool? If so, please describe the
exact command line that reproduces the problem. (Being able to reproduce the
problem using only OIIO components makes it a lot easier for the developers
investigate and makes it clear it's not your application's fault.)

**Evidence**

- Error messages:
```bash
...
-- Found FFmpeg 6.0 
-- Found GIF 5.2.2 
CMake Error at src/cmake/modules/FindLibheif.cmake:34 (string):
  string sub-command REGEX, mode MATCHALL needs at least 5 arguments total to
  command.
Call Stack (most recent call first):
  src/cmake/dependency_utils.cmake:376 (find_package)
  src/cmake/externalpackages.cmake:169 (checked_find_package)
  CMakeLists.txt:178 (include)


-- Found Libheif  
CMake Error at src/cmake/dependency_utils.cmake:450 (if):
  if given arguments:

    "VERSION_LESS" "1.16"

  Unknown arguments specified
[...]

PR #4357, opened 2024-07-23
build(deps): Fix FindLibheif.cmake failed on gentoo

vowstar:
## Description

- Fixed the issue where FindLibheif.cmake failed to obtain the libheif version on gentoo.

``media-libs/libheif`` in ``::gentoo`` has an extra layer of warpper, which causes the regular expression to fail. Here I used compiler preprocessing to solve the problem. Regardless of whether there is a warpper or a space is added before the macro definition, it can be processed correctly.

Bug: https://bugs.gentoo.org/936472
Closes: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4356

## Checklist:

  nonconforming.

### DIFF
--- a/src/cmake/modules/FindLibheif.cmake
+++ b/src/cmake/modules/FindLibheif.cmake
@@ -30,7 +30,19 @@ find_library (LIBHEIF_LIBRARY heif
               DOC "The directory where libheif libraries reside")
 
 if (LIBHEIF_INCLUDE_DIR)
-    file(STRINGS "${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h" TMP REGEX "^#define LIBHEIF_VERSION[ \t].*$")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+        # msvc
+        set(PREPROCESS_COMMAND ${CMAKE_CXX_COMPILER} /E "${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h")
+    else ()
+        # clang, gcc or icc
+        set(PREPROCESS_COMMAND ${CMAKE_CXX_COMPILER} -E -dD -P "${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h")
+    endif ()
+    execute_process(
+        COMMAND ${PREPROCESS_COMMAND}
+        OUTPUT_VARIABLE PREPROCESS_OUTPUT
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    string(REGEX MATCH "#define LIBHEIF_VERSION[ \t]+[^\n]+" TMP "${PREPROCESS_OUTPUT}")
     string(REGEX MATCHALL "[0-9.]+" LIBHEIF_VERSION ${TMP})
 endif ()
 


src/cmake/modules/FindLibheif.cmake
```diff
@@ -30,7 +30,12 @@ find_library (LIBHEIF_LIBRARY heif
               DOC "The directory where libheif libraries reside")
 
 if (LIBHEIF_INCLUDE_DIR)
-    file(STRINGS "${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h" TMP REGEX "^#define LIBHEIF_VERSION[ \t].*$")
+    execute_process(
+        COMMAND ${CMAKE_CXX_COMPILER} -E -dD -P "${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h"
```
lgritz, line 34: Does this do the right thing with all compilers? MSVS?

vowstar: No, if MSVS, sould ``set(PREPROCESS_COMMAND "${CMAKE_CXX_COMPILER} /E ${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h")``

Maybe I should change the code to something like this?

```bash
    if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
        # msvc
        set(PREPROCESS_COMMAND "${CMAKE_CXX_COMPILER} /E ${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h")
    else ()
        # icc, gcc or clang
        set(PREPROCESS_COMMAND "${CMAKE_CXX_COMPILER} -E -dD -P ${LIBHEIF_INCLUDE_DIR}/libheif/heif_version.h")
    endif ()
```

vowstar:
Hi @lgritz, 
I have committed a fix regarding multi-compiler support (especially msvc), please review it, thank you for your help!

vowstar:
ping @lgritz

lgritz (approved):
Seems to work, LGTM.

lgritz:
I think this is fine, but for my curiosity, can you please post what the libheif/heif_version.h looks like on gentoo?

The header generally contains this line that matched the original regex:

```
#define LIBHEIF_VERSION "1.18.2"
```

and I'd like to understand how gentoo has modified the file so that the original regex failed and you need to actually run the whole header through the preprocessor.

vowstar:
The gentoo distribution uses ``multilib-build.eclass`` to generate ``/usr/include/libheif/heif_version.h``, which is intended to be multilib-friendly. Although it solves the multi-architecture problem, it leads to FindLibheif Regular expression fails to obtain version information.

``media-libs/libheif`` version installed: ``1.17.6``

```bash
$ ~ qfile /usr/include/libheif/heif_version.h
media-libs/libheif: /usr/include/libheif/heif_version.h
```

The following is the content of ``/usr/include/libheif/heif_version.h``

```bash
$ ~ cat /usr/include/libheif/heif_version.h
/* This file is auto-generated by multilib-build.eclass
 * as a multilib-friendly wrapper. For the original content,
 * please see the files that are #included below.
 */

#if defined(__x86_64__) /* amd64 */
#	if defined(__ILP32__) /* x32 ABI */
#		error "abi_x86_x32 not supported by the package."
#	else /* 64-bit ABI */
#		include <x86_64-pc-linux-gnu/libheif/heif_version.h>
#	endif
#elif defined(__i386__) /* plain x86 */
#	include <i686-pc-linux-gnu/libheif/heif_version.h>
#elif defined(__mips__)
#	if(_MIPS_SIM == _ABIN32) /* n32 */
#		error "abi_mips_n32 not supported by the package."
#	elif(_MIPS_SIM == _ABI64) /* n64 */
#		error "abi_mips_n64 not supported by the package."
#	elif(_MIPS_SIM == _ABIO32) /* o32 */
#		error "abi_mips_o32 not supported by the package."
#	endif
#elif defined(__sparc__)
#	if defined(__arch64__)
#		error "abi_sparc_64 not supported by the package."
#	else
#		error "abi_sparc_32 not supported by the package."
#	endif
#elif defined(__s390__)
#	if defined(__s390x__)
#		error "abi_s390_64 not supported by the package."
#	else
#		error "abi_s390_32 not supported by the package."
#	endif
#elif defined(__powerpc__) || defined(__ppc__)
#	if defined(__powerpc64__) || defined(__ppc64__)
#		error "abi_ppc_64 not supported by the package."
#	else
#		error "abi_ppc_32 not supported by the package."
#	endif
#elif defined(SWIG) /* https://sourceforge.net/p/swig/bugs/799/ */
#	include <x86_64-pc-linux-gnu/libheif/heif_version.h>
#else
#	error "No ABI matched, please report a bug to bugs.gentoo.org"
#endif
```

lgritz:
Does Gentoo substitute this kind of wrapper for headers of other packages as well?

Considering how common it is for FindFoo.cmake modules to match lines in found headers for version numbers definitions (and other things), this seems like it will break all sorts of packages if the headers as found don't contain their expected contents. I'm surprised libheif is the only one you ran into trouble with for OIIO -- try `grep define src/cmake/modules/*cmake` to see how common this is.

I'm sorry to say this after you've worked to prepare this PR, but I think that this is the wrong approach for how to deal with this. We shouldn't have to run any header we want to match through a separate invocation of the C++ compiler to fully preprocess it. It's a needless expense for every build to incur just because Gentoo did a weird thing. (At the least, this extra step and expense should only be for when Gentoo is detected.)

It also seems inevitable that you or others will come back asking for the same change to be made for all the other Find modules, both in this project and all the other projects, right?

I would propose that we instead fix it by modifying the search paths that CMake uses. 

Does Gentoo have a full shadow directory of these things under /usr/x86_64-pc-linux-gnu? Or is every packages stored in a different place like /usr/blahblah/libheif/x86_64-pc-linux-gnu ?

Can you verify whether doing an OIIO build without any of these modifications, but setting `CMAKE_SYSTEM_PREFIX_PATH` and/or `CMAKE_PREFIX_PATH` to include this special Gentoo path where the real headers live, gives you a successful build or not?

I really think that a simpler, better, more general solution to the problem is to leave the individual projects and Find modules alone, and instead have Gentoo's build recipes simply add the `x86_64-pc-linux-gnu` (or whatever the full path for the right architecture is) to the right CMake search paths, if that works, or send a PR to CMake itself to add "if running on Gentoo, use this logic to set the default value of these search paths".

Or if none of that works and we simply have to change something on the OIIO side, I'd rather do it once in a simple place that looks like

    if (running on Gentoo)
        prepend the right path to CMAKE_PREFIX_PATH
    endif ()

so that all searches check for the "real" headers instead of getting the wrappers.

vowstar:
What you said makes perfect sense. 

> Does Gentoo substitute this kind of wrapper for headers of other packages as well?

By reading the source code, I found that ``libheif-1.18.1.ebuild`` uses ``MULTILIB_WRAPPED_HEADERS`` on gentoo to implement automated wrappers.[1]
I noticed that the ``MULTILIB_WRAPPED_HEADERS`` of gentoo's ``multilib-build.eclass`` also emphasized: ``Please note that header wrapping is *discouraged*. It is preferred to install all headers in a subdirectory of libdir and use pkg-config to locate the headers. Some C preprocessors will not work with wrapped headers.``[2]



Therefore, the problem is the ``libheif-1.18.1.ebuild`` package maintained by gentoo. Maybe I should contact the maintainers of the libheif package under gentoo to find a better solution. This is not a problem that OpenImageIO should solve.

[1]: https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libheif/libheif-1.18.1.ebuild
[2]: https://devmanual.gentoo.org/eclass-reference/multilib-build.eclass/index.html

PR #4363, opened 2024-07-31
build: Don't link libOpenImageIO against OpenCV

lgritz:
Fully linking OpenCV against libOpenImageIO is burdensome for people downstream, especially those who want a static version of libOpenImageIO.

There were only three IBA functions that use OpenCV: to_OpenCV(), from_OpenCV(), and capture_image(), and all three are fairly short. Move them to a separate header, imagebufalgo_opencv.h, make them fully INLINE (so they don't actually produce object code in libOpenImageIO, which therefore does not need to link against OpenCV).

Applications wanting these three functions should include this header and will be responsible themselves for ensuring that the include paths and linkage of their application makes provisions for finding and using OpenCV. All other applications don't need to deal with OpenCV dependency at all.

Our own oiiotool (for --capture) and our python bindings still link against OpenCV to support this functionality (again, still as an optional dependency, enabled only if OpenCV is found at build time). But downstream apps using libOpenImageIO and who do not themselves need our OpenCV functioality now no longer need to link against OpenCV.

We have removed capture_image() from the Python bindings -- Python scripts that need to capture live camera images can use OpenCV or any other capture API of their choice without going through OIIO.

### DIFF
--- a/docs/Deprecations-3.0.md
+++ b/docs/Deprecations-3.0.md
@@ -86,6 +86,14 @@ about being deprecated will be removed in the final 3.0 release.
   deprecated since OIIO 1.5, now has deprecation warnings. Use
   `interppixel_NDC()` instead.
 
+## imageio.h
+
+* The global OIIO::attribute query "opencv_version" has been removed. The
+  libOpenImageIO library itself no longer has OpenCV as a dependency or links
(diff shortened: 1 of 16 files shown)


lgritz:
Does anybody object to this approach?

PR #4365, opened 2024-08-01
fix: copy/paste error in the ImageBuf iterator copy constructor

antond-weta:
## Description

Fixes a simple copy/paste error in a copy constructor where the y coordinate gets initialised twice instead of y and z.

## Tests

This is a trivial change, I don't think it needs any additional testing.

## Checklist:

  nonconforming.


lgritz (approved):
Oh no, that sure looked bad.

PR #4366, opened 2024-08-01
implementation of linear bayer pattern demosaicing

antond-weta:
## Description

WIP: ImageBufAlgo_demosaic

## Checklist:

  nonconforming.


lgritz:
Rather than have separate bayer_demosaic_linear and bayer_demosaic_MHC and a custom enum here -- which could grow to a larger bayer-related zoo of entry points, arguments, and enum values later -- I'd prefer to employ the new KWArgs idiom like I recently rolled out to IBA::resize(), fit(), and warp().

That would make the function signature look like this:

```c++
ImageBuf OIIO_API bayer_demosaic(const ImageBuf& src, KWArgs options = {},
                                 ROI roi, int nthreads = 0);

bool OIIO_API bayer_demosaic(ImageBuf& dst, const ImageBuf& src, KWArgs options = {},
                             ROI roi, int nthreads = 0);
```

and then the linear vs MHC and bayer pattern can be passed through the keyword args, and any future controls you desire can be added without changing the ABI again. So a call might look like

```c++
    // customize everything
    ImageBuf r = bayer_demosaic(src, { { "algorithm", "linear" }, { "pattern", "RGGB" } },
                                roi, nthreads);

    // defaults across the board
    ImageBuf r = bayer_demosaic(src);
```

looking at fit() or warp() should make it pretty obvious how to extract args from the options on the implementation side.

lgritz:
Yeah, I like it better with the kwargs, thanks.

antond-weta:
I have started porting Kimball's X-Trans demosaicing algorithm, but that seems to be a pretty big task. I think I'll skip that for now and add later as a separate PR.

Vlad, I have made some changes to the public interface, you may need to tweak your code if you have started using this already. There is now a single entry point for all (future) pattern types:
`ImageBuf r = demosaic(src, { { "pattern", "bayer" }, { "algorithm", "linear" }, { "layout", "GBRG" } }, roi, nthreads);`

Also did some performance tweaks.

Larry, Vlad, If you are happy with this code, I'll finish it up by adding some unit tests and documentation.

src/libOpenImageIO/imagebufalgo_test.cpp
```diff
@@ -1264,6 +1264,59 @@ test_simple_perpixel()
...
 
 
+static void
+test_demosaic()
+{
+    print("Testing demosaic\n");
+
+    std::string inputs[] = { "/Users/ad/Desktop/TestRaw/Image1.NEF",
+                             "/Users/ad/Desktop/TestRaw/Image2.DNG",
+                             "/Users/ad/Desktop/TestRaw/Image3.RW2" };
```
lgritz, line 1274: This is obviously not going to work for anybody but you. :-)

Not every IBA function needs to be tested (or can be done in a non-awkward way) in imagebufalgo_test.cpp. That's a decent place to put a self-contained unit test if you can programmatically construct an ImageBuf containing a low-resolution synthetic input designed so that you can inspect its output to be sure it generates the numerically correct results. As you can see from most of the rest of that file.

In general, the more robust tests on real images tend to be the result of implementing the Python bindings and exposing the IBA function as an oiiotool command (both of which will be necessary anyway, as direct use of the C++ APIs are a tiny sliver of typical OIIO use), and then it's much easier to construct more complex python or oiiotool command line based tests on real data (which if they are very big can be stored in the OpenImageIO-images project so as not to bulk up the main source code repo unnecessarily).

lgritz:
The code and approach you have looks good to me, I'm in favor of proceeding with this.

I'm a big believer in "minimum viable product" PRs. They are easier to review and understand and it keeps a big feature coming in steadily rather than being in a single hunk so large that nobody knows if it's correct or good. It's only important that each PR is a step forward, even if a small one, doesn't break any existing functionality that people rely on, and makes some sense as a single logical unit.

So it's fine with me if you try to get this into minimal shape (as long as the API seems like it will stick) and add the other parts in subsequent PRs.

As a reminder, a new IBA function usually needs to be followed with:

* Python bindings
* Expose as an oiiotool command
* Documentation
* Tests

My own style is to do the oiiotool portion fairly early (usually as part of the first PR if I can) because it dramatically improves the convenience of testing. Since oiiotool will call the C++ API underneath, as long as there's an oiiotool-based test, you don't need to test the raw IBA C++ function separately, unless you want to do a small specific unit test that rigorously verifies the correctness by using carefully selected values on a low-rez image and checks for expected results.

antond-weta:
I have added an oiiotool action. It seems to be working fine. I've tested like this: `oiiotool -iconfig raw:Demosaic none -i INPUT_PATH --demosaic:layout=GRBG --mulc 6 -o OUTPUT_PATH`

I've also added the python bindings, but have not tested yet. I'm a total newbie with python. I'm on a Mac, build OIIO and install to /usr/local. How can I make python find my module?

lgritz:
> I've also added the python bindings, but have not tested yet. I'm a total newbie with python. I'm on a Mac, build OIIO and install to /usr/local. How can I make python find my module?

Just make sure that PYTHONPATH env variable includes the site-packages directory of your OIIO build.

src/oiiotool/oiiotool.cpp
```diff
@@ -3420,6 +3420,17 @@ OIIOTOOL_OP(warp, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
...
+OIIOTOOL_OP(demosaic, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
+    std::string pattern   = op.options().get_string("pattern");
+    std::string algorithm = op.options().get_string("algorithm");
+    std::string layout    = op.options().get_string("layout");
+
+    return ImageBufAlgo::demosaic(*img[0], *img[1],
+                                  { { "pattern", pattern },
+                                    { "algorithm", algorithm },
+                                    { "layout", layout } });
+});
```
lgritz, line 3433: Nice! I'm very pleased by how compact it was to turn it directly into an exposed oiiotool action.

src/python/py_imagebufalgo.cpp
```diff
@@ -3033,7 +3060,14 @@ declare_imagebufalgo(py::module& m)
...
         .def_static("make_texture", &IBA_make_texture_ib, "mode"_a, "buf"_a,
-                    "outputfilename"_a, "config"_a = ImageSpec());
+                    "outputfilename"_a, "config"_a = ImageSpec())
+
+        .def_static("demosaic", &IBA_demosaic, "dst"_a, "src"_a,
+                    "pattern"_a = "", "algorithm"_a = "", "layout"_a = "",
+                    "roi"_a = ROI::All(), "nthreads"_a = 0)
+        .def_static("demosaic", &IBA_demosaic_ret, "src"_a, "pattern"_a = "",
+                    "algorithm"_a = "", "layout"_a = "", "roi"_a = ROI::All(),
+                    "nthreads"_a = 0);
```
lgritz, line 3070: This looks correct to me

lgritz:
The oiiotool and python parts look right to me.

antond-weta:
I've added documentation (untested, can't build the docs on my machine).
I've tested the python bindings.
Will add a unit test a bit later.

antond-weta:
Hi Larry,

I've added unit tests, however they fail on all platforms apart from macos-14, which I used to generate the ref/out.txt, due to slight differences in calculations, I guess:
```
-  Mean error = 3.44931e-05
+  Mean error = 3.44911e-05
```

I'm wondering though, assuming all tests in the run.py are supposed to run successfully, do we need the ref/out.txt at all?

lgritz:
Your test is a really clever approach, but the run.py was a little hard to follow and understand how it might be failing, so I took the liberty of pushing an update, in which I modified oiiotool-demosaic/run.py to simplify it and speed it up, and also to write out intermediate files:

- `testimage.exr` is the original test image (color gradient)
- `{pattern}.exr`   is each raw bayer pattern
- `{pattern}-testimage.exr`  is the bayered mosaic of the test image, for each pattern
- `{pattern}-{algo}-result.exr`  is the de-mosaiced version of each bayered test

Now if run test test and cd to build/testsuite/oiiotool-demosaic you can see those temp files left around (that's not harmful) and you compare, say

    iv testimage.exr BGGR-linear-result.exr

zoom in on the upper left corner (left mouse click several times) and then down-arrow to switch between the images, you can see that the results are really great in the interior, but right at the image edges (where you can't use bayer pixel values in all directions) that's where all the significant deviations are.

I don't know much about image debayering, so you'll need to look at those files and make your own judgment. If this is worse than you expect on the edges, maybe the algorithm needs some work? But if this is expected, you may simply need to loosen up the error thresholds to let them pass. OR, keep thresholds tight but check in the approved reference output into a \ref subdirectory and do the comparison on the images (with a tighter threshold). OR, maybe --crop to strip out the border rows and columns so you're doing a rigorous comparison of the interior pixels while allowing for the boundary pixels to be pretty stinky?

lgritz:
But I think the code is all solid and as soon as you decide on the right methodology for how to do the comparison on the test, I think you can change this from draft to grown up PR and we can merge when the tests pass.

antond-weta:
Thanks Larry! I've improved the edge processing a bit. The pixels along the edges are still not as clean as the rest of the image, which is expected. I have decided to ignore the edges when calculating the error, as pretty much all cameras crop the image anyway. This allowed me to lower the error threshold significantly.

I'll mark the PR as ready to go now. Thanks for the help!

lgritz:
Oh, nice, good tight bounds on the diff now.

Looks ready to me, I will merge.

lgritz (approved):
LGTM

PR #4367, opened 2024-08-06
deps: Make OCIO 2.2+ a required dependency

lgritz:
Make OCIO a hard dependency, at least version 2.2.

Revise and improve the OCIO "auto-build" scripts.

Remove OCIO 1.x testsuite reference output.

No longer need FindOpenColorIO.cmake module

As part of this, I split the CI "oldest/hobbled" test into two different tests:

* oldest -- test against the oldest supported versions of all dependencies.
* hobbled -- test against most optional dependencies missing and optional features disabled.


lgritz:
Does anybody object to making OpenColorIO >=2.2 be a required dependency for OIIO 3.0? (Especially since we both have the ability to self-build as well as the build_opencolorio.bash script, so two painless ways to build it if it's not already available.)

Speak now if this will create problems for you, otherwise I will merge this early next week.

PR #4368, opened 2024-08-06
int: Rename env OIIOTOOL_METADATA_HISTORY -> OPENIMAGEIO_METADATA_HISTORY

lgritz:
It's not just in oiiotool. This seems clearer and adheres to the env variable naming convention we chose.

Reminder: This controls whether command line history gets written to output image metadata by default by oiiotool and maketx. We historically did it, but recently stopped because of security concerns.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -53,8 +53,8 @@ using pvt::print_info_options;
 #    define OIIO_UNIT_TESTS 1
 #endif
 
-#ifndef OIIOTOOL_METADATA_HISTORY_DEFAULT
-#    define OIIOTOOL_METADATA_HISTORY_DEFAULT 0
+#ifndef OPENIMAGEIO_METADATA_HISTORY_DEFAULT
+#    define OPENIMAGEIO_METADATA_HISTORY_DEFAULT 0
 #endif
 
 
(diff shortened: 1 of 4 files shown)


src/oiiotool/oiiotool.cpp
```diff
@@ -53,8 +53,8 @@ using pvt::print_info_options;
 #    define OIIO_UNIT_TESTS 1
 #endif
 
-#ifndef OIIOTOOL_METADATA_HISTORY_DEFAULT
-#    define OIIOTOOL_METADATA_HISTORY_DEFAULT 0
+#ifndef OPENIMAGEIO_METADATA_HISTORYDEFAULT
```
jessey-git, line 56: Nit / question: Unsure if the intent was to spell this as "...HISTORYDEFAULT" rather than "...HISTORY_DEFAULT"?

lgritz: Not intentional, that appears to be a search-and-replace problem. Thanks so much for spotting.

jessey-git (approved):
Looks good.  The 1 CI failure for sanitizers looks unrelated; it first popped up a few weeks ago seems like.

lgritz:
Yes the sanitizer failure is happening in all branches, and I've been trying to address it separately.

PR #4372, opened 2024-08-11
test: Adjust RB-2.5 simd_test to allow a little more slop on fast_exp

lgritz:


ThiagoIze:
Why 1e-4 when dev-2.6 is using 1e-5?

lgritz:
It still failed at 1e-5 and I was tired of fighting it.

ThiagoIze (approved):
Yes, it doesn't seem worth the effort to get slightly better accuracy in 2.5 when we already have it working better in 2.6. If we're going to spend time on it, I'd rather spend it further improving the accuracy in 2.6 (probably by making the scalar and simd fast_exp both use or not use FMA.

PR #4373, opened 2024-08-12
fix(oiiotool): `-i:ch=...` didn't fix up alpha and z channels

lgritz:
Fix a bug in oiiotool wherein `-i:ch=...` (input only a subset of channels when reading from a file) was not adjusting the spec's alpha_channel and z_channel -- that is, they still contained the OLD channel numbers of those things, which not only could be wrong due to the reordering, but they might be out of range of the new number of channels.

Also a related fix in the TIFF reader, to fix a possible dereference of an empty vector. This was being triggered by that out-of-range alpha_channel. Maybe that can't happen anymore with the above bug fix, but it still feels like the safer way to proceed.

This was all found by the sanitizers in CI -- and only recently, despite this code having this bug for a long time. So yay for sanitizer CI!

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_channels.cpp
+++ b/src/libOpenImageIO/imagebufalgo_channels.cpp
@@ -76,9 +76,9 @@ ImageBufAlgo::channels(ImageBuf& dst, const ImageBuf& src, int nchannels,
         return false;
     }
 
-    // If channelorder is NULL, it will be interpreted as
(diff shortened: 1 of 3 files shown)


lgritz:
Merging immediately because it demonstrably passes all tests including the sanitizer tests that had previously been failing for the last couple weeks.

If anybody post-reviews this and has concerns, just let us know and we can quickly fix.

PR #4380, opened 2024-08-14
build(deps): Raise libheif minimum to 1.11

lgritz:
This version dates from early 2021 (so still supporting back 3.5 years).

The reason for the bump is to simplify a bit on our part and avoid older versions known to be broken or lacking certain features.

This is for master/2.6/3.0 only, will not be backported to 2.5, since we never raise minimum dependencies in already-released branches.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -59,10 +59,9 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * If you want support for GIF images:
      * **giflib >= 5.0** (tested through 5.2)
  * If you want support for HEIF/HEIC or AVIF images:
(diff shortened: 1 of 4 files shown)


jessey-git (commented):
There 2 additional LIBHEIF_HAVE_VERSION checks that can be removed in heifoutput.cpp as well.

lgritz:
> There 2 additional LIBHEIF_HAVE_VERSION checks that can be removed in heifoutput.cpp as well.

Right, thanks for the reminder.

jessey-git (approved):
Looks good. Approving though watch-out for the clang-format error now.

PR #4381, opened 2024-08-16
fix for setenv() on Windows

ssh4net:
## Description
Microsoft's runtime library doesn't support the standard setenv() function. _putenv_s() should be used instead.

Without fix iv do not compiling:
`E:\GH\OpenImageIO\src\iv\ivmain.cpp(119,9): error C3861: 'setenv': identifier not found [E:\GH\OpenImageIO\build\src\iv\iv.vcxproj]`

## Checklist:

  nonconforming.

### DIFF
--- a/src/iv/ivmain.cpp
+++ b/src/iv/ivmain.cpp
@@ -75,6 +75,7 @@ getargs(int argc, char* argv[])
       .action(ArgParse::store());
     
     ap.parse(argc, (const char**)argv);
+    // clang-format on
     return ap;
 }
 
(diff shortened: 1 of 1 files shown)


lgritz:
Why did this ever work?

src/iv/ivmain.cpp
```diff
@@ -115,8 +115,13 @@ main(int argc, char* argv[])
...
     bool use_ocio = color_space != "" && display != "" && view != "";
     std::string ocioenv = Sysutil::getenv("OCIO");
-    if (ocioenv.empty() || !Filesystem::exists(ocioenv))
-        setenv("OCIO", "ocio://default", 1);
+    if (ocioenv.empty() || !Filesystem::exists(ocioenv)) {
+#ifdef _MSC_VER
+    _putenv_s("OCIO", "ocio://default");
+#else
+    setenv("OCIO", "ocio://default", 1);
+#endif
```
lgritz, line 123: I suspect this needs some indentation to pass clang-format.

lgritz: hahaha, because we turn clang format off on line 39 and forget to turn it back on and the end of that function :-)

ssh4net: added
// clang-format 
on on 78 line

and format a fix code

lgritz:
Vlad, why don't you forget changing the clang-format line. It will trigger too many other problems. I'll do that separately after you merge this. Just stick to the few lines you originally wrote.

ssh4net:
@lgritz is this fine now?

lgritz (approved):
LGTM

lgritz:
I see now, our CI for Windows doesn't find Qt, so doesn't compile iv at all. And I'm guessing that vcpkg doesn't enable iv either. I suppose that's why this wasn't caught before.

PR #4384, opened 2024-08-19
refactor: de-virtualize IC and TS

lgritz:
Background: The ImageCache and TextureSystem had historically been implemented as pure virtual abstract base classe defining an interface, then a hidden concrete subclass. The flexibility this was intended to provide was never really used; instead the imagined main client for the flexibility, OSL, has a different mechanism for customizing the texture system (via their RendererServices class). Meanwhile, this has made the interfaces to IC and TS impossible to change without breaking the ABI.

So this PR does the following:

* ImageCache and TextureSytem are fullly de-virtualized, and are now concrete classes, no hidden subclassing, no virtual methods. As such, removing or changing existing methods will be an ABI break, but adding new methods is not.

* Each now use a PIMPL idiom, where all the data and internal methods are hidden from the public interface and do not affect the ABI.

* Roughly speaking, we added the PIMPL pointers and made the ImageCacheImpl/TextureSystemImpl -- what used to be the hidden concrete subclasses -- into the hidden PIMPL classes.

* Moved a lot of classes that used to be in a "pvt" namespace into the main namespace. If they are opaque types not exposed in public headers or as symbols exported from the library, it doesn't matter, so this just removes some arbitrary clutter.

* Had to rearrange some of the recently added heapsize/footprint code, moved some inline functions in the headers into methods of the classes, defined in the cpp files.

This itself is a huge ABI change, so will only be merged into master, to become part of the 3.0 release. This is not an API change, though! This is all about the internals, and should not require any client software to change a single line of code.

I may do further simplification or refactoring of this code in the future, but that will all be smaller and have no further API/ABI changes that break compatibility.

Although the structure of the class hierarchy has changed around, none of the logic about how IC and TS actually *work* has changed, so there should be no change in observable behavior.

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -25,10 +25,7 @@ OIIO_NAMESPACE_BEGIN
 class ImageBuf;
 class ImageBufImpl;  // Opaque type for the unique_ptr.
 class ImageCache;
-
-namespace pvt {
 class ImageCacheTile;
-};  // namespace pvt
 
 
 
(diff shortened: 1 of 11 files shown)


lgritz:
Comments or concerns?

Note: though this doesn't change any public APIs, it is ABI breaking, and so will only go in master.

lgritz:
Since there were no objections raised, I will merge.

PR #4392, opened 2024-08-26
build(deps): Add libjpeg-turbo recipe

zachlewis:
Build and link missing libjpeg-turbo shared + static libs


lgritz (approved):
LGTM! Thanks!

PR #4394, opened 2024-08-26
build: Add new FFmpeg versions

darbyjohnston:
## Description

Detect new versions of FFmpeg (6.1 and 7.0). The FFmpeg versions page was used for reference:
https://www.ffmpeg.org/download.html#releases

## Tests

Are the FFmpeg versions tested?

## Checklist:

- [x ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/CONTRIBUTING.md).
- [ x] I have updated the documentation, if applicable.
- [ ?] I have ensured that the change is tested somewhere in the testsuite
- [ x] If I added or modified a C++ API call, I have also amended the
  nonconforming.


lgritz:
Just to clarify: it would have been building correctly against those new ffmpeg versions before, but until your fix it would have been mis-reporting the version (in the logs and whatnot) as a release older than it really was.

> Are the FFmpeg versions tested?

I think that a few versions are tested simply because we run on a few different Ubuntu releases in CI (plus the latest ffmpeg should be getting pulled by homebrew in the Mac CI tests, and also I'm implicitly testing yet another version on my work Linux machines), but we don't comprehensively or systematically test against every point on the range of supported ffmpeg versions.

darbyjohnston:
Yes, it was building against the newer versions, that's how I spotted it. I was compiling against FFmpeg 7.0 and wondering how it found 6.0.

I was just wondering if there were some tests I had missed and needed to add.

lgritz (approved):
LGTM, thanks for the fix.

PR #4395, opened 2024-08-27
fix: Stop suppressing fmtlib exceptions

lgritz:
It can't be worse than terminating.


lgritz:
Withdrawing, I have a better solution coming

Issue #4388, opened 2024-08-24
[BUG] assert_fail is called 


DougRogers:
I keep getting assert_fail called in my app.  I am using vcpkg, version  2.5.14.0

I don't have a complete stack track, just 

>	OpenImageIO.dll!fmt::v11::detail::assert_fail(const char * file, int line, const char * message) Line 38	C++
 	OpenImageIO.dll!fmt::v11::detail::fwrite_fully(const void * ptr, unsigned __int64 count, _iobuf * stream) Line 77	C++
 	[Inline Frame] OpenImageIO.dll!fmt::v11::detail::print(_iobuf *) Line 1707	C++
 	OpenImageIO.dll!fmt::v11::vprint_buffered(_iobuf * f, fmt::v11::basic_string_view<char> fmt, fmt::v11::basic_format_args<fmt::v11::context> args) Line 1714	C++

Should I define FMT_THROW to something, so it will throw vs. crash?  It is currently crashing my app.  
I do not have a repro case, it occurs every 10 minutes or so in my app.

Windows 11

![image](https://github.com/user-attachments/assets/c20ef082-41ed-4867-81c8-0c9e0c478de9)

lgritz:
My guess is that somewhere there is an error message or some other formatting that isn't needed very commonly and isn't encountered in the testsuite, in which the arguments don't match the format string.

You seem to be in a debugger, can you go up the stack frame and fine the format/print statement in OIIO that triggers the assert?

Also, in our fmt.h, there's a line that says `#    define FMT_EXCEPTIONS 0`, maybe you can change that to 1 to trigger the exception rather than the terminate?

DougRogers:
It’s a release mode build with symbols in the debugger.  The stack that is displayed in the image is all I have.  I am trying to repro in debug mode, but I have not been able to so far.  

I use vcpkg, I will see if there is a simple way to set that define.

DougRogers:
Is this behavior something that could be moved to runtime vs. compile time?

lgritz:
I think that in sufficiently advanced C++ standard, a bunch of it may become constexpr and thus compile-time checkable. Short of that, I think it is possible to fix this to be compile checkable, if we can find the place where it happens (or the much larger job of fixing EVERY place where strings are formatted).

[...]

PR #4400, opened 2024-08-28
fix: Don't let fmtlib exceptions crash the app

lgritz:
When fmt arguments don't match the format string, fmt will throw an exception, or terminate if we disable exceptions (which we do).  For gcc11+, we tried intercepting these, but let's do it for all platforms, and let's not terminate ourselves, but insted print and log the error.

But, oof, to do this properly, I needed to move some error recording functionality from libOpenImageIO to libOpenImageIO_Util and make it owned more properly by strutil.cpp, so that this all works even when only using the util library. The logic isn't changing, it's just moving over to the other library.

This all helps to address #4388

Unfortunately, the exception thrown by fmt doesn't tell us the bad format string itself. That would have really allowed to probably zero right in on it. But at least we know it's occurring, and one could put a breakpoint on pvt::log_fmt_error to catch it in the act and see where it's being called, revealing the bad line.

### DIFF
--- a/src/include/OpenImageIO/detail/fmt.h
+++ b/src/include/OpenImageIO/detail/fmt.h
@@ -19,13 +19,18 @@
 #    define FMT_EXCEPTIONS 0
 #endif
 
-// Redefining FMT_THROW to something benign seems to avoid some UB or possibly
-// gcc 11+ compiler bug triggered by the definition of FMT_THROW in fmt 10.1+
-// when FMT_EXCEPTIONS=0, which results in mangling SIMD math. This nugget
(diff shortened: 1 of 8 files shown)


lgritz:
No objections, and I can't see how deliberately calling terminate() here is helpful inside a library. So I'm going to merge.

DougRogers:
> Unfortunately, the exception thrown by fmt doesn't tell us the bad format string itself.
I'm not sure the contents of the string are relevant.  It looks like error_msg has a value, then when it is used to do a print, it has been cleared by another thread.  A mutex may fix this crash, but fix is still valid.

PR #4401, opened 2024-09-01
api(IB): Add span-based constuctors to ImageBuf

lgritz:
The flavor of ImageBuf that wraps caller-owned memory is constructed or reset by merely passing a raw pointer and it assumes that the range of memory implied by the strides is entirely safe to read from or write to. And even if correct, there is not much within ImageBuf internals to prevent bugs from buffer overruns.

This PR introduces a new constructor and reset() method to ImageBuf where instead of taking a raw pointer, it takes a span describing the extent of the safely addressible memory (and also an optional "pointer to first pixel" within the buffer, which is necessary if you're using negative strides or for some other reason are describing safe memory that starts before the location of pixel 0). That span is stored in the ImageBuf, and also it remembers if it was a cspan (which should indicate a read-only IB) or a non-const span (which should wrap writable memory).

At this moment, I'm just adding this API and will start using and documenting it as the perferred method, with the raw pointer versions being regarded as unsafe for now (and maybe some day will be deprecated?).  But we're not yet using the span internally for safety checks.

In the future, my intent is to gradually add safeguards to do bounds checking against the span for IB operations that touch the wrapped memory. Because we'll most likely bounds check only in debug mode, it shouldn't ever impact performance, but it will make debugging, sanitizer, and fuzz tests more robust to catch and zero in on buffer overruns. I also like how it forces the caller, at the call site, to think hard about exactly what range of memory the IB is wrapping, and not merely pass a single pointer and hope it works out.

I also changed all ImageBuf ctrs and resets within the OIIO codebase to use the new span-based versions.

This is part of an overall trend to migrate from raw pointer based API calls to span-based API design as much as possible.


lgritz:
Any objections or concerns?

lgritz:
I will merge this tomorrow if nobody has comments by then.

PR #4402, opened 2024-09-02
api: Fold span_util.h contents into span.h

lgritz:
span_util.h is relatively new. I'm honestly no longer sure why I thought it was necessary to make a second file, but so little is in it that I now believe its contents should be folded into span.h to keep all the span-related definitions in once place. It's just not big enough to need to be divided into several headers.


lgritz:
No objections all week -> merging

PR #4409, opened 2024-09-06
api: Adjust IBA::perpixel_op function signature

lgritz:
When I added perpixel_op recently, I didn't get the function signature right. It only would work for plain old function pointers, and would not work properly for "functors" or lambdas with capture. I should have used function_view which can do a light-weight wrapper of all three, so now I do. Also modified imagebufalgo_test to explicitly test all three forms to be sure they work.

Also update function_view ever so slightly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 cmake_minimum_required (VERSION 3.15)
 
-set (OpenImageIO_VERSION "2.6.5.1")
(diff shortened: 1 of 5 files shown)


lgritz:
@ssh4net

ssh4net:
@lgritz just checked both variants with:
[v]  with lambda with a capture list
[v] struct
both are works!

PR #4411, opened 2024-09-08
Ensure that the tests find the built python module vs an install

kdt3rd:
This may have a slightly messy un-intended consequence of "snap-shotting" pythonpath when configured in a large python env, but eases testing confusion on systems where pythonpath is not yet set, but an (older) copy of OpenImageIO is installed.

This fixes the python tests on my home machine.

## Checklist:

  nonconforming.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -110,8 +110,21 @@ macro (oiio_add_tests)
                              "OIIO_TESTSUITE_ROOT=${_testsuite}"
                              "OIIO_TESTSUITE_SRC=${_testsrcdir}"
                              "OIIO_TESTSUITE_CUR=${_testdir}"
-                             "PYTHONPATH=$<SHELL_PATH:${CMAKE_BINARY_DIR}/lib/python/site-packages;$ENV{PYTHONPATH}>"
                              ${_ats_ENVIRONMENT})
+            if (NOT DEFINED ENV{GITHUB_ACTIONS})
+              # the github actions run all the tests from the dist tree
+              # and do an install prior to running the tests so skip
+              # setting these here
+              if (NOT DEFINED ENV{OpenImageIO_ROOT})
+                # this will ensure the appropriate fonts are found
+                set_property(TEST ${_testname} APPEND PROPERTY ENVIRONMENT
+                  "OpenImageIO_ROOT=${CMAKE_SOURCE_DIR}/src")
+              endif()
+              if (USE_PYTHON)
+                set_property(TEST ${_testname} APPEND PROPERTY ENVIRONMENT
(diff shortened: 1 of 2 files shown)


src/cmake/testing.cmake
```diff
@@ -110,6 +110,7 @@ macro (oiio_add_tests)
                              "OIIO_TESTSUITE_ROOT=${_testsuite}"
                              "OIIO_TESTSUITE_SRC=${_testsrcdir}"
                              "OIIO_TESTSUITE_CUR=${_testdir}"
+                             "PYTHONPATH=${CMAKE_BINARY_DIR}/lib/python/site-packages:$ENV{PYTHONPATH}"
```
lgritz, line 113: Does the fact that you used `:` here account for the Windows CI failures?

kdt3rd: most likely, switching to use generator expression which should be fine in cmake 3.14+ - the pathListSep config variable is not available until 3.24 or somesuch

kdt3rd:
With the latest commit, I can cleanly run the test target from the build folder (assuming I have previously downloaded the test images) without running an install, which eases the development workflow. I do not have a windows machine to test with, but windows (and ninja multiconfig) have multiple configs, so to avoid potential slicing, put the python modules in the appropriate config path - and copied the init.py in there such that it might work to load the dlls under windows (untested). Given that the CI runs the tests using the installed dist tree, I have disabled overriding the env variables in this scenario

src/cmake/testing.cmake
```diff
@@ -111,6 +111,20 @@ macro (oiio_add_tests)
...
                              ${_ats_ENVIRONMENT})
+            if (NOT DEFINED ENV{GITHUB_ACTIONS})
+              # the github actions run all the tests from the dist tree
+              # and do an install prior to running the tests so skip
+              # setting these here
+              if (NOT DEFINED ENV{OpenImageIO_ROOT})
+                # this will ensure the appropriate fonts are found
+                set_property(TEST ${_testname} APPEND PROPERTY ENVIRONMENT
+                  "OpenImageIO_ROOT=${CMAKE_SOURCE_DIR}/src")
+              endif()
```
lgritz, line 122: I'm not 100% sure I understand all the implications -- now or after future changes we aren't considering -- of setting OpenImageIO_ROOT to point to the src directory, of all places. 

And what happens if somebody is building in an environment where OpenImageIO_ROOT is already set to someplace other than this build area (maybe to an existing install)? In that case, this would not be set, and what would that mean?

Wouldn't it be more direct to set OPENIMAGEIO_FONTS to exactly the right spot? And much less likely to affect absolutely anything else?

Also, sorry about this nit, but would you mind using 4-space indentation like the surrounding code? Thanks.

PR #4413, opened 2024-09-09
ci: add a workflow that builds docs

lgritz:
This allows somebody who is enhancing the docs to check the results entirely through CI without having to set up the dependencies necessary for docs on their end. (Perfect for Dev Days.) It also ensures that any changes we make to the code base don't break the documentation somehow, because now a successful docs build will be part of the CI.

Note that the docs building job leave behind an artifact, which will be a zip file of the directly containing the HTML docs, just like one would fine on RTD, so you can download that zip file and view the docs it made to verify that everything looks right.

I made the docs builder a separate workflow, but I also changed some things in the CI workflow for the sake of making them more similar.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,8 @@ on:
     paths-ignore:
       - '**.md'
       - '**.rst'
-      - '**.analysis.yml'
+      - '**/analysis.yml'
+      - '**/docs.yml'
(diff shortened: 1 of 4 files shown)


grdanny:
I downloaded the new artifact from the CI job.
this looks great! This is going to be really useful.

grdanny (approved):
lgtm

PR #4415, opened 2024-09-10
api: IOProxy const method adjustments

lgritz:
With 3.0 approaching, it seems like the only time to make these changes that I made note of a while back:

* It seems to me that `tell()` should be const because it isn't going to modify or change the state of the IOProxy.

* It seems to me that `flush()` should NOT be const, because it probably does change internal state.


lgritz:
Does this seem right to you? Is my intuition correct about changing these two methods?

jessey-git (approved):
Agreed, and looks good from here.

PR #4417, opened 2024-09-10
build: Don't change CMAKE_XXX_OUTPUT_DIRECTORY when built as subdir

touraill-adobe:
When building OIIO as a subdirectory of a super-project (with `add_subdirectory`), it is better to honor the output directories specified by the parent project.

In our project, it was causing issues because OIIO libs were ending in a different folder from USD plugins, which thereby failed to load correctly in the build tree.

  nonconforming.


lgritz (approved):
Looks reasonable

lgritz:
Sorry, I didn't notice that the CLA had been fixed up here. Merging.

PR #4418, opened 2024-09-11
test: Fix tests that were writing into ref

lgritz:
openexr_copy and openexr-decreasingy

These recently added test were writing into the ref/ directory, which they certainly should not do. Write into the build test area, compare to what's in the ref area.


lgritz:
The bleeding edge failure is unrelated to this PR. It looks like it's having intermittent failures to install a yum package during dependency setup.

PR #4420, opened 2024-09-13
deps: Add build recipe for ZLIB

zachlewis:
Required for building redistributable python wheels on Windows Wheels runners.


lgritz:
I looked at the "all local builds" CI test, and it's still using the system zlib, not exercising this new builder script.

I think it's because you named this file build_zlib.cmake, but that doesn't match the `checked_find_package(ZLIB)`. Needs to be build_ZLIB.cmake.

I think maybe if it was working for you, it was only because you were using a case-insensitive file system?

lgritz (approved):
LGTM

PR #4421, opened 2024-09-13
deps: Add a build recipe for yaml-cpp

zachlewis:
Apparently necessary for producing working Python binary distributions on Windows (if OpenColorIO is not already available).

I've briefly discussed this in the ASWF/openimageio and OpenColorIO/dev Slack channels. It's not immediately clear why, but the Windows Wheel workflow runners seem to run into trouble when the OCIO build system tries to build the missing yaml-cpp dependency itself. It ends up building both static and dynamic libraries, and getting confused shortly after.

### DIFF
--- a/src/cmake/build_OpenColorIO.cmake
+++ b/src/cmake/build_OpenColorIO.cmake
@@ -15,6 +15,9 @@ set_cache (OpenColorIO_BUILD_SHARED_LIBS  ON
 # it all work with the static dependencies, it just makes things complicated
 # downstream.
 
+checked_find_package (yaml-cpp
+                      VERSION_MIN 0.6.0 )  # Used by OpenColorIO
+
 # Clear variables from the failed find_package
 unset (OPENCOLORIO_LIBRARY)
 unset (OPENCOLORIO_INCLUDE_DIR)
--- a/src/cmake/build_yaml-cpp.cmake
+++ b/src/cmake/build_yaml-cpp.cmake
@@ -1,0 +1,41 @@
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+######################################################################
+# yaml-cpp by hand!
+######################################################################
+
+set_cache (yaml-cpp_BUILD_VERSION 0.8.0 "yaml-cpp version for local builds")
+set (yaml-cpp_GIT_REPOSITORY "https://github.com/jbeder/yaml-cpp")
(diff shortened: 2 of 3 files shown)


lgritz:
I'm slightly confused about why this is necessary. My understanding is that the OCIO build itself will download and build yaml-cpp if not found. So why does OIIO also need to look for and build it?

zachlewis:
Yeah, this is one might not be necessary in the end. It's mostly a workaround for a problem we've been experiencing building OCIO on Windows. 

The exact nature of the problem isn't totally clear to me, since I'm not the one actually experiencing it -- João Silva has been helping me test pip-based builds on Windows, and he ran into trouble with OCIO trying to link the static yaml-cpp libraries it built for itself. He and I have a long thread in the ASWF/openimageio Slack channel that's probably pretty boring to read; and another one in the OpenColorIO/dev channel (https://opencolorio.slack.com/archives/C3Q43TN4Q/p1725017941697509) where I go a little further down the rabbit hole. 

In any case, whatever the underlying problem, it's ideally better dealt with by OCIO, than worked around by OIIO. But given the crunch to get both OCIO-2.4.0 and OIIO-3.0 out the door, presuming the yaml-cpp problem is a blocker to getting Windows wheels built, this is the way I've been handling the issue in the Python wheels PR. 

Anyway, I wanted to get your eyes on this, in case you'd prefer we find another way around the issue.

lgritz:
If the problem is only when OIIO is trying to build OCIO itself, then I would prefer you move the `checked_find_package (yaml-cpp)` from externalpackages.cmake (included/run unconditionally) to putting that inside the build_OpenColorIO.cmake file, so that it only searches for yaml-cpp if it's in that situation where we're building OCIO from scratch. 

As an example of another place that works this way, you can see that in build_TIFF.cmake, we do a `checked_find_package (libdeflate)` because it's a build-time dependency of libtiff.

zachlewis:
Later today, I'm going to try to determine if this PR is indeed necessary.

> If the problem is only when OIIO is trying to build OCIO itself, then I would prefer you move the `checked_find_package (yaml-cpp)` from externalpackages.cmake (included/run unconditionally) to putting that inside the build_OpenColorIO.cmake file, so that it only searches for yaml-cpp if it's in that situation where we're building OCIO from scratch.

On it!

lgritz:
This LGTM now. Just let me know whether it's needed and I will merge, or if it turns out to not be necessary, and we can close it without merging.

zachlewis:
It seems this is _not_ necessary for producing the Windows wheels in the CI environments, so I'm gonna close this. Likewise, I've removed all yaml-cpp stuff from the python_cibuildwheels branch / PR. 

This will likely affect some users' ability to successfully `pip install` from source; one's milage was probably always going to vary with local builds anyway.

Not sure how or why the abi check is failing, but... I guess that's not our problem anymore :)


Thanks for having a look!

Issue #4387, opened 2024-08-22
[FEATURE REQUEST] Add self-building abilities for more dependencies


lgritz:
We recently added (in PR #4242) the ability for OIIO's CMake build system to automatically download and build certain dependencies if they are not found.

(Note: It doesn't happen by default, it requires `OpenImageIO_BUILD_MISSING_DEPS` to be set to a comma-separated list of deps to self-build (or "all"). Also, `OpenImageIO_BUILD_LOCAL_DEPS` can be use to *force* listed dependencies to be built in this way, even if found.)

The original PR #4242 set this up for Imath, OpenEXR, fmt, Robinmap, and OpenColorIO. Subsequent PRs added the self-build ability for Freetype (#4283), libtiff and its libdeflate dependency (#4296), pybind11 (#4297), and improved that of OpenColorIO (#4367).

It would be great to have this ability added for other dependencies! Each can be tackled separately, and each is therefore a fairly self-contained task that doesn't require any prior deep knowledge of the OIIO internals, and could be a good first issue or Dev Days project.

The other high priority dependencies that would be good to add these self-build scripts for are:

- [ ] LibJPEG (maybe unnecessary because we can auto-build jpeg-turbo, which is better?)
- [x] libPNG
- [ ] GIFlib
- [x] LibWEBP
- [ ] Ptex
- [ ] OpenJPEG
- [ ] libheif (may also require libde265 to be made)
- [ ] openjph

If you'd like to work on one of these, just comment below 👇 and say which one you will work on so that two people aren't trying to do the same one. Also feel free to ask below if you have any questions about what needs to be done, where to find examples of the ones that already work, or anything else that would help you get started.

hisergiorojas:
I would like to work on GIFlib

hisergiorojas:
Hi @lgritz 
The GIFLIB project repository at https://gitlab.com/giflib/giflib.git appears to be no longer accessible.

lgritz:
It's one of the foundational libraries of the web. It must live somewhere, even if that particular location is no longer used. I'm very confident that you will be able to find it.

lgritz:
Also, make sure it's not just some kind of temporary glitch. When you run dozens of CI tests a day that download dependencies, you learn pretty quickly that any repo out there (or even a huge service like GitHub) will experience a surprising number of glitches and spurious failures to resolve a repo.

zachlewis:
[...]

PR #4423, opened 2024-09-13
deps: Add build recipe for PNG

zachlewis:
As requested in #4387.

### DIFF
--- a/src/cmake/build_Freetype.cmake
+++ b/src/cmake/build_Freetype.cmake
@@ -17,15 +17,24 @@ set_cache (Freetype_BUILD_SHARED_LIBS  OFF
 
 string (MAKE_C_IDENTIFIER ${Freetype_BUILD_VERSION} Freetype_VERSION_IDENT)
 
+set( _freetype_CMAKE_ARGS
+    -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
+    -D CMAKE_POSITION_INDEPENDENT_CODE=ON
+    -D CMAKE_INSTALL_LIBDIR=lib
+)
+
+# Conditionally disable support for PNG-compressed OpenType embedded bitmaps on Apple Silicon
+# https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4423#issuecomment-2455034286
+if ( APPLE AND ( CMAKE_SYSTEM_PROCESSOR MATCHES "arm64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" ) )
+    list( APPEND _freetype_CMAKE_ARGS -D FT_DISABLE_PNG=ON )
+endif ()
+
 build_dependency_with_cmake(Freetype
     VERSION         ${Freetype_BUILD_VERSION}
     GIT_REPOSITORY  ${Freetype_GIT_REPOSITORY}
     GIT_TAG         ${Freetype_GIT_TAG}
-    CMAKE_ARGS
-        -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
-        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
-        -D CMAKE_INSTALL_LIBDIR=lib
-    )
+    CMAKE_ARGS      ${_freetype_CMAKE_ARGS}
+)
 
 # Set some things up that we'll need for a subsequent find_package to work
 
--- a/src/cmake/build_Freetype.cmake
+++ b/src/cmake/build_Freetype.cmake
@@ -33,7 +33,11 @@ build_dependency_with_cmake(Freetype
     VERSION         ${Freetype_BUILD_VERSION}
     GIT_REPOSITORY  ${Freetype_GIT_REPOSITORY}
     GIT_TAG         ${Freetype_GIT_TAG}
-    CMAKE_ARGS      ${_freetype_CMAKE_ARGS}
+    CMAKE_ARGS
+        -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
+        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
+        -D CMAKE_INSTALL_LIBDIR=lib
+        ${_freetype_EXTRA_CMAKE_ARGS}
 )
 
 # Set some things up that we'll need for a subsequent find_package to work
--- a/src/cmake/build_Freetype.cmake
+++ b/src/cmake/build_Freetype.cmake
@@ -17,16 +17,10 @@ set_cache (Freetype_BUILD_SHARED_LIBS  OFF
 
 string (MAKE_C_IDENTIFIER ${Freetype_BUILD_VERSION} Freetype_VERSION_IDENT)
 
-set( _freetype_CMAKE_ARGS
-    -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
-    -D CMAKE_POSITION_INDEPENDENT_CODE=ON
-    -D CMAKE_INSTALL_LIBDIR=lib
-)
-
 # Conditionally disable support for PNG-compressed OpenType embedded bitmaps on Apple Silicon
 # https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4423#issuecomment-2455034286
 if ( APPLE AND ( CMAKE_SYSTEM_PROCESSOR MATCHES "arm64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" ) )
-    list( APPEND _freetype_CMAKE_ARGS -D FT_DISABLE_PNG=ON )
+    set (_freetype_EXTRA_CMAKE_ARGS -DFT_DISABLE_PNG=ON )
 endif ()
 
 build_dependency_with_cmake(Freetype
--- a/src/cmake/build_PNG.cmake
+++ b/src/cmake/build_PNG.cmake
@@ -1,0 +1,54 @@
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+######################################################################
+# PNG by hand!
+######################################################################
+
+set_cache (PNG_BUILD_VERSION 1.6.44 "PNG version for local builds")
+set (PNG_GIT_REPOSITORY "https://github.com/glennrp/libpng")
+set (PNG_GIT_TAG "v${PNG_BUILD_VERSION}")
+
+set_cache (PNG_BUILD_SHARED_LIBS ${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
+           DOC "Should execute a local PNG build, if necessary, build shared libraries" ADVANCED)
+
+string (MAKE_C_IDENTIFIER ${PNG_BUILD_VERSION} PNG_VERSION_IDENT)
+
+unset (PNG_FOUND)
+unset (PNG_LIBRARIES)
+unset (PNG_INCLUDE_DIRS)
+unset (PNG_INCLUDE_DIR)
+
+build_dependency_with_cmake (PNG
+    VERSION         ${PNG_BUILD_VERSION}
+    GIT_REPOSITORY  ${PNG_GIT_REPOSITORY}
+    GIT_TAG         ${PNG_GIT_TAG}
+    CMAKE_ARGS
+        -D PNG_SHARED=${PNG_BUILD_SHARED_LIBS}
+        -D PNG_STATIC=ON
+        -D PNG_EXECUTABLES=OFF
+        -D PNG_TESTS=OFF
+        -D PNG_FRAMEWORK=OFF
+        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
+        -D CMAKE_INSTALL_LIBDIR=lib
+    )
+
+    
+set (PNG_REFIND FALSE)
+set (PNG_REFIND_VERSION ${PNG_BUILD_VERSION})
+
+
+set (PNG_REFIND_ARGS EXACT REQUIRED)
+
+
+if (PNG_BUILD_VERSION VERSION_GREATER 1.6.43)
+    list (APPEND PNG_REFIND_ARGS CONFIG)
+endif ()
+
+
+find_package(PNG ${PNG_REFIND_VERSION} ${PNG_REFIND_ARGS})
+
+if (PNG_BUILD_SHARED_LIBS)
+    install_local_dependency_libs (PNG png)
+endif ()
--- a/src/cmake/build_PNG.cmake
+++ b/src/cmake/build_PNG.cmake
@@ -19,19 +19,24 @@ unset (PNG_FOUND)
 unset (PNG_LIBRARIES)
 unset (PNG_INCLUDE_DIRS)
 unset (PNG_INCLUDE_DIR)
+unset (PNG_VERSION_STRING)
+unset (PNG_DEFINITIONS)
+unset (PNG_VERSION)
 
 build_dependency_with_cmake (PNG
     VERSION         ${PNG_BUILD_VERSION}
     GIT_REPOSITORY  ${PNG_GIT_REPOSITORY}
     GIT_TAG         ${PNG_GIT_TAG}
     CMAKE_ARGS
-        -D PNG_SHARED=${PNG_BUILD_SHARED_LIBS}
+        -D PNG_SHARED=OFF
         -D PNG_STATIC=ON
         -D PNG_EXECUTABLES=OFF
         -D PNG_TESTS=OFF
+        -D PNG_TOOLS=OFF
         -D PNG_FRAMEWORK=OFF
         -D CMAKE_POSITION_INDEPENDENT_CODE=ON
         -D CMAKE_INSTALL_LIBDIR=lib
+        -D PNG_PREFIX=oiio
     )
 
     
@@ -40,14 +45,22 @@ set (PNG_REFIND_VERSION ${PNG_BUILD_VERSION})
 
 
 set (PNG_REFIND_ARGS EXACT REQUIRED)
-
+set (PNG_DIR ${PNG_LOCAL_INSTALL_DIR}/lib/cmake/PNG)
+set (PNG_FIND_VERSION_EXACT ON)
 
 if (PNG_BUILD_VERSION VERSION_GREATER 1.6.43)
     list (APPEND PNG_REFIND_ARGS CONFIG)
 endif ()
 
+find_package(PNG ${PNG_REFIND_VERSION} ${PNG_REFIND_ARGS}
+             HINTS 
+                    ${PNG_LOCAL_INSTALL_DIR}/lib/cmake/PNG
+                    ${PNG_LOCAL_INSTALL_DIR}
+             NO_DEFAULT_PATH
+            )
 
-find_package(PNG ${PNG_REFIND_VERSION} ${PNG_REFIND_ARGS})
+set (PNG_INCLUDE_DIRS ${PNG_LOCAL_INSTALL_DIR}/include)
+include_directories(BEFORE ${PNG_INCLUDE_DIRS})
 
 if (PNG_BUILD_SHARED_LIBS)
     install_local_dependency_libs (PNG png)
--- a/src/png.imageio/png_pvt.h
+++ b/src/png.imageio/png_pvt.h
@@ -4,7 +4,7 @@
 
 #pragma once
 
-#include <png.h>
+#include <libpng16/png.h>
 #include <zlib.h>
 
 #include <OpenImageIO/Imath.h>
--- a/src/cmake/add_oiio_plugin.cmake
+++ b/src/cmake/add_oiio_plugin.cmake
@@ -70,7 +70,7 @@ macro (add_oiio_plugin)
             target_compile_definitions (${_plugin_NAME} PRIVATE
                                         ${_plugin_DEFINITIONS}
                                         OpenImageIO_EXPORTS)
-            target_include_directories (${_plugin_NAME} PRIVATE ${_plugin_INCLUDE_DIRS})
+            target_include_directories (${_plugin_NAME} BEFORE PRIVATE ${_plugin_INCLUDE_DIRS})
             target_link_libraries (${_plugin_NAME} PUBLIC OpenImageIO
                                                    PRIVATE ${_plugin_LINK_LIBRARIES})
             set_target_properties (${_plugin_NAME} PROPERTIES PREFIX "" FOLDER "Plugins")
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -120,6 +120,11 @@ endif ()
 # we will continue building, but the related functionality will be disabled.
 
 checked_find_package (PNG VERSION_MIN 1.6.0)
+if (TARGET PNG::png_static)
+    set (PNG_TARGET PNG::png_static)
+elseif (TARGET PNG::PNG)
+    set (PNG_TARGET PNG::PNG)
+endif ()
 
 checked_find_package (Freetype
                       VERSION_MIN 2.10.0
--- a/src/ico.imageio/CMakeLists.txt
+++ b/src/ico.imageio/CMakeLists.txt
@@ -2,9 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
-if (TARGET PNG::PNG)
+if (PNG_TARGET)
     add_oiio_plugin (icoinput.cpp icooutput.cpp
-                     LINK_LIBRARIES PNG::PNG ZLIB::ZLIB)
+                     INCLUDE_DIRS ${PNG_INCLUDE_DIRS}
+                     DEFINITIONS ${PNG_DEFINITIONS}
+                     LINK_LIBRARIES ${PNG_LIBRARIES} ${PNG_TARGET} ZLIB::ZLIB)
 else ()
     message (WARNING "libpng not found, so ICO support will not work")
     set (format_plugin_definitions ${format_plugin_definitions} DISABLE_ICO=1 PARENT_SCOPE)
--- a/src/libOpenImageIO/CMakeLists.txt
+++ b/src/libOpenImageIO/CMakeLists.txt
@@ -88,7 +88,7 @@ if (EMBEDPLUGINS)
                                 PRIVATE
                                     EMBED_PLUGINS=1
                                     ${format_plugin_definitions})
-    target_include_directories (OpenImageIO
+    target_include_directories (OpenImageIO BEFORE
                                 PRIVATE ${format_plugin_include_dirs})
 
     # Organize the embedded plugins into source folders
--- a/src/png.imageio/CMakeLists.txt
+++ b/src/png.imageio/CMakeLists.txt
@@ -2,9 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
-if (TARGET PNG::PNG)
+if (PNG_TARGET OR PNG_LIBRARIES)
     add_oiio_plugin (pnginput.cpp pngoutput.cpp
-                     LINK_LIBRARIES PNG::PNG ZLIB::ZLIB)
+                     INCLUDE_DIRS ${PNG_INCLUDE_DIRS}
+                     DEFINITIONS ${PNG_DEFINITIONS}
+                     LINK_LIBRARIES ${PNG_LIBRARIES} ${PNG_TARGET} ZLIB::ZLIB)
 else ()
     message (WARNING "libpng not found, so PNG support will not work")
     set (format_plugin_definitions ${format_plugin_definitions} DISABLE_PNG=1 PARENT_SCOPE)


lgritz:
Looks like we're still failing CI here

zachlewis:
indeed... it's the strangest thing. Still trying to get to the bottom of this. 

I think what's happening is, for systems that have a "too old" version of dynamic PNG libraries installed, now that there's a build recipe available, even if we build newer static PNG libs, the build system is still preferring to link the older shared libraries (even though they're too old). 

I dunno if this is a clue or a red herring, but it seems the build system might be getting confused about which version of PNG it's found versus the version it's trying to build (at least nominally):

```
 Could NOT find PNG: Found unsuitable version "1.5.13", but required is at least "1.6.0" (found /usr/lib64/libpng.so)
-- Building package PNG 1.5.13 locally
--         PNG_BUILD_VERSION = 1.6.44
--         PNG_BUILD_SHARED_LIBS = OFF
--         Building local PNG 1.6.44 from https://github.com/glennrp/libpng
--         Downloading local https://github.com/glennrp/libpng
Note: checking out 'f5e92d76973a7a53f517579bc95d61483bf108c0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at f5e92d7 Release libpng version 1.6.44
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_COMPILE_WARNING_AS_ERROR
    CMAKE_MESSAGE_INDENT
    CMAKE_RULE_MESSAGES


-- Build files have been written to: /__w/OpenImageIO/OpenImageIO/build/deps/PNG-build
[1/31] Generating pngprefix.h
[2/31] Generating scripts/pnglibconf.c
[3/31] Generating pnglibconf.c
[4/31] Generating scripts/symbols.out
[5/31] Generating scripts/symbols.chk
[6/31] Generating pnglibconf.out
[7/31] Generating pnglibconf.h
[8/31] Generating scripts/vers.out
[9/31] Generating scripts/sym.out
[10/31] Generating scripts/prefix.out
[11/31] Generating libpng.sym
[12/31] Generating libpng.vers
[13/31] Generating scripts/intprefix.out
[14/31] Building C object CMakeFiles/png_static.dir/pngmem.c.o
[15/31] Building C object CMakeFiles/png_static.dir/pngrio.c.o
[16/31] Building C object CMakeFiles/png_static.dir/pngpread.c.o
[17/31] Building C object CMakeFiles/png_static.dir/pngerror.c.o
[18/31] Building C object CMakeFiles/png_static.dir/pngget.c.o
[19/31] Building C object CMakeFiles/png_static.dir/pngtrans.c.o
[20/31] Building C object CMakeFiles/png_static.dir/pngset.c.o
[21/31] Building C object CMakeFiles/png_static.dir/png.c.o
[22/31] Building C object CMakeFiles/png_static.dir/pngwio.c.o
[23/31] Building C object CMakeFiles/png_static.dir/pngread.c.o
[24/31] Building C object CMakeFiles/png_static.dir/intel/intel_init.c.o
[25/31] Building C object CMakeFiles/png_static.dir/pngwtran.c.o
[26/31] Building C object CMakeFiles/png_static.dir/pngwrite.c.o
[27/31] Building C object CMakeFiles/png_static.dir/pngrutil.c.o
[28/31] Building C object CMakeFiles/png_static.dir/pngrtran.c.o
[29/31] Building C object CMakeFiles/png_static.dir/intel/filter_sse2_intrinsics.c.o
[30/31] Building C object CMakeFiles/png_static.dir/pngwutil.c.o
[31/31] Linking C static library libpng16.a
[0/1] Install the project...
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/libpng16.a
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/libpng.a
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/png.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/pngconf.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/pnglibconf.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/libpng16/png.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/libpng16/pngconf.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/include/libpng16/pnglibconf.h
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/bin/libpng-config
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/bin/libpng16-config
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/share/man/man3/libpng.3
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/share/man/man3/libpngpf.3
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/share/man/man5/png.5
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/pkgconfig/libpng.pc
-- Up-to-date: /__w/OpenImageIO/OpenImageIO/build/deps/dist/bin/libpng-config
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/pkgconfig/libpng16.pc
-- Up-to-date: /__w/OpenImageIO/OpenImageIO/build/deps/dist/bin/libpng16-config
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/libpng/libpng16.cmake
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/libpng/libpng16-release.cmake
-- Up-to-date: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/libpng16.a
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/cmake/PNG/PNGTargets.cmake
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/cmake/PNG/PNGTargets-release.cmake
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/cmake/PNG/PNGConfig.cmake
-- Installing: /__w/OpenImageIO/OpenImageIO/build/deps/dist/lib/cmake/PNG/PNGConfigVersion.cmake
-- Refinding PNG with PNG_ROOT=/__w/OpenImageIO/OpenImageIO/build/deps/dist
-- Found PNG: /usr/lib64/libpng.so (found version "1.5.13") 
-- Found PNG 1.5.13 
--     PNG_INCLUDE_DIR = /usr/include;/usr/include
--     PNG_INCLUDE_DIRS = /usr/include;/usr/include
--     PNG_LIBRARIES = /usr/lib64/libpng.so;/usr/lib64/libz.so
```

so... in the second line in the above log snippet, it claims to be "Building package PNG 1.5.13 locally", even though it's building 1.6.44. 

I'm not sure what, if anything, that has to do with the more serious problems:
  1. That it can't "refind" the freshly-built static PNG given PNG_ROOT, PNG_REFIND_VERSION, or even PNG_REFIND_ARGS CONFIG 
  2. That it proceeds to link the system-level dynamic PNG 1.5.13 that it previously detected as too old

A lesser problem, but one worth figuring out -- currently, I'm forcing the recipe to only build static libraries, cuz Windows has trouble figuring out what to do with / how to install the dynamic PNG library it just built. (I was having similar trouble on macOS until I set `-DPNG_FRAMEWORK=OFF`; but I've not yet investigated what specifically is (or isn't) happening on Windows. 

Ideally, we'd only link the static libraries, since that seems to behave properly across platforms, given the opportunity... 
I believe the OCIO missing-dependency-build-system business allows one to forcibly prefer static libraries instead of dynamic libraries if both are found (on a per-library basis). I don't suppose there's a way to do that here?

zachlewis:
I'm kind of baffled by the CI build failures... I can't seem to pinpoint anything in the logs or the configs that would explain why or how the failures occur on some runs, but not others... 

The problematic runs are:
- CI / VFX2021 oldest gcc9.3/C++17 py3.7 exr-3.1
- CI / VFX2021 gcc9/C++17 py3.7 exr3.1 ocio2.0
- CI / VFX2022 gcc9/C++17 py39 exr3.1 ocio2.1
- CI / VFX2022 clang13/C++17 py39 avx2 exr3.1 ocio2.1 

And they all fail the same way:

```
FAILED: bin/simd_test 
: && ccache /usr/local/bin/_ccache/clang++ -O3 -DNDEBUG  src/libutil/CMakeFiles/simd_test.dir/Unity/unity_0_cxx.cxx.o -o bin/simd_test  -Wl,-rpath,/__w/OpenImageIO/OpenImageIO/build/lib:/usr/local/lib64:/__w/OpenImageIO/OpenImageIO/build/deps/dist/lib  lib/libOpenImageIO.so.2.6.6  lib/libOpenImageIO_Util.so.2.6.6  -lpthread  /usr/local/lib64/libImath-3_1.so.29.4.0  -lm  -Wl,-rpath-link,/usr/local/lib64:/__w/OpenImageIO/OpenImageIO/build/deps/dist/lib && :
/opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: lib/libOpenImageIO.so.2.6.6: undefined reference to `png_set_eXIf_1'
/opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: lib/libOpenImageIO.so.2.6.6: undefined reference to `png_get_eXIf_1'
/opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../bin/ld: lib/libOpenImageIO.so.2.6.6: undefined reference to `png_set_option'
``` 

But... considering the rest of the checks pass, it doesn't seem to be an issue with the compiler or compiler version, the VFX202X build image, the version of Python, the version of EXR, the version of OCIO, or whether sse instructions are used. And I can't seem to ascertain a particular constellation of settings that produces a failure. 

The only thing I could identify that looked suspicious to me is the ordering of paths in the `-Wl,-rpath-link,/usr/local/lib64:/__w/OpenImageIO/OpenImageIO/build/deps/dist/lib` part of the build command for bin/simd_test. Earlier in the command, the path order is the reverse, with OIIO's build/deps/dist/lib path preceding /usr/local/lib64  (`-Wl,-rpath,/__w/OpenImageIO/OpenImageIO/build/lib:/usr/local/lib64`).

The logs aren't verbose enough for successful builds for me to tell if this is the difference that explains the failures of unsuccessful runs.

zachlewis:
Okay, this doesn't seem to be working properly -- CMake is having trouble finding the correct PNG library, and my understanding is, the issue stems from PNG's own build system idiosyncracies. The failures above occur on systems with libpng v1.5 installed, which are initially detected as too-old; but are later "refound" after find_package fails to find the libraries under $PNG_ROOT. This seems to be a problem with either kitware's FindPNG.cmake, PNG's CMake config, or both. 

This said, it looks like the libpng team has been very actively working on improving and modernizing their CMake-based build system recently, seemingly in anticipation of an imminent libpng 1.8 release. 

Importantly, two weeks ago, libpng v1.6.44 made some improvements and changes to the CMake build system (see https://github.com/pnggroup/libpng/pull/545 among others) -- which is great -- but the difference between 1.6.43 and 1.6.44 seems to have serious implications for how find_package behaves (or fails to behave) and where targets can be found -- namely,  we need find_package to consider the CONFIG introduced in 1.6.44, and likewise set PNG_DIR.

However, _sometimes_, the find_package _still_ fails to (re)find the correct libpng library. Locally, I can produce conditions with the VFX2021 image where find_package is able to find the proper self-built libraries... but only with incremental builds that reuse a build directory, and only after a first build attempt that manages to self-build PNG, but still proceeds to link the too-old system dynamic PNG, as described above. 

We do have other means for finding PNG -- for one, PkgConfig / png_check_module seems to find our PNG without difficulty -- but I understand this might be problematic on Windows. Alternatively, we can use the output of the libpng-config cli tool for setting PNG_LIBRARIES and PNG_INCLUDE_DIRS (and PNG_FOUND)... but setting those values doesn't seem to sufficiently inform the rest of the build process. And I don't really know enough about CMake or building and linking stuff in general to know what exactly to do next.

I feel the best solution involves coercing CMake's own Find mechanism to do the right thing. I'm not sure if that means creating a custom FindPNG.cmake module or patching PNG's CMake config, or what; but I suspect the path of least resistance is contingent on the build system improvements the libpng team is currently working on.

This PR was initially motivated by Wheels workflow tasks failing on vanilla macos-14 runners; and while this build_PNG.cmake recipe does seem to help with building the macos-arm64 wheels, it obviously introduces problems under other conditions. So, for the time being, we're bypassing the PNG plugin entirely for the Wheels / Build MacOS ARM64 tasks until we can figure out a more robust solution for PNG.

lgritz:
@zachlewis Any update on this one?

zachlewis:
I'm afraid not. I'll have another look, but I think I might be stuck... 

The failures occur when the build system detects a too-old version of libpng:
 1. Build PNG locally with build_PNG.cmake
 2. Fail to find "our" locally-built libpng
 3. Build the PNG plugin against the "too old" system libpng
 
I'm trying to find a way to force the build system to always find and link "our" PNG; but `find_package` has trouble finding PNG to begin with. 

Also, libpng 1.6.43 and 1.6.44 behave very differently.

kaarrot:
> The failures above occur on systems with libpng v1.5 installed, which are initially detected as too-old; but are later "refound" after find_package fails to find the libraries...

@zachlewis, I run into similar problem during 'Refind' step - the too old version of OpenEXR has been used even thought the locally we build another version according to the current minimal requirements. What I've found in this case is to explicitly instruct `find_package()` to use the local version instead. I feel [this change](https://github.com/kaarrot/oiio/blob/42b282edde453cd79b5f65eb1762445bf07515e2/src/cmake/dependency_utils.cmake#L429) may work for you as well.

lgritz:
If that change doesn't break anything, I'm fine with it.

What do you think about adding

    PATHS ${pkgname}_LOCAL_INSTALL_DIR NO_DEFAULT_PATH

which would, if I understand correctly, make it look ONLY in the place where we locally built the package, and none of the other usual locations. I believe that might make it impossible to accidentally find the wrong one installed elsewhere in the system.

lgritz:
To clarify, I mean those are potential additional arguments to the find_package that is called in the re-find step.

kaarrot:
> What do you think about adding `PATHS ${pkgname}_LOCAL_INSTALL_DIR NO_DEFAULT_PATH`

I run a quick test replacing `${${pkgname}_REFIND_VERSION} REQUIRED` with:
```
find_package (${pkgname} ${_pkg_UNPARSED_ARGUMENTS} ${${pkgname}_REFIND_ARGS} PATHS ${${pkgname}_LOCAL_INSTALL_DIR} NO_DEFAULT_PATH)
```
Sadly, this gives me the same result as the [original issue](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4493#discussion_r1807115623) - where 'Refind' detects existing system version of library - as if `NO_DEFAULT_PATH` has no effect.
I dug out some comments to also add `NO_CMAKE_FIND_ROOT_PATH` or rely on setting `CMAKE_PREFIX_PATH` but both errored detecting locally built (Robinmap).

> If that change doesn't break anything, I'm fine with it.

So far only[ this version](https://github.com/kaarrot/oiio/blob/42b282edde453cd79b5f65eb1762445bf07515e2/src/cmake/dependency_utils.cmake#L429) gives expected result and builds. Note I removed  `REQUIRED` as it still finds locally built library.
`find_package (${pkgname} ${${pkgname}_REFIND_VERSION} ${_pkg_UNPARSED_ARGUMENTS} ${${pkgname}_REFIND_ARGS})`

kaarrot:
> The failures occur when the build system detects a too-old version of libpng:
> 1. Build PNG locally with build_PNG.cmake
> 2. Fail to find "our" locally-built libpng
> 3. Build the PNG plugin against the "too old" system libpng

I was able to reproduce the problem with a system version of PNG 1.5 installed - which forces to build 1.6 version locally.
The build errors with missing symbols (`png_set_eXIf_1`) while linking some tests (simd_test).

Although I didn't encounter issue of not detecting local build of PNG. For consistency though, I [moved the 'refinding' step](https://github.com/kaarrot/oiio/commit/f898d0c4586431e7432afa6715f0636f66c9427c) into `checked_find_package()`.

To address the missing symbols in simd_test, it appears that [OpenImageIO does not apparently link with PNG](https://github.com/kaarrot/oiio/commit/7d4180210f0f554a90975ea9b7627816f7e5e925) if static version of the library is built locally. I'm not sure why this is not necessary in cases when the system version is used.

zachlewis:
Hey @kaarrot, thanks for having a look!

Super tricky, right?! 

I've been doing a ton of experiments... hard to remember what I've tried and not tried, but I've definitely run into the same trouble as you with Robinmap after setting NO_DEFAULT_PATH in the (re)find_package step. I know I've messed with including ${pkgname}_REFIND_VERSION and REQUIRED in the (re)find_package step as well, to no avail.


I have found something that does seem to work, but I'm not sure if it's the direction we want to go, nor do I understand why this method works where others haven't... In any case, I'll push another commit to see what fails.

With a little help from ChatGPT, I put together a `build_dependency_with_fetchcontent_populate` macro that attempts to do what `build_dependency_with_cmake` does, but with FetchContent_Define and [FetchContent_Populate](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_populate) instead of executing the configure / build / install steps as external processes. In particular, it uses a CMake 3.24 feature, the FetchContent_Define option OVERRIDE_FIND_PACKAGE to force future uses of `find_package` to find the named package exactly where FetchContent_Populate places it.

zachlewis:
hmmm... getting closer...!

Not sure why vfx2022-gcc9 and the vfx2021-hobbled tests are still failing, though...

zachlewis:
Actually, looks like those successes were false positives. re-finding with ${pkgname}_REFIND_VERSION REQUIRED reveals elicits proper failures (in that find_package is still finding the system libpng).

I think this works for me locally cuz I'm using a newer version of CMake.

zachlewis:
Wow. It looks like I mangled the commit history pretty good here, huh. 

Instead of trying to reconcile with main and clean stuff up, I've started a fresh branch here:
https://github.com/zachlewis/OpenImageIO/tree/build_PNG_2

Right now, there's a single added build_PNG.cmake file, which is at parity with the one here (so, same 8 failing checks). 

If / when I have something worth sharing, I'll close this draft PR and open another. 

I'm not sure what the next step here would be. I guess we could try a custom FindPNG module?

lgritz:
If you rebase on top of main and squash the whole thing down to one commit, do you get that nice compact addition of the build_PNG.cmake only?

lgritz:
Still weird here.

You should be able to

    git push origin build_PNG_2:build_PNG --force

zachlewis:
I had no idea I could do that. Should be less weird now.

lgritz:
I've been fooling around with this myself. No matter what I do, it's explicitly including the system libpng.so. I'm starting to wonder if some other dependency is pulling it in under our noses.

kaarrot:
> I've been fooling around with this myself. No matter what I do, it's explicitly including the system libpng.so. I'm starting to wonder if some other dependency is pulling it in under our noses.

Indeed, I've been playing around with the [order of linked libraries](https://github.com/kaarrot/oiio/commit/9fbc24d39fe83d15819a810abae67a3208f2f4cf#diff-25bb08ef30745f790e4ebe6c0ec3c1ee873750d550153682f92d74877fcd7c62R159) for OpenImageIO target. What turns out to help is to place locally png library before ${format_plugin_libs}.

lgritz:
I think I have a fix!

I've got some testing and removing all my debug scaffolding. But if it's really working, I will push some minor changes on top of your commits, if that's ok with you?

kaarrot:
> I  will push some minor changes on top of your commits, if that's ok with you?

Perfect, yes please go ahead.

lgritz:
I pushed an update with the following comment:

---

build: Improve finding locally built PNG

* Adjustments to lock down a better ordering of include directories.
* Always prefer a static library png when available, and when building
  local PNG, for sure only build the static version.
* Give the symbols an "oiio" prefix so they can't link against symbols
  in the wrong library.
* Give a bunch of extra hints about the directory locations and
  config location of the local build to make extra sure it's the
  one found and chosen.

This is still not perfect. It's plagued by one remaining nasty
problem: If there is a too-old version of libpng installed in a system
area, although we will diligently build a newer local copy, it's
extremely hard to ensure that it doesn't accidentally get the headers
corresponding to the old version, because OTHER dependencies may be
correctly found in the system area and therefore put it in the include
path where it will be found rather than the one we intend. I don't
know how to fix this robustly.

For now, it remains the case that the dependency self-build works very
well if the dependency is missing, but can be hit-or-miss if there are
headers for a too-old version installed in the same area as those of
other dependencies.

But at least for now, this setup works fine for all of our CI cases.

lgritz:
If you think this is good enough for now, please change it from draft to ready to review, and give it your review accept, and then I can merge it. We can always add further improvements later.

src/cmake/build_PNG.cmake
```diff
@@ -0,0 +1,67 @@
...
+
+if (PNG_BUILD_VERSION VERSION_GREATER 1.6.43)
+    list (APPEND PNG_REFIND_ARGS CONFIG)
+endif ()
+
+find_package(PNG ${PNG_REFIND_VERSION} ${PNG_REFIND_ARGS}
+             HINTS 
+                    ${PNG_LOCAL_INSTALL_DIR}/lib/cmake/PNG
+                    ${PNG_LOCAL_INSTALL_DIR}
+             NO_DEFAULT_PATH
```
zachlewis, line 59: Have you found that NO_DEFAULT_PATH works as expected? I haven't been having much luck getting find_package to ignore system paths....

src/png.imageio/png_pvt.h
```diff
@@ -4,7 +4,7 @@
 
 #pragma once
 
-#include <png.h>
+#include <libpng16/png.h>
```
zachlewis, line 7: I guess this will break if/when we switch to libpng 1.8 (I'm not sure if libpng 1.7 is going to see an official release).
Should we document this somewhere?

zachlewis (commented):
Good work! This is something I definitely would not have been able to figure out on my own, so I appreciate the help!

zachlewis:
(I don't seem to be able to approve the changes in github, but I most certainly do approve!)

zachlewis:
> Indeed, I've been playing around with the [order of linked libraries](https://github.com/kaarrot/oiio/commit/9fbc24d39fe83d15819a810abae67a3208f2f4cf#diff-25bb08ef30745f790e4ebe6c0ec3c1ee873750d550153682f92d74877fcd7c62R159) for OpenImageIO target. What turns out to help is to place locally png library before ${format_plugin_libs}.

This is very interesting. FWIW, Freetype does seem to require (or at least use if available) PNG, so I wonder if PNG's placement below Freetype here is specifically what's causing problems for us; and it makes me wonder if the build system picking up a dynamic Freetype lib that itself links a dynamic libpng.so is somehow fighting our ability to find _our_ libpng. 


Man, if we can force the build system to build static deps the whole way down for all plugins... :chefs_kiss:

zachlewis:
I may have spoken too soon... 

I'm having trouble building the MacOS ARM64-only Python wheels, wherein we're setting -DCMAKE_OSX_ARCHITECTURES="arm64"

```
  FAILED: lib/libOpenImageIO.3.1.0.dylib
  : && /Applications/Xcode.app/Contents/Developer/usr/bin/g++ -Os -DNDEBUG -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -mmacosx-version-min=11 -dynamiclib -Wl,-headerpad_max_install_names -compatibility_version 3.1.0 -current_version 3.1.0 -o lib/libOpenImageIO.3.1.0.dylib -install_name @rpath/libOpenImageIO.3.1.0.dylib src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_pixelmath.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_channels.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_compare.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_copy.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_deep.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_draw.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_addsub.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_muldiv.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_mad.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_minmaxchan.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_orient.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_xform.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_demosaic.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebufalgo_yee.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/deepdata.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/exif.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/exif-canon.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/formatspec.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/icc.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imagebuf.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imageinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imageio.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imageioplugin.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/imageoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/iptc.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/xmp.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/color_ocio.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/maketexture.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/bluenoise.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/printinfo.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/oiio_gpu.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/libtexture/texturesys.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/libtexture/texture3d.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/libtexture/environment.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/libtexture/texoptions.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/libtexture/imagecache.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/bmp.imageio/bmpinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/bmp.imageio/bmpoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/bmp.imageio/bmp_pvt.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/cineoninput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/Cineon.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/OutStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/Codec.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/Reader.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/CineonHeader.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/ElementReadStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/cineon.imageio/libcineon/InStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dds.imageio/ddsinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/dpxinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/dpxoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/DPX.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/OutStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/RunLengthEncoding.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/Codec.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/Reader.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/Writer.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/DPXHeader.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/ElementReadStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/InStream.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/dpx.imageio/libdpx/DPXColorConverter.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/fits.imageio/fitsinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/fits.imageio/fitsoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/fits.imageio/fits_pvt.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/hdr.imageio/hdrinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/hdr.imageio/hdroutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/iff.imageio/iffinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/iff.imageio/iffoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/jpeg.imageio/jpeginput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/jpeg.imageio/jpegoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/null.imageio/nullimageio.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/openexr.imageio/exrinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/openexr.imageio/exroutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/openexr.imageio/exrinput_c.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/png.imageio/pnginput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/png.imageio/pngoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/pnm.imageio/pnminput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/pnm.imageio/pnmoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/psd.imageio/psdinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/rla.imageio/rlainput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/rla.imageio/rlaoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/sgi.imageio/sgiinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/sgi.imageio/sgioutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/softimage.imageio/softimageinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/softimage.imageio/softimage_pvt.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/targa.imageio/targainput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/targa.imageio/targaoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/term.imageio/termoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/tiff.imageio/tiffinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/tiff.imageio/tiffoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/webp.imageio/webpinput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/webp.imageio/webpoutput.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/zfile.imageio/zfile.cpp.o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/include/OpenImageIO/detail/pugixml/pugixml.cpp.o  -Wl,-rpath,/var/folders/w7/h_1h043d5jlbq4gz7rzxdwdm0000gn/T/tmplu6tsuow/build/lib  lib/libOpenImageIO_Util.3.1.0.dylib  deps/dist/lib/libImath_v_3_1_10_OpenImageIO_v3_1_0.a  deps/dist/lib/libOpenEXR_v_3_2_4_OpenImageIO_v3_1_0.a  deps/dist/lib/libpng16.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  deps/dist/lib/libjpeg.a  deps/dist/lib/libOpenEXR_v_3_2_4_OpenImageIO_v3_1_0.a  deps/dist/lib/libOpenEXRCore_v_3_2_4_OpenImageIO_v3_1_0.a  deps/dist/lib/libpng16.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  deps/dist/lib/libjpeg.a  deps/dist/lib/libwebp.a  deps/dist/lib/libwebpdemux.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  deps/dist/lib/libOpenColorIO_v_2_4_0_OIIO.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  deps/dist/lib/libIlmThread_v_3_2_4_OpenImageIO_v3_1_0.a  deps/dist/lib/libIex_v_3_2_4_OpenImageIO_v3_1_0.a  deps/dist/lib/libtiff.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd  deps/dist/lib/libjpeg.a  deps/dist/lib/libdeflate.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libm.tbd  deps/dist/lib/libwebp.a  deps/dist/lib/libsharpyuv.a  deps/dist/lib/libImath_v_3_1_10_OpenImageIO_v3_1_0.a  -lm  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libexpat.tbd  deps/dist/lib/libpystring.a  deps/dist/lib/libyaml-cpp.a  deps/dist/lib/libminizip-ng.a  -framework ColorSync  -framework CoreFoundation  -framework CoreGraphics  -framework IOKit  deps/dist/lib/libfreetype.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libbz2.tbd  deps/dist/lib/libpng.a  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk/usr/lib/libz.tbd && :
  ld: warning: ignoring duplicate libraries: 'deps/dist/lib/libImath_v_3_1_10_OpenImageIO_v3_1_0.a', 'deps/dist/lib/libOpenEXR_v_3_2_4_OpenImageIO_v3_1_0.a', 'deps/dist/lib/libjpeg.a', 'deps/dist/lib/libpng16.a', 'deps/dist/lib/libwebp.a'
  Undefined symbols for architecture arm64:
    "_png_create_info_struct", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_create_read_struct", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_destroy_read_struct", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_error", referenced from:
        _read_data_from_FT_Stream in libfreetype.a[35](sfnt.c.o)
    "_png_get_IHDR", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_get_error_ptr", referenced from:
        _error_callback in libfreetype.a[35](sfnt.c.o)
        _read_data_from_FT_Stream in libfreetype.a[35](sfnt.c.o)
    "_png_get_io_ptr", referenced from:
        _read_data_from_FT_Stream in libfreetype.a[35](sfnt.c.o)
    "_png_get_valid", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_read_end", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_read_image", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_read_info", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_read_update_info", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_expand_gray_1_2_4_to_8", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_filler", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_gray_to_rgb", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_interlace_handling", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_longjmp_fn", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
        _error_callback in libfreetype.a[35](sfnt.c.o)
    "_png_set_packing", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_palette_to_rgb", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_read_fn", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_read_user_transform_fn", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_strip_16", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
    "_png_set_tRNS_to_alpha", referenced from:
        _Load_SBit_Png in libfreetype.a[35](sfnt.c.o)
  ld: symbol(s) not found for architecture arm64
  clang: error: linker command failed with exit code 1 (use -v to see invocation)
  ```

It seems libpng has had a little trouble with MacOS ARM64 CMake-based builds (https://github.com/pnggroup/libpng/issues/372)

It's suggested in the thread that

**Comment by zachlewis** [comment]
I wonder if the problems I'm experiencing somehow reflect an issue with Freetype, and _not_ PNG. 

I can verify that arm64 libpng libraries _are_ built successfully...

```
$  lipo -info libpng.a 
Non-fat file: libpng.a is architecture: arm64
```

The problem goes away if I **disable support for PNG-compressed OpenType embedded bitmaps in Freetype** by passing the following CMake argument to Freetype's build_dependency_with_cmake command: `-D FT_DISABLE_PNG=ON`

I dunno what a PNG-compressed OpenType embedded bitmap is, or how common they are in the wild... but in my very brief testing, it does seem that oiiotool is able to read and write PNGs and print DroidSans text without issue...

File: src/png.imageio/png_pvt.h (hunk 2 above)
Comment (lgritz) [reply src/png.imageio/png_pvt.h:7]: I'm not sure this line is critical, I was thinking at the time that it might help to guard against accidentally getting libpng15 headers (which was one battle I had at one point).

Let's keep it. I think we'll know as soon as this breaks with libpng18.

File: src/cmake/build_PNG.cmake (hunk 1 above)
Comment (lgritz) [reply src/cmake/build_PNG.cmake:59]: I ended up with a collection of fixes that together worked. It's possible that some of them individually could be reverted, but I ran out of time or patience to continue to experiment with turning the all off one by one. I think this is certainly not harmful, and might be important.

**Comment by lgritz** [comment]
> Good work! This is something I definitely would not have been able to figure out on my own, so I appreciate the help!

I feel like you're giving me too much credit! This was far from a skilled application of actual knowledge. Basically, I just tinkered with changing anything and everything I could think of that might be related, all weekend long over the course of 34 CI runs, until finally getting a combination that picked up the right libraries and passed the tests for each of our test matrix cases. It was very frustrating and I felt like I was flying blind most of the time.

**Comment by lgritz** [comment]
I think your assessment is correct -- Freetype is pulling in its own libpng and the exact order that the linker tries to resolve the symbols is very fragile and can break one or the other.

libpng gives us a build-time option to give a prefix to all the symbols (which is one of the changes I made), which prevents the wrong symbols from being linked, but they don't have a build option to change the name of the *library*, which really would have been handy and solved the whole thing, because then the linker wouldn't pick just one -- it would look like two differently named libraries with totally differently named symbols.

I'm almost tempted to propose a PR against libpng to add this option, but it's a pretty big pain for me to get legal clearance to contribute to a project. It's a one-time bureaucratic cost, so well worth it for a project we have a continued need to contribute to, but it's a high threshold for something where I want to make a 3 line change and then probably never send anything to the project again.

OK, so I think you're on the right track with the idea that for the python wheels, you can probably build freetype locally without the png option for now and see if anybody ever complains. Let's try that as an additional safeguard.

I will also note for the record that for quite some time, I've had it on my list to see if we should jettison libpng entirely in favor of "spng" which is just one .h and one .c file so can easily be vendored, and is supposed to be higher performance read/write as well. I was about one day of frustration away from just doing this and cutting libpng out of our lives. Maybe it's something we should still aim for in the relatively near future to eliminate all future libpng woes.

**Comment by zachlewis** [comment]
oh my god i broke everything

**Comment by zachlewis** [comment]
Never mind, I don't _think_ this one's my fault...

Between now and when things last worked, it looks like our CI / Ubuntu checks are now pulling a newer version of libheif (now 1.19, previously 1.17); and they seem to be running an older version of the OS (now "focal", previously "noble"). It seems unrelated to the changes I've just made (to disable PNG support in Freetype), but I'll happily revert them if needed... but I'm guessing I might see these failures in other PRs now.

> I'm almost tempted to propose a PR against libpng to add this option

That seems like a really good idea. If I have some time and if I can figure out what needs to be done, I'll see if I can submit a PR to the libpng folks. 

> Maybe it's something we should still aim for in the relatively near future to eliminate all future libpng woes.

I'd come across "spng" in my desperation too not long ago, and was gonna ask about it on Slack if we'd completely run out of momentum here... I guess it's really a matter of how much more pain you're willing to put up with if this PR doesn't solve everything foreverish.

**Comment by lgritz** [comment]
Let's just say that for the purpose of this PR, any test matrix entry that only fails the "heif" test is unrelated and won't prevent us from merging. We will investigate that separately.

> I'd come across "spng" in my desperation too not long ago, and was gonna ask about it on Slack if we'd completely run out of momentum here... I guess it's really a matter of how much more pain you're willing to put up with if this PR doesn't solve everything foreverish.

I think if this PR is working for now, we should merge it as-is. 

I think that the most important application of this PR is to help put the final touches on the Python wheel generation so that we are sure it supports PNG, a commonly encountered format (despite being awful).  Correct me if I'm wrong, but since WE (via our GHA workflow) are the ones generating the wheel, don't we have the option to make that workflow uninstall any existing problematic/conflicting dependencies from the VM image before we start our own builds? So we should be able to build conflict-free static libraries we need for the python wheel. 

Sure, somebody building on their own system that contains conflicting dependency versions can still have trouble in some cases, but that's no worse than where we are today, and we can return to the problem to make the build system even more robust after 3.0 release including allowing PyPI installs.

**Comment by zachlewis** [comment]
Let's stick a fork in this sucker. LGTM.

> Correct me if I'm wrong, but since WE (via our GHA workflow) are the ones generating the wheel, don't we have the option to make that workflow uninstall any existing problematic/conflicting dependencies from the VM image before we start our own builds? So we should be able to build conflict-free static libraries we need for the python wheel.

Yes, you're absolutely correct. I was in fact uninstalling homebrew'd installed packages in the Wheels workflow at one point specifically to get rid of Freetype + friends; but upon review, we felt a more robust way to go would be to implement a  `IGNORE_HOMEBREWED_DEPS` build option which appends typical homebrew + macports prefixes to CMAKE_IGNORE_PATH. For your consideration, of course. 

Anyway -- yes, we can do whatever we want. Except install Docker on the MacOS images, apparently.

**Comment by lgritz** [comment]
Certainly, if there's a reliable way to ignore the paths, use that rather than uninstalling (which would take more time).

**Comment by lgritz** [comment]
OK, I will merge this, then. It's no big deal to have further changes to the PNG build if need be.

But give me a couple hours to chase down the heif thing, if I can get that solved I will feel better.

**Comment by lgritz** [comment]
I merged in the unrelated CI fix, and that does appear to have gotten this all unstuck.

File: src/cmake/build_Freetype.cmake [hunk 3]
```diff
@@ -17,15 +17,24 @@ set_cache (Freetype_BUILD_SHARED_LIBS  OFF
...
     VERSION         ${Freetype_BUILD_VERSION}
     GIT_REPOSITORY  ${Freetype_GIT_REPOSITORY}
     GIT_TAG         ${Freetype_GIT_TAG}
-    CMAKE_ARGS
-        -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
-        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
-        -D CMAKE_INSTALL_LIBDIR=lib
-    )
+    CMAKE_ARGS      ${_freetype_CMAKE_ARGS}
+)
```
lgritz, line 37: I think I prefer the clarity of having all the unconditional arguments here, like they are in all the other build_foo.cmake files. So let me make the following suggestion:

```suggestion
    CMAKE_ARGS
        -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
        -D CMAKE_INSTALL_LIBDIR=lib
        ${_freetype_EXTRA_CMAKE_ARGS}
)
```

src/cmake/build_Freetype.cmake
```diff
@@ -17,15 +17,24 @@ set_cache (Freetype_BUILD_SHARED_LIBS  OFF
...
+    -D BUILD_SHARED_LIBS=${Freetype_BUILD_SHARED_LIBS}
+    -D CMAKE_POSITION_INDEPENDENT_CODE=ON
+    -D CMAKE_INSTALL_LIBDIR=lib
+)
+
+# Conditionally disable support for PNG-compressed OpenType embedded bitmaps on Apple Silicon
+# https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4423#issuecomment-2455034286
+if ( APPLE AND ( CMAKE_SYSTEM_PROCESSOR MATCHES "arm64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" ) )
+    list( APPEND _freetype_CMAKE_ARGS -D FT_DISABLE_PNG=ON )
+endif ()
```
lgritz, line 30:
```suggestion
# Conditionally disable support for PNG-compressed OpenType embedded bitmaps on Apple Silicon
# https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4423#issuecomment-2455034286
if ( APPLE AND ( CMAKE_SYSTEM_PROCESSOR MATCHES "arm64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64" ) )
    set (_freetype_EXTRA_CMAKE_ARGS -DFT_DISABLE_PNG=ON )
endif ()
```

lgritz (changes requested):
Suggestion diffs

lgritz:
I believe this is ready to go. The one test failure seems unrelated, let's look into it separately. You ready for me to merge this, @zachlewis ?

lgritz (approved):
LGTM

zachlewis:
Merge away! Thanks again.

PR #4426, opened 2024-09-14
api(IB)!: Add span based ImageBuf methods for getpixel/setpixel

lgritz:
Next round of making safer interfaces. Primarily, this adds new span-baesd versions of ImageBuf::get_pixels, set_pixels, setpixel, getpixel, interppixel, interppixel_NDC, interppixel_bucibuc, interppixel_bicubic_NDC.

The old ones took raw pointers and implied sizes -- these still exist but are considered "unsafe" and aimed at experts and special cases. (If a downstream project wants to be sure not to use them, they can define symbol "OIIO_IMAGEBUF_DEPRECATE_RAW_PTR" before including imagebuf.h to force deprecation warnings.)

The new additions instead take a span -- a combined pointer and length that can be bounds-checked (in debug mode). I strongly encourage people to use the new interface, as it forces the caller to think hard about the exact region of memory that's ok to read from or write to, and then for the OIIO side to honor that and enfoce/debug with bounds checks against that if we add such plumbing.

Along the way, I added some interesting helper functions:

In span.h:
* `span_within()` checks that one byte span lies entirely within another.
* `check_span()` checks if a pointer and length lies entirely within a span.
* `OIIO_ALLOCA_SPAN` is a macro much like OIIO_ALLOCA, except what it returns is a span rather than a raw pointer to the stack-allocated region.

In imageio.h:
* `span_from_buffer()`/`cspan_from_bufffer()` : Given a raw pointer, pixel data type (TypeDesc), and image dimensions and strides, returns a byte stride describing the extent of memory that encompasses all pixels laid out by that description.

I hope the switch to spans is not very onerous for people. I really think that this will help us write more solid and deliberate code to begin with, and when things do go wrong, the various debug-mode checks and assertions that spans provide should help us more quickly identify exactly where buffer overruns and the like occurred.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 cmake_minimum_required (VERSION 3.15)
 
-set (OpenImageIO_VERSION "2.6.6.0")
+set (OpenImageIO_VERSION "2.6.7.0")
 set (OpenImageIO_VERSION_OVERRIDE "" CACHE STRING
      "Version override (use with caution)!")
 mark_as_advanced (OpenImageIO_VERSION_OVERRIDE)
--- a/src/doc/imagebuf.rst
+++ b/src/doc/imagebuf.rst
@@ -163,10 +163,10 @@ Getting and setting pixel values
 .. doxygenfunction:: OIIO::ImageBuf::getchannel
 .. doxygenfunction:: OIIO::ImageBuf::getpixel(int x, int y, int z, float *pixel, int maxchannels = 1000, WrapMode wrap = WrapBlack) const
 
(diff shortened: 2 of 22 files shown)


lgritz:
I'm very curious to hear feedback on this. Do people like the direction this is moving, to generally prefer public APIs that use spans rather than raw pointers? Or do you think I'm going down a dark road?

If this is acceptable, I'd like to get it merged. This is probably the last compatibility breaking change I'll make before tagging a 3.0 beta.

lgritz:
I'm going to merge this by the end of the weekend if nobody objects to the general direction of these changes.

PR #4428, opened 2024-09-16
feat: Python wheels workflow and build backend

zachlewis:
### Summary
This PR is the spiritual successor to #4011. It implements a `scikit-build-core`-based python build-backend, making it possible to use `pip install .` to build from source; and it adds a Github workflow for building with `cibuildwheel` and  publishing to pypi.org binary distributions (bdists) of the Python module / extensions / CLI tools for cpython 3.8-3.13, across major operating systems and architectures. 

When you `pip install OpenImageIO`, pip attempts to retrieve an OpenImageIO bdist from pypi.org for the host's platform / architecture / Python interpreter. If it can't find something appropriate, pip will attempt to build locally from the OpenImageIO source distribution (sdist), downloading and temporarily installing cmake and ninja if necessary.

### PEP-Compliant Packaging: `pyproject.toml`

The `pyproject.toml` file is organized in three parts:
 1. **Package metadata**: standard attributes identifying and describing the Python package, its run-time and build-time requirements, entry-points to executable scripts, and so forth.
 2. **scikit-build-core options**: governs how `pip install ...` interacts with `cmake`. 
 3. **cibuildwheel options**: additional steps and considerations for building, packaging, and testing relocatable wheel build artifacts in isolated environments. 

### Additions to `__ init __.py` 

Previously, we were using a custom OpenImageIO/__ init__.py file to help Python-3.8+ on Windows load the shared libraries linked by the Python module (i.e., the .dll files that live alongside oiiotool.exe under $PATH). 

This PR adds an additional method for loading the DLL path, necessitated by differences between pip-based and traditional CMake-based installs. 

It also adds a mechanism for invoking binary executables found in the .../site-packages/OpenImageIO/bin directory. This provides a means for exposing Python script "shims" for each CLI tool, installed to platform-specific locations under $PATH, while keeping the actual binaries in a static location relative to the dynamic libraries. Upshot is, in `pyproject.toml`,
each item under `[project.scripts]` is turned into a Python script upon installation that behaves indistinguishably to the end user to the CLI binary executable of the same name. 

### Relocatable Binary Distributions with `cibuildwheel` + `repairwheel`

[cibuildwheel](https://github.com/pypa/cibuildwheel) is a widely-used tool for drastically streamlining the process of building, repairing, and testing Python wheels across platforms, architectures, interpreters, and interpreter versions. 

Additionally, the cibuildwheel-based builds set CMAKE_BUILD_TYPE to "MinSizeRel" to optimize for size (instead of speed) -- this seems to shave ~1.5MB off each .whl's size, compared to "Release"

### "Wheels" Github workflow

I straight-up copied `.github/workflows/wheel.yml` from OpenColorIO and made a few OIIO-specific modifications. When pushing a commit tagged v3*, the workflow will invoke a platform-agnostic "build sdist" (source distribution) task, followed by a series of tasks for building OIIO wheels for cpython-3.8-3.13 on Windows, Linux (x86_64 + aarch64, new libstdc++), and MacOS (x86_64 + arm64) and persisting build artifacts; followed finally by a task for publishing the build artifacts to pypi.org

Note: For the sake of simplicity and troubleshooting, I've made as few changes to OpenColorIO's wheel.yml as I could get away with; but in the future, we can also build wheels for the PyPy interpreter, and possibly pyodide. 

Note: A "trusted publisher" must be set up on pypi.org. See https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/

### Other Changes

I made some minor adjustments to `pythonutils.cmake` and `fancy_add_executable.cmake` that only affect scikit-build-core-based installs:
 -  -- namely, on Linux and macOS, I'm setting the INSTALL_RPATH property to point to the relative path to the dynamic libraries, for the Python module and CLI tools, respectively. This helps ensure that pip-based builds and installs from source (as opposed to installs from repaired, pre-built wheels) yield relocatable, importable packages, without needing to mess with $LD_LIBRARY_PATH etc. 

## Tests

`cibuildwheel`  tests if `oiiotool --buildinfo` runs. If that command elicits code zero, it means the "oiiotool" Python script installed by the wheel is able to `import OpenImageIO`; that the actual binary executable `oiiotool` is properly packaged and exists in the expected location (e.g., at `.../site-packages/OpenImageIO/bin`); and that all runtime dependencies are found.

## Inspiration, Credit, Prior Art
 - @aclark4life's and @JeanChristopheMorinPerso's efforts + direction + discussion + advice. See [#3249](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/3249), and [#4011](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4011), as well as JCM's [python_wheels](https://github.com/JeanChristopheMorinPerso/oiio/tree/python_wheels) and [python_wheels_windows](https://github.com/JeanChristopheMorinPerso/oiio/tree/python_wheels_windows) branches. This PR is an attempt to leverage OIIO-2.6+ self-building-dependency features with # 4011's minimalist and modern approach to packaging.
  - OpenColorIO -- I tried to copy as much as I could from @remia et al's fantastic work with all things wheels-related. The __init __.py modifications, the way we're wrapping the CLI tools, and the github Wheels workflow are lifted almost-verbatim from OCIO. Insert pun about reinventing the wheel here. 
  - @joaovbs96's help and patience with testing stuff on Windows.


zachlewis:
An update: 
After much trial and error, I've determined two critically important things re: the wheel workflow runs:

1. The linux x86_64 runners aren't able to perform the wheels tests for the emulated aarch64 builds. 
I'm disabling the manylinux_aarch64 tests for now, and added a #TODO about re-enabling the tests when linux-arm runners are available. 
(I've tested the build artifacts locally, and they behave as expected).

2. The macOS ARM runners are having trouble with the system-installed PNG-1.4. 
(This is what initially prompted #4423, which apparently has its own troubles).


For posterity, here are the relevant parts of the logs:
```
...
    -- Could NOT find PNG (missing: PNG_DIR)
    -- PNG was not found
    -- PNG  is outside the required range 1.6.0...10000.0.0 
    -- PNG library not found 
    --     Try setting PNG_ROOT ?
    --     Maybe this will help:  src/build-scripts/build_libpng.bash

...

 --
    -- =========================================================================
    -- = Dependency report                                                     =
    -- =========================================================================
    --
    -- The following dependencies found externally:
    --     BZip2 1.0.8
    --     Freetype 2.13.3
    --     libjpeg-turbo 3.0.4
    --     pybind11 2.13.6
    --     Python3 3.13.0
    --     ZLIB 1.2.12
    -- The following dependencies were found but were too old:
    --     GIF (found 4, needed >= 5.0)
    --     PNG (found 1.4.12, needed >= 1.6.0)
    -- The following dependencies were not found:
    --     DCMTK
    --     FFmpeg
    --     fmt 7.0...<10.99  (10.2.1 BUILT LOCALLY)
    --     Imath 3.1  (3.1.10 BUILT LOCALLY)
    --     JXL
    --     libdeflate 1.18  (1.20 BUILT LOCALLY)
    --     Libheif
    --     LibRaw
    --     Nuke
    --     OpenColorIO 2.2...<2.9  (2.3.2 BUILT LOCALLY)
    --     OpenCV
    --     OpenEXR 3.1  (3.2.4 BUILT LOCALLY)
    --     OpenJPEG
    --     Ptex
    --     Qt5
    --     Qt6
    --     Robinmap 1.2.0  (1.3.0 BUILT LOCALLY)
    --     TBB
    --     TIFF 4.0  (4.6.0 BUILT LOCALLY)
    --     WebP
    --
    -- =========================================================================
    --
    -- Configuring done (105.4s)
    -- Generating done (0.1s)
    -- Build files have been written to: /tmp/build_oiio_wheels
    *** Building project with Ninja...
    Change Dir: '/tmp/build_oiio_wheels'
  
    Run Build Command(s): /private/var/folders/0g/hj_q_pzx65bbjnslxz9n0src0000gn/T/pip-build-env-fd58mh83/normal/lib/python3.13/site-packages/ninja/data/bin/ninja -v
    
...


    [88/151] /Applications/Xcode_15.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -DEMBED_PLUGINS=1 -DOIIO_FREETYPE_VERSION=\"2.13.3\" -DOIIO_INTERNAL=1 -DOIIO_OPENEXR_CORE_DEFAULT=0 -DOIIO_PYTHON_VERSION=\"3.13.0\" -DOIIO_QT_VERSION=\"\" -DOIIO_TBB_VERSION=\"\" -DOIIO_USE_EXR_C_API=1 -DOpenImageIO_EXPORTS -DUSE_FREETYPE=1 -DUSE_JPEG_TURBO=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/tmp/build_oiio_wheels/include/OpenImageIO -I/tmp/build_oiio_wheels/include -I/tmp/build_oiio_wheels/src/include -I/Users/runner/work/OpenImageIO/OpenImageIO/src/include -I/tmp/build_oiio_wheels/deps/include -I/OpenEXR -isystem /tmp/build_oiio_wheels/deps/dist/include -isystem /tmp/build_oiio_wheels/deps/dist/include/Imath -isystem /tmp/build_oiio_wheels/deps/dist/include/OpenEXR -isystem /Library/Frameworks/Mono.framework/Headers -isystem /opt/homebrew/include -isystem /opt/homebrew/include/freetype2 -Os -DNDEBUG -std=c++17 -arch arm64 -isysroot /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -mmacosx-version-min=11.0 -fPIC -fvisibility=hidden -Wall -Werror -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -fno-math-errno -MD -MT src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o -MF src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o.d -o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o -c /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icoinput.cpp
    FAILED: src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o
    /Applications/Xcode_15.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -DEMBED_PLUGINS=1 -DOIIO_FREETYPE_VERSION=\"2.13.3\" -DOIIO_INTERNAL=1 -DOIIO_OPENEXR_CORE_DEFAULT=0 -DOIIO_PYTHON_VERSION=\"3.13.0\" -DOIIO_QT_VERSION=\"\" -DOIIO_TBB_VERSION=\"\" -DOIIO_USE_EXR_C_API=1 -DOpenImageIO_EXPORTS -DUSE_FREETYPE=1 -DUSE_JPEG_TURBO=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/tmp/build_oiio_wheels/include/OpenImageIO -I/tmp/build_oiio_wheels/include -I/tmp/build_oiio_wheels/src/include -I/Users/runner/work/OpenImageIO/OpenImageIO/src/include -I/tmp/build_oiio_wheels/deps/include -I/OpenEXR -isystem /tmp/build_oiio_wheels/deps/dist/include -isystem /tmp/build_oiio_wheels/deps/dist/include/Imath -isystem /tmp/build_oiio_wheels/deps/dist/include/OpenEXR -isystem /Library/Frameworks/Mono.framework/Headers -isystem /opt/homebrew/include -isystem /opt/homebrew/include/freetype2 -Os -DNDEBUG -std=c++17 -arch arm64 -isysroot /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -mmacosx-version-min=11.0 -fPIC -fvisibility=hidden -Wall -Werror -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -fno-math-errno -MD -MT src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o -MF src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o.d -o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icoinput.cpp.o -c /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icoinput.cpp
    In file included from /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icoinput.cpp:9:
    /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/../png.imageio/png_pvt.h:251:9: error: no matching function for call to 'png_get_iCCP'
            png_get_iCCP(sp, ip, &profile_name, &compression_type, &profile_data,
            ^~~~~~~~~~~~
    /Library/Frameworks/Mono.framework/Headers/png.h:2398:24: note: candidate function not viable: no known conversion from 'png_bytep *' (aka 'unsigned char **') to 'png_charpp' (aka 'char **') for 5th argument
    PNG_EXPORT(png_uint_32,png_get_iCCP) PNGARG((png_const_structp png_ptr,
                           ^
    /Library/Frameworks/Mono.framework/Headers/pngconf.h:1361:58: note: expanded from macro 'PNG_EXPORT'
    #  define PNG_EXPORT(type,symbol) PNG_IMPEXP type PNGAPI symbol
                                                             ^
    In file included from /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icoinput.cpp:9:
    /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/../png.imageio/png_pvt.h:641:13: error: no matching function for call to 'png_set_iCCP'
                png_set_iCCP(sp, ip, "Embedded Profile", 0, icc_profile, length);
                ^~~~~~~~~~~~
    /Library/Frameworks/Mono.framework/Headers/png.h:2405:17: note: candidate function not viable: no known conversion from 'unsigned char *' to 'png_charp' (aka 'char *') for 5th argument
    PNG_EXPORT(void,png_set_iCCP) PNGARG((png_structp png_ptr,
                    ^
    /Library/Frameworks/Mono.framework/Headers/pngconf.h:1361:58: note: expanded from macro 'PNG_EXPORT'
    #  define PNG_EXPORT(type,symbol) PNG_IMPEXP type PNGAPI symbol
                                                             ^
    2 errors generated.
    [89/151] /Applications/Xcode_15.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -DEMBED_PLUGINS=1 -DOIIO_FREETYPE_VERSION=\"2.13.3\" -DOIIO_INTERNAL=1 -DOIIO_OPENEXR_CORE_DEFAULT=0 -DOIIO_PYTHON_VERSION=\"3.13.0\" -DOIIO_QT_VERSION=\"\" -DOIIO_TBB_VERSION=\"\" -DOIIO_USE_EXR_C_API=1 -DOpenImageIO_EXPORTS -DUSE_FREETYPE=1 -DUSE_JPEG_TURBO=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/tmp/build_oiio_wheels/include/OpenImageIO -I/tmp/build_oiio_wheels/include -I/tmp/build_oiio_wheels/src/include -I/Users/runner/work/OpenImageIO/OpenImageIO/src/include -I/tmp/build_oiio_wheels/deps/include -I/OpenEXR -isystem /tmp/build_oiio_wheels/deps/dist/include -isystem /tmp/build_oiio_wheels/deps/dist/include/Imath -isystem /tmp/build_oiio_wheels/deps/dist/include/OpenEXR -isystem /Library/Frameworks/Mono.framework/Headers -isystem /opt/homebrew/include -isystem /opt/homebrew/include/freetype2 -Os -DNDEBUG -std=c++17 -arch arm64 -isysroot /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -mmacosx-version-min=11.0 -fPIC -fvisibility=hidden -Wall -Werror -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -fno-math-errno -MD -MT src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o -MF src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o.d -o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o -c /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icooutput.cpp
    FAILED: src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o
    /Applications/Xcode_15.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -DEMBED_PLUGINS=1 -DOIIO_FREETYPE_VERSION=\"2.13.3\" -DOIIO_INTERNAL=1 -DOIIO_OPENEXR_CORE_DEFAULT=0 -DOIIO_PYTHON_VERSION=\"3.13.0\" -DOIIO_QT_VERSION=\"\" -DOIIO_TBB_VERSION=\"\" -DOIIO_USE_EXR_C_API=1 -DOpenImageIO_EXPORTS -DUSE_FREETYPE=1 -DUSE_JPEG_TURBO=1 -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/tmp/build_oiio_wheels/include/OpenImageIO -I/tmp/build_oiio_wheels/include -I/tmp/build_oiio_wheels/src/include -I/Users/runner/work/OpenImageIO/OpenImageIO/src/include -I/tmp/build_oiio_wheels/deps/include -I/OpenEXR -isystem /tmp/build_oiio_wheels/deps/dist/include -isystem /tmp/build_oiio_wheels/deps/dist/include/Imath -isystem /tmp/build_oiio_wheels/deps/dist/include/OpenEXR -isystem /Library/Frameworks/Mono.framework/Headers -isystem /opt/homebrew/include -isystem /opt/homebrew/include/freetype2 -Os -DNDEBUG -std=c++17 -arch arm64 -isysroot /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.5.sdk -mmacosx-version-min=11.0 -fPIC -fvisibility=hidden -Wall -Werror -Wno-unused-function -Wno-overloaded-virtual -Wno-unneeded-internal-declaration -Wno-unused-private-field -Wno-tautological-compare -Qunused-arguments -Wunknown-warning-option -Wno-unused-local-typedefs -fno-math-errno -MD -MT src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o -MF src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o.d -o src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/ico.imageio/icooutput.cpp.o -c /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icooutput.cpp
    In file included from /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icooutput.cpp:10:
    /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/../png.imageio/png_pvt.h:251:9: error: no matching function for call to 'png_get_iCCP'
            png_get_iCCP(sp, ip, &profile_name, &compression_type, &profile_data,
            ^~~~~~~~~~~~
    /Library/Frameworks/Mono.framework/Headers/png.h:2398:24: note: candidate function not viable: no known conversion from 'png_bytep *' (aka 'unsigned char **') to 'png_charpp' (aka 'char **') for 5th argument
    PNG_EXPORT(png_uint_32,png_get_iCCP) PNGARG((png_const_structp png_ptr,
                           ^
    /Library/Frameworks/Mono.framework/Headers/pngconf.h:1361:58: note: expanded from macro 'PNG_EXPORT'
    #  define PNG_EXPORT(type,symbol) PNG_IMPEXP type PNGAPI symbol
                                                             ^
    In file included from /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/icooutput.cpp:10:
    /Users/runner/work/OpenImageIO/OpenImageIO/src/ico.imageio/../png.imageio/png_pvt.h:641:13: error: no matching function for call to 'png_set_iCCP'
                png_set_iCCP(sp, ip, "Embedded Profile", 0, icc_profile, length);
                ^~~~~~~~~~~~
    /Library/Frameworks/Mono.framework/Headers/png.h:2405:17: note: candidate function not viable: no known conversion from 'unsigned char *' to 'png_charp' (aka 'char *') for 5th argument
    PNG_EXPORT(void,png_set_iCCP) PNGARG((png_structp png_ptr,
                    ^
    /Library/Frameworks/Mono.framework/Headers/pngconf.h:1361:58: note: expanded from macro 'PNG_EXPORT'
    #  define PNG_EXPORT(type,symbol) PNG_IMPEXP type PNGAPI symbol
                                                             ^
    2 errors generated.
```
As you can see, the build correctly finds and identifies the system PNG as "too old", but seems to continue trying to build with `/Library/Frameworks/Mono.framework/Headers/png.h`. 

Two things worth noting:
 1. This isn't causing problems during the `CI / macos-14` runs because a dynamic libpng-1.6 is provided by the homebrew steps
 2. The build_PNG.cmake from #4423 does seem to fix this problem; but I've temporarily removed the build recipe from this PR so I could better identify and isolate the various problems causing check failures.

(I've also removed the yaml-cpp build recipe from this PR, because it's apparently not as required for building Windows wheels as I'd previously thought.)


There are some CI workflow failures that I can't make heads nor tails of -- they don't seem related to anything this PR is affecting. It's possible the failures reflect something I screwed up when merging the main branch into this branch. This PR has gotten kind of hairy, and it might be worth my time to create a new, cleaner PR, with all changes rebased on top of the current main branch.

zachlewis:
Alright -- to move things along, I've disabled the PNG plugin for the macos-arm64 wheels. Annoying. We'll need to address this in a future PR. 

Here's where we're at:

- All the Wheel / Build checks are finally passing.
- The CI / ABI check is failing, which is fine for now.
- The CI / VFX2021 oldest and VFX2021 hobbled checks are failing, which is not fine for now.

The next steps:
 - Figure out what's going on with those "oldest" and "hobbled" CI checks
 - Test the pypi.org stuff in a local branch / test pypi server
 - Figure out why the macos-x86_64 wheels are twice as large as they should be. (We're trying not to build universal binaries)
  - Clean up, simplify, clarify, and document whatever needs a little more love

aclark4life:
Thank you @zachlewis! Looking good 👍 

> This PR is the spiritual successor to #3249.

I view it as the spiritual successor to #4011 both of which attempt to resolve #3249 and were based on the work of @JeanChristopheMorinPerso [here](https://github.com/JeanChristopheMorinPerso/oiio/tree/python_wheels) and [here](https://github.com/JeanChristopheMorinPerso/oiio/tree/python_wheels_windows). 😄 

> The next steps:

I have a [Test PyPI account](https://test.pypi.org/user/aclark/) so maybe we can try to test this there. Also reviewing [all things self-building](https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4387).

tasks.py
```diff
@@ -0,0 +1,35 @@
...
+
+from pathlib import Path
+import shutil
+import invoke
+from repairwheel._vendor.auditwheel.wheeltools import InWheelCtx
+
+
+DIRS_TO_REMOVE = ['lib', 'lib64', 'share', 'include']
+
+def _wheel_edit(wheel_path, dirs_to_remove=DIRS_TO_REMOVE):
```
JeanChristopheMorinPerso, line 13: I think we can avoid this by using cmake components. In OpenEXR, we did https://github.com/AcademySoftwareFoundation/openexr/commit/222bc3d44959b8eabd135f6cb4c0adffee835394. Basically, the cmake build system should only install what's needed. That's the most reliable way to manage this and also the recommended way from the scikit-build-core team.

JeanChristopheMorinPerso (commented):
Thanks @zachlewis and great work! I did a quick review pass and left some comments. Some minors and some others that I think are worth fixing (numpy and audit wheel mainly). I can do a more in depth review in a week if you want, just let me know!

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,416 @@
...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cibw-wheels-${{ matrix.python }}-${{ matrix.manylinux }}
+          path: ./wheelhouse/*.whl
+
+  # ---------------------------------------------------------------------------
+  # Linux ARM Wheels
+  # ---------------------------------------------------------------------------
+
+  linux-arm:
```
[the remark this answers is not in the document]
zachlewis: So, the reason why I broke this out into a separate job is cuz I want to make it as straightforward as possible to switch the aarch64 builds to a linux ARM runner as soon as they become available -- in a few months, I'm guessing (ARM runners were recently made widely available to enterprise accounts; maybe ASWF has access?). These emulated builds take an order of magnitude longer than the native builds.

The other main difference between the linux x86_64 and aarch64 jobs is that I'm skipping the tests for the aarch64 jobs. It seems CIBW sometimes has trouble running tests on cross-compiled wheels. I've verified locally that the aarch64 build artifacts behave properly.

(likewise, the macos-arm64 tests failed until I moved the builds to the macos-14 (ARM) runner)

Beyond that, I do have some thoughts about other ways we could run these builds -- I think I'd like to switch to incremental builds governed by the CIBW config rather than discrete github jobs -- but for this initial PR, I thought I'd try to match as closely as possible to what OCIO is doing (simply cuz I'm familiar with OCIO, and things seem to be working pretty well over there).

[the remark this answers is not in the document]
remia: I don't remember the exact reasons why we split the jobs like this, I like the simplicity it gives when reviewing workflows and seeing what works or doesn't at a single glance. One other benefits is that it runs in parallel which is important as OCIO is a bit long to compile. But maybe cibuildwheel can do that too from a single job?

OCIO setup is working but might be a bit old at this point and no longer following latest recommendations. I think we should move to using scikit-build for example.

Amazing work @zachlewis by the way! Will try to have a look this week.

tasks.py
zachlewis: Yeah, I noticed you'd done this with OpenEXR...

The real reason for doing what I'm doing here is to get rid of the versioned copies of the libraries -- libOpenImageIO.so.2.6, etc. -- cuz the symlinks get copied over to the wheel as discrete files. It just happened to be a convenient place to strip out everything else at the same time, too.

Would using components prevent those versioned copies from getting created?

The other thing is, we're building libOpenImageIO as a dynamic library, so it can be shared by the Python module and the CLI tools; and we're also building OpenColorIO as a dynamic library for the time being "to prevent problems downstream" (Larry can elaborate). I think in the end, we'd only use components to shave off the headers, CMake configs, and fonts -- and even those things are things I'd like to provide some day as part of the bdist, if we can correct the CMake configs and provide helpers for downstream builds.

If we can get rid of the "versioned copies" of the libraries, whether by using components or some other method, we could skip the repair-wheel step entirely.

.github/workflows/wheel.yml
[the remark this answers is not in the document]
zachlewis: > But maybe cibuildwheel can do that too from a single job?

Sort of. The incremental builds run in serial, but they can share a build cache. If you use a persistent build directory, I think it'll reuse the static dependency libs from the first build for all the subsequent Python module builds (although I haven't tried it myself). It is particularly helpful here, because the OIIO build system is also building OCIO, OpenEXR, TIFF, and handful of others (~12 min on the current runners), all of which can just be built once and reused for all the module builds. It doesn't take too much longer (unless you're doing emulated builds), and it seems like it would save a lot of energy...

zachlewis:
Okay, I think I was able to prevent OIIO from linking x265. It seems the system TIFF was built against WebP... so now we're always building all missing dependencies (on MacOS), and LINKSTATIC is enabled, in order to force the build system to pick up our locally-built static dependencies, where possible. We're also forcibly uninstalling the runner image's Freetype, (installed with homebrew) in order to avoid linking some of its dependencies.

The x86_64 .dylibs directory looks like this:
```
.
├── libbrotlicommon.1.1.0.dylib
├── libbrotlidec.1.1.0.dylib
├── libbrotlienc.1.1.0.dylib
├── libhwy.1.2.0.dylib
├── libIex-3_2.31.3.2.4.dylib
├── libIlmThread-3_2.31.3.2.4.dylib
├── libImath-3_1.29.11.0.dylib
├── libjpeg.8.3.2.dylib
├── libjxl.0.11.0.dylib
├── libjxl_cms.0.11.0.dylib
├── libjxl_threads.0.11.0.dylib
├── liblcms2.2.dylib
├── libOpenColorIO_v_2_3_2_OIIO.2.3.2.dylib
├── libOpenEXR-3_2.31.3.2.4.dylib
├── libOpenEXRCore-3_2.31.3.2.4.dylib
├── libOpenImageIO.2.6.7.dylib
├── libOpenImageIO_Util.2.6.7.dylib
├── libpng16.16.dylib
├── libsharpyuv.0.1.0.dylib
├── libwebp.7.1.9.dylib
└── libwebpdemux.2.0.15.dylib
```

This is still larger than the ARM64 wheels, and some of these dependencies seem to be coming from the system level (libwebp, ...), and others from homebrew (libpng16, jpeg i think). Interestingly, OCIO is linking the shared lcms2 library, and probably the shared IMath library as well, both of which are provided by the runner image. 

As long as there are no GPL-licensed dependencies, I'm cool with leaving this as-is.


I believe at this point, the only failing check is the CI / bleeding edge run.

.github/workflows/wheel.yml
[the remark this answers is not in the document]
remia: Interesting. Assuming you have some limitations like you can't re-use the same dependencies build between manylinux 2014 and manylinux 2.18 for example?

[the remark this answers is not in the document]
zachlewis: Yeah, you do -- I guess "persistent build directory" is kind of a loaded term. I'm setting `$SKBUILD_BUILD_DIR=/tmp/build_oiio`, which, for linux, is container-specific storage (as opposed to a volume mapped to host storage), so it kind of works itself out neatly. On MacOS, things get a bit hairier, as far as universal2 / multi-arch builds are concerned. But as long as I'm using a macos-14 runner for the ARM64 wheels, and macos-13 or macos-12 runners for the x64 wheels, there doesn't seem to be a problem.

tasks.py
JeanChristopheMorinPerso: >Would using components prevent those versioned copies from getting created?

No, for that you'd have to if/else setting SOVERSION. [Here](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/a946288db47df47b257bf79732a0de40ccf1adae/src/libutil/CMakeLists.txt#L99-L100) and [here](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/a946288db47df47b257bf79732a0de40ccf1adae/src/libOpenImageIO/CMakeLists.txt#L218-L219) I think.

Using components would only allow you to only install what you want and nothing else. That with not setting VERSION and SOVERSION I think is the cleanest and most future proof and easy to maintain option.

.github/workflows/wheel.yml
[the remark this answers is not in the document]
JeanChristopheMorinPerso: >So, the reason why I broke this out into a separate job is cuz I want to make it as straightforward as possible to switch the aarch64 builds to a linux ARM runner as soon as they become available -- in a few months, I'm guessing (ARM runners were recently made widely available to enterprise accounts; maybe ASWF has access?). These emulated builds take an order of magnitude longer than the native builds.

You already set `arch` as a variant, so you should be able to do

```yaml
runs-on: ${{ matrix.runs-on }}
strategy:
  matrix:
    include:
    - build: CPython 3.8 ARM 64 bits manylinux2014
      manylinux: manylinux2014
      python: cp38-manylinux_aarch64
      arch: aarch64
      runs-on: <image tag>
```

Anyway, I think the way it is is fine. It's just duplicate configs, but if Larry is fine with that, then that's what's important.

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,423 @@
...
+        uses: pypa/cibuildwheel@d4a2945fcc8d13f20a1b99d461b8e844d5fc6e23 # v2.21.1
+        env:
+          CIBW_BUILD: ${{ matrix.python }}
+          CIBW_ARCHS: ${{ matrix.arch }}
+          # TODO: Re-enable HEIF when we provide a build recipe that does 
+          # not include GPL-licensed dynamic libraries.
+          USE_Libheif: 'OFF'
+          # Make sure the OIIO build's version of freetype is linked
+          CIBW_BEFORE_ALL: |
+            brew uninstall --ignore-dependencies harfbuzz fontconfig cairo freetype
```
JeanChristopheMorinPerso, line 283: It would be better to force CMake to ignore brew's provided dependencies. I thnink this can be achieved using https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_USE_CMAKE_SYSTEM_PATH.html#variable:CMAKE_FIND_USE_CMAKE_SYSTEM_PATH (see https://cmake.org/cmake/help/latest/command/find_package.html#index-0-variable:CMAKE_INSTALL_PREFIX).

pyproject.toml
```diff
@@ -0,0 +1,125 @@
...
+    "Topic :: Multimedia :: Video",
+    "Topic :: Multimedia :: Video :: Display",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+requires-python = ">= 3.8"
+
+[project.urls]
+Homepage = "https://openimageio.org/"
+Source = "https://github.com/AcademySoftwareFoundation/OpenImageIO"
+Documentation = "https://openimageio.readthedocs.io"
```
JeanChristopheMorinPerso, line 36:
```suggestion
Documentation = "https://docs.openimageio.org"
```
I'm not sure if OIIO have fully switched to their vanity URL yet though.

lgritz: I think the new URLs all work. I believe this can be verfied with any browser. But it does look like there are a few places in the docs where we use the rtd URL instead of the openimageio.org link. Those should be updated.

.github/workflows/wheel.yml
zachlewis: Interesting... I'll look into both of these approaches. Thanks for the tips!
Seems like it should just be a matter of appending /opt/homebrew to CMAKE_IGNORE_PATH...

.github/workflows/wheel.yml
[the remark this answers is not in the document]
zachlewis: oooh, I see what you're saying. That's a pretty good idea.

.github/workflows/wheel.yml
zachlewis: I added a IGNORE_HOMEBREWED_DEPS option which appends homebrew prefixes to the CMAKE_IGNORE_PATH. I also made a slight modification to the src/cmake/dependency_utils.cmake:build_dependency_with_cmake, to make sure dependency builds inherit CMAKE_IGNORE_PATH too. It seems to work alright.

tasks.py
aclark4life, line 1: Well played 😄

zachlewis: ha -- we'll see... what I've been trying has apparently broken everything. I'm not sure if I have enough time to debug this myself. We may have to split off the attempt to get rid of tasks.py into its own PR. Right now, it's solving a few problems at once.

JeanChristopheMorinPerso:
Great work with the latest changes @zachlewis! I like how the extract step to fix the wheels are not needed anymore.

I started to review the generated artifacts. I see that the versioned libraries are not yet fully removed:
```
$ unzip -l openimageio-3.1.0.0-cp313-cp313-linux_x86_64.whl | grep libOpenImageIO
 11355768  2024-10-17 14:19   OpenImageIO/lib/libOpenImageIO.so
 11355768  2024-10-17 14:19   OpenImageIO/lib/libOpenImageIO.so.3.1.0
  1171672  2024-10-17 14:19   OpenImageIO/lib/libOpenImageIO_Util.so
  1171672  2024-10-17 14:19   OpenImageIO/lib/libOpenImageIO_Util.so.3.1.0
```

I also see
```
    33700  2024-10-17 14:11   OpenImageIO/share/doc/OpenImageIO/CHANGES-0.x.md
   223954  2024-10-17 14:11   OpenImageIO/share/doc/OpenImageIO/CHANGES-1.x.md
    60831  2024-10-17 14:11   OpenImageIO/share/doc/OpenImageIO/CHANGES.md
    11357  2024-10-17 14:11   OpenImageIO/share/doc/OpenImageIO/LICENSE.md
    13418  2024-10-17 14:11   OpenImageIO/share/doc/OpenImageIO/THIRD-PARTY.md
    42480  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSans-Bold.ttf
    41028  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSans.ttf
   117072  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSansMono.ttf
    48880  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSerif-Bold.ttf
    45652  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSerif-BoldItalic.ttf
    40416  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSerif-Italic.ttf
    43648  2024-10-17 14:11   OpenImageIO/share/fonts/OpenImageIO/DroidSerif.ttf
```
Are these really needed?

JeanChristopheMorinPerso:
About the rpath, if I had to take a guess, I would guess that the culprit is https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c3c6b1e01b12c9ebaef343e55c61500357920595/src/cmake/compiler.cmake#L657-L658

zachlewis:
Hey @JeanChristopheMorinPerso  -- once again, I can't tell you how much I appreciate you having a look!

I apologize, I left things in an inconsistent state over the weekend. Good news is, I do believe I've fixed remaining issues...!

Major changes: 
 1. Added a NAMELINK_SKIP option to the custom install_targets macro in compiler.cmake, and I'm using that option during scikit-build-core builds for installing libOpenImageIO and libOpenImageIO_Util libraries
 2. Changed the default CMAKE_INSTALL_RPATH to $ORIGIN:$ORIGIN/lib
 3. Explicitly set INSTALL_RPATH to $ORIGIN/lib or $ORIGIN/../lib for certain targets
 4. Re-enabled the default CIBW step. (Fixes bad filenames)
 
 

> About the rpath, if I had to take a guess, I would guess that the culprit is
> 
> https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c3c6b1e01b12c9ebaef343e55c61500357920595/src/cmake/compiler.cmake#L657-L658

You were totally right about this -- now I'm doing this:

```cmake
if (NOT CMAKE_INSTALL_RPATH)
        if(APPLE)
            set(BASEPOINT @loader_path)
        else()
            set(BASEPOINT $ORIGIN)
        endif()
        set (CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
endif ()
```


And now the OIIO libs have a relative RPATH:
```
[zach@alma OpenImageIO]$ readelf -d lib/libOpenImageIO.so.3.1.0 

Dynamic section at offset 0xc5df10 contains 34 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libOpenImageIO_Util.so.3.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libz.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libOpenColorIO_v_2_3_2_OIIO.so]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [libOpenImageIO.so.3.1.0]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN:$ORIGIN/lib]
 ```

:metal:


I've gotten rid of the docs. For the fonts, actually, what I'd like to do adjust __init __.py and append the relative dir to the env var "[OPENIMAGEIO_FONTS](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/3b70f1d2d2fa82e0753017103535d38c3cec3855/src/libOpenImageIO/imagebufalgo_draw.cpp#L737C10-L737C27)". I like the idea of a cross-platform default font shipping with the module...

File: src/cmake/compiler.cmake [hunk 5]
```diff
@@ -655,7 +655,12 @@ if (CMAKE_SKIP_RPATH)
     unset (CMAKE_INSTALL_RPATH)
 else ()
     if (NOT CMAKE_INSTALL_RPATH)
-        set (CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}")
+        if(APPLE)
+            set(BASEPOINT @loader_path)
```
JeanChristopheMorinPerso, line 659: This is kind of reverting what was made before. You could pass `-DCMAKE_INSTALL_RPATH` on a per platform basis instead.

JeanChristopheMorinPerso:
Thanks @zachlewis. I'm always happy to help!

MrLixm:
Hey there,
First I want to say thanks to everyone involved in those wheels, the outcome will really be invaluable for some of us.

By curiosity I wanted to test early some those wheels so I downloaded one from the last action's artefacts. However, once installed I do notice that the whell doesn't include support for `png`, I only see:
```
openexr,tiff,jpeg,bmp,cineon,dds,dpx,fits,hdr,iff,null,pnm,psd,rla,sgi,softimage,targa,term,zfile
```
Looking at the previous comments I see that there was some trouble with png so I assume its removal is intentional for now. However, I see those comments seems to be for MacOS and I'm currently testing it on Windows. Did I just misunderstand and png is not yet available for all platform ?

Job I used: https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/11464854677/job/31902028113
_(wheel openimageio-3.1.0.0-cp310-cp310-win_amd64.whl)_

Feel free to ignore as I fully understand that this PR is still wip and I'm not supposed to use those wheels yet.
Best,
Liam.

aclark4life:
> Feel free to ignore as I fully understand that this PR is still wip and I'm not supposed to use those wheels yet.

Feedback welcome! @zachlewis or @JeanChristopheMorinPerso can probably provide more detail but yeah I think PNG is a WIP and FWIW here's Pillow's PNG build: https://github.com/python-pillow/Pillow/blob/main/.github/workflows/wheels-dependencies.sh

zachlewis:
Heya -- 

This PR doesn't intend to do anything specific with PNG (or any other dependency) per se -- it assumes that all dependencies are provided by OIIO's own dependency self-building mechanisms.  The reason why PNG isn't included yet is because we're still working on getting the PNG self-builds to work (see #4423). Once a working solution is merged, the Python wheels will automatically inherit PNG support.


(It's true, not long ago, we were disabling PNG for the MacOS ARM wheels earlier here, but that was just a temporary workaround for handling conflicting homebrew-managed libraries. Now we have a more robust CMake option `IGNORE_HOMEBREWED_DEPS` which forces the build system (and dependency sub-builds) to ignore typical homebrew prefixes.)


As for the status of this PR, I don't really intend to change anything, unless there's something specific you guys would like me to change.

pyproject.toml
```diff
@@ -0,0 +1,132 @@
...
+skip = [
+    # Skip 32-bit builds
+    "*-win32",
+    "*-manylinux_i686",
+    # Building with musl seems to work, but the repair-wheel step seems to fail...
+    # This may be a bug in repairwheel (or auditwheel)?
+    "*musllinux*",  
+]
+test-command = "oiiotool --buildinfo"
+# Prevent CIBW's default repair-wheel command from running, because it fails to
```
JeanChristopheMorinPerso, line 109: I think this is outdated right?

pyproject.toml
```diff
@@ -0,0 +1,132 @@
...
+# recognize that the wheel isn't broken to begin with, and ends up creating
+# additional copies of the shared libraries in the wheel.
+
+[tool.cibuildwheel.macos.environment]
+# Specify a persistent build directory so the repairwheel command can (re)find
+# dependency libraries built by OpenImageIO.
+SKBUILD_BUILD_DIR = "/tmp/build_oiio_wheels"
+# Always (re)build the TIFF library; see #4439 for more details.
+SKBUILD_CMAKE_ARGS = "-DOpenImageIO_BUILD_LOCAL_DEPS=TIFF; -DOpenImageIO_BUILD_MISSING_DEPS=all; -DLINKSTATIC=1; -DIGNORE_HOMEBREWED_DEPS=1; -DCMAKE_INSTALL_LIBDIR=lib"
```
JeanChristopheMorinPerso, line 118: Is it necessary to pass build missing here since it's already defined at the top level?

src/python/__init__.py
```diff
@@ -2,16 +2,50 @@
...
 
-# This works around the python 3.8 change to stop loading DLLs from PATH on Windows.
-# We reproduce the old behaviour by manually tokenizing PATH, checking that the directories exist and are not ".",
-# then add them to the DLL load path.
-# This behviour can be disabled by setting the environment variable "OIIO_LOAD_DLLS_FROM_PATH" to "0"
-if sys.version_info >= (3, 8) and platform.system() == "Windows" and os.getenv("OIIO_LOAD_DLLS_FROM_PATH", "1") == "1":
-    for path in os.getenv("PATH", "").split(os.pathsep):
-        if os.path.exists(path) and path != ".":
-            os.add_dll_directory(path)
+logger = logging.getLogger(__name__)
```
JeanChristopheMorinPerso, line 11: This seems unused.

pyproject.toml
zachlewis: Good call. 

I'm also removing stuff related to incremental building with cibuildwheel. I've come to realize that `checked_find_package` installs libraries after it builds 'em; but successive OIIO builds that are able to find previously-built libraries in <build-dir>/deps/dist never have the opportunity to be installed (cuz their build recipes are never invoked). This is how and why the custom repair-wheel-cmd worked -- `repairwheel -l $SKBUILD_BUILD_DIR/deps/dist ...` finds and effectively "installs" any shared libraries that `checked_find_package` successfully bundled into the cpython-3.8 wheel, but failed to bundle into the cpython-3.9+ wheels. 

In any case, it's a problem outside the scope of this PR, I think.

src/cmake/compiler.cmake
zachlewis: Well, before, I think CMAKE_INSTALL_RPATH was defaulting to an absolute path; whereas now, it uses loader-path-relative defaults. We could still pass the option on a per-platform basis, but I'm not sure how that would be advantageous...?

src/build-scripts/build_libtiff.bash
```diff
@@ -45,6 +45,7 @@ if [[ -z $DEP_DOWNLOAD_ONLY ]]; then
                -Dtiff-tests=${LIBTIFF_BUILD_TESTS:-OFF} \
                -Dtiff-docs=${LIBTIFF_BUILD_TESTS:-OFF} \
                -Dlibdeflate=ON \
+               -Dwebp=OFF \
```
lgritz, line 48: Is this something we should be able to control? Like

`-Dwebp=${LIBTIFF_WEBP_SUPPORT:-ON}`

but set the environment variable to disable it, only when building the wheels (if that's the only instance it's a problem)?

src/cmake/build_OpenColorIO.cmake
```diff
@@ -45,6 +45,8 @@ build_dependency_with_cmake(OpenColorIO
         -D OCIO_BUILD_TESTS=OFF
         -D OCIO_USE_OIIO_FOR_APPS=OFF
         -D OCIO_INSTALL_DOCS=OFF
+        # Disable versioning in OCIO shared library name
+        -D OCIO_USE_SOVERSION=OFF
```
lgritz, line 49: This feels like another one where I'm not sure what the implications are of changing this. Maybe make it something controllable and only turn it off for the wheels?

src/iconvert/CMakeLists.txt
```diff
@@ -2,4 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
-fancy_add_executable (LINK_LIBRARIES OpenImageIO)
+# Don't build if installing as a Python wheel
+if (SKBUILD)
+    return ()
+endif ()
+
```
lgritz, line 9: I think this idiom is not needed. The `fancy_add_executable` honors a variable called `ENABLE_binaryname` -- if set and is not on, that executable will not build. So instead of this return happening in each one, I think that in one central place where you set up SKBUILD, you can simply 

    set (ENABLE_iconvert OFF)
    set (ENABLE_idiff OFF)
    ... etc

lgritz (changes requested):
I pointed out just a few things that are mostly stylistic: some places where you could have done something simpler, or added a settable control instead of changing a default behavior.

The only major concern I have is that the Windows builds are broken in the CI with this patch.

I'm not really enough of a Python guru or have ever made wheels myself, so don't have much basis to comment on the logic of the wheel building and whether it works.

src/build-scripts/build_libtiff.bash
zachlewis: Oh, weird... to be honest, I don't remember why I added this in the first place; or why I made the adjustment to the bash build-script, and not the cmake install recipe, since the wheels workflows aren't really concerned with the bash build scripts. 
Judging from the surrounding commits, it looks like I was trying to prevent the WebP plugin from linking dynamic x265 libs found at the system level on the macos-13 runner (cuz x265 is GPL-licensed); and I guess I was fearful of libtiff linking dynamic WebP libraries at the system level, which, in turn, link x265? 

I dunno. I'm gonna try removing this line entirely, cuz I believe I'm taking care of avoiding (or at least de-prioritizing) the system libraries in more robust ways in later commits.

If I find that removing the line is causing problems, then yeah, I'll definitely do exactly what you're suggesting, and opt out of providing WebP support to libtiff only where problems arise.

src/cmake/build_OpenColorIO.cmake
zachlewis: OH. Ignore the above.

I just remembered, this option does not control for what I described above; it dictates whether freshly-built dynamic OCIO libs have "namelink copies" installed -- normally, these are just symlinks to the library with comparatively abbreviated filenames (i.e., sans version info); but for the wheels, the symlinks are hardened into actual copies of the library. We want to avoid this, to keep the .whl size as tiny as possible.

Again, since we're now building OCIO statically, this is moot; and, in fact, at the top of the recipe, I have OpenColorIO_BUILD_SHARED_LIBS hardcoded to OFF instead of ON. But for the sake of keeping this as simple as possible, I'll remove this line entirely. 


NB: Currently, _all_ of OIIO's plugin dependencies for which we have cmake recipes are statically built; but in the near future, we'll have to figure out how to disable creation of "namelink copies" for GPL-licensed dependencies of libheif + avif, cuz they most likely do not expose convenient cmake options for handling that,

zachlewis:
I do believe we're in a good place now.

> The only major concern I have is that the Windows builds are broken in the CI with this patch.

I've implemented the adjustments from #4590, which addresses the `CI / Windows-2022 VS2022` failure. 

The `CI / Windows-2019 VS2019` failure isn't induced by the changes here; the failures seem to be occurring on other PRs as well.

JeanChristopheMorinPerso:
I did an analysis of the generated wheels and they look good to me. I looked at:

* Included files in all wheels
* Name of each wheel
* DSOs on Linux (checked the RPATHs and linked libraries).
* I did not check DSOs on macOS and Windows.

I'm making the assumption that they should be fine if the linux ones are fine since tests are run on the produced wheels on all platforms.

Outstanding work @zachlewis!

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,414 @@
...
+  # Source Distribution
+  # ---------------------------------------------------------------------------
+
+  sdist:
+    name: Build SDist
+    runs-on: ubuntu-latest
+    # Don't run on OIIO forks
+    if: |
+      github.event_name != 'schedule' ||
+      github.repository == 'AcademySoftwareFoundation/OpenImageIO'
```
lgritz, line 51: You sure this logic is correct? You definitely only want it to run on the main repo, not people's forks. But doesn't this say that as long as it's not the scheduled event, it will run on forks?

CMakeLists.txt
```diff
@@ -123,6 +123,8 @@ set (PLUGIN_SEARCH_PATH "" CACHE STRING "Default plugin search path")
 file (TO_NATIVE_PATH "${PLUGIN_SEARCH_PATH}" PLUGIN_SEARCH_PATH_NATIVE)
 set (CMAKE_DEBUG_POSTFIX "_d" CACHE STRING "Library naming postfix for Debug builds")
 
+
+
```
lgritz, line 127: Intentional?

lgritz (approved):
A few nits pointed out in comments, but nothing high stakes that couldn't be fixed later. I trust you to decide what to do with the suggestions. Let me know when you've changed anything you are inclined to and are ready for a merge.

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,414 @@
...
+    # Workflow run on pull_request only when related files change.
+    branches-ignore:
+      - dev-*
+      - lg-*
+      - rtd
+      - sonar-1
+    paths:
+      - .github/workflows/wheel.yml
+      - pyproject.toml
+      - src/python/OpenImageIO/*.py
```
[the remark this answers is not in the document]
zachlewis: Good call!

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,414 @@
...
+    tags:
+      - v2.6.*
+      - v3.*
+  pull_request:
+    # Workflow run on pull_request only when related files change.
+    branches-ignore:
+      - dev-*
+      - lg-*
+      - rtd
+      - sonar-1
```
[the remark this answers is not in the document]
zachlewis: I believe this means the branch names in the official repo into which people are requesting the PRs merge. So, yeah, the `lg` branches wouldn't be relevant... 

It sounds like I should remove this section entirely. I'll do that.

CMakeLists.txt
zachlewis: nope!

.github/workflows/wheel.yml
zachlewis: Hmmm.... that's a good point. I think you're right... if anything, we'd want an `&&` operand (which I guess would negate the cron stuff); or we'd want to use only the github.repository bit. 

I based this entire workflow off of OCIO's [wheel_workflow.yml](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/.github/workflows/wheel_workflow.yml), cuz I barely know what I'm doing with github actions -- maybe @remia can elaborate?

remia: I think our goal was to allow the workflow to run on people's forks if trigger by a push / PR but not by the nightly cron job, if that makes sense.

lgritz: I think you definitely want the second clause (repository) to keep it from running on everybody's fork push or whatever. But I don't think you want the first clause at all, because we want it to run for all of scheduled, PRs, and pushes to the official repo, right?

remia: It can be useful to allow the workflow to run on people forks before submitting on the main repo, especially for first time contributors that have to wait for approval every time they push. This is only for building the wheel obviously.

remia: I think @zachlewis set it similar to OCIO where only PRs that are updating Python related code will trigger the wheel build. Combined with the nightly (which run only on the main repo) this has been sufficient for OCIO needs at least.

zachlewis: On it!

zachlewis:
Thanks everyone, I really appreciate the feedback!

@lgritz, unless you (or anyone else) have any remaining questions / comments / doubts / fears / concerns, I think we're good to merge 🤞!

.github/workflows/wheel.yml
```diff
@@ -10,21 +13,17 @@ permissions:
...
-      - dev-*
-      - lg-*
-      - rtd
-      - sonar-1
     paths:
       - .github/workflows/wheel.yml
       - pyproject.toml
       - src/python/OpenImageIO/*.py
+      - src/python/OpenImageIO/*.h
+      - src/python/OpenImageIO/*.cpp
```
lgritz, line 26: Sorry for not spotting this earlier, but I think all of these should be `src/python/*.h`, .cpp, and .py. There is no OpenImageIO subdirectory in that path.

zachlewis: ha -- good point, adjusting now

lgritz (approved):
LGTM!

Thanks for all the hard work on this, it's a very exciting addition to merge!

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,407 @@
...
+      id-token: write
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v3')
+    steps:
+      - uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          pattern: cibw-*
+          path: dist
+          merge-multiple: true
```
lgritz, line 405: I'm sorry, one last question here:
Is this correct to ensure that the upload to pypi is only going to happen on tagged releases from the main repo?

zachlewis: No need to apologize!

You mean, is there anything preventing folks from submitting PRs to their own forks' main branches and pushing a v3 tag locally to kick off an "unsanctioned" upload_pypi step?

Welp... on the pypi.org side, only AcademySoftwareFoundation/OpenImageIO is set up as a "[trusted publisher]( https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/)" for the OpenImageIO project; so PyPi wouldn't even consider attempted uploads from other forks. As I understand it.

zachlewis: That being said, PyPi also strongly advocates for setting up a dedicated Github Actions environment for the publishing workflow, and I don't really know what that means...!

aclark4life: I don't think lines 392 to 405 ensures that, but if you look at the [entire file](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4428/files/b54e873ead5ed2bc29e164c3e28b79c4d1b48416#diff-ac3c74ee584d913de4395ec4eac7256dbd38ced6e1fe789ec9f119529ef7e5e8), LGTM. The uploads actually happen on [upload-artifact](https://github.com/actions/upload-artifact).

lgritz: Yeah, I took for granted that others can't actually publish the wheels on PyPI, but what I meant was, are we sure that (a) user forks won't attempt to do so and look like failed jobs, and (b) we are only pushing to PyPI for release tags, and not for every merged PR? Also, now that I mention that, we probably want to restrict the PyPI publishes to v3.0.* tags, and not accidentally appear to publish new releases if we happen to make a developer preview tag in main (which would be called something like v3.1.2.0-dev).

aclark4life: > That being said, PyPi also strongly advocates for setting up a dedicated Github Actions environment for the publishing workflow, and I don't really know what that means...!

Maybe this, something to do with running jobs from context created in other jobs I think: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#defining-a-workflow-job-environment

aclark4life: > Yeah, I took for granted that others can't actually publish the wheels on PyPI, but what I meant was, are we sure that (a) user forks won't attempt to do so and look like failed jobs, and (b) we are only pushing to PyPI for release tags, and not for every merged PR? Also, now that I mention that, we probably want to restrict the PyPI publishes to v3.0.* tags, and not accidentally appear to publish new releases if we happen to make a developer preview tag in main (which would be called something like v3.1.2.0-dev).

If it's not there already, you may want to add a [step for Test PyPI ](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#separate-workflow-for-publishing-to-testpypi)and merge with *only* that one enabled until you get used to the workflow. Then enable PyPI after sufficient testing.

.github/workflows/wheel.yml
```diff
@@ -0,0 +1,407 @@
...
+          name: cibw-wheels-${{ matrix.python }}
+          path: ./wheelhouse/*.whl
+
+
+  upload_pypi:
+    needs: [sdist, linux, linux-arm, macos, macos-arm, windows]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v3')
```
lgritz, line 397:
```suggestion
    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v3.0.')
```
I think this change will ensure that we only upload to PyPI for tagged releases in the 3.0 branch, which are the supported releases. We can modify this in the future as new branches become supported, but for now, tags that look like v3.1 are just developer releases and so should not go to PyPI. Does that sound right ot everybody?

lgritz (approved):

lgritz:
OK, I'm out of questions or suggested changes. Let's do this!

MrLixm:
Thanks so much to everyone involved, I cannot overstate how game-changer this is for some of us !
It makes OpenImageIO so much more accessible and easier to recommend to less tech-savvy people.

If oiiotool is still shipped with the wheels (since I tested them), I'm pretty sure that using [uv](https://docs.astral.sh/uv/#highlights), its possible to make it one command away from being available to anyone from any context.

src/cmake/compiler.cmake
```diff
@@ -693,7 +693,12 @@ if (CMAKE_SKIP_RPATH)
...
+        else()
+            set(BASEPOINT $ORIGIN)
+        endif()
+        set (CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
```
lgritz, line 701: @zachlewis I'm only realizing now that this change seems quite broken on my Mac. 

oiiotool runs fine, but iv now no longer runs, with an error message of

```
$ iv
dyld[72343]: Library not loaded: @rpath/libOpenImageIO.3.1.0.dylib
  Referenced from: <9D486E88-B93B-3453-ACF8-8424860A81C3> /Users/lg/code/oiio/oiio.lg/dist/bin/iv
  Reason: tried: '/usr/local/lib/libOpenImageIO.3.1.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/lib/libOpenImageIO.3.1.0.dylib' (no such file), '/Users/lg/code/oiio/oiio.lg/dist/bin/libOpenImageIO.3.1.0.dylib' (no such file), '/Users/lg/code/oiio/oiio.lg/dist/bin/lib/libOpenImageIO.3.1.0.dylib' (no such file), '/usr/local/lib/libOpenImageIO.3.1.0.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/local/lib/libOpenImageIO.3.1.0.dylib' (no such file), '/Users/lg/code/oiio/oiio.lg/dist/bin/libOpenImageIO.3.1.0.dylib' (no such file), '/Users/lg/code/oiio/oiio.lg/dist/bin/lib/libOpenImageIO.3.1.0.dylib' (no such file)
```

I believe those funny dist/bin/lib paths (which don't exist) are because of the `${BASEPOINT}/${CMAKE_INSTALL_LIBDIR}` where BASEPOINT is `@loader_path` which in turn is the bin directory. Somehow, the old `${CMAKE_INSTALL_FULL_LIBDIR}` is no longer part of the rpath.

I think our CI doesn't catch the problem because it never launches 'iv'. Can it really be that this is the first I've run iv by hand using a build that incorporates your python wheel changes?

src/cmake/compiler.cmake
lgritz: It wasn't any old absolute path, it was the install destination's lib directory. 

I think this variable can be a list, though. I'm trying now to add the install lib area back in and see if it fixes things for me.

PR #4430, opened 2024-09-17
feat(jpeg): Support encoding/decoding arbitrary metadata as comments

lukasstockner:
This is needed to port Blender's current JPEG IO code to using OIIO, but is also a useful feature to have in general.

For reading, the code tries to parse comments as colon-separated key-value pairs and sets metadata accordingly. For writing, this needs to be explicitly enabled by setting jpeg:com_attributes to 1 in order to avoid accidentally bloating files for existing applications.

## Tests
I've added a small (~10KB) JPEG file containing Blender metadata and a basic test that parses it, checks that the metadata was read correctly, writes it twice (once with and once without `jpeg:com_attributes`), and then checks that those files are also parsed as expected.
In case you're wondering why the info for "no-attribs.jpg" still contains one Blender attribute - that's because the first COM field is still put into `ImageDescription` just like before, so even without `jpeg:com_attributes` it ends up being written to the output file and recognized during parsing.

## Checklist:

  nonconforming.


lgritz:
The Mac failures are unrelated and fixed by a different PR that has already been merged.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -284,10 +284,24 @@ JpgInput::open(const std::string& name, ImageSpec& newspec)
                    && !strcmp((const char*)m->data, "Photoshop 3.0"))
             jpeg_decode_iptc((unsigned char*)m->data);
         else if (m->marker == JPEG_COM) {
+            std::string data((const char*)m->data, m->data_length);
             if (!m_spec.find_attribute("ImageDescription", TypeDesc::STRING))
-                m_spec.attribute("ImageDescription",
-                                 std::string((const char*)m->data,
-                                             m->data_length));
+                m_spec.attribute("ImageDescription", data);
+            // Additional string metadata can be stored in JPEG files as
```
lgritz, line 290: Maybe it's worth explicitly starting by spelling out:

    // The first COM block encountered will be interpreted as the image description.
    // Subsequent COM blocks, if in the form "key:value", ... blah blah

By the way, is this exactly what we want? What if the first COM looks like "key:value", should that always be slotted into ImageDescription?

lukasstockner: I would prefer only setting ImageDescription if the parsing fails, but I figured that's a potentially breaking change so I kept it safe for now.

lgritz: What's the verdict here, @lukasstockner? Do you want to make any more changes, or do you want to keep the logic as-is and we can always revise later if it causes trouble?

lukasstockner: If it's fine with you, I'll go ahead and add some more logic to only set the ImageDescription if a global option is set and/or the matched key fails a heuristic.

lgritz: Sounds good to me.

lgritz:
@lukasstockner Is this ready to merge, or is there more you wanted to do at this stage?

lukasstockner:
> @lukasstockner Is this ready to merge, or is there more you wanted to do at this stage?

I still want to implement the ImageDescription logic discussed above. I didn't find time yet, but hope to do so tomorrow.

lgritz:
@lukasstockner Just a reminder, you were going to make a minor adjustment here, and then I think we are ready to merge. You may also need to rebase on the current main and resolve some very minor conflicts that have occurred since you started this PR.

lgritz:
Ping @lukasstockner

lukasstockner:
So sorry for the delay here, I've finally implemented the improvement we discussed.
There's still some ambiguity (e.g., `Example: A tree` will be parsed as a key-value pair), but now there is a setting to disable that.

lgritz (approved):
LGTM, thanks for getting this one across the finish line.

PR #4435, opened 2024-09-18
admin: change docs and comments references master -> main

lgritz:


jessey-git (approved):
Looks good. I was just about to submit something similar for fetching the correct oiio-images branch but looks like you got that here too.

PR #4440, opened 2024-09-25
Add JXL as a req'd dep.

stolk:
The INSTALL.md documentation lists dependencies that are required, and dependencies that are optional.

JXL is a required dependency that is not in the list.

This MR adds that dependency to the documentation.

This MR also sets an upper bound on libfmt, because fmt version 11 is incompatible with OpenImageIO.

N/A

N/A

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,9 +26,11 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * **OpenColorIO >= 2.2** (tested through 2.3 and main)
  * libjpeg >= 8 (tested through jpeg9e), or **libjpeg-turbo >= 2.1** (tested
    through 3.0)
- * **[fmtlib](https://github.com/fmtlib/fmt) >= 7.0** (tested through 10.2).
+ * **[fmtlib](https://github.com/fmtlib/fmt) >= 7.0 < 11**
    If not found at build time, this will be automatically downloaded unless
    the build sets `-DBUILD_MISSING_FMT=OFF`.
+ * **JXL >= 0.10.1**
+
 
 ### Optional dependencies -- features may be disabled if not found
  * If you are building the `iv` viewer (which will be disabled if any of


lgritz:
I'm pretty sure JXL is an optional dependency: if not found, the rest of the project will build, it just won't have JPEG-XL support. 

This is different from how libtiff is a *required* dependency -- without it, the whole build cannot continue. This is partly by necessity (many parts of the code base depend on things defined by libtiff's headers, even when no actual TIFF files are being read), and partly a policy decision (TIFF files are so common and fundamental, we don't even want an OIIO to be built that is unable to read TIFF files).

INSTALL.md
```diff
@@ -26,9 +26,11 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * **OpenColorIO >= 2.2** (tested through 2.3 and main)
  * libjpeg >= 8 (tested through jpeg9e), or **libjpeg-turbo >= 2.1** (tested
    through 3.0)
- * **[fmtlib](https://github.com/fmtlib/fmt) >= 7.0** (tested through 10.2).
+ * **[fmtlib](https://github.com/fmtlib/fmt) >= 7.0 < 11**
```
lgritz, line 29: You've changed the meaning of this line. "Tested through 10.2" is not just a claim about what versions we think we support, it's making a more exact statement about what the highest version is that is tested by our CI.

INSTALL.md
```diff
@@ -26,9 +26,11 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
...
    If not found at build time, this will be automatically downloaded unless
    the build sets `-DBUILD_MISSING_FMT=OFF`.
+ * **JXL >= 0.10.1**
```
lgritz, line 32: Your instincts are basically right here -- JXL should be listed in the INSTALL.md as a dependency. My only complaint is that by our definition, it's an optional dependency, not a required one.

lgritz (changes requested):

stolk:
> I'm pretty sure JXL is an optional dependency: if not found, the rest of the project will build, it just won't have JPEG-XL support.

I tried again, after removing jxl, and you are right, it is optional. I must have confused it for another cause to stop cmake.

I will close this PR.

INSTALL.md
stolk: Yes, you are right. I found that v11 was incompatible, so tried to make it more restrictive, but by doing so, information is lost.

lgritz: Oh no, I had not realized that version 11 had been released!  It's probably not really incompatible, this rejection is only because we say
```
checked_find_package (fmt REQUIRED
                      VERSION_MIN 7.0
                      VERSION_MAX 10.99
                      BUILD_LOCAL missing
                     )
```

I don't suppose you're interested in trying out doing this fix yourself?

- [ ] Edit the lines above in externalpackages.cmake to raise VERSION_MAX.
- [ ] Edit the CI "linux-latest-releases" test to specify the latest release.
- [ ] Push to your branch, triggering CI to verify that this all works.
- [ ] (Hopefully unnecessary) Fix anything that's broken!
- [ ] Submit PR.

lgritz: Wait, there is something that will need fixing, and it's not at all obvious. I'll take care of this one.

Issue #4436, opened 2024-09-20
[MEMORY] Reduce file descriptors memory footprint in the ImageCache 


bfraboni:
We would like to reduce the memory usage of file descriptors in the `ImageCache` as we observed a large amount of duplicated data. This is mainly due to image descriptors (`ImageSpec` ) being stored for each mip map (`LevelInfo` ) of each sub-image (`SubImageInfo`).

Clearing up some memory should be possible in that area ( `LevelInfo / SubImageInfo` ) but needs a bit a refactor, that I'm happy to start during DevDays with help from @lgritz.

More to come in the next comments about the approach to follow.
```[tasklist]
### Tasks
- [ ] https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4442
```

lgritz:
Yes, I agree that we want to store the ImageSpec per subimage, and the individual MIP levels really only need to know their resolutions and tile sizes, I think those are the only things that meaningfully differ from one MIP level to the next. They don't all need their own ImageSpec's at all.

lgritz:
Additionally, even at the subimage level, we currently store both a "spec" and a "nativespec".

The original concept was that nativespec reflects what was in the file, whereas spec describes what's in memory in the cache. The main way they can differ is (a) in the cache, only a few pixel data types are allowed (uint8, uint16, half, float) and all channels must be the same data type, whereas in the file they can be other types and can differ among the channels; (b) in the cache, the image always appears "tiled", sometimes the real tiling of the file, sometimes looking like one big tile for the whole image, sometimes looking tiled even though it's not ("autotile"). But the real metadata will not change.

(In my defense, when this was designed, we saw a lot less metadata in files, certainly not big things like ICC profiles, thumbnails, or other big things. So having a second ImageSpec didn't seem like a big deal, but now it is.)

OK, so what I'm thinking is that we could store only ONE ImageSpec per subimage, and then only the couple of fields that explain how it differs between the file and the buffer. 

I'm not yet of a strong opinion whether the one spec should be the file, with separate fields for the data format and tile size of the buffer? Or if the spec should be the buffer, with additional fields saying what the data formats and tiling was like in the file?  (Probably the first option, but I'm not 100% sure.)

The question is what to do about the get_imagespec() and imagespec() methods, both of which take a `native` flag to say which it's selecting.

IC::get_imagespec makes a copy, so it could copy the one imagespec and then modify those few extra fields if they're asking for the other kind.

[...]

PR #4442, opened 2024-09-25
api: remove redundant ImageSpec from ImageCache (issue #4436)

bfraboni:
## Description

First draft of frontend changes for  issue #4436.

## Checklist:

- [x] change `imagespec`, `get_imagespec` API calls in ImageCache, ImageCacheImpl, TextureSystem and all calls.
- [x] add `get_cache_dimensions` to access cache internal mip level dimensions
- [x] add `get_cache_dimensions` unit tests once we validate the new version
- [ ] change backend to reduce memory usage (might be a different PR)

### DIFF
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -766,7 +766,7 @@ public:
         return get_imagespec(file, thread_info, spec, subimage);
     }
 
-    /// Return a pointer to an ImageSpec associated with the named image file.
+    /// Return a pointer to an ImageSpec that describes the named image file.
     /// If the file is found and is an image format that can be read,
     /// otherwise return `nullptr`.
     ///
@@ -814,11 +814,13 @@ public:
     }
 
     /// Copy the image dimensions (x, y, z, width, height, depth, full*,
-    /// nchannels, format) and data types associated with the named image
+    /// nchannels, format) and data types that describes the named image
     /// cache file for the specified subimage and miplevel. It does *not*
     /// copy arbitrary named metadata or channel names (thus, for an
     /// `ImageSpec` with lots of metadata, it is much less expensive than
-    /// copying the whole thing with `operator=()`).
+    /// copying the whole thing with `operator=()`). The associated
+    /// metadata and channels names can be retrieved with `imagespec()`
+    /// or `get_imagespec`.
     ///
     /// @param  filename
     ///             The name of the image, as a UTF-8 encoded ustring.
--- a/src/include/OpenImageIO/imagecache.h
+++ b/src/include/OpenImageIO/imagecache.h
@@ -724,8 +724,15 @@ public:
                         int miplevel, ustring dataname, TypeDesc datatype,
                         void* data);
 
-    /// Copy the ImageSpec associated with the named image file.
+    /// Copy the ImageSpec that describes the named image file.
     ///
+    /// Note that the spec returned describes the file as it exists in the
+    /// file, at the base (highest-resolution) MIP level of that subimage.
+    /// Certain aspects of the in-cache representation may differ from the
+    /// file (due to ImageCache implementation strategy or options like
+    /// `"forcefloat"` or `"autotile"`). If you really need to know the
+    /// in-cache data type, tile size, or how the resolution or tiling changes
+    /// on a particular MIP level, you should use `get_cache_dimensions()`.
     /// @param  filename
     ///             The name of the image, as a UTF-8 encoded ustring.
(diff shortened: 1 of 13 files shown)


src/include/OpenImageIO/imagecache.h
```diff
@@ -724,42 +724,29 @@ class OIIO_API ImageCache {
                         int miplevel, ustring dataname, TypeDesc datatype,
                         void* data);
 
-    /// Copy the ImageSpec associated with the named image (the first
-    /// subimage & miplevel by default, or as set by `subimage` and
-    /// `miplevel`).
+    /// Copy the ImageSpec associated with the named image file.
```
lgritz, line 727: I would like to suggest elaborating a bit more here:

```suggestion
    /// Copy the ImageSpec that describes the named image file.
    ///
    /// Note that the spec returned describes the file as it exists in the
    /// file, at the base (highest-resolution) MIP level of that subimage.
    /// Certain aspects of the in-cache representation may differ from the
    /// file (due to ImageCache implementation strategy or options like
    /// `"forcefloat"` or `"autotile"`). If you really need to know the
    /// in-cache data type, tile size, or how the resolution or tiling changes
    /// on a particular MIP level, you should use `get_cache_dimensions()`.
```

lgritz:
Other than the comment I made about elaborating the description a bit, this is all looking great to me.

We talked in a different channel about adding back some very minimal inline methods to keep back compatibility for now, so I know that's still on its way, but if that ends up straightforward, we should be able to merge this without any trouble. Then the API will be nailed down and you can do the further changes to the internals without any additional compatibility breaks during our upcoming beta period.

src/include/OpenImageIO/imagecache.h
bfraboni: Done in [this commit](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4442/commits/f1c97b62efc06a2707810cd0cd2ae13cd7df3153)

bfraboni:
I've added the backwards compatible methods for ImageCache `get_imagespec()` and `imagespec()`, and TextureSystem `get_imagespec()`. I think the CI is be happy now, I just need to write unit tests for the newer `get_cache_dimensions()' method and it'll be good to go.
Thank you @lgritz for the directions !

bfraboni:
I've just added the unit tests for `get_cache_dimensions()`

src/include/OpenImageIO/imagecache.h
```diff
@@ -724,42 +724,51 @@ class OIIO_API ImageCache {
...
-    ///
+    /// Copy the ImageSpec that describes the named image file.
+    ///
+    /// Note that the spec returned describes the file as it exists in the
+    /// file, at the base (highest-resolution) MIP level of that subimage.
+    /// Certain aspects of the in-cache representation may differ from the
+    /// file (due to ImageCache implementation strategy or options like
+    /// `"forcefloat"` or `"autotile"`). If you really need to know the
+    /// in-cache data type, tile size, or how the resolution or tiling changes
+    /// on a particular MIP level, you should use `get_cache_dimensions()`.
```
lgritz, line 735: I'm not sure, but I think maybe this needs a blank line or the `@param` will get folded into that paragraph?

```suggestion
    /// on a particular MIP level, you should use `get_cache_dimensions()`.
    ///
```

src/include/OpenImageIO/imagecache.h
```diff
@@ -768,29 +777,70 @@ class OIIO_API ImageCache {
     /// (even other threads) calls `invalidate()` on the file, or
     /// `invalidate_all()`, or destroys the ImageCache.
     ///
+    /// Note that the spec returned describes the file as it exists in the
+    /// file, at the base (highest-resolution) MIP level of that subimage.
+    /// Certain aspects of the in-cache representation may differ from the
+    /// file (due to ImageCache implementation strategy or options like
+    /// `"forcefloat"` or `"autotile"`). If you really need to know the
+    /// in-cache data type, tile size, or how the resolution or tiling changes
+    /// on a particular MIP level, you should use `get_cache_dimensions()`.
```
lgritz, line 786:
```suggestion
    /// on a particular MIP level, you should use `get_cache_dimensions()`.
    ///
```

lgritz (approved):
LGTM!

I suggested two changes to docs -- just inserting a blank line to be sure it understands the parameter command. I will add those for you by simply accepting the suggestions, then merge.

Thanks, this is great. Looking forward to the eventual change to the internal implementation, but this at least gets the new API locked into place.

Issue #308, opened 2012-04-10
iv - togglable overlay of the display & data windows


lgritz:
Add a toggleable visual overlay to outline the location of the
display window, and/or the data window. (I can suggest color for these
to match other apps).  By default, only show the display window, but
should a dotted-line outside the display window, for the data window,
demonstrating what additional pixels are available.

lgritz:
Coalescing

tkchanat:
PR created here: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4443

PR #4443, opened 2024-09-25
feat: add iv data and display windows overlay

tkchanat:
Adding data and display window overlay in iv.
Dashed line: Display window
Solid line: Data window

Screenshot 2024-09-25 at 15 07 13

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -386,15 +386,10 @@ ImageViewer::createActions()
     //    toggleImageAct->setEnabled(true);
     connect(toggleImageAct, SIGNAL(triggered()), this, SLOT(toggleImage()));
 
-    showDataWindowAct = new QAction(tr("Show Data Window (solid line)"), this);
-    showDataWindowAct->setCheckable(true);
-    connect(showDataWindowAct, SIGNAL(triggered()), this,
-            SLOT(toggleDataWindow()));
-
-    showDisplayWindowAct = new QAction(tr("Show Display Window (dotted line)"), this);
-    showDisplayWindowAct->setCheckable(true);
-    connect(showDisplayWindowAct, SIGNAL(triggered()), this,
-            SLOT(toggleDisplayWindow()));
+    toggleWindowGuidesAct = new QAction(tr("Show display and data window borders"), this);
+    toggleWindowGuidesAct->setCheckable(true);
+    connect(toggleWindowGuidesAct, SIGNAL(triggered()), this,
+            SLOT(toggleWindowGuides()));
 
     slideShowAct = new QAction(tr("Start Slide Show"), this);
     connect(slideShowAct, SIGNAL(triggered()), this, SLOT(slideShow()));
@@ -684,8 +679,7 @@ ImageViewer::createMenus()
     viewMenu->addAction(prevImageAct);
     viewMenu->addAction(nextImageAct);
     viewMenu->addAction(toggleImageAct);
-    viewMenu->addAction(showDataWindowAct);
-    viewMenu->addAction(showDisplayWindowAct);
+    viewMenu->addAction(toggleWindowGuidesAct);
     viewMenu->addSeparator();
     viewMenu->addAction(zoomInAct);
     viewMenu->addAction(zoomOutAct);
@@ -1362,15 +1356,7 @@ ImageViewer::toggleImage()
 
 
 void
-ImageViewer::toggleDataWindow()
-{
-    ((QOpenGLWidget*)(glwin))->update();
-}
-
-
-
-void
-ImageViewer::toggleDisplayWindow()
+ImageViewer::toggleWindowGuides()
 {
     ((QOpenGLWidget*)(glwin))->update();
 }
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -2142,8 +2142,8 @@ ImageViewer::fitWindowToImage(bool zoomok, bool minsize)
     // (or we failed to open it).
     if (!img || !img->image_valid())
         return;
-    // FIXME -- figure out a way to make it exactly right, even for the
-    // main window border, etc.
+        // FIXME -- figure out a way to make it exactly right, even for the
+        // main window border, etc.
 #ifdef __APPLE__
     int extraw = 0;  //12; // width() - minimumWidth();
     int extrah = statusBar()->height()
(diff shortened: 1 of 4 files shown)


lgritz:
Neat!

Do any onlookers have opinions on the dashed vs dotted lines? This way, or reversed? Or is there a different convention that should be used? Does this align with or contradict other viewers that support a similar visuazliation of the windows?

src/iv/ivgl.cpp
```diff
@@ -903,6 +937,44 @@ IvGL::paint_pixelview()
...
+
+    glDisable(GL_TEXTURE_2D);
+    glUseProgram(0);
+
+    glColor4f(0.5f, 0.5f, 0.5f, 1.f);
+    const float xmin = spec.x;
+    const float xmax = spec.x + spec.width;
+    const float ymin = spec.y;
+    const float ymax = spec.y + spec.height;
+    gl_rect_border(xmin, ymin, xmax, ymax);
```
lgritz, line 954: I'm not so sure this is correct. `(spec.x, spec.y)` is the upper-left pixel. But `(spec.x + spec.width, spec.y + spec.height)` is ONE PAST the lower right pixel in each direction.

So first of all, I think we should either draw the rectangle RIGHT ON the boundary pixels, or SURROUNDING the boundary pixels, but definitely not one choice on one side and the other choice on the other side. 

I'm also not sure which is more helpful. Right on actually obscures the top and bottom row and left and right columns. Maybe it should be just beyond the boundary?

Or am I misunderstanding the whole issue?

src/iv/imageviewer.cpp
```diff
@@ -386,6 +386,17 @@ ImageViewer::createActions()
...
+    showDataWindowAct = new QAction(tr("Show Data Window"), this);
+    showDataWindowAct->setCheckable(true);
+    connect(showDataWindowAct, SIGNAL(triggered()), this,
+            SLOT(toggleDataWindow()));
+
+    showDisplayWindowAct = new QAction(tr("Show Display Window"), this);
+    showDisplayWindowAct->setCheckable(true);
+    connect(showDisplayWindowAct, SIGNAL(triggered()), this,
+            SLOT(toggleDisplayWindow()));
+    showDisplayWindowAct->setChecked(true);  // Show display window by default
```
lgritz, line 398: I think maybe I prefer that the default be to not draw either one, and you can toggle it on if you want to see where they are? But I'm willing to be talked out of that if you had a reason to show it by default.

src/iv/ivgl.cpp
```diff
@@ -903,6 +937,44 @@ IvGL::paint_pixelview()
...
+void
+IvGL::paint_datawindow()
+{
+    IvImage* img = m_current_image;
+    const ImageSpec& spec(img->spec());
+
+    glDisable(GL_TEXTURE_2D);
+    glUseProgram(0);
+
+    glColor4f(0.5f, 0.5f, 0.5f, 1.f);
```
lgritz, line 949: What do other viewer apps do here? Is 0.5 grey the right choice? Should it be a color that is unusual in natural imagery, like bright magenta, so there's almost no conceivable image in which it wouldn't be clearly visible?

src/iv/ivgl.cpp
tkchanat: If lines were rasterized at pixel center, what you said would be true (with last row and column being covered). But in this case, the bound `(spec.x, spec.y)`, `(spec.x + spec.width, spec.y + spec.height)` lie on the pixel edges. It's uncertain which side the fragments are going to be rasterized. And the actual line segment rasterization algorithm isn't strictly enforced by the OpenGL specification, so it can differ from device to device. This results in the following cases, with the same zoom level but panning the camera slightly to the right yields different rasterized lines.
|Border shown on top and left|Border shown on top and right|
|:--:|:--:|
|image|image|

I don't see this as a major concern, as modern displays have high dpi standards. It should be quite trivial for users to notice.

src/iv/imageviewer.cpp
tkchanat: I don't have a strong preference for it, it was mostly to follow the description in https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/308. Happy to disable them by default.

src/iv/ivgl.cpp
tkchanat: Good question, I am not sure either. But I did try to use the `GL_XOR` logical operation while drawing the borders, which gives great contrast in non gray-tone images. The GrayRemapHorizontal example shows where it fails around gray pixels. 
|MtTamWest|GrayRemapHorizontal|
|:--:|:--:|
|image|image|

Nonetheless, the border is still very visible and it adapts to the viewing image without having to set a constant color.

src/iv/imageviewer.cpp
lgritz: Oh boy, after 12 years, I didn't realize it was me who made the request! (Though I was probably transcribing what somebody told me.)

I'm fine with the dash/solid choice, we can easily change that later if people object after trying it.

I think if we guaranteed that the lines were *just outside* the window in question, and thus the default view of showing exactly the display window would not show it, I would be ok with defaulting display window guide to be on. But since we can't control its exact placement, I believe I prefer it to be off so that it doesn't obscure any pixels or interfere with the view of the image, by default. So let's default to off and people can turn on when they want to check where the windows are.

src/iv/ivgl.cpp
lgritz: My bad, I didn't notice the XOR. I think that's fine. We can always revise if people complain that there are significant cases where it's unclear.

src/iv/ivgl.cpp
lgritz: OK, let's roll with this oddity for now. We can always revisit later if it's causing actual trouble for people.

src/iv/imageviewer.cpp
```diff
@@ -386,6 +386,17 @@ ImageViewer::createActions()
     //    toggleImageAct->setEnabled(true);
     connect(toggleImageAct, SIGNAL(triggered()), this, SLOT(toggleImage()));
 
+    showDataWindowAct = new QAction(tr("Show Data Window"), this);
```
lgritz, line 389: Can I suggest that the menu text itself say "Show Display Window (dotted line)" and "Show Data Window (solid line)" to make it self-documenting.

src/iv/imageviewer.cpp
tkchanat: No problem, that sounds fair to me. I can disable them both by default.

src/iv/ivgl.cpp
tkchanat: No, not at all. I was experimenting the XOR option yesterday, so my changes weren't pushed yet. I agree we can revise this later if needed.

src/iv/imageviewer.cpp
tkchanat: Surely do!

lgritz:
I think we're very close on this one. I believe outstanding issues are:
* At least one of your commits needs a DCO sign-off.
* Some clang-format cleanup needed.
* On chat, we talked about turning the two controls into just one, since people probably want to view both window guides, or neither.
* Not a problem -- the "bleeding edge" failure, that's fixed in main (you could rebase if you want, or just ignore it).

lgritz:
I like the new combined behavior. Simple and easy to use!

lgritz (approved):
LGTM -- this is a nice feature

Issue #4361, opened 2024-07-29
[BUG] Sony ARW medium and small compression 15 bit per sample files are unsupported


lingyukongt:
**Describe the bug**

Importing a Sony ARW Medium or Small compression file with 15 bits per sample is unsupported in OIIO.

Other Sony ARW Medium or Small compression files which are not 15 bits per sample open correctly and are supported.

**OpenImageIO version and dependencies**

```
oiiotool --buildinfo
OIIO 2.5.12.0 | MacOS/ARM
    Build compiler: Apple clang 15.0 | C++14/201402
    HW features enabled at build: neon
Dependencies: Boost 1.85.0, BZip2 1.0.8, FFmpeg 6.0, fmt 10.2.1, Freetype
    2.13.2, GIF 5.2.2, Libheif 1.17.6, libjpeg-turbo 3.0.3, LibRaw 0.21.2,
    OpenColorIO 2.3.2, OpenEXR 3.2.4, OpenVDB NONE, PNG 1.6.43, pugixml 1.14,
    pybind11 2.12.0, Python 3.12.3, Robinmap, TBB 2021.12.0, TIFF 4.6.0, WebP
    1.4.0, ZLIB 1.2.12
```

**To Reproduce**

1. Download the ARW file
2. Try to convert to a different format with OIIO
3. This error appears.
4. I expected it to output.

```
oiiotool /Users/topazlabs/Downloads/A1A04973.ARW -o ~/Downloads/arw.png
oiiotool ERROR: read : Could not open file "/Users/topazlabs/Downloads/A1A04973.ARW", Unsupported file format or not RAW file
Full command line was:
> oiiotool /Users/topazlabs/Downloads/A1A04973.ARW -o /Users/topazlabs/Downloads/arw.png
```

**Evidence**

https://www.dropbox.com/scl/fi/y19jqq4q3rh1wv7810048/A1A04973.ARW?rlkey=pnrdm0oo0hlpbkep1goczk1da&st=ij4br4bn&dl=0

lgritz:
We rely on libraw to decode that file type. I think that maybe this means that libraw itself doesn't handle that bit depth for that file? 

I don't think there's a lot we can directly do until libraw supports it. Maybe you should file an issue against that project?

LibRaw:
What version of LibRaw do you use?
Sony Small/Medium Pseudo-RAW format is supported since 202403 snapshot published in our github repo on Mar 29, 2024.
[...]

PR #4448, opened 2024-09-26
fix(raw) check for nullptr in raw input plugin

### LINKED ISSUE #3115: [BUG] Invalid write crash when opening certain RAW images

antond-weta:
## Description

This adds a null pointer check in the raw input plugin when the user requested raw bayer data for a file format which doesn't have it. With this fix OIIO shows an error instead of crashing in such case.

This fixes the issue mentioned in the comments of #4361. The title issue of #4361 has been resolved in LibRaw 0.22.0, so we can close that issue, I guess.

This also fixes the crash on 5 out of 6 examples in #3115. The last one is unrelated, and I don't have a good fix for that at this time (crash due to the raw buffer having half resolution, with the pixel aspect ratio of 0.5). Perhaps we could create a separate issue and close #3115. 

## Tests

I have not added any tests for this change specifically, as it is fairly trivial. 
I have started working on a more comprehensive test suite for the raw plugin / rawtoaces, but it's not ready yet, and I'm not sure where it should live at this stage.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -1516,6 +1516,13 @@ RawInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
     if (!m_process) {
         // The user has selected not to apply any debayering.
 
+        if (m_processor->imgdata.rawdata.raw_image == nullptr) {
+            errorfmt(
+                "Raw undebayered data is not available for this file \"{}\"",
+                m_filename);
+            return false;
+        }
+
         // The raw_image buffer might contain junk pixels that are usually trimmed off
         // we must index into the raw buffer, taking these into account
         auto& sizes        = m_processor->imgdata.sizes;


lgritz (approved):
LGTM!

Don't worry about the failed "bleeding edge" test -- that's unrelated and is because OCIO just checked some changes into their main that require a test output reference update on our end.

PR #4451, opened 2024-09-27
fix (iv): iv shows constant brown and GL error messages on start-up.

dadler:
This fixes a problem seen on RHEL 9.2 Linux when launching an image in iv, eg:
  iv testsuite/texture-overscan/ref/out-exact.exr

The image displayed as a constant brown and the terminal showed:
  GL error create program 1281 - Invalid value
  GL error After attach vertex shader. 1281 - Invalid value

The fix is to initialize m_shader_program and m_vertex_shader in the class constructor.

ERROR ANALYSIS:

The bug was introduced in 5fe44415 "feat(iv): OCIO color managed display (#4031)".

Before that commit, create_shaders() initialized those members to 0 before they were used.

After that commit:

* uninitialized m_shader_program is passed to glDeleteProgram() (near the top of create_shaders) ), causing the
"create program" warning (there isn't an print_error call after glDeleteProgram so the error is reported as "create program").

* ununitialized m_vertex_shader is non-0 when checked by if (!m_vertex_shader), causing the vertex shader not to be created and causing the later call to glAttachShader() with uninitialized m_vertex_shader to fail.  That causes the second error message, and lack of a vertex shader apparently causes the image to display as constant brown.


lgritz (approved):
LGTM!

Issue #3974, opened 2023-09-04
[FEATURE REQUEST] Add check_open validity plausibiity checks to format readers


lgritz:
There is a new helper function, ImageInput::check_open(), that can be used by ImageInput::open() implementations to do some sanity checking to try to discover which files have certain header values so preposterous that they seem likelier to be corrupted or even malicious than they are to be legit.

You can see how this is used in the TIFF reader (search tiffinput.cpp for "check_open") or the Targa reader (targainput.cpp). Check also in imageio.h for the check_open helper method of ImageInput. (N.B. Not to be confused for the ImageOutput::check_open helper, of a similar theme, but for output.)

Task: Add this to the other file format's ImageInput implementations.

A "good first issue" is adding the check to any **ONE** format reader. You don't need to do them all.

It does need some format-specific care to be sure that the resolution limits you pass to check_open match what seems likely or possible for the given format. So don't just blindly cut and paste the call from one format reader to another without understanding it and whether it reflects the right choices for that format.

DharshanV:
Hellooo! I was thinking of picking this one up for dev days :)

lgritz:
Adding a comment to ensure this gets picked up by Clotributor as recently active. This would be a great Dev Days item.

hkgulka:
Hi @lgritz, it seems like there may still be some format readers without this check. If this is the case, I would love to help out with this for Dev Days!

lgritz:
Absolutely!

PR #4452, opened 2024-09-27
Added check_open for PNG, JPEG, and EXR readers

DharshanV:
## Description
This PR is ment to address this feature request: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3974

I added `check_open` for these file format's ImageInput while following each file specfication:
- PNG: https://www.w3.org/TR/png/#11IHDR
- JPEG: https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format#JFIF_APP0_marker_segment
- EXR: https://openexr.com/en/latest/OpenEXRFileLayout.html#header-attributes-all-files

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -1024,7 +1024,7 @@ OpenEXRInput::seek_subimage(int subimage, int miplevel)
     m_miplevel = miplevel;
     m_spec     = part.spec;
 
-    if (!check_open(m_spec, { 0, 1 << 30, 0, 1 << 30, 0, 1 << 16, 0, 1 << 16 }))
+    if (!check_open(m_spec, { 0, 1 << 20, 0, 1 << 20, 0, 1 << 16, 0, 1 << 12 }))
         return false;
 
     if (miplevel == 0 && part.levelmode == Imf::ONE_LEVEL) {
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -1024,6 +1024,9 @@ OpenEXRInput::seek_subimage(int subimage, int miplevel)
     m_miplevel = miplevel;
     m_spec     = part.spec;
 
+    if (!check_open(m_spec, { 0, 1 << 30, 0, 1 << 30, 0, 1 << 16, 0, 1 << 16 }))
+        return false;
+
     if (miplevel == 0 && part.levelmode == Imf::ONE_LEVEL) {
         return true;
     }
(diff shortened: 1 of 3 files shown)


src/png.imageio/pnginput.cpp
```diff
@@ -162,7 +162,8 @@ PNGInput::open(const std::string& name, ImageSpec& newspec)
     bool ok = PNG_pvt::read_info(m_png, m_info, m_bit_depth, m_color_type,
                                  m_interlace_type, m_bg, m_spec,
                                  m_keep_unassociated_alpha);
-    if (!ok || m_err) {
+    if (!ok || m_err
+        || check_open(m_spec, { 0, 1 << 30, 0, 1 << 30, 0, 1, 0, 4 })) {
```
DharshanV, line 166: I am not sure what is a reasonable max size here is. According to the doc, the width and height are four-byte unsigned integer. So the range is range 0 to 2^31 - 1.

lgritz: For OpenEXR output, here is what we did, maybe that's the right approach here too?

        // Arbitrarily limit res to 1M x 1M and 4k channels, assuming anything
        // beyond that is more likely to be a mistake than a legit request. We
        // may have to come back to this if these assumptions are wrong.
        if (!check_open(Create, srcspec,
                        { 0, 1 << 20, 0, 1 << 20, 0, 1, 0, 1 << 12 }))
            return false;

So basically, we just decided that although it *could* be a full uint32 size, we're going to assume nobody needs ore than 1M pixels in each direction.

If that makes us nervous, then as a next step, perhaps instead of hard-coding it, we could make a global OIIO attribute that provides a reasonable resolution limit, defaulting to 1M say, and then if somebody really has a legit need for larger dimensions, there is a way to override it without changing the source.

lgritz, line 166: I believe there is a bug here: you need `!check_open()`, right? That's probably why CI is failing.

src/openexr.imageio/exrinput.cpp
```diff
@@ -1024,6 +1024,9 @@ OpenEXRInput::seek_subimage(int subimage, int miplevel)
     m_miplevel = miplevel;
     m_spec     = part.spec;
 
+    if (!check_open(m_spec, { 0, 1 << 30, 0, 1 << 30, 0, 1 << 16, 0, 1 << 16 }))
```
lgritz, line 1027: `1 << 30` seems high to me. Maybe `1 << 20`? I think we should run with the assumption that we're not going to encounter an OpenEXR file that is more than 1 million pixels wide or high, and that if a file says it is, it's most likely corrupted somehow. Similarly `1 << 16` is an awful lot of channels. For OpenEXR output, I used `1 << 12` which is still pretty generous.

src/png.imageio/pnginput.cpp
DharshanV: Oh yes! First time doing this, forgot there is `make test`.

DharshanV: Okay than I think I will try to match the outputs limits (if they exist). And yea, 1 << 30 might be bit too big for general use cases.

lgritz (approved):
LGTM! Thanks for the fix-ups.

Issue #3881, opened 2023-06-13
feat(raw): Expose max_raw_memory_mb setting


JenusL:
**Is your feature request related to a problem? Please describe.**
We have some huge DNG-files that we're unable to develop with the default memory allocation in libraw.

**Describe the solution you'd like**
I would like the `imgdata.rawparams.max_raw_memory_mb` in libraw to be exposed as a `raw:max_raw_memory_mb` setting.

**Describe alternatives you've considered**
We have recompiled libraw with a higher default memory allocation to get around the problem but that isn't optimal.

**Additional context**
We have tested it to make sure that this setting will actually solve our problem.

lgritz:
🎣 Mackerel!

ankmachine:
going to take this one up for "dev days"

ankmachine:
I have created PR to address this feature https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4454 , but not able to add any reviewer for some reason.

lgritz:
> I have created PR to address this feature #4454 , but not able to add any reviewer.

Sigh. It's been a very busy day for the one volunteer who has been doing all the reviews.

lgritz:
This was closed by #4454

PR #4454, opened 2024-09-27
feat(raw): raw reader - exposing max_raw_memory_mb

ankmachine:
## Description

PR is to address https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3881 feature request. The reported wanted to expose `imgdata.rawparams.max_raw_memory_mb` as `raw:max_raw_memory_mb`

## Tests

I was able to compile my code successfully, no new test were added for this.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -453,12 +453,9 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // Output 16 bit images
     m_processor->imgdata.params.output_bps = 16;
 
-    // exposing max_raw_memory_mb setting
-    static_assert(sizeof(m_processor->imgdata.rawparams.max_raw_memory_mb)
-                      == sizeof(int),
-                  "Sizes must match");
-    config.getattribute("raw:max_raw_memory_mb", TypeDesc::INT,
-                        &m_processor->imgdata.rawparams.max_raw_memory_mb);
+    // Exposing max_raw_memory_mb setting. Default max is 2048.
+    m_processor->imgdata.rawparams.max_raw_memory_mb
+        = config.get_int_attribute("raw:max_raw_memory_mb", 2048);
 
 
     // Disable exposure correction (unless config "raw:auto_bright" == 1)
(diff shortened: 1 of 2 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -453,6 +453,14 @@ RawInput::open_raw(bool unpack, const std::string& name,
...
     m_processor->imgdata.params.output_bps = 16;
 
+    // exposing max_raw_memory_mb setting
+    static_assert(sizeof(m_processor->imgdata.rawparams.max_raw_memory_mb)
+                      == sizeof(int),
+                  "Sizes must match");
+    config.getattribute("raw:max_raw_memory_mb", TypeDesc::INT,
+                        &m_processor->imgdata.rawparams.max_raw_memory_mb);
+
+
```
lgritz, line 463: Might it be simpler to express this more like

    m_processor->imgdata.rawparams.max_raw_memory_mb = 
        config.get_int_attribute("raw:max_raw_memory_mb", <default value>);

?? Then you wouldn't need the static assert, and various type conversions would happen automatically if the types ever change.

Also, would you mind please documenting the existence and meaning of this attribute in the raw section of builtinplugins.rst ? Thanks.

lgritz (changes requested):
This is great, I just made a couple suggestions.

src/raw.imageio/rawinput.cpp
lgritz: What's a good default for this? The other reason I like my suggestion above is that it's probably a good idea to by default impose some reasonable limit, even if this parameter is not passed from the user. We shouldn't allow big files to consume unlimited memory in libraw unless specifically asked to do so.

JenusL: Default in libraw is 2048 (2GB)
https://github.com/LibRaw/LibRaw/blob/9bcb8a1d9593ba67e4eb67fed716efc5e1353d5c/libraw/libraw_const.h#L24
and
https://github.com/LibRaw/LibRaw/blob/9bcb8a1d9593ba67e4eb67fed716efc5e1353d5c/doc/API-datastruct.html#L556-L558
Libraw itself doesn't have a upper limit.

lgritz: Great, let's just express that default as I suggested above, then you don't need the static_assert or the needing pointers.

ankmachine:
> This is great, I just made a couple suggestions.

@lgritz I have addressed the feedback and added the documentation. Let me know if its good :) .

lgritz (approved):
LGTM, thanks!

PR #4460, opened 2024-09-27
docs: convert to test for ImageBufAlgo Partial Image Arithmetic

lydia-zheng:
## Description

This a documentation to testsuite conversion for cpp and python for the ImageBufAlgo Chapter, Section: Image Arithmetic up to the div function.

The PR also fixes a python testing image file reference in ..docs-examples-python/ref/out.txt, where reference images "simple.tif" and "scanlines.tif" should have the docs-examples-cpp prefix (see ..docs-examples-python/run.py line 11 for details).

## Tests
Verified in SPI build environment that existing tests are working.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/imagebufalgo.rst
+++ b/src/doc/imagebufalgo.rst
@@ -56,7 +56,7 @@ Most ImageBufAlgo functions that produce image data come in two forms:
 
 
 2. Pass a destination ImageBuf reference as the first parameter.
-   
+
    The function is passed a *destination* ImageBuf where the results will
    be stored, and the return value is a `bool` that is `true` if the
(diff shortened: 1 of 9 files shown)


grdanny (approved):
we built the docs and looked at the results.
it all looks good to me.
Larry is looking into that failed CI action for us, but if it's all good then this look good to me.

lgritz (approved):
Looks good to me!

Great work, @lydia-zheng

lydia-zheng:
Thank you for checking and merging! Was a great learning experience :)

Issue #4461, opened 2024-09-29
[BUG] Floating  point operations performed on unitialized values.


stolk:
**Describe the bug**

When image output is writing tiles, then the tiles at the edges can be partially filled.

The unused pixels of the partial tile will never have their values set.

But they still undergo float conversion, causing floating point ops being performed on uninitialized values.

This has all sorts of bad effects, being UB, NaN generation, and possibly floating point exceptions if the application has those enabled. (Side note: we should run all our tests with FPE enabled!)

**OpenImageIO version and dependencies**

```
$ ./oiiotool --buildinfo
OIIO 2.6.7.0dev | Linux/x86_64
    Build compiler: clang 18.1 | C++17/201703
    HW features enabled at build: sse2
    No CUDA support (disabled / unavailable at build time)
Dependencies: BZip2 1.0.8, DCMTK NONE, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF 5.2.2, Imath 3.1.9, JPEG 80, JXL
    0.12.0, Libheif NONE, libjpeg-turbo NONE, LibRaw NONE, OpenColorIO 2.4.0, OpenCV NONE, OpenEXR 3.1.5, OpenGL, OpenJPEG
    NONE, OpenVDB NONE, PNG 1.6.43, Ptex NONE, Ptex NONE, pybind11 2.11.1, Python3 3.12.3, Qt5 NONE, Qt6 NONE, Robinmap
    1.3.0, TBB 2021.11.0, TIFF 4.5.1, WebP 1.3.2, ZLIB 1.3

```

**To Reproduce**

Run valgrind on the maketx binary with an image that has a width that is not a multiple of tile size (64.)

**Evidence**

```
==150208== Command: ./maketx --threads 1 /home/stolk/Desktop/test.jpeg
==150208== 
==150208== Conditional jump or move depends on uninitialised value(s)
==150208==    at 0x5A9C8A6: roundf (s_roundf.c:38)
==150208==    by 0x4FE3A97: round (src/include/OpenImageIO/simd.h:7614)
==150208==    by 0x4FE3A97: void OpenImageIO_v2_6_7::convert_type<float, unsigned char>(float const*, unsigned char*, unsigned long, unsigned char, unsigned char) (src/include/OpenImageIO/fmath.h:915)
[...]

PR #4462, opened 2024-09-29
[image output] Initialize pixels of partial tile conversion buffer.

stolk:
When writing a partial tile, the unused pixels still go through float conversion.

This means, that floating point operations are done on uninitialized data.

This can easily lead to NaN and to floating point exceptions, if those were to be enabled.

This change will set the shared buffer used for all partial tiles to all zero pixels.

Tested by running valgrind before and after the change

FIXES: #4461

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imageoutput.cpp
+++ b/src/libOpenImageIO/imageoutput.cpp
@@ -177,8 +177,13 @@ ImageOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
                     ok &= write_tile(x, y, z, format, tilestart, xstride,
                                      ystride, zstride);
                 } else {
-                    if (!buf.get())
-                        buf.reset(new char[pixelsize * m_spec.tile_pixels()]);
+                    if (!buf.get()) {
+                        const size_t sz = pixelsize * m_spec.tile_pixels();
+                        buf.reset(new char[sz]);
+                        // Not all pixels will be initialized, so we set them to zero here.
+                        // This will avoid generation of NaN, FPEs and valgrind errors.
+                        memset(buf.get(), 0, sz);
+                    }
                     OIIO::copy_image(m_spec.nchannels, xw, yh, zd, tilestart,
                                      pixelsize, xstride, ystride, zstride,
                                      &buf[0], pixelsize,


src/libOpenImageIO/imageoutput.cpp
```diff
@@ -177,8 +177,14 @@ ImageOutput::write_tiles(int xbegin, int xend, int ybegin, int yend, int zbegin,
...
-                    if (!buf.get())
-                        buf.reset(new char[pixelsize * m_spec.tile_pixels()]);
+                    if (!buf.get()) {
+                        const size_t sz = pixelsize * m_spec.tile_pixels();
+                        char* mem       = new char[sz];
+                        // Not all pixels will be initialized, so we set them to zero here.
+                        // This will avoid generation of NaN, FPEs and valgrind errors.
+                        memset(mem, 0, sz);
+                        buf.reset(mem);
+                    }
```
lgritz, line 187: I'm confused about where the access of the uninitialized memory is. In copy_image? In write_tile?

Also, I wonder what you think about the following slightly alternate idiom:

                    if (!buf.get()) {
                        const size_t sz = pixelsize * m_spec.tile_pixels();
                        buf.reset(new char[sz]);
                        // Not all pixels will be initialized, so we set them to zero here.
                        // This will avoid generation of NaN, FPEs and valgrind errors.
                        memset(buf.get(), 0, sz);
                    }

I think this is a smaller change from the previous code's approach and doesn't need to introduce an additional raw pointer variable?

stolk: The callstack of the use of uninitialized data is in the bug report of issue #4462

Note that valgrind gives you two callstacks if you issue the `--track-origins=yes` option:

First it shows you where the access of garbage data happened.

Then it shows you where the data in question was allocated.

The invalid data is used in `parallel_convert_from_float()` when it converts **all** 64x64 FP pixels, including the ones that were never assigned a value. The actual leaf-func that triggers a branch on invalid data (and thus is UB) is the rounding function `roundf()`.

I will replace the explicit allocation variable as you suggested, but I find them equally valid.

stolk: Oh, and I see what you mean about the confusion... so the UB was triggered not in the main thread, but in a thread from the thread pool. So it is not traced all the way back to `main()`

(That's why I was asking about a mode that uses no threading at all, and does all the work on the main thread. Sadly, there is no such mode: Even with `--threads 1` the work will be performed on a separate thread. I think it is better form to have a switch to disable all threading, even if it is only for easier debugger sessions. But fixing that is probably non trivial and outside the scope of this issue.)

lgritz: It surprises me that --threads 1 doesn't run all in the main thread. I wonder if there is someplace that is failing to honor that number and using the full thread ppool.

lgritz: What about if in addition to --threads 1, you set env `OPENIMAGEIO_THREADS=1` ?

lgritz: Yeah, I consider this a bug.

lgritz (approved):
LGTM

PR #4464, opened 2024-09-30
ci: Remove old MacOS-12 CI test

lgritz:
It's an old flavor not well supported by homebrew any longer, so what's happening is that some of the dependency packages we ask to install have to build from source, which means it takes a long long time. Just get rid of it, we have two other Mac test cases -- a newer MacOS 13 Intel one, and a MacOS 14 ARM one. That's enough coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -506,14 +506,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - desc: MacOS-12
-            runner: macos-12
-            nametag: macos12-py310
-            cc_compiler: clang
-            cxx_compiler: clang++
-            cxx_std: 17
-            python_ver: "3.10"
-            aclang: 13
           - desc: MacOS-13
             runner: macos-13
             nametag: macos13-py311
@@ -525,11 +517,11 @@ jobs:
             simd: sse4.2,avx2
           - desc: MacOS-14-ARM
             runner: macos-14
-            nametag: macos14-arm-py311
+            nametag: macos14-arm-py312
             cc_compiler: clang
             cxx_compiler: clang++
             cxx_std: 20
-            python_ver: "3.11"
+            python_ver: "3.12"
             aclang: 15
     runs-on: ${{ matrix.runner }}
     env:


ThiagoIze:
The https://vfxplatform.com/ says we should target macOS 12.0 SDK. We can do that on newer macOS by I think setting in cmake [CMAKE_OSX_DEPLOYMENT_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html). Do we have that? If not, should we add it to say the MacOS-13 runner as part of this PR?

lgritz:
> The https://vfxplatform.com/ says we should target macOS 12.0 SDK. We can do that on newer macOS by I think setting in cmake [CMAKE_OSX_DEPLOYMENT_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html). Do we have that? If not, should we add it to say the MacOS-13 runner as part of this PR?

I'm not sure I understand what that does, in light of the fact that we don't rely on anything other than the standard C++ library (nothing newer than C++17) and the other open source dependencies we install ourselves. We're not using any native MacOS APIs, so I'm not sure what there is to nail down specifically to OSX 12.x.

I also have an alternate design that works fine, which is to KEEP this MacOS-12 test, but instruct it not to build OpenCV or Qt. So that means that only for MacOS-12, we wouldn't be building in support for the OpenCV interop or osltoy -- neither of which is actually tested by the automated testsuite anyway. If you think that's a better approach, I can replace the present PR by that version.

lgritz:
Alternate approach: #4665

ThiagoIze:
It's for more than just macos API. Take this random link I found https://stackoverflow.com/a/19819591 which says it's used for selecting the C++ STL implementation. 

Unlike linux where we need to build on older distros so we build against an old enough glibc so the binary works for all users, on macs you can build on the latest macos and instead use this flag to tell the compiler/linker what to target.

lgritz:
OK, I'll close this PR and keep building specifically on MacOS-12.

PR #4468, opened 2024-10-01
docs: Convert ImageInput-unassociatedalpha examples from docs to tests

pfranz:
## Description

Convert error checking C++ and Python examples from the "imageinput" chapter into tests within the "docs-examples" testsuites (https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/3992).

I grabbed unpremut.tif from the output of another test.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
I think this looks good, but it looks like it has merge conflicts because other people have had PRs merged recently in the same areas of those files. Can you please rebase on the current main and resolve any conflicts?

lgritz:
All looked good except needed an update of the MacOS ARM outputs of two tests. (You updated the reference output used for all the other platforms.) I took the liberty of pushing the fix to your branch. If this passes CI, I will merge.

lgritz (approved):
LGTM, merging.

PR #4479, opened 2024-10-06
Color management nomenclature improvements

lgritz:
A round of color management improvements (I hope).

We were pretty squirrelly about what "linear" meant. Get more exact. Now we will treat "linear" as a legacy synonym for "lin_rec709" (sRGB primaries, but linear response). Change file format readers to say "lin_rec709" when they mean it, and not use the more generic alias "linear".

"scene_linear", on the other hand, is the internal working color space for math. Which concrete color space it is an alias for is determined by the OCIO config in use.

Use "g18_rec709" and "g22_rec709" for rec709 + gamma 1.8 or 2.2, respectvely.

A few new API calls help out: equivalent_colorspace(), set_colorspace(spec,name), set_colorspace_rec709_gamma().

OpenEXR always tagged files as linear by default. It was too aggressive about this, so with this PR, now it defaults to lin_rec709 only for files that appear to be RGB and not have any other clues about the color space.

Similarly, the raw reader was too eager to set the Exif:ColorSpace to look like sRGB even when it wasn't.

Lots of reference outputs needed to be updated.

This is a potentially big behavior change. But I think it's for the better.

### DIFF
--- a/src/cineon.imageio/cineoninput.cpp
+++ b/src/cineon.imageio/cineoninput.cpp
@@ -169,7 +169,7 @@ CineonInput::open(const std::string& name, ImageSpec& newspec)
     // This is not very smart, but it seems that as a practical matter,
     // all Cineon files are log. So ignore the gamma field and just set
     // the color space to KodakLog.
-    m_spec.attribute("oiio:ColorSpace", "KodakLog");
+    m_spec.set_colorspace("KodakLog");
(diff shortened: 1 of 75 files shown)


lgritz:
No comments here, but on chat and elsewhere, I've gotten positive feedback on this, so let's do it.

PR #4482, opened 2024-10-07
api: first draft for removing nativespec from ImageBuf

bfraboni:
Second half of API changes to remove `nativespec` from `ImageCache` and `ImageBuf` for issue: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4436

- it currently does not compile, I'm just opening this on to discuss with @lgritz about the changes required.
- I have removed `nativespec()` from the ImageBuf API but this creates a bunch of problems I have not answers for in `ImageBuf::read`

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -1064,11 +1064,10 @@ public:
     /// knowledge of its pixel memory layout.  USE WITH EXTREME CAUTION.
     ImageSpec& specmod();
 
-    /// Return a read-only (const) reference to the "native" image spec
-    /// (that describes the file, which may be slightly different than
-    /// the spec of the ImageBuf, particularly if the IB is backed by an
-    /// ImageCache that is imposing some particular data format or tile
-    /// size).
+    /// Return the "native" channel format that describes the file,
+    /// which may be slightly different than the spec of the ImageBuf,
+    /// particularly if the IB is backed by an ImageCache that is imposing
+    /// some particular data format or tile size.
     ///
     /// This may differ from `spec()` --- for example, if a data format
     /// conversion was requested, if the buffer is backed by an ImageCache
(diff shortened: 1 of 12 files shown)


src/include/OpenImageIO/imagebuf.h
```diff
@@ -1064,19 +1064,25 @@ class OIIO_API ImageBuf {
...
     /// This may differ from `spec()` --- for example, if a data format
     /// conversion was requested, if the buffer is backed by an ImageCache
     /// which stores the pixels internally in a different data format than
     /// that of the file, or if the file had differing per-channel data
     /// formats (ImageBuf must contain a single data format for all
     /// channels).
-    const ImageSpec& nativespec() const;
+    TypeDesc file_format() const;
+    /// Same as the above for channel specific formats.
+    std::vector<TypeDesc> file_channelformats() const;
```
lgritz, line 1080: Not sure how much it matters, but do you think returning a vector is good here? As opposed to returning a `const vector&` or a `cspan<TypeDesc>`? (But those are only options it's stored somewhere in the IB and we can make a stable reference to it.

src/python/py_imagebuf.cpp
```diff
@@ -361,7 +361,9 @@ declare_imagebuf(py::module& m)
              "height"_a = 0, "depth"_a = 0)
         .def("spec", &ImageBuf::spec,
              py::return_value_policy::reference_internal)
-        .def("nativespec", &ImageBuf::nativespec,
+        .def("file_format", &ImageBuf::file_format,
```
lgritz, line 364: I think here, like with C++, for back compatibility we'll want to preserve a nativespec() call, even if it does something different than before.

lgritz:
I think this is going in the right direction. Looks like failed tests? And you need a DCO?

lgritz:
Just a remind that there are some pending items to finish this

bfraboni:
@lgritz I'll get back to this task soon (on the side but not with a high priority because it is API breaking), and see how I can fix the failing tests. I have more experience now that I know how to run and debug them.

Issue #4424, opened 2024-09-13
[FEATURE REQUEST] HDR Gain map support


filmkorn:
**Is your feature request related to a problem? Please describe.**
- Providing compatible HDR content for an increasing number of HDR capable devices.

**Describe the solution you'd like**
Proposing to add support for gain maps in OIIO to read/write backwards compatible jpg images with gain maps.
- Add reader and writer plugins for libultrahdr. Existing support in Chrome to view these seems beneficial.

**Describe alternatives you've considered**
- ImageMagick appears to have support for UltraHDR.

**Additional context**
- https://www.iso.org/standard/86775.html
- https://github.com/google/libultrahdr/
- https://github.com/libjpeg-turbo/libjpeg-turbo/issues/746

lgritz:
This is apparently an extension to JPEG that stashes an extra part to the file that includes the HDR information (readers that don't know about this just see the part that's the ordinary SDR JPEG image.

This would be great for somebody to tackle as a modification to our JPEG reader.

🎣 Mackerel!

mugulmd:
I can take a stab at it if that's alright with everyone!

lgritz:
Yes, please, @mugulmd !

lgritz:
Implemented by #4484

PR #4484, opened 2024-10-08
feat(jpeg): Support reading Ultra HDR images

mugulmd:
## Description

Initial feature request: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4424

Add support in the `jpeg` plugin for reading Ultra HDR images using the reference codec `libultrahdr`: https://github.com/google/libultrahdr

## Tests

Images used for testing during development: https://github.com/MishaalRahmanGH/Ultra_HDR_Samples

## Checklist:

  doesn't change the API.)
  nonconforming.


src/jpeg.imageio/jpeg_pvt.h
```diff
@@ -32,6 +32,8 @@ extern "C" {
 #    define OIIO_JPEG_LIB_VERSION JPEG_LIB_VERSION
 #endif
 
+#include "ultrahdr_api.h"
```
lgritz, line 35: Certainly this should only happen if the ultrahdr library was found at build time.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -453,6 +493,17 @@ JpgInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+        uhdr_raw_image_t* uhdr_raw = uhdr_get_decoded_image(m_uhdr_dec);
+
+        const size_t row_size = uhdr_raw->stride[UHDR_PLANE_PACKED] * 8;
+        unsigned char* top_left = static_cast<unsigned char *>(uhdr_raw->planes[UHDR_PLANE_PACKED]);
+        unsigned char* row_data_start = top_left + row_size * y;
+        memcpy(data, row_data_start, row_size);
+
+        return true;
+    }
+
```
lgritz, line 506: This is going to need to be guarded by some `#if` to only compile if the uhdr library is found at build time.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -494,6 +545,9 @@ JpgInput::close()
         if (m_decomp_create)
             jpeg_destroy_decompress(&m_cinfo);
         m_decomp_create = false;
+        if (m_use_uhdr)
+            uhdr_release_decoder(m_uhdr_dec);
```
lgritz, line 549: Needs an `#if` guard

lgritz (changes requested):

src/cmake/externalpackages.cmake
```diff
@@ -85,6 +85,8 @@ else ()
     checked_find_package (JPEG REQUIRED)
 endif ()
 
+checked_find_package (libuhdr REQUIRED)
```
lgritz, line 88: If it's required, we're going to need an auto-builder for it. But I would recommend that it be an optional dependency for the time being.

src/jpeg.imageio/jpeg_pvt.h
mugulmd: Completely agree, I made the `libuhdr` dependency optional and updated the code accordingly :)

src/cmake/externalpackages.cmake
zachlewis: I made a little build recipe that I've tested with the vfx2022 ci image:

save this as `src/cmake/build_libuhdr.cmake`:

```
# Copyright Contributors to the OpenImageIO project.

# libuhdr by hand!

set_cache (libuhdr_BUILD_VERSION 1.2.0 "libuhdr version for local builds")
set (libuhdr_GIT_REPOSITORY "https://github.com/google/libultrahdr")
set (libuhdr_GIT_TAG "v${libuhdr_BUILD_VERSION}")

set_cache (libuhdr_BUILD_SHARED_LIBS ${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
           DOC "Should execute a local libuhdr build, if necessary, build shared libraries" ADVANCED)

build_dependency_with_cmake(libuhdr
    VERSION         ${libuhdr_BUILD_VERSION}
    GIT_REPOSITORY  ${libuhdr_GIT_REPOSITORY}
    GIT_TAG         ${libuhdr_GIT_TAG}
    CMAKE_ARGS
        -D BUILD_SHARED_LIBS=${libdeflate_BUILD_SHARED_LIBS}
        -D CMAKE_INSTALL_LIBDIR=lib
    )

set (libuhdr_ROOT ${libuhdr_LOCAL_INSTALL_DIR})

find_package(libuhdr REQUIRED)

set (libuhdr_VERSION ${libuhdr_BUILD_VERSION})

if (libuhdr_BUILD_SHARED_LIBS)
    install_local_dependency_libs (uhdr uhdr)
endif ()
```

mugulmd: Thanks a lot !
Should I add it even though `libuhdr` is an optional dependency ?

zachlewis: ah -- you know what, if it's causing problems, maybe the build recipe should be a separate PR. 
 (although, if I'm understanding correctly, it might just be a matter of adding one more line to CMAKE_ARGS: `-D CMAKE_POSITION_INDEPENDENT_CODE=ON`

mugulmd: Seems like this is fixing *almost* everything, however the ICC and Windows builds are still failing.
Should I remove these commits from the PR and create a separate one later for the build recipe ?

mugulmd:
@lgritz quick question: I will need to add at least one Ultra HDR image for the test suite. The ones I use for testing locally weigh approximately 2M. Should I add them directly to the OpenImageIO repo, or should I create a PR in OpenImageIO-images to add them there ?

src/cmake/externalpackages.cmake
zachlewis: Hmmmm. Might be easier to troubleshoot the build recipe if everything else were comfortably in place; and I don't want the build recipe stuff to halt progress on the rest of your PR, as it's kind of a separate problem. Up to you guys.

As for the errors, I'm realizing that I accidentally left in `-D BUILD_SHARED_LIBS=${libdeflate_BUILD_SHARED_LIBS}` instead of `${libuhdr_BUILD_SHARED_LIBS}`...! I think fixing that will permit Windows to build, install, and link dynamic libuhdr instead of static, which seems to be less troublesome for Windows.

For the ICC check, not quite sure what the fix is...

lgritz:
Images repo, please

src/cmake/externalpackages.cmake
mugulmd: Fine by me !
Unfortunately that didn't fix the Windows build, it looks like CMake has troubles locating the libs...
Do you know if there's a way to inspect the CI images easily ? Could be a good starting point for troubleshooting

zachlewis: I don't have a very sophisticated way of testing locally - I'm working on top of a `aswf/ci-osl:2022-clang13` docker image, and looking at `src/build_scripts/gh-installdeps.bash` and `ci-startup.bash`. But I'm not sure if there's a way to test `icc` outside the CI environment...

Windows is another problem, I don't have a way to test locally. 

It's not pretty, but it might be a matter of just trying stuff and rifling through github's build logs.

The [linux-vfx2022-icc failures](https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/11281030497/job/31375343120?pr=4484#step:8:958) all seem to pertain to C++17 aligned allocation functions; and the problems are only occurring with icc-2022.1, not with icx-2023.1, which is interesting.

In a worst case scenario, would it be cheating if we disabled uhdr support just for the linux-vfx2022-icc check? 

Looking at https://github.com/google/libultrahdr/blob/main/docs/building.md, we could try setting `-D UHDR_ENABLE_LOGS=ON` to increase build verbosity... beyond that, I'm not sure what to suggest.

Also, it looks like there's a constraint on maximum resolution that defaults to 8192x8192. Maybe we should set `-D UHDR_MAX_DIMENSION=65535` or something?

mugulmd: Okay so little investigation updates:
- Windows build failure -> looks like `libultrahdr` does not have any `install` target for Windows: https://github.com/google/libultrahdr/blob/b35a75dfbaf67e066b69f79860dfe7b429ad1485/CMakeLists.txt#L739 (or did I understand it the wrong way ?)
- ICC -> maybe the same bug with precompiled headers as in this issue: https://stackoverflow.com/questions/74541813/why-do-i-get-error-reference-to-align-val-t-is-ambiguous-with-intels-c-2
- Ubuntu gcc14 -> this one is new but from the log this doesn't seem related to the PR, no idea what's going on...

mugulmd: Okay so it seems I fixed the windows build, but now the test suite is failing and the error is a bit more obscure, not sure I understand the problem

mugulmd: Windows CI is fixed !
However the ICC build and now the API check are failing. @lgritz @zachlewis any ideas ?

src/cmake/build_libuhdr.cmake
```diff
@@ -0,0 +1,47 @@
...
+        -D BUILD_SHARED_LIBS=${libuhdr_BUILD_SHARED_LIBS}
+        -D CMAKE_INSTALL_LIBDIR=lib
+        -D CMAKE_POSITION_INDEPENDENT_CODE=ON
+        -D UHDR_BUILD_EXAMPLES=FALSE
+        -D UHDR_ENABLE_LOGS=TRUE
+    )
+
+if (WIN32)
+    execute_process (COMMAND mkdir -p ${libuhdr_LOCAL_INSTALL_DIR}/lib)
+    execute_process (COMMAND cp ${libuhdr_LOCAL_BUILD_DIR}/Release/*.lib ${libuhdr_LOCAL_INSTALL_DIR}/lib)
```
lgritz, line 30: 'cp' feels like it might not reliably work on Windows, no? Would it be more portable to use

    file (COPY ...)

or

    file (INSTALL ...)

??? (And that might even create the directory automatically, I'm not sure?

src/jpeg.imageio/CMakeLists.txt
```diff
@@ -2,8 +2,19 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
-add_oiio_plugin (jpeginput.cpp jpegoutput.cpp
-                 LINK_LIBRARIES
-                     $<TARGET_NAME_IF_EXISTS:libjpeg-turbo::jpeg>
-                     $<TARGET_NAME_IF_EXISTS:JPEG::JPEG>
-                 )
+if (libuhdr_FOUND)
+    add_oiio_plugin (jpeginput.cpp jpegoutput.cpp
```
lgritz, line 6: I think all you need here is

    if (libuhdr_FOUND)
        set (UHDR_DEFS USE_UHDR)
    endif()

and then that first clause of your 'if' is valid for all cases, because LIBUHDR_INCLUDE_DIR and LIBUHDR_LIBRARY and UHDR_DEFS all will be defied only if libuhdr_FOUND is true, and will safely be empty if not.

src/jpeg.imageio/jpeg_pvt.h
```diff
@@ -32,6 +32,10 @@ extern "C" {
 #    define OIIO_JPEG_LIB_VERSION JPEG_LIB_VERSION
 #endif
 
+#if defined(USE_UHDR)
+#    include "ultrahdr_api.h"
```
lgritz, line 36: Shouldn't this be `<ultrahdr_api.h>`?

lgritz:
What's happening with icc is that the libuhdr's build is passing parameters to the compiler that it doesn't recognize. Its build system probably assumes that if it's building on Linux, it must be either gcc or clang. But icc doesn't take those particular arguments.

I think the easiest thing is to just set the icc test (in ci.yml) to disable the use of libuhdr by adding to its `setenvs: ` the extra definition `DISABLE_libuhdr=1`. The icc compiler is basically discontinued, so it's really not important to test that particular case with libuhdr support. It's not worth the trouble to try to fix it for an optional dependency.

src/jpeg.imageio/jpeg_pvt.h
mugulmd: Good spot, thanks !

src/cmake/build_libuhdr.cmake
mugulmd: You're right it's definitely a better alternative, code corrected !

src/jpeg.imageio/CMakeLists.txt
mugulmd: Indeed that's cleaner, thanks :)

mugulmd:
Once I get the CLA authorized by my company I'll be able to merge the PR containing the test image and then I'll re-introduce the test case in the test suite :)

lgritz (approved):
LGTM, thanks!

We'll leave this pending the CLA and adding a proper test including source in OpenImageIO-images. When those items are completed, I think it's ready to merge.

lgritz:
I think the libheif failures are unrelated to this PR. Just ignore those for the moment, I'll address them separately.

mugulmd:
It seems that the test I added for reading Ultra HDR images is not passing in the CI for some reason. I just checked it's passing on my local setup, so it must be an issue with how we build/link libultrahdr. I'll try to fix that :)

src/cmake/build_libuhdr.cmake
```diff
@@ -0,0 +1,53 @@
...
+set (libuhdr_GIT_TAG "v${libuhdr_BUILD_VERSION}")
+
+set_cache (libuhdr_BUILD_SHARED_LIBS OFF
+           DOC "Should execute a local libuhdr build, if necessary, build shared libraries" ADVANCED)
+
+find_package(JPEG REQUIRED)
+message("[libuhdr] JPEG_FOUND=${JPEG_FOUND}")
+message("[libuhdr] JPEG_INCLUDE_DIRS=${JPEG_INCLUDE_DIRS}")
+message("[libuhdr] JPEG_LIBRARIES=${JPEG_LIBRARIES}")
+message("[libuhdr] JPEG_VERSION=${JPEG_VERSION}")
```
lgritz, line 20: I think you shouldn't need this here at all, since in externalpackages.cmake `checked_find_package(libuhdr)` is called after the JPEG libraries are already found.

src/cmake/externalpackages.cmake
```diff
@@ -226,6 +226,10 @@ checked_find_package (fmt REQUIRED
 get_target_property(FMT_INCLUDE_DIR fmt::fmt-header-only INTERFACE_INCLUDE_DIRECTORIES)
 
 
+# Ultra HDR
+checked_find_package (libuhdr)
+
+
```
lgritz, line 232: This is fine here, but maybe for organizational clarity, I would recommend finding libuhdr immediately after the search for JPEG, since they are intimately related?

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -406,6 +414,86 @@ JpgInput::read_icc_profile(j_decompress_ptr cinfo, ImageSpec& spec)
...
+        desc      = TypeDesc::UINT8;
+        break;
+    default:
+        errorfmt("Unsupported Ultra HDR image format: {}", int(uhdr_raw->fmt));
+        uhdr_release_decoder(m_uhdr_dec);
+        return false;
+    }
+
+    ImageSpec newspec = ImageSpec(uhdr_raw->w, uhdr_raw->h, nchannels, desc);
+    newspec.extra_attribs = m_spec.extra_attribs;
```
lgritz, line 486: I think that instead of a copy here, you can just move it.

```suggestion
    newspec.extra_attribs = std::move(m_spec.extra_attribs);
```

lgritz (commented):
This isn't a full review, since all the tests aren't passing yet. But looking it over, I have just a few comments. The rest looks fine so far.

src/cmake/build_libuhdr.cmake
mugulmd: Ah yeah sorry I'm just trying to debug the CI, I did not intend to leave that here !

lgritz:
Just out of curiosity, what was the fix that addressed the failed CI tests?

lgritz:
You have just one test (icx) failing. Once that is fixed, I think this is ready to get final review and be merged.

mugulmd:
@lgritz it seems libuhdr had trouble locating the jpeg library. I think (but I'm not 100% sure) that it comes from the `find_package(JPEG)` instruction in libuhdr's CMakeLists.txt, which will redefine the `JPEG_INCLUDE_DIRS` and `JPEG_LIBRARIES` variables, and make them point to a different JPEG install (as it seems most of the containers used in the CI already have a JPEG install by default, but we're trying to use the jpeg-turbo that we manually installed).
By looking into CMake's official `FindJPEG.cmake`, I noticed that the `JPEG_INCLUDE_DIR` and `JPEG_LIBRARY` variables are not necessarily redefined, therefore we can use them to point to the jpeg-turbo install, and pass them to libuhdr's build process.
I also noticed that something similar was done for libTIFF, so I figured this might be a good idea, and apparently I got lucky :)

mugulmd:
Hey @lgritz 
I just did a few checks on the CI and it seems the icx test is failing only because libuhdr gets compiled with icx, so it seems to be a compiler issue. I have to admit I don't know much about icx, so can I suggest we override the compiler with g++ for the moment on that specific CI run ? I see we're already doing that for OCIO.
(we can also create an issue on the libuhdr repo, it's possible they never tested compiling with icx and there might be some extra work to do on their side to support it properly)

src/cmake/build_libuhdr.cmake
```diff
@@ -20,6 +20,18 @@ if (TARGET libjpeg-turbo::jpeg)
...
+else ()
+    set (UHDR_CMAKE_C_COMPILER ${CMAKE_C_COMPILER})
+endif()
+
+if (DEFINED ENV{UHDR_CXX})
+    set (UHDR_CMAKE_CXX_COMPILER $ENV{UHDR_CXX})
+else ()
+    set (UHDR_CMAKE_CXX_COMPILER ${CMAKE_CXX_COMPILER})
+endif()
+
```
lgritz, line 34: I think this will be simpler, using our `set_cache` utility that we can give a default but allow automatic override by matching env variable:

```
set_cache (UHDR_CMAKE_CXX_COMPILER ${CMAKE_CXX_COMPILER} "libuhdr build compiler override" ADVANCED)
set_cache (UHDR_CMAKE_C_COMPILER ${CMAKE_C_COMPILER} "libuhdr build compiler override" ADVANCED)
```

.github/workflows/ci.yml
```diff
@@ -121,6 +121,8 @@ jobs:
             simd: "avx2,f16c"
             setenvs: export USE_OPENVDB=0
                             OPENCOLORIO_CXX=g++
+                            UHDR_C=gcc
+                            UHDR_CXX=g++
```
lgritz, line 125: No need to add 2 env variables with yet additional names.

```
UHDR_CMAKE_C_COMPILER=gcc
UHDR_CMAKE_CXX_COMPILER=g++
```

lgritz (changes requested):
Minor suggestion, but otherwise this is great.

lgritz (approved):
LGTM, this looks great.

Thanks, this is a really interesting feature. I hope that this convention is embraced and we see lots of images that look like old school JPEGs to naive viewers but are HDR to the ones that know better.

mugulmd:
There's still quite a few features from libuhdr we could expose in oiio, but yeah that felt like a pretty good starting point !

Issue #4483, opened 2024-10-08
[API CHANGE SUGGESION] Make `TextureOpt::missingcolor` a `cspan`


virtualritz:
I heard most pointers + length would be replaced with `span`s of some sort going forward.

In that spirit I would like to suggest a (breaking) change to `TextureOpt`. The title kinda sums it up already. :grin:

The change would be that *the length of the `missingcolor` does not have to match `nchannels`* any more.
It would simply be using the last channel value as fill if the cspan is shorter than nchannels or ignore channels at the end, if it is longer.

On that note: the OIIO API has quite afew naming inconistencies with its use of underscores. For example, the `set` prefix in some some methods is sometimes postfixed with an underscore and sometimes not. 

In `TextureOpt` itself the obvious example is `conservative_filter` (uses an underscore) vs `missingcolor`, `firstchannel`, `subimage`, subimagename` (don't).
It should be `conservative_filter` & `missing_color`, `first_channel`, `sub_image` and `sub_image_name`. Or neither should have an underscore.

As a non-native speaker I prefer the underscore version.

If the `missingcolor` as `cspan` change happens that may be an opprtunity to clean up naming in this struct too.

In the Rust version I btw. also postfixed any `s` & `t` settings, i.e. `s_blur` etc.

virtualritz:
No wuckers. I wasn't expecting this to make 3.0 anyway.

I track all renaming suggestions I have in a Google sheet. I figured a while ago that there are too many for any chance to make it into 3.0. :grin: And I haven't shared this sheet as I like to take the time to write a rationale for each.

But I would hope the Rust wrapper could be used drive the discussion as to what/if sth. gets renamed how & why. Some of the suggestions are already reflected there.

On that note: the wrapper now builds w/o running `babble` to generate the bindings (`bbl-build-rs` is still needed though but it will pull that in from GH).

You can have a peek at the docs by doing:

```
git clone https://github.com/phyrondev/openimageio
cd openimageio
cargo doc --no-deps --open
```

No need to have a C++ toolchain & OIIO installed at all. Issue weldome if that doesn't work as expected.

ThiagoIze:
How long do you think we'd need to wait for the next chance to add this in?

For the cspan, that sounds nice, but this would I suspect require users to modify their code to integrate this, which does sound like an imposition at this stage for a small benefit.

For naming cleanups, I think we can survive waiting longer, though I would love to see that later.

[...]

PR #4485, opened 2024-10-10
api: TextureOpt refactor

lgritz:
* Version the class for easier future expansion! Let's call it `TextureOpt_v2` and we make a `using` alias so that the name `TextureOpt` means this. We can then revise the struct in the future by not losing link compatibility.

* The macro OIIO_TEXTUREOPT_VERSION (now 2) can be used to test which is the current version.

* Get rid of TextureOpt-local enums by finally switching to the freestanding ones in the Tex namespace. (We've wanted to do this for a long time, but couldn't break compatibility.) Preserve API compatibility for now with clever use of `static constexpr` and `using`.

* Make more efficient use of space in the TextureOpt by: field order rearrangement to eliminate empty space, switching a couple of the enums that needed only a few values to be `class enum xxx : uint8_t` so they take only 1 byte instead of 4, switch `anisotropic` to be 16 bits (do we really need an anisotropic ratio > 65535?), elimiate the time and samples fields which have never been used (they were reserved for shadow maps, which we never implemented).

I think this set of changes is minimally onerous to downstream projects for the 3.0 release, while laying a path forward (via the versioning) to allow future changes to the struct to happen with a lower pain threshold.

I don't claim that this perfectly preserves downstream source compatibility, but the fact that outside of these changes to delcarations in texture.h, only 8 lines in the entire rest of the OIIO codebase needed to change to accommodate them. So I think that the impact downstream will be minimal. (OSL might be the exception, since its LLVM code generation is exquisitely sensitive to the exact layout of the TextureOpt struct, but I'm working on those changes, too.)


lgritz:
Related to #4483

src/include/OpenImageIO/texture.h
```diff
@@ -187,82 +191,72 @@ class TextureOptions;  // forward declaration
...
+        mipmode(MipMode::Default), interpmode(InterpMode::SmartBicubic),
+        conservative_filter(true), anisotropic(32),
+        sblur(0.0f), tblur(0.0f), rblur(0.0f),
+        swidth(1.0f), twidth(1.0f), rwidth(1.0f),
         fill(0.0f), missingcolor(nullptr),
-        time(0.0f), rnd(-1.0f), samples(1),
-        rwrap(WrapDefault), rblur(0.0f), rwidth(1.0f),
+        rnd(-1.0f),
         colortransformid(0),
         envlayout(0)
```
ThiagoIze, line 237: Since you're already changing it, should we replace this with a default constructor and instead do in class initialization?

ThiagoIze (approved):
This looks good to me, though I did mention an optional cleanup you could do if you want.

src/include/OpenImageIO/texture.h
lgritz: Good idea. Pushed an update.

lgritz: oops, NOW the update is pushed

ThiagoIze (approved):

PR #4488, opened 2024-10-12
api: Remove deprecations

lgritz:
These are all things long marked as deprecated (often since before the official launch of 2.0) that we've waited for 3.0 to finally remove. Now is the time.


jessey-git (approved):
All seems reasonable.  Neat trick there with OIIO_DEPRECATED_EXTERNAL. +1

lgritz:
> All seems reasonable. Neat trick there with OIIO_DEPRECATED_EXTERNAL. +1

Yeah, sometimes there are situations where I want to start warning downstream projects to stop using something, but there are... reasons... why it's extremely difficult to stop using it internally quite yet. So the simple thing is to rig it to not give that particular warning when compiling OIIO itself.

PR #4489, opened 2024-10-13
TS: Version TextureOptBatch like we did for TextureOpt

lgritz:
No changes at this time, but it's setting us up for future changes.


lgritz:
Withdrawing. I ran into something that indicates I'm going to have to make more changes here before I can beta. This was much more difficult to accommodate on the OSL side than I thought, need to rethink a couple things on the batch side. I will submit a new PR when it's ready.

PR #4490, opened 2024-10-14
TS: Version TextureOptBatch like we did for TextureOpt

lgritz:
I got part way through the OSL side changes to accommodate what we did
recently with TextureOpt and TextureOptBatch. Ugh, it's a lot harder than
I thought and I don't know how to finish the job without postponing the
3.0 release even more, which is unacceptable at this point. So here I do
to TextureOptBatch to get out of this mess:

* "Version" TextureOptBatch like we did for TextureOpt (should have done
  that all along).
* Revert "v1" to the same byte-for-byte layout we used to have, which
  involves using `int` for some fields as it implicitly did before, because
  the new enum classes totally change the field sizes and struct layout.
* Sketch out in comments what v2 might look like, to more fully match the
  TextureOpt. The introduction of struct versioning means we can make that
  change for 3.1, it doesn't need to wait years for 4.0.

### DIFF
--- a/src/include/OpenImageIO/texture.h
+++ b/src/include/OpenImageIO/texture.h
@@ -32,6 +32,16 @@
 
 // Revision of the TextureOpt class
 #define OIIO_TEXTUREOPT_VERSION 2
+#define OIIO_TEXTUREOPTBATCH_VERSION 1
+
+// Preprocessor utility: Concatenation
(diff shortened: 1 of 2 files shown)


lgritz:
In the interests of trying to get a 3.0 beta out the door today, I'm going to merge this. Hopefully it is fine with everybody, and if not, we can tweak it during the beta period if necessary.

PR #4493, opened 2024-10-15
BUILD_MISSING_DEPS - fix minor configuration options

kaarrot:
## Description
Address a few small issues encountered during CMake configuration step and during build using Makefile `BUILD_MISSING_DEPS` option enabled.
- Mismatch between expected libraw versions specified in `build_libraw.bash` 0.21.3 and `externalpackages.cmake` 0.20.0
- OpenEXR link error in libutil tests
- Update top level Makefile to accept `BUILD_MISSING_DEPS` as string instead of bool. Pass the string option down to CMake

## Test
Verified build completes with no errors: `BUILD_MISSING_DEPS=all make`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,11 +202,6 @@ ifneq (${CLANG_FORMAT_EXCLUDES},)
   MY_CMAKE_FLAGS += -DCLANG_FORMAT_EXCLUDES:STRING=${CLANG_FORMAT_EXCLUDES}
 endif
 
-ifneq (${BUILD_MISSING_DEPS},)
-  MY_CMAKE_FLAGS += -DBUILD_MISSING_DEPS:BOOL=${BUILD_MISSING_DEPS}
-endif
-
-
 #$(info MY_CMAKE_FLAGS = ${MY_CMAKE_FLAGS})
 #$(info MY_MAKE_FLAGS = ${MY_MAKE_FLAGS})
 
@@ -395,7 +390,6 @@ help:
 	@echo "                                  0, sse2, sse3, ssse3, sse4.1, sse4.2, f16c,"
 	@echo "                                  avx, avx2, avx512f)"
 	@echo "      TEX_BATCH_SIZE=16        Override TextureSystem SIMD batch size"
-	@echo "      BUILD_MISSING_DEPS=1     Try to download/build missing dependencies"
 	@echo "  make test, extra options:"
 	@echo "      TEST=regex               Run only tests matching the regex"
 	@echo ""
--- a/src/cmake/dependency_utils.cmake
+++ b/src/cmake/dependency_utils.cmake
@@ -426,7 +426,7 @@ macro (checked_find_package pkgname)
         #   ${pkgname}_REFIND_ARGS    : additional arguments to pass to find_package
         if (${pkgname}_REFIND)
             message (STATUS "Refinding ${pkgname} with ${pkgname}_ROOT=${${pkgname}_ROOT}")
-            find_package (${pkgname} ${_pkg_UNPARSED_ARGUMENTS} ${${pkgname}_REFIND_ARGS})
+            find_package (${pkgname} ${${pkgname}_REFIND_VERSION} REQUIRED ${_pkg_UNPARSED_ARGUMENTS} ${${pkgname}_REFIND_ARGS})
             unset (${pkgname}_REFIND)
         endif()
         # It's all downhill from here: if we found the package, follow the
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -453,10 +453,11 @@ RawInput::open_raw(bool unpack, const std::string& name,
     // Output 16 bit images
     m_processor->imgdata.params.output_bps = 16;
 
+#if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0, 21, 0)
     // Exposing max_raw_memory_mb setting. Default max is 2048.
     m_processor->imgdata.rawparams.max_raw_memory_mb
         = config.get_int_attribute("raw:max_raw_memory_mb", 2048);
-
+#endif
 
     // Disable exposure correction (unless config "raw:auto_bright" == 1)
     m_processor->imgdata.params.no_auto_bright
@@ -715,7 +716,9 @@ RawInput::open_raw(bool unpack, const std::string& name,
             crop_top    = p->get_int_indexed(1);
             crop_width  = p->get_int_indexed(2);
             crop_height = p->get_int_indexed(3);
-        } else if (m_processor->imgdata.sizes.raw_inset_crops[0].cwidth != 0) {
+        }
+#if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0, 21, 0)
+        else if (m_processor->imgdata.sizes.raw_inset_crops[0].cwidth != 0) {
             crop_left   = m_processor->imgdata.sizes.raw_inset_crops[0].cleft;
             crop_top    = m_processor->imgdata.sizes.raw_inset_crops[0].ctop;
             crop_width  = m_processor->imgdata.sizes.raw_inset_crops[0].cwidth;
@@ -723,7 +726,16 @@ RawInput::open_raw(bool unpack, const std::string& name,
             left_margin = m_processor->imgdata.sizes.left_margin;
             top_margin  = m_processor->imgdata.sizes.top_margin;
         }
-
+#else
+        else if (m_processor->imgdata.sizes.raw_inset_crop.cwidth != 0) {
+            crop_left   = m_processor->imgdata.sizes.raw_inset_crop.cleft;
+            crop_top    = m_processor->imgdata.sizes.raw_inset_crop.ctop;
+            crop_width  = m_processor->imgdata.sizes.raw_inset_crop.cwidth;
+            crop_height = m_processor->imgdata.sizes.raw_inset_crop.cheight;
+            left_margin = m_processor->imgdata.sizes.left_margin;
+            top_margin  = m_processor->imgdata.sizes.top_margin;
+        }
+#endif
         if (crop_width > 0 && crop_height > 0) {
             ushort image_width  = m_processor->imgdata.sizes.width;
             ushort image_height = m_processor->imgdata.sizes.height;


src/cmake/externalpackages.cmake
```diff
@@ -57,7 +57,6 @@ checked_find_package (Imath REQUIRED
 
 checked_find_package (OpenEXR REQUIRED
     VERSION_MIN 3.1
-    NO_FP_RANGE_CHECK
```
lgritz, line 60: Why? I must have had that in for a reason (though now I'm not sure what problem it solved).

src/cmake/externalpackages.cmake
```diff
@@ -161,7 +160,7 @@ checked_find_package (Libheif VERSION_MIN 1.11
                       RECOMMEND_MIN_REASON "for orientation support")
 
 checked_find_package (LibRaw
-                      VERSION_MIN 0.20.0
+                      VERSION_MIN 0.21.3
```
lgritz, line 163: No, we don't want to do this. Changing it here means "we will not accept a version older than this." The minimum 0.20.0 was chosen purposefully.

Makefile
```diff
@@ -203,7 +203,7 @@ ifneq (${CLANG_FORMAT_EXCLUDES},)
 endif
 
 ifneq (${BUILD_MISSING_DEPS},)
-  MY_CMAKE_FLAGS += -DBUILD_MISSING_DEPS:BOOL=${BUILD_MISSING_DEPS}
+  MY_CMAKE_FLAGS += -DOpenImageIO_BUILD_MISSING_DEPS:STRING=${BUILD_MISSING_DEPS}
 endif
```
lgritz, line 207: Actually, I'd recommend removing these 3 lines entirely. Let's just document that people use `OpenImageIO_BUILD_MISSING_DEPS`. Because that variable is set up in dependency_util.cmake using a call to set_cache, it will automatically pick up a value in an environment variable, which is naturally set up by make. In other words

    make OpenImageIO_BUILD_MISSING_DEPS=all

will just work, even if Makefile doesn't mention it at all.

lgritz: set_cache is a recent addition, but I'm really liking it, so my goal is to slowly change lots more things that are `set(... CACHE)` to set_cache (and `option` to `set_option`) and then as we do so, eliminate any special processing of those variables in the Makefile. My goal is to eventually get this Makefile really down the the minimum wrapper needed to have `make` give you a decent default build without needing to know anything about cmake.

src/cmake/externalpackages.cmake
lgritz: To explain this even better:

The versions in externalpackages.cmake express the oldest version we will accept. This is some kind of compromise between wanting it convenient for users who for various reasons are in an environment with versions of these packages that may be as much as a few years old, versus not wanting to have to require OIIO to have too much extra logic to cater to older APIs or lack of newer features in those dependencies. Also, we never raise minimum versions except at the annual release (like 2.4 -> 2.5), never at an intermediate patch (like 2.4.1 -> 2.4.2) which are guaranteed not to change build requirements.

The versions in build_foo.cmake or build_foo.bash express the version of the dependency that we will choose to download and build ourselves if an adequate one is not found in the environment. Those will tend to be fairly recent, though not necessarily the very latest one -- depends on how confident we feel that the latest has been thoroughly tested and is solid.

So in general, we expect the "oldest accepted" and "default for self-building" to be different from one another.

kaarrot: While building from master I run into [recent changes not available in 0.20.0](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4454/files) Had to use build_libraw.bash to get the updates.

lgritz: Ah, I see. I think that this means that we should have a `#if` around just those couple lines, like

    #if LIBRAW_VERSION >= LIBRAW_MAKE_VERSION(0, 21, 0)

And then in the docs where we describe `raw:max_raw_memory_mb`, just point out that it requires LibRaw >= 21 for it to be honored.

Can you add that to this PR, please, so then we don't have to raise the minimum libraw version just for that one feature? Thanks.

kaarrot: >So in general, we expect the "oldest accepted" and "default for self-building" to be different from one another.

Thanks, for detailed explanation - that makes sense. 

>Can you add that to this PR  so then we don't have to raise the minimum libraw version

Of course, I'll add it.

src/cmake/externalpackages.cmake
kaarrot: I think I see what's happening. I ran into this problem, as I currently have another version of `OpenEXR 2.5.7` installed on the system, which is too old compared to the one specified by `checked_find_package()` -[ which is 3.1](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/db9fe349d007feeceb25f3b6fec8f12ff54d6ea3/src/cmake/externalpackages.cmake#L59).

With  `NO_FP_RANGE_CHECK` enabled, we should handle and always update [_pkg_VERSION_MIN](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/db9fe349d007feeceb25f3b6fec8f12ff54d6ea3/src/cmake/dependency_utils.cmake#L352) as well in order to set version range to the maximum range available (ie 0.0.1). Currently we don't do that which puts `2.5.7` outside of the valid range - which triggers a local build for version `3.x` .

Now, with two versions available the '[REFIND](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/db9fe349d007feeceb25f3b6fec8f12ff54d6ea3/src/cmake/dependency_utils.cmake#L429)' step does not enforce which version `find_package()` should select. It happens (at least in my case) that it picks the original one and not the one just built.
So in other words I'm wondering if we should also specify  `${${pkgname}_REFIND_VERSION} REQUIRED` to the `find_package()` in order to enforce that?

lgritz: I remember why the NO_FP_RANGE_CHECK now. So just as background, this is a signal to checked_find_package to NOT pass a version number (or range) to the underlying call to find_package. Now, why would we want to do that?

The dependency's exported BlahConfig.cmake/etc configuration will verify its declared version against that being demanded by the find_package call. There are a few methods that can be specified to write_basic_package_version_file when you have it create the BlahConfigVersion: AnyNewerVersion, SameMajorVersion, SameMinorVersion, ExactVersion.

The problem is that most packages export using SameMajorVersion. And that's probably the right choice for them, in the sense that the authors of a package such as OpenEXR know they have good source compatibility within a major version, but not between different major versions.

But we often want to make OIIO able to accept quite a wide range of possible versions of a dependency -- including spanning across major version boundaries, which we handle explicitly through the necessary `#if` statements and whatnot so that we really are able to use either major version. So what do we ask for? If we ask for OpenEXR 2.5 but find 3.1, we don't want the rejection we would get with OpenEXR's declaration that it's not compatible across major versions. 

This is partly due to kind of a misunderstanding. I think that when people write `find_package (Mypackage 2.5)` they probably *mean* "a minimum of 2.5", and don't realize that this won't work with 3.1 unless the upstream package said AnyNewerVersion. You just can't override the matching rules on the downstream/requesting end.

So to get around it, we use NO_FP_RANGE_CHECK to suppress our declaring the range at all, so it's not rejected by find_package. We do our own range check instead.




The problem is that with

kaarrot: I see, thanks for clarification.
One more question regarding the 'Refind' step. If during the build we determine to build local version of the library, as the existing one (found on the system) is too old. Should the `find_package` in the 'Refind' step actually only search for the local version. I'm not sure in what order CMake performs the scan but it feels there's a chance to pick up the version we don't want to use, no?

lgritz: Yes, if we build our own, we should try hard to ensure that the re-find only can get the one we just built.

I think we're doing great when the dependency isn't found at all and we build it.

But I've had some trouble in the past with the case where it's found but too old. So it's possible that we are still missing the right cmake arguments somewhere to 100% ensure that we can't pick up the old one.

lgritz: Specifically, we want the re-find to be rigged to ONLY look in the `${PROJECT_NAME}_LOCAL_DEPS_ROOT` destination where we put the auto-built version, and not in the system areas, and also we probably want the include area under there to absolutely be listed before any system include areas so that we don't accidentally pick up the system (wrong version) headers for anything we tried to build locally.

lgritz:
This all looks great, I think the only missing piece is that one or more of your commits lacks the DCO signoff. Maybe the easiest fix is to squash your comments and make sure the one remaining commit has the proper sign-off info?

src/cmake/externalpackages.cmake
kaarrot: I wasn't able to post this earlier but I can reproduce the 'Refind' related problem mentioned above. Here is the annotated output without the current changes:
```
make OpenImageIO_BUILD_MISSING_DEPS=all

-- Found Imath 3.1.10
-- OpenEXR  is outside the required range 3.1...10000.0.0
-- Building package OpenEXR  locally  <-------------------- Build our own version
--         Building local OpenEXR 3.2.4 from https://github.com/AcademySoftwareFoundation/OpenEXR
--         Downloading local https://github.com/AcademySoftwareFoundation/OpenEXR
HEAD is now at a1a00ff Add PR 1674 to notes
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/kuba/SRC/OIIO/oiio-kaarrot/build/deps/OpenEXR-build
gmake[1]: Entering directory '/home/kuba/SRC/OIIO/oiio-kaarrot/build/deps/OpenEXR-build'
...
...
...
-- Up-to-date: /home/kuba/SRC/OIIO/oiio-kaarrot/build/deps/dist/include/OpenEXR/ImfUtilExport.h
gmake[1]: Leaving directory '/home/kuba/SRC/OIIO/oiio-kaarrot/build/deps/OpenEXR-build'
-- Refinding OpenEXR with OpenEXR_ROOT=/home/kuba/SRC/OIIO/oiio-kaarrot/build/deps/dist
-- Found OpenEXR 2.5.7    <-------------------- Still selects system library 
-- Found libjpeg-turbo 3.0.4
```
Passing  `${${pkgname}_REFIND_VERSION} REQUIRED` addresses that.

kaarrot:
Squashed and signed-off.

lgritz (approved):
LGTM

PR #4498, opened 2024-10-22
fix(iinfo): iinfo was not reading MIP levels correctly

lgritz:
A logical error was causing us to not correctly cycle through the successive MIP levels. Somehow, we missed this all these years.

You can tell that there was something wrong all along by examining the testsuite/iinfo test output (which we have updated here) -- in the old/wrong one, the statistics "FiniteCount" suspiciously stayed the same for all MIP levels, which is clearly wrong. Now you can see that it does correctly reflect the reduced resolution of each MIP level.

This bug was limited to 'iinfo' itself. There wasn't any inherent problem with reading MIP levels directly with ImageInput or ImageBuf.

The line with the bug has been unchanged for 14 years! We were trying to be clever about not double-reading, except that the only context in which this was called, that would never have  happened anyway, so that cleverness (which was incorrectly implemented) was unnecessary.

### DIFF
--- a/src/iinfo/iinfo.cpp
+++ b/src/iinfo/iinfo.cpp
@@ -60,9 +60,6 @@ static bool
 read_input(const std::string& filename, ImageBuf& img, int subimage = 0,
            int miplevel = 0)
 {
-    if (img.subimage() >= 0 && img.subimage() == subimage)
-        return true;
-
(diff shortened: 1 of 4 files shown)


lgritz:
No objections -> merging

PR #4499, opened 2024-10-22
[feat]IBA:demosaic add white balancing

antond-weta:
## Description

Adds white balancing functionality to IBA:demosaic

## Tests

Updated unittests

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (commented):
This LGTM. It's still marked as "draft". Is it ready to be reviewed and merged?

antond-weta:
Hi Larry, if you're happy with the code, I just need to update the docs and add some tests.

I've also just realised we could add an "auto" or "meta" value to the Bayer layout and white balance parameters, which would pull the values from the ImageBuf attributes.

lgritz:
I'm happy with the code. Docs+tests should take it over the finish line.

I'm not quite sure what you mean about the ImageBuf attributes? Do you mean that when the image is read in, you expect the raw reader would drop hints in the spec about the default white balancing method to use, so IBA::demosaic should honor that if it gets no other overriding instructions? Or did you mean something different?

antond-weta:
Yep, exactly that. The raw reader currently adds the "raw:FilterPattern" attribute to the output when asked not to debayer. We could do something similar for white balancing. We can leave this for later, as adding an auto white balancing mode should not change the default behaviour, I think. If we were to add the "auto" layout pattern, it should become the default, as the current default behaviour is to assume RGGB. This is beyond the scope of this change though. 

I'll update the docs, tests and the python bindings.

antond-weta:
Done!
Larry, could you please check my python bindings, I'm not sure if there is a better way.

src/doc/imagebufalgo.rst
```diff
@@ -1166,19 +1166,24 @@ Shuffling channels
           ImageSpec hint;
           hint["raw:Demosaic"] = "none";
           ImageBuf Src ("test.cr3", 0, 0, nullptr, &hint);
-          ParamValue options[] = { { "layout", "GRBG" } };
+          float wb[4] = {2.0, 0.8, 1.2, 1.5};
+          ParamValue options[] = {
+                { "layout", "GRBG" },
+                ParamValue("white-balance", TypeFloat, 4, wb)
+          };
```
lgritz, line 1173: Not requesting this as part of this PR, but just noting that at some point, we should do the thing where we move these actual code snippets to the appropriate parts of testsuite/docs-examples-cpp and testsuite/docs-examples-cpp, and then just include them by reference in the docs, so that it is confirmed that they compile and work correct with every CI run.

src/oiiotool/oiiotool.cpp
```diff
@@ -3458,14 +3458,30 @@ OIIOTOOL_OP(warp, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
...
+    std::string wb = op.options()["white_balance"];
+    string_view str(wb);
+    float white_balance[4] = { 1.0f, 1.0f, 1.0f, 1.0f };
+    if (str.size() && Strutil::parse_float(str, white_balance[0])
+        && Strutil::parse_char(str, ',')
+        && Strutil::parse_float(str, white_balance[1])
+        && Strutil::parse_char(str, ',')
+        && Strutil::parse_float(str, white_balance[2])
+        && Strutil::parse_char(str, ',')
+        && Strutil::parse_float(str, white_balance[3])) {
```
lgritz, line 3479: I think there is a more compact alternative, and also more correct, because it will fail if the string contains any extra non-whitespace after the 4th value.


```suggestion
    string_view str(wb);
    Strutil::remove_trailing_whitespace(str);
    float white_balance[4] = { 1.0f, 1.0f, 1.0f, 1.0f };
    if (Strutil::parse_values(str, "", white_balance, ",") && str.size()) {
```

src/python/py_imagebufalgo.cpp
```diff
@@ -3106,10 +3112,12 @@ declare_imagebufalgo(py::module& m)
 
         .def_static("demosaic", &IBA_demosaic, "dst"_a, "src"_a,
                     "pattern"_a = "", "algorithm"_a = "", "layout"_a = "",
-                    "roi"_a = ROI::All(), "nthreads"_a = 0)
+                    "wb_r"_a = 1.0f, "wb_g1"_a = 1.0f, "wb_g2"_a = 1.0f,
+                    "wb_b"_a = 1.0f, "roi"_a = ROI::All(), "nthreads"_a = 0)
```
lgritz, line 3116: This part, I'm not so sure about. Would it be better to pass white_balance as one parameter that expects a tuple or array (checked to be 4 floats), versus 4 separate float parameters?

I don't have a strong feeling one way or the other. I'm just wondering if that would be more symmetric with the C++ API, and/or more idiomatic for a Python native speaker?

testsuite/oiiotool-demosaic/run.py
```diff
@@ -10,75 +10,109 @@
...
-        "ext" : "exr",
-        "threshold": 0.0000003
-    },
-    {
-        "type" : "half",
-        "ext" : "exr",
-        "threshold": 0.0005
-    },
+# Create a test image with color gradient
+make_testimage = f"--pattern fill:topleft=0,0,1:topright=0,1,0:bottomleft=1,0,1:bottomright=1,1,0 {width}x{height} 3 "
```
lgritz, line 14: This run.py has changed a lot. I'm not entirely sure I understand how it works. I totally trust you on this, but I urge you to run this test in isolation and make sure it doesn't seem like it takes too long. There was a point there I visited this test not long ago because it was extraordinarily expensive, noticed a ton of redundant work being done, rewrote it a bit, and sped up by 3x (see #4429). I just want to make sure we aren't altering it in a way that goes back to this test being disproportionately expensive versus the narrow feature it's trying to test.

lgritz (changes requested):
Also, I think before this is ready to merge, we will need the docs to be updated for all of C++ (imagebufalgo.h), Python (pythonbindings.rst), and oiiotool (oiiotool.rst), to explain the new parameters.

src/oiiotool/oiiotool.cpp
antond-weta: Done! I've also decided to allow 3 values, for RGB, and extend them to RGGB automatically.

src/python/py_imagebufalgo.cpp
antond-weta: I've changed that to use tuples, seems to be more inline with other python bindings

testsuite/oiiotool-demosaic/run.py
antond-weta: It works exactly the same way as before, I just changed the code to generate the 4 layouts (RGGB, GRBG, GBRG, BGGR) programmatically, using a nested 2x2 loop. That reduces the code duplication, mostly in the part where we generate the bayer patterns. Since some tests now require un-whitebalanced images, there would have been even more code duplication without that.

The code still uses your trick with pre-generating images. It works slightly slower, as it writes more files, but it doesn't write more files than necessary. Runs pretty fast for me.

src/doc/pythonbindings.rst
```diff
@@ -3680,6 +3680,24 @@ Color manipulation
         ImageBufAlgo.unpremult (A, A)
 
 
+.. py:method:: ImageBuf ImageBufAlgo.demosaic (src, pattern="", algorithm="", layout="", white_balance=py::none(), "roi=ROI.All, nthreads=0)
+                bool ImageBufAlgo.demosaic (dst, src, pattern="", algorithm="", layout="", white_balance=py::none(), "roi=ROI.All, nthreads=0)
+    Demosaic a raw digital camera image.
```
lgritz, line 3685: I think there is an unbalanced quote here:

```suggestion
.. py:method:: ImageBuf ImageBufAlgo.demosaic (src, pattern="", algorithm="", layout="", white_balance=py::none(), roi=ROI.All, nthreads=0)
                bool ImageBufAlgo.demosaic (dst, src, pattern="", algorithm="", layout="", white_balance=py::none(), roi=ROI.All, nthreads=0)
    Demosaic a raw digital camera image.
```

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -2163,6 +2163,9 @@ bool OIIO_API repremult (ImageBuf &dst, const ImageBuf &src,
 ///
 ///     The order the color filter array elements are arranged in, pattern-specific.
 ///
+///   - "white-balance" : float[3] or float[4] (default: {1.0, 1.0, 1.0, 1.0})
+///
+///     Optional white-balancing weights. Can contain either three (RGB), or four (RGGB) values.
```
lgritz, line 2168: Is there a use case for specifying two green values?

src/doc/oiiotool.rst
```diff
@@ -4184,6 +4184,35 @@ current top image.
 
 
 
+.. option:: --demosaic <words>
```
lgritz, line 4187: There are no arguments that follow --demosaic, right?

```suggestion
.. option:: --demosaic
```

lgritz (changes requested):

lgritz:
I'm confused about the white balance values. 

1. When you do give 4 values, are they always in the order RGGB, or are they in the order implied by the "layout"? I think we need to be more clear about this.

2. Will anybody ever really want 4 values, with two different white balance values for the two greens? If not, a lot of things get easier if we just document the type as a Color3 instead of an array of floats that might be either 3 or 4 elements long.

src/include/OpenImageIO/imagebufalgo.h
antond-weta: Yes, I have seen some images recently with the greens in red and blue rows having distinctly different levels. Sony A99, I think.

antond-weta:
Hi Larry,

I've been mostly following exiftool, which shows WB as either RGB or RGGB, regardless of the layout:

```
exiftool G85.RW2 | grep 'WB RG'
WB RGB Levels 1                 : 612 256 399
WB RGB Levels 2                 : 657 256 372
WB RGB Levels 3                 : 703 256 354
WB RGB Levels 4                 : 420 256 575
WB RGB Levels 5                 : 632 256 393
WB RGB Levels 6                 : 576 256 400
WB RGB Levels 7                 : 420 256 575
```

```
exiftool R5.CR3 | grep 'WB RG'
WB RGGB Levels As Shot          : 2241 1024 1024 1771
WB RGGB Levels Auto             : 2241 1024 1024 1771
WB RGGB Levels Measured         : 2241 1024 1024 1771
WB RGGB Levels Daylight         : 1960 1024 1024 1913
WB RGGB Levels Shade            : 2294 1024 1024 1636
WB RGGB Levels Cloudy           : 2118 1024 1024 1777
WB RGGB Levels Tungsten         : 1348 1024 1024 2962
WB RGGB Levels Fluorescent      : 1667 1024 1024 2819
WB RGGB Levels Kelvin           : 1960 1024 1024 1913
WB RGGB Levels Flash            : 2203 1024 1024 1708
WB RGGB Levels                  : 2241 1024 1024 1771
```

Libraw stores the weights as RGBG, regardless of the layout. I find the RGGB easier to understand, because the order also represents a valid Bayer layout, so it is clear where each G sits.

lgritz:
Great, precedence from another widely used tool is a good reason for doing it a particular way.

src/include/OpenImageIO/imagebufalgo.h
antond-weta: > Is there a use case for specifying two green values?

RAW_PENTAX_K200D.PEF in the test suite has 2 different weights for the green sub-channels

antond-weta:
Thanks @kmilos, 
That is good info. I'm happy to change to RGB(G), if that's the preferred way.

lgritz:
Where are we on this? Are there any intended changes still pending, or is it in its intended final form for review?

antond-weta:
Hi @lgritz, let me change that to RGB(G) as Miloš suggested.

Also I think we should change the default ROI to be the data window, not the full size. Do you mind if I do it here instead of creating a separate PR?

lgritz:
> Also I think we should change the default ROI to be the data window, not the full size. Do you mind if I do it here instead of creating a separate PR?

Up to you. Fine to do it here.

src/doc/imagebufalgo.rst
antond-weta: I can take a look on the weekend. Marking as a draft for now.

PR #4500, opened 2024-10-22
fix: fix incorrect vertical alignment in text render

peterhorvath111:
## Description

The `y` argument in `OIIO::ImageBufAlgo::render_text()` should represent the top and bottom edge of the text if vertical alignment is set to `TextAlignY::Top` and `TextAlignY::Bottom`. However currently the rendered text is off. The current code adjusts y to point to the top of the text box before rendering the text, but it has to be adjusted to point to the baseline of the text.

## Tests

There is a existing test case, testing vertical alignment, but the reference image seems wrong and has now been updated. See the "Topleft" and "Rightbot" texts.

| Current    | Fixed |
| -------- | ------- |
| image | image |

Note: The testsuite contains a reference image for freetype v2.4.11 (aligned-freetype2.4.11.tif). I don't have that version built, and haven't been able to update the image. I'm wondering if this is still used or could be removed. I guess CI will tell.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
Lacking DCO sign-off, you'll need to add it ("git commit --amend -s" + force push).

lgritz (approved):

lgritz:
Unless you have a strong objection, @peterhorvath111, what I'm thinking is that this fix will be for 3.0 only, not backported to 2.5. Although technically a bug fix, this very much changes behavior and image appearance that somebody might be relying on.

peterhorvath111:
> Unless you have a strong objection, @peterhorvath111, what I'm thinking is that this fix will be for 3.0 only, not backported to 2.5. Although technically a bug fix, this very much changes behavior and image appearance that somebody might be relying on.

Sure, that sounds reasonable.

PR #4501, opened 2024-10-22
fix: Windows newline shows invalid character in text render

peterhorvath111:
## Description

Newline in strings on Windows is encoded as '\r\n', that is currently not handled by the text render code and so carriage return (\r) shows up in the rendered text. As a fix we simply ignore this character.
We came across this issue while exporting user specified text from DCCs to render on Windows.

## Tests

I'm not sure how to make a proper test case for this. Current tests specify a command as a string that is passed to oiiotool as arguments and executed in the command line. But adding a newline to the text breaks the command.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
Looks reasonable. Can you fix the DCO signature?

lgritz (approved):

lgritz:
This seems like a pure bug fix, so I will backport it to 2.5 for the November 1 patch release.

Issue #4504, opened 2024-10-22
Problems with font enumeration


peterhorvath111:
While using OIIO in Arnold to render text over our rendered images, we discovered a few issues regarding the current (2.5.16) font enumeration code.

- HOME environment variable is not defined on Windows. Windows instead uses `%HOMEDRIVE%%HOMEPATH%`.
- User specific system font directory is not searched on Windows. That is `%LOCALAPPDATA%/Microsoft/Windows/Fonts`.
- `$HOME/.fonts` and `$HOME/.local/share/fonts` are valid system folders on some linux distros, that are currently not searched.
- Duplicated entries in the search folders. For instance on Windows, both, `%SystemRoot%` and `C:/Windows` is added, but they are the same.
- The current implementation looks for fonts in subfolders one level deep. That allows folders that are not standard system folders. For instance on Windows subfolders are not supported natively. Also this might be expensive, for instance the current code reads all folders from `C:\Windows` and the system root on other platforms, unnecessarily.
- Fonts are now loaded from folders that are not standard system font folders. For instance `%HOME%/usr/share/fonts` on Windows, that is not a standard font folder.
- The code is hard to read, hard to tell which folders are searched/valid on each platform.

We've got fixes for some of the issues above that we can port back. 

I'd also suggest to simplify the code and instead of using prefixes and suffixes, adding only known system font folders (global and user) per each platform. And allow defining custom font folders via the `OPENIMAGEIO_FONTS` environment variable. What do you think?

lgritz:
Sounds good to me. I welcome a principled overhaul.

zachlewis:
Would you mind checking for fonts in `$OpenImageIO_ROOT/share/fonts` too?

peterhorvath111:
Created PR #4508 with the suggested changes.

PR #4508, opened 2024-10-24
Update font enumeration

peterhorvath111:
## Description

Suggestion for issue #4504.

The PR removes the code that looks for font folders based on a prefix and suffix list. Instead, adds all known system font folders per each platform. The global "font_searchpath" attribute, `OPENIMAGEIO_FONTS` and `OpenImageIO_ROOT` environment variables are still searched as before, such as fonts directories relative to the currently running binary. However $HOME and $SystemRoot folders (with all previous suffixes) are not searched anymore, thus this is a breaking change, and the documentation might need to be updated accordingly.
The search order did not change. It's still "font_searchpath" attribute, environment variables, system font folders, folders relative to binary.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_draw.cpp
+++ b/src/libOpenImageIO/imagebufalgo_draw.cpp
@@ -7,6 +7,7 @@
 #include <limits>
 #include <set>
 #include <unordered_map>
+#include <unordered_set>
 
 #include <OpenImageIO/half.h>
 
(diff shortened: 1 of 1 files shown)


lgritz:
There's still one extremely minor clang-format fix that needs to be applied...

peterhorvath111:
I'm not sure why it wants to change the indentation on that line, the current indentation should be correct. Maybe it's the macro or the one line for loop. But anyways, let's change it to pass the check.

lgritz (approved):
LGTM

lgritz:
Am I correct in understanding that the set of places that are checked on any one OS is more or less the same as always? So the primary change is that there's a more crisp per-platform search strategy, rather than a single mish-mash pile of all directories that might hold fonts on any platform?

lgritz:
Fixes #4504

peterhorvath111:
> Am I correct in understanding that the set of places that are checked on any one OS is more or less the same as always? So the primary change is that there's a more crisp per-platform search strategy, rather than a single mish-mash pile of all directories that might hold fonts on any platform?

Basically yes. Most of the system folders, that are listed now, were searched before, except a few that was missing, such as:
- `%LOCALAPPDATA%/Microsoft/Windows/Fonts` on Windows
- `$HOME/.fonts` and `$HOME/.local/share/fonts` on Linux

And folders that are not specific to the given platform are now removed. For instance if you had fonts in `$HOME/share/fonts` on Mac for some reason, now it won't load automatically. You have to explicitly add that folder (or `$HOME`) to the `OPENIMAGEIO_FONT` env var.

Issue #4502, opened 2024-10-22
[FEATURE REQUEST] Lookup font in text render based on family and style name


peterhorvath111:
Currently, fonts are identified by the name of the file that defines the font. In Arnold we extended this logic to be able to find fonts based on the font family and style name. This makes it easier for end users to select fonts. For instance "Arial Bold" is defined by arialbd.ttf on Windows. Specifying arialbd is not that obvious and also file names could differ based on the platform.

If you find this useful, we can port our code back here.

I'd also like to point out that it's not that obvious from the current documentation that *fontname* supposed to be the file name and not the family name.

> **fontsize/fontname** – Size and name of the font. If the name is not a full pathname to a font file, it will search for a matching font, defaulting to some reasonable system font if not supplied at all), and with a nominal height of fontsize (in pixels).

Specially because some examples suggest the family name should work, like this:
```
ImageBufAlgo::render_text(A, x, y, "Centered", 48, "Courier New");
```
Not sure about other platforms, but for instance on Windows this does not work currently, because "Courier New" is defined by `cour.ttf`.

peterhorvath111:
Created PR #4509 with the suggested changes.

PR #4509, opened 2024-10-24
feat: Lookup font in text render based on family and style name

peterhorvath111:
## Description

PR for issue #4502.

The PR adds the new `init_font_families()` function, that iterates over the available font files, reads family and style name via the freetype API and caches them. The function is called from `resolve_font()` on demand.

## Tests

Added a new test to the oiiotool-text folder, that loads a font by the name "Droid Serif Bold".

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
Fixes #4502.

lgritz:
I don't think it works for me to just add that comment above to link to the issue, since I'm not the PR author. But I did it explicitly with the control panel on the right.

For future reference, you need to use very specific language in your original PR description:

```
Closes #xxxx
```
or
```
Fixes #xxxx
```

and that tells GitHub to close the issue automatically when the PR is merged.

lgritz (approved):
LGTM, thanks for this enhancement!

PR #4510, opened 2024-10-25
fix: address fmt exceptions for left justification

lgritz:
I don't understand why, but for certain combinations of gcc, fmt, etc., we were getting exceptions from fmt. I finally caught it in the act! It seems related to formatting specs that looks like `{:-20s}`. Maybe this used to be allowed but isn't officially supported? Anyway, left justification is default anyway, so there's no point to the `-`. Removing it makes fmt happy, no more exceptions.


jessey-git (approved):
Seems good and looks like it was just these locations.

FWIW this seems to have changed starting with fmt 8.x. The "-" is only valid for floating point formats, not strings, so I'm guessing this started failing when fmt started doing more type validation.

lgritz:
That sounds like a good guess.

Yes, these were the only locations that used this construct, and also those lines were only executed when using "verbose" mode of maketx / IBA::make_texture(), so it's understandable why even though it's been problematic for a while, people rarely encountered it and didn't know why it was happening.

PR #4513, opened 2024-10-29
IBA::Unsharp_mask() speed and memory optimization

ssh4net:
## Description

Replacing 3x IBA + Helper function that generate 4 fulls size image buffers with single unsharp_mask_impl() that use parallel_image() to compute unsharp: 
src + contr * (((src - blur) < threshold) ? 0.0 : (src - blur))

Added two pass 1D convolution for a kernels higher than 3x3

## Tests

```
	ImageBuf sharped(input.spec());
	const int repeats = 50;

	std::cout << "Start sharpening\n";
	auto start = std::chrono::high_resolution_clock::now();

	for (int i = 0; i < repeats; i++)
	{
		//ok = ImageBufAlgo::unsharp_mask(sharped, input, "gaussian", 15.0f, 10.0f, 0.01f);
		ok = ImageBufAlgo::unsharp_mask(sharped, input, "gaussian", 5.0f, 2.0f, 0.05f);
		std::cout << ".";
	}

	std::cout << "\n";

	auto part1 = std::chrono::high_resolution_clock::now();
	std::chrono::duration<double> elapsed_part1 = part1 - start;
	std::cout << "Elapsed time: " << elapsed_part1.count() << " s\n";
```

both single threaded (one IB at time) and multithreaded (multiply IB at time) show pretty good speedup:
~30-40% with less memory use.

for 5x5 gaussian kernels two pass mode should add at least 20% speedup.

(if someone can do independent benchmark, will be great. As soon as I had a big differences on them depend on real or synthetic use)

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo.cpp
+++ b/src/libOpenImageIO/imagebufalgo.cpp
@@ -1008,7 +1008,7 @@ ImageBufAlgo::unsharp_mask(ImageBuf& dst, const ImageBuf& src,
             return false;
         }
     } else {
-        ImageBuf K = make_kernel(kernel, 1, width);
+        ImageBuf K = make_kernel(kernel, width, width);
         if (K.has_error()) {
             dst.errorfmt("{}", K.geterror());
             return false;
--- a/src/libOpenImageIO/imagebufalgo.cpp
+++ b/src/libOpenImageIO/imagebufalgo.cpp
@@ -971,15 +971,14 @@ unsharp_impl(ImageBuf& dst, const ImageBuf& blr, const ImageBuf& src,
 {
     OIIO_DASSERT(dst.spec().nchannels == src.spec().nchannels
                  && dst.spec().nchannels == blr.spec().nchannels);
-    // source + contrast * ((source - blurred) > threshold ? (source - blurred) : 0)
-    // -> source + (contrast * (source - blurred)) > (threshold * contrast) ? (contrast * (source - blurred) : 0)
+
     ImageBufAlgo::parallel_image(roi, nthreads, [&](ROI roi) {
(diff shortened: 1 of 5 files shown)


ssh4net:
@lgritz what to do?
There are 3 pixels difference between reference tahoe-unsharp.tif made with old single pass 5x5 convolution and new 2x 1D convolution:
this is amplified view of those pixels:
image

difference is 0.004 for one of channels.

I guess an issue happen due to float rounding errors. When convolution_() estimate a scale for normalization and use it later.
Already normalized kernel from kernel function already has 
1.000000075 sum
that give scale equal to 0.999999925 in double (in excel)
but c++ in float has it 0.999999881
and small error can a bit amplified in 2x 1D convolution.

As an alternative we can try to use double for kernels and normalization and cast to float before apply to a real pixels.

lgritz (approved):
LGTM

The changes resulted in 3 pixels having a single LSB difference versus the old code, which I think is totally understandable because you changed the code in a way that should be mathematically equivalent, but is a different sequence of floating point operations. I think this is acceptable, especially given the performance gain.

I took the liberty of updating the reference output and pushing that change to your tree, that should clear things up for the tests.

There are still some failed CI tests -- unrelated to your code -- in the heif test, which is the result of the GitHub runners upgrading the libheif version they are using. We've already fixed this in OIIO main, you branched off a little earlier than that for this fix, so as soon as we merge this PR, all will be ok.

I'll double check that CI passes in all cases *except* for the heif failures, and if that's the case, I will merge.

Sorry to take a while with this review, it's been a really busy last several days for me. But this is good work, thanks.

Issue #4406, opened 2024-09-04
[FEATURE REQUEST] Link to a source code from OpenImageIO documentation


ssh4net:
Too often, documentation is not enough to understand how some API functions are working.
For example, Colour-Science Python library usually provides a link to a source code where possible to check how it works under the hood. (for example https://colour.readthedocs.io/en/master/generated/colour.CHROMATIC_ADAPTATION_METHODS.html )

OpenImageIO can also win from a similar approach. Like in case when user wants to check if there is a additional optimisation in function code, or not all input parameters are documented.

lgritz:
That's interesting. Does anybody have a good idea about how to achieve this in a fairly automatic way for our documentation system?

lgritz:
Yes, please!

We build docs with Sphinx, using the Breathe plugin which takes Doxygen-generated XML that extracts API documentation from the C++ headers.

aarondemolder:
Came across this and saw this didn't happen, so made a PR - hope that's alright 😄

aarondemolder:
> If the docs are built with sphinx adding 'sphinx.ext.viewcode' to `extensions` in src/doc/conf.py should do the trick? Shall I make a pull request?

Unfortunately it wasn't quite so simple and will need some more attention - I found some more discussion on similar here: https://github.com/breathe-doc/breathe/issues/388

lgritz:
This has been stale for a while, but does anybody want to pick it up and prototype it to see if we like the result?

PR #4515, opened 2024-10-29
Add Sphinx Viewcode to docs

aarondemolder:
## Description

One-line change that addresses the request in #4406 to add links to source code from the docs as per https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4406#issuecomment-2331959930

Will confirm that the change is correctly built by checks in a comment below!

## Tests

n/a

## Checklist:

- [X] I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/conf.py
+++ b/src/doc/conf.py
@@ -68,6 +68,7 @@ sys.path.insert(0, os.path.join(os.path.abspath(os.path.dirname(__file__)), '../
 #import latex_mods
 
 extensions = [
+              'sphinx.ext.viewcode',
               'breathe',
               'sphinx_tabs.tabs'
  ]


lgritz:
Is this implementing the links to code, or is it just adding the sphinx module that will allow it if we make some other changes later?

Where exactly do you see the links to the code appearing? I'm not seeing it in the docs generated. Am I just looking in the wrong place?

aarondemolder:
I was also disappointed to discover that indeed it was not as easy as adding 'sphinx.ext.viewcode' to extensions in src/doc/conf.py - the docs appear to be identical. It seems this works well in python only...

https://github.com/breathe-doc/breathe/issues/388 is also a dead end

I can do some more reading and update anything in the original issue, but it doesn't appear to be a simple change.

PR #4516, opened 2024-10-29
[fix] raw: fix channel layout

antond-weta:
## Description

The function libraw_filter_to_str() converts the `imgdata.idata.filters` field to human readable form, by mapping a few hardcoded values. I found the issue by converting images with disabled demosaicing, which writes the Bayer pattern into the ImageBuf metadata. The Panasonic GX9 didn't have any pattern recognised; The Canon R5 was incorrectly recognised as GMCY.

The `imgdata.idata.filters` actually contains a sequence of 2-bit values, where each represents an index in a 4-character long string `imgdata.idata.cdesc`. I have modified the code to extract the channels from `imgdata.idata.cdesc` as expected.

I have also disabled the normalisation code, which I believe was incorrect, and never executed due to the issue above.

## Tests

The current tests pass, however, I'll start a discussion on the piece of code I have concerns with.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -248,7 +248,7 @@ namespace {
 std::string
 libraw_filter_to_str(unsigned int filters, const char* cdesc)
 {
-    char result[5] = { 0 };
+    char result[5] = { 0, 0, 0, 0, 0 };
     for (size_t i = 0; i < 4; i++) {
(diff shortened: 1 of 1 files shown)


lgritz (commented):
LGTM pending one minor comment I made asking for clarification about the array initialization.

lgritz (approved):
LGTM

PR #4517, opened 2024-10-30
build: support static OCIO self-builds

zachlewis:
* Add build recipes for certain dependencies needed for linking static OpenColorIO libs: expat, minizip-ng, pystring, and yaml-cpp

 * Add SOURCE_SUBDIR option to `build_dependency_with_cmake` that mimics the behavior of the FetchContent option of the same name 
 
 * Tweak OpenColorIO build recipe to always build static libraries by default.


lgritz:
Problem on Windows?

zachlewis:
Problems on Windows, a tale older than time itself. 

Just a little trouble with finding Expat's config. I haven't looked into this too carefully yet -- I'll have another go today.

zachlewis:
Nice, putting some distance between building pystring and REQUIRED-finding it seems to more reliably pass all CI checks.

lgritz (approved):
LGTM and seems to pass all tests now!

PR #4518, opened 2024-11-01
perf: oiiotool --line --text --point --box speedups

lgritz:
These commands, which are intended to alter the top image on the stack, were each allocating and copying an image unnecessarily for each operation. By changing the implementation to the right calls that know it doesn't need to make a copy (already a feature I'd implemented in the OiiotoolOp helper class) greatly speeds them up.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5016,7 +5016,8 @@ OIIOTOOL_INPLACE_OP(point, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
 
 // --text
 OIIOTOOL_INPLACE_OP(text, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
-    img[0]->copy(*img[1]);
+    OIIO_ASSERT(img[0] == img[1]);  // Assume we're operating in-place
+    // img[0]->copy(*img[1]);  // what we'd do if we weren't in-place
     const ImageSpec& Rspec(img[0]->spec());
     int x            = op.options().get_int("x", Rspec.x + Rspec.width / 2);
     int y            = op.options().get_int("y", Rspec.y + Rspec.height / 2);
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -76,6 +76,18 @@ using pvt::print_info_options;
         op();                                                        \
     }
 
+// Lke OIIOTOOL_OP, but designate the op as "inplace" -- which means it
+// uses the input image itself as the destination.
+#define OIIOTOOL_INPLACE_OP(name, ninputs, ...)                      \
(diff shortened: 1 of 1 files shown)


lgritz:
Comments?

lgritz:
Can somebody review this? I'm trying to get out of the habit of merging my own patches without positive proof that somebody else's eyes have been on it.

lgritz:
I'm less worried about the testing -- the CI ought to vouch for the correctness of the result -- but more just looking over the code to see if it seems reasonable. I just don't like the optics of merging my own PR without so much as a single comment from anybody else. We do it too often, but it's something that any healthy project should only do in dire emergencies.

src/oiiotool/oiiotool.cpp
```diff
@@ -4940,7 +4952,7 @@ OIIOTOOL_OP(contrast, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
 
 // --box
 // clang-format off
-OIIOTOOL_OP(box, 1, nullptr, ([&](OiiotoolOp& op, span<ImageBuf*> img) {
+OIIOTOOL_INPLACE_OP(box, 1, nullptr, ([&](OiiotoolOp& op, span<ImageBuf*> img) {
     img[0]->copy(*img[1]);
```
mugulmd, line 4956: Just checking: since the Op now runs in-place, do we still need this copy ?

lgritz: Internal to copy, it checks if it's doing a self-assignment and takes an early out. So there is not actually a needless copy here. 

But that's a good point, maybe for clarity it's better to remove it. Let me get rid of it and run a test to be sure we're correct that it can go.

lgritz: Yes, you were right, I'm able to eliminate that line, as it is not currently necessary. Will push an update.

src/oiiotool/oiiotool.cpp
```diff
@@ -5000,7 +5015,7 @@ OIIOTOOL_OP(point, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
 
 
 // --text
-OIIOTOOL_OP(text, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
+OIIOTOOL_INPLACE_OP(text, 1, [&](OiiotoolOp& op, span<ImageBuf*> img) {
     img[0]->copy(*img[1]);
```
mugulmd, line 5019: I think you forgot to remove the copy here (like you've done in the 3 ops above)

mugulmd: (for consistency)

lgritz: you're right, I will fix

mugulmd (approved):
LGTM 🙂

PR #4522, opened 2024-11-03
docs: Clarify 'copy_image' example

lgritz:
It was pointed out on Slack that this example was unclear as to what should be passed to open().


lgritz:
Can somebody approve this? I'm trying to get out of the habit of merging my own patches without positive proof that somebody else's eyes have been on it.

mugulmd (approved):
LGTM :)

PR #4523, opened 2024-11-04
api: Add API to query available font families and styles

### LINKED ISSUE #4503: [FEATURE REQUEST] API to query available font families and styles
In Arnold we use OIIO to render text over our rendered images. We have a user interface where end users can configure text render, such as specify the font, font size, alignment, etc. They can select the font from a combo box that lists available font familes (e.g. "Arial", "times New Roman", etc.) and styles (e.g. "Regular", "Bold", "Italic", etc).

image

Therefore we extended the API to return available font families and styles per family. If you think this would be interesting to have, we can port our code back.

peterhorvath111:
## Description

Fixes #4503.

Add an API to query available fonts based on family and style name. The API can be used to build a UI where users can select a font from a list.

- **font_families()**: Return the list of available font families in alphabetical order.
- **font_styles(family)**: Return the list of available font styles of a given font family.
- **font_filename(family, style)**: Return the font filename that defines the font with the given family and style.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -386,6 +386,37 @@ ROI OIIO_API text_size (string_view text, int fontsize=16,
                         string_view fontname="");
 
 
+/// Return the list of available font families in alphabetical order,
+/// such as "Arial", "Times New Roman", etc.
+/// Fonts are read from system font folders, folders defined by
+/// the global `font_searchpath` attribute or the `OPENIMAGEIO_FONTS`
+/// or `OpenImageIO_ROOT` environment variables.
+const std::vector<std::string>& OIIO_API font_families();
+
+
+/// Return the list of available font styles of a given font family,
+/// such as "Regular", "Bold", "Italic", etc.
+///
+/// @param  family
+///             Font family name.
+///
+const std::vector<std::string> OIIO_API font_styles(string_view family);
+
+
+/// Return the font filename that defines the font with the given 
+/// family and style, or an empty string if the font does not exist.
+///
+/// @param  family
+///             Font family name.
+///
+/// @param  style
+///             Font style name. If not given that means the 'Regular'
+///             style (if exists).
+///
+const std::string OIIO_API font_filename(string_view family, 
+                                         string_view style = "");
+
+
 /// Generic channel shuffling: return (or store in `dst`) a copy of `src`,
 /// but with channels in the order `channelorder[0..nchannels-1]` (or set to
 /// a constant value, designated by `channelorder[i] = -1` and having the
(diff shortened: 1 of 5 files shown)


src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -1597,9 +1628,7 @@ bool OIIO_API convolve (ImageBuf &dst, const ImageBuf &src, const ImageBuf &kern
 /// Laplacian kernel,
 ///
 ///                     [ 0  1  0 ]
-///                     [ 1 -4  1 ]
 ///                     [ 0  1  0 ]
-///
 ImageBuf OIIO_API laplacian (const ImageBuf &src, ROI roi={}, int nthreads=0);
```
lgritz, line 1632: Whoa, what's up with this?

lgritz:
Are you aware of the existing `OIIO::getattribute()` queries related to fonts?

```
/// - `string font_list`
/// - `string font_file_list`
/// - `string font_dir_list`
///
///   A semicolon-separated list of, respectively, all the fonts that
///   OpenImageIO can find, all the font files that OpenImageIO can find (with
///   full paths), and all the directories that OpenImageIO will search for
///   fonts.  (Added in OpenImageIO 2.5)
```

Historically, we've tended to avoid adding new single-use API calls for individual queries, instead preferring to just add new tokens that getattribute recognizes to cover as many cases as possible.

Your `font_families()` and `font_styles()` are trivially amenable to this approach, since they just return a list of strings. The `font_filename()` is a little trickier, but we could probably do it by accepting special query names composed of `"font_filename:FAMILY:STYLE"` and parsing out the family and style name. Internally, they would still be implemented by more or less the functions you implemented but they would be turned into something in the pvt namespace and hidden from public headers.

I'm not saying I'm necessarily set against doing it as functions as you have... but I wanted to give you the chance to consider sticking with the idiom we've been using, which I think has been successful in keeping ABI churn to a minimum. For functionality that's never really in a performance critical loop, it's really easy to add, remove, backport, or change the name or meaning of these tokens without any hard compatibility breaks, compared to the approach where a different set of exposed API functions is required.

src/include/OpenImageIO/imagebufalgo.h
peterhorvath111: Oops, my bad, I don't know what happened there. Fixed now.

peterhorvath111:
I don't mind switching to `getattribute()` queries, if that's the preferred approach.

peterhorvath111:
API functions are converted to the following `getattribute()` queries:

- **font_family_list**: Return a semicolon-separated list of all available font families.
- **font_style_list:family**: Return a semicolon-separated list of all available font styles of a given font family.
- **font_filename:family:style**: Return the font filename that defines the font of the given family and style.

lgritz (approved):
I like! Thanks.

lgritz:
Because 3.0.0 is already marked as a release candidate, my goal is only to merge critical bug fixes for the final non-rc tag on Friday. So I'll merge this into main today, then backport it into the dev-3.0 branch after tagging on Friday, so technically this will be released as part of the next (3.0.1) patch release.

If it's really important to you to get this in Friday's first 3.0.0 final release, then I can do it at essentially no real risk (since as far as public API, all this does is add some new getattribute query tokens and changes no other behavior, which was the whole point!). But if you can wait for 3.0.1 or operate from main for now, then I think it's better release policy to not add anything unnecessary once an RC is declared, no matter how harmless it appears.

ThiagoIze:
Waiting for 3.0.1 is perfectly fine for us.

PR #4525, opened 2024-11-04
deps: Add new ref output for libheif updates

lgritz:
The latest libheif patch release changed something very slightly that requires us to update some saved reference output.


lgritz:
Comments?

lgritz:
Because this fixes a hot CI break that will cause all other PRs to spuriously appear to fail, I'm going to reluctantly merge this immediately without review. But of course you all know that this is something we'd like to avoid in all but emergency situations.

PR #4526, opened 2024-11-05
ci: increased the macos timeout slightly to fix spurious failures

lgritz:
Slow Mac runners sometimes take too long and fail needlessly.


lgritz:
This PR has been posted for 8 days without comment. I will pre-emptively merge now, since it only affects CI and doesn't alter any code, and it's causing other PR's CI to spuriously fail.

PR #4528, opened 2024-11-08
fix(exr): allow an empty "name" metadata to be read

lgritz:
When the exr "name" metadata (used for the names of subimages/parts) existed but was the empty string, we were treating it like it didn't exist (and upon write substituted our own "subimageXY" name, as we are wont to do for unnamed parts). But now we've seen multi-part exrs where the name is "" (exists, but empty string) and we want a read-write cycle to preserve that. Even though it's not a very sensible or helpful part name, it's technically legal, so we should honor it.

### DIFF
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -652,7 +652,7 @@ OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in,
     }
 
     // EXR "name" also gets passed along as "oiio:subimagename".
-    if (header->hasName() && header->name() != "")
+    if (header->hasName())
         spec.attribute("oiio:subimagename", header->name());
 
     spec.attribute("oiio:subimages", in->m_nsubimages);


lgritz:
@meimchu

lgritz:
Does anybody object to this? In short, the change allows multi-part exr files to have a part whose name is the empty string "", which is technically allowed by OpenEXR itself. Our old behavior was to treat an empty string part name as if there was no name at all (in which case we substituted something else).

lgritz:
N.B. I always say "does anybody object" but what I really mean is "somebody PLEASE review and approve (or critique) this, because it's bad open source practice to merge PRs with no approval by somebody who isn't the author of the patch and I don't want to do that except in cases of real emergencies."

JGoldstone:
For what it's worth, in the OpenEXR subset documented in SMPTE's ACES Container (ST 2065-4:2023 ACES Image Container File Layout) zero-length attributes are everywhere -- unless we missed a spot -- disallowed. Point of information only. As for the change itself, LGTM

PR #4531, opened 2024-11-09
doc: Quickstart guide

lgritz:
Having such a document is one of the remaining steps to getting our OpenSSF Best Practices silver level badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,11 @@ is the best place to start if you are interested in how to use OpenImageIO,
 its APIs, its component programs (once they are built). There is also a [PDF
 version](https://readthedocs.org/projects/openimageio/downloads/pdf/latest/).
 
+Additional resources:
+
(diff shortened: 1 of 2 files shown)


lgritz:
This has been posted for a week without receiving any comments. Does anybody have feedback? It's meant to be an extremely brief "quickstart" that just shows how to do a simple image operation through each of the several interfaces and bindings of the project.

ThiagoIze (commented):
Does the quickstart guide also need to explain how to build OIIO or is it just about using a tool? Could we add a link for where to go for more info?

lgritz:
We already have the INSTALL.md (how good that is, perhaps, is debatable), so I was assuming this was the "use it" quickstart. But I should cross-link those documents.

ThiagoIze (approved):

PR #4532, opened 2024-11-10
docs: code review guidelines and tips

lgritz:


docs/dev/CodeReview.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Submitter guidelines -- how to encourage good reviews and timely merges
+
+A few tips for submitters to help ensure that your PRs get reviewed and merged
+in a timely manner and are respectful of the reviewers' time and effort:
+
+- Aim for small PRs that do a specific thing in a clear way. It is better to
+  implement a large change with a series of PRs that each take a small,
+  obvious step forward, than to have one huge PR that makes changes so
+  extensive that nobody quite knows how to evaluate it.
```
dgovil, line 61: Might be worth mentioning that PRs should link to any other PRs that need to land first?

docs/dev/CodeReview.md
```diff
@@ -0,0 +1,177 @@
...
+- Make sure your PR commit summary message clearly explains the why and how of
+  your changes. You want someone to read that and judge whether it's a
+  sensible approach, know what to look for in the code, and have their
+  examination of the code make them say "yes, I see, of course, great, just as
+  I expected!"
+- Make sure your PR includes a test (if not covered by existing tests).
+- Make sure your PR fully passes our CI tests.
```
dgovil, line 68: Might also be good to suggest it passes any linters that your project requires too?

docs/dev/CodeReview.md
```diff
@@ -0,0 +1,177 @@
...
+  GitHub, or by adding a comment that says "LGTM" (looks good to me).
+- If possible, reviewers should have some familiarity with the code being
+  changed (though for a large code base, this is not always possible).
+- At least a few work days should elapse between posting the PR and merging
+  it, even if an approving review is received immediately. This is to allow
+  additional reviewers or dissenting voices to get a chance to weigh in.
+
+Not all patches need the highest level of scrutiny. Reviews can be cursory and
+quick, and merges performed without additional delay, in some common low-risk
+circumstances:
```
dgovil, line 94: I might also suggest that in these scenarios , it's worth doing the following:

- leave a message on the PR as to why you're merging without review. 

- broadcast that you did so in the community (slack)

Those just allow someone to historically review things without blocking it.

docs/dev/CodeReview.md
```diff
@@ -0,0 +1,177 @@
...
+These are anti-patterns that generally are indications of an unhealthy open
+source project:
+
+* An author merging their own PR without any review comments that indicate
+  meaningful scrutiny or approval from another party, and without giving
+  adequate warning that a merge will occur if no reviews are received.
+* PRs languishing for weeks or months without any review or merge.
+* PRs that don't adequately test their changes (basically crossing your
+  fingers and hoping it works).
+* PRs that are merged despite failing CI that might be related or, if
```
dgovil, line 172: One more that I might suggest adding is something along the lines of avoid unilateral rejection of ideas. 

Obviously this would be discretionary as there are some ideas that would never pass muster (e.g rewrite the project in a new language) or the submitter cannot defend their viewpoint beyond basic scrutiny. 

However,  in the case where the submitter is able/willing to defend their suggestion, it would be recommended to bring in a second reviewer or bring it to the TSC.

dgovil (approved):
Left a few minor thoughts based on experiences with internal projects , but overall I really like this. I think it would be cool to have as an example recommendation for all ASWF projects. 

Thanks for putting this together!

docs/dev/CodeReview.md
```diff
@@ -0,0 +1,177 @@
...
+**EVERY** pull request:
+
+- At least one reviewer critiques and eventually (possibly after changes are
+  made) approves the pull request. Reviewers, by definition, are not the same
+  person as the author.
+- Reviewers should indicate their approval by clicking the "Approve" button in
+  GitHub, or by adding a comment that says "LGTM" (looks good to me).
+- If possible, reviewers should have some familiarity with the code being
+  changed (though for a large code base, this is not always possible).
+- At least a few work days should elapse between posting the PR and merging
```
dgovil, line 88: Might also be worth adding something like "submitters should be given some time (a day?) to review changes made to their PR by a reviewer, in case they have useful context/thoughts/objections to add"

docs/dev/CodeReview.md
lgritz: I was more thinking of the case of the same person having subsequent PRs queued up but not yet submitted, waiting for earlier work to be accepted.

docs/dev/CodeReview.md
lgritz: CI doesn't pass unless all that stuff works, so it's redundant in this project. Even not having correct formatting fails a CI test.

docs/dev/CodeReview.md
lgritz: Yes, good suggestion, thanks.

docs/dev/CodeReview.md
dgovil: Ah that's fair. I've done it both ways when trying to have dependent features submitted.

docs/dev/CodeReview.md
lgritz: In general, if I make suggestions in the form of an explicit diff, I expect authors to be the one to click the "accept" button. It feels rude to push something to their branch, unless they ask me to or it's clearly the only way to move the PR forward. But yeah, I agree that even if the reviewer has to push a change, they should give the author a chance to okay it before a merge in almost all circumstances. I will say something to that effect.

docs/dev/CodeReview.md
lgritz: Thanks, I will amend. A "rejection" should always be one step in a conversation, with the next step being the author either fixing the patch, trying to explain/justify it as-is, or realizing themselves that a different approach is needed and closing their own PR. I don't believe a reviewer should ever *close* a PR, unless it's clearly been abandoned by the author, or unless it utterly fails to conform to the minimum standards of any submission (i.e. based on form, not simply disagreeing with the content).

lgritz:
Thanks for the thoughtful comments, @dgovil. I pushed an update with some amendments. I think this is a PR that I'll leave up for quite a while, as I'd love to have plenty of time for people to digest this and make additional suggestions.

lgritz:
If there are no further comments forthcoming, I will merge this over the weekend (which will have given it a full 2 weeks of comment period). We can always continue to modify both the document and the policy goals over time if needed.

PR #4535, opened 2024-11-12
Added support for negative `nthreads` and resp. documentation.

virtualritz:
## Description

This makes `nthreads` accept negative numbers to mean `m_pool->size()` + `nthreads`. I.e. on a `16` core machine specifying `nthreads` as `-2` would make OIIO use `14` cores.
If `m_pool->size()` < `-nthreads` the result is clamped at `1`. I.e. a single thread would be used.

On that note: the original `resolve()` does not have a check for the case where `m_maxthreads` > `m_pool->size()` and should possibly be clamped to the latter, if so. I kept it that way. Is that intended/a feature or an oversight?

## Checklist:

  nonconforming.

### DIFF
--- a/src/doc/imagebufalgo.rst
+++ b/src/doc/imagebufalgo.rst
@@ -143,6 +143,14 @@ that the number of thread should be the OIIO global default set by
 itself defaults to be the detected level of hardware concurrency (number
 of cores available).
 
+Specifying a negative number for `nthreads` will reserve `-nthreads`
+threads. In this case the actual number of threads used is the number of
+cores + `nthreads` + 1.  One thread is always reserved for the caller.
+For example, on a 32 core system specifying `nthreads` as -2 means the
+OpenImageIO will use 31 cores: 32 + `nthreads` + 1. If `-nthreads` is
+specified larger than the number of available cores only a single core will
+be used.
+
 Generally you can ignore this parameter (or pass 0), meaning to use all
 the cores available in order to perform the computation as quickly as
 possible.  The main reason to explicitly pass a different number
@@ -3242,4 +3250,3 @@ General functions that also work for deep images
     regardless of resolution), for each pixel merely copying the closest
     deep pixel of the source image (no true interpolation is done for deep
     images).
-
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -129,6 +129,14 @@ class Filter2D;
 /// `OIIO::attribute()`, which itself defaults to be the detected level of
 /// hardware concurrency (number of cores available).
 ///
+/// Specifying a negative number for `nthreads` will reserve `-nthreads`
+/// threads. In this case the actual number of threads used is the number of
+/// cores + `nthreads` + 1.  One thread is always reserved for the caller.
+/// For example, on a 32 core system specifying `nthreads` as -2 means the
+/// OpenImageIO will use 31 cores: 32 + `nthreads` + 1. If `-nthreads` is
+/// specified larger than the number of available cores only a single core will
+/// be used.
+///
 /// Generally you can ignore this parameter (or pass 0), meaning to use all
 /// the cores available in order to perform the computation as quickly as
 /// possible.  The main reason to explicitly pass a different number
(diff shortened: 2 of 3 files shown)


ssh4net:
Why this can be better than 

`std::thread::hardware_concurrency() - 2`
?

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -680,14 +688,14 @@ bool OIIO_API resize(ImageBuf &dst, const ImageBuf &src, KWArgs options = {},
 /// corresponding portion of `src` (mapping such that the "full" image
 /// window of each correspond to each other, regardless of resolution).  If
 /// `dst` is not yet initialized, it will be sized according to `roi`.
-/// 
+///
```
lgritz, line 691: Can you please remove all spurious formatting changes that are not directly related to the PR? Thanks.

lgritz:
> Why this can be better than
> 
> `std::thread::hardware_concurrency() - 2` ?

1. Because it means 2 less than the pool size, not the HW threading size.
2. Because it can get passed from all sorts of places where it's not convenient to call C++ std::thread_hardware_concurrency(). Like Python? Or from command line args?

src/include/OpenImageIO/imagebufalgo.h
virtualritz: Yeah, sorry about that. It seems removing trailing white space on save is a default feature in modern IDEs.
I'll get to it sthe next week.
That said: wouldn't it make sense to remove trailing white space in the entire code base to save anyone from going though this on a PR?
I guess I'm asking if you'd consider a separate PR for just that?

lgritz: Yes, certainly. I just want code changes to not have formatting changes to any code that's not changing. A PR that is nothing but formatting changes is fine as long as no code changes. (Among other things, that allows us to add the formatting-only commit hashes to `.git-blame-ignore-revs` so that they don't change the apparent history or authorship of any actual code.)

src/doc/imagebufalgo.rst
```diff
@@ -143,6 +143,14 @@ that the number of thread should be the OIIO global default set by
 itself defaults to be the detected level of hardware concurrency (number
 of cores available).
 
+Specifying a negative number for `nthreads` will reserve `-nthreads`
+threads. In this case the actual number of threads used is the number of
+cores + `nthreads` + 1.  One thread is always reserved for the caller.
+For example, on a 32 core system specifying `nthreads` as -2 means the
+OpenImageIO will use 31 cores: 32 + `nthreads` + 1. If `-nthreads` is
+specified larger than the number of available cores only a single core will
+be used.
```
lgritz, line 152: Just to clarify, because I myself was not sure: Does `-2` mean "num_hardware_cores - 2" or does it mean "thread_pool_size - 2"?

src/libutil/thread.cpp
```diff
@@ -595,10 +595,15 @@ paropt::resolve()
...
         m_pool = default_thread_pool();
-    if (m_maxthreads <= 0)
-        m_maxthreads = m_pool->size() + 1;  // pool size + caller
-    if (!m_recursive && m_pool->is_worker())
+
+    if (!m_recursive && m_pool->is_worker()) {
         m_maxthreads = 1;
+    } else {
+        if (m_maxthreads < 0)
+            m_maxthreads = fmax(m_pool->size() + m_maxthreads + 1, 1);
```
lgritz, line 603: This seems to do "pool_size - n"

lgritz:
I think that your changes only alter the meaning of the nthreads parameter to IBA functions. It is worth noting that there are a few other places where number of threads are passed:

* The setting of the thread pool size with `OIIO::attribute("threads", n)`
* ImageInput/ImageOutput::threads()
* ImageBuf::threads()
* Utility functions such as `parallel_convert_image()`
* The `paropt` structure used by parallel loops and such in parallel.h

Do we want to change those also? If we don't, will this create confusion?

Do those other places misbehave if passed negative thread counts? That's something we may never have considered before because... why would anybody do that? But now, we're giving people a reason to pass negative thread counts, at least in some places, which means we should expect people to do it inadvertently it in any of the other places.

PR #4536, opened 2024-11-12
fix(exr): allow an empty "name" metadata to be read

lgritz:
Continuation of PR #4528. I forgot to change the spot in the OpenEXR "core" API as well.


EmilDohne:
LGTM! (commenting since I saw you wanted to move away from no reviews)

lgritz:
Thank you!

Yes, please, it was never our intention to merge review-less PRs, outside of true necessity, but I'm afraid the habit started to grow over time. But now we're trying to re-emphasize that it is undesirable. At the very least, having a second pair of eyes doing even a very cursory look before a merge ensures that no one person (including one trusted person whose credentials are stolen) can easily sneak problems into the repo -- something has actually happened to cause some high-profile debacles in other projects.

EmilDohne:
Glad to help! Even if it's not through direct commits :)

PR #4541, opened 2024-11-25
feat(IBA): IBA::scale()

antond-weta:
## Description

Implements IBA::scale, an algo which takes two images, one of which is required to be a single channel image, each channel of the other image gets multiplied by that single channel, per pixel.

Note that this is subtly different from IBA::mul(), which is a channel-by-channel multiplication, and so doesn't have a provision for allowing you to scale all channels of one image by the single channel of a second image.

## Tests

I have added tests for the C++ code, python code, oiiotool, and the documentation examples

## Checklist:

  doesn't change the API.)
  nonconforming.


src/libOpenImageIO/imagebufalgo_muldiv.cpp
```diff
@@ -91,6 +149,8 @@ ImageBufAlgo::mul(ImageBuf& dst, Image_or_Const A_, Image_or_Const B_, ROI roi,
     pvt::LoggedTimer logtime("IBA::mul");
     if (A_.is_img() && B_.is_img()) {
         const ImageBuf &A(A_.img()), &B(B_.img());
+        if (A.nchannels() == 1 || B.nchannels() == 1)
+            return scale(dst, A, B, roi, nthreads);
```
lgritz, line 153: This is a behaviour changes for mul(). Maybe that's fine, but we should certainly document it, and maybe think hard about if we want it.

Are we prepared to say that any time somebody multiplies an n-channel image by a 1-channel image, they mean to scale all channels by that one value? Is there ever a case where people would not want that? (I honestly don't know.)

antond-weta: I does look weird when the alpha is scaled, as you mentioned, but on the other hand this behaviour is consistent with mul() used with a single scalar.

antond-weta:
Yes, that was the idea. I guess, we could introduce an option to treat the alpha channel differently, if needed. I'm not sure if that would be more convenient than using the ROI.

src/libOpenImageIO/imagebufalgo_muldiv.cpp
lgritz: I think that's fine.

I wonder... if maybe we should get in the habit of adding a KWArgs parameter to all new IBA functions (and, slowly, adding them to exiting ones when the occasion pops up that we have to modify them for other reasons), even if we don't have any keyword args to respond to at this moment, as a placeholder for future expansion without needing to alter the ABIs or add more functions. In this case, that would make it more palatable to pick a reasonable default behavior (just scale all channels) for now and then in the future, we can very easily accommodate an option that would cause the alpha to not be scaled.

lgritz: Let's not add it to mul (which would either break compatibility, or cause yet another set of mul versions to be added) until we have a desired use.

But let's add it to scale() now -- it costs nothing and has no compatibility issues for a "new" IBA function, and future-proofs the API.

antond-weta: I've added a KWArgs parameter to scale(), and removed the changes from mix() for now.

antond-weta:
This should be ready for review now

src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -953,6 +953,17 @@ ImageBuf OIIO_API abs (const ImageBuf &A, ROI roi={}, int nthreads=0);
 bool OIIO_API abs (ImageBuf &dst, const ImageBuf &A, ROI roi={}, int nthreads=0);
 
 
+/// Compute per-pixel product `A * B`, returning the result image.
+///
+/// All channels of one of the images get multiplied by the value in the only channel of the other image.
+/// Either `A` or `B` must be a single channel image.
```
lgritz, line 959:
```suggestion
/// Compute per-pixel product `A * B`, returning the result image. At least
/// one of `A` and `B` must be a single channel image, whose value is used to
/// scale all channels of the other image.
///
/// @param  options
///         Optional ParamValue's that may control the reconstruction.
///         (Reserved for future expansion.)
///
```

lgritz (changes requested):
Code and design LGTM.

Very minor documentation changes suggested.

src/include/OpenImageIO/imagebufalgo.h
antond-weta: sorry, forgot about that

lgritz (approved):
LGTM, thanks

PR #4542, opened 2024-11-25
fix(exr): Avoid integer overflow for large deep exr slice strides

### LINKED ISSUE #4540: [BUG] Writing a very large deep EXR scanlines crashes
**Describe the bug**

Writing a large deep scanline EXR crashes.

OIIO 2.5.17.0 | Linux/x86_64
    Build compiler: gcc 11.4 | C++17/201703
Dependencies:  , Boost 1.74.0, BZip2 NONE, DCMTK NONE, FFmpeg NONE, Freetype 2.11.1, GIF NONE, JPEG 80, Libheif
    NONE, libjpeg-turbo NONE, LibRaw NONE, OpenColorIO NONE, OpenCV NONE, OpenEXR 3.3.2, OpenGL, OpenJPEG NONE, PNG
    1.6.37, Ptex NONE, Ptex NONE, pybind11 2.9.1, Python 3.10.12, Qt5 NONE, Qt6 NONE, Robinmap, TBB NONE, TIFF 4.3.0,
    WebP NONE, ZLIB 1.2.11

It was also reproducible in older version: OIIO 2.3.14 using OpenEXR 3.2 and also 2.2.18.

**To Reproduce**

1. Extract the attached python script
2. Run (eg 'python3 dcmbug.py')
3. Wait until it crashes

[dcmbug.zip](https://github.com/user-attachments/files/17874000/dcmbug.zip)

The exact same crash can be observed when doing the same via C++ API also. If you reduce the image resolution significantly (or reduce the number of channels significantly) then it no longer crashes.
I'm reasonably certain it's OIIO-related issue and not OpenEXR because the similar test in OpenEXR does *not* crash.

lgritz:
Fixes #4540

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -1729,15 +1729,15 @@ OpenEXROutput::write_deep_scanlines(int ybegin, int yend, int /*z*/,
         frameBuffer.insertSampleCountSlice(countslice);
         std::vector<void*> pointerbuf;
         dd->get_pointers(pointerbuf);
+        size_t slchans      = size_t(m_spec.width) * size_t(nchans);
+        size_t xstride      = sizeof(void*) * size_t(nchans);
+        size_t ystride      = sizeof(void*) * slchans;
+        size_t samplestride = dd->samplesize();
         for (int c = 0; c < nchans; ++c) {
             Imf::DeepSlice slice(m_pixeltype[c],
                                  (char*)(&pointerbuf[c] - m_spec.x * nchans
-                                         - ybegin * m_spec.width * nchans),
-                                 sizeof(void*)
-                                     * nchans,  // xstride of pointer array
-                                 sizeof(void*) * nchans
-                                     * m_spec.width,  // ystride of pointer array
-                                 dd->samplesize());   // stride of data sample
+                                         - ybegin * slchans),
+                                 xstride, ystride, samplestride);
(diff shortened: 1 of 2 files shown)


lgritz:
Original reporter of #4540 reported there that they applied this patch on their end and it fixes the problem.

src/openexr.imageio/exroutput.cpp
```diff
@@ -1729,15 +1729,15 @@ OpenEXROutput::write_deep_scanlines(int ybegin, int yend, int /*z*/,
         frameBuffer.insertSampleCountSlice(countslice);
         std::vector<void*> pointerbuf;
         dd->get_pointers(pointerbuf);
+        size_t slchans      = size_t(m_spec.width) * size_t(nchans);
```
jessey-git, line 1732: The only thing that caused me to do a double-take in this PR is that this line doesn't mirror the form used in exrinput. Can we declare nchans as size_t above (assuming m_spec.nchannels is already known to be >= 0) so that this entire block of code will match what is used in exrinput? i.e. there's 3 casts to size_t in this block whereas the ones in exrinput has none.

lgritz: Yeah, that makes sense

lgritz: So updated (delay was for fixing the great GHA failure of 2024).

jessey-git (approved):

PR #4543, opened 2024-11-28
ci: Fix broken CI for ASWF 2021 and 2022 containers

lgritz:
The background here takes some explaining...

The ASWF's prepared Docker containers that reproduce a CentOS7-based environment corresponding to VFX Platform 2021 and 2022 years contain a glibc that is too old to run a newer version of "node" -- which GitHub Actions themselves use. The old one is no longer receiving support, and GitHub finally changed the policy that was allowing its GHA runners to still run the old version that worked with those ASWF containers.  The new node version that is happy on the GHA runners requires a glibc newer than the one in the containers. So that breaks all our CI runs in 2021 and 2022 VFX Platform configurations.

See this thread, if you dare:
https://academysoftwarefdn.slack.com/archives/C0169RX7MMK/p1732574400981949

Anyway, Jean-Francois Panisset proposed and tested a fix, which lets us continue working (for now) by downloading an old node implemenation (which is happy with the old glibc) and installing it atop the conainers, so it's this old one that the GH actions will use. It feels like it's all held together by chewing gum and duct tape, but it unbreaks our CI and hopefully it will keep working until all the studios have switched fully to RHEL/Alma/Rocky 9.x and we no longer care about testing on CentOS-7 based containers or supporting the VFX Platform years that are too old to be relevant to the new OS that the studios are using.


lgritz:
Simplified a bit

lgritz:
Because this addresses a major CI breakage that prevents any other PRs from correctly assessing whether they build and run properly in all environments, it's an emergency and I will merge this in the morning even if there are no reviews.

PR #4545, opened 2024-11-30
ci: refactor using a single steps workflow

lgritz:
The logic for the GHA steps was repeated across multiple workflows and platforms. Consolidate it all into a common "ci-steps" workflow, and have all the other workflows reference that when possible. This should hopefully make maintenance of the CI scripts easier in the future, with less redundancy of the step logic.

The only major place where we deviate is that we have a separeately specified set of steps for the "old" VFX Platform 2021 and 2022 based tests (because we need to specify different container arguments to it).

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -16,10 +16,14 @@ on:
       - main
       - '*analysis*'
       - '*sonar*'
-    paths-ignore:
-      - '**.md'
-      - '**.rst'
-      - 'docs/**'
+    paths:
+      - '**'
+      - '!**.md'
+      - '!**.rst'
+      - '!**/analysis.yml'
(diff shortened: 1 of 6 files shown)


lgritz:
Any opinions on this?

Again what's happening here is that we're merging all the (mostly redundant) step lists of the various workflows into a single step list, then referencing it from the other places. Flags are used to skip the steps for workflows where they are not needed (for example, the "build docs" workflow skips actually building or testing the code). Mostly, it's just eliminating redundancy and the possibility of the different workflows diverging and doing different things even for steps that we intend to be the same across all of them workflows.

lgritz:
This has been sitting around for a while without any comment. I will merge it by the end of the week if there are no objections.

PR #4548, opened 2024-12-01
jpeg2000: valid_file implementation, much faster than trying to open

aras-p:
## Description

Primary cost of detecting whether a given file is valid jpeg2000 file is the thread pool initialization and shutdown that  `Jpeg2000Input::open` does (the actual thread pool part is inside OpenJpeg code).

So add a dedicated `valid_file()` implementation that only needs to check 12 bytes of the header. While at it, I changed already existing `isJp2File()` to `is_jp2_header()` to better match naming conventions used elsewhere, and instead of trying to handle both little and big endian cases by manual repetition of two sets of magic integers, let's do just byte comparisons with `memcmp` instead.

On my PC (Ryzen 5950X, SSD, Windows VS2022), doing `ImageInput::create()` on 1138 files where they are not images at all (so OIIO in turns tries all the input plugins on them):

- Before: **3.4 seconds** spent in `Jpeg2000Input::open` (1.9s `opj_thread_pool_create`, 1.3s `opj_thread_pool_destroy`)
- After: **33 milliseconds** spent in `Jpeg2000Input::valid_file`

## Tests

No new tests. I checked behavior on the official Jpeg2000 conformance data set (https://github.com/uclouvain/openjpeg-data/tree/master/input/conformance), seems to work.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/jpeg2000.imageio/jpeg2000input.cpp
+++ b/src/jpeg2000.imageio/jpeg2000input.cpp
@@ -80,6 +80,7 @@ public:
         return feature == "ioproxy";
         // FIXME: we should support Exif/IPTC, but currently don't.
     }
+    bool valid_file(Filesystem::IOProxy* ioproxy) const override;
     bool open(const std::string& name, ImageSpec& spec) override;
     bool open(const std::string& name, ImageSpec& newspec,
               const ImageSpec& config) override;
@@ -97,7 +98,8 @@ private:
 
     void init(void);
 
-    bool isJp2File(const int* const p_magicTable) const;
+    static bool is_jp2_header(const uint8_t header[12]);
+    static bool is_j2k_header(const uint8_t header[5]);
 
     opj_codec_t* create_decompressor();
     void destroy_decompressor();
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM, great find!

(Note to self, or anybody else reading: I wonder if the openjpeg library has a facility to the caller to provide its own thread pool instead of having a new one created and potentially over-threading.)

lgritz:
Oof, the OpenJPEG library creates and destroys a threadpool for every file it reads (or tries to read)? It really doesn't just let the thread pool persist so that if multiple files are read, it doesn't do a full build/teardown each time?

lgritz:
Always good to see you helping out here, @aras-p. You never fail to turn up some really interesting thing that we had previously overlooked.

aras-p:
> Oof, the OpenJPEG library creates and destroys a threadpool for every file it reads (or tries to read)?

Looks like it. It's... not a design I would recommend, let's put it that way :)

PR #4565, opened 2024-12-17
fix: more ICC protections against invalid tag sizes

lgritz:


testsuite/jpeg-corrupt/ref/out.txt
```diff
@@ -20,3 +20,27 @@ src/corrupt-exif-1626.jpg :  256 x  256, 3 channel, uint8 jpeg
     oiio:ColorSpace: "sRGB"
 corrupt-icc-4551.jpg
 DCT coefficient (lossy) or spatial difference (lossless) out of range
+Reading src/corrupt-icc-4552.jpg
```
mugulmd, line 23: Quick question: since the file we're reading here is supposed to be corrupted, shouldn't we be failing when reading it ?

lgritz: Well, there's a balance -- if there's something wrong with one tag in the ICC profile, which is just one (non-essential to many apps) piece of metadata in the file, do we press on and try to read everything else that we can for the file? Or upon the first problem, no matter how minor, do we conservatively abort because it might be a sign that the whole file is suspicious or maybe even malicious? We usually err on the side of trying to read as much as we can, unless the whole file is unrecoverably broken.

This is the point of #4560: adding an attribute that, in the future, will give an app using OIIO more control over tradeoff I just described.

mugulmd: That makes sense !
LGTM then :)

mugulmd (approved):

lgritz:
Noting for the record that this turns out to also have fixed #4450.

PR #4567, opened 2024-12-22
api: make a 2-level namespace scheme

lgritz:
TL;DR: In main, change the enclosing namespace from `OpenImageIO_v3_1` to a 2-level scheme: `OpenImageIO::v3_1`.  This can't be backported to the current release for obvious ABI cmopatibility reasons; it is a forward-looking change only.

More ponderous explanation follows:

We have for some time had a single namespace that incorporates the version, defaulting to `OpenImageIO_v3_0` in the current release, for example.  A namespace alias, `OIIO`, always aliases the current namespace, so client applications can just say `OIIO::foo` without needing to change source code for every minor release (let alone if a build-time option sets up a custom namespace).

By bumping the namespace for every minor release, we use the symbol names themselves to enforce a rigid ABI compatibility test, so you can't accidentally compile against one minor release and link against an older minor release. This gives us the freedom to break the ABI (link compatibility) for each minor release, without subtle user errors. (Doing it wrong makes a total failure to link, which is hard to miss.)

But being able to introduce ABI changes annually by changing the enclosing namespace comes at the cost of a *complete* ABI compatibility break with every minor release. Even classes or functions that haven't changed at all will be incompatible by virtue of their changed symbol names.

This is an unfortunate limitation, and in an ideal world, we would like downstream users to be able to upgrade to a newer minor release more painlessly, and confident that they could even relink or perhaps compile with a request to use an old ABI.

I haven't fully worked out all the details, or even if this is going to be worth the trouble, but I think that a change we can introduce now that will allow more flexibiity in the future is to switch to a 2-level namespace scheme, for example, `OpenImageIO::v3_1` instead of the current `OpenImageIO_v3_1`.

The reason this might be helpful in the future is that we can use "inline" namespaces that default to finding things that are in, say, OpenImageIO::v3_2 without needing to specify it explicitly, while giving the ability to explicitly give an alternate inner versioned namespace. I haven't implemented that part here, it's reserved for future expansion (though I have experimented with it, it does work, but I want to discuss separately whether or not to actually do it).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -360,7 +360,7 @@ jobs:
             simd: "avx2,f16c"
             skip_tests: 1
             # abi_check: v2.6.2.0
-            abi_check: 7929c81619ea8201065c475761546d51091f6773
+            abi_check: 889cf33b61e84025750bdc2c827608fe6157c834
             setenvs: export OIIO_CMAKE_FLAGS="-DOIIO_BUILD_TOOLS=0 -DOIIO_BUILD_TESTS=0 -DUSE_PYTHON=0"
                             USE_OPENCV=0 USE_FFMPEG=0 USE_PYTHON=0 USE_FREETYPE=0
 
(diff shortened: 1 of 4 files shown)


lgritz:
Comments? I'd like to discuss this PR at the TSC meeting on Monday. I can give an outline I can present for how we might build on this going forward to achieve a very high degree of back compatibility for ABI even as we move from one minor release to another, but I seek feedback about whether that goal is worth the amount of clutter that it will introduce.

PR #4568, opened 2024-12-23
fix: IPTC fields have length limits

### LINKED ISSUE #4342: [HELP] "Damaged" JPGs in Photoshop
I first noticed this issue using Gaffer and since Gaffer uses OIIO for image writing I checked _oiiotool_ and found the same behavior. What happens is that Photoshop objects to JPGs exported with _oiiotool_: "This document contains Adobe Photoshop data which appears to be damaged." You can just go ahead and click OK and it opens fine. Affinity Photo or any other picture viewer are fine too, so this is mostly just about the inconvenience of having to click the warning in PS. Is there a workaround, maybe an _oiiotool_ parameter? Thanks!

lgritz:
It escaped our notice before that the IPTC spec dictates length limits for many fields. Getting this wrong can confuse some software, including Photoshop, apparently.

This patch enforces length limits wherever we could figure them out from the IPTC spec -- it will simply truncate those strings that are too long before writing them to a binary IPTC tag.

For this reason, we also are ending the practice of automatically translating several IPTC fields to and from what we figured were the equivalent generic metadata names. This was perhaps a dubious practice to begin with, but now the enforcement of length limits for IPTC (but not the generic metadata) makes it even more frought. There is probably nobody depending on this behavior (and maybe few OIIO users depending on IPTC support at all?), so now anybody purposely using IPTC metadata is fully responsible for setting it and dealing with any issues of whether it's "out of sync" with any other metadata that OIIO stores in or reads from a file.

Fixes #4342

### DIFF
--- a/src/doc/stdmetadata.rst
+++ b/src/doc/stdmetadata.rst
@@ -1063,6 +1063,16 @@ avoid conflicts with other plugins or standards.
     The history of the image or document.
 
 
+References for more information on IPTC metadata:
+
+* https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata
+* https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#iptc-core-schema-1-5-specifications
+  This is the one where you can find the length limits
+* ExifTool's documentation about IPTC tags (caveat: not a definitive
+  reference, could be outdated or incorrect):
+  https://exiftool.org/TagNames/IPTC.html
+
+
 SMPTE metadata
 ==============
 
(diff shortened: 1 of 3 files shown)


BrianHanke:
Just tested with my build, confirmed working great in Photoshop now!

jessey-git (approved):
I spot checked about a dozen of the newly added length limits and they all seem to match.  Do you have a link to the spec you were using as the only one I found was an unfriendly to parse through PDF.  It would be good to list the URL to the spec in the PR description for future archeologists if one is available.

lgritz:
I just added some URLs for reference both to the comments in iptc.cpp and the documentation (metadata chapter).

jessey-git (approved):

PR #4569, opened 2024-12-24
feat: Filesystem::getline()

lgritz:
Convenience function to safely read one line of text from an open FILE* and return it as a string.


lgritz:

PR #4570, opened 2024-12-25
ci: Upgrade to newer actions

lgritz:
* Upgrade upload-artifact action, old one is being deprecated.
* Update checkout and cache actions to latest.
* Fix workflow warning: save-always is not really an input to
  actions/cache.


lgritz:
Over a week with no objections + CI only change = merge

PR #4571, opened 2024-12-25
fix: some LoggedTimer instances lacked a variable name

lgritz:
It was syntactically correct, but without an actual variable, there was no RAII scope and so it destructs immediately, causing the timers to capture 0 time.

Found by Sonar warnings.


jessey-git (approved):
Looks like they're all taken care of now.

PR #4572, opened 2024-12-26
ci: fixups of analysis workflow

lgritz:
* Fix missing secrets that accidentally disabled the analysis workflow a few commits back. After splitting into reusable steps, we needed the main analysis workflow to let the steps workflow inherit secrets.
* Run analysis on TOT dev branches.
* For analysis, only build OIIO as debug, not dependencies.


lgritz:
Pre-emptively merging since this only affects CI, irrefutably fixes something that is broken, and doesn't change code or user build scripts.

PR #4573, opened 2024-12-27
testing: improve Ptex testing

lgritz:
* Ptex testsuite entry had been accidentally disabled somewhat recently when we switched to using Ptex exported cmake config instead of our own find module, because the "found" variable changed from PTEX_FOUND to Ptex_FOUND.
* Change the test to print stats, thus forcing the ptex test to read pixels instead of just header info, which improves the testing coverage of ptexinput.cpp.


lgritz:
Over a week without objections + testing/CI only change = merge

PR #4574, opened 2024-12-27
ci: test and document support for WebP 1.5 and fmt 11.1

lgritz:


lgritz:
Over a week with no objections + CI only change = merge

PR #4576, opened 2024-12-28
ci: Only pass build-steps the secrets it needs

lgritz:
Safer than letting it inherit all secrets.


JeanChristopheMorinPerso (approved):
This looks great! I'm all in for more secure workflows.

PR #4577, opened 2024-12-28
fix: various internal fixes to address Sonar and other warnings

lgritz:
Chipping away at minor warnings and areas identified by static analysis as potentially problematic or hard to analyze.

* benchmark.h: suppress warnings about uninitialized Benchmarker members.
* dpxoutput.cpp: change some raw pointers to spans
* dead code removal (color_ocio.cpp, filesystem.cpp, sysutil.cpp)
* exclude build-scripts from static analysis
* some misc warning suppression

### DIFF
--- a/src/dpx.imageio/dpxoutput.cpp
+++ b/src/dpx.imageio/dpxoutput.cpp
@@ -104,7 +104,7 @@ private:
 
     /// Helper function - set keycode values from int array
     ///
-    void set_keycode_values(int* array);
+    void set_keycode_values(cspan<int> array);
 };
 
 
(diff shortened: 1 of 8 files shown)


src/dpx.imageio/dpxoutput.cpp
```diff
@@ -370,8 +370,7 @@ DPXOutput::open(const std::string& name, const ImageSpec& userspec,
 
     ParamValue* kc = spec0.find_attribute("smpte:KeyCode", TypeKeyCode, false);
     if (kc) {
-        int* array = (int*)kc->data();
-        set_keycode_values(array);
+        set_keycode_values(cspan<int>((int*)kc->data(), 7));
```
jessey-git, line 373: How sure are we that there's actually 7 ints worth of data to walk through?

lgritz: Because the find() will only succeed for an item of TypeKeyCode, which is represented by 7 ints.

jessey-git: Ah yes, find does end up checking the type as well. Alright.

jessey-git (approved):

PR #4578, opened 2024-12-29
testing: better testing coverage of null image reader/writer

lgritz:
This improves the mediocre testing line coverage in nullimageio.cpp from 50% to 95%.

* Test writing
* Test reading of tiled and MIP-mapped null images
* Test setting and reading of attributes
* Simplify some code
* Along the way, identified and fixed a nullimage bug -- it was answering that it supported "rectangles" but had no write_rectangle method. Testing works!


grdanny (approved):
looks reasonable to me.

PR #4579, opened 2025-01-01
[feat]IBA:demosaic add X-Trans demosaicing

antond-weta:
## Description

Implements X-Trans pattern demosaicing using a simple bilinear algorithm. Adds the Xtrans pattern layout to the ImageBuf attributes in the raw reader plugin. 

## Tests

I have implemented unittests in imagebufalgo_test.cpp which cover all permutations of the Bayer and X-Trans layouts, as the X-Trans pattern is a bit too complicated to generate using oiiotool.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/imagebufalgo.h
+++ b/src/include/OpenImageIO/imagebufalgo.h
@@ -2164,7 +2164,9 @@ bool OIIO_API repremult (ImageBuf &dst, const ImageBuf &src,
 ///
 ///   - "pattern" : string (default: "bayer")
 ///
-///     The type of image sensor color filter array. Currently only the Bayer-pattern images are supported.
+///     The type of image sensor color filter array. Currently suported patterns:
+///     - `bayer` - Bayer-pattern image.
+///     - `xtrans` - X-Trans-pattern image.
 ///
 ///   - "algorithm" : string (default: "linear")
 ///
@@ -2174,11 +2176,14 @@ bool OIIO_API repremult (ImageBuf &dst, const ImageBuf &src,
 ///     - `MHC` - Malvar-He-Cutler linear demosaicing algorithm. Slower than `linear`, but produces 
 ///       significantly better results.
 ///
-///   - "layout" : string (default: "RGGB")
+///     The following algorithms are supported for X-Trans-pattern images:
+///     - `linear` - simple linear demosaicing. Fast, but can produce artefacts along sharp edges.
+///
+///   - "layout" : string (default: "RGGB" for Bayer, "GRBGBR BGGRGG RGGBGG GBRGRB RGGBGG BGGRGG" for X-Trans)
 ///
 ///     The order the color filter array elements are arranged in, pattern-specific.
 ///
-///   - "white-balance" : float[3] or float[4] (default: {1.0, 1.0, 1.0, 1.0})
+///   - "white-balance" : float[3] or float[4], (default: {1.0, 1.0, 1.0, 1.0})
 ///
 ///     Optional white-balancing weights. Can contain either three (R,G,B), or four (R,G1,B,G2) values.
 ///     The order of the white balance multipliers does not depend on the matrix layout.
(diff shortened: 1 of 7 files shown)


src/include/OpenImageIO/imagebufalgo.h
```diff
@@ -2174,11 +2176,14 @@ bool OIIO_API repremult (ImageBuf &dst, const ImageBuf &src,
 ///     - `MHC` - Malvar-He-Cutler linear demosaicing algorithm. Slower than `linear`, but produces 
 ///       significantly better results.
 ///
-///   - "layout" : string (default: "RGGB")
+///     The following algorithms are supported for X-Trans-pattern images:
+///     - `linear` - simple linear demosaicing. Fast, but can produce artefacts along sharp edges.
+///
+///   - "layout" : string (default: "RGGB" for Bayer, "GRBGBR BGGRGG RGGBGG GBRGRB RGGBGG BGGRGG" for X-Trans)
```
lgritz, line 2182: I'm not sure the text alignment is correct here. Can you double check this whole surrounding area? Also might be good to look at the artifacts saved by the "docs CI" to be sure that the resulting documentation definitely look as you imagine it should.

src/libOpenImageIO/imagebufalgo_demosaic.cpp
```diff
@@ -146,143 +173,246 @@ template<class Rtype, class Atype, int size> class BayerDemosaicing {
...
+        if (whitespaces)
+            length += pattern_size - 1;
+
+        layout.resize(length);
+
+        size_t i = 0;
+        for (size_t y = 0; y < pattern_size; y++) {
+            for (size_t x = 0; x < pattern_size; x++) {
+                int c = channel_at_offset(x + x_offset, y + y_offset);
+                assert(c < 4);
```
lgritz, line 286: Let's use OIIO_ASSERT (or OIIO_DASSERT if you are sure you want it to do nothing at all for non-debug builds).

lgritz (approved):
LGTM pending two very minor comments

src/include/OpenImageIO/imagebufalgo.h
antond-weta: Yes, I had checked the artefacts, those looked fine. 
I think I've spent a few hours fixing that bit, as I can't build the docs locally. The reason that alignment looks weird is because there are nested bullet point lists, and the 'linear' thingy is a single entry list (made it that way to be consistent with other things) belonging to the previous parameter ('algorithm').

antond-weta:
Larry, with the added unit tests covering all permutations, would it be worth simplifying the oiiotool-based tests to only check that every parameter gets passed to the algo?

src/libOpenImageIO/imagebufalgo_demosaic.cpp
antond-weta: Done!

lgritz:
> Larry, with the added unit tests covering all permutations, would it be worth simplifying the oiiotool-based tests to only check that every parameter gets passed to the algo?

I guess it's not critical to test it all twice. Either way is fine, I guess -- either have thorough C++ unit tests and a perfunctory oiiotool-based test to ensure that the params are all passed, OR if the oiiotool tests can exercise all corners of the algorithm, then the pure C++ tests are actually redundant. Whichever you think is better is fine with me.

src/include/OpenImageIO/imagebufalgo.h
lgritz: OK, thanks, I was just making sure since it was a 2-level list and I wasn't sure from looking these diffs whether everything was aligned as you intended. If they looked ok when viewing the artifacts, then all is fine.

lgritz (approved):
LGTM

lgritz:
N.B.: The Windows CI failure is unrelated to this patch and has already been fixed in main.

PR #4580, opened 2025-01-01
admin: Sign release artifacts

### LINKED ISSUE #4404: Consider signing release artifacts
The OpenSSF Best Practices Badge suggests [signing release artifacts](https://www.bestpractices.dev/en/criteria). Consider using OpenEXR's [release-sign.yml](https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/release-sign.yml) workflow as a template. It's triggered on release creation and does these steps:
1. Runs `get archive` to generate a `<release>.tar.gz` artifact
2. Signs the `<release>.tar.gz` via [sigstore](https://www.sigstore.dev/) 
3. Uploads the resulting sigstore signature file along with the tarball.

lgritz:
Approach copied from OpenEXR. Let's try it!

Fixes #4404


lgritz:
I'm going to merge this without review, since it's basically copied from OpenEXR, where it's known to work, and it doesn't affect our code, and there's not really any way to test it before merging and seeing if it does the right thing with today's releases.

PR #4582, opened 2025-01-03
paramlist.h: ParamValue::as_span(), as_cspan()

lgritz:
Convenience methods that give a bounded span of the data that a PV holds.


lgritz:
Any objections?

lgritz:
In review for 2 week, no objections -> will merge at the end of the day today if nobody has further comments.

PR #4583, opened 2025-01-03
testing: at long last, set up a softimage reading test

lgritz:
Also some very minor code cleanup that was helpful in setting up the test cases and getting better code coverage.


lgritz:
Any comments from the Softimage aficionados?

lgritz:
Two weeks to review with no objections -> I will merge at the end of the day if there are no further comments.

Issue #4586, opened 2025-01-06
[BUG] Build issue with libraw


OgreTransporter:
In PR #4556 the option `COMPILE_OPTIONS` was added to the macro `add_oiio_plugin` which adds these to the plugin with `target_compile_options`. The reason for this was to pass the variable `LibRaw_DEFINITIONS` to the compiler. Unfortunately a mistake was made, because in `FindLibRaw.cmake` lines 99 and 100 this variable is used as a list for definitions as the name suggests.
https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/e964a6af111c07387387bd24bd324b5618630e38/src/cmake/modules/FindLibRaw.cmake#L99

The original code in the CMakeLists.txt of the raw plugin `DEFINITIONS "USE_LIBRAW=1" ${LibRaw_r_DEFINITIONS}` was therefore correct. The new variant `COMPILE_OPTIONS ${LibRaw_r_DEFINITIONS} DEFINITIONS "USE_LIBRAW=1"` treats the list of definitions as compiler flags, which does not work. `LIBRAW_NODLL` is not a valid compiler argument, so an attempt is made to compile the file `LIBRAW_NODLL`. As this file does not exist, an error occurs.

It is easy to fix by adding `-D` to lines 99 and 100 of `FindLibRaw.cmake`:

```
        set (LibRaw_r_DEFINITIONS ${LibRaw_r_DEFINITIONS} -D LIBRAW_NODLL)
        set (LibRaw_DEFINITIONS ${LibRaw_DEFINITIONS} -D LIBRAW_NODLL)
```

OgreTransporter:
Here I think there is a misunderstanding of the functions `target_compile_options` and `target_compile_definitions`. `target_compile_options` adds options to the compiler without any changes. The user is responsible for ensuring that these are correct. If you want to set a definition, you must therefore also specify `-D`. `target_compile_definitions` treats all passed values as definitions and automatically appends them to the compiler with `-D`. This is the normal behaviour as intended by CMake.

PR #4587, opened 2025-01-06
fix: Build issue with libraw #4586

OgreTransporter:
## Description
Fix issue #4586

## Checklist:

  doesn't change the API.)
  nonconforming.


OgreTransporter:
Someone else will have to patch this, EasyCLA doesn't work for me.

lgritz:
Do you want to try to fix the EasyCLA? The most common problem people have is that they made the commit using a different email than the one they used with the EasyCLA system. If that's not it and there is some other problem (it happens), we can open a support ticket with LF to get it straightened out.

lgritz:
I'm reopening for now so this isn't forgotten while we straighten this out or wait for somebody else to submit a replacement PR.

lgritz:
Oops, I can't re-open because the contributor's branch is already deleted. I will resubmit it on my end.

OgreTransporter:
Ops, I'm sorry. I must have deleted the fork too quickly.

PR #4588, opened 2025-01-06
build: Fix libraw definitions (again)

### LINKED ISSUE #4586: [BUG] Build issue with libraw

lgritz:
Fixes #4586
Replaces #4587


lgritz:
@OgreTransporter

OgreTransporter:
This is not necessary in this way. If the definitions are added via `COMPILE_OPTIONS`, `-D` must be specified because `target_compile_options` is used in the background which adds pure compiler commands. If the definitions are added via `DEFINITIONS`, `-D` does not have to be specified because `target_compile_definitions` is used in the background, which automatically adds `-D` if necessary.

lgritz:
Oh, I see now. The libraw definitions aren't really just "definitions" (in the preprocessor symbol sense), they may have compiler flags mixed in.

lgritz:
Amended the PR.

OgreTransporter:
> The libraw definitions aren't really just "definitions" (in the preprocessor symbol sense), they may have compiler flags mixed in.

This depends on the platform and the compiler. In my MSVC version of libraw, no flags are set in the package, but this can be different depending on the build options.

The PR should fix that problem. Thanks!

lgritz:
Yeah, it's obviously rare. I'm not actually sure what combination of factors makes it have any extra compile flags, but it obviously did for somebody to prompt the PR that started all this.

PR #4589, opened 2025-01-07
doc: fix typo in oiiotool's gradient fill argument

mugulmd:
## Description

Correct `botromright` typo in `oiiotool` documentation.


lgritz (approved):
Thanks for catching that typo!

Over time, we are trying to migrate all of the code examples in the docs into actual code that's executed as part of the testsuite and merely included by reference into the docs, guaranteeing that this kind of mistake can't happen. But we don't have it all converted yet.

Issue #4546, opened 2024-11-30
[FEATURE REQUEST] oiiotool command to split "layers" into subimages or separate images


lgritz:
Just like we have oiiotool command `--sisplit` which splits an image's subimages (what exr calls "parts") into separate images on the oiiotool stack, it might be handy to have `--layersplit` (split channel-name-based layers into separate images on the stack) and/or `--layertosi` (split the layers in an image into an image organized by subimages/parts). In both cases, the layers would need to be deduced by looking at and grouping the channel names.

Background:

Modern (>= 2.0) OpenEXR has "multi-part" files that can contain a different raster image in each part (for example, a beauty pass, reflections, isolated lights, mattes, etc). The parts are read separately, efficiently. When dinosaurs were alive and used OpenEXR 1.x, the lack of multi-part led to a convention called "multi-layer" exr, in which a ton of channels were jammed into a (single part of a) single image, and channels were grouped simply according to a naming convention, for example "LAYERNAME.channelname", and all the channels with the same LAYERNAME were considered a "layer" or group that conceptually goes together.

Ideally, nobody would know or care about this for the last several years, because they'd be using parts instead. But there is some software out there still making "multi-layer exr files" where dozens of channels are jammed into a single part. So maybe we need the equivalent of --sisplit for layers? (Note: we can already group separate images into a single image with multiple "layers" by using the existing --chappend plus ensuring that the channels are named according to the layer naming conventions, so we don't need a new command for going in that direction.)

mugulmd:
Hey @lgritz, I can take a look at that !
Do you have some example images in mind for testing ? Or should I go on and find/make some on my own ?

lgritz:
I don't, but it should be easy to construct one with oiiotool -- it's really just a matter of making a low-res image with a bunch of channels that have the right names to group them into "layers."

```
oiiotool -pattern constant:color=1,2,3,4,5,6,7,8,9,10 4x4 10 -chnames "R,G,B,A,diffuse.R,diffuse.G,diffuse.B,specular.R,specular.G,specular.B" -d half -o layers.exr
```

Then what we want is a new command like

```
oiiotool layers.exr --layersplit -o parts.exr
```

And then `oiiotool --info -v -a --stats parts.exr` should show that parts.exr has 3 parts, the first containing the original R,G,B,A channels, the second containing the three diffuse channels, and the third containing the three specular channels.

[...]

PR #4591, opened 2025-01-13
feat(oiiotool): oiiotool --layersplit, new command to split layers

mugulmd:
## Description

Implement FR from https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4546

Add a new `--layersplit` command to `oiiotool` to split an image into its channel-name-based layers onto the stack.

The extracted layer names are then stored in the `oiio:subimagename` metadata (in case it is needed for later use, e.g merging these "layers" as sub-images), and the extracted channel names replace the old channel names in the new images.

Example:
an image with channels `R, G, B, A, diffuse.R, diffuse.G, diffuse.B` will be split into two images, the first one with channels `R, G, B, A` and the second one named `diffuse` with channels `R, G, B`.

> Note: we did not implement the `--layertosi` command suggested in the FR as it can already be done with a combination of `--layersplit` and `--siappendall`

## Tests

We add a test in the test suite that takes a "multi-layer" image, splits the layers and merges them as sub-images. The final image should have 3 sub-images.

## Checklist:

  doesn't change the API.)
  nonconforming.


src/oiiotool/oiiotool.cpp
```diff
@@ -2880,6 +2880,50 @@ action_subimage_split(Oiiotool& ot, cspan<const char*> argv)
...
+    ImageRecRef A = ot.pop();
+    ot.read(A);
+
+    // Push the individual channel-name-based layers onto the stack
+    ImageSpec* spec = A->spec();
+    int chbegin = 0;
+    for (size_t i = 0; i < spec->channelnames.size(); ++i) {
+        const std::string & chname          = spec->channelnames[i];
+        const std::string::size_type dotpos = chname.find(".");
+        const std::string layername         = (dotpos == std::string::npos) ? "" : chname.substr(0, dotpos);
```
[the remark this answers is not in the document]
mugulmd: Definitely 😅 Thanks for the tip !

lgritz:
Closes #4546

src/doc/oiiotool.rst
```diff
@@ -2193,6 +2193,16 @@ current top image.
     image comprised of the subimages of all original images appended
     together.
 
+.. option:: --layersplit
+
+    Remove the top image from the stack, split it into its consituent
+    channel-name-based layers, and push them all onto the stack
+    (first to last).
```
lgritz, line 2200:
```suggestion
    Remove the top image from the stack, split it into a separate image for
    each of its constituentchannel-name-based layers, and push them all
    onto the stack (first to last).
```

lgritz: A spelling fix, and also added a clause to clarify that you are splitting into multiple images, not just subimages or whatever.

src/doc/oiiotool.rst
```diff
@@ -2193,6 +2193,16 @@ current top image.
...
+
+    By "layer" we mean a subset of the initial channels which, when named
+    using the convention "LAYERNAME.channelname", all share the same layer
+    name.
+
```
lgritz, line 2205: Do you want to clarify what should happen to the channels that don't have any '.' at all? Are they gathered into one image together? Is it guaranteed to be the first image?

lgritz (commented):
The code LGTM, great job!

On the docs, I made one spelling and phrasing suggestion, as well as a question/suggestion aimed at clarifying what happens to channels that aren't part of a named layer. That's about all I can think of to make this PR better, it's really great.

Breaking out an entirely new test for this as you did is fine, though I wanted to let you know (in case you're inclined to change it) that it would have also been ok to add one more command line to an existing test, whichever of the oiiotool related tests looked like it contained the most related commands (maybe wherever the tests for --siappend and --sisplit are, so that related commands have their tests together).

lgritz:
One more FYI: When mentioning an issue in a PR, GitHub expects you to use a particular phrasing in order to truly link them in such a way that accepting the PR will automatically close the issue:

    Closes #xxx

or

    Fixes #xxx

Anything else is just talk, and doesn't really link the issue.

src/doc/oiiotool.rst
mugulmd: Updated, thanks !

src/doc/oiiotool.rst
mugulmd: Yes and yes ! (I've just checked the exr input code just to make sure)
I've updated the docs :)

mugulmd:
Hey @lgritz 
Thanks for the review (and all the tips on how the repo works, super helpful !), I think the code should be ready to go now :)

lgritz (approved):
LGTM!

PR #4592, opened 2025-01-13
ci: Fix Windows 2019 CI -- make python version match the runner

lgritz:
For reasons that are unclear (maybe a change to which python versions are installed where on the GH runners?) this seems to have broken within the past couple days, in a way that did not correspond to any change we made on the OIIO side.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -588,7 +588,7 @@ jobs:
             runner: windows-2019
             vsver: 2019
             generator: "Visual Studio 16 2019"
-            python_ver: 3.7
+            python_ver: "3.9"
           - desc: Windows-2022 VS2022
             runner: windows-2022
             vsver: 2022


lgritz:
@zachlewis This should solve your broken PR as well. Can you review?

zachlewis:
Ah, interesting. Since we're not constraining the VS2019 build to python 3.7, I'll have to add a setenvs bit to this task as well (export OPENIMAGEIO_PYTHON_LOAD_DLLS_FROM_PATH=1).

LGTM!

lgritz:
I think it all works ok when there's just one python version, but when there might be multiple (like on the GH runners), we use PYTHON_VERSION in our build system to select a particular one. I think what happened is that the python 3.7 installation maybe was removed from the runner? There's a 3.9 there, but the explicit request for 3.7 we were using for this CI run started failing just within the last couple days.

PR #4594, opened 2025-01-15
ci(deps): Adjust some pugixml versions

lgritz:
* Bump the 'latest' test to test the actual latest, 1.15.
* Be sure that the 'oldest' tests actually test the oldest we support, 1.8.
* Sprinkle around a couple other versions that we don't test currently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,7 @@ jobs:
             setenvs: export  CMAKE_VERSION=3.18.2
                              PTEX_VERSION=v2.3.2
                              WEBP_VERSION=v1.1.0
+                             PUGIXML_VERSION=v1.8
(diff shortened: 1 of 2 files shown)


jessey-git (approved):
Seems alright to me - test the oldest and test the most recent 3, I can't argue much with that. CI failure doesn't seem related?

lgritz:
The CI failure appears to be an inability to clone a copy of pybind11 from github, which is unrelated to this PR and I assume is a spurious failure. I'll rerun that CI job just to be sure, it was probably a very temporary glitch.

It's absolutely shocking how many non-repeating failures happen in these GHA CI runs on a daily basis, usually the temporary inability to `git clone` something, or else the instanced runner image not having something that's supposed to be there.

PR #4596, opened 2025-01-15
fix: Recent change broke when using non-Apple clang on Apple

lgritz:
This fixes a problem introduced with #4582


lgritz:
Merging without review because it fixes an important build break for OSL.

PR #4597, opened 2025-01-15
feat(span.h): span_memcpy is a safer memcpy when you know the span boundaries

lgritz:


lgritz:
No objections for 2 weeks -> merge

PR #4600, opened 2025-01-17
fix(pdf): Perform endian byteswap on correct buffer area for PSD RLE

### LINKED ISSUE #4599: [BUG] Crash when reading certain PSD files
**Describe the bug**

A crash occurs when reading certain PSD files. Looks like it's due to RLE encoding... PR incoming

Two versions were tested:
```
OIIO 2.5.11.0 | unknown arch?
    Build compiler: MSVS 1929 | C++14/199711

And

OIIO 3.1.0.0dev | unknown arch?
    Build compiler: MSVS 1943 | C++17/199711
    HW features enabled at build: sse2,sse3,ssse3,sse41,sse42,avx,avx2
```

**To Reproduce**

1. Try reading the attached file using OIIO
2. Crash inside `PSDInput::read_native_scanline`

Python repro
```
import OpenImageIO as oiio
inp = oiio.ImageInput.open("psd-test.psd")

spec = inp.spec()
nchannels = spec.nchannels
pixels = inp.read_image(0, 0, 0, nchannels, "float")  # does not matter if uint16 or uint8 is attempted either
```

[psd-test.zip](https://github.com/user-attachments/files/18446771/psd-test.zip)

jessey-git:
## Description

Perform the endian byteswap on the correct buffer area for PSD RLE decoding.  The original code used `dst` but that pointer is modified from its original value as it progresses through decoding.

Fixes #4599 

## Tests

It might be prudent to add a test for this.  Is the repro file attached to the bug too big for this repo(842kb) or do you want them in the oiio-images repo maybe?

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -2233,6 +2233,7 @@ PSDInput::decompress_packbits(const char* src, char* dst,
     int16_t header;
     int length;
 
+    char* dst_start = dst;
     while (src_remaining > 0 && dst_remaining > 0) {
         header = *reinterpret_cast<const signed char*>(src);
         src++;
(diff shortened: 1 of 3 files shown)


lgritz (approved):
Good catch!

lgritz:
Yeah, I would say that anything several hundred KB or more is probably better in the oiio-images repo so that it's only needed of people want to run the testsuite and not if they're just making an untested build.

This is approved as-is as far as I'm concerned. Do you want me to merge now, and we can add the test later, or would you like me to wait while you add the test?

jessey-git:
I added a PR to the images repo, and if/when that goes in, I'll add a test line for it over here.

lgritz (approved):
LGTM!

PR #4601, opened 2025-01-21
feat(raw): add black level and BPS metadata

antond-weta:
## Description

Adds attributes to the output ImageBuf for the black level and bit depth when the raw bayer buffer is requested.

## Tests

I believe we don't test the retrieved metadata currently. We may want to start doing that eventually?

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (approved):
LGTM

PR #4605, opened 2025-01-26
ci: Unbreak the scorecard workflow

lgritz:
* Fix syntax error, other minor touch-ups
* Don't run other workflows when only scorecard.yml changes


lgritz:
Merging to unbreak CI

PR #4606, opened 2025-01-27
cleanup: Get rid of some compiler symbols no longer needed

lgritz:
The newer versions of the compilers we now use no longer need this.


lgritz:
Any objections? I think there is no longer any need for this removed code.

lgritz:

But I'm not going to backport it because there is a small chance that something downstream is depending on the fact that we included that header (even though we don't require it).

PR #4607, opened 2025-01-27
cleanup: no longer need OIIO_INLINE_CONSTEXPR macro

lgritz:
Our minimum C++ version allows `inline constexpr`, just use it.

### DIFF
--- a/src/include/OpenImageIO/platform.h
+++ b/src/include/OpenImageIO/platform.h
@@ -173,13 +173,10 @@
 // See https://en.cppreference.com/w/cpp/compiler_support
 //
 // OIIO_CPLUSPLUS_VERSION : which C++ standard is compiling (14, 17, ...)
-// OIIO_CONSTEXPR14 :
 // OIIO_CONSTEXPR17 :
 // OIIO_CONSTEXPR20 : constexpr for C++ >= the designated version, otherwise
 //                    nothing (this is useful for things that can only be
 //                    constexpr for particular versions or greater).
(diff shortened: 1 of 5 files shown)


src/include/OpenImageIO/platform.h
```diff
@@ -192,15 +189,16 @@
...
 #elif (__cplusplus >= 201703L) || (defined(_MSC_VER) && _MSC_VER >= 1914)
 #    define OIIO_CPLUSPLUS_VERSION 17
 #    define OIIO_CONSTEXPR20 /* not constexpr before C++20 */
-#    define OIIO_INLINE_CONSTEXPR inline constexpr
 #else
 #    error "This version of OIIO is meant to work only with C++17 and above"
 #endif
 
+// DEPRECATED(3.1): use C++17 inline constexpr
+#define OIIO_INLINE_CONSTEXPR inline constexpr
```
mugulmd, line 200: Do we really need to keep this macro ? From what I can see you've updated all the `OIIO_INLINE_CONSTEXPR` in the code base, so maybe we can just get rid of it, this way we have one less macro :)

lgritz: It's in a public header file, we can't break downstream source compatibility, even if it's something we don't use internally any longer.

mugulmd: Ah you're right, good point !

mugulmd (approved):

PR #4610, opened 2025-01-30
feat(oiiotool): oiiotool new expression eval tokens IS_CONSTANT, IS_BLACK

lydia-zheng:
## Description

Included two additional metadata boolean checks for constant and black frame images, so that handling for those image cases are facilitated.

IS_CONSTANT and IS_BLACK return 1 when true, and 0 when false.

## Tests

No new tests added to testsuite for now, but tested with local build copy of oiiotool and test images to ensure it behaved as expected.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/oiiotool/expressions.cpp
+++ b/src/oiiotool/expressions.cpp
@@ -334,8 +334,7 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
                     // trusting that the constantcolor check means all channels have the same value, so we only check the first channel
                     if (color[0] == 0.0f) {
                         result = "1";
-                    }
-                    else {
+                    } else {
                         result = "0";
                     }
                 } else {
--- a/src/oiiotool/expressions.cpp
+++ b/src/oiiotool/expressions.cpp
@@ -317,7 +317,6 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
                 result = out.str();
                 if (result.size() && result.back() == '\n')
                     result.pop_back();
-
             } else if (metadata == "IS_CONSTANT") {
                 std::vector<float> color((*img)(0, 0).nchannels());
                 if (ImageBufAlgo::isConstantColor((*img)(0, 0), 0.0f, color)) {
@@ -325,8 +324,6 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
                 } else {
                     result = "0";
                 }
-
-
             } else if (metadata == "IS_BLACK") {
                 std::vector<float> color((*img)(0, 0).nchannels());
                 // Check constant first to guard against false positive average of 0 with negative values i.e. -2, 1, 1
--- a/src/oiiotool/expressions.cpp
+++ b/src/oiiotool/expressions.cpp
@@ -317,6 +317,32 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
                 result = out.str();
                 if (result.size() && result.back() == '\n')
                     result.pop_back();
+
+            } else if (metadata == "IS_CONSTANT") {
+                std::vector<float> color((*img)(0, 0).nchannels());
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), 0.0f, color)) {
+                    result = "1";
+                } else {
+                    result = "0";
+                }
+
+
+            } else if (metadata == "IS_BLACK") {
+                std::vector<float> color((*img)(0, 0).nchannels());
+                // Check constant first to guard against false positive average of 0 with negative values i.e. -2, 1, 1
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), 0.0f, color)) {
+                    // trusting that the constantcolor check means all channels have the same value, so we only check the first channel
+                    if (color[0] == 0.0f) {
+                        result = "1";
+                    }
+                    else {
+                        result = "0";
+                    }
+                } else {
+                    // Not even constant color case -> We don't want those to count as black frames.
+                    result = "0";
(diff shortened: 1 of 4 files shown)


src/oiiotool/expressions.cpp
```diff
@@ -317,6 +317,35 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
                 result = out.str();
                 if (result.size() && result.back() == '\n')
                     result.pop_back();
+
+            } else if (metadata == "IS_CONSTANT") {
+                std::vector<float> color((*img)(0, 0).nchannels());
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), color)) {
```
lgritz, line 323: I think this is not quite correct for the declaration of isConstantColor. I believe you want

```suggestion
                if (ImageBufAlgo::isConstantColor((*img)(0, 0), 0.0f, color)) {
```

The 2nd parameter is the threshold amount (0.0f would mean a "true constant", exact same value everywhere), and the 3rd parameter is the place to store the constant value it finds.

src/oiiotool/expressions.cpp
```diff
@@ -317,6 +317,35 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
...
+                } else {
+                    result = "0";
+                }
+
+
+            } else if (metadata == "IS_BLACK") {
+                auto pixstat = ImageBufAlgo::computePixelStats((*img)(0, 0));
+                std::vector<float> color((*img)(0, 0).nchannels());
+                // Check constant first to guard against false positive average of 0 with negative values i.e. -2, 1, 1
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), color)) {
```
lgritz, line 334:
```suggestion
                if (ImageBufAlgo::isConstantColor((*img)(0, 0), 0.0f, color)) {
```

src/oiiotool/expressions.cpp
```diff
@@ -317,6 +317,35 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
...
+                std::vector<float> color((*img)(0, 0).nchannels());
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), color)) {
+                    result = "1";
+                } else {
+                    result = "0";
+                }
+
+
+            } else if (metadata == "IS_BLACK") {
+                auto pixstat = ImageBufAlgo::computePixelStats((*img)(0, 0));
```
lgritz, line 331: I don't think you need to call computePixelStats. You can save the expense, since you are calling isConstantColor() and that will deposit the constant color it finds into `color`, then you can just check if the values of `color[0..nchans-1]` are all 0.0f to know if it's black.

src/oiiotool/expressions.cpp
```diff
@@ -317,6 +317,35 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
...
+                std::vector<float> color((*img)(0, 0).nchannels());
+                // Check constant first to guard against false positive average of 0 with negative values i.e. -2, 1, 1
+                if (ImageBufAlgo::isConstantColor((*img)(0, 0), color)) {
+                    // Do we need this loop if we know that all channels should be the same value in constant frame?
+                    for (size_t i = 0; i < pixstat.avg.size(); ++i) {
+                        // If any of the pixel doesn't have an average of (0,0,0), then we don't have a black frame
+                        // Is this the best way to check? Feels fishy to have the specified number...should we look at 0.0f or other formats?
+                        if (pixstat.avg[i] == 0.000000) {
```
lgritz, line 339: `0.0f` is sufficient. But you don't need to look at the image's average color -- you can just use color[0] since isConstantColor returns that constant color.

src/oiiotool/expressions.cpp
```diff
@@ -317,6 +317,35 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
...
+                            result = "1";
+                        } else {
+                            result = "0";
```
lgritz, line 342: I think this logic is incorrect. Doesn't this end up setting result based solely on whether the LAST channel is black or not? You want the result to be "0" if *any* channel is not black, right?

lgritz (changes requested):
Great feature addition, Lydia. I made a few comments, I think there is a logic error here, but it should all be an easy fix for you.

lydia-zheng:
> Great feature addition, Lydia. I made a few comments, I think there is a logic error here, but it should all be an easy fix for you.

Thank you so much for taking such a thorough look, Larry! 
Seeing all the comments and reading them over now.
Some of these are definitely what I was suspecting I got wrong, glad it's confirmed and I'll start looking into the fix in my free cycles :D

lydia-zheng:
Hi @lgritz !

Took another jab at this with your amazing feedback, and fixed the feedstock/git commits such that CLs are all green.

Let me know if this is good to merge or if the new logic still looks off!

lgritz (approved):
LGTM, thanks Lydia. Ready to merge.

lgritz:
Just before merging, I got cold feet. I didn't feel good about this not having a test, so I took the liberty of adding one and amending your PR (it's just a few lines). I'll merge after it passes PR.

lydia-zheng:
> Just before merging, I got cold feet. I didn't feel good about this not having a test, so I took the liberty of adding one and amending your PR (it's just a few lines). I'll merge after it passes PR.

That’s super fair and thanks for adding the tests and checking, Larry!

Actually had some local files in test suite but was waiting for the base PR to go through first, though what you did was much better for ensuring it worked.

PR #4616, opened 2025-02-01
ci: Switch to using ARM-native linux runners

zachlewis:
## Description

Use the publicly-available ARM-native linux runners for CI checks and building Python wheels


lgritz (approved):
LGTM!

Seems to be working, I will merge when the last CI tests are done.

PR #4618, opened 2025-02-02
fix(build): Fix recently broken rpath setting

lgritz:
As part of the recent big python wheel merge, compiler.cmake got changed in a subtle way in how it sets CMAKE_INSTALL_RPATH.

This broken iv for me on a Mac with dynamic libraries. Strangely (and I'm too lazy to chase down exactly why), oiiotool seems fine.  So this was missed entirely by the CI, which never runs 'iv', since it's an interactive app. (Note to self: we should find some way to at least verify that it can execute as part of the testsuite.)

Anyway, I believe that this change adds the install lib area back to the path, and it does seem to repair my ability to run iv on my end.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -698,9 +698,9 @@ else ()
         else()
             set(BASEPOINT $ORIGIN)
         endif()
-        set (CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR}
-                                 ${BASEPOINT}
-                                 ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
+        set (CMAKE_INSTALL_RPATH ${BASEPOINT}
+                                 ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR}
+                                 ${BASEPOINT}/../${CMAKE_INSTALL_LIBDIR})
     endif ()
     # add the automatically determined parts of the RPATH that
     # point to directories outside the build tree to the install RPATH
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -698,7 +698,9 @@ else ()
         else()
             set(BASEPOINT $ORIGIN)
         endif()
-        set (CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
+        set (CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR}
+                                 ${BASEPOINT}
+                                 ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
     endif ()
     # add the automatically determined parts of the RPATH that
     # point to directories outside the build tree to the install RPATH
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -699,7 +699,6 @@ else ()
             set(BASEPOINT $ORIGIN)
         endif()
         set (CMAKE_INSTALL_RPATH ${BASEPOINT}
-                                 ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR}
                                  ${BASEPOINT}/../${CMAKE_INSTALL_LIBDIR})
     endif ()
     # add the automatically determined parts of the RPATH that
--- a/src/cmake/fancy_add_executable.cmake
+++ b/src/cmake/fancy_add_executable.cmake
@@ -53,12 +53,6 @@ macro (fancy_add_executable)
         target_link_libraries (${_target_NAME} PRIVATE ${_target_LINK_LIBRARIES})
         target_link_libraries (${_target_NAME} PRIVATE ${PROFILER_LIBRARIES})
         set_target_properties (${_target_NAME} PROPERTIES FOLDER ${_target_FOLDER})
-        if (SKBUILD) 
-            # The installed bin and lib directories are at the same level, so we
-            # need to set the rpath to look for the libraries in ../lib
-            set_target_properties (${_target_NAME} PROPERTIES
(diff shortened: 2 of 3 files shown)


lgritz:
@zachlewis @JeanChristopheMorinPerso Opinions/review desired.

src/cmake/compiler.cmake
```diff
@@ -698,7 +698,9 @@ else ()
         else()
             set(BASEPOINT $ORIGIN)
         endif()
-        set (CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
+        set (CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR}
+                                 ${BASEPOINT}
+                                 ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
```
JeanChristopheMorinPerso, line 703: I'm tempted to suggest something like this.
```suggestion
        set (CMAKE_INSTALL_RPATH ${BASEPOINT} ${BASEPOINT}/${CMAKE_INSTALL_LIBDIR})
        if (NOT SKBUILD)
            set (CMAKE_INSTALL_RPATH ${CMAKE_INSTALL_FULL_LIBDIR} ${CMAKE_INSTALL_RPATH})
        endif() 
```
This would avoid having the full path when building the wheels. The wheels should be as hermetic as possible.

lgritz: Sure, that makes sense.

lgritz: Does it make sense to add `${BASEPOINT}/../lib` to the SKBUILD clause?
BASEPOINT is going to be `something/bin`, where the executable lives. `${BASEPOINT}/${CMAKE_INSTALL_LIBDIR}` is therefore `something/bin/lib` which is useless, that's never where the libraries will be, right?

JeanChristopheMorinPerso: Umm, won't `BASEPOINT` be `lib` for the library? We technically need two different RPATH don't we? One for the executables and one for the libs no?

lgritz: RPATH is only used to find libraries, whether the thing needing the library is a binary or another library. So RPATH always needs to include the lib directory.

JeanChristopheMorinPerso: Sure, what I meant was that the executable will have a different RPATH. The executable needs `$ORIGIN/../lib` while the library(ies) need `$ORIGIN`.

lgritz: Right. But  ${BASEPOINT}/../${CMAKE_INSTALL_LIBDIR} points to the same correct place in both cases, and seems to me less error-prone than trying to set rpath attributes separately for every library and binary target -- seems easy for somebody adding libraries or binaries later or modifying the cmake scripts to mess it up by not understanding this subtlety.

lgritz: I pushed an amendment to the PR that swaps the absolute path for what I was discussing above.

lgritz:
Any further comments on this?

src/cmake/compiler.cmake
zachlewis: As far as executables go, we're actually editing the RPATH in a couple of other places as well:

In [src/cmake/fancy_add_executable.cmake:](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c4873d57e78c349ac19c3af50165ec866698fafb/src/cmake/fancy_add_executable.cmake#L56-L61)
```
        if (SKBUILD) 
            # The installed bin and lib directories are at the same level, so we
            # need to set the rpath to look for the libraries in ../lib
            set_target_properties (${_target_NAME} PROPERTIES
                INSTALL_RPATH "$<IF:$<BOOL:${APPLE}>,@loader_path,$ORIGIN>/../${CMAKE_INSTALL_LIBDIR}")
        endif ()

```

and in [src/libutil/CmakeLists.txt](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c4873d57e78c349ac19c3af50165ec866698fafb/src/libutil/CMakeLists.txt#L108-L115)
```
 if (SKBUILD)
        set (PYTHON_SITE_DIR .)
        set_target_properties (${targetname} PROPERTIES
                INSTALL_RPATH "$<IF:$<BOOL:${APPLE}>,@loader_path,$ORIGIN>/${CMAKE_INSTALL_LIBDIR}")
        install_targets (NAMELINK_SKIP ${targetname})
    else ()
        install_targets (${targetname})
    endif ()
    ```

lgritz: OH, I didn't realize we were tweaking it in those spots as well. So it seems to me that the real problem is that the one in fancy_add_executable is ONLY happening for SKBUILD. When I build on my Mac from source, it's not getting this logic that includes the ../lib suffix, and that's why it's not finding the library.

lgritz: So you have some special logic in libutil (for libOpenImageIO_Util), but not in libOpenImageIO (for the main libOpenImageIO)?

Can all these scattered special cases be removed and just use the one in compiler.cmake for everything?

zachlewis: > So you have some special logic in libutil (for libOpenImageIO_Util), but not in libOpenImageIO (for the main libOpenImageIO)?

Yep, to make sure libOpenImageIO_Util can find libOpenImageIO.
(libOpenImageIO doesn't link against libOpenImageIO_Util)
(Without doing this, the default cibuildwheel repair wheel step would flag libOpenImageIO_Util as missing dependencies (even though libOpenImageIO was _right there in the same directory_. So frustrating) 

> Can all these scattered special cases be removed and just use the one in compiler.cmake for everything?

I do believe so!!

lgritz: OK, updated once again. Got rid of the special cases in fancy_add_executable and in libutil/CMakeLists.txt. Just the one place in compiler.cmake, and it has both the basepoint and basepoint/../lib, which I think covers all of our bases.

zachlewis (approved):

PR #4624, opened 2025-02-04
fix(rla): More robust to corrupted RLA files that could overrun buffers

lgritz:


lgritz:
Comments?

lgritz:
2 weeks with no objections -> merging

PR #4625, opened 2025-02-04
fix(ico): More robust to corrupted ICO files

lgritz:


lgritz:
Any comments?

src/ico.imageio/icoinput.cpp
```diff
@@ -190,6 +190,17 @@ ICOInput::seek_subimage(int subimage, int miplevel)
...
     }
 
+    // some sanity checking
+    if (subimg.bpp != 1 && subimg.bpp != 4 && subimg.bpp != 8
+        && subimg.bpp != 16 && subimg.bpp != 24 && subimg.bpp != 32) {
+        errorfmt("Unsupported image color depth, probably corrupt file");
+        return false;
+    }
+    if (subimg.reserved != 0) {
+        errorfmt("Probably corrupt file");
```
EmilDohne, line 200: [nitpick] This could be a little more descriptive on why the file is corrupt for logging purposes

EmilDohne:
LGTM otherwise

src/ico.imageio/icoinput.cpp
lgritz: So fixed, thanks for the suggestion.

PR #4627, opened 2025-02-06
oiiotool: better handling of wildcards that match no files

### LINKED ISSUE #4623: [BUG] oiiotool not returning correct error code for wildcard + missing file
Paraphrased from a discussion on the Slack where a user said:

Today I stumbled on an interesting behaviour in oiiotool that shadowed another (internal) bug and got my head scratching for a bit.
Image you’re doing a simple operation, but none of the input files exist. oiiotool will nicely report that as an issue:
```
$ oiiotool --runstats non-existent.exr non-existent.exr  --over -o test.exr
oiiotool ERROR: read : File does not exist: "non-existent.exr"

$ echo $?
1
```

However, if you use the frame syntax to refer to non existent frames, it will not flag any issue, and exit with 0:
```
$ oiiotool --runstats  non-existent.%04d.exr non-existent.exr --over -o test.exr
oiiotool WARNING : oiiotool produced no output. Did you forget -o?

$ echo $?
0
```

I tested this is with
```
$ oiiotool --version
2.5.14.0
```

Not a biggie, but I wonder if the single-file check that happens in the first case could be expanded to work with the supported syntaxes for frames too?

lgritz:
When you use frame or view wildcards, for example:

    oiiotool foo.%04d.tif -o bar.%04d.exr

we noticed that if NO files matched, no iteration would occur. You would get this output:

```
$ oiiotool foo.%04d.tif -o bar.%04d.exr
oiiotool WARNING : oiiotool produced no output. Did you forget -o?
```

But maybe this is not helpful and it should be some kind of error to ask for a match and not get one? Or at least give a more clear explanation of what happened.

This PR detects the case where there is no match of the wildcard at all, warns about that, and then tries interpreting the command line literally (no wildcard expansion). So now the output is

```
$ oiiotool foo.%04d.tif -o bar.%04d.exr
oiiotool WARNING : No frame number or views matched the wildcards
oiiotool ERROR: read : File does not exist: "foo.%04d.tif"
Full command line was:
> oiiotool foo.%04d.tif -o bar.%04d.exr
```

Closes #4623

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -7329,6 +7329,12 @@ handle_sequence(Oiiotool& ot, int argc, const char** argv)
         nfilenames = frame_numbers[0].size();
     }
 
+    if (!nfilenames) {
+        // No filenames matched the first wildcard pattern
+        ot.warning("", "No frame number or views matched the wildcards");
+        return false;
+    }
+
     // Make sure frame_numbers[0] has the canonical frame number list
     if (sequence_args.size() && frame_numbers[0].empty())
         frame_numbers[0] = frame_numbers[sequence_args[0]];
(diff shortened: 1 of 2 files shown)


grdanny (approved):
that's a nice change, thank you for that!

PR #4630, opened 2025-02-09
fix(ImageBuf): Fix bug in ImageBuf construction from ptr + neg strides

lgritz:
There was a bug in span_from_buffer when strides were negative (basically, we were subtracting, when adding was correct since those strides were negative values).

Also added a negative stride test to imagebuf_test.cpp. Embarrassed that we didn't already have this, since it would have caught the problem.

### DIFF
--- a/src/libOpenImageIO/imagebuf_test.cpp
+++ b/src/libOpenImageIO/imagebuf_test.cpp
@@ -303,10 +303,12 @@ ImageBuf_test_appbuffer_strided()
 
     // Test negative strides by filling with yellow, backwards
     {
+        ImageBufAlgo::fill(wrapped, cspan<float>(green));
         // Use the ImageBuf constructor from a pointer to the last pixel and
-        // negative strides.
-        ImageBuf neg(ImageSpec(res, res, nchans, TypeFloat),
-                     &mem[res - 1][res - 1][0] /* point to last pixel */,
+        // negative strides. But don't include the edge pixels of the original
+        // buffer.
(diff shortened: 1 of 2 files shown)


src/libOpenImageIO/imagebuf_test.cpp
```diff
@@ -300,6 +300,25 @@ ImageBuf_test_appbuffer_strided()
...
+    {
+        // Use the ImageBuf constructor from a pointer to the last pixel and
+        // negative strides.
+        ImageBuf neg(ImageSpec(res, res, nchans, TypeFloat),
+                     &mem[res - 1][res - 1][0] /* point to last pixel */,
+                     -nchans * sizeof(float) /* negative x stride */,
+                     -res * nchans * sizeof(float) /* negative y stride*/);
+        const float yellow[nchans] = { 1.0f, 1.0f, 0.0f };
+        ImageBufAlgo::fill(neg, cspan<float>(yellow));
+
```
ThiagoIze, line 314: What do you think of adding to the test a guard (say {0.5, 0.5, 0.5}) to check for overflow outside the expected span?

jessey-git (approved):
Confirmed locally in my test app that things are working now. I can't test Blender-proper immediately though but it looks solved from review.

src/libOpenImageIO/imagebuf_test.cpp
lgritz: Good idea. So amended.

ThiagoIze (approved):
Looks good to me

PR #4631, opened 2025-02-11
api: ImageSpec::scanline_bytes, tile_bytes, image_bytes

lgritz:
Existing versions could return the expected bytes for a scanline, tile, or image for either the native per-channel types, or the single composite channel type.

The new flavor lets you pass in any data type and it tells you how many bytes you'd need for a scanline, tile, or image if all channels were the passed type. (As a special case, the answer for TypeUnknown is the "native" size, same as `thing_bytes(true)`).

This is handy for computing how big your buffer needs to be to hold a type-converted scanline or tile.


lgritz:
I will merge this soon if there are no objections.

PR #4634, opened 2025-02-13
fix(wheels): Ensure python-based builds use maj.min.patch SO versioning

### LINKED ISSUE #4621: [BUG] Wheels contain duplicated shared libraries causing bloat
**Describe the bug**

The newly built and released python wheels contain the "symlinked" shared libraries which get replace by full copies, causing bloat.

We fixed that problem in #4428 and it looks like the problem is back. Not sure if it was accidentally re-introduced after my last review of #4428 or after the merge.

We don't need the versioned libraries.

3.0.3.1

**Evidence**

You can see that in effect by using the PyPI Inspector. For example, https://inspector.pypi.io/project/openimageio/3.0.3.1/packages/84/71/6429ab84a3863dadb55a67a70d58d4833c1cabe935a3acbc60710cc3713a/openimageio-3.0.3.1-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl/.

Downloading the wheel and running `unzip -l` shows
```
   721328  2025-02-02 02:52   OpenImageIO/lib/libOpenImageIO_Util.so.3.0.3
 16002984  2025-02-02 02:52   OpenImageIO/lib/libOpenImageIO.so.3.0
 16002984  2025-02-02 02:52   OpenImageIO/lib/libOpenImageIO.so.3.0.3
   721328  2025-02-02 02:52   OpenImageIO/lib/libOpenImageIO_Util.so.3.0
```

zachlewis:
Always use patch-specific SO versioning for scikit-build-core-based builds, even when OpenImageIO_SUPPORTED_RELEASE=ON.

Fixes #4621

## Description

Ensures that Python-based builds only bundle a single copy each of libOpenImageIO and libOpenImageIO_Util shared libraries. 

## Tests

Previously, `$ uv sync -v --reinstall -Ccmake.define.OpenImageIO_SUPPORTED_RELEASE=1` would bundle both .so.3.0 and .so.3.0.3 copies of libraries when installing to a local virtual environment. 

With this fix, only .so.3.0.3 copies of libraries are installed.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -617,7 +617,7 @@ set (EXTRA_DSO_LINK_ARGS "" CACHE STRING "Extra command line definitions when bu
 ###########################################################################
 # Set the versioning for shared libraries.
 #
-if (${PROJECT_NAME}_SUPPORTED_RELEASE)
+if (${PROJECT_NAME}_SUPPORTED_RELEASE AND NOT SKBUILD)
     # Supported releases guarantee ABI back-compatibility within the release
     # family, so SO versioning is major.minor.
     set (SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}


lgritz (approved):
LGTM

PR #4635, opened 2025-02-13
ci: Move away from soon-to-be-deprecated ubuntu-20.04 GHA runner

lgritz:


lgritz:
Withdrawing, I don't have this ready yet, didn't mean to send

PR #4636, opened 2025-02-14
ci: Move away from soon-to-be-deprecated ubuntu-20.04 GHA runner

lgritz:
In the process, port a more modern build_llvm.bash from osl.


lgritz:
Any objections? This is a CI-only change.

PR #4638, opened 2025-02-14
ci(wheels): Add `numpy` as a runtime requirement

### LINKED ISSUE #4637: [BUILD] Optional numpy dependency missing on pyproject.toml
When installing the Python OpenImageIO bindings there isn't a way to install optional dependencies. 

There is a reference to the optional numpy dependency for Python bindings on the install documentation, but there don't seem to be any optional dependencies specified in `pyproject.toml`.

Would it be possible to add numpy as an optional dependency?

zachlewis:
Adds "numpy" as a runtime requirement for the Python wheels.

Fixes #4637

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,9 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 requires-python = ">= 3.8"
+dependencies = [
+    "numpy>=1.19",
+]
 
 [project.urls]
 Homepage = "https://openimageio.org/"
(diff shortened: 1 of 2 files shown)


lgritz:
You have some CI failures.

zachlewis:
Not really sure what inspired the CI failures -- the logs suggest that one of the runs terminated very, very early, which caused a parallel run to fail (also very, very early).... but nothing beyond that. I modified the init.py file a tiny bit to re-kick everything, and everything seems to be passing. Not clear on what to do to prevent this from happening in the future.

lgritz (approved):

PR #4640, opened 2025-02-15
fix(sysutil): improve the implementation of Sysutil::put_in_background on macOS

mikaelsundell:
## Description

Replaces the system() call with posix_spawn() on macOS to preserve environment variables. This is essential for Debug builds using DYLD_IMAGE_SUFFIX=_debug to prevent Qt from loading mixed configurations.

## Tests

Run the iv application in background mode on macOS without the -F flag, using a Debug configuration. Previously, the system() call did not inherit the parent environment variables, causing a mix of Debug and Release versions of Qt to be loaded.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -75,11 +75,10 @@ OIIO_UTIL_API void
 usleep(unsigned long useconds);
 
 /// Try to put the process into the background so it doesn't continue to
-/// tie up any shell that it was launched from.  The arguments are the
-/// argc/argv that describe the program and its command line arguments.
+/// tie up any shell that it was launched from.
 /// Return true if successful, false if it was unable to do so.
 OIIO_UTIL_API bool
-put_in_background(int argc, char* argv[]);
+put_in_background();
 
 /// Number of virtual cores available on this platform (including
 /// hyperthreads). Note that this is just a wrapper/synonym for C++
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -74,10 +74,18 @@ getenv(string_view name, string_view defaultval = {});
 OIIO_UTIL_API void
 usleep(unsigned long useconds);
 
+/// Puts the process into the background, detaching it from the shell
+/// to prevent it from occupying the terminal or blocking the shell
+/// it was launched from. This allows the process to continue running
+/// independently in the background.
+OIIO_UTIL_API bool
+put_in_background();
+
 /// Try to put the process into the background so it doesn't continue to
 /// tie up any shell that it was launched from.  The arguments are the
 /// argc/argv that describe the program and its command line arguments.
 /// Return true if successful, false if it was unable to do so.
+/// DEPRECATED(3.0) old API.
 OIIO_UTIL_API bool
 put_in_background(int argc, char* argv[]);
 
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -75,10 +75,11 @@ OIIO_UTIL_API void
 usleep(unsigned long useconds);
 
 /// Try to put the process into the background so it doesn't continue to
-/// tie up any shell that it was launched from.
+/// tie up any shell that it was launched from.  The arguments are the
+/// argc/argv that describe the program and its command line arguments.
 /// Return true if successful, false if it was unable to do so.
 OIIO_UTIL_API bool
-put_in_background();
+put_in_background(int argc, char* argv[]);
 
 /// Number of virtual cores available on this platform (including
 /// hyperthreads). Note that this is just a wrapper/synonym for C++
--- a/src/include/OpenImageIO/sysutil.h
+++ b/src/include/OpenImageIO/sysutil.h
@@ -78,13 +78,13 @@ usleep(unsigned long useconds);
 /// to prevent it from occupying the terminal or blocking the shell
 /// it was launched from. This allows the process to continue running
 /// independently in the background.
+/// Return true if successful, false if it was unable to do so.
 OIIO_UTIL_API bool
 put_in_background();
 
-/// Try to put the process into the background so it doesn't continue to
-/// tie up any shell that it was launched from.  The arguments are the
-/// argc/argv that describe the program and its command line arguments.
-/// Return true if successful, false if it was unable to do so.
+/// Obsolete version. The argc, argv parameters are not used. We suggest
+/// switching to the version of `put_in_background()` that takes no
+/// arguments.
 /// DEPRECATED(3.0) old API.
 OIIO_UTIL_API bool
 put_in_background(int argc, char* argv[]);
--- a/src/libutil/sysutil.cpp
+++ b/src/libutil/sysutil.cpp
@@ -565,11 +565,7 @@ Sysutil::put_in_background()
 
 
 bool
-#ifdef _WIN32
-Sysutil::put_in_background(int, char*[])
-#else
 Sysutil::put_in_background(int argc, char* argv[])
-#endif
 {
     return put_in_background();
 }
--- a/src/libutil/sysutil.cpp
+++ b/src/libutil/sysutil.cpp
@@ -529,9 +529,9 @@ Term::ansi_bgcolor(int r, int g, int b)
 
 bool
 #ifdef _WIN32
-Sysutil::put_in_background()
+Sysutil::put_in_background(int, char*[])
 #else
-Sysutil::put_in_background()
+Sysutil::put_in_background(int argc, char* argv[])
 #endif
 {
 #if defined(__linux__) || defined(__GLIBC__)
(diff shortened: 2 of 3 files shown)


mikaelsundell:
@lgritz Why is this function located in libutil when it assumes -F on the command line? It mixes application-specific arguments (to prevent an infinite loop) with system-level process spawning. Should we consider moving it to the application itself or finding a better way to eliminate the -F dependency?

lgritz:
> @lgritz Why is this function located in libutil when it assumes -F on the command line? It mixes application-specific arguments (to prevent an infinite loop) with system-level process spawning. Should we consider moving it to the application itself or finding a better way to eliminate the -F dependency?

Such an interesting question!

Looking at the last PR to touch that code, which was in 2012 (!), it used to live in ivmain.cpp, and moved to libutil, probably because I imagined that putting the process of a CLI utility in the background sounded like something I might want to use elsewhere, so elevated it to sysutil.h / libutil. (It's not used anywhere else in OIIO, but it is used in OSL! That only dates from 7 years ago, so it's not the *reason* I moved the OIIO code around.)

Anyway, what I suspect happened is just that I decided to move the function from inside iv to libutil so it could be more widely used, and at that time had forgotten and didn't notice that inside it, it was messing with command line arguments in a way that was iv specific. Maybe this is because it's not necessary for Linux, which simply puts the current process in the background, whereas I couldn't find a reliable way to do it on Apple without re-launching (for which I needed the command line args, and to modify them to instruct it to run in the foreground).

If there's a better way to put the current process in the background on Apple that doesn't need to shell out or care about command line args, we should switch to that. In its absence, maybe we should modify put_in_background to *assume* that the command line arguments have already been altered to indicate foreground use before being passed, so that put_in_background doesn't need to guess that `-F` will do it? Or, alternately, maybe the "put in foreground" command should be passed in as an argument, for applications that want to use this library function but don't use `-F` for the purpose?

mikaelsundell:
Let me prepare a modified version. I was also considering passing arguments directly to remove the iv dependency from the function. Since put_in_background doesn’t seem to be used elsewhere, it shouldn’t be an issue unless it's widely relied upon by applications using the API.

I’m out for now but will be back later tonight. Thanks for the quick response!

mikaelsundell:
This commit will break the ABI as I changed the argc/ argv parameters of put_in_background.
https://github.com/mikaelsundell/oiio/commit/a92ff5b2c0e6808407eb8aef934bc9cd7d39da0e

Updates:
Eliminated explicit argc and argv parameters by using _NSGetArgc() and _NSGetArgv().
Added a check for parent process ID (getppid()) to prevent infinite respawning.

lgritz:
> This commit will break the ABI as I changed the argc/ argv parameters of put_in_background. [mikaelsundell@a92ff5b](https://github.com/mikaelsundell/oiio/commit/a92ff5b2c0e6808407eb8aef934bc9cd7d39da0e)
> 
> Updates: Eliminated explicit argc and argv parameters by using _NSGetArgc() and _NSGetArgv(). Added a check for parent process ID (getppid()) to prevent infinite respawning.

This doesn't just change the ABI, it breaks the API, and we're not allowed to do that until OIIO 4.0, which hopefully won't be for another 5 years. You can give the arguments defaults so that people don't have to pass them, or you can make a second version of put_in_background that doesn't take arguments, but you can't eliminate the version that accepts them; we have to keep the existing call to preserve source compatibility for downstream users.

Anybody using this already passes the arguments, you can probably ignore that part unless you think it's really important to eventually migrate away from it. The important part here is using getppid to prevent recursive spawning of processes rather than needing to alter the command line arguments (which we were doing in a way that assumed the particular command line arguments that iv uses, that was the real design bug).

mikaelsundell:
Sorry, just for testing purposes only—never meant to break ABI.

Anyway, after some thought, should we deprecate the old function and add a second one with no arguments, then update ivmain.cpp to use that? The argc, argv parameters only apply to Mac configurations, so it would make sense to introduce a new function that puts the process into daemon mode without requiring arguments. That way, we have consistency across Mac and Linux, while Windows behaves differently and builds with SUBSYSTEM:WINDOWS?

src/libutil/sysutil.cpp
```diff
@@ -532,37 +534,25 @@ Sysutil::put_in_background(int, char*[])
...
 
 #elif defined(__APPLE__) && TARGET_OS_OSX
-    std::string newcmd = std::string(argv[0]) + " -F";
-    for (int i = 1; i < argc; ++i) {
-        newcmd += " \"";
-        newcmd += argv[i];
-        newcmd += "\"";
+    if (getppid() == 1) {
+        return true;
     }
```
lgritz, line 545: Could you please add a comment here explaining how this early return based on the parent process ID is the reason you don't end up recursively spawning new iv's?

lgritz (approved):
LGTM, though I would recommend a comment explaining how the recursive launching of iv copies is avoided.

src/include/OpenImageIO/sysutil.h
```diff
@@ -74,10 +74,18 @@ getenv(string_view name, string_view defaultval = {});
 OIIO_UTIL_API void
 usleep(unsigned long useconds);
 
+/// Puts the process into the background, detaching it from the shell
+/// to prevent it from occupying the terminal or blocking the shell
+/// it was launched from. This allows the process to continue running
+/// independently in the background.
```
lgritz, line 80:
```suggestion
/// independently in the background.
/// Return true if successful, false if it was unable to do so.
```

src/include/OpenImageIO/sysutil.h
```diff
@@ -74,10 +74,18 @@ getenv(string_view name, string_view defaultval = {});
...
+OIIO_UTIL_API bool
+put_in_background();
+
 /// Try to put the process into the background so it doesn't continue to
 /// tie up any shell that it was launched from.  The arguments are the
 /// argc/argv that describe the program and its command line arguments.
 /// Return true if successful, false if it was unable to do so.
```
lgritz, line 87:
```suggestion
/// Obsolete version. The argc, argv parameters are not used. We suggest
/// switching to the version of `put_in_background()` that takes no
/// arguments.
```

src/libutil/sysutil.cpp
```diff
@@ -575,6 +564,17 @@ Sysutil::put_in_background(int argc, char* argv[])
 }
 
 
+bool
+#ifdef _WIN32
+Sysutil::put_in_background(int, char*[])
+#else
+Sysutil::put_in_background(int argc, char* argv[])
+#endif
```
lgritz, line 572: I'm not sure this is necessary. Did you get warnings on non-Windows platforms if you just stuck to `(int, char*[])` across the board?

lgritz (changes requested):
Minor suggestions for the documentation comments. Otherwise, looks great.

src/libutil/sysutil.cpp
mikaelsundell: Running CI tests, removed legacy WIN32 arguments inherited from master.

mikaelsundell:
All checks passed – ready to go!

lgritz (approved):
LGTM!

lgritz:
@mikaelsundell Does this actually work for you? I'm finding that on Mac, iv no longer properly puts itself in the background since this PR was integrated.

lgritz:
Weird! It does not do that for me -- it seems to remain fully in the foreground (it used to work). I'm running Ventura 13.7.3. You?

I'm ok with it sending output to the terminal while in background (which I think only happens for debug builds, so not worried about it).

mikaelsundell:
What macOS version are you using? I've got a pile of old Macs here—I can probably match it and fix the issue. Feels like something I should take care of.

lgritz:
I'm running Ventura 13.7.3

mikaelsundell:
Intel or Silicon?

lgritz:
Intel

mikaelsundell:
Great, this is a case for my old friend in the garage!

lgritz:
If you can't reproduce it on a machine with that CPU and OS version, don't feel the need to chase it any farther than that. Nobody has complained, and I can learn to type `&`. We can sweep it under the rug until somebody other than me actually encounters it and cares.

mikaelsundell:
Let me give it a try tomorrow, no problem.

mikaelsundell:
Good news, I have a matching Ventura/ Intel Mac. Rebuilding all dependencies and doing a test after that.

mikaelsundell:
https://github.com/user-attachments/assets/1092731f-a657-4ab8-8042-b3ca1cc825ea

Hi, this is not what you see on your Intel Mac? Keep in mind we're still redirecting the output to the Terminal even if it's put in background (can be changed of course).

lgritz:
> @lgritz this is not what you see on your Intel Mac? Keep in mind we're still redirecting the output to the Terminal even if it's put in background (can be changed of course).

No, I don't get a command prompt back at all until I exit iv.

PR #4642, opened 2025-02-19
feat: add self-build functionality for OpenJPEG

InisPromise:
## Description

This is my first contribution! I attempted to implement a self-build functionality for the OpenJPEG dependency. I looked at cmake documentation, specifically "Find Module" and followed OIIO CONTRIBUTING guide as much as I could. I am just getting back into coding and also have not worked too much with cmake before, so I am very open to criticism and tips! 

##Image of Changes 
image

## Checklist:

- [ x] I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
- [ x] If I added or modified a C++ API call, I have also amended the
  nonconforming.


lgritz:
Hi, welcome!

There are a number of oddities in your PR. Most of the changes appear to be additions that seem entirely irrelevant. Maybe you accidentally checked in a bunch of files related to your editor and that are not meant to be in the code base?

But the parts that do seem like additions to the cmake files, I don't think that they will work to address the problem. All it does is search for OpenJPEG. In the cmake file we have, externaldependencies.cmake, we already do this [here](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/cmake/externalpackages.cmake#L171)

The issue of "adding self-build functionality" is about what happens when that search fails, because OpenJPEG (or an adequately new version of it) cannot be found. In that case, we want it to do a download and build on its own. This is handled by the various build_xxx.cmake files, for example you can see how we do it for libtiff [here](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/cmake/build_TIFF.cmake).

I'm particularly curious about this line

    find_package(OpenJPEG REQUIRED)
    target_link_libraries(build PRIVATE ${OpenJPEG_LIBRARIES})    <----------
    include_directories(${OpenJPEG}_INCLUDE_DIRS)

because "build" is not a target name anywhere in our library, so if you actually try to build the project with this code in place, you should get this error message:

    CMake Error at CMakeLists.txt:226 (target_link_libraries):
      Cannot specify link libraries for target "build" which is not built by this
      project.

I'm really curious how you came up with this solution and how you tested it. Did you not observe this error?

InisPromise:
Hello! 

I'm so sorry for all the issues. First, I'm not sure how the additional files got in there? I think your guess is probably correct and I did accidentally check those in but I did not mean to and don't know how that occurred. As for the cmake file addition, I used the cmake documentation (the find modules section, since OIIO has a FindOpenJPEG.cmake file in cmake/modules) to come up with the syntax, but could not find many more resources. This is my first time working with such a big cmake file and I was really in over my head (as I'm sure you have gathered). I think I also must've pushed the wrong version, because when I ran it (which, at this point, I don't know if I even did that correctly), I didn't receive any errors. I think the target is supposed to be "make build" in that line (according to the Makefile "help" message) but I see now that even if there were no errors with my code, it would not have mattered much anyway given my misinterpretation of how to resolve the self-building issue.  

If you have the time, could you point me to a few resources to learn how to implement this change?  I want to learn very badly, I'm just not sure how to tackle it. 

Thank you!

lgritz:
> If you have the time, could you point me to a few resources to learn how to implement this change? I want to learn very badly, I'm just not sure how to tackle it.

Well, any of the existing `src/cmake/build_*.cmake` files are examples of how we do the auto-build for various other packages. We just don't have one for OpenJPEG yet.

The test to validate that you did it correctly involves building OIIO when there is NO installation of OpenJPEG, and verifying that it build and uses it even though it can't already be found on the system.

lgritz:
This PR has sat idle for several months and still doesn't actually implement the builder module that's necessary for it to work.

@InisPromise Do you intend to return to this work, or should we close the PR?

lgritz:
This PR as presented does not solve the issue and there has been no response or update from the author for months. Closing.

PR #4644, opened 2025-02-19
ICC profile reading improvements, especially for PSD

lgritz:
The main thing here is that it turns out that when trying to parse the ICC profiles from Photoshop PSD/PSB files, we weren't storing the attributes in the correct way. Everything else, we shoved in both the "common" and "composite" imagespecs, but we left ICC data out of the composite spec.

While I'm at it, I made two other ICC related changes:

* Among the several formats with ICC profiles, we were inconsistent about whether we report errors for corrupt ICC profiles and treat the whole image as broken, or just ignore the errors entirely. Fixed in various places to make a consistent policy of always checking the errors, and reporting/terminating if the "imageinput:strict" global attribute is true, allowing it to just ignore the broken ICC record if that attribute is false.

* Found a spot in JPEG handling of ICC where we were dangerously trusting of the sizes of things. Replace a risky memcpy with a safer spancpy that can't overrun the buffer bounds when copying.

### DIFF
--- a/src/jpeg.imageio/jpeginput.cpp
+++ b/src/jpeg.imageio/jpeginput.cpp
@@ -427,8 +427,13 @@ JpgInput::read_icc_profile(j_decompress_ptr cinfo, ImageSpec& spec)
         if (m->marker == (JPEG_APP0 + 2)
             && !strcmp((const char*)m->data, "ICC_PROFILE")) {
             int seq_no = GETJOCTET(m->data[12]);
-            memcpy(icc_buf.data() + data_offset[seq_no],
-                   m->data + ICC_HEADER_SIZE, data_length[seq_no]);
+            if (data_offset[seq_no] + data_length[seq_no] > icc_buf.size()) {
+                errorfmt("Possible corrupt file, invalid ICC profile\n");
+                return false;
+            }
+            spancpy(make_span(icc_buf), data_offset[seq_no],
+                    make_cspan(m->data + ICC_HEADER_SIZE, data_length[seq_no]),
+                    0, data_length[seq_no]);
         }
     }
     spec.attribute("ICCProfile", TypeDesc(TypeDesc::UINT8, total_length),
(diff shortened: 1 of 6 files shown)


lgritz:
@zachlewis That would be a really helpful addition, and it is one beyond my very thin knowledge of ICC and OCIO. Honestly, I'm just a plumber who can, upon request, connect an OIIO API call to an obvious OCIO one, but I do so with minimal understanding about how it works (beyond what you pick up by osmosis with a career in VFX studios). I'm regularly surprised to discover the existence of features that have been in OCIO for years without my awareness. I'm sure that somebody well versed in the ICC support in OCIO and how it should be used could rig up a half dozen extremely useful OIIO features in a matter of days, if not hours.

lgritz:
Anybody object to the changes in this PR?

lgritz:
Yes, all things being equal, I'd prefer to just outsource the problem to OCIO and assume that there's a future OCIO API call that directly does what we need. OIIO's pieces of the puzzle should really be limited to:
- knowing how to read/write the ICC profile for each type of image file
- exposing/wrapping the OCIO functionality via an OIIO API call and an oiiotool action/cli argument (much as we do for other important OCIO functionality)

lgritz:
> (No objections from me, although I honestly haven't really reviewed this PR)

My theory of code review is that most of the value is just getting a second pair of eyes on it so that we are truly minimizing the number of things directly committed without any 2nd party oversight. High priority items are:

* Does the approach/design -- revealed mostly by the commit comment -- seem like the right way to solve the problem and the right direction for the project, and does it appear to correspond to what the code seems to be doing at a high level.
* Is there anything that appears extraneous or malicious sneaking in that seems to not be a legit part of the PR's purpose.
* Is the functionality tested somehow, if not plainly obvious.
* Are any public API changes getting the scrutiny they deserve, and are any backward compatibility breaks or behavior changes appropriate for the current project stage or the likely release branch backports we'll do of this PR.

All of that can usually be done with a cursory look by anybody who is a competent programmer even if not previously familiar with that section of the code. A deep line-by-line expert understanding of exactly how it works (to the degree that you could identify any bugs just by reading the code) would be ideal, but realistically it is rare to achieve in practice, and that's ok because it's only a tiny portion of the code review value.

lgritz:
> (Although if this is exhaustive of all the plugins that currently implement ICC support, I guess it's worth at least noting somewhere that HEIF, AVIF, and JPEG XL all permit embedded ICC profiles as well...!)

Good point! I just spot-checked, and currently our heif and ffmpeg(avif) readers do not consider icc at all, and jpegxl reads it from the file then but does not add it to the ImageSpec. :-) So yeah, let's remember that there are other ICC-related touch-ups that somebody should do (ideally, the people who are most active in each of those format reader implementations).

PR #4645, opened 2025-02-20
admin: Add ".vs" to .gitignore

lgritz:
Prevents people from accidentally committing their Visual Studio configuration files.


lgritz:
No objections + doesn't change code -> merge

PR #4646, opened 2025-02-22
ci: for docs workflow, lock down versions and speed up

lgritz:
When test building the docs, lock down breathe and furo versions to exactly what we use on RTD.

Also, to speed up the workflow: run on native Ubuntu rather than on an ASWF container (eliminates needless container download), and make sure that we don't waste time installing dependencies that we don't need.

These changes speed up the docs workflow from around 3:40 to around 2:00.


lgritz:
Any objections?

lgritz:
No objections -> merging

PR #4648, opened 2025-02-24
admin: Set up .gitattributes file

lgritz:
Starter purpose is giving hints to GitHub's language analysis.


jessey-git (approved):
Seems reasonable.  So very odd that of all the languages that have `.inc`, we ended up with "Pawn" :)

PR #4653, opened 2025-03-04
deps: Add build recipe for GIF

hisergiorojas:
## Description
As requested in https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4387.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/build_gif.cmake
+++ b/src/cmake/build_gif.cmake
@@ -2,42 +2,139 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
-######################################################################
-# GIF by hand!
-######################################################################
+###########################################################################
+# Build GIF library locally
+#
+# This script downloads and builds the GIF library from source since GIF
+# doesn't provide a CMake-based build system like other dependencies.
+###########################################################################
 
-set_cache (GIF_BUILD_VERSION 5.2.1 "GIFLIB version for local builds")
-set (GIF_URL "https://downloads.sourceforge.net/project/giflib/giflib-${GIF_BUILD_VERSION}.tar.gz")
-set (GIF_URL_HASH SHA256=31da5562f44c5f15d63340a09a4fd62b48c45620cd302f77a6d9acf0077879bd)
+# Set version and source information
+set_cache(GIF_BUILD_VERSION 5.2.1 "GIFLIB version for local builds")
+set(GIF_URL "https://downloads.sourceforge.net/project/giflib/giflib-${GIF_BUILD_VERSION}.tar.gz")
+set(GIF_URL_HASH SHA256=31da5562f44c5f15d63340a09a4fd62b48c45620cd302f77a6d9acf0077879bd)
 
-set_cache (GIF_BUILD_SHARED_LIBS ${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
-           DOC "Should execute a local GIF build, if necessary, build shared libraries" ADVANCED)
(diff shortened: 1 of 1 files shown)


lgritz:
How did you test this and verify that it works?

hisergiorojas:
I'm still working on the pull request - 

Here's what I'm currently doing to test and verify the CMake files:
1. I've turned on the "ignore homebrew-installed dependencies" option in the CMakeLists.txt file.
2. Then I'm building with these commands:
```
mkdir build
cd build
cmake .. -DOpenImageIO_BUILD_MISSING_DEPS=all

```

This makes CMake to build all dependencies locally instead of using system-installed versions.

hisergiorojas:
Screenshot 2025-03-05 at 11 00 27 PM

hisergiorojas:
Screenshot 2025-03-05 at 11 01 04 PM

hisergiorojas:
Another way to test and verify: 

1. To build oiiotool specifically, you can continue the build process with:

`make -j8   # Adjust based on the number of CPU cores`

2. Once the build completes, you can verify GIF support by:

  a. Converting an image to GIF format:
`./bin/oiiotool path/to/some/image.jpg -o test_output.gif`

b. You can also try to convert a GIF to another format to further verify that it's working:
`./bin/oiiotool ~/Downloads/earth.gif -o earth.jpg`

src/cmake/build_gif.cmake
```diff
@@ -0,0 +1,43 @@
...
+    VERSION         ${GIF_BUILD_VERSION}
+    URL             ${GIF_URL}
+    URL_HASH        ${GIF_URL_HASH}
+    CMAKE_ARGS
+        -D CMAKE_INSTALL_PREFIX=${GIF_LOCAL_INSTALL_DIR}
+        -D BUILD_SHARED_LIBS=${GIF_BUILD_SHARED_LIBS}
+        -D CMAKE_INSTALL_LIBDIR=lib
+    PATCH_COMMAND
+        curl -L https://sourceforge.net/p/giflib/bugs/_discuss/thread/4e811ad29b/c323/attachment/Makefile.patch -o Makefile.patch &&
+        patch -p0 < Makefile.patch
```
lgritz, line 29: Should we be bothered by the fact that the `build_dependency_with_cmake` function doesn't take a "PATCH_COMMAND" argument, and so this extra patch won't be applied?

lgritz: Also, build_dependency_with_cmake doesn't take a URL or URL_HASH argument, and in fact the only kind of download it knows how to do is to clone a git repo? I just don't understand how the code as presented here can work. 

Are you sure you aren't locally running and testing a version of this patch that is different from the one currently in this PR?

This hypothesis is also supported by the fact that the screenshots you attached have messages about FetchContent, which we do not call anywhere and also aren't mentioned in this patch.

hisergiorojas: just realized I forgot to push my latest change where I removed the build_dependency_with_cmake method since it only accepts a git repository

lgritz:
This PR has been idling for a while, I think because it never quite passed CI. Is this something you're going to return to, @hisergiorojas?  Maybe we should start by rebasing it on top of the current main and see what, if anything, is still broken?

lgritz:
Closing. This never fully passed CI, the PR has sat idle for months without the author fixing it, and in the mean time, somebody else has fully implemented it in #4921. Sincere thanks for trying to implement this; it's ok, not all PRs work out. The effort is appreciated.

PR #4655, opened 2025-03-08
fix(png): increase allowed width/height limit

jessey-git:
## Description

The current width/height limit enforced by the `check_open` API is too low for PNG files. Each dimension can technically be up to `2^31 - 1` but using that value would defeat the purpose of the check here. Use `1 << 20` like what is used for TIFF and EXR instead.

The situation was triggered while trying to write an 85000x28400 px image, which is smaller than 65536*65536 but the dimensions are checked individually rather than as a total pixel count or similar.

Also looks like the PNG output was allowing too many channels so fix that too.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (approved):
LGTM, thanks for the fix!

PR #4657, opened 2025-03-08
build: Better disabling of work when USE_PYTHON=0

lgritz:
When USE_PYTHON is disabled, don't even bother looking for python.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -113,7 +113,9 @@ else ()
 endif()
 
 # From pythonutils.cmake
-find_python()
+if (USE_PYTHON)
+    find_python()
+endif ()
 if (USE_PYTHON)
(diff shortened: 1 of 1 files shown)


src/cmake/externalpackages.cmake
```diff
@@ -113,7 +113,9 @@ else ()
 endif()
 
 # From pythonutils.cmake
-find_python()
+if (USE_PYTHON)
```
jessey-git, line 116: Is there any particular reason not to use the existing `if (USE_PYTHON)` conditional 3 lines below this?

lgritz: Yes, because find_python can turn off USE_PYTHON if not found.

jessey-git: Ah, sneaky, I see.

jessey-git (approved):

PR #4658, opened 2025-03-09
ci: step naming adustments

lgritz:
* Give names to some CI steps that didn't have them -- makes it nicer to read the logs and CI status dashboard.
* Make sure no job nametag is a substring of another, or it might find the wrong cache, which is based on the job name.


lgritz:
No objections + CI output change only = merge

PR #4659, opened 2025-03-10
ci: Update libPNG address and version for ci & autobuild

lgritz:
* The GitHub account has changed (the old one redirects, but still, let's update to the current canonical name).
* Update the version we use in CI and for "autobuild" to the latest tagged release instead of one that's years old.


lgritz:
No objections, only affects CI -> merging

PR #4663, opened 2025-03-11
Bug fix: Updated tags for psb

lydia-zheng:
## Description

There were some psb files that were not opening with oiiotool, failing on the signature.
Some keys were missing for the additional layer information, leading to the pointer to misalign when reading the file, expecting the next 4 bytes rather than 8 bytes where needed for parsing.

## Tests

No additional test suites added.

After deliberating, it is unclear if we should have a strict mode, or keep parsing the rest of the file after a failed signature, given that in this case the signature is not the part that's actually causing the issue. Might be useful as a future feature, if there is an use case for that?

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -509,7 +509,7 @@ const PSDInput::ResourceLoader PSDInput::resource_loaders[]
 
 const char* PSDInput::additional_info_psb[]
     = { "LMsk", "Lr16", "Lr32", "Layr", "Mt16", "Mt32", "Mtrn",
-        "Alph", "FMsk", "Ink2", "FEid", "FXid", "PxSD", "cinf" };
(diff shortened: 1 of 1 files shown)


lgritz:
Wow, that is not at all what I thought the problem was!

So this fixes the psd files you were having trouble reading?

lydia-zheng:
> Wow, that is not at all what I thought the problem was!
> 
> So this fixes the psd files you were having trouble reading?

Me neither at first!
Yes, this fixes it! Was testing to see the keys, and looks like at least with the 60 or so files that Maxwell flagged, those were no longer erroring out with oiiotool —info once the new keys were added. This also fixes the previous typo with INK2 as LNK2.

There’s not any explicit documentation from adobe file format on lnkE, but is supposedly a form of linked format too when referencing some other open source PSD parsers for keys.

Also pths was not parsing correctly for the elements after, so added it as well to the list. Tested that after adding pths, the following cinf keys parsed in the psbs.

lgritz (approved):
LGTM, great find!

PR #4664, opened 2025-03-14
imagecache: remove redundant ImageSpec from ImageCache internals

bfraboni:
First draft of backend changes for issue https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4436.

## Description

- add `LevelSpec` struct to store only a subset of `ImageSpec` descriptors, which can be different per mip level.
   - currently stores all fields that could differ, we might want a more compact version ?
- add member `std::shared_ptr<ImageSpec> m_spec` in `SubimageInfo` to store the descriptors that are common across all mip level
- add member `std::shared_ptr<ImageSpec> m_spec` in `LevelInfo` to link to the associated subimage spec
- add member`std::unique_ptr<LevelSpec> m_spec` in `LevelInfo`, allocated only if some fields differ from `m_spec`
- add accessors to `LevelInfo` internals, such that it can be used in `ImageCache` and ` TextureSystem`

## Checklist:

  doesn't change the API.)

### DIFF
--- a/src/libtexture/imagecache_pvt.h
+++ b/src/libtexture/imagecache_pvt.h
@@ -177,18 +177,22 @@ public:
     {
         return (int)m_subimages[subimage].levels.size();
     }
-    const ImageSpec& spec(int subimage, int miplevel) const
+
+    void get_level_dimensions(int subimage, int miplevel, ImageSpec& spec) const
     {
-        return levelinfo(subimage, miplevel).spec();
+        levelinfo(subimage, miplevel).get_level_dimensions(spec);
     }
-    ImageSpec& spec(int subimage, int miplevel)
+
+    ImageSpec& get_subimage_spec(int subimage)
     {
-        return levelinfo(subimage, miplevel).spec();
+        return subimageinfo(subimage).get_subimage_spec();
     }
-    const ImageSpec& nativespec(int subimage, int miplevel) const
+
+    const ImageSpec& get_subimage_spec(int subimage) const
     {
-        return levelinfo(subimage, miplevel).nativespec;
+        return subimageinfo(subimage).get_subimage_spec();
     }
+
     ustring filename(void) const { return m_filename; }
     ustring fileformat(void) const { return m_fileformat; }
     TexFormat textureformat() const { return m_texformat; }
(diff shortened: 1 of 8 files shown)


bfraboni:
Hey @lgritz ,

I'm almost done with this PR, apologies for the massive changes, but I've reworked a bit the ownership of the ImageSpec and LevelSpec.
 
I still have one unknown right now, keeping or getting rid of the compressed LevelSpec, which uses a bitfield and an allocation instead of 15 integers for the non compressed counterpart. I'm still not convinced it is worth it, it is slogjtly slower and the ImageSpec/LevelSpec reuse across subimages and miplevels saves much more memory. Also getting rid of it will simplify the complicated getter/setter logic a fair bit.

Furthermore, the LevelInfo footprint is at its minimum so I can't really compress further and that's what dominates the current version; shared ImageSpec / LevelSpec takes few tens of mb while few millions of SubimageInfo / LevelInfo takes few hundreds of mb now.

I still have to rebase, sign off all commits (I missed one), move functions definitions from imagecache_pvt.h to .cpp but I think that gets closer.

If you have some time, could you have a quick look at the code and let me know what you think ?
Thank you 🙏

bfraboni:
@lgritz I've updated the comment with what's in the PR, the tests are all green, the history is not clean but I wait for your input and the list of  required changes before squashing all that together.
Should I do anything else before review ? (I can't really ask on github, the area for review is grayed)

bfraboni:
I just cleaned up the commit history, kept only the major changes between versions, this reads better.

src/include/OpenImageIO/memory.h
```diff
@@ -48,7 +48,7 @@ template<typename T>
 inline size_t
 footprint(const T* t)
 {
-    return sizeof(T) + (t ? footprint(*t) : 0);
+    return sizeof(T*) + (t ? footprint(*t) : 0);
 }
```
bfraboni, line 52: Fix footprint evaluation for pointers

src/include/OpenImageIO/memory.h
```diff
@@ -111,6 +101,18 @@ footprint(const std::unique_ptr<T>& ref)
...
+inline size_t
+heapsize(const std::vector<T>& vec)
+{
+    size_t size = 0;
+    for (const T& elem : vec)
+        size += footprint(elem);
+    return size;
+}
+
+
```
bfraboni, line 115: Fix heapsize evaluation for `std::vector<std::shared_ptr<T>>` ; before the heapsize specialization for shared_ptr wouldn't be used.

src/libtexture/texture3d.cpp
bfraboni, line 1: Lot of local variable renaming, to plumb the new usage to access miplevel descriptors:
```c++
const SubimageInfo& si(texfile->subimageinfo(subimage));
const Dimensions& dims(si.dimensions(miplevel));
// do things with DImensions
int w = dims.width;
```

src/libtexture/texturesys.cpp
bfraboni, line 1: Lot of local variable renaming, to plumb the new usage to access miplevel descriptors:
```c++
const SubimageInfo& si(texfile->subimageinfo(subimage));
const Dimensions& dims(si.dimensions(miplevel));
// do things with DImensions
int w = dims.width;
```

src/libtexture/imagecache_pvt.h
```diff
@@ -262,30 +262,63 @@ class OIIO_API ImageCacheFile final : public RefCnt {
...
+        static Dimensions& convert(ImageSpec& s)
+        {
+            // is that evil ?
+            OIIO_STATIC_ASSERT(alignof(ImageSpec) >= alignof(Dimensions));
+            OIIO_DASSERT(sizeof(Dimensions)
+                         == ((size_t)&s.format - (size_t)&s.x));
+            return *((Dimensions*)&s.x);
+        }
+    };
+
```
bfraboni, line 306: New internal structure that can be used standalone for miplevels or as a view onto an existing ImageSpec, using the `Dimensions::convert` method.

src/libtexture/imagecache_pvt.h
```diff
@@ -262,30 +262,63 @@ class OIIO_API ImageCacheFile final : public RefCnt {
...
-        LevelInfo(const ImageSpec& nativespec)
-            : LevelInfo(nullptr, nativespec)
-        {
-        }
+        uint16_t nchannels;             ///< Number of channels in the image
+        bool full_pixel_range : 1;   ///< pixel data window matches image window
+        bool onetile : 1;            ///< Whole level fits on one tile
+        bool polecolorcomputed : 1;  ///< Pole color was computed
+
+        LevelInfo(ImageSpec* spec, Dimensions* dims = nullptr);
```
bfraboni, line 319: I make sure LevelInfo is as small as possible to reduce memory usage, by removing `m_spec` and `nativespec` and compressing the newly added nchannels with the booleans in a bitfield.

I also removed the unnecessary `mutable`, because LevelInfo can be accessed by reference where polecolor is computed.

src/libtexture/imagecache_pvt.h
bfraboni: **TODO**: we just need to make sure the static_assert for testing the aligment of ImageSpec and Dimensions are correct.

src/libtexture/imagecache_pvt.h
```diff
@@ -262,30 +262,101 @@ class OIIO_API ImageCacheFile final : public RefCnt {
...
+        int full_depth;   ///< depth of the full (display) window
+        int tile_width;   ///< tile width (0 for a non-tiled image)
+        int tile_height;  ///< tile height (0 for a non-tiled image)
+        int tile_depth;   ///< tile depth (0 for a non-tiled image,
+                          ///<             1 for a non-volume image)
+        int nchannels;    ///< number of image channels, e.g., 4 for RGBA
+
+        static const Dimensions& convert(const ImageSpec& s)
+        {
+            validate_alignment();
```
lgritz, line 290: I don't think you need to *call* validate_alignment. It just will be a compile-time error for it merely to contain the failing static_assert. 

In fact, I'm not even sure you need it in a static function. I think you can just have the STATIC_ASSERT just "bare" inside the class definition.

src/libtexture/imagecache_pvt.h
```diff
@@ -262,30 +262,101 @@ class OIIO_API ImageCacheFile final : public RefCnt {
     // success, false on failure.
     bool get_average_color(float* avg, int subimage, int chbegin, int chend);
 
+    //! Dimensions is a minified ImageSpec that only store the fields
+    //! that can change per mip map level.
+    //! It is used as standalone container or can be used as a view
+    //! over the same fields of ImageSpec via the Dimensions::convert() function.
+    struct Dimensions {
```
lgritz, line 269: Can I ask you to rename this to ImageDimensions or ImageDims? That seems a little more symmetric to ImageSpec. 

I can see at a future that that maybe ImageSpec will be redefined to inherit from ImageDims (while adding the other metadata). But I don't think we can do it now without possibly breaking ABI.

src/openexr.imageio/exrinput.cpp
```diff
@@ -1077,6 +1077,11 @@ OpenEXRInput::seek_subimage(int subimage, int miplevel)
     m_miplevel = miplevel;
     m_spec     = part.spec;
 
+    //! Add the number of miplevels as an attribute for the first miplevel.
+    //! TOFIX: adding the following attribute breaks unit tests
+    // if (m_miplevel == 0 && part.nmiplevels > 1)
+    //     m_spec.attribute("oiio:miplevels", part.nmiplevels);
+
```
lgritz, line 1084: Does it only break unit tests by printing the "oiio:miplevels" in certain test output and we just need to update that test output reference? Do you need exr files to do this trick?

lgritz:
Sorry it's taken me a while to get to this. My comments are extremely minor, and other than that, this all LGTM. Make those minor changes and rebase to the current main (there seems to be some minor conflict) and then I am ready to merge it.

src/openexr.imageio/exrinput.cpp
bfraboni: It seems to break more than the just output, i.e. for Ptex I just had to update the output text and the test passed again, for Exr it is beyond my understanding so far..

src/libtexture/imagecache_pvt.h
bfraboni: I just renamed to `ImageDims` , thank you for the suggestion

src/libtexture/imagecache_pvt.h
bfraboni: Right, I just needs to put it somewhere the class is complete, I can't just from within the class ImageDims, but I can from within the upper class ImageCacheFile.

PR #4665, opened 2025-03-14
fix(build): Improve OpenJpeg version detection.

jreichel-nvidia:
Also detect patch level version. E.g. useful when checking for included security fixes.

## Tests

I ran `oiiotool --help` manually, and checked that the macro is available already in 2.1.0.

## Checklist:

- [N/A] I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
- [N/A] I have ensured that the change is tested somewhere in the testsuite
- [N/A] If I added or modified a C++ API call, I have also amended the
  nonconforming.

### DIFF
--- a/src/cmake/modules/FindOpenJPEG.cmake
+++ b/src/cmake/modules/FindOpenJPEG.cmake
@@ -83,10 +83,8 @@ if (OPENJPEG_INCLUDE_DIR)
 endif ()
 
 # Search for opj_config.h -- it is only part of OpenJpeg >= 2.0, and will
-# contain symbols OPJ_VERSION_MAJOR and OPJ_VERSION_MINOR. If the file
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM

lgritz:
I'll backport this to the 3.0 release branch as well.

I noticed that our current minimum of OpenJPEG >= 2.0 is quite old (2015!). We "recommend" OpenJPEG >= 2.2, and even that version is from 2017. We can't change minimum requirements in the release branch, but in main (future 3.1), should we raise the OpenJPEG minimum and simplify a bit of this logic where it notes that the availability of `OPJ_VERSION_BUILD` is only in 2.1+?

lgritz:
My notes indicate that 2.2 is when they introduced multithreaded decoding.

PR #4667, opened 2025-03-15
api(span.h): OIIO::span improvements

lgritz:
Additions to `OIIO::span<>`:

* New flavors of `as_bytes()`, `as_writable_bytes()` that take a raw pointer + length, turning them into `cspan<byte>` and `span<byte>`, respectively.
* `span_cast<T>()`: "cast" a span of one type into a span of a different type covering the same memory. Use with caution!

Safety improvements to existing functionality:

* front(), back(): Add a debug assertion to validate that there is at least one element.
* first(), last(), subspan(): clamp offsets and counts so that the result can't possibly return a span referencing memory that wasn't inside the original bounds.
* The copy constructor that takes a non-matching span was much too flexible. Now it uses enable_if to ensure that it can't be used to change data types or lengths, it can only turn span of non-const into span of const, and span of fixed extent into span of dynamic extent.

### DIFF
--- a/src/include/OpenImageIO/span.h
+++ b/src/include/OpenImageIO/span.h
@@ -19,6 +19,13 @@
 #include <OpenImageIO/platform.h>
 #include <OpenImageIO/detail/fmt.h>
 
+// Span notes and helpful links:
+// - cppreference on std::span:
+//   https://en.cppreference.com/w/cpp/container/span
+// - Another implementation, for reference:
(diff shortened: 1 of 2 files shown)


lgritz:
@EmilDohne Some preliminary work to add a few span utilities I needed in the implementation of span-ifying the APIs.

EmilDohne (approved):
LGTM! I like the addition of the std::min() calls in `first()` and `last()` as it's quite easy to forget

PR #4668, opened 2025-03-15
ci: Fix wheel building on Mac

lgritz:
This used to work until a few days ago. Then it stopped. I think that GitHub changed the Mac runner configuration in some way that broke it, now it's complaining about not being able to find ninja, so easily fixed by instructing it to generate regular makefiles instead.

Along for the ride: If ONLY the wheel making workflow changes, there's no need to rerun the regular CI, docs, or analysis workflows.


zachlewis:
Interesting. No objections at all to the stuff that's come along for the ride.

I would like to dive deeper into whatever's causing the problems with ninja, because ninja should be coming from scikit_build_core (i.e. as a dependency needed for python-based builds, same as with cmake itself). This could be indicative of a change introduced by a newer version of scikit_build_core itself, in which case, we might be better served by more tightly constraining the version requirement upper bound. 

(Still, no problem with changing the generator to "Unix Makefiles", that's a clever workaround)

lgritz:
Maybe it was scikit_build_core? In any case, it only failed on Mac Intel.
This seemed to be the expedient fix, but if you'd like to dig into it more, feel free. You can reproduce the failure case just by pushing the current (pre- this PR) main to your fork, using any branch name that contains the word "wheel".

PR #4669, opened 2025-03-17
api: start span-ifying ImageInput

lgritz:
Add new ImageInput varieties of read_native_scanlines, read_native_tiles, and a new read_native_voluemetric_tiles (all-channel and channel-subset varieties of each) that take a `span<byte>` describing the area to deposit the contiguous native-datatype values.

Moving forward, these will be the primary pixel-reading customization points of the format readers. For now, these default to calling the pointer-based varieties (so that old plugins keep working), but eventually, the pointer-based varieties will be deprecated.

Notes:

* Omit the single-scanline and single-tile versions. Just keep the more general multi-scanline, multi-tile versions, which of course people are free to call with ranges that request just one scanline or tile.

* Omit the z coordintes for read_native_scanlines, since we know of no volumetric format that allows reading of individual "scanlines".

* Omit the z coordinates from read_native_tiles to simplify the common case of 2D images, but add a new read_native_volumetric_tiles() to handle the volumetric case separately. This only needs to be supplied for volume formats,

* The default implementations of the new methods is simply to call the old, unsafe, raw pointer based versions. Thus, an ImageInput that still supplies pointer based versions but not the span-based ones will not break. Eventually, we will switch to the other way around, where all ImageInput subclasses are expected to fully implement the span version, and the pointer version will default to just be wrappers around the span version for backward incompatibility. Way, way down the line, we'll probably remove the pointer based varieties entirely.

Note that for this PR, we are only tinkering with the `read_native_` functions that the ImageInput format readers supply. In a later PR, we'll handle all the client-facing API calls to have span-based varieties, as well as the analogous functions on the ImageOutput side.

In this PR, we do the overhaul on the JPEG and TIFF readers, to prove that the whole scheme works. Over time, we will do the similar overhaul of the other format readers.

### DIFF
--- a/src/jpeg.imageio/jpeginput.cpp
+++ b/src/jpeg.imageio/jpeginput.cpp
@@ -557,14 +557,47 @@ bool
 JpgInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
                                void* data)
 {
+    return read_native_scanlines(subimage, miplevel, y, y + 1,
+                                 as_writable_bytes(data, m_spec.scanline_bytes(
+                                                             true)));
+}
+
+
+
+bool
+JpgInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
+                                int yend, int z, void* data)
+{
+    size_t size = m_spec.scanline_bytes(true) * size_t(yend - ybegin);
+    return read_native_scanlines(subimage, miplevel, ybegin, yend,
+                                 as_writable_bytes(data, size));
+}
+
+
+
+bool
+JpgInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
+                                int yend, span<std::byte> data)
+{
     lock_guard lock(*this);
     if (!seek_subimage(subimage, miplevel))
         return false;
     if (m_raw)
         return false;
-    if (y < 0 || y >= (int)m_cinfo.output_height)  // out of range scanline
+    if (ybegin < 0 || yend > (int)m_cinfo.output_height || ybegin >= yend) {
+        // out of range scanlines
+        errorfmt(
+            "JPEG read_native_scanlines: Out of valid range scanline indices (b={} e={}).",
+            ybegin, yend);
         return false;
-    if (m_next_scanline > y) {
+    }
+
+    // Validate that the span provided can hold the requested scanlines.
+    if (!valid_raw_span_size(data, m_spec, m_spec.x, m_spec.x + m_spec.width,
+                             ybegin, yend))
+        return false;
+
+    if (m_next_scanline > ybegin) {
         // User is trying to read an earlier scanline than the one we're
         // up to.  Easy fix: close the file and re-open.
         // Don't forget to save and restore any configuration settings.
@@ -594,8 +627,8 @@ JpgInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         const size_t row_size   = uhdr_raw->stride[UHDR_PLANE_PACKED] * nbytes;
         unsigned char* top_left = static_cast<unsigned char*>(
             uhdr_raw->planes[UHDR_PLANE_PACKED]);
-        unsigned char* row_data_start = top_left + row_size * y;
-        memcpy(data, row_data_start, row_size);
+        unsigned char* row_data_start = top_left + row_size * ybegin;
+        memcpy(data.data(), row_data_start, row_size * (yend - ybegin));
 
         return true;
     }
@@ -607,27 +640,47 @@ JpgInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         return false;
     }
 
-    void* readdata = data;
+    int nscanlines     = yend - ybegin;
+    size_t sl_bytes    = m_spec.scanline_bytes(true /*native*/);
+    JSAMPLE** readdata = OIIO_ALLOCA(JSAMPLE*, nscanlines);
+    for (int i = 0; i < nscanlines; ++i)
+        readdata[i] = reinterpret_cast<JSAMPLE*>(&data[i * sl_bytes]);
+
     if (m_cmyk) {
         // If the file's data is CMYK, read into a 4-channel buffer, then
         // we'll have to convert.
-        m_cmyk_buf.resize(m_spec.width * 4);
-        readdata = &m_cmyk_buf[0];
+        m_cmyk_buf.resize(m_spec.width * 4 * nscanlines);
+        for (int i = 0; i < nscanlines; ++i)
+            readdata[i] = reinterpret_cast<JSAMPLE*>(m_cmyk_buf.data()
+                                                     + i * m_spec.width * 4);
         OIIO_DASSERT(m_spec.nchannels == 3);
     }
 
-    for (; m_next_scanline <= y; ++m_next_scanline) {
+    // If we need to skip scanlines to get to the first requested one...
+    for (; m_next_scanline < ybegin; ++m_next_scanline) {
         // Keep reading until we've read the scanline we really need
-        if (jpeg_read_scanlines(&m_cinfo, (JSAMPLE**)&readdata, 1) != 1
-            || m_fatalerr) {
+        if (jpeg_read_scanlines(&m_cinfo, readdata, 1) != 1 || m_fatalerr) {
             errorfmt("JPEG failed scanline read (\"{}\")", filename());
             return false;
         }
     }
 
+    for (int y = ybegin; y < yend;) {
+        int toread = y - ybegin;
+        int r      = jpeg_read_scanlines(&m_cinfo, readdata + toread, yend - y);
+        if (r == 0 || m_fatalerr) {
+            errorfmt(
+                "JPEG failed scanline reading scanlines {}-{} from \"{}\" (returned {})",
+                y, yend, filename(), r);
+            return false;
+        }
+        y += r;
+    }
+    m_next_scanline = yend;
+
     if (m_cmyk)
-        cmyk_to_rgb(m_spec.width, (unsigned char*)readdata, 4,
-                    (unsigned char*)data, 3);
+        cmyk_to_rgb(m_spec.width * nscanlines, (unsigned char*)readdata, 4,
+                    (unsigned char*)data.data(), 3);
 
     return true;
 }
--- a/src/libOpenImageIO/imageinput.cpp
+++ b/src/libOpenImageIO/imageinput.cpp
@@ -233,22 +233,25 @@ ImageInput::read_scanline(int y, int z, TypeDesc format, void* data,
         m_spec.auto_stride(xstride, format, m_spec.nchannels);
     // Do the strides indicate that the data area is contiguous?
     bool contiguous = (xstride == buffer_pixel_bytes);
+    // Bytes a scanline takes in native formats
+    size_t native_sl_bytes = m_spec.scanline_bytes(true);
 
     // If user's format and strides are set up to accept the native data
     // layout, read the scanline directly into the user's buffer.
     if (native_data && contiguous)
-        return read_native_scanline(current_subimage(), current_miplevel(), y,
-                                    z, data);
+        return read_native_scanlines(current_subimage(), current_miplevel(), y,
+                                     y + 1,
+                                     as_writable_bytes(data, native_sl_bytes));
 
     // Complex case -- either changing data type or stride
     int scanline_values = m_spec.width * m_spec.nchannels;
 
     unsigned char* buf;
-    OIIO_ALLOCATE_STACK_OR_HEAP(buf, unsigned char,
-                                m_spec.scanline_bytes(true));
+    OIIO_ALLOCATE_STACK_OR_HEAP(buf, unsigned char, native_sl_bytes);
 
-    bool ok = read_native_scanline(current_subimage(), current_miplevel(), y, z,
-                                   buf);
+    bool ok = read_native_scanlines(current_subimage(), current_miplevel(), y,
+                                    y + 1,
+                                    as_writable_bytes(buf, native_sl_bytes));
     if (!ok)
         return false;
     if (m_spec.channelformats.empty()) {
@@ -331,16 +334,19 @@ ImageInput::read_scanlines(int subimage, int miplevel, int ybegin, int yend,
     bool contiguous                = (xstride == (stride_t)buffer_pixel_bytes
                        && ystride == (stride_t)buffer_scanline_bytes);
 
+    auto dataspan = make_span(static_cast<std::byte*>(data),
+                              buffer_scanline_bytes * size_t(yend - ybegin));
+
     // no_type_convert is true if asking for data in the native format
     bool no_type_convert = (format == spec.format
                             && spec.channelformats.empty());
     if ((native || no_type_convert) && contiguous) {
         if (chbegin == 0 && chend == spec.nchannels)
-            return read_native_scanlines(subimage, miplevel, ybegin, yend, z,
-                                         data);
+            return read_native_scanlines(subimage, miplevel, ybegin, yend,
+                                         dataspan);
         else
-            return read_native_scanlines(subimage, miplevel, ybegin, yend, z,
-                                         chbegin, chend, data);
+            return read_native_scanlines(subimage, miplevel, ybegin, yend,
+                                         chbegin, chend, dataspan);
     }
 
     // No such luck.  Read scanlines in chunks.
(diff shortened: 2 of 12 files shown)


lgritz:
The ABI test is expected to fail since this... intentionally changes the ABI.

EmilDohne:
Loving these changes! Will try and take a look before the end of the week :)

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -556,15 +556,48 @@ cmyk_to_rgb(int n, const unsigned char* cmyk, size_t cmyk_stride,
...
+                                                             true)));
+}
+
+
+
+bool
+JpgInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
+                                int yend, int z, void* data)
+{
+    size_t size = m_spec.scanline_bytes(true) * size_t(yend - ybegin);
```
EmilDohne, line 571: [nit] c-style cast to static_cast

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -556,15 +556,48 @@ cmyk_to_rgb(int n, const unsigned char* cmyk, size_t cmyk_stride,
...
+JpgInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
+                                int yend, span<std::byte> data)
 {
     lock_guard lock(*this);
     if (!seek_subimage(subimage, miplevel))
         return false;
     if (m_raw)
         return false;
-    if (y < 0 || y >= (int)m_cinfo.output_height)  // out of range scanline
+    if (ybegin < 0 || yend > (int)m_cinfo.output_height || ybegin >= yend) {
```
EmilDohne, line 587: [nit] c-style cast to static_cast

src/jpeg.imageio/jpeginput.cpp
EmilDohne, line 571: I would like to add a check here if yend < ybegin if the user passed bad data to catch it earlier than underflowing the size_t

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -607,27 +640,47 @@ JpgInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+            return false;
+        }
+        y += r;
+    }
+    m_next_scanline = yend;
+
     if (m_cmyk)
-        cmyk_to_rgb(m_spec.width, (unsigned char*)readdata, 4,
-                    (unsigned char*)data, 3);
+        cmyk_to_rgb(m_spec.width * nscanlines, (unsigned char*)readdata, 4,
```
EmilDohne, line 682: [nit] c-style cast to reinterpret_cast

src/libOpenImageIO/imageinput.cpp
```diff
@@ -350,14 +356,16 @@ ImageInput::read_scanlines(int subimage, int miplevel, int ybegin, int yend,
     int chunk = std::max(1, (1 << 26) / int(spec.scanline_bytes(true)));
     chunk     = std::max(chunk, int(oiio_read_chunk));
     chunk     = round_to_multiple(chunk, rps);
-    std::unique_ptr<char[]> buf(new char[chunk * native_scanline_bytes]);
+    std::unique_ptr<std::byte[]> buf(
```
EmilDohne, line 359: Any particular reason we don't use std::vector here? If it is to avoid the initialization cost I think something like [this](https://stackoverflow.com/questions/21028299/is-this-behavior-of-vectorresizesize-type-n-under-c11-and-boost-container/21028912#21028912) could be helpful instead

src/libOpenImageIO/imageinput.cpp
```diff
@@ -843,6 +899,106 @@ ImageInput::read_native_tiles(int subimage, int miplevel, int xbegin, int xend,
...
+                              int ybegin, int yend, int chbegin, int chend,
+                              span<std::byte> data)
+{
+    if (pvt::oiio_print_debug
+#ifndef NDEBUG
+        || true
+#endif
+    ) {
+        // Only for debug builds or if OIIO "debug" is enabled at runtime,
+        // validate that the span size.
```
EmilDohne, line 936: [nit] This sentence seems to be missing "is correct"

src/libOpenImageIO/imageinput.cpp
```diff
@@ -1273,6 +1429,27 @@ ImageInput::check_open(const ImageSpec& spec, ROI range, uint64_t /*flags*/)
...
+                                int zbegin, int zend, int chbegin, int chend)
+{
+    if (chend < 0 || chend > spec.nchannels)
+        chend = spec.nchannels;
+    size_t sz = spec.pixel_bytes(chbegin, chend, true /*native*/)
+                * size_t(xend - xbegin) * size_t(yend - ybegin)
+                * size_t(zend - zbegin);
+    if (sz > buf.size()) {
+        errorfmt(
+            "Buffer span size is inadequate to read the requested {} pixels ({} bytes vs {} needed)",
```
EmilDohne, line 1444: [question] Does errorfmt have something like `{:L}` that e.g. `std::format` would have to display large numbers more nicely?

src/tiff.imageio/tiffinput.cpp
```diff
@@ -1765,17 +1815,19 @@ TIFFInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+        if (TIFFReadScanline(m_tif, m_scratch.data(), y) < 0) {
 #if OIIO_TIFFLIB_VERSION < 40500
             errorfmt("{}", oiio_tiff_last_error());
 #endif
             return false;
         }
+        size_t n(m_spec.width);
         if (m_bitspersample <= 8)
-            palette_to_rgb(m_spec.width, &m_scratch[0], (unsigned char*)data);
+            palette_to_rgb(n, make_cspan((uint8_t*)m_scratch.data(), n),
```
EmilDohne, line 1826: [nit] prefer static_cast/reinterpret_cast over c-style casts

src/tiff.imageio/tiffinput.cpp
```diff
@@ -1765,17 +1815,19 @@ TIFFInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+                           span_cast<uint8_t>(data));
         else if (m_bitspersample == 16)
-            palette_to_rgb(m_spec.width, (uint16_t*)&m_scratch[0],
-                           (unsigned char*)data);
+            palette_to_rgb(n, make_cspan((uint16_t*)m_scratch.data(), n),
```
EmilDohne, line 1829: [nit] prefer static_cast/reinterpret_cast over c-style casts

src/tiff.imageio/tiffinput.cpp
```diff
@@ -1784,7 +1836,7 @@ TIFFInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
     size_t input_bytes = plane_bytes * m_inputchannels;
     // Where to read?  Directly into user data if no channel shuffling, bit
     // shifting, or CMYK conversion is needed, otherwise into scratch space.
-    unsigned char* readbuf = (unsigned char*)data;
+    unsigned char* readbuf = (unsigned char*)data.data();
```
EmilDohne, line 1839: [nit] prefer static_cast/reinterpret_cast over c-style casts

EmilDohne (commented):
Out of my own curiosity, are templated read variants planned as a separate MR? I for one find myself quite often writing a function like this and then having to convert from a typed vector to now a bytespan:

```cpp
template <typename T>
constexpr OIIO::TypeDesc get_type_desc()
{
	if constexpr (std::is_same_v<T, uint8_t>)
	{
		return OIIO::TypeDesc::UINT8;
	}
	else if constexpr (std::is_same_v<T, uint16_t>)
	{
		return OIIO::TypeDesc::UINT16;
	}
	else if constexpr (std::is_same_v<T, uint32_t>)
	{
		return OIIO::TypeDesc::UINT32;
	}
	else if constexpr (std::is_same_v<T, float>)
	{
		return OIIO::TypeDesc::FLOAT;
	}
	else if constexpr (std::is_same_v<T, half>)
	{
		return OIIO::TypeDesc::HALF;
	}
	return OIIO::TypeDesc::UNKNOWN;
}
```

lgritz:
> Out of my own curiosity, are templated read variants planned as a separate MR?

Yes, this first one is just for the methods that get overloaded by the format reader implementations (the ones named read_native_blah). These aren't called by applications using OIIO, and they are very bare bones -- the only support reading "native" data types into contiguous buffers.

A subsequent PR will add the methods called by client software -- such as read_scanlines(), etc. Those do data type conversion, can handle placing data into strided noncontiguous positions, etc. These will have templated overloads for all the types, as well as a new image_span way of passing the pointers, bounds, and strides all together.

> I for one find myself quite often writing a function like this and then having to convert from a typed vector to now a bytespan:
> 
> ```c++
> template <typename T>
> constexpr OIIO::TypeDesc get_type_desc()
> ```

Oh, we already have that:

    BaseTypeFromC<T>::value
    TypeDescFromC<T>::value
    CType<T_basetype>::type

they're all in typedesc.h.

EmilDohne:
> Oh, we already have that:
   ```
   BaseTypeFromC<T>::value
   TypeDescFromC<T>::value
   CType<T_basetype>::type
   they're all in typedesc.h.
   ```

Thanks I didnt know that!

src/jpeg.imageio/jpeginput.cpp
lgritz: good idea, yes

src/jpeg.imageio/jpeginput.cpp
lgritz: Good idea, yes. Done.

lgritz: Can I ask why that is a superior solution? I think this kind of cast is compact and easy to read. What's the advantage of the more verbose `static_cast<size_t>` syntax here?

src/jpeg.imageio/jpeginput.cpp
lgritz: In this case, I can see why `int(foo)` is a nicer notation than the original C `(int)foo`, but I'm unclear on what advantage there is to the significantly more verbose `static_cast<int>(foo)` when it means exactly the same thing. Can you enlighten me?

I'm not asking this to be a jerk. I want to be convinced. In 33 years of using C++ as my primary language, I've never understood why using the more cumbersome notation is better in a case like this of converting a simple value between two integer types.

src/libOpenImageIO/imageinput.cpp
lgritz: Yes, exactly, I did that to not pay the cost of zeroing out the allocated arrays when it's not necessary.

But that is a very interesting link! I think I knew theoretically that this was possible with a special allocator, but in multiple decades of C++ development, I'm not sure I've ever used anything but the default allocator, and I feel like I never would have been able to reproduce that link's code.

I don't want to derail this PR, but I'll look separately about putting that no-init allocator someplace where we can use it regularly and then I can phase out the unique_ptr trick.

src/libOpenImageIO/imageinput.cpp
lgritz: good catch, thanks

src/libOpenImageIO/imageinput.cpp
lgritz: errorfmt is just fmtlib underneath, which is equivalent to std::format for all practical purposes. I admit that until just now, I did not know about `{:L}` (which for an integer means to format with locale-aware digit separators).

I have mixed feelings about this -- I understand the desire to make large numbers more readable, but we try very hard not to allow any locale-specific behavior.

Not only is this an error message, but it's one that only comes into play if there is a programming error (not a user error or a bad file). So I think the practical value of making this look nicer is very low.

src/tiff.imageio/tiffinput.cpp
lgritz: done

src/tiff.imageio/tiffinput.cpp
lgritz: done

src/tiff.imageio/tiffinput.cpp
lgritz: done

src/jpeg.imageio/jpeginput.cpp
EmilDohne: The main advantage is compile time casting vs runtime casting, so while a c-style cast may fail at runtime, a cpp style cast will fail at compile time (or emit a warning)

src/jpeg.imageio/jpeginput.cpp
EmilDohne: > I'm not asking this to be a jerk.

Not at all! Tbh it took me a while to understand the differences as well but from what I understand there is two main advantages:

`static_cast<T>` can be easier found during e.g. replacement as well as being more type safe. This page is quite standardese but it essentially boils down that a `T(var)` will start off with a static_cast but then also try const_cast reinterpret_cast etc. allowing for undefined behaviour while `static_cast` will only allow well defined conversions.

https://en.cppreference.com/w/cpp/language/explicit_cast

For two ints it likely doesn't matter as the behaviour would be the same to truncate on narrowing but it would be for consistency and ensuring that you can rely on the program being well-formed.

Taking this example:

```cpp
const int x = 42;
int* ptr1 = (int*)&x; 
```

the compiler will happily let you violate the constness of x while this would not be the case for static_cast<>

```
<source>:8:17: error: static_cast from 'const int *' to 'int *' is not allowed
    8 |     int* ptr1 = static_cast<int*>(&x); 
      |                 ^~~~~~~~
```

Hopefully that gave some insight :)


FYI :
```
T(x)
(T)x
``` 
are almost entirely identical except that `T(x)` will only allow for a `simple-type-specifier` so no pointers but that can be gotten around by typedefing

```
const int x = 42;
typedef int* foo;
int* ptr1 = foo(x);
```

src/libOpenImageIO/imageinput.cpp
EmilDohne: Yeah this would definitely be as part of a different MR but just wanted to make you aware of it in case you weren't, I also would never be able to come up with something like this. One thing to keep in mind is that these uninitialized vectors are different types than `std::vector<T>` with `std::allocator<T>` so one cannot be assigned or referenced by the other. That being said e.g. a `cspan<T>` or `span<T>` should get around this

src/libOpenImageIO/imageinput.cpp
EmilDohne: Good point, I also agree that it is quite unfortunate to have it be locale specific, to get around this I usually roll my own function to format these numbers but as programmer error it might even be better without so you can copy paste

src/jpeg.imageio/jpeginput.cpp
EmilDohne: As a slight continuation of that, actually only `T(x)` is equivalent to a c-style cast, any of these would be fine:

`T val(x);`
`auto val = T{x}`

lgritz: Yeah, I understand that C cast is a blunt "cast all the things" whereas the C++ casts cleanly separate concerns of casting constness (const_cast), versus pointer type (reinterpret_cast), versus class inheritance hierarchy (dynamic_cast), versus value type conversion (static_cast). In most cases -- certainly all but the last -- I do tend to prefer the C++ style (and usually use it, though sometimes I forget and revert to old habits, and you caught a few of these).

My comments here are strictly regarding simple value type conversion where there are no pointers, no class hierarchy, no constness issues (it's just a value). In this particular case, `T(x)` does exactly the same thing as `static_cast<T>(x)` and I tend to prefer the more compact notation.

I think I'll leave this for now, because it's the prevailing style of the code base. I don't mind when people use the more verbose cast and would never make them change, but it just doesn't roll off the tongue for me. (I do appreciate you pointing out all of these, and I did change the pointer casts to reinterpret_cast, I think that's a case where the C++ syntax is very helpful because it prevents you from accidentally casting away constness.)

lgritz:
Any other comments on this from anyone?

EmilDohne:
LGTM!

jessey-git:
It turns out this commit breaks DDS reading of 3d textures. DDS is a non-tiled format so what is the recommended way to fix now?

lgritz:
Do you have a file you can send me that's broken?

jessey-git:
Yes sir, see attached for an image and a python script that prints some data about it to get started.

Namely for "good" builds you should see something like this at the very bottom of the file:
```
unique values found: 219
overall min (0.000 0.000 0.000 0.000)
overall max (1.000 1.000 1.000 0.000)
overall avg (0.855 0.748 0.804 0.000)
overall std (0.271 0.338 0.292 0.000)
```

And for "bad" builds you'll see:
```
unique values found: 24
overall min (0.000 0.000 0.000 0.000)
overall max (1.000 1.000 0.769 0.000)
overall avg (0.839 0.740 0.297 0.000)
overall std (0.294 0.346 0.261 0.000)
```
[ddstest.zip](https://github.com/user-attachments/files/26494623/ddstest.zip)

lgritz:
Let's continue the DDS breakage discussion on #5132

PR #4671, opened 2025-03-17
fix: invalid loop bound when appending mipmap textures using oiiotool

bfraboni:
The following command fails for images with mip map levels due to a bug:
```
oiiotool mipmap_images.*.exr --siappendall -o all_images.exr
```

If the option `-a` is not appended to the command, mip levels higher than 0 should be ignored, instead it just segfault due to out of bound access; I just fixed the it in the PR cf. [my comment below](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4671#discussion_r1999705885).

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -2982,7 +2982,7 @@ action_subimage_append_n(Oiiotool& ot, int n, string_view command)
     for (int i = 0; i < n; ++i) {
         ImageRecRef A = images[i];
         for (int s = 0; s < A->subimages(); ++s, ++sub) {
-            for (int m = 0; m < A->miplevels(s); ++m) {
+            for (int m = 0; m < allmiplevels[s]; ++m) {
                 bool ok = (*R)(sub, m).copy((*A)(s, m));
                 if (!ok) {
                     ot.error(command, (*R)(sub, m).geterror());


src/oiiotool/oiiotool.cpp
```diff
@@ -2982,7 +2982,7 @@ action_subimage_append_n(Oiiotool& ot, int n, string_view command)
     for (int i = 0; i < n; ++i) {
         ImageRecRef A = images[i];
         for (int s = 0; s < A->subimages(); ++s, ++sub) {
-            for (int m = 0; m < A->miplevels(s); ++m) {
```
bfraboni, line 2985: The number of allocated miplevels for subimage `s` can differ from `A->miplevels(s)`, hence we have to use `allmiplevels[s]` instead:
https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/415fe634ba1e37953c31cc507f0282b6f96ba046/src/oiiotool/oiiotool.cpp#L2970-L2971

src/oiiotool/oiiotool.cpp
```diff
@@ -2982,7 +2982,7 @@ action_subimage_append_n(Oiiotool& ot, int n, string_view command)
...
+            for (int m = 0; m < allmiplevels[s]; ++m) {
```
bfraboni, line 2985: The number of allocated miplevels for subimage `s` can differ from `A->miplevels(s)`, hence we have to use `allmiplevels[s]` instead:
https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/415fe634ba1e37953c31cc507f0282b6f96ba046/src/oiiotool/oiiotool.cpp#L2970-L2971

lgritz (approved):
LGTM. Good catch!

PR #4672, opened 2025-03-19
feat: ImageBuf::merge_metadata and oiiotool --mergemeta

lgritz:
Clarify that IB::copy_metadata and oiiotool --pastemea fully replace all existing metadata with that of the other IB.

Add a new IB::merge_metadata() and oiiotool --mergemeta that merges in metadata from the other IB, preserving metadata in `this` that was not in the other IB.  Optionally, metadata that existed in both can be overridden.  Optionally, a subset of metadata to be merged can be selected via a regex.


ianianda:
I've reviewed this implementation of this --pastemeta and its optional features. The approach looks exactly what we need for production, and the code is clean.

kato87:
`185240`

PR #4673, opened 2025-03-19
Improved IFF support reading and writing z buffers

mikaelsundell:
## Description

Added support for reading and writing ZBUF chunks, also known as Z-buffers or depth maps
Removed noproxy versions, no longer needed

## Checklist:

  doesn't change the API.)
  nonconforming.


src/iff.imageio/iffinput.cpp
```diff
@@ -385,48 +436,60 @@ IffInput::read_header()
...
                                     return false;
+                                }
                             }
                         }
                         // TBHD done, break
                         break;
+                    } else {
+                        // skip to the next block.
+                        if (!ioseek(chunksize, SEEK_CUR)) {
+                            errorfmt("IFF error io seek failed");
```
lgritz, line 480: The ImageInput helper methods ioseek() and ioread() handle calling errorfmt for you. If they fail, you just need to return false, you don't need to clutter with another error message call.

src/iff.imageio/iffinput.cpp
```diff
@@ -452,27 +514,30 @@ IffInput::read_native_tile(int subimage, int miplevel, int x, int y, int /*z*/,
     if (!seek_subimage(subimage, miplevel))
         return false;
 
-    if (m_buf.empty())
-        readimg();
+    if (m_buf.empty()) {
+        if (!readimg()) {
+            errorfmt("IFF error could not read image for tile");
```
lgritz, line 519: If readimg() itself calls errorfmt for each way that it can fail, you don't need to do it again here.

src/iff.imageio/iffoutput.cpp
```diff
@@ -144,28 +145,57 @@ IffOutput::open(const std::string& name, const ImageSpec& spec, OpenMode mode)
...
+
+    // IFF image files only supports UINT8 and UINT16.  If something
+    // else was requested, revert to the one most likely to be readable
+    // by any IFF reader: UINT8
+
+    TypeDesc base_format = spec.format;
+    if (base_format != TypeDesc::UINT8 && base_format != TypeDesc::UINT16) {
+        errorfmt("Unsupported format {}. Converting to UINT8.", base_format);
+        base_format = TypeDesc::UINT8;
+    }
```
lgritz, line 177: I think that most of the readers pick the supported data type that loses the least precision versus the one requested. So if they ask for something at least as big as uint16, we should probably give them uint16 rather than uint8, which is even less fidelity.

mikaelsundell:
There will be an update on this one today!

src/iff.imageio/iffinput.cpp
```diff
@@ -713,23 +787,87 @@ IffInput::readimg()
...
+                    }
+                }
+            }
+            ztiles++;
 
         } else {
             // skip to the next block
-            if (!ioseek(chunksize))
+            if (!ioseek(chunksize, SEEK_CUR)) {
+                errorfmt("IFF error io seek failed");
```
lgritz, line 858: Not necessary -- the ioseek method itself calls errorfmt if it fails. All you have to do here is return.

lgritz:
Unfortunately, this dumps core on some example iff files I have. It's not something I can send. Let me see if I can at least point you to the part of the code where it crashes.

src/iff.imageio/iffinput.cpp
```diff
@@ -713,23 +787,87 @@ IffInput::readimg()
...
+                for (uint32_t py = ymin; py <= ymax; py++) {
+                    uint8_t* out_dy = static_cast<uint8_t*>(m_buf.data())
+                                      + (py * m_header.width)
+                                            * m_header.pixel_bytes();
+
+                    for (uint16_t px = xmin; px <= xmax; px++) {
+                        uint8_t* out_p = out_dy + px * m_header.pixel_bytes()
+                                         + m_header.rgba_channels_bytes() + c;
+                        *out_p++ = *in_p++;
+                    }
```
lgritz, line 850: For an RGBAZ iff file (uint8 RGBA channels, float Z channel), I'm getting crashes at line 852 when releasing a std::vector (presumably the "in" vector declared at line 834).

I suspect that this loop may be the culprit. Are you sure about those offets? I believe that here, we are taking the compressed single z channel (stored at this point in `scratch` and copying it to the right postion in the output buffer. But the

    ... + m_header.rgba_channels_bytes() + c

looks wrong to me, and also the next line

    *outp++ = *in_p++;

These are uint8_t pointers, so you're incrementing them by 1 each here, but isn't the data in "in" *really* float (the Z channel) and isn't the spacing of the output buffer going to be m_header.pixel_bytes() ?

Or am I misunderstanding what's happening here?

mikaelsundell: The reason this happens is that each call to uncompress_rle_channel decompresses floats and uint16 values byte by byte, while also handling planar storage and byte order at the same time. That’s why uint8_t is used across uint8, uint16, and float decompression and compression—it's one RLE run per byte, not per channel in this case.

pixel_bytes() is full RGB(A)+Z while rgba_channel_bytes() is RGB(A) only cause RGBA and Z chunks are stored in the actual IFF layout while we store interleaved/ packed.

My guess is that there’s a mismatch between the expected chunk size and what’s actually in the chunk (RLE vs uncompressed etc), which might be causing the issue. Are you able to open your sample files in Nuke without any problems? Or maybe try FCheck on a Windows machine? I could be wrong of course.

I could write out the binary layout if we can't reproduce it with other non-confidential files. May I ask—what software wrote the IFF files, maybe that's an easier way to the issue. After all, you are probably right there is a run over somewhere and a byte is all it needs. There must be a glitch in the code unless the original software you used writes IFF in a different way or with incorrect or non-aligned chunk sizes. If it works in Nuke etc it should work in my code.

This old legacy format is fun—if you had asked me to write it today, I would’ve done things _slightly_ differently. 😄

lgritz: The good news is that I have at least 2 different iff files that crash here -- they are RGBAZ (uint8 for RGBA, float for Z), RLE compression. The bad news is that they are production frames and I can't send them out.

They appear to be Maya playblast output. They're from different shows at different times, so I think maybe it won't be difficult for you to reproduce from scratch on your end?

Will try loading with Nuke and report back...

lgritz: BTW, those failing frames do not crash in the mainline OIIO (not using this PR's patch)... but then again, I think it may simply be ignoring the Z channel entirely in that case, so I wouldn't say they are "correct" exactly.

lgritz: Nuke does load it. I'm not sure how to verify that it interprets the depth values correctly.

mikaelsundell: Must be fixed! if the test files are coming from Maya it should be an easy thing to reproduce. You are able to open the earlier test file from Nicholas, it's also Maya and RGBAZ.

mikaelsundell: Nuke will scream in red if the Z buffer reading is off, my code must be off/ wrong somewhere. Let's try to find it, try to open the Nicholas file and see if you get the same problem. I will also switch between a few different configuration and see if I can trigger the issue.

Are you doing something like a RGBAZ channels from iff => exr with oiiotool when testing?

lgritz: Which earlier test file from Nicholas?

lgritz: I fail just doing a straightforward `iconvert foo.iff foo.exr` or `iinfo -v --stats` or oiiotool, yes.

mikaelsundell:
New version in place with both support for uncompressed data (not verified with real IFF/z-buffer files) and the use of spans to avoid buffer runovers. I also removed some redundant error messages.

lgritz:
(Mikael and I have been discussing this off-channel, but wanted to be sure people knew progress was being made here.)

mikaelsundell:
Thanks Larry, will be back this weekend to finish up, out traveling this week.

lgritz:
Excuse my cluttering things up, but I'm using this PR as a guinea pig to request a GH Copilot review just to see what it says. Feel free to ignore any of it, this is just an experiment.

mikaelsundell:
Great! I'm back now, will finish this up in the next couple of days, will make sure I copilot my changes before commit. Let the bot battle begin! :-D

mikaelsundell:
New commits, spans + reading and writing of uncompressed IFF data verified. I've tested both uncompressed + rle with nuke and photoshop and both works for RGBAZ in 8/16-bit.

src/iff.imageio/iffinput.cpp
```diff
@@ -50,7 +50,8 @@ class IffInput final : public ImageInput {
     bool readimg(void);
 
     // helper to uncompress a rle channel
-    size_t uncompress_rle_channel(const uint8_t* in, uint8_t* out, int size);
+    size_t uncompress_rle_channel(span<uint8_t> in, span<uint8_t> out,
```
lgritz, line 53: Let's make that `cspan<uint8_t> in` since it's const data?

src/iff.imageio/iffinput.cpp
```diff
@@ -526,210 +593,391 @@ IffInput::readimg()
...
 
                         std::vector<uint8_t> in(tw * th);
-                        uint8_t* in_p = &in[0];
+                        span<uint8_t> in_span(in);
+
+                        size_t used = uncompress_rle_channel(scratch_span,
+                                                             in_span, tw * th);
+                        if (used > scratch_span.size()) {
+                            errorfmt(
+                                "RLE uncompress exceeds span size (channel byte %d)",
```
lgritz, line 740: All the functions ending in `fmt` take the new std::format idioms, so this should be:

```suggestion
                                "RLE uncompress exceeds span size (channel byte {})",
```

src/iff.imageio/iffinput.cpp
```diff
@@ -526,210 +593,391 @@ IffInput::readimg()
...
-                                                    * m_spec.pixel_bytes();
+                        size_t offset = 0;
+                        for (uint16_t py = ymin; py <= ymax; ++py) {
+                            uint8_t* out_dy = m_buf.data()
+                                              + (py * m_header.width)
+                                                    * m_header.pixel_bytes();
+
+                            for (uint16_t px = xmin; px <= xmax; ++px) {
+                                if (offset >= in_span.size()) {
+                                    errorfmt("in_span underflow at (%u, %u)",
```
lgritz, line 755:
```suggestion
                                    errorfmt("in_span underflow at ({}, {})",
```

src/iff.imageio/iffinput.cpp
```diff
@@ -526,210 +593,391 @@ IffInput::readimg()
...
+                                memcpy(&pixel, pixel_in.data() + c * 2, 2);
+
                                 if (littleendian()) {
                                     swap_endian(&pixel);
                                 }
-                                *sl_p++ = pixel;
+
+                                if (sl_span.empty()) {
+                                    errorfmt(
+                                        "scanline span overflow at (%u, %u)",
```
lgritz, line 808:
```suggestion
                                        "scanline span overflow at ({}, {})",
```

src/iff.imageio/iffinput.cpp
```diff
@@ -526,210 +593,391 @@ IffInput::readimg()
...
+
+                    int sx = 0;
+                    for (uint16_t px = xmin; px <= xmax; px++, sx++) {
+                        size_t pixel_index  = sy * tw + sx;
+                        size_t pixel_offset = pixel_index
+                                              * m_header.zbuffer_bytes();
+
+                        if (pixel_offset + m_header.zbuffer_bytes()
+                            > scratch_span.size()) {
+                            errorfmt("in span overflow at pixel (%u, %u)", px,
```
lgritz, line 946:
```suggestion
                            errorfmt("in span overflow at pixel ({}, {})", px,
```

I'm not going to mark all of them, you can do it as well as I can

lgritz (changes requested):
Overall, LGTM and works on my iff files that once had trouble.

I did spot some problems with errorfmt calls getting confused about printf vs the newer format specifiers (throughout OIIO, all formatted string functions ending in "fmt" use the new syntax, not the old printf one). Those will need to get cleaned up, so some searching to be sure to find all of them.

With those changes, I think we are ready to merge this.

lgritz (approved):
LGTM!

mikaelsundell:
Awesome 🙂

PR #4674, opened 2025-03-21
feat: oiiotool --mergemeta pt 2: fold into --pastemeta

lgritz:
After merging #4672, we noticed some things that could be improved, so this is a further revision (and partial retraction).

There was already an oiiotool flag, `--metamerge`, which in retrospect make it seem unnecessarily confusing to add --mergemeta. So this PR removes this extra new command, instead simply piggy-backing on the existing --pastemeta by adding to it optional modifiers:

- `:merge=` controls the merge strategy, 0 (default) means discard the existing metadata entirely and use only the pasted, matching the original --pastemeta behavior; 1 means merge nondestructively (no existing metadata will be altered or removed); 2 means destructive merge, where pasted metadata will overwrite existing metadata having the same name, but conflict-free existing metadata will remain.

- `:pattern=` regex selects only some metadata.

Additionally, we noticed an interesting corner case that surprised users when testing the new metadata merging feature from the last PR: What happens if the metadata source image has fewer subimages than the pixel source images? The behavior generally for oiiotool is that when binary operations (those taking two images as input) are on images with differing numbers of subimages, the output gets the number of subimges as the FIRST input (assuming -a instructed us to consider subimages at all). But that's not what we want here! If the metadata source has one subimage and the pixel source has many, clearly the intuitive result is for that one metadata set to be pasted into all of the subimages from the pixel-supplying image.

So we fix that so that for --pastemeta, it's the second input image (not the first) that takes the lead in determining the number of subimages we will process. This in turn took a bit of extra infrastructure fixing in the OiiotoolOp class that controls this logic.


ianianda:
This is working with images with different subimage counts and improve the metadata handling functionality in oiiotool. Looks good to me! Thanks!

lgritz:
The CI failure appears to be a false positive or some kind of abi checker confusion. It happens sometimes for things that aren't really ABI breaks, I think it gets confused by certain templates.

PR #4675, opened 2025-03-22
ci: Fix wheel building on Mac ARM (continuation of #4668)

lgritz:
The Intel Macs had this problem last week and #4668 fixed it. Now suddenly the ARM Macs had it this week. Seems to be something changing on the skbuild side.


EmilDohne:
LGTM! On an unrelated note, I always find macs the most tedious platform to build on with all its quirks and specialties

PR #4676, opened 2025-03-22
fix: Partial reversion to fix ABI checker false positive

lgritz:
I still think that the "abi break" is a false positive. But I really didn't need the new constructor that seemed to confuse the abi checker, so just to be extra sure that there is no break that I simply misunderstand, change to make abicheck happy.


lgritz:
Fixes CI break, merging immediately.

PR #4677, opened 2025-03-22
int: Experimental default_init_allocator and default_init_vector

lgritz:
Inspired by a link from EmilDohne in an unrelated PR, this introduces `default_init_allocator`, an experimental custom allocator that default initializes, rather than value-0-initializes, POD/non-class types.

This is useful for being able to make a std::vector with this allocator (`default_init_vector`), thus having the property of not incurring a large cost of needlessly initializing the elements when they are allocated, as a default std::vector does.

I picked two spots sort of randomly, in which we previously avoided using std::vector and instead did a unique_ptr / new[] dance to do scoped heap allocation without paying the vector initialization tax. By using default_init_vector, we avoid the extra cost while still using the much more convenient vector type.

These two places are just to prove it works. There's no particular need to proactively root out every use of the unique_ptr trick. But for future development, we can use now use std::vector without the construction cost when we allocate.

### DIFF
--- a/src/include/imageio_pvt.h
+++ b/src/include/imageio_pvt.h
@@ -223,67 +223,6 @@ print_stats(std::ostream& out, string_view indent, const ImageBuf& input,
 
 
 
-/// Allocator adaptor that interposes construct() calls to convert value
-/// initialization into default initialization.
-///
-/// This is a way to achieve a std::vector whose resize does not force a value
-/// initialization of every allocated element. Put in more plain terms, the
-/// following:
-///
-///     std::vector<int> v(Nlarge);
-///
-/// will zero-initialize the Nlarge elements, which may be a cost we do not
-/// wish to pay, particularly when allocating POD types that we are going to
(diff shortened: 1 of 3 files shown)


lgritz:
(If somebody wants to propose better names for these things, I'm all ears.)

jessey-git:
Yeah, my mind went to the 'di' prefix often used in chemistry to mean 2. I could imagine someone thinking this is actually a structure containing 2 parallel vectors maybe. I can only offer up "raw_vector" or the longer "default_init_vector".

lgritz:
Updated and renamed to default_init_vector

lgritz:
How do people like this with the new name? Does this seem like a good idiom to have in the toolbox? (I would rather that STL / std::vector just gave us this ability instead of our having to make it as a nonstandard thing, but such is the burden of C++.)

EmilDohne:
Im happy with default_init_vector so LGTM from my end. Agree about the vector thing but I suppose thats just one of the contracts of vector

PR #4681, opened 2025-03-28
feat(oiiotool): add --render_text modifiers measure= and render=

lgritz:
`--render_text:measure=1` causes the render_text command to set user variables `TEXT_X`, `TEXT_Y`, `TEXT_WIDTH`, and `TEXT_HEIGHT` (as if set by `--set`) to the origin offset and dimensions of the rendered text size. These can then be accessed in subsequent expression evaluation.

`--render_text:render=0` causes the text to not actually be rendered into the image. (The default is 1.) This is primarily of use in conjunction with `measure=1` if you want to measure the size that text would have taken, then use the variables for subsequent commands, but not actually draw the text.

### DIFF
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -4203,6 +4203,16 @@ current top image.
       `shadow=` *size*
         if nonzero, will make a dark shadow halo to make the text more clear
         on bright backgrounds.
+      `measure=` *int*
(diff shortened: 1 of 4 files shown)


lgritz:
Any comments or suggestions here? Does this seem useful?

lgritz:
No objections -> merging

PR #4685, opened 2025-04-01
fix(python): fix handle leak

lgritz:
Seems that for quite some time, there have been particular circumstances in which the Python bindings have leaked ImageInput or ImageOutput objects.  That's bad enough on its own, but it also leaves file handles open if code was depending on the fact that the ImageInput and ImageOutput destructors close the files if they haven't already been closed.

I think the offending part is that the `unique_ptr<ImageInput>` extracts its raw pointer and gives it to Python, on the mistaken assumption that it would be reference counted as a python object.  But apparently, it's not. Oops.

Leaving it as a unique_ptr turns out to be fine, as far as pybind11 is concerned! And on top of that, if you give an uninitialized unique_ptr to pybind11, it considers that a Python `None`, which means we can further simplify some places where we explicitly had to return a py::object and figure out if it should be None or not, but actually it's all ok to just give the unique_ptr and pybind11 figures the whole thing out for us.

### DIFF
--- a/src/python/py_imageinput.cpp
+++ b/src/python/py_imageinput.cpp
@@ -114,7 +114,7 @@ ImageInput_read_tiles(ImageInput& self, int subimage, int miplevel, int xbegin,
 
 
 
-py::object
+std::unique_ptr<DeepData>
 ImageInput_read_native_deep_scanlines(ImageInput& self, int subimage,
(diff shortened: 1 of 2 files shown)


EmilDohne:
LGTM! pybind11 handles smart pointers really nicely overall returning the proper objects to the python domain and handling releases etc. itself

PR #4686, opened 2025-04-01
build: cmake 4.0 compatibility

lgritz:
The latest cmake in some package repos is 4.0. CMake 4.0 drops support for compatibility with cmake < 3.5, and thus errors when building our own dependencies that happen to have their own cmake scripts asking for compatibility with something older than that.

Setting CMAKE_POLICY_VERSION_MINIMUM=3.5 environment and/or cmake variable seems like a hack that could hide true incompatibilities, but will get us past this for now.

Change our own minimum to the range of 3.18.2...4.0 to correctly reflect the range of versions that we believe will work.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -383,6 +383,7 @@ jobs:
         env:
           CIBW_BUILD: ${{ matrix.python }}
           CIBW_ARCHS: ${{ matrix.arch }}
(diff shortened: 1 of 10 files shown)


lgritz:
The failure of the "bleeding edge" case is unrelated and I'll investigate that separately. If all the other CI tests pass, I will merge this immediately because everything else is paralyzed by the failing CI.

PR #4687, opened 2025-04-01
ci: Update ABI standard after PR 4669

lgritz:


lgritz:
All the failures are because of unrelated cmake nonsense being addressed in another PR. This just fixes the fact that the immediately prior merge intentionally broke ABI and we have to update the standard. Merging preemptively because it's the obviously necessary thing.

PR #4688, opened 2025-04-02
build: More changes to be robust for cmake 4.0

lgritz:
zlib needed the treatment, too.


lgritz:
Merging this needed fix. The "bleeding edge" failure is unrelated and I will address it separately.

PR #4689, opened 2025-04-02
int: Switch to spans for some exif manipulation, fixing warnings

lgritz:
We recently started getting some CI failures on the "bleeding edge" test that uses the new gcc 14, seems that the newest cut of that compiler was giving us new warnings where it suspected we were writing past the valid bounds of some memory areas, but it was very hard to decipher the circumstances and either find the bug or convince myself that is was a false positive.

To help me debug it, I switched some of the internals of exif.cpp that passed around raw pointers in very hard-to-validate ways to instead using spans with explicit lengths, in the hopes that the debug-mode bounds checking it provides would allow me to more easily track down the problem.

Now it no longer gets compiler warnings.

Did I fix a very subtle bug in the process?

Does the reformulation with spans make it easier for the compiler to verify that we aren't stomping on memory, and that suppresses the warning?

Either way, it now passes all CI tests and I think the new code is safer and aligned with our strategy of slowly changing raw unbounded pointers to using spans in order to make these kinds of bugs harder to introduce.

### DIFF
--- a/src/include/OpenImageIO/span.h
+++ b/src/include/OpenImageIO/span.h
@@ -575,6 +575,17 @@ as_writable_bytes(T* ptr, size_t len) noexcept
 
 
 
+/// Convert a reference to a single variable to a const byte span of that
+/// object's memory.
+template<typename T>
+span<const std::byte>
+as_bytes_ref(const T& ref) noexcept
+{
+    return make_cspan(reinterpret_cast<const std::byte*>(&ref), sizeof(T));
+}
+
+
+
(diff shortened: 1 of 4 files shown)


lgritz:
I'm going to need to merge this shortly if CI passes, as it fixes the last CI failure standing my way of doing the release that was supposed to have been completed yesterday.

lgritz:
Aha, passing full CI now for the first time since several things simultaneously broke a couple days ago!

Issue #4682, opened 2025-03-29
[FEATURE REQUEST] Python Stubs for Type Checking and Completion


KelSolaar:
With the newly released Pypi wheels, thanks you so much for this btw, it would be awesome to see Python stubs for type checking and completion. I need to pepper the code with `# pyright: ignore` pragma and thus, the code is not checked:

```python
from OpenImageIO import DOUBLE  # pyright: ignore
from OpenImageIO import FLOAT  # pyright: ignore
from OpenImageIO import HALF  # pyright: ignore
from OpenImageIO import UINT8  # pyright: ignore
from OpenImageIO import UINT16  # pyright: ignore
from OpenImageIO import ImageInput  # pyright: ignore
from OpenImageIO import ImageOutput  # pyright: ignore
from OpenImageIO import ImageSpec  # pyright: ignore
```

Not an issue but this would be a great improvement.

@zachlewis for 👀

JGoldstone:
Should the required Python version for OpenImageIO's bindings be listed somewhere in the online docs? I don't see it.

Contrairwise (or is it conversely?) what is the most recent version of Python that could be used in the OIIO bindings that Colour would support? I was using the 3.12 `type` syntax in `camdkit` but had to back it down to 3.11 when it came to submitting the PR.

lgritz:
> Should the required Python version for OpenImageIO's bindings be listed somewhere in the online docs? I don't see it.

https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/INSTALL.md?plain=1#L38

Actually, now that I see it, it's slightly out of date: we build fine against Python 3.13 as well.

lgritz:
I'm not enough of a Python master to know what we actually need to do on our side to provide the stubs. But if somebody knows, please do make a PR.

@chadrik I seem to remember you corresponding with me about stubs a few months back? Is this something you already have?

chadrik:
Hi all,  here's the PR that adds OIIO stubs to my cg-stubs project:  https://github.com/LumaPictures/cg-stubs/pull/19


Larry, I copied in some info our previous email conversation into the PR above as a refresher for why I had to maintain a set of manual corrections.  I think we both agree it would be good to reduce these as much as possible. 

The majority of the issues that I had to correct are due to functions with might return `None` / `py:none` but which pybind11 reports as returning `object`.  For example:


```c++
       .def(
            "imagespec",
[...]

PR #4692, opened 2025-04-05
feat: Add python stub files

chadrik:
## Description

Add python stub files for code completion and static analysis.

Fixes https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4682

## Tests

No new test cases, since there is not any change to the behavior.   I do have an idea for a kind of test that might validate the stubs against the python unit tests, but I don't know if it's worth it. 

## Checklist:

  doesn't change the API.)
  nonconforming.


src/cmake/pythonutils.cmake
```diff
@@ -158,5 +160,44 @@ macro (setup_python_module)
...
+    install(FILES ${_stub_file} DESTINATION ${PYTHON_SITE_DIR} RENAME __init__.pyi COMPONENT user)
+    # install the marker file
+    install(FILES ${CMAKE_SOURCE_DIR}/src/python/py.typed DESTINATION ${PYTHON_SITE_DIR} COMPONENT user)
+
+    # Ensure this runs after PyOpenImageIO
+    add_custom_target(
+        PyOpenImageIO_stubs ALL 
+        DEPENDS "${PYTHON_VENV_DIR}/bin/activate" ${_stub_file} ${CMAKE_SOURCE_DIR}/src/python/py.typed
+    )
+    add_dependencies(PyOpenImageIO_stubs PyOpenImageIO)
```
chadrik, line 200: I'm not sure if making a custom target is the right approach here.  My goal is to make sure that the stubs are built after the python bindings.

chadrik (commented):
Here are my takeaways from this effort:

1. While `std::optional` is generally a win in terms of reducing the need for maintaining manual fixes to the stubs, there are many other scenarios where we cannot pass enough information through to pybind11 to guide the stubgen process via C++, such as unbounded tuples (`tuple[T, ...]` vs `tuple[A, B, C]`) or type unions (`T | list[T] | tuple[T, ...]`).  If we want good stubs then we're signing up for permanantly maintaining these manual corrections external to our C++ code. 
2. While I was hoping to publish a separate `types-OpenImageIO` with just the stubs, it will have a snowball effect on the CI build and release process, so for now I'd like to set a simpler goal of adding the `__init__.pyi` file to the existing `OpenImageIO` distribution.

src/python/generate_stubs.py
```diff
@@ -0,0 +1,115 @@
...
+            ("*.ImageBufAlgo.*", "bottomright", "object"): PY_TO_STDVECTOR_RESULT,
+            ("*.ImageBufAlgo.*", "color", "object"): PY_TO_STDVECTOR_RESULT,
+            # BASETYPE & str are implicitly converible to TypeDesc
+            ("*", "*", "*.TypeDesc"): "Union[TypeDesc, BASETYPE, str]",
+            # list is not strictly required
+            ("*.ImageOutput.open", "specs", "list[ImageSpec]"): "Iterable[ImageSpec]",
+        },
+        result_type_overrides={
+            # FIXME: is there a way to use std::optional for these?
+            ("*.ImageOutput.create", "object"): "ImageOutput | None",
```
chadrik, line 50: I tried to modify the wrapper code to use `std::optional`, but it complained about the classes being abstract.

src/cmake/pythonutils.cmake
```diff
@@ -158,5 +160,45 @@ macro (setup_python_module)
...
+    if(WIN32)
+        set(PYTHON_VENV_EXE "${PYTHON_VENV_DIR}/Scripts/python.exe")
+    else()
+        set(PYTHON_VENV_EXE "${PYTHON_VENV_DIR}/bin/python")
+    endif()
+
+    # Create the virtualenv if it doesn't exist
+    add_custom_command(
+        COMMAND ${Python3_EXECUTABLE} -m venv "${PYTHON_VENV_DIR}"
+        COMMAND ${PYTHON_VENV_EXE} -m pip install mypy~=1.15.0 stubgenlib~=0.1.0
```
chadrik, line 176: I need python 3.9+ to install and run a new-ish version of mypy.  The version of python does not need to match the version used to build the bindings.  Any ideas?  Does the docker image come with a newer version of python installed at the system level?

chadrik:
Idea:  converting `py_to_stdvector` to use `std::variant` could be a big win for adding more type transparency to drive more of the stub types from C++.

src/cmake/pythonutils.cmake
zachlewis: I think the runners have `uv` installed, so, theoretically, it should be possible to automatically download and invoke any version of python. 

e.g., `$ uv venv --python 3.9 --seed --no-config "${PYTHON_VENV_DIR}"`

`uv venv` is a little different from Python's venv module. The `--seed` option forces the venv to install pip and setuptools (which normally doesn't happen by default with `uv venv`), and `--no-config` forces `uv` to ignore the `pyproject.toml` at the top-level root (because otherwise, `uv` would attempt to build OpenImageIO with the python build backend, presumably recursively, forever and ever).

However, you could probably avoid a few steps by adding PEP 723 inline script metadata to the top of `generate_stubs.py` and using the `uv run` command.


For example, just above your imports, add:

```
# /// script
# requires-python = ">=3.9"
# dependencies = [
#   "mypy~=1.15.0",
#   "stubgenlib~=0.1.0",
# ]
# ///

from __future__ import absolute_import, annotations, division, print_function
...
```

And then instead of creating a virtual environment and pip installing stuff manually, you should be able to swing a command like 
`$ uv run --no-config --directory ${PYTHON_SITE_DIR} ${_stub_gen} -p OpenImageIO -o ${PYTHON_BUILD_SITE}`

The above `uv run` command should create an appropriate ad-hoc venv for running `generate_stubs.py`. As above, the `--no-config` flag ensures the top-level `pyproject.toml` is ignored. The `--directory` option forces `uv` to change directories to ${PYTHON_BUILD_SITE} before running `generate_stubs.py` -- hopefully, this allows the script to find (and import?) the OIIO Python bindings. (As opposed to setting PYTHONPATH; which would probably be a fine alternate approach if the above doesn't work).


Of course, I've never tried any of this...

chadrik: @zachlewis  Using PEP 723 entrypoint is a great idea.  Thanks!

.github/workflows/wheel.yml
```diff
@@ -375,9 +389,14 @@
~
~      - uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
~        name: Install Python
         with:
           python-version: '3.9'
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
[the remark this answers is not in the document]
lgritz: @chadrik What this alert is trying to say is to use a full commit hash instead of "v5" here, which is just a tag that could be spoofed by somebody taking control of the astral-sh repo.

zachlewis:
> In theory we could build the stubs just once on a trusted platform and python version and publish the stubs separately as a types-OpenImageIO python distribution.

This may be the cleanest and simplest thing to do; and, like you mentioned earlier, it allows for type-checking in extremely light environments. Ostensibly, we could even run a building and publishing step for the types-OpenImageIO package immediately after publishing the OpenImageIO package, assuming we're not too quick for pypi. 

Alternatively, if we limit our scope to only generating stubs when building wheels with `cibuildwheel`, perhaps we could exploit CIBW's "repair-wheel" step; i.e., we'd invoke a custom script that:
 1. repairs the wheels
 2. runs the stub generation script with the PEP 723 OpenImageIO dependency pointing to the relative location of the repaired wheels
 3. properly patches the __init__.pyi stub into the .whl

> Here's an idea for a solution to this problem:
> 
>     - commit the __init__.pyi stub to the repo (vs generating it on the fly)
>     - add a manually run script that updates the stub file (vs running it every build)

I like this, too. There's something to be said for not attempting to be too clever about all this. We can simply say, "If you're updating the Python bindings, make sure to run the script that generates new stubs". If it makes things easier for folks, we could even expose a stub-generating script as part of the OpenImageIO package itself.

zachlewis:
I messed around with the cibuildwheel "repair-wheel" step hack as described above... I haven't tested on windows, but it seems like it's doing what it's supposed to be doing at first glance...

If you're interested, I posted a proof of concept here:
https://gist.github.com/zachlewis/67ee0a8faae8c25240d8fbaf315bd7e1

chadrik:
> I messed around with the `cibuildwheel` "repair-wheel" step hack as described above... I haven't tested on windows, but it seems like it's doing what it's supposed to be doing at first glance...
> 
> If you're interested, I posted a proof of concept here: https://gist.github.com/zachlewis/67ee0a8faae8c25240d8fbaf315bd7e1

Thanks for the great suggestion @zachlewis.  I had a look and it gave me some ideas to improve my approach.  After some consideration, I think `test-command` will be better for my case than `repair-wheel-command` because the repair command is actively used and has defaults that vary per OS -- I don't want to accidentally break anything.  The `test-command` actually works a lot like a "post-build callback", so it seems like a good fit.

Have a look at the current state of things.  I'm happy with where things landed. 

There are the requirements I was trying to hit when designing this workflow:

- DO NOT complicate the default build process:  the stubs represent a niche artifact and we don't want it causing failures for folks during day to day development.  The new workflow only runs during local development if explicitly called via `make pystubs`
- notify developers if the changes in their PR will result in a change to the stubs.  I don't want to opaquely auto-generate stubs in CI and blindly shuttle them into a wheel.  I want developers to know when their C++ change results in a degradation of the stubs, so that they can go through a review process, and the `generate_stubs` rules can be updated to accommodate. 
- install stubs automatically when `pip install OpenImageIO`  is run.  We don't want developers to have to know to install this or else many people will miss out on this quality of life improvement.   I'm including the stub within every wheel distribution. 

I left two goals for later PRs:
- create a wheel that can be installed separately from the C-extension wheels, to keep lightweight environments. 
- modify the python bindings to avoid obscuring types (e.g. using modern C++ types like `std::optional`

pyproject.toml
```diff
@@ -122,4 +122,21 @@ CXXFLAGS = "-Wno-error=stringop-overflow -Wno-pragmas"
...
+# The test command acts as a kind of "post-build" callback where it's possible
+# for the stub-generator to import the freshly-built wheel.
+# There are two entry-points which are designed to call generate_stubs.py through
+# this test command:
+# - `make pystubs` is called during local development to generate the
+#   stubs and copy them into the git repo to be committed and reviewed.
+# - in CI, the cibuildwheel action is used to validate that the stubs match what
+#   has been committed to the repo.
+select = "cp311-manylinux_*64"
+test-requires = "mypy~=1.15.0 stubgenlib~=0.1.0"
```
chadrik, line 138: I decided use this instead of a PEP 723 script for `generate_stubs.py` because cibuildwheel creates a venv for testing with the OpenImageIO wheel pre-installed, and using `uv run` would create a new ad-hoc venv.

chadrik:
@zachlewis  @lgritz   This is ready for review.  I've removed the C++ changes, so I think what I have is quite safe.   It would be good to test out the local workflows on a platform other than MacOS, which is what I'm using.

zachlewis:
This is fantastic. Just a couple of questions inline, but I'm quite happy with everything you've put together. Thank you! This was quite an effort, and I appreciate the time you put into making stub generation as simple as possible to slot into OIIO's existing CI workflows. I'd definitely like to do something similar for OCIO!

cdombrova-scanlinevfx:
@zachlewis did you have some comments you were going to add?  Maybe you forgot to complete the review?

Now that we’ve got a workflow design for this that show work for C++ projects I doubt it would be much work to add this to OCIO.

zachlewis:
Ah, sorry. Yes, all of this looks good to me; although I was wondering if the stubs should use `numpy.typing.NDArray` instead of `numpy.ndarray` directly.

zachlewis (approved):

src/python/stubs/__init__.pyi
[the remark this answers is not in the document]
chadrik: > I'm guessing probably not, since the runtime interpreter doesn't really care

correct

src/python/generate_stubs.py
```diff
@@ -0,0 +1,135 @@
...
+    sys.path.append(out_path)
+    os.environ.setdefault("OPENIMAGEIO_PYTHON_LOAD_DLLS_FROM_PATH", "1")
+    # stubgen.main will performa an import, but we do this now so we can see the traceback
+    # if it fails.
+    import OpenImageIO
+    sys.argv[1:] = ["-p", "OpenImageIO", "-o", out_path]
+    mypy.stubgen.main()
+    dest = os.path.join(out_path, "OpenImageIO", "__init__.pyi")
+    print(f"Renaming to {dest}")
+    os.rename(os.path.join(out_path, "OpenImageIO", "OpenImageIO.pyi"), dest)
```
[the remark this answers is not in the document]
chadrik: This comment is on an outdated commit.  The latest changes use argparse.  

One note on the suggestion:  I'm wary of adding support for wheel patching because it's not something we explicitly need, so it's more complexity without much value.

src/python/stubs/__init__.pyi
```diff
@@ -0,0 +1,1617 @@
+#
+# This file is auto-generated. DO NOT MODIFY!  Run `make pystubs` to regenerate
+#
+
+import numpy
```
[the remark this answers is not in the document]
chadrik: There's no runtime overhead here, but I had a look at `numpy.typing.NDArray` and I think it makes sense here, but I need to test it out.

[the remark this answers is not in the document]
chadrik: from what I can tell, when it comes to static typing the primary difference between `numpy.ndarray` and `numpy.typing.NDArray` comes when you specify types and especially shape/dimensions.  Right now we're not tracking this, though there is a future opportunity to track `ImageSpec` type and dimension similar to `numpy.typing.NDArray`, in which case we could allow mypy to infer the type and shape of an `ImageSpec` from the ndarray used to instantiate it. 

All of that said, `numpy.typing.NDArray` was added in numpy 1.20 and as of this writing OIIO still supports 1.19.  Even though the stubs are only used for analysis and aren't imported at runtime, mypy does rely on the versions of libraries that are installed (e.g. into a venv).   

tl;dr  I think we should use `numpy.ndarray` until OIIO drops support for numpy 1.19.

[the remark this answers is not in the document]
lgritz: > tl;dr I think we should use numpy.ndarray until OIIO drops support for numpy 1.19.

If you'd asked me 5 minutes ago, I wouldn't have been able to tell you what the current version of numpy is or how old 1.19 was, it's just not really my world -- as far as I knew, any old numpy would be fine. Our INSTALL says we support up to 2.2.4 (should probably be updated to 2.2.5 which was just released 2 weeks ago) but doesn't specify a minimum version, which I suppose is just the minimum version supported by pybind11, whatever that turns out to be.

Turns out that 1.20.0 dates from Jan 2021, and VFX Platform specified Numpy 1.20 since CY 2022. Our policy is not to change minimum dependency versions for release branches (3.0.x currently), but if you want to raise the numpy minimum in main (i.e. for 3.1 released later this year), I think that's fine based on its age. Or, if there are particular features you want to disable when older versions are encountered, that's fine too.

Makefile
```diff
@@ -243,11 +243,18 @@ build: config
...
 pystubs: config
 	@ ( cd ${build_dir} ; \
 	    ${CMAKE} --build . --config ${CMAKE_BUILD_TYPE} --target pystubs \
 	  )
 
+# run mypy on python tests to confirm the stubs are working
+test-pystubs: config
+	@ ( uv export --quiet --no-dev --no-build --no-emit-project --no-hashes -o ${build_dir}/requirements.txt ; \
+	    uvx --with-requirements ${build_dir}/requirements.txt mypy==1.15.0 \
+	  )
```
chadrik, line 256: @zachlewis   heads up that I added one more change after your review:  I added type annotations to the python test code so that it can be used to manually test the stubs.  It's not full coverage, and I haven't added it to CI, but I found this useful in vetting the stubs against some real code.

src/python/generate_stubs.py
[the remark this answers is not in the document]
zachlewis: (Completely outdated, by all means, ignore! Definitely agree, no need to patch the wheel, since we're committing the pre-generated stubs to the repo.)

src/python/stubs/__init__.pyi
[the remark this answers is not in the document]
zachlewis: To be honest, the reason we support Numpy 1.19 is cuz it's the oldest numpy version supported for Python 3.9. 
But it sounds like, for our purposes, `numpy.ndarray` should be totally fine!

chadrik:
Everything is green!   I downloaded a wheel from the CI artifacts and confirmed that `__init__.pyi` is there. 

Ready to merge?

I will follow up this PR with one that changes the bindings in order to put the workflow through its paces.

src/python/stubs/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+
+# Setup the pystub target, which is not built by default.
+
+set (_stub_file "${CMAKE_SOURCE_DIR}/src/python/stubs/OpenImageIO/__init__.pyi")
+
+# Note: the python version must be kept in sync with `[[tool.cibuildwheel.overrides]]` in pyproject.toml.
```
nrusch, line 6: Might be worth adding a symmetrical comment in `pyproject.toml` as well.

src/python/stubs/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+# The stubs are generated within a container so the version of python does not need to match
+# the version being built.
```
nrusch, line 8:
```suggestion
# the version of python that OpenImageIO is being built against.
```

src/python/stubs/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+# Note: the version of cibuildwheel should be kept in sync with .github/workflows/wheel.yml
```
nrusch, line 9: Ditto here for comment symmetry?

src/python/stubs/CMakeLists.txt
chadrik: done

src/python/stubs/CMakeLists.txt
chadrik: done

lgritz:
There seems to be a commit with missing DCO.

Aside from that, where do we stand on this? I'm not enough of a Python power user or native speaker to really judge it. What say you, @zachlewis? Is this complete as far as you're concerned, @chadrik?

testsuite/python-texturesys/src/test_texture_sys.py
```diff
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # https://github.com/AcademySoftwareFoundation/OpenImageIO
 
+from __future__ import annotations
```
lgritz, line 7: Just out of curiosity, at which Python version is the annotations stuff included and no longer needs to come from `__future__`?

testsuite/python-deep/src/test_deep.py
```diff
@@ -68,7 +60,7 @@ def print_deep_image (dd, prefix="After init,") :
                 print ()
 
 
-def print_deep_imagebuf (buf, prefix) :
+def print_deep_imagebuf (buf: oiio.ImageBuf, prefix: str) :
```
lgritz, line 63: Note to mostly-self: we should remember to take a pass over the docs and any python examples in the docs and change all the function declarations to use modern python3 syntax with type annotations.

src/python/stubs/OpenImageIO/__init__.pyi
```diff
@@ -0,0 +1,1617 @@
+#
+# This file is auto-generated. DO NOT MODIFY!  Run `make pystubs` to regenerate
+#
+
```
lgritz, line 4: I'm ok with checking this into the repo, but do we have an automated way to detect that something has changed about the python bindings in a way that necessitates a re-generation? Should we add a note to the RELEASING.md guide reminding us to regenerate on every release as part of the release staging process?

testsuite/python-texturesys/src/test_texture_sys.py
chadrik: This can be removed once OCIO drops support for python 3.9, and only supports 3.10+

src/python/stubs/OpenImageIO/__init__.pyi
chadrik: If a developer changes anything in the code that results in the type stubs changes, the `cp311-manylinux_x86_64` 
 and `cp311-manylinux_aarch64` while build jobs in CI will fail, and will print the instructions on how to update the stubs.  I also updated the INSTALL.md with the instructions on how to update the stubs.

testsuite/python-texturesys/src/test_texture_sys.py
lgritz: OK, that will be a while. I think for 3.1 (~Sept release), we will probably raise our minimum from 3.7 to 3.9. But I think we can't go beyond that. Even at my company, I anticipate needing active 3.9 support through the end of the calendar year to be safe.

chadrik:
Yes, I think this is ready to go.  I've also started porting this workflow over to OCIO.

lgritz:
OK, I'd love to just get a very last ok from Zach, then I'm ready to merge.

zachlewis:
> What say you, @zachlewis?

I think everything about this PR is fantastic.

testsuite/python-texturesys/src/test_texture_sys.py
chadrik: The longer answer is that there are ways to avoid adding this import, but IMHO they are not worth the effort.   For example, if you commit to using `typing.List[X]` instead of `list[str]`, and using `typing.Union[X, Y]` instead of `X | Y`.  It's (understandably) difficult for people to keep track of [which version introduced which feature](https://docs.google.com/document/d/1tCX4rXHUST1Y_4EKTSvoqPiTtYVtLCdGJKlKvaxD6JM/edit?tab=t.0), so using `from __future__ import annotations` is a catchall that let's us not care so we can use these newer features even in python 3.7.

chadrik:
One last question:  is it worth cherry-picking this onto other release branches, so that folks can benefit from this even if they're stuck on older versions?   If yes, would I introduce an MR for each release branch?

src/python/stubs/OpenImageIO/__init__.pyi
lgritz: ok, good enough

Makefile
lgritz: That's great. And it inspired me to add this just now as well: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4752

testsuite/python-deep/src/test_deep.py
lgritz: Just added this: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4752

testsuite/python-texturesys/src/test_texture_sys.py
lgritz: No, I'm not worried about the annotation being there at all. It was mostly a curiosity of when in the future we need to remember that they are no longer needed.

lgritz:
> One last question: is it worth cherry-picking this onto other release branches, so that folks can benefit from this even if they're stuck on older versions? If yes, would I introduce an MR for each release branch?

I was planning to backport to 3.0, the currently supported release family. You don't need to do it. All new code goes into main, and I cherry-pick the things that look like they can safely merge into the release branch and make sure CI verifies that they don't break anything.

lgritz:
BTW, don't worry about the failing "bleeding edge" test. That is the test that builds against top-of-tree of several of our dependencies, and it frequently breaks when dependencies change or momentarily introduce bugs on their side. It looks like something changed in freetype today that is responsible. I'll let that keep failing for a few days to see if it's just a freetype regression that they fix on their own.

lgritz:
@chadrik I tried cherry-picking into dev-3.0 but it fails (you were right about it catching changes!) because the stubs aren't quite right and need to be built separately for that branch. I tried the `make pystubs` but actually had some trouble (on my Mac laptop). It's probably something about my setup, but I don't have time to track it down.

So I'd like to request two things from you, if you don't mind:

1. On your end (and in your fork), please cherry pick into dev-3.0, regen the stubs, and then submit the combo as a PR **against dev-3.0** and I'll do that merge.
2. Anticipating that this may come up in the future, not just for me but possibly others, I'm thinking that a good strategy might be to modify the check so that if it fails, it should generate new stubs (can it do that reliably from CI?) and save it as an artifact, so that from the GH page showing the test failure, the user who submitted the PR can download the file, inspect it, and add it to their PR even if they didn't have the capacity to generate it on their end. (This is basically mimicking our "clang-format" test, which when it fails, allows people to download the diffs to apply on their end even if it's not possible for them to actually run "clang-format" with the right version on their own machine.)

lgritz:
@chadrik Not a huge rush on this, BTW. I think we are just aiming to be able to cherry-pick it into dev-3.0 successfully before the next scheduled release from that branch, on June 1.

cdombrova-scanlinevfx:
Yeah, that all seems reasonable.  

If you have a chance to post your error I’d love to check it out. I’d like to make it more bulletproof.

lgritz:
Sure. Here's what happened for me. Keep in mind I have some unclean set of multiple pythons, whatever is on the system, whatever homebrew installed, etc. 
```
$ make pystubs
[1/2] pystubs: Generating python stubs
FAILED: wheelhouse/OpenImageIO/__init__.pyi /Users/lg/code/oiio/oiio.lg/build/wheelhouse/OpenImageIO/__init__.pyi
cd /Users/lg/code/oiio/oiio.lg/build/src/python/stubs && /usr/local/Frameworks/Python.framework/Versions/3.13/bin/python3.13 /Users/lg/code/oiio/oiio.lg/src/python/stubs/generate_stubs_local.py --repo-root /Users/lg/code/oiio/oiio.lg --python-version="3.11" --cibuildwheel-version="2.21.1" --output-dir /Users/lg/code/oiio/oiio.lg/build/wheelhouse
Installed 7 packages in 11ms

     _ _       _ _   _       _           _
 ___|_| |_ _ _|_| |_| |_ _ _| |_ ___ ___| |
|  _| | . | | | | | . | | | |   | -_| -_| |
|___|_|___|___|_|_|___|_____|_|_|___|___|_|

cibuildwheel version 2.21.1

Build options:
  platform: linux
  allow_empty: False
  architectures: aarch64, i686, ppc64le, s390x, x86_64
  build_selector:
    build_config: cp311-manylinux_x86_64
    skip_config:
    requires_python: >=3.8
    prerelease_pythons: True
    free_threaded_support: True
  output_dir: /Users/lg/code/oiio/oiio.lg/build/wheelhouse
  package_dir: /Users/lg/code/oiio/oiio.lg
  test_selector:
    skip_config:
  before_all:
  before_build:
  before_test:
  build_frontend: None
  build_verbosity: 1
  config_settings:
  container_engine: docker
  dependency_constraints: pinned
  environment:
    SKBUILD_CMAKE_ARGS="-DLINKSTATIC=1"
    CXXFLAGS="-Wno-error=stringop-overflow -Wno-pragmas"
    SKBUILD_CMAKE_BUILD_TYPE="MinSizeRel"
  manylinux_images:
    x86_64: quay.io/pypa/manylinux2014_x86_64:2024.09.16-1
    i686: quay.io/pypa/manylinux2014_i686:2024.09.16-1
    pypy_x86_64: quay.io/pypa/manylinux2014_x86_64:2024.09.16-1
    aarch64: quay.io/pypa/manylinux2014_aarch64:2024.09.16-1
    ppc64le: quay.io/pypa/manylinux2014_ppc64le:2024.09.09-0
    s390x: quay.io/pypa/manylinux2014_s390x:2024.09.09-0
    pypy_aarch64: quay.io/pypa/manylinux2014_aarch64:2024.09.16-1
    pypy_i686: quay.io/pypa/manylinux2014_i686:2024.09.16-1
  musllinux_images:
    x86_64: quay.io/pypa/musllinux_1_2_x86_64:2024.09.16-1
    i686: quay.io/pypa/musllinux_1_2_i686:2024.09.16-1
    aarch64: quay.io/pypa/musllinux_1_2_aarch64:2024.09.09-0
    ppc64le: quay.io/pypa/musllinux_1_2_ppc64le:2024.09.09-0
    s390x: quay.io/pypa/musllinux_1_2_s390x:2024.09.09-0
  repair_command: auditwheel repair -w {dest_dir} {wheel}
  test_command:
    *: oiiotool --buildinfo
    cp311-manylinux_x86_64: oiiotool --buildinfo && python {project}/src/python/stubs/generate_stubs.py --out-path '/output' --validate-path '{project}/src/python/stubs/OpenImageIO/__init__.pyi'
  test_extras:
  test_requires:
    *:
    cp311-manylinux_x86_64:
      mypy~=1.15.0
      stubgenlib~=0.1.0

Cache folder: /Users/lg/Library/Caches/cibuildwheel

Here we go!

  File "/Users/lg/.cache/uv/archive-v0/UlJa7mfI6eeKXlO_lDHv5/bin/cibuildwheel", line 12, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/lg/.cache/uv/archive-v0/UlJa7mfI6eeKXlO_lDHv5/lib/python3.13/site-packages/cibuildwheel/__main__.py", line 49, in main
    main_inner(global_options)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/Users/lg/.cache/uv/archive-v0/UlJa7mfI6eeKXlO_lDHv5/lib/python3.13/site-packages/cibuildwheel/__main__.py", line 184, in main_inner
    build_in_directory(args)
    ~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/lg/.cache/uv/archive-v0/UlJa7mfI6eeKXlO_lDHv5/lib/python3.13/site-packages/cibuildwheel/__main__.py", line 352, in build_in_directory
    platform_module.build(options, tmp_path)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/lg/.cache/uv/archive-v0/UlJa7mfI6eeKXlO_lDHv5/lib/python3.13/site-packages/cibuildwheel/linux.py", line 434, in build
    subprocess.run(
    ~~~~~~~~~~~~~~^
        [build_step.container_engine.name, "--version"],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        check=True,
        ^^^^^^^^^^^
        stdout=subprocess.DEVNULL,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/subprocess.py", line 554, in run
    with Popen(*popenargs, **kwargs) as process:
         ~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/subprocess.py", line 1039, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                        pass_fds, cwd, env,
                        ^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
                        gid, gids, uid, umask,
                        ^^^^^^^^^^^^^^^^^^^^^^
                        start_new_session, process_group)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/subprocess.py", line 1969, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'docker'
Building cp311-manylinux_x86_64
  File "/Users/lg/code/oiio/oiio.lg/src/python/stubs/generate_stubs_local.py", line 62, in <module>
    main()
    ~~~~^^
  File "/Users/lg/code/oiio/oiio.lg/src/python/stubs/generate_stubs_local.py", line 43, in main
    subprocess.check_call(
    ~~~~~~~~~~~~~~~~~~~~~^
        [
        ^
    ...<9 lines>...
        cwd=args.repo_root,
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/Cellar/python@3.13/3.13.3/Frameworks/Python.framework/Versions/3.13/lib/python3.13/subprocess.py", line 419, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['uv', 'tool', 'run', '--python=/usr/local/opt/python@3.13/bin/python3.13', 'cibuildwheel@2.21.1', '--output-dir=/Users/lg/code/oiio/oiio.lg/build/wheelhouse', '--only=cp311-manylinux_x86_64', '.']' returned non-zero exit status 1.
ninja: build stopped: subcommand failed.
make: *** [pystubs] Error 1
        3.05 real         1.06 user         0.37 sys
```

lgritz:
Oh yeah, this part sure isn't going to work on my mac:
```
FileNotFoundError: [Errno 2] No such file or directory: 'docker'
```

PR #4694, opened 2025-04-08
Docs: specify the units for DPX scanned size

ideasman42:
From the PDF:
https://pub.smpte.org/doc/st268/20030505-pub/st0268-2003.pdf

3.31 X scanned size: Defines the horizontal size of the original scanned optical image in millimeters. [42.1]
3.32 Y scanned size: Defines the vertical size of the original scanned optical image in millimeters. [42.2]

----

This is a small PR that simply clarifies the units in the documentation.


lgritz (approved):
Thanks for the fix!

PR #4695, opened 2025-04-08
Docs: correct the type for BMP x/y density

ideasman42:
These values are read & written as integers
and don't work unless they are accessed using int accessors.

m_dib_header.hres = m_spec.get_int_attribute("XResolution");
## Description

Correct documentation.

BMP uses integer access:
- `./src/bmp.imageio/bmpinput.cpp:`, `m_spec.attribute("XResolution", (int)m_dib_header.hres);`
- `./src/bmp.imageio/bmpoutput.cpp`, `m_dib_header.hres = m_spec.get_int_attribute("XResolution");`


lgritz (approved):
Thanks for the correction!

PR #4696, opened 2025-04-10
build: Include Windows version information on produced binaries

jessey-git:
## Description

This change embeds version information into our Windows binaries. This is done by ways of a `version_win32.rc` resource file[1] populated at CMake configure time with our major.minor.patch version information.

This not only allows at a glance to discover what version a particular OIIO binary is (see screenshot), but it also impacts certain MSI installer behaviors. If this is a MSI install which updates a prior installation, and the shared libs/dlls don't have version information, then the installer assumes the binaries are equal and just skips them. This can be problematic for applications that package OIIO etc.

Old on left, New on right:
old-vs-new

[1] https://learn.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/build-scripts/version_win32.rc.in
+++ b/src/build-scripts/version_win32.rc.in
@@ -1,0 +1,48 @@
+#include <winver.h>
+
+#define VER_FILEVERSION             @PROJECT_VERSION_MAJOR@,@PROJECT_VERSION_MINOR@,@PROJECT_VERSION_PATCH@,0
+#define VER_FILEVERSION_STR         "@PROJECT_VERSION_MAJOR@.@PROJECT_VERSION_MINOR@.@PROJECT_VERSION_PATCH@.0\0"
+
+#define VER_PRODUCTVERSION          @PROJECT_VERSION_MAJOR@,@PROJECT_VERSION_MINOR@,@PROJECT_VERSION_PATCH@,0
+#define VER_PRODUCTVERSION_STR      "@PROJECT_VERSION_MAJOR@.@PROJECT_VERSION_MINOR@\0"
+
+#ifndef DEBUG
+#define VER_DEBUG                   0
+#else
+#define VER_DEBUG                   VS_FF_DEBUG
+#endif
+
+VS_VERSION_INFO VERSIONINFO
+FILEVERSION     VER_FILEVERSION
+PRODUCTVERSION  VER_PRODUCTVERSION
+FILEFLAGSMASK   VS_FFI_FILEFLAGSMASK
+FILEFLAGS       (VER_DEBUG)
+FILEOS          VOS__WINDOWS32
+FILETYPE        VFT_DLL
+FILESUBTYPE     VFT2_UNKNOWN
+BEGIN
+    BLOCK "StringFileInfo"
+    BEGIN
+        BLOCK "040904E4"
+        BEGIN
+            VALUE "FileDescription",  "OpenImageIO"
+            VALUE "FileVersion",      VER_FILEVERSION_STR
+            VALUE "InternalName",     "OpenImageIO"
+            VALUE "ProductName",      "OpenImageIO"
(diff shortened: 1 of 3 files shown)


lgritz:
Thanks, I never would have known how to do this, or even that it was a thing.

It looks fine to me, inasmuch as it's all enclosed in `if (WIN32)` so can't break any other platforms, and CI passes including on Windows.

But I'm not qualified to know if you've done it right. Would anybody else like to get their eyes on this and weigh in? (I will merge it in a couple days unless somebody objects.)

Should we have something in the testsuite (obviously that only runs for Windows) that validates this in some way? You posted a screenshot from a GUI tool, but is there anything scriptable that can extract that information from the libraries and compare to a reference output, so we can be sure via CI that we don't break this in the future?

ThiagoIze (approved):
This looks essentially good to me. There's a few parts where I don't know enough about this to say if it's totally right. For instance, VarFileInfo is listed as 1252. Why that instead of 1200, which is unicode? I don't know the answer. Maybe it doesn't really matter for our purposes?

I'm approving on the assumption that this is working properly for you in Windows. I do like what the new properties window is showing.

jessey-git:
> Should we have something in the testsuite (obviously that only runs for Windows) that validates this in some way?

It's trivially accessible using Powershell. Let me see if I can make it work somehow.

jessey-git:
> This looks essentially good to me. There's a few parts where I don't know enough about this to say if it's totally right. For instance, VarFileInfo is listed as 1252. Why that instead of 1200, which is unicode? I don't know the answer. Maybe it doesn't really matter for our purposes?

It's a good question that I can't find a definitive answer to. The VarFileInfo block relates to the StringFileInfo block above it; i.e. the `0x409, 1252` == `040904E4` once both fields are hex'd.  I *think* this is basically used to provide language-specific strings in the File properties box if the user happens to be running Windows in a different locale.  Like if the description of the application had translatable elements in it.  I honestly don't know if its better or worse to use Unicode for this one :-/

jessey-git:
Alright, added a test to verify the embedded version information and that seems to pass.  Here's 2 examples of what you might see from the output of the script during both a Pass and a Fail:

Pass
```
182: Test command: C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe "C:/Users/jesse/source/OpenImageIO/testsuite/win32/version_check.ps1" "C:/Users/jesse/source/OpenImageIO/build/bin/RelWithDebInfo/" "3.1.1.0"
182: Working Directory: C:/Users/jesse/source/OpenImageIO/build/src/libOpenImageIO
182: Test timeout computed to be: 1500
182: Arguments : C:/Users/jesse/source/OpenImageIO/build/bin/RelWithDebInfo/ 3.1.1.0
182: Expected : OpenImageIO|3.1.1.0|3.1
182: Actual   : OpenImageIO|3.1.1.0|3.1
182:          : OpenImageIO|3.1.1.0|3.1
1/1 Test #182: win32_versioning .................   Passed    0.43 sec
```

Fail
```
182: Test command: C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe "C:/Users/jesse/source/OpenImageIO/testsuite/win32/version_check.ps1" "C:/Users/jesse/source/OpenImageIO/build/bin/RelWithDebInfo/" "3.1.1.0"
182: Working Directory: C:/Users/jesse/source/OpenImageIO/build/src/libOpenImageIO
182: Test timeout computed to be: 1500
182: Arguments : C:/Users/jesse/source/OpenImageIO/build/bin/RelWithDebInfo/ 3.2.0.0
182: Expected : OpenImageIO|3.2.0.0|3.2
182: Actual   : OpenImageIO|3.1.1.0|3.1
182:          : OpenImageIO|3.1.1.0|3.1
1/1 Test #182: win32_versioning .................***Failed    0.43 sec
```

lgritz (approved):
LGTM!

PR #4697, opened 2025-04-11
fix(sgi): Fix valid_file to properly swap bytes on little-endian platforms

jessey-git:
## Description

The SGI IRIS format is big-endian so incoming multi-byte fields need to be swapped. This wasn't being done in the `valid_file` call so the magic test would always fail on little-endian platforms.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/sgi.imageio/sgiinput.cpp
+++ b/src/sgi.imageio/sgiinput.cpp
@@ -89,6 +89,9 @@ SgiInput::valid_file(Filesystem::IOProxy* ioproxy) const
 
     int16_t magic {};
     const size_t numRead = ioproxy->pread(&magic, sizeof(magic), 0);
+    if (littleendian()) {
+        swap_endian(&magic);
+    }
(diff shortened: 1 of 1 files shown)


lgritz:
How did the old code not fail on little endian? Does this mean that the valid_file was never called?

lgritz (approved):
LGTM, thanks for the fix

lgritz:
BTW, the failure for MacOS 13 is unrelated, it's failing for master, too. Looks like something in homebrew or GH's Mac image changed in the last couple days. I will look into that separately.

jessey-git:
> How did the old code not fail on little endian? Does this mean that the valid_file was never called?

Yeah, there are ways through the ImageInput::create code path that will not call valid_file for you. In our case we happen to call valid_file ourselves anyhow and ended up running into this one. Currently we don't use OIIO for SGI but I was looking to try. Figured it would be an easy friday exercise... hit a bug instead :)

PR #4698, opened 2025-04-12
fix(build): Fix new dcmtk 3.6.9 vs C++ warning

lgritz:
Some new changes in DCMTK 3.6.9 that triggers warnings.


jessey-git (approved):
Yeah, seems this is all we can do until upstream fixes it.

PR #4701, opened 2025-04-14
build: Address Robin-map vs CMake 4.0 compatibility

lgritz:
* Raise auto-build version of Robinmap to 1.4.
* Add the magic CMAKE_POLICY_VERSION_MINIMUM to our auto-build of Robinmap in case an older version is selected for auto-build. (Mimicking the fix we did for other packages.)
* Add Robin-map to the INSTALL.md list of dependencies, somehow it had been overlooked before.


lgritz:
Look ok?

fpsunflower (approved):

PR #4704, opened 2025-04-15
feat(raw): add raw:ForceLoad hint

antond-weta:
## Description

The raw input plugin populates the attributes in the ImageSpec from the libraw structures in the RawInput::open_raw() method. That is done before the actual image data being decompressed on demand in RawInput::read_native_scanline(). Libraw re-calculates some values after the pixels get available, it is impossible to access those, as the ImageBuf's spec does not get updated, and the updated ImageSpec is not available from outside of RawInput when/after the pixels are being processed.

An example: the initial values for the white balancing weights get populated from the image file metadata, those get copied into the ImageSpec. If the libraw was configured to white balance by averaging the pixels of the whole image or a region, the white balancing weights would get re-calculated by libraw after processing the pixels, but the updated numbers are impossible to access from outside of the RawInput plugin.

This change adds a hint "raw:ForceLoad", which, if set to 1, forces libraw to decompress and process the image in RawInput::open_raw(), so the image attributes returned from that method would contain the final values.

I am not certain if this is the best way to achieve what I needed. I have also looked into making the ImageBuf retain the ImageInput object instead of spawning multiple transient objects, but that seemed too complicated.

## Tests

I have not added any unit tests, not sure how to approach that. I have tested that the image does not get decompressed multiple times in a bunch of scenarios. 

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -2228,7 +2228,12 @@ options are supported:
      - Maximum memory allocation for processing of raw images. Stop processing if
        raw buffer size grows larger than that value (in megabytes).
        (Default: 2048)
-
+   * - ``raw:ForceLoad``
+     - int
+     - If 1, forces libraw to decompress and process the image during
+       initialization. This populates the image attributes which depend on the
+       pixel values.
+       (Default: 0)
 
 |
 
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -72,7 +72,6 @@ public:
                               void* data) override;
 
 private:
-    bool process();
     bool m_process  = true;
     bool m_unpacked = false;
     std::unique_ptr<LibRaw> m_processor;
@@ -87,9 +86,10 @@ private:
     std::string m_make;
 
     bool do_unpack();
+    bool do_process();
 
     // Do the actual open. It expects m_filename and m_config to be set.
-    bool open_raw(bool unpack, const std::string& name,
+    bool open_raw(bool unpack, bool process, const std::string& name,
                   const ImageSpec& config);
     void get_makernotes();
     void get_makernotes_canon();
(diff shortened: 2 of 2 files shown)


lgritz (approved):
LGTM. It's a hack, but seems like the best one available to us.

lgritz:
Ick, it's frustrating that you can get metadata from libraw that looks correct upon reading the header, but then changes when you decode the pixels.

Just throwing this out there: Another possible approach -- no less ugly, and I'm not necessarily advocating this -- might be:
1. RawInput modifies its stored imagespec with the changed metadata upon reading the pixels. (Or adds those items with slightly different names that indicate it's a post-read final value, versus a pre-read estimate?)
2. ImageBuf, upon reading the pixels, JUST FOR "raw" images, updates its stored ImageSpec from the ones in in the ImageInput it used for the read (which updated itself in step 1).
It adds a tiny bit of abstraction leaking by making ImageBuf internals need to be aware of what kind of file it's reading from (but we do similar thing in some other isolated areas).

That doesn't really change the fact that an app (even 'iinfo', say) has cases where it CAN'T print all the correct final metadata just by reading the header only, that's the heartbreaking thing about it. The app just has to know that if it might be reading raw files with certain options, it needs to set this extra config hint and be prepared to eat the extra cost of the decode that it doesn't otherwise need.

lgritz:
I'm fine with this as it is. I think it's ready to merge, but let me know if after sleeping on it, you want to stand with this as-is and merge, or to hold off for other changes.

antond-weta:
Thanks Larry! 
I've been sitting on this for over a month already. I'm a bit reluctant to force the user to fetch the pixels if all they needed was the metadata. I'm 'happy' with the change as it is. If you can't think of a more elegant way, let's merge it in.

lgritz:
No, I certainly don't have a more elegant solution. I will merge.

Do you work primarily out of main? We can postpone backporting this to 3.0 for a while, in case a better solution comes to mind later. In main, we're pretty free to try different things and back out of something if we realize it's not the best way. We're a little more cautious about removing features or changing existing behavior in release branches.

PR #4706, opened 2025-04-21
fix: clean up Windows compilation warnings

lgritz:


jessey-git (commented):
I can confirm that most* of the build warnings are gone for me locally. So this looks good from that angle.

*There's another set of warnings that have been popping up for me:
```
9>C:\Users\jesse\source\OpenImageIO\src\include\OpenImageIO\simd.h(6906,66): warning C4556: value of intrinsic immediate argument '8' is out of range '0 - 7'
9>(compiling source file '../../../src/iv/ivmain.cpp')
9>C:\Users\jesse\source\OpenImageIO\src\include\OpenImageIO\simd.h(8797,69): warning C4556: value of intrinsic immediate argument '8' is out of range '0 - 7'
9>(compiling source file '../../../src/iv/ivmain.cpp')
```

This is because we're feeding `_MM_FROUND_NO_EXC` into `_mm_cvtps_ph` and `_mm256_cvtps_ph` which isn't a valid value for that parameter: 
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_cvtps_ph&ig_expand=2110,2107
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_cvtps_ph&ig_expand=2110

lgritz:
@jessey-git Thanks for the tip about simd.h. I will have a separate PR to address that.

jessey-git (approved):
All fine from my side.

PR #4722, opened 2025-04-26
int: Address some nickpick sonar warnings about TileID initialization

lgritz:


jessey-git (approved):
Looks alright to me. You can also `= default` the TileID destructor too as that should be another sonar warning I think. I don't have to see the next revision if you choose to do so though, should be fine.

PR #4723, opened 2025-04-26
fmath.h: add span-based bit_unpack() utility

lgritz:
This also fixes a sonar warning about being unsure of a raw pointer was used out of range.


lgritz:
Any concerns about this?

lgritz:
I will merge this over the weekend if nobody objects or voices concerns by then.

PR #4726, opened 2025-04-27
refactor: Rearrange initialize_cuda() for better err check, no warn

lgritz:
Sonar was complaining that initialize_cuda() contained a loop that would only iterate once. Upon looking, this was true, but mainly because we just didn't adequately check errors. Fix up, which should both make the code more robust and also silence the Sonar complaint.


lgritz:
I will merge this at the end of the week if nobody has comments or objections.

PR #4729, opened 2025-04-29
build: Detect libultrahdr version and enforce minimum of 1.3

lgritz:
I discovered, by testing against the latest libultrahdr, that only starting in 1.3 does it support the "64 bit 'half' RGBA" variety of UHDR that is the flavor used by the testsuite example. But our "auto-build" is pegged to 1.2, which means that all along, we were not decoding that test file properly as UHDR and our test reference output was therefore wrong. We also didn't detect libuhdr's version at all, let alone exclude versions that were too old.

Remediations:

* Amend Findlibuhdr.cmake to detect the version from symbols in its header files.

* Have our search for libuhdr enforce a minimum version of 1.3. (I feel bad about "changing minimum dependencies" in a release branch, which we usually disallow, but I think this time is an exception since we know versions < 1.3 are broken.)

* Change the auto-build selected version to 1.4, the latest.

* Mention libuhdr in the INSTALL description of dependencies and acknowledgement in the docs, where it had previously been omitted in those places.

* Update the testsuite/jpeg-ultrahdr test reference output. (You can tell it was wrong before because the "max" channel value was 1.0, and now it's 4.6.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -74,6 +74,8 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
      * Ptex >= 2.3.1 (probably works for older; tested through 2.4.3)
  * If you want to be able to do font rendering into images:
      * Freetype >= 2.10.0 (tested through 2.13)
+ * If you want to be able to read "ultra-HDR" embedded in JPEG files:
+     * libultrahdr >= 1.3 (tested through 1.4)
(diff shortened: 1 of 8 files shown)


lgritz:
@mugulmd Can you take a look and give a review, please?

lgritz:
Any objections from anyone?

lgritz:
No objections -> merging

PR #4730, opened 2025-04-29
build: fix fmt throwing behavior warnings

lgritz:
Certain version combinations of fmt and gcc were warning, and it all boiled down to the knots we tied ourselves into to replace the function fmt uses instead of throwing exceptions, when they are disabled. I can't seem to find any harm in letting it do its default thing, so simplify our side to silence the warnings.


lgritz:
This fixes some build problems on certain toolchain version combinations. Anybody have concerns?

lgritz:
No objections for more than a week -> merging

PR #4731, opened 2025-04-30
ci: save time by not checking out entire project history

lgritz:
Seems to shave about 5s off of each and every CI run variant. Every little bit helps.

Also update the versions of some actions we're using to the latest.


lgritz:
No comments for 4 days + only affects CI (not code or behavior) -> I will merge this by the end of the weekend, if there are no objections voiced.

PR #4735, opened 2025-05-02
build: Restore OIIO_AVX512ER_ENABLED preprocessor symbol

lgritz:
This useless symbol was removed in PR #4724, but maybe downstream things used it and will break simply because it was removed, even though it never would have been nonzero. It should not have been completely removed except at a compatibility-breaking version boundary.


lgritz:
Merging PRs without review and without leaving at least several days for reviews to occur is considered bad in all but the most pressing cases, but I'm going to do it in this case because it's repairing an unintentional compatibility break that affects downstream projects and is in fact breaking OSL CI and holding up a release. If somebody has an objection after the fact, please speak up and we will amend.

PR #4736, opened 2025-05-03
docs: Online docs improvements, mostly formatting

lgritz:


lgritz:
Docs only change, no objections -> merging

PR #4737, opened 2025-05-03
int: Address safety warnings in pvt::append_tiff_dir_entry

lgritz:
I think what we were doing here was fine all along, but the idiom was confusing to static analyzers who identified a danger that we were memcpy'ing into a field that was potentially not big enough. A minor restructuring of the code and a new assertion should verify that it's safe and also make it clear to the static analyzer that we aren't falling into the case it warned about.


scott-wilson:
Looks good to me. Did a test compile, and it compiled without issues.

scott-wilson (approved):

PR #4738, opened 2025-05-04
Handle various bits per sample values in JPEG XL files

1div0:
I was wondering why the 12 bits per channel images were rejected by the image viewer?

## Description

Extend the ranges of bits per sample.


lgritz:
If you get something like 12 bps, are you just leaving those values as they are but storing them in a uint16? So, like, fully bright is 4095?

The design philosophy we adopted is that the app shouldn't need to consider every possible bit depth. On the app side of the interface, we pretend all values are UINT8, UINT16, HALF, etc. So the OIIO reader is expected to "round up" and promote a "non-whole" bit depth to the next bigger whole one -- e.g., 12 bit in the file becomes UINT16. But we want it to look like a proper UINT16, so what we do is scale the values up so that the max brightness is always 65535 for UINT16 data.

Also when we do this scaling, we set in the ImageSpec an attribute "oiio:BitsPerSample" that gives the origin number of bits in the file.

You can see how this works in tiffinput.cpp, for example. Look circa line 1053 (in current main), and 1859 for where the bit depth scaling takes place.

lgritz:
@1div0 ping on this, I had made some comments/questions

1div0:
Last two weeks were extremely busy.

The various bit depth in JPEG XL image files in the range from 1 to 31 bits per channel are handled by the decoder in the reference libjxl transparently according to selected data type. That means 8, 16 and 32 bits per channel in output from decoder are scaled properly, so there is not needed to do more.

In other words, 12 bpc images are scaled to next bigger type UINT16.

src/jpegxl.imageio/jxlinput.cpp
```diff
@@ -249,20 +250,18 @@ JxlInput::open(const std::string& name, ImageSpec& newspec)
...
-                break;
-            case 32:
+                bits          = 16;
+            } else if (info.bits_per_sample <= 32) {
                 jxl_data_type = JXL_TYPE_FLOAT;
                 m_data_type   = TypeDesc::FLOAT;
-                break;
-            default: errorfmt("Unsupported bits per sample\n"); return false;
+                bits          = 32;
             }
```
lgritz: We're getting at least one CI failure for gcc11 that is pointing out an error where 'bits' may be uninitialized. I would recommend one more final `else` clause here that calls errfmt and returns false, catching any situation where info.bits_per_sample could possibly have a value > 32.

1div0: Will take care of it in the evening.

jessey-git:
Gentle nudge. How's this PR going?

src/jpegxl.imageio/jxlinput.cpp
lgritz:
```suggestion
                bits          = 32;
            } else {
                errorfmt("Unsupported bits per sample\n");
                return false;
            }
```

I think that this restores the old catching of invalid bps values.
@1div0 or @jessey-git, if you want to ok and hit the button to add this minor correction, we can see if it passes CI, and if it does, I believe this may be ready to merge.

src/jpegxl.imageio/jxlinput.cpp
```diff
@@ -221,6 +221,7 @@ JxlInput::open(const std::string& name, ImageSpec& newspec)
     JxlPixelFormat format;
     JxlDataType jxl_data_type;
     TypeDesc m_data_type;
+    uint32_t bits;
```
lgritz: I think you're going to need this to finally make gcc shut up:

```suggestion
    uint32_t bits = 0;
```

1div0:
C001! Finally all checked. Lesson learned. Will do better next time_t.

Good Night. And Good Luck.

lgritz (approved):

PR #4739, opened 2025-05-04
simd.h: renaming and fixing of shuffle template

lgritz:
While I was investigating some SonarQube warnings about our simd `shuffle<>` templates (which are a false positive and I have a separate PR to simply silence it), I did get to thinking about the naming, and want to make a couple changes:

* I decided that the 1-template-argument version of this function, `shuffle<int i>(simd_type)` would actually be more clear and self-documenting if renamed `broadcast_element` to emphasize that it is taking just one simd lane/element and broadcasting it to all lanes. (The multi-argument shuffle really is doing a true shuffle, giving an index for each lane to make a permutation of swizzle, so I'm not renaming that one.) To avoid breaking source compatibility, I am leaving the old name as well as a synonym, but commenting it as deprecated and I will phase out its use. It will disappear entirely from a future OIIO version that's safe to break compatibility.

* For 16-wide simd, the 1-arg template we called shuffle was not doing the same operation -- it was replicating one group of 4 elements instead of a single element. We didn't use it anywhere, so I'm redefining it to do the analogous thing as it does for 4-wide and 8-wide.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -614,8 +614,9 @@ private:
 template<int i0, int i1, int i2, int i3>
 OIIO_FORCEINLINE vbool4 shuffle (const vbool4& a);
 
-/// shuffle<i>(a) is the same as shuffle<i,i,i,i>(a)
-template<int i> OIIO_FORCEINLINE vbool4 shuffle (const vbool4& a);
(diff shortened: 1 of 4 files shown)


lgritz:
This has been posted for a week. If there are no objections by the end of Monday, I will merge.

PR #4743, opened 2025-05-07
docs: Fix furo requirement after recently breaking docs

lgritz:
PR #4736 bumped the version of sphinx we use on readthedocs, and that was incompatible with the old version of furo we specified.

I'm not sure how the problem was not caught by our "docs" workflow.


lgritz:
Merging immediately because this fixes broken docs and makes no changes except to readthedocs configuration.

PR #4744, opened 2025-05-08
ci: new variants for VFXP 2025, Windows 2025

lgritz:


lgritz:
CI only change, demonstrably passes. Would be great to get an approval so we can merge this right away.

jessey-git (approved):
Looks ok to me.

PR #4745, opened 2025-05-10
ci: add facility for benchmarking as part of CI

lgritz:
Make it so that CI test cases that set the GHA variable "benchmark" to 1 will add a benchmarking step to the workflow that runs a new ci-benchmark.bash script.

The script runs selected unit tests containing benchmarks (currently, only image_span_test, but we can amend later as needed). Those designated tests are run, and their output both echoed to the log for that step and also put in build/benchmarks/TESTNAME and saved as a build artifact for optional download.

Most test cases will not turn benchmarking on -- it probably will end up adding a few minutes so do it very selectively (once per major platform or compiler version is plenty).

I would have previously guessed that any attempts at benchmarking on GHA runners was doomed, but in practice, I'm surprised to find that there's almost as much run-to-run consistency as I find doing casual benchmarks on my own machine. As such, I think this can be a handy way to do some rough benchmarking using CI, to compare platforms or compilers, or verify that changes we want to make don't introduce performance regressions.

Caveats to remember in the future:

* Take it all with a big grain of salt, and watch the benchmark numbers for the trial-to-trial range of times -- wide variation means that the numbers probably can't be trusted.
* The GH runners themselves may change without warning, so beware benchmark stability over time, or if they ever have pools of heterogeneous machine generations/configurations.
* While my results indicate a decent amount of timing reliability for purely computational tests, I assume that there will be enormous run-to-run variation in anything involving I/O or networking. So this is unlikely to be a fruitful way of testing for performance regressions in image format I/O speed (but probably is useful for a variety of in-memory operations).
* As we add more unit tests to what we benchmark in the future, keep an eye how much time we're spending running these benchmarks.  A few minutes on a small subset of the test jobs is probably fine, but I wouldn't want it to make the overall wait for a full CI run to become substantially longer because of it.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -50,6 +50,8 @@ on:
         type: string
       abi_check:
         type: string
+      benchmark:
+        type: string
       build_docs:
         type: string
       clang_format:
@@ -145,6 +147,10 @@ jobs:
         if: inputs.skip_tests != '1'
         shell: bash
         run: src/build-scripts/ci-test.bash
+      - name: Benchmarks
+        if: inputs.benchmark == '1'
+        shell: bash
+        run: src/build-scripts/ci-benchmark.bash
       - name: clang-format
         if: inputs.clang_format == '1'
         shell: bash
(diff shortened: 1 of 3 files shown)


lgritz:
Any objections or comments before I merge this?

lgritz:
Over a week in review, no objections, CI-only change ==> merge

PR #4747, opened 2025-05-11
ci: Bump abi_check standard after ImageOutput changes

lgritz:


lgritz:
Merging immediately because it fixes CI to complete the last PR #4727

PR #4750, opened 2025-05-12
fix: ImageInput/ImageOutput did not set per-file threads correctly

lgritz:
II and IO were documented as initializing their internal "how many threads to use when reading/writing this particular file" to the global "threads" OIIO attribute. But they did not, they always initialized to 0 ("all the threads") even if the global attribute had been set differently prior to creation of the reader or writer.


lgritz:
Any objections here?

EmilDohne:
None, lgtm

EmilDohne (approved):

PR #4751, opened 2025-05-13
fix(exr): did not properly allocate 'missingcolor' vector

lgritz:
Oops, this internal vector called `reserve()` and then filled in results. Should have been `resize()`.


lgritz:
No comments for 2 days and badly needed fix -> merging

PR #4753, opened 2025-05-14
feat(python): _repr_png_ implementation for ImageBuf

### LINKED ISSUE #4680: [FEATURE REQUEST] ImageBuf should know how to display itself to a Jupyter Notebook (IPython)
As it is, when you do `oiio.ImageBuf("input.tif")` in your Jupyter Notebook, it displays as this: `<OpenImageIO.OpenImageIO.ImageBuf at 0x77d2b912dbb0>`

...really not the most illuminating representation, and a particularly big letdown compared to a `PIL.Image.Image`, which just _displays the image_ in the notebook.

IPython has a protocol for how to customize the display of your objects: https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

In short, the Python class for ImageBuf should have a `_repr_png_()` method that returns the PNG-encoded image data. (Or perhaps the `_repr_mimebundle_` method so it could return whichever of WebP, PNG, or JPEG seems appropriate.)

It's possible to add a third-party hook to get this effect, but notebooks are popular enough that its worth considering supporting them out of the box.

(I [hacked](https://gist.github.com/keturn/daed603ffa95b4d33c3cd8a1baa77cc4) together proof of concept for this, but it got pretty kludgey when I couldn't figure out how to use ioproxy from Python.)

Glowies:
Fixes #4680

## Description
Adds the ability for ImageBuf objects to be displayed in Jupyter notebooks, as requested in #4680 
Uses the IOVecOutput IOProxy to write the ImageBuf as a png into memory. Then, reads and returns that as a python bytes object.
image

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (changes requested):
Looks great! Simple, really cool feature that unlocks some new use cases, and it's also a great example of IOProxy to write image files to memory.

LGTM on the code!

I would like to ask for you to add something in the docs to the python chapter? This needs to be more discoverable by users. Also, is there any way to test this as part of the testsuite?

Glowies:
> Just thinking aloud... if we're writing 8bit pngs for the sake of ephemeral preview, we should enable dithering as well. 

I'll look into adding dithering there! Not exactly sure how to do it, but I'm sure its somewhere in the docs 😀 

And a follow-up issue for the Display/View transforms sounds like a great idea. Definitely sounds more involved than I thought!

lgritz:
This is a great start (modulo needing a mention in the docs). Don't feel like you need to keep piling on additional features into the first PR. I'm a big fan of small incremental PRs.

OIIO's png output already has a provision for dithering the output when converting from float to uint8 upon output. All you have to do (I think) is add the attribute "oiio:dither" (with a nonzero value) to the metadata, which you can also do through the ImageBuf interface.

lgritz:
> For the testsuites, I think I can add some very simple tests to see if the correct bytes get output from _repr_png_ and that it doesnt crash for 'dimensionless' images.

Yes, exactly what I had in mind.

zachlewis:
> Don't feel like you need to keep piling on additional features into the first PR. I'm a big fan of small incremental PRs.

(Totally agree, feel free to ignore me!)

lgritz:
The dither addition looks right.

src/doc/pythonbindings.rst
```diff
@@ -35,6 +35,38 @@ of brevity, we will alias the package name as follows:
 
 .. _sec-pythontypedesc:
 
+
+Jupyter Notebooks
+=================
+
+Like any other Python package, OpenImageIO can be used in `Jupyter notebooks <https://jupyter.org/install>`_. 
+The ImageBuf objects support getting displayed inline within notebooks.
+
```
lgritz, line 44: This is a very odd place for this discussion to go -- it's literally inside the section that's talking about the TypeDesc class.

I would recommend that in the ImageBuf section of this chapter, you put a brief description for the `_repr_png_` method itself to document its presence. 

Then move this part you've written here, which is really a "how to" for incorporating OIIO into Jupyter Notebooks and viewing the results, to the *very end* of the chapter, because it actually fits well in the section called "Python Recipes".

lgritz: The text of what you've written here is great. I just think you should move it to the end of the "Recipes" section.

src/doc/pythonbindings.rst
```diff
@@ -4027,6 +4027,43 @@ following boilerplate:
 
 |
 
+
+Jupyter Notebooks
+-----------------
+
+Like any other Python package, OpenImageIO can be used in `Jupyter notebooks <https://jupyter.org/install>`_.
```
lgritz, line 4034: Hi, sorry, one more request for change. 
The "recipes" section is progressing from simple common tasks to more rare or complex examples.
So instead of making this the first item in the section, can you literally make it the very last example (and thus the last text in this file)? Thanks.

src/doc/pythonbindings.rst
```diff
@@ -4260,3 +4260,37 @@ add an alpha channel that is 1 everywhere**
             out.open ("multipart.exr", specs[s], "AppendSubimage")
         bufs[s].write (out)
     out.close ()
+
+
+
+Jupyter Notebooks
+-----------------
```
lgritz, line 4267: Fixing some appearance issues..

```suggestion
|

**Running OpenImageIO in Jupyter Notebooks and displaying ImageBuf**

```

src/doc/pythonbindings.rst
```diff
@@ -4260,3 +4260,37 @@ add an alpha channel that is 1 everywhere**
...
+
+.. warning::
+    
+    Currently, ImageBuf objects get displayed as **uint8 PNGs** inside of notebooks.
+    ImageBuf objects that store images with higher bit depths get dithered to account for this.
+    Keep in mind that directly saving the inline image to disk will not preserve the original image within the ImageBuf.
+
+
+Running a Local Jupyter Notebook
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
lgritz, line 4282:
```suggestion
Running a Local Jupyter Notebook:

```

lgritz (approved):
After previewing the docs, I made two extremely small changes to for you, didn't change any text, just to fix it so the formatting was more like the surrounding section.

The rest all looks good to me, I think we are ready to merge this cool new feature!

PR #4754, opened 2025-05-14
Upload python stubs as an artifact in CI

chadrik:
Provides a path for users who can't build the stubs locally

## Description

if the stub check fails, it should generate new stubs and save it as an artifact, so that from the GH page showing the test failure, the user who submitted the PR can download the file, inspect it, and add it to their PR even if they didn't have the capacity to generate it on their end. (This is basically mimicking our "clang-format" test, which when it fails, allows people to download the diffs to apply on their end even if it's not possible for them to actually run "clang-format" with the right version on their own machine.)

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -163,7 +163,11 @@ jobs:
       - uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
         with:
           name: cibw-wheels-${{ matrix.python }}-${{ matrix.manylinux }}
-          path: ./wheelhouse/*.whl
+          path: |
+            ./wheelhouse/*.whl
+            ./wheelhouse/OpenImageIO/__init__.pyi
+        # if stub validation fails we want to upload the stubs for users to review
+        if: success() || failure()
 
   # ---------------------------------------------------------------------------
   # Linux ARM Wheels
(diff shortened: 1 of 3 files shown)


lgritz:
I'm testing this now on my 3.0 backport...

lgritz:
Is it possible... that this change is making it no longer actually fail the test when the stub file is incorrect?

When I previously cherry-picked your merged stub PR into dev-3.0, I failed the "wheels" CI because the stub file was wrong in that branch due to divergence of some APIs from main, I guess.

My plan was to test by cherry-picking this PR on top of that, push to my fork, run CI, download those artifacts, and then check the stub file into that branch. 

But upon cherry-picking this PR and pushing... now I seem to fully pass CI. Maybe the `continue-on-error: true` is not the right approach?

chadrik:
> Is it possible... that this change is making it no longer actually fail the test when the stub file is incorrect?  Maybe the continue-on-error: true is not the right approach?

Yeah, seems like the wrong approach.   I just pushed a different method.

lgritz:
It did, however, generate a stubs file that I can use for 3.0!
So that part of it definitely works as hoped.

lgritz:
Works!

Would you mind fixing the DCO, please? It won't let me merge without it.

lgritz (approved):
Works and LGTM, thanks for jumping on this!

Issue #4598, opened 2025-01-16
[FEATURE REQUEST] oiiotool to be able to create output directories that don't yet exist


tanant:
**Is your feature request related to a problem? Please describe.**
When running oiiotool to do some reformatting work I noticed that if I specify an output path, where the directory does not exist oiiotool will perform the operation and consume the time but then fail on the output step _after_ the work is done. 

As a trivial example:
```
B:\> oiiotool.exe -v -i .\inputfile.exr -o subdirectory\outputfile.exr 
oiiotool ERROR: -o : Could not open "subdirectory\outputfile.e204dc78.temp.exr" (No such file or directory)
Full command line was:
> oiiotool.exe -v -i ./inputfile.exr -o subdirectory\\outputfile.exr 
```

When combined with multiple operations the impact is a bit worse as I can potentially get partial completion as the stack is evaluated in order and terminates at the first error, in this case (which is slightly contrived, I admit) I'd take the source image, resize to 500px, output to a subdir, then output the jpg version of that alongside the full resolution, and finally make a jpg version from the full res. 
```
oiiotool.exe -i .\inputfile.exr --label "source" -o outputfile.exr --resize 500x0 -o reformatted\outputfile.exr  -o outputfile.jpg -i source -o another_version.jpg
```
I'll get outputfile.exr, it'll then consume the time to calculate the resize and then fail on the output for the missing subdirectory and terminate at that point. 

**Describe the solution you'd like**
I'd ideally like to pass a `--make-paths` flag in the same way i can pass `--no-clobber` so that the output stage would attempt to make the full path provided. If that failed, then behaviour as current where the remainder of the command for that frame doesn't get executed.

In the same vein it'd be nice to have a way to fast-fail if it knew the output directory wasn't there and `--make-paths` wasn't supplied to save on computation time and IO _but_ due to expressions that could depend on a characteristic of the image I'm not sure this is feasible (but would be nice). Wouldn't be that much of a saving overall, but I would prefer to have an all-or-nothing for the command as partial completion is harder to detect (and since it's a behaviour change from defaults it'd probably need to go behind a flag incase people are relying on current partial behaviour)

**Describe alternatives you've considered**
Another option would be to pre-create directories or do a saning of the input command but I don't think this feels correct in the spirit of oiiotool. and due to the expression capability, is also I suspect fraught.

**Additional context**
[...]

PR #4756, opened 2025-05-15
feat(oiio): Added preflight action process for output action

DharshanV:
## Description
This PR is for a feature request: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4598.
It was noted in the issue that running `oiiotool` with output filename inside a folder that doesn't exist will cause OIIO to consume time and resources, but fail eventually due to the folder not existing.

This PR address this problem by having a new flag `--pre-flight`, that run's thru all the command line arugments, calling a `pre_action` step to the original `action` step.

Take for example the command:
```
oiiotool --pre-flight -i input.png -o ./output_folder/output.png -o ./output_folder_2/output.png
```

In the `-o` action step, the pre output action (`pre_output_file`) will create both the `./output_folder` and `./output_folder_2` folders, and the whole operations completes without failure.

In the future, this can be further extended to other actions that can do validation before the main action (`-i` etc).

## Tests
I can look into adding unit test, but tested it locally, and it creates both the above folders and outputs the image.

## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/include/OpenImageIO/argparse.h
+++ b/src/include/OpenImageIO/argparse.h
@@ -277,6 +277,8 @@ public:
     /// @defgroup Parsing arguments
     /// @{
 
+    void parse_args_preflight(int argc, const char** argv);
+
     /// With the options already set up, parse the command line. Return 0 if
     /// ok, -1 if it's a malformed command line.
     int parse_args(int argc, const char** argv);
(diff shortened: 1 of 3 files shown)


DharshanV:
I initially did have just the a new cmd flag `--create-dir`, that will create the directory if needed during the output action.
After, posting in Slack and trying to figure out how to parse the whole cmd line args, the `--pre-flight` came up, and I thought it was good idea (also can be extended in the future if needed).

But I can close this PR and branch, and this is the new PR for `--create-dir` cmd line arg: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4762.

Also, I didn't want to purse the `-n` and a modifier to `-o` because:
1. The user needs to run with `-n` first, and run again without `-n` to get the output, causing them to waste time.
2. Having a modifier on individual `-o` might be noisy and user may forget, causing whole operation to fail.

lgritz:
> The user needs to run with -n first, and run again without -n to get the output, causing them to waste time.

This is the part I didn't understand. How is that different than the proposed `--pre-flight`?

PR #4757, opened 2025-05-15
fix(exr): not honoring 'missingcolor' for scanline files

lgritz:
As a reminder, the global OIIO attribute "missingcolor", or the file-specific ImageInput open hint "oiio:missingcolor" is designed to treat missing scanlines or tiles as not a true error, but instead just pass over them and substitute the missingcolor value for those missing areas on the file.

The nature of this PR is that while we had implemented this behavior fully for tile-based exr files, we never did so for scanline-based files. So here we do.

Now then, unfortunately, it does not work. I think it's a bug in the OpenEXR library itself, but we're still trying to figure that out. But nonetheless, I believe this change is correct.

### DIFF
--- a/src/openexr.imageio/exr_pvt.h
+++ b/src/openexr.imageio/exr_pvt.h
@@ -220,6 +220,10 @@ private:
         m_missingcolor.clear();
     }
 
+    bool read_native_scanlines_individually(int subimage, int miplevel,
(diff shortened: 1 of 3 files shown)


lgritz:
No objections for two weeks -> merging

PR #4758, opened 2025-05-15
build: fix failed test with old fmt

lgritz:
Our fmath_test (a.k.a. unit_fmath) test was failing with an exception thrown by the fmt library because we used a width specifier when formatting a TypeDesc.

It was really only an issue when using old fmt < 10.0, so I'm just solving it trivially by ignoring that specifier for old fmt. It doesn't seem necessary to make it work correctly with width specifiers for old versions, when we don't in any way rely on it.


lgritz:
Pre-emptively merging because this fixes an important test breakage for some people using old versions of fmtlib. (But can't possibly break anybody using more recent versions, since the changes are entirely inside an `#if` that is only active for those old versions.)

PR #4762, opened 2025-05-16
feat(oiiotool): Added create-dir cmd arg to create dir if it doesn't exist

### LINKED ISSUE #4598: [FEATURE REQUEST] oiiotool to be able to create output directories that don't yet exist

DharshanV:
Fixes #4598.

## Description

It was noted in the issue that running `oiiotool` with output filename inside a folder that doesn't exist will cause OIIO to consume time and resources, but fail eventually due to the folder not existing.

Take for example the command:
```
oiiotool --create-dir -i input.png -o ./output_folder/output.png -o ./output_folder_2/output.png
```

During the output action, it creates the directories if needed (`output_folder` and `output_folder_2`), and outputs the images into the folder.

## Tests
Added unit tests. Tested it locally, and it creates both the above folders and outputs the image.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5514,6 +5514,18 @@ output_file(Oiiotool& ot, cspan<const char*> argv)
         return;
     }
 
+    if (ot.createdirectories) {
+        const std::string outputdirpath = Filesystem::parent_path(filename);
+        if (!Filesystem::exists(outputdirpath)) {
+            const bool result = Filesystem::create_directory(outputdirpath);
+            if (!result) {
+                ot.errorfmt(command, "Failed to create output directory: {}",
+                            outputdirpath);
+                return;
+            }
+        }
+    }
+
     if (ot.noclobber && Filesystem::exists(filename)) {
         ot.warningfmt(command, "{} already exists, not overwriting.", filename);
         return;
@@ -6458,6 +6470,8 @@ Oiiotool::getargs(int argc, char* argv[])
       .help("Do not overwrite existing files");
     ap.arg("--noclobber", &ot.noclobber)
       .hidden(); // synonym
+    ap.arg("--create-dir", &ot.createdirectories)
+      .help("Create output directories if it doesn't exists");
     ap.arg("--threads %d:N")
       .help("Number of threads (default 0 == #cores)")
       .OTACTION(set_threads);
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5514,15 +5514,23 @@ output_file(Oiiotool& ot, cspan<const char*> argv)
         return;
     }
 
-    if (ot.createdirectories) {
-        const std::string outputdirpath = Filesystem::parent_path(filename);
-        if (!Filesystem::exists(outputdirpath)) {
-            const bool result = Filesystem::create_directory(outputdirpath);
-            if (!result) {
-                ot.errorfmt(command, "Failed to create output directory: {}",
-                            outputdirpath);
+    const std::string outputdirpath = Filesystem::parent_path(filename);
+    if (!outputdirpath.empty() && !Filesystem::exists(outputdirpath)) {
+        if (ot.create_dir) {
+            std::string err;
+            const bool ok = Filesystem::create_directories(outputdirpath, err);
+            if (!ok) {
+                ot.errorfmt(command,
+                            "Failed to create output directory: {}\n\tError: {}",
+                            outputdirpath, err);
                 return;
             }
+        } else {
+            ot.errorfmt(command,
+                        "Non-existent output directory: {}\n"
+                        "\t--create-dir to create missing output directories",
+                        outputdirpath);
+            return;
         }
     }
 
@@ -6470,7 +6478,7 @@ Oiiotool::getargs(int argc, char* argv[])
       .help("Do not overwrite existing files");
     ap.arg("--noclobber", &ot.noclobber)
       .hidden(); // synonym
-    ap.arg("--create-dir", &ot.createdirectories)
+    ap.arg("--create-dir", &ot.create_dir)
       .help("Create output directories if it doesn't exists");
     ap.arg("--threads %d:N")
       .help("Number of threads (default 0 == #cores)")
--- a/src/oiiotool/oiiotool.h
+++ b/src/oiiotool/oiiotool.h
@@ -77,11 +77,12 @@ public:
     int cachesize;
     int autotile;
     int frame_padding;
-    bool eval_enable;              // Enable evaluation of expressions
-    bool parallel_frames = false;  // Parallelize over frame iteration
-    bool skip_bad_frames = false;  // Just skip a bad frame, don't exit
-    bool nostderr        = false;  // If true, use stdout for errors
-    bool noerrexit       = false;  // Don't exit on error
+    bool eval_enable;                // Enable evaluation of expressions
+    bool parallel_frames   = false;  // Parallelize over frame iteration
+    bool skip_bad_frames   = false;  // Just skip a bad frame, don't exit
+    bool nostderr          = false;  // If true, use stdout for errors
+    bool noerrexit         = false;  // Don't exit on error
+    bool createdirectories = false;
     std::string dumpdata_C_name;
     std::string full_command_line;
     std::string printinfo_metamatch;
--- a/src/oiiotool/oiiotool.h
+++ b/src/oiiotool/oiiotool.h
@@ -77,12 +77,12 @@ public:
     int cachesize;
     int autotile;
     int frame_padding;
-    bool eval_enable;                // Enable evaluation of expressions
-    bool parallel_frames   = false;  // Parallelize over frame iteration
-    bool skip_bad_frames   = false;  // Just skip a bad frame, don't exit
-    bool nostderr          = false;  // If true, use stdout for errors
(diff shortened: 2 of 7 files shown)


lgritz:
Don't worry about the failing "bleeding edge" test. It's unrelated to your PR, and I have a separate patch already submitted that addresses the problem.

You need a DCO signoff of your commits.

src/oiiotool/oiiotool.h
```diff
@@ -77,11 +77,12 @@ class Oiiotool {
...
-    bool parallel_frames = false;  // Parallelize over frame iteration
-    bool skip_bad_frames = false;  // Just skip a bad frame, don't exit
-    bool nostderr        = false;  // If true, use stdout for errors
-    bool noerrexit       = false;  // Don't exit on error
+    bool eval_enable;                // Enable evaluation of expressions
+    bool parallel_frames   = false;  // Parallelize over frame iteration
+    bool skip_bad_frames   = false;  // Just skip a bad frame, don't exit
+    bool nostderr          = false;  // If true, use stdout for errors
+    bool noerrexit         = false;  // Don't exit on error
+    bool createdirectories = false;
```
lgritz, line 85: I know this is super picky, but I think it would be a little more readable as `create_directories`. Or `create_dir` is also fairly clear and corresponds very directly to the command line argument?

src/oiiotool/oiiotool.cpp
```diff
@@ -5514,6 +5514,18 @@ output_file(Oiiotool& ot, cspan<const char*> argv)
...
+        const std::string outputdirpath = Filesystem::parent_path(filename);
+        if (!Filesystem::exists(outputdirpath)) {
+            const bool result = Filesystem::create_directory(outputdirpath);
+            if (!result) {
+                ot.errorfmt(command, "Failed to create output directory: {}",
+                            outputdirpath);
+                return;
+            }
+        }
+    }
```
lgritz, line 5527: What happens now when this new flag is not used, and the directory path doesn't exist? Do you get an understandable error message?

Aha, it just says "cannot create [filename]" but doesn't say why.

So what do you think about reversing the order of these "if" statements, turning it inside out so the existence test is on the outside:

    if (!exists(outputdirpath)) {
        if (ot.create_directories) {
            // create_cirectory(), and error if you can't
        } else {
            // error with clear message about missing directory
            // including alerting user about --create-dir
        }
    }

lgritz (changes requested):
This is good. I made a couple fairly minor suggestions for improvement.

Also, it would be helpful if you could add a test of this feature to the testsuite. Maybe just two additional invocations in testsuite/oiiotool/run.py, testing something simple like

    oiiotool --create 2x2 1 -o 1/2/out.tif
    # ^^ should make the right error message as output

    oiiotool --create-dir --create 2x2 1 -o 1/2/out.tif
    oiiotool --info 1/2/out.tif
    # ^^ should make the right output showing that the file is found

Also, the new command really needs to be documented in src/doc/oiiotool.rst in an appropriate section.

lgritz:
Hi, @DharshanV just checking in on this. I had made a couple minor suggestions and wondered what you think. I believe we are very close to being able to wrap this up and merge it.

DharshanV:
Hello @lgritz! Sorry, just been away the whole weekend! I might also be away Monday.

But, I do really want to get this in and will address your comments once I am back. Thanks!

lgritz:
Great, thanks. No rush at all!

src/oiiotool/oiiotool.h
DharshanV: Done. Reworked to `create_dir`.

src/oiiotool/oiiotool.cpp
DharshanV: Done. Reversed the `if` statements, added description error messages.

DharshanV:
> ```
> oiiotool --create 2x2 1 -o 1/2/out.tif
> # ^^ should make the right error message as output
> 
> oiiotool --create-dir --create 2x2 1 -o 1/2/out.tif
> oiiotool --info 1/2/out.tif
> # ^^ should make the right output showing that the file is found
> ```

I created this test. But, on the first run of the tests (`make test TEST=^oiiotool$`) passes, but consecutive run's fails due to difference in the `out.txt` because the folder's `folder1/folder2` already exists, and it doesn't error on the first:
```
oiiotool ("--create 2x2 1 -o folder1/folder2/out.tif")
```

In order to make consecutive run's pass, I need to delete the `folder1` in my `build`, before running `make test ...`.
What would you suggests I do here?
1. Delete `folder1` using `shutil.rmtree('folder1')` in `testsuite/oiiotool/run.py`?
2. Or in `testsuite/runtest.py` delete folder's if `cleanup_on_success` is true?

lgritz:
Aha, yes. I think choice 1 is better -- just leave the funny deletion business local to this one test.

DharshanV:
Just started looking into the first choice, but it looks like I need to add a new command in `runtest.py`:
```
def rm_tree_command (args):
    shutil.rmtree(args)
```

and in my `oiiotool/run.py`, I than need to do:
```
...
command += oiiotool ("--info folder1/folder2/out.tif")
command += rm_tree_command ("folder1")
```
____

But I need to figure out the absolute path for my `shutil.rmtree`. Tried understanding the code what variable has this path but couldn't parse it. Can you let me know?

Thanks!

lgritz:
I don't think you need to do that. Instead of adding a command to the commands list that removes the directory after the test is run, I think you just need to remove the directory in run.py, which will execute BEFORE the command list is run, guaranteeing that the directory does not exist when the features is tested. No?

DharshanV:
@lgritz ahhh okay, I was trying to remove it after adding the command. Added it before, it works, thanks!

lgritz (approved):
LGTM!

Great work, helpful feature, thanks.

PR #4763, opened 2025-05-16
feat(oiiotool): eraseattrib_fromfile

lydia-zheng:
## Description

Extended eraseattrib to a prototyped standalone flag "--eraseattrib_fromfile" that allows for erasure of metadata from a text file with a list of regex.
Erases all metadata attribute fields that match.

## Tests

Added a new test to `oiiotool-attribs` which has been verified painstakenly to be working!
Also tried running the command directly from dist/bin and verified with `oiiotool --info -v`.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
Hi, Lydia, can you please fix the formatting? If you look at the CI clang-format run (look for the red and green highlights) you can see it's just a few lines that you could easily fix by hand if you don't have clang-format set up to run locally.
https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/15058621627/job/42329426760?pr=4763
(I hope that links shows you what it shows me)

lydia-zheng:
> Hi, Lydia, can you please fix the formatting? If you look at the CI clang-format run (look for the red and green highlights) you can see it's just a few lines that you could easily fix by hand if you don't have clang-format set up to run locally. https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/15058621627/job/42329426760?pr=4763 (I hope that links shows you what it shows me)

Yes Larry! Already working on a clang-format fix right now!

What I'm more worried about is the bleeding edge gcc14 test that failed? Digging into the log seems to imply that there are pixel differences from a test field that's completely untouched...

testsuite/oiiotool-attribs/run.py
```diff
@@ -7,6 +7,12 @@
 # Test oiiotool's ability to add and delete attributes
 
 
+# Test --eraseattrib_fromfile to erase from a list of regex matching attributes
+
+command += oiiotool (OIIO_TESTSUITE_IMAGEDIR+"/tahoe-gps.jpg --eraseattrib_fromfile src/regex_list.txt -o nomakegps.jpg")
+command += info_command ("nomakegps.jpg", safematch=True)
+
+
```
lgritz, line 15: Great test!

lgritz:
The "bleeding edge" failure is unrelated. I just submitted a fix here: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4765

lydia-zheng:
Taking a look at incorporating all the comments after redo-ing the format fix!

lgritz:
Because Lydia and I had talked about this offline, I want to fill in a bit of the motivation here for others who are reading:

It's common in a studio to have tools insert all sorts of metadata into files as they're rendered, etc. But a lot of that are things that SHOULD NOT leave the facility -- like security issues from internal pathnames or network info, or secrets like plot details revealed by sequence or character names. So generally you want to scrub the metadata from the final images before they are shipped to the client or to marketing or whatever.

oiiotool's `--eraseattrib` is handy for this, but it takes a single regex. But we were thinking that:

1. You may have several patterns to fully describe the metadata you want to be sure not to ship. This could be an unwieldy regex or a cluttered series of multiple --eraseattrib commands.
2. That list of patterns may change over time, and it would be error prone to have to change it in every script that calls oiiotool.
3. That list of patterns may need to be different per show.

We thought a solution to facilitate this might be a version of --eraseattrib that takes a filename that contains a list of regexes, one per line of the text file. So that master list of things to scrub could be maintained in a central place and not need to update every oiiotool-calling script in order to update it. Furthermore, you could have the command say something like

    $ oiiotool ready_to_ship.exr --eraseattrib_fromfile /blah/$SHOW/nometadata.txt -o clean.exr

to make a single command that allows per-show customization.

lgritz:
Question for onlookers: Lydia implemented this as a new command, `--eraseattrib_fromfile`.

An alternate design we discussed was just using the existing command but with a new optional modifier: `--eraseattrib:fromfile=1` that causes the subsequent argument to be interpreted as the name of a file containing the regexes, instead of as a regex itself.

Does anybody have a strong preference? Or any other feedback on the basic design?

build
```diff
@@ -0,0 +1 @@
+/net/soft_scratch/users/lzheng/openimageio/metadata/build
```
lgritz, line 1: This looks suspiciously like a mistakenly added file. Maybe it's responsible for all the tests suddenly failing?

src/oiiotool/oiiotool.cpp
```diff
@@ -1494,6 +1494,22 @@ set_oiio_attribute(Oiiotool& ot, cspan<const char*> argv)
...
+    // helper function that takes a text file and return all regex attribname entries as a list
+    std::vector<std::string> regex_list;
+    std::ifstream inputFile(filename);
+    if (inputFile.is_open()) {
+        std::string line;
+        while (std::getline(inputFile, line)) {
+            regex_list.push_back(line);
+            //std::cout << "list element:" << regex_list.back() << std::endl;
+        }
+    }
```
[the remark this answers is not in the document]
lydia-zheng: That's amazing! Cheers, will update as such!

src/oiiotool/oiiotool.cpp
```diff
@@ -1503,7 +1519,16 @@ class OpAttribSetter final : public OiiotoolOp {
         : OiiotoolOp(ot, opname, argv, 1)
     {
         inplace(true);  // This action operates in-place
-        attribname = args(1);
+        erase_from_file = "--eraseattrib_fromfile";
```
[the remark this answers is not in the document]
lydia-zheng: As in we want to allow for both `-` and `--`? Thanks for the clarification, was thinking of adding an or case with `||`, but if there's a more efficient solution, would be happy to change it up as such!

build
lydia-zheng: Yes, checking right now but definitely on me for being hasty!

src/oiiotool/oiiotool.cpp
```diff
@@ -1577,6 +1613,14 @@ erase_attribute(Oiiotool& ot, cspan<const char*> argv)
     action_attrib_helper(ot, argv[0], argv);
 }
 
+// --eraseattrib_fromfile
+static void
+erase_attribute_fromfile(Oiiotool& ot, cspan<const char*> argv)
```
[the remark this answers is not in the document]
lydia-zheng: That's very true, was thinking it was redundant but had added it for the comment clarity... but would be more compact/cleaner with your suggestion for sure

build
lydia-zheng: Hi Larry,

That should be removed now and just pushed with the latest updates, hopefully these will show all green soon!

lgritz (commented):
It's really close. I pointed out one line that is no longer needed.

I'd like to hold this in review for a few days, simply for the sake of allowing others the chance to comment on design of the feature.

Also, before merging, we really should make sure that the documentation mentions this feature.

lydia-zheng:
> It's really close. I pointed out one line that is no longer needed.
> 
> I'd like to hold this in review for a few days, simply for the sake of allowing others the chance to comment on design of the feature.
> 
> Also, before merging, we really should make sure that the documentation mentions this feature.

Sounds good Larry, I was checking the tests and it might have updated the branch to merge, feel free to ignore.

Adding a list of todos since there’s a chance that I might not be able to work on this soon:

- [ ] remove unneeded line from get_regex_list_from_file
- [ ] add documentation for eraseattrib_fromfile

lgritz:
Thanks for the feedback, Joseph. 

I don't believe that oiiotool has any other commands at the moment that read arguments from files. But maybe it will in the future, so I agree that we want to think carefully about what we establish as the precedent.

Are there other utilities (outside of oiio) that have similar functionality? Ah, the one that popped into my mind for some reason is "rsync", and its man page includes:
```
            --exclude=PATTERN       exclude files matching PATTERN
            --exclude-from=FILE     read exclude patterns from FILE
            --include=PATTERN       don't exclude files matching PATTERN
            --include-from=FILE     read include patterns from FILE
            --files-from=FILE       read list of source-file names from FILE
```

And I checked `tar` (the modern gnu version), which has:
```
     -T filename, --files-from filename
             In x or t mode, tar will read the list of names to be extracted
             from filename.  In c mode, tar will read names to be archived
             from filename.
```

So they are using separate commands, and seem to follow a common "--foo-from filename"  pattern.

But they don't the kind of "optional command modifiers" that we do, so maybe they didn't consider that part of the design space?

```
    # Existing command:
    oiiotool ... --eraseattrib pattern ...

    # Design 1: new command
    oiiotool ... --eraseattrib-from filename-containing-patterns ...

    # Design 2: optional modifier
    oiiotool ... --eraseattrib:fromfile=1 filename-containing-patterns ...

```

I'm mildly preferring the second option, simply because if someday we have many commands we want to operate like this, we don't have to make a completely new command for each instance. We can just make more commands use the same "fromfile" optional modifier. Use of optional modifiers in this way is already a very common oiiotool idioms that users expect.

That's my current thought, anyway. If everybody strongly prefers Design 1, I won't object.

nrusch:
> I'm mildly preferring the second option, simply because if someday we have many commands we want to operate like this, we don't have to make a completely new command for each instance. We can just make more commands use the same "fromfile" optional modifier. Use of optional modifiers in this way is already a very common oiiotool idioms that users expect.

I personally prefer the second option as well, for exactly the reasons you've alluded to:
1. Scales better with the (already large) number of `oiiotool` options.
2. Consistent with prior art for command modifiers.

src/doc/oiiotool.rst
```diff
@@ -119,10 +119,10 @@ contents of an expression may be any of:
 * *imagename.metadata*
 
   The named metadata of an image.
-  
+
   If there is no metadata whose name matches, the expression will not have any
   substitution made and an error will be issued.
-  
+
   The *imagename* may be one of:
```
lgritz, line 126: Something odd happened here, your editor seems to have introduced whitespace changes all through this document.

lgritz:
Your docs are great, thanks.

To do:
- [ ] One very minor suggestions about how to do the string comparison on the command name (I made the suggestion via diffs, if you're ok with that, you can just click to accept the change)
- [ ] I would like you to back out the unintended whitespace changes in the docs. There's no reason to touch so many lines in this file.
- [ ] Let's sit on this until early next week to give people a little more time to think about the design choice and weigh in if they have a preference.

src/oiiotool/oiiotool.cpp
```diff
@@ -1503,8 +1511,17 @@ class OpAttribSetter final : public OiiotoolOp {
         : OiiotoolOp(ot, opname, argv, 1)
     {
         inplace(true);  // This action operates in-place
-        attribname = args(1);
-        value      = (nargs() > 2 ? args(2) : "");
+        erase_from_file     = "--eraseattrib_fromfile";
+        erase_from_file_alt = "-eraseattrib_fromfile";
+        if (opname != erase_from_file && opname != erase_from_file_alt) {
```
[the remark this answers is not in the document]
lydia-zheng: Much cleaner and is super obvious once you pointed it out, thanks a bunch!

src/doc/oiiotool.rst
lydia-zheng: This is embarrassing, apologies Larry! 

Was looking at this and I think that's still affected, will take a look again later.

lydia-zheng: Hi Larry, 

I believe the whitespace issue is walked back, let me know if those still look wrong?
Seems like the editor was trimming them, but I've verified with an older commit hash of the file that this should be what we started with.

Also the string comparison suggestion should be there as well!

JGoldstone:
Amazing how far behind one can get on an issue by taking just a single day off. But: I wanted to say I really like design #2, as well.

lgritz:
Lydia, this looks great now, the stray whitespace changes are definitely cleaned up.

It feel like consensus is leaning toward Design 2. We'll discuss in today's TSC meeting, but I suspect that is the way we'd rather go.

src/doc/oiiotool.rst
```diff
@@ -2043,6 +2050,13 @@ current top image.
         # Remove all metadata
         oiiotool in.exr --eraseattrib:subimages=all ".*" -o no_metadata.exr
 
+        #Remove all attribute that match any regex in text file
```
lgritz, line 2053: Just a spacing change:
```suggestion
        # Remove all attribute that match any regex in text file
```

src/doc/oiiotool.rst
```diff
@@ -2043,6 +2050,13 @@ current top image.
...
+        oiiotool in.exr --eraseattrib:fromfile=1 no_gps_make.txt -o no_gps_make_metadata.exr
+
+        no_gps_make.txt content:
```
lgritz, line 2056:
```suggestion
        Example contents of file no_gps_make.txt:
```

src/oiiotool/oiiotool.cpp
```diff
@@ -1503,8 +1511,17 @@ class OpAttribSetter final : public OiiotoolOp {
...
         inplace(true);  // This action operates in-place
-        attribname = args(1);
-        value      = (nargs() > 2 ? args(2) : "");
+        auto options  = ot.extract_options(opname);
+        bool fromfile = options.get_int("fromfile", 0);
+        if (!fromfile) {
+            attribname = args(1);
+        } else {
+            // handle erase attribute using regex text file case
+            // defined it outside to avoid magic number, still not very elegant solution wise
```
lgritz, line 1520: I don't understand this comment. Magic number?

lgritz (commented):
Looks great!

I made a couple very minor suggestions about documentation wording and formatting. The diffs should allow you to click to accept them without needing to re-edit and push (assuming you agree with my suggestions).

src/oiiotool/oiiotool.cpp
lydia-zheng: Feel free to disregard! Second sentence was leftover commentary from the old old prototype design, doesn't apply here at all after design 2. Will be removing that

lydia-zheng:
> Looks great!
> 
> I made a couple very minor suggestions about documentation wording and formatting. The diffs should allow you to click to accept them without needing to re-edit and push (assuming you agree with my suggestions).

Hi Larry! Those all looked good, I've accepted the documentation wording updates, cheers.

lgritz (approved):
Congrats, Lydia, LGTM!

lydia-zheng:
Sweet! Thanks for the double checking, and for everyone's input on the design of the feature!

PR #4764, opened 2025-05-16
fix(iv) fix crash on .DS_Store; fix uppercase extensions

antond-weta:
## Description

I've found a couple issues with iv when launched with a directory name as input, while iterating through the files in the directory:
1. Crash on .DS_Store file
2. The files with extension in capital letters are not recognised as valid images, due to case-sensitive compare. Capitalised extensions are common in raw camera images.

## Tests

No tests, the changes are trivial, and in UI-specific code, not easy to test.

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (approved):
LGTM

PR #4766, opened 2025-05-16
Fix #4718 do not resize on open and other zoom fixes

### LINKED ISSUE #4718: iv: "Open" shouldn't resize the window
originally was #289

soswow:
## Description

Fixes #4718

See the demo of problems that were fixed and of fixes in action over here: https://www.loom.com/share/3bdee20b32434e0bbba47d393928d4f7

There are fixed for three problems:
1) Window is not resizing when new image is loaded in (it still does when image being loaded is the only one)
2) When you using scroll mouse wheel it is taking into account where the cursor is (there is some corner case when it doesn't do it perfectly, that I can't catch when it doesn't follow it exactly, but a bit of panning fixes it)
~~3) Switching between images recenters (if image is smaller) or making some other panning nessery to have image in the viewport. I believe it is doing same stuff what happens when window is resized.~~ This has been reverted.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ gastest.o
 /*.exr
 /*.tif
 /*.jpg
+/*.png
 /*.jxl
 /*.tx
 /*.log
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -1252,13 +1252,14 @@ ImageViewer::displayCurrentImage(bool update)
 
     if (update) {
         glwin->update();
+        glwin->clamp_view_to_window();
     }
     float z = zoom();
     if (fitImageToWindowAct->isChecked())
         z = zoom_needed_to_fit(glwin->width(), glwin->height());
     zoom(z);
     //    glwin->trigger_redraw ();
-
+    
     updateTitle();
     updateStatusBar();
     if (infoWindow)
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -2015,7 +2015,7 @@ ImageViewer::zoomIn(bool smooth)
     float current_zoom = zoom();
     if (current_zoom >= 64)
         return;
-    
+
     float newzoom = ceil2f(current_zoom);
 
     this->zoomToCursor(newzoom, smooth);
@@ -2028,11 +2028,11 @@ ImageViewer::zoomOut(bool smooth)
     IvImage* img = cur();
     if (!img)
         return;
-    
+
     float current_zoom = zoom();
     if (current_zoom <= 1.0f / 64)
         return;
-    
+
     float newzoom = floor2f(current_zoom);
 
     this->zoomToCursor(newzoom, smooth);
@@ -2051,10 +2051,11 @@ ImageViewer::zoomToCursor(float newzoom, bool smooth)
     float yoffset = yc - ym;
 
     float maxzoomratio = std::max(oldzoom / newzoom, newzoom / oldzoom);
-    int nsteps = smooth ? (int)OIIO::clamp(20 * (maxzoomratio - 1), 2.0f, 10.0f) : 1;
+    int nsteps = smooth ? (int)OIIO::clamp(20 * (maxzoomratio - 1), 2.0f, 10.0f)
+                        : 1;
     for (int i = 1; i <= nsteps; ++i) {
-        float a = (float)i / (float)nsteps;  // Interpolation amount
-        float z = OIIO::lerp(oldzoom, newzoom, a);
+        float a         = (float)i / (float)nsteps;  // Interpolation amount
+        float z         = OIIO::lerp(oldzoom, newzoom, a);
         float zoomratio = z / oldzoom;
         view(xm + xoffset / zoomratio, ym + yoffset / zoomratio, z, false);
         if (i != nsteps) {
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -1252,14 +1252,13 @@ ImageViewer::displayCurrentImage(bool update)
 
     if (update) {
         glwin->update();
-        glwin->clamp_view_to_window();
     }
     float z = zoom();
     if (fitImageToWindowAct->isChecked())
         z = zoom_needed_to_fit(glwin->width(), glwin->height());
     zoom(z);
     //    glwin->trigger_redraw ();
-    
+
     updateTitle();
     updateStatusBar();
     if (infoWindow)
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -2005,67 +2005,58 @@ ImageViewer::print()
 }
 
 
-
 void
-ImageViewer::zoomIn()
+ImageViewer::zoomIn(bool smooth)
 {
     IvImage* img = cur();
     if (!img)
         return;
-    if (zoom() >= 64)
-        return;
-    float oldzoom = zoom();
-    float newzoom = ceil2f(oldzoom);
 
+    float current_zoom = zoom();
+    if (current_zoom >= 64)
+        return;
+    
+    float newzoom = ceil2f(current_zoom);
+
+    this->zoomToCursor(newzoom, smooth);
+}
+
+
+void
+ImageViewer::zoomOut(bool smooth)
+{
+    IvImage* img = cur();
+    if (!img)
+        return;
+    
+    float current_zoom = zoom();
+    if (current_zoom <= 1.0f / 64)
+        return;
+    
+    float newzoom = floor2f(current_zoom);
+
+    this->zoomToCursor(newzoom, smooth);
+}
+
+
+void
+ImageViewer::zoomToCursor(float newzoom, bool smooth)
+{
+    float oldzoom = zoom();
     float xc, yc;  // Center view position
     glwin->get_center(xc, yc);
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -23,6 +23,7 @@ gastest.o
 /*.exr
 /*.tif
 /*.jpg
+/*.png
```
soswow, line 26: since I had my own test images and they were .png I thought it would be ok

src/iv/imageviewer.cpp
```diff
@@ -854,14 +854,11 @@ ImageViewer::open()
         if (filename.empty())
             continue;
         add_image(filename);
-        //        int n = m_images.size()-1;
-        //        IvImage *newimage = m_images[n];
-        //        newimage->read_iv (0, false, image_progress_callback, this);
```
soswow, line 859: not sure if this and all the other commented parts of the code are there for a reason, so feel free to ask to revert if you believe it is needed

src/iv/imageviewer.cpp
```diff
@@ -887,7 +884,6 @@ ImageViewer::openRecentFile()
         if (m_images.size() > 1) {
             // Otherwise, add_image already did this for us.
             current_image(m_images.size() - 1);
-            fitWindowToImage(true, true);
```
soswow, line 890: that stops window from resizing when new images are loaded in

src/iv/imageviewer.cpp
```diff
@@ -1256,13 +1252,14 @@ ImageViewer::displayCurrentImage(bool update)
 
     if (update) {
         glwin->update();
+        glwin->clamp_view_to_window();
```
soswow, line 1255: that what make sure loaded image (when switching between images for example) is in the view. if images is smaller then viewport this will center it as well

src/iv/imageviewer.cpp
```diff
@@ -2009,67 +2006,58 @@ ImageViewer::print()
...
         return;
-    if (zoom() <= 1.0f / 64)
+    
+    float current_zoom = zoom();
+    if (current_zoom <= 1.0f / 64)
         return;
+    
+    float newzoom = floor2f(current_zoom);
+
+    this->zoomToCursor(newzoom, smooth);
```
soswow, line 2039: Refactor zoomIn and zoomOut to use common code (new zoomToCursor function) that to be able to zoom at cursor with or without animation.

src/iv/ivgl.cpp
```diff
@@ -1328,9 +1328,11 @@ IvGL::wheelEvent(QWheelEvent* event)
...
     if (abs(angdelta.y()) > abs(angdelta.x())      // predominantly vertical
         && abs(angdelta.y()) > 2) {                // suppress tiny motions
-        float oldzoom = m_viewer.zoom();
-        float newzoom = (angdelta.y() > 0) ? ceil2f(oldzoom) : floor2f(oldzoom);
-        m_viewer.zoom(newzoom);
+        if (angdelta.y() > 0) {
+            m_viewer.zoomIn(false);
+        } else {
+            m_viewer.zoomOut(false);
+        }
```
soswow, line 1335: that fixes the zooming with mouse wheel to use mouse cursor position

src/iv/ivgl.h
```diff
@@ -96,6 +96,8 @@ class IvGL : public QOpenGLWidget, protected QOpenGLExtraFunctions {
                             GLenum& gltype, GLenum& glformat,
                             GLenum& glinternal) const;
 
+    void clamp_view_to_window();
```
soswow, line 99: This is the only change I Am not confident about. In order to use `clamp_view_to_window` I had to move it from protected to public. I don't know if this is a big deal. I seek some guidance here.

lgritz:
Overall, the code looks great (needs a clang-format pass) and you've improved a number of things.

Honestly, the only thing I'm unsure about how you've done is the centering of each image in the view. I think if one is viewing a number of unrelated images of different sizes, that's probably fine. But if you are looking at related images that are occupying the same abstract pixel coordinate system, if you know what I mean, perhaps it's better for them to stay aligned so that as you switch between images, pixel (2,42) or each image is always at the same spot on screen as you cycle through. 

I think that the old behavior kept the same coordinate system as you cycled through images instead of repositioning it as you go from one image to another. It's not clear to me whether or not people will view that as a welcome change.

Actually, as I'm thinking about it, there are at least a few different (and contradictory) behaviors people may desire for different use cases, when there are images of different sizes:
1. A single coordinate system that stays in the same position relative to the view window (I mean you can scroll around, but merely Up/Dn to cycle through the image list doesn't move the coord system).
2. Images always align themselves *individually* to be best viewed (centered, as you come to each image? at least until until you use the mouse to drag it to a different position, then maybe that sticks?).
3. Images align but also zoom to the window size, so that, for example, if you are viewing one 1k image and one 2k image of the same scene, they both appear the same size in the window, so you can flip back and forth between them and they stay aligned. (At least until you zoom in or out, then the zoom level you selected sticks?)

I'm kind of rambling. I don't want to keep expanding the scope of this one PR. Maybe the best policy is to stick to option 1 (as we had before, I think) but then consider a subsequent pass (or at least filing a reminder issue) about allowing users to choose any of those strategies. That is, unless other viewers in general agree on a well established convention for doing this that I'm unaware of, in which case we should align to that for the sake of user expectations.

Thoughts?

soswow:
@lgritz 
> , if you know what I mean, perhaps it's better for them to stay aligned so that as you switch between images, pixel (2,42) on each image is always at the same spot on screen as you cycle through.

yeah ... When I saw various feature requests that involve some sort of relationship between images in the set (showing a diff or having a split) I already thought about the fact that `iv` might be often used with images of the same size.

hm ... 

I just tried and it works actually well. The only difference from master is when you see image edge in the image (and some black void behind it). Before it would just switch between keeping same position. Now it would make sure whole viewport is filled out. See norated video for illustration: https://imgur.com/a/hY0EVuv

Let me see if I can improve by combing best of both worlds and if not - I Can drop that part for now.

lgritz:
> yeah ... When I saw various feature requests that involve some sort of relationship between images in the set (showing a diff or having a split) I already thought about the fact that `iv` might be often used with images of the same size.

And keep in mind an important special case of this: images that are different *sizes* (in terms of the actual pixel array, what we'd call "data window" for openexr files) but are using a common coordinate system overall and should be aligned to it.

An example of that is that it's common for openexr files to be "shrink wrapped" where it only stores the pixel/data window of the smallest rectangle that completely encloses all the non-black pixels. So if you have two frames from a shot of a space ship that's moving, and those frames are shrink wrapped, maybe on frame 1 the image's data window is (100,100) - (200,200) [thus a "resolution" of 100x100] and on the second frame it is farther right and down an has changed orientation so now the nonzero pixels that are stored are (250,250)-(400,350) [thus a data "resolution" of 150x100, and with a different upper-left origin]. 

As you flip between those two images, they shouldn't display "on top of" each other, nor should they be aligned to the upper left corner of the screen. One has its upper left corner of pixels at coordinate (100,100) and the other at coordinate (250,250) on a common coordinate system that should not shift around within the viewing window just because you are switching from one image to the other.

I hope that makes some kind of sense.

lgritz:
I'm just thinking aloud here, don't consider this a firm position. :-) But I'm wondering what people think about this strategy I'm about to outline:

Initially, you inventory the images specified on the command line, and take basically the union of their respective "full" resolutions (that's the OIIO term, it's what OpenEXR calls "display window", and is taken from the ImageSpec's full_x, full_y, full_width, full_height). Not the regular pixel data resolution (ImageSpec's x, y, width, height, which correspond to what OpenEXR calls the "data window").

If that combined full size fits on the screen, you start with zoom = 1.0 and make the screen window the size of that combined full size, but if it's too big to fit on the display, you make the window as big as can fit on the display, and set the zoom to whatever makes the combined full size just barely fully fit, centered, in that window (zoom will thus will be something < 1.0). Either way, this establishes a mapping between the pixel coordinate system and the on-screen display showing the images.

At that point, the user dragging with the mouse, or zooming (mouse wheel or clicks or hotkeys) moves or zooms the coordinate system relative to the on-screen viewer window. But up/down to switch between images does not -- they all just display relative to what their ImageSpec says their placement should be on that pixel coordinate system. Opening/loading a new image that we didn't start with also should not change the coordinate system at all at that point. But there is a hotkey ('alt-f', I think) that recenters and/or resizes the mapping between pixel coordinate system and viewer window to optionally center and fill the current viewer window with the currently displayed image. And also hotkey 'f' changes the viewer window to exactly enclose the currently displayed image in its current position and zoom as it appears on screen. 

Reading this over, I suspect that this is what I was originally going for all those years ago, though I may not have correctly implemented all those behaviors.

src/libtexture/imagecache_pvt.h
soswow, line 1: 🤷 blame clang-format

src/iv/imageviewer.cpp
soswow: reverted

lgritz:
Yeah, the behavior looks reasonable to me now. It's certainly a step forward from what we had before.

soswow:
@lgritz I am glad to hear that. Tell me what can I do else? Are there any unit tests/integration tests that needs to be added? I've assumed since nothing is failing there is no coverage there.

lgritz (approved):
All LGTM now. Thanks, this is a solid improvement!

Issue #4714, opened 2025-04-25
iv: area probe


lgritz:
Allow a way to sweep out a rectangle and report the min/max/avg values within the rectangle. 

(originally #310)

dann1m:
Hello! This is my first year as a contributor to ASWF Dev Days. I would like to work on this issue as a first task. Is it too late to have this task assigned to me? Thank you!

lgritz:
We don't assign tasks so much as people claim them for themselves. Which you just did!

dann1m:
I have a quick question. I have the area probe set up, but I'm facing an issue where the mouse input is only routing to zoom rather than allowing for selection. I was wondering if you have any suggestions on this

lgritz:
I think maybe you want a menu selection and/or hotkey that toggles on this "area sample" mode. By default, click zooms. But when in "area sample" mode, a click will let you drag out an area.

Does that help?

PR #4767, opened 2025-05-16
feat:area probe

dann1m:
## Description

This PR is for a feature request: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4714.

This PR adds support for selecting a rectangular region within an image to compute and display per-channel minimum, maximum, and average values.

The feature can be activated through the Tools menu or by using a keyboard shortcut (E) . Once active, the user can click and drag with the left mouse button to define a rectangular area, which is visually indicated by a blue outline. Upon releasing the mouse button, the tool calculates statistics for all pixels within the selected area and displays the results in the status bar.

## Checklist:

- [x ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
- [x ] I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
- [ x] I have ensured that the change is tested somewhere in the testsuite
  nonconforming.

### DIFF
--- a/src/iv/ivgl.cpp
+++ b/src/iv/ivgl.cpp
@@ -698,7 +698,8 @@ IvGL::paintGL()
         
         gl_rect( left, bottom, right, top, -0.1f);     
         
-        glPopAttrib();     
+        glPopAttrib();  
+        glPopMatrix();     
     
     }
     glPopMatrix();  
@@ -1321,25 +1322,34 @@ IvGL::mousePressEvent(QMouseEvent* event)
 {
     remember_mouse(event->pos());
     int mousemode = m_viewer.mouseModeComboBox->currentIndex();
-    // mousemode = 3;
+    bool areaMode = m_viewer.areaSampleMode();
     bool Alt      = (event->modifiers() & Qt::AltModifier);
     m_drag_button = event->button();
     if (!m_mouse_activation) {
         switch (event->button()) {
         case Qt::LeftButton:
-            if (mousemode == ImageViewer::MouseModeZoom && !Alt)
-                m_viewer.zoomIn();
-            else if (mousemode == ImageViewer::MouseModeSelect && !Alt){
+            if (areaMode){
+                std::cerr << areaMode;
                 m_select_start = event->pos();
                 m_select_end = m_select_start;
                 m_selecting = true;
                 parent_t::update();
+            } else if (mousemode == ImageViewer::MouseModeZoom && !Alt && !areaMode){
+                m_viewer.zoomIn();
             }
+            // if (mousemode == ImageViewer::MouseModeZoom && !Alt && !areaMode)
+            //     m_viewer.zoomIn();
+            // else if (mousemode == ImageViewer::MouseModeSelect && !Alt && areaMode){
+            //     m_select_start = event->pos();
+            //     m_select_end = m_select_start;
+            //     m_selecting = true;
+            //     parent_t::update();
+            // }
             else
                 m_dragging = true;
             return;
         case Qt::RightButton:
-            if (mousemode == ImageViewer::MouseModeZoom && !Alt)
+            if (mousemode == ImageViewer::MouseModeZoom && !Alt && !areaMode)
                 m_viewer.zoomOut();
             else
                 m_dragging = true;
@@ -1371,7 +1381,6 @@ IvGL::mouseReleaseEvent(QMouseEvent* event)
         m_select_end = event->pos();
         m_selecting = false;
         analyze_selected_area();
-        parent_t::update();
 
     }
     parent_t::mouseReleaseEvent(event);
@@ -1383,6 +1392,19 @@ void
 IvGL::mouseMoveEvent(QMouseEvent* event)
 {
     QPoint pos = event->pos();
+
+    // Area probe override
+    if (m_viewer.areaSampleMode() && m_selecting) {
+        m_select_end = event->pos();
+        parent_t::update();
+        remember_mouse(pos);
+        if (m_viewer.pixelviewOn()){
+            parent_t::update();
+        }
+        parent_t::mouseMoveEvent(event);
+        return;
+    }
+
     // FIXME - there's probably a better Qt way than tracking the button
     // myself.
     bool Alt      = (event->modifiers() & Qt::AltModifier);
--- a/src/iv/ivgl.cpp
+++ b/src/iv/ivgl.cpp
@@ -1321,7 +1321,7 @@ IvGL::mousePressEvent(QMouseEvent* event)
 {
     remember_mouse(event->pos());
     int mousemode = m_viewer.mouseModeComboBox->currentIndex();
-    mousemode = 3;
+    // mousemode = 3;
     bool Alt      = (event->modifiers() & Qt::AltModifier);
     m_drag_button = event->button();
     if (!m_mouse_activation) {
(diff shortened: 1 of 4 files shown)


lgritz:
Hi, Danielle, it's great to see you here, and this is a good start!

I tried this out on my end, and I have a number of comments. I'm just going to list them all, in no particular order. Some of this is going to be pretty picky about minor UI issues. I think they are all fairly minor in the grand scheme of things, so don't be discouraged that it's not 100% on the first try. (Nothing I do with UIs ever is.)

* Is there a reason for 'E' as the hotkey? Does it stand for something?
* When I hit 'E', there is no immediate visual feedback that I'm in "sweep an area to sample" mode. I think that perhaps if the cursor changes at that point to a different shape (especially if it somehow indicates or invites the user is expected to sweep out an area), it would be helpful. Or, maybe it's enough to change the status bar (or display the overlay) when 'E' is hit, instead of only after the area is swept? Perhaps that would make it more visually obvious that you're in a new mode.
* After the sweep, the display changes in a fairly jarring way:
  - The status bar at the bottom goes from being strictly one line, to being vertically thick with each channel taking an additional line. This isn't necessarily bad by itself, though we should consider whether it "scales", for example, is it going to look really bad if you're viewing an image with 27 channels? This may be turn out to be the best argument for showing the area info in an overlay, like the 'P' pixel mode display, instead of on the status bar.
  - The viewed image itself jumps to a different position. I think this is a real problem. We definitely want the displayed image to stay in the previous spot, even if the size of the status line changes.
* The channels are listed as "ch0", "ch1". We should use the real names, like the 'P' pixel view overlay does.
* Not a problem: I like the fact that you can actually enable the pixel view mode and the area sweep mode simultaneously!
* When you hit 'E' again to go back to the usual mode again, the image pops position again, and also changes color (!) until you move the mouse again, and then goes back to normal. Not sure what that's all about.

dann1m:
@lgritz, I made the suggested changes to the feature if you would like to take a look at it. 
There is an issue with one of the commits not recognizing the signature, so I attempted rebasing a couple of times, and also a different pr entirely, but to no avail.

lgritz:
I think the problem is that you tried to "merge". That's rarely necessary for a PR unless you're pretty sure there is a direct conflict between what you're working on and something committed since you branched off. I think that doing a "rebase" on top of the new main is much better, and yields a simple linear history.

Hang on, I think I can fix this for you... Do you mind if I push a fix to your branch?

lgritz:
By the way, it's good to see you back here, Danielle! I was really hoping you were going to pick up this work again and be able to get it fully merged.

dann1m:
I don't mind at all, thank you so much! Sorry, I took time with it; school was winding down. It's good to be back!

lgritz:
ok, I pushed the changes to your fork -- cleanly rebased on top of main, fixed the one broken signature that was holding up the DCO. 

You'll need to fetch or pull on your end to get my changes to your local machine, and move your branch marker to the new one. (Or do a fresh checkout.)

dann1m:
Thank you @lgritz! I did a bit of cleaning, so everything should work now. When you have time, I would appreciate feedback

lgritz:
I pulled this to my machine, did a build, and played around with it a bit. Behavior-wise, this looks really good to me! I like how it works. It's also handy that when you're in area mode, if you just click, it shows you the value of that pixel. I don't have any additional complaints or suggestions about the behavior at this time. I'll take one more look at the code itself and see if I have any suggestions there.

src/iv/ivgl.cpp
```diff
@@ -664,12 +665,54 @@ IvGL::paintGL()
         paint_windowguides();
     }
 
+    if (m_selecting) {
+        glPushMatrix();
+        glLoadIdentity();
+        // Transform is now same as the main GL viewport -- window pixels as
+        // units, with (0,0) at the center of the visible window.
+
+
```
lgritz: Such a minor nit, but we usually try to avoid multiple blank lines in a row *within* a function body.

src/iv/ivgl.cpp
```diff
@@ -938,6 +981,89 @@ IvGL::paint_pixelview()
...
+    float closeup_height = closeupsize * (0.06f * (spec.nchannels + 1));
+
+    // Position the close-up box
+    const float status_bar_offset = 35.0f;
+    glTranslatef(closeup_width * 0.5f + 5 - width() / 2,
+                 closeup_height * 0.5f + status_bar_offset - height() / 2, 0);
+
+    // Movement logic
+    // if (m_pixelview_left_corner) {
+    //     glTranslatef(closeup_width * 0.5f + 5 - width() / 2,
```
lgritz: If this commented-out code is no longer needed, can we remove it?

src/iv/ivgl.cpp
```diff
@@ -938,6 +981,89 @@ IvGL::paint_pixelview()
...
+    // } else {
+    //     glTranslatef(-closeup_width * 0.5f - 5 + width() / 2,
+    //                 closeup_height * 0.5f + status_bar_offset - height() / 2, 0);
+
+    //     if (xw > (width() - closeup_width - 5) && yw > (height() - closeup_height - status_bar_offset))
+    //         m_pixelview_left_corner = true;
+    // }
+
+
+
```
lgritz: Another spot where just one blank line is plenty.

```suggestion

```

src/iv/ivgl.cpp
```diff
@@ -938,6 +981,89 @@ IvGL::paint_pixelview()
...
+    glPushAttrib(GL_ENABLE_BIT | GL_CURRENT_BIT);
+    glDisable(GL_TEXTURE_2D);
+    if (m_use_shaders)
+        glUseProgram(0);
+
+
```
lgritz:
```suggestion

```

src/iv/ivgl.cpp
```diff
@@ -938,6 +981,89 @@ IvGL::paint_pixelview()
...
+    float extraspace = 10 * (1 + spec.nchannels) + 4;
+    glColor4f(0.1f, 0.1f, 0.1f, 0.5f);
+    gl_rect(-0.5f * closeup_width - 2, 0.5f * closeup_height + 10 + 2,
+            0.5f * closeup_width + 2, -0.5f * closeup_height - extraspace,
+            -0.1f);
+
+
```
lgritz:
```suggestion

```

src/iv/ivgl.cpp
```diff
@@ -938,6 +981,89 @@ IvGL::paint_pixelview()
...
+    // Draw probe text
+    QFont font;
+
+    // int textx = m_pixelview_left_corner ? 9 : width() - closeup_width - 1;
+    // int texty = 30;
```
lgritz: Are these commented-out lines something that should be removed?

src/iv/ivgl.cpp
```diff
@@ -1207,25 +1333,123 @@ IvGL::clamp_view_to_window()
...
                 m_viewer.zoomIn(true);  // Animated zoom for mouse clicks
+            }
+            // if (mousemode == ImageViewer::MouseModeZoom && !Alt && !areaMode)
+            //     m_viewer.zoomIn();
+            // else if (mousemode == ImageViewer::MouseModeSelect && !Alt && areaMode){
+            //     m_select_start = event->pos();
+            //     m_select_end = m_select_start;
+            //     m_selecting = true;
+            //     parent_t::update();
+            // }
```
lgritz: Keep or delete?

src/iv/ivgl.cpp
```diff
@@ -1357,7 +1608,30 @@ IvGL::get_focus_window_pixel(int& x, int& y)
     y = m_mousey;
 }
 
-
+void
```
lgritz, line 1586: I know this is incredibly picky, but we like 3 blank lines between standalone functions, to give a little visual clarity when quickly scanning code. (This is one example here, but please check the spacing between other functions you added.)

```suggestion


void
```

dann1m:
No problem! I'll deal with those now

src/iv/ivgl.cpp
dann1m: i'll delete it

src/iv/ivgl.cpp
dann1m: yes, taking it out now

lgritz (approved):
LGTM, Danielle. Thanks for making the corrections quickly. Looks great, congrats!

PR #4768, opened 2025-05-17
Use screen pixel ratio to render sharp text in pixel view tool

soswow:
## Description

Improve sharpness of text in pixel view tool for HiDPI screen

Before:
Screenshot 2025-05-17 at 4 45 50 pm

After:
Screenshot 2025-05-17 at 4 47 11 pm

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/iv/ivgl.cpp
+++ b/src/iv/ivgl.cpp
@@ -690,7 +690,9 @@ IvGL::shadowed_text(float x, float y, float /*z*/, const std::string& s,
      * Paint on intermediate QImage, AA text on QOpenGLWidget based
      * QPaintDevice requires MSAA
      */
-    QImage t(size(), QImage::Format_ARGB32_Premultiplied);
+    qreal dpr = devicePixelRatio();
+    QImage t(size() * dpr, QImage::Format_ARGB32_Premultiplied);
+    t.setDevicePixelRatio(dpr);
     t.fill(qRgba(0, 0, 0, 0));
     {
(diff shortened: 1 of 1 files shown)


src/iv/ivgl.cpp
```diff
@@ -690,7 +690,9 @@ IvGL::shadowed_text(float x, float y, float /*z*/, const std::string& s,
      * Paint on intermediate QImage, AA text on QOpenGLWidget based
      * QPaintDevice requires MSAA
      */
-    QImage t(size(), QImage::Format_ARGB32_Premultiplied);
+    qreal dpr = devicePixelRatio();
```
soswow, line 693: I don't know if there are some checks for verson of QT needs to happen here.

src/iv/ivgl.cpp
```diff
@@ -690,7 +690,9 @@ IvGL::shadowed_text(float x, float y, float /*z*/, const std::string& s,
...
+    QImage t(size() * dpr, QImage::Format_ARGB32_Premultiplied);
+    t.setDevicePixelRatio(dpr);
```
soswow, line 695: hm. I can see in a different place there is 
```
float dpr = m_viewer.devicePixelRatio();
glScalef(dpr, dpr, 1.0f);
```
I need to understand better what all of this is doing ...

lgritz (approved):
Nice results!

LGTM

PR #4769, opened 2025-05-17
docs: Update Windows build instructions to rely on deps auto-build

lgritz:
The INSTALL.md Windows building instructions were out of date in two important ways: (1) they neglected to mention OpenColorIO as a (newly) required dependency nor give instructions for how to build it; (2) it doesn't mention the new BUILD_MISSING_DEPS capabilities.

So basically, this PR modifies our Windows build guidance to set OpenImageIO_BUILD_MISSING_DEPS=all and rely on that without needing to document any additional steps for dependencies.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -280,64 +280,64 @@ Building on Windows
 
 You will need to have Git, CMake and Visual Studio installed.
 
-The minimal set of dependencies for OIIO is: zlib, libTIFF, Imath, OpenEXR, and libjpeg or libjpeg-turbo. If you have them built somewhere then you skip
-the section below, and will only have to point OIIO build process so their locations.
+The minimal set of dependencies for OIIO is: zlib, libTIFF, Imath, OpenEXR,
+OpenColorIO, and libjpeg or libjpeg-turbo. If you have them built somewhere
+then if you set the environment variable `CMAKE_PREFIX_PATH` to include those
+areas so CMake will find them, the OpenImageIO build will find those pre-built
+packages and use them. If you don't have them build, or are unsure, the
+`OpenImageIO_BUILD_MISSING_DEPS=all` CMake option will cause the build process
+to download the sources of those dependencies, build them, and use them.
 
-* zlib: this will build it, and then delete the non-static library, so they don't get picked up:
-  ```
-  cd {ZLIB_ROOT}
-  git clone https://github.com/madler/zlib .
-  cmake -S . -B build -DCMAKE_INSTALL_PREFIX=.
-  cmake --build build --config Release --target install
-  del build\Release\zlib.lib
-  ```
-* libTIFF:
-  ```
-  cd {TIFF_ROOT}
-  git clone https://gitlab.com/libtiff/libtiff.git .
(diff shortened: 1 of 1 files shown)


aras-p:
> Can you please look this over and make sure I've said the right things and that none the instructions I removed are still essential and should be left in?

Yes, much simpler and easier now with the auto-deps build. I tried the instructions as they are in this PR, everything works fine!

> Is the "auto dependency build" (which we use in CI for the Python wheels, among other places) something we should enable by default on Windows? Everywhere? 

I would go for having it on by default, at least on windows. Maybe everywhere. Obviously everywhere you need a way to turn it off, but having the simplest out of the box experience would be a big plus. That said, I don't know the usual Linux workflows.

> The Windows build instructions from Aras specify USE_PYTHON=0. Is that the right choice?

I frankly forget why I used both Python and Qt off in those instructions. Most likely because for my use case I did not need any of them. The same reason why I had "no shared libs, use static linking". Maybe the readme could say that "this builds statically linked OIIO libraries and utilities, without Qt or Python integrations for simplicity" or similar.

A complete side note: the same readme says that visual studio needs to be 2017-2019 version, but I think that should be extended to 2017-2022.

lgritz:
> I would go for having it on by default, at least on windows. Maybe everywhere. Obviously everywhere you need a way to turn it off, but having the simplest out of the box experience would be a big plus. That said, I don't know the usual Linux workflows.

My only worry was that while we rely on it for CI, it's new enough that I'm not sure a lot of users have tried it in the wild or what ways it can fail that we haven't anticipated. I guess at least on Windows, the number of people whose builds will break if it's turned on will probably not be worse than the number of people whose builds break now because it can't find a dependency that it needs.

> > The Windows build instructions from Aras specify USE_PYTHON=0. Is that the right choice?
> 
> I frankly forget why I used both Python and Qt off in those instructions. Most likely because for my use case I did not need any of them.

I think, then, I will change the instructions to re-enable them but point out how they can be disabled.

> A complete side note: the same readme says that visual studio needs to be 2017-2019 version, but I think that should be extended to 2017-2022.

Thanks. I will fix.

lgritz:
I revised the instructions a bit based on the above discussion.

I'm holding off on changing the default BUILD_MISSING_DEPS for Windows, that feels like it should be a totally separate PR to isolate it.

INSTALL.md
```diff
@@ -280,64 +280,71 @@ Building on Windows
...
 ```
 
-This will produce `{OIIO_ROOT}/build/OpenImageIO.sln` that can be opened in Visual Studio IDE. Note that the solution will be
-only for the Intel x64 architecture only; and will only target "min-spec" (SSE2) SIMD instruction set.
+If that command succeeds, you can either do the full build and install 
+from the command line, or open the generated Visual Studio solution.
+
+Note that you can speed up the build by disabling certain components if you
+know you won't need them: Adding `-DUSE_PYTHON=0` to the command above will
+skip building the Python bindings, `-DUSE_QT=0` will disable looking foor and
```
aras-p, line 327: `foor` -> `for` maybe

lgritz: yep

lgritz:
Merging these docs improvements, Aras said they worked for him.

PR #4773, opened 2025-05-18
fix(iv): Bug fix for iv window loosing focus on mac on startup

soswow:
## Description

Master version of the iv today when opening on mac sometimes opens without a focus, which causes top menu bar not to be populated (just `iv` on the left). Workaround is to switch back and forth once to and from another app for menu bar too get populated.

It is unclear to me how reliably replicate the bug so that I have a solid prove, that this fix works. But since building with this line it never happend to me yet.

Maybe more knowledgable in Qt can verify if it makes sense.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/iv/ivmain.cpp
+++ b/src/iv/ivmain.cpp
@@ -138,6 +138,8 @@ main(int argc, char* argv[])
     if (ap["rawcolor"].get<int>())
         mainWin->rawcolor(true);
 
+    QApplication::processEvents();  // Process any pending events
+
     // Make sure we are the top window with the focus.
     mainWin->raise();
     mainWin->activateWindow();


lgritz:
Thanks. I encounter this bug, too, but had no idea how to fix. I will test on my end.

lgritz (approved):
LGTM

I don't understand how or why this works, but I can totally confirm that it fixes the problem for me. I'd been encountering this for a while, but assumed it was some interaction between my still being on an old version of MacOS, and whatever version of Qt I had, or some other combination specific to me. I never would have known how to fix it entirely on our end with just one more Qt call. Thanks for fixing!!

PR #4774, opened 2025-05-18
feat(iv): Implement files drag and drop into an iv window

### LINKED ISSUE #4771: iv: Drag and drop file into an iv window to open it
It woudl be nice to have an ability to drag and drop an image file into an open iv window

soswow:
Drag and drop of files. Tested on my own Mac and it seem to work just fine.

Fixes #4771

### DIFF
--- a/src/iv/imageviewer.cpp
+++ b/src/iv/imageviewer.cpp
@@ -152,7 +152,13 @@ ImageViewer::ImageViewer(bool use_ocio, const std::string& image_color_space,
 
     setWindowTitle(tr("Image Viewer"));
     resize(m_default_width, m_default_height);
+    
     setAcceptDrops(true);
+    // Disable drag and drop on child widgets
+    for (QWidget* child : findChildren<QWidget*>()) {
+        child->setAcceptDrops(false);
+    }
+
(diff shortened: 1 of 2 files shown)


ssh4net:
In my experience, the drag and drop event is better disabled on child widgets in the main window. Otherwise, the view might not work.

I don't see something like this in the code. But did you test not only an empty window case but also the case when the image is already open?

soswow:
> In my experience, the drag and drop event is better disabled on child widgets in the main window. Otherwise, the view might not work.

I've added removal of DnD from children components
 
> I don't see something like this in the code. But did you test not only an empty window case but also the case when the image is already open?

I've tested with images already loaded and it was still working.

lgritz:
Tip for future PRs: I think you need to say "Closes #xxxx" or "Fixes #xxxx" in the body of the PR description to trigger GitHub to link and automatically close an issue. Other language like "implements" or "addresses" or other things that are clear to humans will not be recognized by GH.

lgritz (approved):
Works beautifully when I tested on my end! Thanks, this is a very cool feature.

LGTM

PR #4775, opened 2025-05-18
CI pkg config libdir fix

scott-wilson:
## Description

This was causing the Rust bindings to fail because the pkgconfig was pointing to a path that didn't exist (the libdir was the full root appended to itself).

## Tests

This should not make the CI fail,  but will affect the Rust build.

## Checklist:

  doesn't change the API.)
  nonconforming.


scott-wilson:
Ack, sorry for the extra noise. I have a commit from 2015 that wasn't DSO signed. I followed the recommendations and now there's almost 500 commits for a single change. 🙃

lgritz (approved):
LGTM

merge history aside, this is a one-line fix and it will all be squashed and rebased as part of the merge

lgritz:
How did you even do this?

It's not letting me merge because of the DCO check. I don't understand how you got into this state. Can you try this:

* Reset the ci_pkgconfig_build_fix branch hard to the current main
* Re-apply the one line change again
* `git commit -s` to sign it
* `git push origin ci_pkgconfig_build_fix --force`

scott-wilson:
Sorry about that, for some reason my Git-fu failed me. Commit recreated and pushed.

lgritz:
I was able to fix it up and push back to your tree.

lgritz (approved):
LGTM

PR #4777, opened 2025-05-19
fix(python): IBA.demosaic had GIL release in wrong spot

lgritz:
Only symptomatic on debug builds, but I was seeing exceptions from pybind11 about bad GIL release. I traced it to the recent implementations of demosaic. Apparently it's not safe to release the GIL until AFTER the call to py_to_stdvector. Which makes sense when you think about it -- we need the lock whenever we are interacting directly with Python calls or data structures (such as what py_to_stdvector is doing), but we release it when we're done with the python and only calling OIIO internals (such as IBA::demosaic() itself).

### DIFF
--- a/src/python/py_imagebufalgo.cpp
+++ b/src/python/py_imagebufalgo.cpp
@@ -2397,9 +2397,9 @@ IBA_demosaic_ret(const ImageBuf& src, const std::string& pattern = "",
                  py::object white_balance = py::none(), ROI roi = ROI::All(),
                  int nthreads = 0)
 {
-    py::gil_scoped_release gil;
     std::vector<float> wb;
     py_to_stdvector(wb, white_balance);
+    py::gil_scoped_release gil;
     return ImageBufAlgo::demosaic(src,
                                   { { "pattern", pattern },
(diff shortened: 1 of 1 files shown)


lgritz:
I've only seen it symptomatic on my Mac laptop. We obviously run the testsuite in debug mode in CI, not sure why it wasn't showing up. Maybe only certain combos of python version + pybind11 version catch the unsafe behavior and throw an exception for it?

EmilDohne:
I'm curious why the GIL is released in the first place, it should only ever be released if you want to make the c++ function parallelizable from python (if e.g. you have c++ code to get something from a server) but seeing as demosaic will likely saturate all threads this shouldn't really be necessary.

lgritz:
This allows any parallelization happening in the python program that's calling it to proceed in other threads and not get locked up during the long execution of the IBA function. It's not to parallelize this call, it's to allow other python threads to make progress during the expensive part of this operation.

EmilDohne:
Right I see, thanks for clearing up!

EmilDohne (approved):

PR #4779, opened 2025-05-21
fix: ImageOutput::check_open logic was flawed

lgritz:
We documented this function as checking the valid range of pixel coordinates, but we CALLED it everywhere as if we were only checking the resolution.

But we IMPLEMENTED as if we were checking the range, generating errors for outputting exr files with negative pixel window origins. But why was nobody seeing those errors when writing exr files with negative origins?  Because the openexr writer BOTCHED it by not noticing the return value and considering it a failure.

Oh boy. All my fault.

So the fix:

* Change the comment documenting check_open() to match the way we had been using it at the call sites.

* Change the implementation of check_open() to match this by no longer checking whether all four corners were in the range -- only check if each dimension's size is within the limit for that dimension.

* Fix the OpenEXR writer's failure to notice the check_open failure.

* Does this eliminate necessary checks? No! We already separately checked that nonzero origins were only allowed for formats that support it, and that negative origins were only allowed for formats that support that. So the flawed check was redundant anyway. Ha!

* Add a test verifying that we can write exr files with negative origins.  (Technically, that would have passed before, but only because we were ignoring the incorrect error.)

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -2164,11 +2164,12 @@ protected:
     ///     The ImageSpec that we are validating.
     ///
     /// @param range
-    ///     An ROI that describes the allowable pixel coordinates and channel
-    ///     indices as half-open intervals.  For example, the default value
-    ///     `{0, 65535, 0, 65535, 0, 1, 0, 4}` means that pixel coordinates
-    ///     must be non-negative and the width and height be representable by
-    ///     a uint16 value, up to 4 channels are allowed, and volumes are not
+    ///     An ROI that describes the allowable resolution and channel count:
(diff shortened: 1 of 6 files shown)


lgritz:
This is needed for a high priority bug fix on my end. I would really love if somebody had a chance to review ASAP.

1div0:
$#17 H4PP3N5

lgritz:
I let this linger for a few days because of a failing test that I didn't have time to chase down, but that turned out to be an unrelated issue I just fixed in #4782, and now that this PR has rebased on top of it, it passes.

So I'd love to get a review on this important bug fix and be able to merge early next week at the latest.

lgritz:
Two weeks old, no objections, fixes important bug -> merging

PR #4782, opened 2025-05-24
ci: Update linux arm clang reference output

lgritz:
Two PRs crossed paths in flight last week -- one adding the ARM tests, the other doing something that changed the reference output of this test slightly. Both passed on their own, but the combo left one platform's output needing the update.


lgritz:
It's not good to preemptively merge without review, but because this is breaking CI on other PRs and does not change any actual code, I'm going to merge immediately upon CI finishing.

PR #4783, opened 2025-05-25
api: image_span-ify ImageCache and TextureSystem

lgritz:
I think this is all the span-izing that the public interface of ImageCache and TextureSystem needs. And it doesn't break ABI compatibility (unlike the ImageInput and ImageOutput changes recently) because these aren't virtual methods.


lgritz:
Any comments? I would love to merge this before next week, since I already have the next follow-on (for ImageBuf) staged and ready to succeed this one.

This PR has no new design elements, it's just doing the things we already approved and changed in ImageInput and Output, and applying the same idioms to ImageCache and TextureSystem.

lgritz:
Last chance, any objections?

PR #4784, opened 2025-05-25
api: image_span_within_span

lgritz:
A utility to confirm that an image_span (possibly with non-contiguous strides) entirely fits within another contiguous span.


lgritz:
Any comments? I would love to merge this before next week.

lgritz:
No objections for a week -> merging

PR #4786, opened 2025-06-02
feat(iba): add 'auto' value for all options of IBA::demosaic

antond-weta:
## Description

# Changes
- The parameter 'pattern' now supports a special value 'auto', which is also default.
- The parameter 'layout' now supports a special value 'auto', which is also default.
- The parameter 'algorithm' now supports a special value 'auto', which is also default.
- New parameter 'white_balance_mode.
- The raw input plugin now adds an attribute 'raw:WhiteBalance' in addition to the libraw-specific 'raw:cam_mul', carrying the same values.

This changes the current behaviour, please read below.

Currently, the parameter 'pattern' defaults to 'bayer'. The new default value of 'auto' makes OIIO look for the input ImageBuf's attribute 'raw:FilterPattern', and deduct the pattern from its value. If the attribute is missing, 'bayer' is assumed.

Currently, the parameter 'layout' defaults to 'RGGB' for Bayer pattern, or 'GRBGBR BGGRGG RGGBGG GBRGRB RGGBGG BGGRGG' for X-Trans. The new default value of 'auto' makes OIIO look for the input ImageBuf's attribute 'raw:FilterPattern', and deduct the layout from its value. If the attribute is missing, the old default values are assumed.

Technically the 2 changes above the current behaviour may change, but only in a very rare use case: when the user relies on the old default values to decode the image, and there is metadata present in the ImageBuf saying that the pattern and/or layout is different from the default.

Currently, the parameter 'algorithm' defaults to 'linear'. The new default value 'auto' maps to 'MHC' for the Bayer pattern, and 'linear' for the X-Trans.

The newly added parameter 'white_balance_mode' can have one of the following three values:
- 'auto' (default) - the white-balancing weights are read from the input ImageBuf's attribute 'raw:WhiteBalance', defaulting to (1, 1, 1, 1).
- 'manual' - the white-balancing weights are read from the parameter 'white_balance'.
- 'none' - same as 'manual' with the weights of (1, 1, 1, 1).

This changes the current behaviour where the white-balancing weights from the parameter 'white_balance' get applied if present. Now, in addition to providing them, you need to explicitly set 'white_balance_mode' to 'manual'.

## Tests

I've updated the unit tests to cover the new 'auto' mode. 

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz:
I took the liberty of pushing an amendment to your fork to update a few test reference outputs, to see if we can get a CI pass.

lgritz:
I'm not quite sure how the "white-balance" is going to work from Python. By my read of it now, it defaults to passing 'none':
```
        .def_static("demosaic", &IBA_demosaic, "dst"_a, "src"_a,
                    "pattern"_a = "", "algorithm"_a = "", "layout"_a = "",
                    "white_balance"_a = py::none(), "roi"_a = ROI::All(),
                    "nthreads"_a = 0)
```

But then you pass
```
                                    ParamValue("white_balance", TypeFloat,
                                               wb.size(), wb.data()) },
```

which for the 'none' case will have no size. So I think this is passing the "white_balance" as if it were a single float? Or an array [0] floats? Something wrong. But then in the actual ImageBufAlgo::demosaic, you handle the case of `float[3]` and `float[4]` and string "auto", but whatever is passed by default from python, is not a case you explicitly handle. I think it just makes an error?

Maybe I'm misunderstanding this code, in which case you can ignore what I'm saying.

So I believe that in IBA_demosaic (the python helper function in py_imagebufalgo.cpp), instead of passing it unconditionally, I think you want to build up the list and only add the "white_balance" ParamValue if there seems to be a valid thing passed.

Now, as for your question about how to pass either the float arrays or strings from Python: While I'm sure there's a way to do it, I think taking very different types is going to be confusing, not just for Python but also C++ and how to document all of this. And if I'm not mistaken, the *only* valid string you can pass is "auto"?

So I wonder if it's simpler to make it recognize another optional parameter: "auto_white_balance" (int) that if 1, will ignore the float array "white_balance" and do the automatic thing. So it's an error to pass both "auto_white_balance" and "white_balance".

I dunno, I'm just throwing ideas out there.

antond-weta:
Thank you Larry, I'll take a look at the bindings bug.

It is correct, at this stage the only recognised string value is "auto". I was considering adding "none", as an alias to white_balance={1,1,1,1}, but might skip that now, if that makes python bindings more cumbersome.

Alternatively, instead of auto_white_balance, we could add white_balance_mode, supporting "auto", "none", and "custom".

lgritz:
Aha, and then so the float array parameter only is used if it's set to "custom"?

antond-weta:
Yes! I think I'm going to add that.

antond-weta:
I've added 'white_balance_mode', updated the tests and documentation.
I think this is ready to go.

lgritz:
Also, does any existing behavior change if you don't use any new options? In other words, is it safe to backport this patch to the release branch without breaking anybody's workflow from when the feature was first introduced? It's ok for us to change behavior for 3.1, but not for 3.0.7 -> 3.0.8.

antond-weta:
The old default behaviour changes in 2 ways:

- If there is no "pattern" or "layout" or "white-balance" specified **AND** the ImageBuf contains the attributes 'raw:FilterPattern' and/or "raw:WhiteBalance".
- The default algo for Bayer changes from "linear" to "MHC". No reason for this change, apart from being consistent with the other parameters' "auto" behaviour: if the user didn't give specifics, do your best. I can change this back, if you like.

I will update the description and let you know.

lgritz:
I don't have any intuition about (a) how common those circumstances are, (b) whether it's likely to make a visible change, or (c) how likely it is that the behavior change will be interpreted as a regression rather than a clear improvement.

I think there are three approaches here, and I leave it in your hands to choose:

1. Merge into main as is, but don't backport to 3.0. Both the new features and the behavior change are slated for 3.1 (release in about 3 months) and won't be available in the current release series.
2. Decide it's always an improvement that will be welcome (or, equivalently, that any changes will not be noticed), and backport as-is to 3.0.
3. Alter the code in the PR with some appropriate `#if OIIO_VERSION_GREATER_EQUAL(3,1,0)` to allow us to safely backport the patch to dev-3.0, with both branches getting new behavior when explicitly asked for. but default behaviors only changing in main (3.1).

lgritz:
A key piece is estimating how many people out there are actually using IBA::demosaic and will scrutinize the results enough to see or care about the difference caused by the default behavior change. If the answer is "probably only Anton would notice", then choice 2 is fine.

antond-weta:
Thanks Larry, 

I'd be happy with option 1, there is no rush on my side, but could you elaborate on the option 3? Do you mean something like this:

```
#if OIIO_VERSION_GREATER_EQUAL(3,1,0)
    //new behaviour
#elif defined(OIIO_USE_NEW_DEMOSAIC_DEFAULT_3_0)
    //new behaviour
#else
    //old behaviour
#endif
```

At what stage would we be able to drop these conditionals?

There is only one user of this code I know of, @ssh4net, who may want to chime in.

lgritz:
Simpler:

```
#if OIIO_VERSION_GREATER_EQUAL(3,1,0)
    //new options + new defaults
#else
    //new options (if selected), old defaults
#endif
```

This will result in main (already designated as "3.1 in progress") having your new behavior entirely. If we backport the PR to the current release branch (3.0) it will allow use of your new options, but for people who don't use them proactively, they'll get old default behavior. We could, then, later eliminate the "else" clause entirely after branching 3.1 and knowing that we're unlikely to want to do any more backports of this area of the code.

But if nobody is needing this change in the 3.0 branch (that is, main / 3.1 is fine, nobody will ever need a tagged, supported 3.0.x release with these changes), then option 1 is by far the simplest. For this, no `#if` is necessary, we simply will not backport the PR to the 3.0 branch.

lgritz:
How about this: Let's go with option 1 -- merge this now into main (to be in 3.1). If somebody comes along asking for a backport to the 3.0 branch, then we can decide which of the other options we prefer to do in order to to make a non-breaking backport.

What do you think?

antond-weta:
Thanks Larry, that works for me.
I've also updated the PR description.

lgritz (approved):

Issue #4759, opened 2025-05-15
[FEATURE REQUEST] Global attrib to enable/disable formats


lgritz:
From a [slack conversation](https://academysoftwarefdn.slack.com/archives/C05782U3806/p1747335172629159?thread_ts=1747334142.587609&cid=C05782U3806), maybe it would be helpful to have a global OIIO:attribute() way to set at runtime a specific list of which file formats OIIO is willing to read or write?

This is settable at build time now, of course. But I could imagine it being helpful, especially in cases where you have to live with whichever libOpenImageIO you find on your system, to be able to restrict which formats you support. One reason might be security or reliability -- essentially limiting the attack surface area to just the formats a particular app anticipates needing?

It would be fairly straightforward to implement.

Rahul-Baradol:
Would like to pick this!
Have been able to build locally, and will figure out the changes...

Could you please assign this to me?

MayankKumarSingh1:
Is this issue still open and needs to be assigned?

lgritz:
No, as you can see above, there is a PR currently in review that implements it, but I think it might need one more round of edits before it's ready to be merged.

PR #4789, opened 2025-06-03
feat: global attribute to enable/disable formats at runtime

Rahul-Baradol:
## Description
A solve for the issue #4759 

Added a OIIO attribute to enable/disable formats at runtime 

## Tests

- [ ] Proving input image formats can be restricted at runtime 
- [ ]  Proving output image formats can be restricted at runtime 

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imageio.cpp
+++ b/src/libOpenImageIO/imageio.cpp
@@ -59,11 +59,13 @@ int limit_imagesize_MB(std::min(32 * 1024,
 int imageinput_strict(0);
 ustring font_searchpath(Sysutil::getenv("OPENIMAGEIO_FONTS"));
 ustring plugin_searchpath(OIIO_DEFAULT_PLUGIN_SEARCHPATH);
-std::string format_list;         // comma-separated list of all formats
-std::string input_format_list;   // comma-separated list of readable formats
-std::string output_format_list;  // comma-separated list of writable formats
-std::string extension_list;      // list of all extensions for all formats
-std::string library_list;        // list of all libraries for all formats
+std::string allowed_input_format_list;      // list of input formats to be allowed at runtime 
+std::string allowed_output_format_list;     // list of output formats to be allowed at runtime 
+std::string format_list;                    // comma-separated list of all formats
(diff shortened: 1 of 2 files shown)


src/libOpenImageIO/imageio.cpp
```diff
@@ -676,6 +686,20 @@ getattribute(string_view name, TypeDesc type, void* val)
         *(float*)val = IB_total_image_read_time;
         return true;
     }
+    if (name == "allowed_input_formats" && type == TypeString) {
+        if (allowed_input_format_list.empty()) {
+            return false;
+        }
+        *(std::string*)val = allowed_input_format_list;
```
lgritz, line 693: If you look at some of the other string examples, you'll see that it's not actually a std::string that val points to, it's really a ustring.

Rahul-Baradol: Got it!

lgritz:
Did you close this on purpose?

Rahul-Baradol:
Yes..
Because I used the wrong email for the CLA signoff

lgritz:
OK, though I think you could have just amended the current PR and force push it.

Rahul-Baradol:
It was EasyCLA signoff, and I couldn't find any option over there..
Then I realized, I could use DCO.

lgritz:
You need both DCO and CLA.

Rahul-Baradol:
So is there a way for me use a different email ID?
Apparently the new PR I created, is using the other email for the CLA then

lgritz:
`git commit --amend --author="my name <my@fixed.email>"`

then force push?

Or I think EasyCLA can be told your github id as the key instead of your email, and then any emails associated with that (can you have multiple?) will work?

Rahul-Baradol:
I tried something similar, and that didn't work...

I reached out to their support, and they revoked the signature.
They told  that  the platform picks any one email from Github account, and it ended up picking the wrong one.

I have removed that mail, and re-signed using the correct mail.

We are good to go!

PR #4794, opened 2025-06-05
fix: build break from untested combo of two PRs

lgritz:
Oops, my bad. Looks like #4664 (merged about a week ago) and #4783 (merged today, but authored before #4664 was merged) interacted in a way that produced a build break. This tiny fix is essential to unbreak it.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -3915,7 +3915,7 @@ ImageCacheImpl::add_tile(ustring filename, int subimage, int miplevel, int x,
     }
     if (chend < chbegin) {  // chend < chbegin means "all channels."
         chbegin = 0;
-        chend   = file->spec(subimage, miplevel).nchannels;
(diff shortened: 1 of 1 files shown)


lgritz:
Need to fast track this emergency fix. If CI fully passes, I will merge it preemptively. 

So sorry, same-day merges, especially those without a review, should be vanishingly rare. But this is an embarrassingly serious build break that has a one-line fix.

lgritz:
Not adding, subtracting -- the (private) call changed in one of the PRs to remove an unneeded argument (and the other PR added a call to the one with the old signature).

We decided that each MIP level didn't need a separately allocated ImageSpec because there's no format we support that allows arbitrary metadata to be set PER MIPLEVEL. Resolution, etc. is the only thing that changes from miplevel to miplevel within one subimage.

ThiagoIze:
oh indeed. I got the two sides confused.

PR #4795, opened 2025-06-06
Use R"()" syntax for glsl shader strings for better readability

soswow:
Small PR to improve redability of shader code parts.


lgritz (approved):
Much nicer, thanks!

PR #4796, opened 2025-06-08
Add raw string syntax modifier for VSCode and Cursor to be able to sy…

soswow:
Sorry for not doing it in a first PR. I thought to see if I can make VSCode / Cursor to actually syntax highlight that raw string literal content. And I found a way. Seems like if you change the delimiter using language of content syntax name you make VSCode to syntax highlight it:

Screenshot 2025-06-08 at 11 12 45 pm

I don't think those clang exceptions are needed now.


lgritz:
Wow, I had no idea about that trick!

PR #4797, opened 2025-06-09
ci: Various ccache save/restore improvements for CI runs

lgritz:
Improvements to the way our CI utilizes ccache to speed up build. Ported from OSL:

- Split caching into separate restore/save, and rearrange the step order so we save cache before tests run. This allows us to save the ccache even for jobs with failed tests (where the all-in-one unsplit approach would skip the cache save step for a failed job), or for jobs that we purposely end early (maybe because we see failing tests), not only when all tests finish and succeed. This makes subsequent builds, that we do on the same branch while attempting to fix failures, go faster by using the cache.

- Make subsequent pushes to the same PR or REF to cancel any previous jobs, so if we push updates before the last CI run of the same branch completes, it won't keep those going as useless zombie jobs.

- Change the cache key names we use to remove the github ref to encourage cache reuse across different commits or branches (I think), and to fix a couple errors where we had mis-named things and were sharing caches between jobs when we shouldn't or vice versa.

- Be more careful about the location of the cache, making sure we were for sure telling the GHA cache action to be saving/restoring the same directory location that we told ccache to use as its cache location.

- Use CCACHE_COMPRESSION env var to be extra sure we're using compressed caches to be sure we get the best bang for buck on the limited cache storage we're allowed.

- Set CMAKE_CXX_COMPILER_LAUNCHER and CMAKE_C_COMPILER_LAUNCHER in the shell for CI in the hopes of making more dependencies we build also use ccache.

- Alternate launcher structure -- use LANG_COMPILER_LAUNCHER, not RULE_LAUNCH.

A sampling of typical results:

    test                         build deps     build oiio
    -----------------------      ----------     ----------
    VP 2024 (cold)                N/A            6:22
    VP 2024 (cached)              N/A            0:55 (7x)

    latest rels (cold)            8:29           5:30
    latest rels (cached)          2:24 (3.5x)    0:45 (7.3x)

(The N/A is because that run gets all its dependencies from the container and doesn't have to build any from scratch.)

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -106,25 +106,23 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - name: Prepare ccache timestamp
-        id: ccache_cache_keys
-        shell: bash
-        run: echo "date=`date -u +'%Y-%m-%dT%H:%M:%SZ'`" >> $GITHUB_OUTPUT
-      - name: ccache
-        id: ccache
-        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
-        with:
-          path: ./ccache
-          key: ${{github.job}}-${{inputs.nametag}}-${{steps.ccache_cache_keys.outputs.date}}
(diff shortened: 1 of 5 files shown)


lgritz:
This CI-only PR has been in review for 5 days without comment. I will merge this over the weekend if nobody objects.

PR #4798, opened 2025-06-10
ci: fix broken bleeding edge CI

lgritz:
Just on the "bleeding edge" CI job, we were refraining from doing the apt-get update. It was breaking something last year.  Now NOT doing it is breaking something in this CI test, just since this week (presumably some default package on the GH runners changed, I haven't been able to isolate any other reason). But bringing all the apt packages up to date makes it go away.


lgritz:
Ping. Since this fixes broken CI, I would like to merge this ASAP. Will do tomorrow if there are no objections before then.

PR #4800, opened 2025-06-13
api: Add new ImageInput::supports() query: "mipmap"

lgritz:
ImageOutput::supports() recognized queries for "multiimage" and "mipmap".  But ImageInput::supports() only recognized "multiimage". Add "mipmap" to make it easier for client software to know if the image file format they're reading from allows the possibility of MIP-mapping.

As far as I can tell, only OpenEXR, TIFF, Ptex, and DDS do, among the file types we currently support.

### DIFF
--- a/src/dds.imageio/ddsinput.cpp
+++ b/src/dds.imageio/ddsinput.cpp
@@ -35,7 +35,7 @@ public:
     const char* format_name(void) const override { return "dds"; }
     int supports(string_view feature) const override
     {
-        return feature == "ioproxy";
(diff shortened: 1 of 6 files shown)


lgritz:
Opinions?

Since it's a new "supports" query, it should not break anything, though authors of any custom ImageInput subclasses are expected to return true for this query if they implemented a format that supports mipmaps.

lgritz:
I would like to merge this by the end of the week, if nobody objects.

jessey-git (approved):

PR #4802, opened 2025-06-14
ci:  Automatically push updates to stubs to the current branch

chadrik:
## Description

This simplifies developer workflow by automatically pushing freshly built stubs to the current branch for review. 

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -147,6 +147,9 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
 
       - name: Install Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
@@ -171,11 +174,24 @@ jobs:
             ./wheelhouse/*.whl
             ./wheelhouse/OpenImageIO/__init__.pyi
         # if stub validation fails we want to upload the stubs for users to review
-        if: success() || failure()
+#         if: success() || failure()
 
-      # Commit all changed files back to the repository
-      - uses: stefanzweifel/git-auto-commit-action@v6
-        commit_message: "Automatic update to python stubs"
+      - name: Copy stubs to repo
+        run: |
+          pwd
+          ls -la .
+          ls -la ./wheelhouse/
+          if [ -f ./wheelhouse/OpenImageIO/__init__.pyi ]; then
+            echo "Copying stubs into repo"
+            cp ./wheelhouse/OpenImageIO/__init__.pyi ./src/python/stubs/OpenImageIO/__init__.pyi
+          fi
+#         if: failure()
+
+      - name: Commit changed stubs back to the repository
+        uses: stefanzweifel/git-auto-commit-action@v6
+        with:
+          commit_message: "Automatic update to python stubs"
+#         if: failure()
 
   # ---------------------------------------------------------------------------
   # Linux ARM Wheels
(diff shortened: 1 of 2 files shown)


lgritz:
I'm not sure I understand the security implications of this approach.

chadrik:
> I'm not sure I understand the security implications of this approach.

I'm not totally sure either, tbh.  Once I get it working I'll explore finer grained permissions, unless you'd like me to stop. 

This approach was inspired by some feedback from @zachlewis over at https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2162#issuecomment-2970617821

I don't have permissions to run any CI in that repo, so I thought I'd test it out here.

chadrik:
Status update. 

My goal is to push updates to the pull request branch in my fork.   

I got pretty close to making this work, but I think I've reached an impasse.  

First, I configured the `checkout` action to pull from my origin and checkout my branch (the default behavior fetches the forked source in a way that hides the fork url and branch name).  

I'm using https://github.com/stefanzweifel/git-auto-commit-action action to auto-commit-and-push changes, and it's doing what I expect -- pushing to my fork and branch -- but it fails with a permissions issue:

```
remote: Permission to chadrik/oiio.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/chadrik/oiio/': The requested URL returned error: 403
```

The PR is set to allow pushes from maintainers, but apparently the actions bot does not count.  I'm not sure how to get past this, but it might require me to setup a personal access token, which would make this workflow prohibitively cumbersome.

Some alternatives to pursue:
- use `pull_request_target` trigger instead of `pull_request`.  This runs with the full perms of the base repo, but it's considered a major security threat.  More info [here](https://github.com/orgs/community/discussions/32032)
- Use [autofix-ci](https://github.com/marketplace/autofix-ci#pricing-and-setup).  This seems designed to solve exactly this problem and it's free for open source projects.

lgritz:
@chadrik, I'm still a little nervous about a fully automatic push to the PR author's branch (50% - will they not like that; 50% - is there any way somebody can abuse it to push things to our own repo before a merge?). But in [this Slack conversation](https://academysoftwarefdn.slack.com/archives/C0169RX7MMK/p1748113465760779?thread_ts=1748109895.870809&cid=C0169RX7MMK), @JeanChristopheMorinPerso outlines a way where, after a PR fails for this particular reason, we could add a tag that would run a special "fix stubs" workflow that would do the fix, push the branch to the author to amend the PR, and then remove the tag. 

Not sure how you feel about that approach, or if you fully understand how to do it (I get it conceptually but would probably have to do a lot of documentation reading and trial-and-error to know how to implement it, because I'm a workflow newbie), but it feels to me like it's inherently safer to require that minor manual step to kick off the fixing and pushing, rather than it being completely automatic and irreversible simply upon receipt of the PR.

OTOH, if @JeanChristopheMorinPerso is not concerned about the security of the "automatic" approach (he both knows more than me about this and seems pretty conservative about it), then I guess I'm willing to go along with that if nobody voices a concern.

JeanChristopheMorinPerso:
>I'm using https://github.com/stefanzweifel/git-auto-commit-action action to auto-commit-and-push changes, and it's doing what I expect -- pushing to my fork and branch -- but it fails with a permissions issue:
>
>```
>remote: Permission to chadrik/oiio.git denied to github-actions[bot].
>fatal: unable to access 'https://github.com/chadrik/oiio/': The requested URL returned error: 403 
>```
>The PR is set to allow pushes from maintainers, but apparently the actions bot does not count. I'm not sure how to get past this, but it might require me to setup a personal access token, which would make this workflow prohibitively cumbersome.

Yes. The GHA token has relatively limited permissions and it can't impersonate you. One way to solve this is to use a PAT like you said. Another way is to create an app, grant it appropriate permissions and I think that would remove some maintenance burden (since the app token would never expire, as opposed to a PAT). But that requires the use of `pull_request_target`.

PR #4803, opened 2025-06-14
fix(oiiotool): fixes to --missingfile behavior

lgritz:
Fix some `oiiotool --missingfile` behavioral corner cases:

When substituting, warn rather than making a true error. This didn't seem like a big deal for a single command line, but when doing a loop over frames, it would prevent correct behavior from pressing on to the next iteration.

How big is the substitute image? It uses the rez and channels of the first image successfully read (if available). But it when in a frame loop, it resets this on every iteration, so any missing frame looks like it doesn't have a good size guess -- but it does, as long as it's not the first iteration of the loop. This involves safely gettingthe first image dimensions from the *parent* (each iteration is a child Oiiotool object).

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -174,8 +174,8 @@ Oiiotool::clear_options()
     printinfo_nometamatch.clear();
     printinfo_verbose = false;
     clear_input_config();
-    first_input_dimensions = ImageSpec();
-    output_dataformat      = TypeDesc::UNKNOWN;
+    m_first_input_dimensions = ImageSpec();
+    output_dataformat        = TypeDesc::UNKNOWN;
     output_channelformats.clear();
     output_bitspersample      = 0;
     output_scanline           = false;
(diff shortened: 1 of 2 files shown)


AndreAnjosTD:
Tested the flag `--missingfile` with command:
`oiiotool --parallel-frames --missingfile black -i path/to/sequence/render.#.exr --frames 1001-1200 --framepadding 4 --if "{SUBIMAGES > 1}" --subimage Beauty --endif --tocolorspace sRGB -o path/to/output/render.#.jpg`
Our pipe worked with no issues after internal unit tests and production testing. This resolves missing frames as well as pixel reading errors, if used with [PR 4804](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4804).

AndreAnjosTD (approved):
Tested the flag --missingfile with command:
`oiiotool --parallel-frames --missingfile black -i path/to/sequence/render.#.exr --frames 1001-1200 --framepadding 4 --if "{SUBIMAGES > 1}" --subimage Beauty --endif --tocolorspace sRGB -o path/to/output/render.#.jpg`
Our pipe worked with no issues after internal unit tests and production testing. This resolves missing frames as well as pixel reading errors, if used with https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4804.

PR #4804, opened 2025-06-14
feat(oiiotool): New expression pseudo-metadata term: SUBIMAGES

lgritz:
When doing oiiotool command line expression substitution, asking for this metadata evaluates to the number of subimages in the referenced image.

The purpose is that asking to retrieve, say, `{TOP.'oiio:subimages'}` might look right, but generally it's only added to images (or format readers) that are potentially multi-image. Whereas SUBIMAGES is not actual image metadata, it's just an evaluation token that will always be the number of subimages.


zachlewis:
maybe call it NSUBIMAGES?

AndreAnjosTD (approved):
The `SUBIMAGES` token is working great and passed our unit and production tests.

lgritz:
@zachlewis I'm on the fence, but I think I'm going to leave it as SUBIMAGES.

PR #4805, opened 2025-06-15
fix(python): python ImageBuf.init_spec did not return correct value

lgritz:


lgritz:
Not many PRs easier to review than this one...

grdanny (approved):
That is true. :-)

PR #4809, opened 2025-06-21
ci: Simplify gh-win-installdeps, no more vcpkg

lgritz:
Among other simplifications, remove all vcpkg builds from windows CI, don't need them.  Just rely on our auto-build. Shortens the build time on Windows!


lgritz:
I'm gonna take Zach's popper as a review and merge this CI-only fix.

PR #4810, opened 2025-06-23
build(jxl): Use correct cmake variables for the include directories

jessey-git:
I'm unsure how this could have worked correctly. There were three different variables used before, now there is only one.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/modules/FindJXL.cmake
+++ b/src/cmake/modules/FindJXL.cmake
@@ -41,5 +41,4 @@ find_package_handle_standard_args(JXL
 
 if(JXL_FOUND)
   set(JXL_LIBRARIES ${JXL_LIBRARY} ${JXL_THREADS_LIBRARY})
-  set(JXL_INCLUDES ${JXL_INCLUDE_DIR})
 endif(JXL_FOUND)
--- a/src/jpegxl.imageio/CMakeLists.txt
+++ b/src/jpegxl.imageio/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 if (JXL_FOUND)
     add_oiio_plugin (jxlinput.cpp jxloutput.cpp
-                     INCLUDE_DIRS ${JXL_INCLUDE_DIRS}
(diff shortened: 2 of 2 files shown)


lgritz:
I think the failing CI is unrelated to this -- we have recently seen sporadic failure that I don't yet understand, related to not finding pystring despite just having built it immediately prior.

When the whole CI run is done, just hit the "rerun failed tests" and see if it clears up.

lgritz:
Explanatory, not asking for changes:

Near as I can tell, it's a fairly common CMake module idiom to use FOO_INCLUDE_DIR as the internal-to-module variable where it stores the find_path results, but then if everything is found in the end, usually they are copied to the FOO_INCLUDES and FOO_LIBRARIES as the main variables used by downstream consumers. Just so you know, that's why that copy was probably there. You are, of course, totally right that in one place it used INCLUDE_DIR and in another INCLUDE_DIRS, that was a legit bug.

But this is fine. The above may be the preferred idiom, but SO MANY modules written prior to that being established as the preference are still circulating that directly use FOO_INCLUDE_DIR the old way. And the new new way is supposed to be setting targets, so downstream projects don't use those variables at all. So who cares, I guess.

lgritz (approved):
LGTM

src/cmake/modules/FindJXL.cmake
```diff
@@ -41,5 +41,4 @@ find_package_handle_standard_args(JXL
 
 if(JXL_FOUND)
   set(JXL_LIBRARIES ${JXL_LIBRARY} ${JXL_THREADS_LIBRARY})
-  set(JXL_INCLUDES ${JXL_INCLUDE_DIR})
```
kmilos, line 44: This should've been `set(JXL_INCLUDE_DIRS ${JXL_INCLUDE_DIR})` as recommended by CMake, and then the second change below would have not been necessary.

Also note that `JXL_INCLUDE_DIR` is `mark_as_advanced`, and isn't really intended to be used outside the find module...

PR #4812, opened 2025-06-26
docs: Clarify description of JPEG XL faster decoding modes

### LINKED ISSUE #4584: [DOCS] `jpegxl:speed` incorrectly listed/explained
In [the docs](https://openimageio.readthedocs.io/en/v3.0.1.0/builtinplugins.html#jpeg-xl) `jpegxl:speed` is described as setting the speed of encoding, even though that's what `jpegxl:effort` directly above it does. The description even says 
> Note: in libjxl it named JXL_ENC_FRAME_SETTING_DECODING_SPEED. But it is about encoding speed and compression quality, not decoding speed.

Which is incorrect. The function is meant to adjust decoding speed, but in it's current implementation in libjxl that means disabling expensive coding tools, which also improves encoding speed by chance. In the case of lossless, it can cause [exponentially larger filesizes](https://github.com/libjxl/libjxl/issues/3532) than just lowering the effort setting.

jonnyawsom3:
Fixes #4584
A minor edit with large consequences.

Previously the setting would encode uncompressed files due to a bug, causing the much faster encoding as stated in the old description.

In the upcoming v0.12 libjxl release, [my PR](https://github.com/libjxl/libjxl/pull/4201) overhauls the setting, to correctly scale decode speed at the cost of density.


jonnyawsom3:
That was one hell of a CLA to sign for a text edit haha.

src/doc/builtinplugins.rst
```diff
@@ -1345,11 +1345,9 @@ control aspects of the writing itself:
        For lossy, higher effort should more accurately reach the target quality.
    * - ``jpegxl:speed``
      - int
-     - Sets the encoding speed tier for the provided options. Minimum is 0
-       (slowest to encode, best quality/density), and maximum is 4 (fastest to
-       encode, at the cost of some quality/density). Default is 0.
-       (Note: in libjxl it named JXL_ENC_FRAME_SETTING_DECODING_SPEED. But it
-       is about encoding speed and compression quality, not decoding speed.)
+     - Sets the decoding speed tier for the provided options. Minimum is 0
```
lgritz: These are output options. Is this really something that controls the decoding speed tier? Or encoding?

lgritz:
> That was one hell of a CLA to sign for a text edit haha.

Yeah, sorry about that. It seems silly to need the CLA for certain kinds of trivial or non-code changes. But logistically it's a lot easier to just require it across the board and enforce it uniformly and automatically, than it would be to try to make a reliable way to capture the subtlety of which cases really need it.

src/doc/builtinplugins.rst
jonnyawsom3: As the name in libjxl suggests, `JXL_ENC_FRAME_SETTING_DECODING_SPEED` increases the decoding speed with higher levels. This is done by using less expensive coding tools, which can have a marginal impact on encode speed, but a major impact on decode speed.
| Level | BPP | Encode (MP/s) |  Decode (MP/s) |
|-------|------|-----------------|------------------|
| Faster Decoding 0 | 5.480 | 6.240 | 38.678 |
| Faster Decoding 1 | 5.534 | 8.401 | 69.939 |
| Faster Decoding 2 | 5.695 | 8.637 | 83.142 |
| Faster Decoding 3 | 6.545 | 3.760 | 162.619 |
| Faster Decoding 4 | 6.685 | 3.906 | 180.708 |

Fresh results just for you. As you can see, level 3 and 4 were slower to encode than 0, while being much faster to decode. These results are based on a PR that will be released in the next version of libjxl https://github.com/libjxl/libjxl/pull/4201.

jonnyawsom3: Heh, by pure chance FD2 is 2x faster and FD4 is 4x faster on that image. Funny how things like that happen.

jonnyawsom3:
I'll be away for the weekend, but hopefully those [benchmark results](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4812#discussion_r2170091808) are evidence enough. Thanks for giving this a look.

src/doc/builtinplugins.rst
lgritz: Maybe then just alter the description a little bit more to make it clear that this is a control that alters all of the compression level, encoding speed, and decoding speed? As written, I think somebody could read this description and get the impression that it only affects decode performance and not size or encoding performance.

Also, just as an aside, do you have any layman's explanation for the behavior illustrated by the chart? I would naively expect that a control that achieves progressively better compression would make encoding slower, but level 0's encoding speed is somewhere in the middle, which seems strange.

lgritz:
Looking at your performance chart... is level 0 the best default choice for us? Most files are read many times more than they are written. Should our default be one that is a more middle-of-the-road tradeoff that accepts slightly worse compression for dramatically better reading speed?

Put another way, the whole span of your chart from best to worst is only a 20% difference in file size, but a 2.3x difference in encode speed, and nearly a 5x difference in decode speed. What should we be optimizing for in OIIO, for users who aren't expert enough in JPEG-XL to know to set this option specifically?

I'm also still really confused by the chart. Going from level 0 to 4, the compression ratio falls monotonically, the decode speed goes up monotonically, but the encode speed is... weird. What's going on with that? I would have expected a straightforward relationship where encode speed is slower the better the compression ratio is.

jonnyawsom3:
I agree, it is a sharp increase in decoding speed for minimal impact on density. We're looking into changing default encoder settings, so they'll be more like FD1 than current, but heuristics change depending on image content, resolution, ect. So it's not always a simple thing.
For now I'd recommend leaving it at level 0, as we'll be changing things upstream. With this corrected documentation, users noticing slow decode speed will be able to find the option and use it.

Results also change based on effort level, image content and even device hardware and compiler from our testing. The faster decoding levels disable expensive compression techniques or replaces them with simpler ones. On some images those simple versions work best or have minimal impact, sometimes it can be quite significant.

To maintain high compression ratios, the simpler compression techniques are put through more expensive implementations. This maintains the same high decode apeed, but regains some lost density too. Our methodology behind the new results are similar to yours. If you're using faster decoding, you probably care less about fast encoding, so we apply heavier encoding per effort level to squeeze the most out of it.

I'm typing this on my phone, on a 3 hour delayed train... So apologies if it's hard to understand. Maybe we could merge this and discuss changing defaults in another issue or PR?

lgritz:
So is it not a single monotonic scale? Are you saying that 1-4 form a true scale, but "0" just means "default" which might not be "less than 1" at all, and may change its value in different versions?

jonnyawsom3:
Right now you can use `cjxl -d 0 -e 2 -g 0` and you get almost the same results as FD3, just slightly less dense. We've not had enough time to find optimal parameters, just fix the completely broken version prior to v0.12 which was uncompressed.

FD0 is indeed defaults, or whatever parameters you specify to libjxl. The defaults will naturally change over time, as improvements are made and parameters tested.

lgritz:
OK, knowing that 0 is not part of a single 0,1,2,3,4 scale makes the table results make a LOT more sense.

jonnyawsom3:
Right, my apologies. It was originally labelled Default, but I think I changed it so I could just have a 'Level' column without overcomplicating it. As usual though, that just made it worse.

lgritz:
Maybe it would be more clear if the explanation was a little more explicit about all of this, for example (you don't need to use my exact language):

> - Sets the decoding speed tier. Values 1 to 4 offer progressively faster decompression speed but moderately lower compression ratio. Encoding speed does not have a clear monotonic relationship to the decoding speed tier. The default value 0 selects the speed vs compression default as determined by libjxl, which as of version X.Y happens to be the best compression ratio and slowest decode speed, but may select a different tradeoff in future releases.

lgritz:
By the way, just to make sure this does not go unsaid: Thanks for helping with this, it's great to have somebody from an upstream project tell us how we can improve use of a dependency. There's a lot to juggle in a project like this and we can't be experts in every dependency, especially for the file formats that are only just beginning to get use in our domain. We really appreciate the extra guidance.

jonnyawsom3:
What about
> Sets the decoding speed tier. Values 1 to 4 offer progressively faster decoding speed but lower compression ratio. Encoding speed is variable between levels, but still moderated with the effort setting. Default value is 0 for highest compression ratio/quality but slowest decoding.

I don't mention libjxl version as if we increase the default decoding speed, faster decoding will also be tweaked to maintain a meaningful difference. It should be assumed that things will change between libjxl versions, but we'll try and make sure that faster decoding is always faster than default.

lgritz:
Yeah, that seems fine, let's go with that.

jonnyawsom3:
Currently in bed and tried to do it on my phone, so hopefully it worked. If there's formatting issues, either you can correct them or I will tomorrow. Thanks for seeing this though and being interested in the format, there's certainly lots to learn.

lgritz (approved):
LGTM

PR #4814, opened 2025-06-28
api: image_span-ify ImageBuf

lgritz:
We've already done ImageInput, ImageOutput, ImageCache, TextureSystem.

This applies the analogous set of changes to the public APIs of ImageBuf -- adding span and image_span versions of all the API calls that previously required raw pointers and strides. Many of the internals are still using the old versions, but at least this takes care of the interface.

### DIFF
--- a/src/include/OpenImageIO/image_span.h
+++ b/src/include/OpenImageIO/image_span.h
@@ -70,7 +70,6 @@ public:
         // Validations:
         // - an image_span<byte> can have any chansize, but any other T must
         //   have the chansize equal to the data type size.
-        OIIO_DASSERT(nchannels > 0 && width > 0 && height > 0 && depth > 0);
         OIIO_DASSERT((std::is_same<std::remove_const_t<T>, std::byte>::value)
                      || chansize == sizeof(T));
 
@@ -380,10 +379,11 @@ template<typename T, size_t Rank>
 image_span<std::byte>
 as_image_span_writable_bytes(const image_span<T, Rank>& src) noexcept
 {
-    return image_span<std::byte>(reinterpret_cast<std::byte*>(src.data()),
-                                 src.nchannels(), src.width(), src.height(),
-                                 src.depth(), src.chanstride(), src.xstride(),
-                                 src.ystride(), src.zstride(), src.chansize());
+    return image_span<std::byte>(
+        const_cast<std::byte*>(reinterpret_cast<const std::byte*>(src.data())),
+        src.nchannels(), src.width(), src.height(), src.depth(),
+        src.chanstride(), src.xstride(), src.ystride(), src.zstride(),
+        src.chansize());
 }
 
 
--- a/src/include/OpenImageIO/image_span.h
+++ b/src/include/OpenImageIO/image_span.h
@@ -373,19 +373,17 @@ as_image_span_bytes(const image_span<T, Rank>& src) noexcept
 }
 
 
-/// Convert an image_span of any type to a mutable span of bytes covering
-/// the same range of memory.
+/// Convert an image_span of any nonconst type to a mutable span of bytes
+/// covering the same range of memory.
 template<typename T, size_t Rank>
 image_span<std::byte>
 as_image_span_writable_bytes(const image_span<T, Rank>& src) noexcept
 {
-    return image_span<std::byte>(
-        const_cast<std::byte*>(reinterpret_cast<const std::byte*>(src.data())),
-        src.nchannels(), src.width(), src.height(), src.depth(),
-        src.chanstride(), src.xstride(), src.ystride(), src.zstride(),
-        src.chansize());
+    return image_span<std::byte>(reinterpret_cast<std::byte*>(src.data()),
+                                 src.nchannels(), src.width(), src.height(),
+                                 src.depth(), src.chanstride(), src.xstride(),
+                                 src.ystride(), src.zstride(), src.chansize());
 }
-
 
 /// Verify that the image_span has all its contents lying within the
 /// contiguous span.
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -496,7 +496,7 @@ ImageBufImpl::ImageBufImpl(string_view filename, int subimage, int miplevel,
         m_nativespec = *spec;
         if (buforigin || bufspan.size()) {
             if (!buforigin)
-                buforigin = bufspan.data();
+                buforigin = static_cast<const void*>(bufspan.data());
             set_bufspan((char*)buforigin, xstride, ystride, zstride);
             m_storage      = ImageBuf::APPBUFFER;
             m_pixels_valid = true;
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -133,6 +133,22 @@ cspan_from_buffer(const void* data, TypeDesc format, int nchannels, int width,
 
 
 
+// Unsafe version of as_image_span_writable_bytes with a const_cast. Keep out
+// of the public APIs, just used as necessary utility here in imagebuf
+// internals.
+template<typename T, size_t Rank>
+image_span<std::byte>
+force_as_image_span_writable_bytes(const image_span<T, Rank>& src) noexcept
+{
+    return image_span<std::byte>(
+        const_cast<std::byte*>(reinterpret_cast<const std::byte*>(src.data())),
+        src.nchannels(), src.width(), src.height(), src.depth(),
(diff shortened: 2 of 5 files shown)


src/libOpenImageIO/imagebuf.cpp
```diff
@@ -685,20 +737,29 @@ ImageBuf::operator=(ImageBuf&& src)
...
+    // size that we previously had, we shouldn't need to free and then
+    // immediatel re-allocate again, right?
     if (m_allocated_size)
         free_pixels();
     try {
         m_pixels.reset(size ? new char[size] : nullptr);
-        // Set bufspan to the allocated memory
-        m_bufspan = { reinterpret_cast<std::byte*>(m_pixels.get()), size };
+        // Set m_bufspan to the allocated memory
+        set_bufspan(m_pixels.get());
```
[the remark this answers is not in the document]
lgritz: I will address this with an amendment.

lgritz:
@EmilDohne any thoughts?

src/include/OpenImageIO/image_span.h
```diff
@@ -380,10 +379,11 @@ template<typename T, size_t Rank>
 image_span<std::byte>
 as_image_span_writable_bytes(const image_span<T, Rank>& src) noexcept
 {
-    return image_span<std::byte>(reinterpret_cast<std::byte*>(src.data()),
-                                 src.nchannels(), src.width(), src.height(),
-                                 src.depth(), src.chanstride(), src.xstride(),
-                                 src.ystride(), src.zstride(), src.chansize());
+    return image_span<std::byte>(
```
EmilDohne: I do have some concerns about blindly allowing conversion from const-qualified objects to non-const ones as part of the public API. I don't necessarily mind them if they are narrow in scope but I think here I'd prefer doing some kind of compile-time check on const objects.

This may be outside of the scope of this PR so feel free to tackle this as part of another one

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -342,53 +356,51 @@ class ImageBufImpl {
...
+    ustring m_name;                 // Filename of the image
+    ustring m_fileformat;           // File format name
+    int m_nsubimages       = 0;     // How many subimages are there?
+    int m_current_subimage = 0;     // Current subimage we're viewing
+    int m_current_miplevel = 0;     // Current miplevel we're viewing
+    int m_nmiplevels       = 0;     // # of MIP levels in the current subimage
+    mutable int m_threads  = 0;     // thread policy for this image
+    ImageSpec m_spec;               // Describes the image (size, etc)
+    ImageSpec m_nativespec;         // Describes the true native image
+    std::unique_ptr<char[]> m_pixels;  // Pixel data, if local and we own it
```
EmilDohne, line 386: [nit] Prefer the previously added `default_init_vector`, potentially even with `std::byte` since if I understand this correctly this is meant to act as pixel data for uint8_t, uint16_t, half etc.

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -427,52 +483,32 @@ ImageBufImpl::ImageBufImpl(string_view filename, int subimage, int miplevel,
...
-    , m_xstride(0)
-    , m_ystride(0)
-    , m_zstride(0)
-    , m_channel_stride(0)
     , m_contiguous(false)
     , m_imagecache(imagecache)
-    , m_allocated_size(0)
-    , m_write_tile_width(0)
-    , m_write_tile_height(0)
-    , m_write_tile_depth(1)
```
EmilDohne, line 451: nothing to add except that I think this is a good change!

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -427,52 +483,32 @@ ImageBufImpl::ImageBufImpl(string_view filename, int subimage, int miplevel,
...
+            if (!buforigin)
+                buforigin = static_cast<const void*>(bufspan.data());
+            set_bufspan((char*)buforigin, xstride, ystride, zstride);
             m_storage      = ImageBuf::APPBUFFER;
             m_pixels_valid = true;
         } else {
-            m_storage = ImageBuf::LOCALBUFFER;
+            m_storage  = ImageBuf::LOCALBUFFER;
+            m_readonly = false;
+            alloc(*spec);
```
EmilDohne: [nit] For something that is so close to c-library names I always prefer being explicit with `this->alloc()`

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -622,6 +652,28 @@ ImageBuf::ImageBuf(const ImageSpec& spec, void* buffer, stride_t xstride,
...
+}
+
+
+
+ImageBuf::ImageBuf(const ImageSpec& spec, const image_span<std::byte>& buffer)
+    : m_impl(new ImageBufImpl("", 0, 0, nullptr /*imagecache*/, &spec, buffer,
+                              true /*readonly*/, nullptr /*config*/,
+                              nullptr /*ioproxy*/),
+             &impl_deleter)
+{
```
EmilDohne, line 688: [nit] For such 'complex' initializations I always prefer writing them explicitly in the constructor body to aid in readability

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -685,38 +737,47 @@ ImageBuf::operator=(ImageBuf&& src)
...
+               ? m_spec.image_bytes()
+               : 0;
+    if (m_allocated_size != size) {
+        if (m_allocated_size)
+            free_pixels();
+        try {
+            m_pixels.reset(size ? new char[size] : nullptr);
+            // Set m_bufspan to the allocated memory
+            set_bufspan(m_pixels.get());
+        } catch (const std::exception& e) {
```
EmilDohne, line 772: [consider] If this is truly meant to only catch out of memory exceptions this should explicitly catch `std::bad_alloc` to still allow other exceptions to gracefully make their way up the stack

src/libOpenImageIO/imagebuf.cpp
```diff
@@ -685,38 +737,47 @@ ImageBuf::operator=(ImageBuf&& src)
...
+                           e.what());
+            error("ImageBuf unable to allocate {} bytes ({})\n", size,
+                  e.what());
+            size      = 0;
+            m_bufspan = {};
+        }
+        m_allocated_size = size;
+        pvt::IB_local_mem_current += m_allocated_size;
+        atomic_max(pvt::IB_local_mem_peak,
+                   (long long)pvt::IB_local_mem_current);
```
EmilDohne: [nit] `static_cast<long long>()`

EmilDohne (commented):
Overall think this is the right direction. Is it perhaps possible to add some unittests to the new functions as well?

src/include/OpenImageIO/image_span.h
lgritz: Do you think it's enough to have an enable_if so that this helper function cannot be accidentally used to turn a `image_span<const T>` into an `image_span<byte>`?

lgritz: Or just change the implementation to 
```
    return image_span<std::byte>(
        reinterpret_cast<std::byte*>(src.data())),
        ...
```
by removing the const_cast, it make it an error to try to instantiate the template if it was a const image_span passed in?

src/libOpenImageIO/imagebuf.cpp
lgritz: ok

src/libOpenImageIO/imagebuf.cpp
lgritz: We try not to let exceptions propagate, since nothing called in OIIO's public APIs is supposed to throw. And I'm not sure what else it would be here? But I'm being over-inclusive.

src/libOpenImageIO/imagebuf.cpp
lgritz: ok

src/libOpenImageIO/imagebuf.cpp
lgritz: Since this has been working for a while, I think I'll save that change (which is a good suggestion) for a follow-on PR.

src/include/OpenImageIO/image_span.h
lgritz: I see now, I introduced that change with this PR. I'm reverting it, and putting the special "force it non-const" to be private in the one module where I needed it internally.

lgritz:
I think, looking at it again, that this doesn't fully image_span-ify ImageBuf. It just does the constructors, reset, and get_pixels. Turns out there is still set_pixels left to do, but I'll do that in another PR.

EmilDohne (approved):
Thanks for getting around to these so fast!

PR #4817, opened 2025-06-29
testing: remove old test reference output we no longer need

lgritz:
Delete some reference test output for older toolchains and dependency versions that seem no longer needed.


jessey-git (approved):
LGTM

PR #4818, opened 2025-06-29
fix(oiiotool): Allow thread control for --parallel-frames

### LINKED ISSUE #4811: [BUG] Crash when using --parallel-frames and --threads together
**Describe the bug**

My intention was to use `oiiotool` to convert a sequence of linear ap0 EXR frames into PNGs. I wanted to restrict how many threads are used during the process but still let it work on the sequence in parallel. Thus the command line looked something like this:
```
oiiotool --parallel-frames -i "linear.%04d.exr" --iscolorspace "ACES2065-1" --ociodisplay "Rec.1886 Rec.709 - Display" "ACES 1.0 - SDR Video" --threads 3 -o "rec709.%04d.png"
```
The addition of the `--threads` argument seems to be crashing the process. I expected that when I set `--threads` to 3, only 3 frames would be read in parallel as there would be only 3 threads available. However, the application seems to ignore this and tries to load 96 frames in parallel due to my machine having 96 cores available. The crash is almost always silent. I was able to see the following message once:
```
oiiotool(33955,0x309d24000) malloc: Incorrect checksum for freed object 0x7f94d2798e08: probably modified after being freed.
Corrupt value: 0x14
oiiotool(33955,0x309d24000) malloc: *** set a breakpoint in malloc_error_break to debug
```

The reason I was going down this path is that I have a memory budget that gets exceeded if I try to run this on a production machine, which has 96 cores. Using `--parallel-frames` starts loading 96 4k images at once, which doesn't fit in the available memory. To make this work, I need to be able to either:
* Specify a memory limit so that `oiiotool` can figure out dynamically when to stop loading new images.
* Restrict the number of images loaded in parallel statically. Possibly via an addition to the `--parallel-frames` argument like `--parallel-frames 3`

I was hoping that `--threads` would do so but that crashes the application.

Installed via `pip install OpenImageIO==3.0.4.0` in both linux and macos:

```
OIIO 3.0.4.0 | MacOS/x86_64
Dependencies: DCMTK NONE, expat 2.5.0, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF NONE, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.20, libjpeg-turbo 3.0.4, LibRaw NONE, libuhdr 1.2.0, minizip-ng 4.0.7, OpenColorIO 2.4.1, OpenCV
    NONE, OpenEXR 3.2.4, OpenGL, OpenJPEG NONE, PNG 1.6.44, Ptex NONE, Ptex NONE, pybind11 2.12.0, pystring 1.1.4, Python3 3.10.11, Qt5 NONE, Qt6 NONE, Robinmap 1.3.0, TBB NONE, TIFF 4.0.9, WebP 1.4.0, yaml-cpp 0.8.0, ZLIB 1.2.12, ZLIB 1.2.12
```
```
OIIO 3.0.4.0 | Linux/x86_64
    Build compiler: gcc 12.2 | C++17/201703
Dependencies: DCMTK NONE, expat 2.6.3, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF NONE, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.20, Libheif NONE, libjpeg-turbo 3.0.4, LibRaw NONE, libuhdr 1.2.0, minizip-ng 4.0.7, OpenColorIO 2.4.1, OpenCV NONE, OpenEXR 3.2.4, OpenGL, OpenJPEG NONE, PNG 1.6.44, Ptex NONE, Ptex NONE, pybind11
    2.13.6, pystring 1.1.4, Python3 3.11.10, Qt5 NONE, Qt6 NONE, Robinmap 1.3.0, TBB NONE, TIFF 4.6.0, WebP 1.4.0, yaml-cpp 0.8.0, ZLIB 1.2.11, ZLIB 1.2.11
```

**To Reproduce**

Find a bunch of EXR frames and execute the following:
```
oiiotool --parallel-frames --threads 3 -i "linear.%04d.exr" --iscolorspace "ACES2065-1" --ociodisplay "Rec.1886 Rec.709 - Display" "ACES 1.0 - SDR Video"  -o "rec709.%04d.png"
```

lgritz:
Fixes #4811

This patch makes oiiotool's initial cursory scan of the command line notice a `--threads` and carfully honor it when parallelizing the frame loop.

Explanatory background info:

The cited issue shows something subtle happening with oiiotool's `--parallel-frames` argument. The symptom was that it was using too many threads and running out of memory, and seemed impervious to the use of `--threads` to properly control it.

The reason is that oiiotool first scans the command line args in a cursory way looking for signs of a "sequence", like a `--frames` or any filename containing frame wildcards (like `linear.%04d.exr`) and also noting if there is a `--parallel-frames`, but not actually taking any actions for any of the commands yet.

Having seen that there are frame wildcards, what it then does is execute the whole command line, once for each frame number that matched the pattern. And since it noted that a `--parallel-frames` argument was present somewhere, it does it with a parallel for loop. But how many threads does it use? Well, enough for all the cores, since in some sense it has not yet noticed the `--threads` argument. Moving the `--threads` to be placed prior to the `--parallel-frames` doesn't help, either.

The suggested workaround was to set the default number of threads with the environment variable `OPENIMAGEIO_THREADS` prior to launching oiiotool. But this patch is a more robust solution.

### DIFF
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -28,6 +28,7 @@
 #include <OpenImageIO/deepdata.h>
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/filter.h>
+#include <OpenImageIO/fmath.h>
 #include <OpenImageIO/imagebuf.h>
 #include <OpenImageIO/imagebufalgo.h>
 #include <OpenImageIO/imagebufalgo_util.h>
@@ -616,9 +617,11 @@ Oiiotool::extract_options(string_view command)
 
 // --threads
 static void
-set_threads(Oiiotool&, cspan<const char*> argv)
+set_threads(Oiiotool& ot, cspan<const char*> argv)
 {
     OIIO_DASSERT(argv.size() == 2);
+    if (ot.in_parallel_frame_loop())
+        return;
     int nthreads = Strutil::stoi(argv[1]);
     OIIO::attribute("threads", nthreads);
     OIIO::attribute("exr_threads", nthreads);
@@ -7228,9 +7231,10 @@ one_sequence_iteration(Oiiotool& otmain, size_t i, int frame_number,
     }
 
     Oiiotool otit;  // Oiiotool for this iteration
-    otit.imagecache      = otmain.imagecache;
-    otit.frame_number    = frame_number;
-    otit.parent_oiiotool = &otmain;
+    otit.imagecache               = otmain.imagecache;
+    otit.frame_number             = frame_number;
+    otit.parent_oiiotool          = &otmain;
+    otit.m_in_parallel_frame_loop = otmain.in_parallel_frame_loop();
     otit.getargs((int)seq_argv.size(), (char**)&seq_argv[0]);
 
     if (otit.ap.aborted()) {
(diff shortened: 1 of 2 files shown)


lgritz:
Thoughts, anyone?

EmilDohne (approved):
Lgtm!

PR #4819, opened 2025-07-01
ci: Bump 'latest' test versions

lgritz:
Some actual bumps in CI, others just bringing the INSTALL documentation up to date with what we have been testing and know works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -412,12 +412,12 @@ jobs:
             cxx_std: 20
             fmt_ver: 11.2.0
             opencolorio_ver: v2.4.2
-            openexr_ver: v3.3.3
+            openexr_ver: v3.3.4
             pybind11_ver: v2.13.6
             python_ver: "3.12"
             simd: avx2,f16c
-            setenvs: export LIBJPEGTURBO_VERSION=3.1.0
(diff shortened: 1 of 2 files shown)


jessey-git (approved):
I think these changes are all true.

Is there a quick way to verify? I looked through both the workflow ci.yml file as well as the actual build output for some of the libraries to determine what's actually being used. Is there an easier way?

lgritz:
The changes to ci.yml is me actually raising the versions that are downloaded and built. A package version is only listed in ci.yml in order to trigger a build of that specific version, it's not merely documentation.

The changes to INSTALL.md that don't directly correspond to those is me looking at our CI build logs which print a summary of the version if found of every dependency -- mostly checking on the MacOS jobs, since they get most dependencies from Homebrew, which is pretty good about staying on the latest version of everything.

lgritz:
It would be neato to have some automated process that could scan the build logs of ALL the CI jobs we have, and print a list of each dependency's min/max that we actually test in practice. 

I bet something like that could be implemented as a GHA workflow that runs nightly or weekly -- find the latest nightly CI run from main, download the logs for each job, and parse them to find the min/max for each dependency and output that into its log for us to check against INSTALL.md.

PR #4820, opened 2025-07-01
build: wheel upload_pypi step should only run from main repo

lgritz:
The upload_pypi step only happened for tags, not for other events like PRs. But it would TRY (and fail) if you pushed a v3.0.* tag to your own fork. This patch makes it not een attempt it, except when the tag is pushed to the offical repo.


zachlewis (approved):
Nice 🤘

PR #4822, opened 2025-07-02
Fix HEIF stride type mismatch in input/output classes

chenrui333:
seeing build failures when building against libheif 1.19+

```
  /private/tmp/openimageio-20250701-18561-96sl0/OpenImageIO-3.0.8.0/src/heif.imageio/heifinput.cpp:390:37: error: no matching member function for call to 'get_plane'
    390 |     const uint8_t* hdata = m_himage.get_plane(heif_channel_interleaved,
        |                            ~~~~~~~~~^~~~~~~~~
  /opt/homebrew/include/libheif/heif_cxx.h:909:32: note: candidate function not viable: no known conversion from 'int *' to 'size_t *' (aka 'unsigned long *') for 2nd argument
    909 |   inline const uint8_t* Image::get_plane(enum heif_channel channel, size_t* out_stride) const noexcept
        |                                ^                                    ~~~~~~~~~~~~~~~~~~
  /opt/homebrew/include/libheif/heif_cxx.h:914:26: note: candidate function not viable: no known conversion from 'int *' to 'size_t *' (aka 'unsigned long *') for 2nd argument
    914 |   inline uint8_t* Image::get_plane(enum heif_channel channel, size_t* out_stride) noexcept
        |                          ^                                    ~~~~~~~~~~~~~~~~~~
  1 error generated.
```

relates to https://github.com/Homebrew/homebrew-core/pull/228740

### DIFF
--- a/src/heif.imageio/heifinput.cpp
+++ b/src/heif.imageio/heifinput.cpp
@@ -385,8 +385,11 @@ HeifInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
         return false;
     if (y < 0 || y >= m_spec.height)  // out of range scanline
         return false;
-
-    int ystride          = 0;
+#if LIBHEIF_NUMERIC_VERSION >= MAKE_LIBHEIF_VERSION(1, 20, 0, 0)
+    size_t ystride = 0;
+#else
+    int ystride = 0;
+#endif
     const uint8_t* hdata = m_himage.get_plane(heif_channel_interleaved,
                                               &ystride);
     if (!hdata) {
--- a/src/heif.imageio/heifoutput.cpp
+++ b/src/heif.imageio/heifoutput.cpp
@@ -149,7 +149,7 @@ bool
 HeifOutput::write_scanline(int y, int /*z*/, TypeDesc format, const void* data,
                            stride_t xstride)
 {
-    data           = to_native_scanline(format, data, xstride, scratch);
+    data = to_native_scanline(format, data, xstride, scratch);
 #if LIBHEIF_NUMERIC_VERSION >= MAKE_LIBHEIF_VERSION(1, 20, 0, 0)
     size_t hystride = 0;
 #else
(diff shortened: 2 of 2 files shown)


lgritz:
We can see the [declaration from 1.19](https://github.com/strukturag/libheif/blob/v1.19.8/libheif/api/libheif/heif_cxx.h#L348) and compare it to the [declaration from 1.20](https://github.com/strukturag/libheif/blob/v1.20.0/libheif/api/libheif/heif_cxx.h#L348) and see that the parameter to that function changed from `int*` to `size_t*`.

Your diagnosis of the problem seems correct, but I fear that your fix will make it work with 1.20 but break it when encountering any older version.

I think that the correct implementation is going to need to look like

    #if LIBHEIF_NUMERIC_VERSION >= MAKE_LIBHEIF_VERSION(1, 20, 0, 0)
        ... new code ...
    #else
        ... old code ...
    #endif

chenrui333:
yeah, make sense, let me update the patch.

lgritz:
I think the `MAKE_LIBHEIF_VERSION` macro is in heifinput.cpp but not in heifoutput.cpp. It will need to be reproduced there.

lgritz:
I took the liberty of making the fix I think is needed and pushing it to your tree. If I can get it to pass CI, I will then merge. 

Homebrew upgraded their libheif to 1.20, so this getting this fix out is absolutely necessary ASAP to avoid a broken builds for lots of people (including on my laptop).

darix:
can the same patch be applied to the 2.x branch?

lgritz:
> can the same patch be applied to the 2.x branch?

I will try. The trouble is that it's been months since we've maintained that branch, so while the cherry-pick of this PR is easy, getting the CI to pass may be a weird hunt for all the fixes we've put in over the last 6 months to keep up with subtly changing GitHub runners and dependencies.

darix:
well i can do the cherry picking also in our rpm without you having to do it upstream.

package prepared here:

https://build.opensuse.org/package/show/home:darix:branches:graphics/OpenImageIO

lgritz:
Backported and released as v2.5.19.0.

But for the record, the 2.5 series is considered obsolete, and only gets occasional and extremely critical fixes. We recommend upgrading to the supported 3.0 release series if at all possible.

darix:
As soon as Blender 4.5 is out we can hopefully drop that version :)

*package update noises* Thank you for your help!

PR #4827, opened 2025-07-12
api: image_span-ify ImageBuf::set_pixels()

lgritz:


lgritz:
@EmilDohne your review would be appreciated, as you have been following this issue closely.

EmilDohne (approved):
LGTM! Only slight nitpick would be to add the oiio nodiscard to the function :)

PR #4828, opened 2025-07-12
ci: Bump 'latest releases' tests to use pybind11 3.0.0

lgritz:
Also found one stray test that was using a pybind11 version older than we claim to use, so bump that up to the advertised minimum.


lgritz:
This seems like a no-brainer, it's just ensuring that we test the newest pybind11 in CI and documenting that we support it.

Because it doesn't change code, I will merge within a couple days even if not reviewed.

PR #4830, opened 2025-07-13
build: Raise minimum supported Python from 3.7 to 3.9

lgritz:
Remove the CI jobs that tested VFXPlatform 2021 era, which we only were still doing for the sake of python 3.7.
Also stop building Python wheels for any python versions older than 3.9 (we never built 3.7 wheels, but we did build 3.8).

This is just in main, a.k.a. OIIO 3.1, to be released in fall 2025 and expected to be the main release series people are using in 2026.

VFX Platform has been 3.9+ since 2022, that's more than the "3 years back" rule dictates. And in the real world, it hasn't been "current" since 2020.

Just to be extra safe, I'm documenting 3.9 as the minimum and removing all 3.7 tests, but I'm not having the build affirmatively reject 3.7 for now, just in case somebody needs it. It should continue to work if someone tries it, with no guarantees or support. But probably by the time we do a 3.1 release in September, we will make it actually reject 3.7.


lgritz:
I don't understand why the wheel test is failing now, looks like the stubs are no longer matching, bud I didn't change the bindings. @chadrik @zachlewis any insight? Am I missing some subtlety where merely removing the Python 3.8 wheel generations would cause the stubs to fail?

chadrik:
It looks like a new version of pybind11 was releases and there's nothing in the cibuildwheel process to restrict it. 

Here's my original PR build process indicating `pybind11 2.13.6`: https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/15009156441/job/42174548321?pr=4692#step:4:1722

And the new build indicating `pyblind11 3.0.0`: https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/16244058754/job/45864325116?pr=4830#step:4:1713

I whipped up this diagram to indicate the dependency chain for generating the stubs:

```mermaid
flowchart TD
 subgraph s1["key"]
        n9["TOOL"]
        n10(["PRODUCT"])
        n11["1ST PARTY"]
  end
    A["src/python/generate_stubs.py"] -- produces --> B(["stubs"])
    n2["stubgenlib"] --> A
    n1["mypy"] --> n2
    n4["src/**"] -- produces --> n3(["OpenColorIO <br>binding docstrings"])
    n5["pybind11"] --> n4
    n3 --> A

    n11@{ shape: rounded}
    A@{ shape: rounded}
    n2@{ shape: rect}
    n1@{ shape: rect}
    n4@{ shape: rounded}
    n5@{ shape: rect}
    style n9 stroke:#FFD600
    style n10 stroke:#00C853
    style n11 stroke:#2962FF
    style A stroke:#2962FF
    style B stroke:#00C853
    style n2 stroke:#FFD600
    style n1 stroke:#FFD600
    style n4 stroke:#2962FF
    style n3 stroke:#00C853
    style n5 stroke:#FFD600
```

The versions of mypy and stubgenllib are locked down, and we're in control of the source code, so my primary suspect was pybind11 and it turns out to have been a good hunch.

chadrik:
This should fix it:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 4507cfa0b..0ad5397ec 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ oiiotool = "OpenImageIO:_command_line"
 build-backend = "scikit_build_core.build"
 requires = [
     "scikit-build-core>=0.10.6,<1",
-    "pybind11",
+    "pybind11>2.13,<3",
 ]
 
 [tool.scikit-build]
```

chadrik:
On second thought, since we're not saving a lock file to the repo I recommend using fully qualified versions for both of these build requirements in order to get deterministic builds.  Unlike runtime deps like numpy, there's little advantage to leaving the build system versions as a range.

lgritz:
Oh wow, so it wasn't really connected to this PR at all!

So, would you like me to apply your suggestion, but maybe just say `pybind11=2.13` and lock it down?

lgritz:
This PR is 3 weeks old. We need to probably merge by the end of next week because of the looming release deadlines. Are there any opinions anybody would like to voice that we should keep the current minimum python 3.7 through the life of OIIO 3.1?

chadrik:
No concerns here.

lgritz:
In addition to 3 weeks in review with no objections, I brought this up at the last couple OIIO and OSL TSC meetings and everyone present was in agreement to move this forward, so I'm going to merge.

PR #4831, opened 2025-07-13
build: For python stub generation, lock pybind11

lgritz:
pybind11 3.0 seems to break this process, so lock down to no later than the last version before it broke.


lgritz:
@chadrik

lgritz:
I got an offline approval from @chadrik on this patch, so merging now in order to fix our currently broken wheel CI.

PR #4832, opened 2025-07-14
perf(exr): Speed up OpenEXR non-core header read time

lgritz:
Recent OpenEXR internals changes make header reads via the C++ library dramatically slower if a Imf::IStream subclass does not implement the size() and stateless-read method.

See https://github.com/AcademySoftwareFoundation/openexr/issues/1984

On my laptop, when forcing OIIO to avoid the "core" library and use the C++ API to OpenEXR, this speeds up the reading of small exr files by about 6x.

These changes were introduced with OpenEXR 3.3 and we never changed our side.

(Also opportunistically added a couple DASSERT's to the existing functions.)

### DIFF
--- a/src/openexr.imageio/exr_pvt.h
+++ b/src/openexr.imageio/exr_pvt.h
@@ -77,13 +77,38 @@ public:
             throw Iex::IoExc("Unexpected end of file.");
         return n;
     }
-    uint64_t tellg() override { return m_io->tell(); }
+
+    uint64_t tellg() override
+    {
+        OIIO_DASSERT(m_io);
+        return m_io->tell();
+    }
+
     void seekg(uint64_t pos) override
     {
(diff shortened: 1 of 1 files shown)


lgritz:
I can see that this closes a big gap between the core and C++ library performance when reading the openexr file headers. Prior to this change, using the core library was 5-6x faster. With this change, they are approximately the same speed.

lgritz:
@kdt3rd Does this seem right to you?

lgritz:
Anybody have comment or objections about this one? It's been outstanding for a week.

ThiagoIze:
It looks reasonable to me, but I am not well-versed in this code. 
While this doesn't fix the openexr slowdown for my use case, it at least fixes it for some cases, which is great.

lgritz:
Kimball contacted me offline and said it looked right to him also.

PR #4835, opened 2025-07-20
build: PNG auto-build improvements

lgritz:
The main thing is no longer using a custom namespace prefix for auto-built libpng. We're already building it statically by default and hiding the symbols, so the custom prefix only serves to make occasional build problems by having different symbols than typically found in the library. We still keep it as an option that can be enabled.

Also, be a little more careful in clearing unneeded variable that may have been set by an earier attempt to find the package but that failed to turn up a version we were happy with.


lgritz:
A quick review would be appreciated. This appears to be a necessary step to addressing some broken OSL CI, for reasons that are actually tricky to explain, I think maybe they changed something on the GHA Mac runners.

src/cmake/build_PNG.cmake
```diff
@@ -36,17 +43,20 @@ build_dependency_with_cmake (PNG
...
-
 set (PNG_REFIND_ARGS EXACT REQUIRED)
-set (PNG_DIR ${PNG_LOCAL_INSTALL_DIR}/lib/cmake/PNG)
 set (PNG_FIND_VERSION_EXACT ON)
+unset (PNG_LIBRARY)
+unset (PNG_LIBRARY_RELEASE)
+unset (PNG_LIBRARY_DEBUG)
+unset (PNG_INCLUDE_DIRS)
+unset (PNG_INCLUDE_DIR)
+unset (PNG_PNG_INCLUDE_DIR)
```
jessey-git, line 62: PNG_PNG... - Is this line correct?  Also, the section above which does `unset` has a few extra variables like PNG_VERSION_STRING - do we want to unset those here too?

lgritz: God help us all, it is correct, and comes from CMake:
https://github.com/Kitware/CMake/blob/master/Modules/FindPNG.cmake#L91

lgritz: I just pushed an amendment that harmonizes the two lists.

lgritz:
This is needed to unbreak CI for OSL, so I will merge this tomorrow morning unless there is a specific objection.

jessey-git (approved):
Looks good.

PR #4836, opened 2025-07-20
fix(build): update jpeg2000input.cpp to include cstdarg

1div0:
Add missing C++ header file

On the F41 and F42 compilation using LLVM based Clang++ failed.

[ 23%] Building CXX object src/libOpenImageIO/CMakeFiles/OpenImageIO.dir/__/jpeg2000.imageio/jpeg2000input.cpp.o /usr/src/github.com/1div0/OpenImageIO/src/jpeg2000.imageio/jpeg2000input.cpp:462:9: error: use of undeclared identifier 'va_start'
  462 |         va_start(args, fmt);
      |         ^
/usr/src/github.com/1div0/OpenImageIO/src/jpeg2000.imageio/jpeg2000input.cpp:464:9: error: use of undeclared identifier 'va_end'
  464 |         va_end(args);
      |         ^
2 errors generated.

## Checklist:

- [ ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/jpeg2000.imageio/jpeg2000input.cpp
+++ b/src/jpeg2000.imageio/jpeg2000input.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/AcademySoftwareFoundation/OpenImageIO
 
+#include <cstdarg>
 #include <cstdio>
 #include <vector>
 


1div0:
I am wondering why this was not catched earlier by the CI.

1div0:
`clang++ -v
clang version 20.1.8 (Fedora 20.1.8-1.fc42)
`

Is that too picky?

lgritz:
> I am wondering why this was not catched earlier by the CI.

It might just be that none of our 25 or so CI jobs happen to hit the specific combination of Linux and clang 20 and have openjpeg available? 

It's very hard to ensure that CI tests every compiler version on every OS and hardware platform, with every version combination and availability of all dependencies. I mean, I guess we could do it, but then a CI run would have hundreds of jobs and probably take days to run. So we have to compromise and "sample" the combination space, hoping that across all the platforms, we test each important version of each tool or dependency at least once, but not necessarily all possible combinations together. Every once in a while, especially with a new compiler (like clang 20.1), we hit an untested combo that is a problem, and in that case, we just try to get it fixed as soon as we can. As you did!

PR #4837, opened 2025-07-23
ffmpeg: Replace deprecated and soon removed avcodec_close with avcodec_free_context in ffmpeg plugin.

ssh4net:
avcodec_close (AVCodecContext *avctx) removing in ffmpeg 7.2 and was deprecated for couple years

## Description

Global avcodec_close (AVCodecContext *avctx)
Do not use this function. Use avcodec_free_context() to destroy a codec context (either open or closed). Opening and closing a codec context multiple times is not supported anymore – use multiple codec contexts instead.

In ffmpeg master, it has already been removed. But still work in v7.1

## Checklist:

- [x ] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -581,7 +581,7 @@ bool
 FFmpegInput::close(void)
 {
     if (m_codec_context)
-        avcodec_close(m_codec_context);
+        avcodec_free_context(&m_codec_context);
     if (m_format_context) {
         avformat_close_input(&m_format_context);
         av_free(m_format_context);  // will free m_codec and m_codec_context


lgritz:
How far back is `avcodec_free_context` available? Does it go all the way back to ffmpeg 4.0, the earliest version we claim to support? (If not, then we either have to make our advertised minimum ffmpeg new enough that this is supported, or else use an `#if` based on the version.)

ssh4net:
Not sure about version, but there is a changelog 
```
2014-05-18 - 68c0518 / fd05602 - lavc 55.63.100 / 55.52.0 - avcodec.h
  Add avcodec_free_context(). From now on it should be used for freeing
  AVCodecContext.
```

lgritz:
Our oldest is ffmpeg 4.0, which corresponds to 58.18.100, so I think we are safe.

PR #4838, opened 2025-07-24
api: span-ify attribute/getattribute everywhere

lgritz:
* span-ify ParamValue (including ctr and init)
* add span versions of attribute() and getattribute() to the classes that have them: ParamValueList, ImageSpec, ImageCache, TextureSystem.

Much like the previous span/image_span additions we've made elsewhere, now each of these calls has 3 flavors:
1. The previously existing -- but acknowledged "unsafe" -- version that takes a TypeDesc and a raw pointer (which had better match the TypeDesc in type and size, but who knows?)
2. A version that takes a TypeDesc and a `span<std::byte>` delineating the raw memory to use, that must match the size implied by the type, and is implemented underneath (for now) by calling the pointer version after checking that the size is adequate.
3. A template version that takes a TypeDesc and a `span<T>` that must match both in total size and appropriate type and asserts in debug builds if they don't. It's implemented by calling the raw byte version after the type check.

The third variety is the preferred one for apps to call, whenever possible, but the others are still exposed for special cases.


lgritz:
@EmilDohne 👀 🙏

EmilDohne:
I'll take a look this weekend!

lgritz:
I'd love to get this merged by the end of the week, if nobody has any objections.

src/include/OpenImageIO/imagecache.h
```diff
@@ -315,18 +315,20 @@ class OIIO_API ImageCache {
...
     ///
     ///     ImageCache *ic;
     ///     ...
     ///     int maxfiles = 50;
-    ///     ic->attribute ("max_open_files", TypeDesc::INT, &maxfiles);
+    ///     ic->attribute ("max_open_files", TypeInt, make_cspan(maxfiles));
     ///
     ///     const char *path = "/my/path";
-    ///     ic->attribute ("searchpath", TypeDesc::STRING, &path);
+    ///     ic->attribute ("searchpath", TypeString, make_cspan(&path, 1));
```
EmilDohne, line 331: [question] Should we instead make this docstring as follows?:

```cpp
///     const std::string = "/my/path";
///     ic->attribute ("searchpath", TypeString, make_cspan(path.c_str, path.size()));
```

src/include/OpenImageIO/imageio.h
```diff
@@ -609,10 +634,58 @@ class OIIO_API ImageSpec {
...
+    ///     spec.getattribute ("compression", TypeString, make_span(compression));
+    ///
+    ///     // Alternately, retrieving a string with a ustring:
+    ///     ustring compression;
+    ///     spec.getattribute ("compression", TypeString, make_span(compression));
+    ///
+    /// Note that when retrieving a string, you need to pass a span of either
+    /// `const char*` or `ustring`, not a pointer to the first character of
+    /// the string.  The caller does not need to ever free the memory that
+    /// contains the characters.
```
EmilDohne: [nit] expand on who owns the string, whether that may be the `ImageSpec` or a `ustring`

src/libutil/paramlist.cpp
```diff
@@ -37,19 +37,20 @@ ParamValue::init_noclear(ustring _name, TypeDesc _type, int _nvalues,
...
             else
                 memset(&m_data, 0, sizeof(m_data));
             m_copy     = false;
             m_nonlocal = false;
         } else {
             void* ptr = malloc(size);
-            if (_value)
-                memcpy(ptr, _value, size);  //NOSONAR
+            if (_value.size())
+                memcpy(ptr, _value.data(), size);  //NOSONAR
```
EmilDohne: [consider] since `_value.size()` can either be == size or == 0 as asserted above we should probably just do `malloc(_value.size())` and then `memcpy(ptr, _value.data(), _value.size())`

EmilDohne (changes requested):

EmilDohne:
> I'll take a look this weekend!

Sorry I only got around to it now, except for those things this LGTM

src/include/OpenImageIO/imagecache.h
lgritz: No, this is correct as it stands, making a span containing one `char*` (one string).
The way you wrote it, you have a span containing n characters.

EmilDohne: Right, I see, I missed the `&path`

src/libutil/paramlist.cpp
lgritz: You mean just to eliminate the "if"?

src/include/OpenImageIO/imageio.h
lgritz: > [nit] expand on who owns the string, whether that may be the `ImageSpec` or a `ustring`

Amended

src/libutil/paramlist.cpp
EmilDohne: I meant more something along the lines of:

```cpp
void* ptr = malloc(_value.size());
if (_value.size())
   memcpy(ptr, _value.data(), _value.size());
```

and then be rid of the `//NOSONAR` since I'm assuming it triggered on copying from a buffer that may be smaller than `size`

lgritz: OH, I see your point now. Yes, I agree. Will fix.

lgritz: So amended.

lgritz:
Seem all set to you know, @EmilDohne?

PR #4839, opened 2025-07-25
fix(ImageInput): incorrect IOProxy logic related to valid_file

lgritz:
Inside ImageInput::create(), there's a section where the created ImageInput tries to do a quick call to its valid_file() method to do an inexpensive check that the file appears to be the right file format corresponding to the ImageInput subclass. However, it always called the variety of valid_file() that takes a string filename, ignoring the fact that create() itself may have been passed an IOProxy, which should always take precedence over trying to open a real file.

### DIFF
--- a/src/libOpenImageIO/imageioplugin.cpp
+++ b/src/libOpenImageIO/imageioplugin.cpp
@@ -630,6 +630,13 @@ ImageInput::create(string_view filename, bool do_open, const ImageSpec* config,
         format = filename;
     }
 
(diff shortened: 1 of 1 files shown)


lgritz:
Any comments on this?

lgritz:
I will merge this shortly if there are no objections.

lgritz:
I got a LGTM offline from somebody who tried it and confirmed the fix, but isn't allowed to comment publicly. So merging.

Issue #4791, opened 2025-06-04
[FEATURE REQUEST] ACES Container compliant writer


antond-weta:
It would be nice to have a mode in the ExrWriter enforcing compatibility with the ACES Container standard (ST 2065-4).

The ACES Container compliant EXR must have these limitations (I may have missed some):
- The 'acesImageContainerFlag' attribute present in the metadata
- The chromaticities are set to ACES AP0
- Only RGB, BGRA channels allowed (either mono or stereo)
- The data type must be half float
- No compression

The way I envision this implemented:

Pass a hint to the ExrWriter, something like 'ACESContainer' with one of the three values:
- "none" (default) - restrict enabling the 'acesImageContainerFlag' attribute, no other changes.
- "strict" - check that the image and writer configuration are compliant, error otherwise. Should automatically set the chromaticities to ACES AP0, and enable the 'acesImageContainerFlag'.
- "relaxed" - check that the image and writer configuration are compliant, issue a warning otherwise. Allow non-compliant settings, like enabling compression, storing extra channels, writing full float data. Should automatically set the chromaticities to ACES AP0. Should enable the 'acesImageContainerFlag' ONLY if the file is fully compliant.

Alternatively to erroring in the 'strict' mode, we could enforce compliance automatically where possible, like overriding the compression and pixel data type settings, discarding the extra channels, etc. I have no preference on this. Or perhaps we could have 2 separate modes.

zachlewis:
Totally agree with every part of this.

> Alternatively to erroring in the 'strict' mode, we could enforce compliance automatically where possible, like overriding the compression and pixel data type settings, discarding the extra channels, etc. I have no preference on this. Or perhaps we could have 2 separate modes.

I like the idea of "strict" mode forcing compliance, I think that strikes a nice balance with "relaxed" mode. Would it make sense to attempt to transform from `oiio:ColorSpace` to AP0 if we need to (in either mode)?

antond-weta:
> Would it make sense to attempt to transform from oiio:ColorSpace to AP0 if we need to (in either mode)

I see how that can be useful, but I personally wouldn't do that. The proposed changes are for the ImageOutput plugin only, I don't think it should mess with the pixels.
Perhaps when this code is being used in places like oiiotool, we could add extra checks and issue a warning, or auto convert there?

JGoldstone:
[...]

PR #4840, opened 2025-07-27
exr: Don't assume unlabeled OpenEXR files are lin_rec709

lgritz:
New logic for reading OpenEXR files:

1. If the new colorInteropID is present, that's the color space.
2. Use the OCIO file naming rules, if that indicates a color space.
3. Otherwise, leave it blank, don't mae any assumptions.

It was a mistake for 3.0 to assume that completely unlabeled openexr files are lin_rec709.

### DIFF
--- a/src/openexr.imageio/exr_pvt.h
+++ b/src/openexr.imageio/exr_pvt.h
@@ -6,6 +6,7 @@
 
 
 #include <OpenImageIO/Imath.h>
+#include <OpenImageIO/color.h>
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/imageio.h>
 #include <OpenImageIO/platform.h>
@@ -228,6 +229,7 @@ private:
     int m_nsubimages;                   ///< How many subimages are there?
     int m_miplevel;                     ///< What MIP level are we looking at?
     std::vector<float> m_missingcolor;  ///< Color for missing tile/scanline
+    std::string m_filename;             // filename, if known
 
     void init()
     {
@@ -243,6 +245,7 @@ private:
         m_io                       = nullptr;
         m_local_io.reset();
         m_missingcolor.clear();
+        m_filename.clear();
     }
 
     bool read_native_scanlines_individually(int subimage, int miplevel,
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -240,6 +240,8 @@ OpenEXRInput::open(const std::string& name, ImageSpec& newspec,
 
     // Check any other configuration hints
 
+    m_filename = name;
+
     // "missingcolor" gives fill color for missing scanlines or tiles.
     if (const ParamValue* m = config.find_attribute("oiio:missingcolor")) {
         if (m->type().basetype == TypeDesc::STRING) {
@@ -391,14 +393,6 @@ OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in,
         return false;
 
     spec.deep = Strutil::istarts_with(header->type(), "deep");
-
-    // Unless otherwise specified, exr files are assumed to be linear Rec709
-    // if the channels appear to be R, G, B.  I know this suspect, but I'm
-    // betting that this heuristic will guess the right thing that users want
-    // more often than if we pretending we have no idea what the color space
-    // is.
-    if (pvt::channels_are_rgb(spec))
-        spec.set_colorspace("lin_rec709");
 
     if (levelmode != Imf::ONE_LEVEL)
         spec.attribute("openexr:roundingmode", roundingmode);
@@ -673,6 +667,17 @@ OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in,
         spec.attribute("oiio:subimagename", header->name());
 
     spec.attribute("oiio:subimages", in->m_nsubimages);
+
+    // Try to figure out the color space
+    if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
+        spec.set_colorspace(c->get_ustring());
+    } else {
+        string_view cs
+            = ColorConfig::default_colorconfig().getColorSpaceFromFilepath(
+                in->m_filename, "unknown");
+        if (cs.size() && cs != "unknown")
+            spec.set_colorspace(cs);
+    }
 
     // Squash some problematic texture metadata if we suspect it's wrong
     pvt::check_texture_metadata_sanity(spec);
--- a/src/openexr.imageio/exrinput.cpp
+++ b/src/openexr.imageio/exrinput.cpp
@@ -668,15 +668,11 @@ OpenEXRInput::PartInfo::parse_header(OpenEXRInput* in,
 
     spec.attribute("oiio:subimages", in->m_nsubimages);
 
-    // Try to figure out the color space
-    if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
+    // Try to figure out the color space for some unambiguous cases
+    if (spec.get_int_attribute("acesImageContainerFlag") == 1) {
+        spec.set_colorspace("lin_ap0_scene");
+    } else if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
         spec.set_colorspace(c->get_ustring());
-    } else {
-        string_view cs
-            = ColorConfig::default_colorconfig().getColorSpaceFromFilepath(
-                in->m_filename, "unknown");
-        if (cs.size() && cs != "unknown")
-            spec.set_colorspace(cs);
     }
 
     // Squash some problematic texture metadata if we suspect it's wrong
--- a/src/openexr.imageio/exrinput_c.cpp
+++ b/src/openexr.imageio/exrinput_c.cpp
@@ -19,6 +19,7 @@
 #include <OpenEXR/openexr.h>
 
 #include "imageio_pvt.h"
+#include <OpenImageIO/color.h>
 #include <OpenImageIO/dassert.h>
 #include <OpenImageIO/deepdata.h>
 #include <OpenImageIO/filesystem.h>
@@ -208,6 +209,7 @@ private:
     std::unique_ptr<Filesystem::IOProxy> m_local_io;
     int m_nsubimages;                   ///< How many subimages are there?
     std::vector<float> m_missingcolor;  ///< Color for missing tile/scanline
+    std::string m_filename;             // filename, if known
 
     void init()
     {
@@ -216,6 +218,7 @@ private:
         m_userdata.m_io  = nullptr;
         m_local_io.reset();
         m_missingcolor.clear();
+        m_filename.clear();
     }
 
     bool valid_file(const std::string& filename, Filesystem::IOProxy* io) const;
@@ -357,6 +360,8 @@ OpenEXRCoreInput::open(const std::string& name, ImageSpec& newspec,
     //KDTDISABLE }
 
     // Check any other configuration hints
+
+    m_filename = name;
 
     // "missingcolor" gives fill color for missing scanlines or tiles.
     if (const ParamValue* m = config.find_attribute("oiio:missingcolor")) {
@@ -526,14 +531,6 @@ OpenEXRCoreInput::PartInfo::parse_header(OpenEXRCoreInput* in,
 
     spec.deep = (storage == EXR_STORAGE_DEEP_TILED
                  || storage == EXR_STORAGE_DEEP_SCANLINE);
-
-    // Unless otherwise specified, exr files are assumed to be linear Rec709
-    // if the channels appear to be R, G, B.  I know this suspect, but I'm
-    // betting that this heuristic will guess the right thing that users want
-    // more often than if we pretending we have no idea what the color space
-    // is.
-    if (pvt::channels_are_rgb(spec))
-        spec.set_colorspace("lin_rec709");
 
     if (levelmode != EXR_TILE_ONE_LEVEL)
         spec.attribute("openexr:roundingmode", (int)roundingmode);
@@ -775,6 +772,17 @@ OpenEXRCoreInput::PartInfo::parse_header(OpenEXRCoreInput* in,
     }
 
     spec.attribute("oiio:subimages", in->m_nsubimages);
+
+    // Try to figure out the color space
+    if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
+        spec.set_colorspace(c->get_ustring());
+    } else {
+        string_view cs
+            = ColorConfig::default_colorconfig().getColorSpaceFromFilepath(
+                in->m_filename, "unknown");
+        if (cs.size() && cs != "unknown")
+            spec.set_colorspace(cs);
+    }
 
     // Squash some problematic texture metadata if we suspect it's wrong
     pvt::check_texture_metadata_sanity(spec);
--- a/src/openexr.imageio/exrinput_c.cpp
+++ b/src/openexr.imageio/exrinput_c.cpp
@@ -773,15 +773,11 @@ OpenEXRCoreInput::PartInfo::parse_header(OpenEXRCoreInput* in,
 
     spec.attribute("oiio:subimages", in->m_nsubimages);
 
-    // Try to figure out the color space
-    if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
+    // Try to figure out the color space for some unambiguous cases
+    if (spec.get_int_attribute("acesImageContainerFlag") == 1) {
+        spec.set_colorspace("lin_ap0_scene");
+    } else if (auto c = spec.find_attribute("colorInteropID", TypeString)) {
         spec.set_colorspace(c->get_ustring());
-    } else {
-        string_view cs
-            = ColorConfig::default_colorconfig().getColorSpaceFromFilepath(
-                in->m_filename, "unknown");
-        if (cs.size() && cs != "unknown")
-            spec.set_colorspace(cs);
     }
 
     // Squash some problematic texture metadata if we suspect it's wrong
--- a/testsuite/dup-channels/ref/out.txt
+++ b/testsuite/dup-channels/ref/out.txt
@@ -6,7 +6,6 @@ out.exr              :   64 x   64, 6 channel, half openexr
     PixelAspectRatio: 1
     screenWindowCenter: 0, 0
     screenWindowWidth: 1
-    oiio:ColorSpace: "lin_rec709"
     oiio:subimages: 1
     openexr:lineOrder: "increasingY"
 Reading out2.exr
@@ -18,7 +17,6 @@ out2.exr             :   64 x   64, 6 channel, half openexr
     PixelAspectRatio: 1
     screenWindowCenter: 0, 0
     screenWindowWidth: 1
-    oiio:ColorSpace: "lin_rec709"
     oiio:subimagename: "Aimg"
     oiio:subimages: 1
     openexr:chunkCount: 4
--- a/testsuite/iinfo/ref/out-fmt6.txt
+++ b/testsuite/iinfo/ref/out-fmt6.txt
@@ -147,7 +147,6 @@ src/tiny-az.exr :     screenWindowCenter: 0, 0
 src/tiny-az.exr :     screenWindowWidth: 1
 src/tiny-az.exr :     Software: "OpenImageIO 2.5.0.0spi : oiiotool -pattern constant:color=0.25,0.5,0.75 4x4 3 --origin +2+2 --fullsize 20x20+1+1 -o tiny-az.exr"
 src/tiny-az.exr :     Exif:ImageHistory: "oiiotool -pattern constant:color=0.25,0.5,0.75 4x4 3 --origin +2+2 --fullsize 20x20+1+1 -o tiny-az.exr"
-src/tiny-az.exr :     oiio:ColorSpace: "lin_rec709"
 src/tiny-az.exr :     oiio:subimages: 1
 src/tiny-az.exr :     openexr:lineOrder: "increasingY"
 src/tiny-az.exr :     Stats Min: 0.250000 0.500000 0.750000 (float)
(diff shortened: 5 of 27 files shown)


zachlewis:
I think this is on the right track. 

I'm gonna talk about some stuff that's probably just a liiiittle bit outside the scope of not assuming unlabeled OpenEXRs are lin_rec709, but... hey, you're the one who brought up colorInteropIDs! 

(I'm not necessarily suggesting we do all of this in a single PR; but I want to get on the same page as you re: "the bigger picture", as far as divining color space names from colorInteropIDs is concerned)



Quoting from Doug Walker's WIP notes on [identifying the color space of OpenEXR files](https://docs.google.com/document/d/1MTH1bq2L67ifvdDf64Amhzg4AbkIM5LG6yPHrB96Vwo/edit?tab=t.0), my commentary unquoted inline:

> 1. If the acesImageContainer attribute is present, this takes precedence, consider the color space ACES2065-1. This should be handled the same as if the colorInteropID is present and set to "lin_ap0_scene".

 Likewise, we should set "oiio:ColorSpace" to the config's canonical name for the "aces_interchange" space. ("aces_interchange" is a required role for OCIO-2.2+ configs)). There's a chance OCIO-2.1 and earlier configs may not contain a linear AP0 color space, which... well, poses a problem for dealing with ACES2065-4 plates. OCIO does have some heuristics for identifying other recognized color spaces in a config, which might be used as a means for creating and adding to the config a "lin_ap0" color space; but the heuristics don't always work, and may not be worth the complexity of trying to solve this problem under the hood for users using OCIO configs that do not include a lin_ap0 space. Upshot, if users are trying to apply ColorSpace-aware IBAs on ACES2065-4 plates with a custom OCIO config that excludes an ACES2065-1 color space... I think it's fine to throw an error. 

> 2.  If the colorInteropID is "data", the file only contains non-color data and should not be color managed.

If "data" is indicated, we can just set "oiio:ColorSpace" to either the config's "data" role, if present, or recurse through the config's color spaces to find a color space whose "isdata" value is True; or, failing both of those, we add a "data" color space to the config in-memory, and use that. Any OCIO transform involving a "data" color space automatically becomes a NoOp in the Processor, so we don't really have to do much special handling inside downstream IBAs. That said, for any IBAs that do expect to change `oiio:ColorSpace`, we should probably ensure we actually are passing through the "data" color space, since we don't want to misinform IBAs further downstream...! The only way to "break out" of 'noop-data' mode is by manually setting `oiio:ColorSpace` to something else. 

> 3. If the colorInteropID is set to "unknown" or is not present, the application may use other mechanisms to assign a default color space. OpenColorIO provides "File Rules" for this purpose.

Pretty straightforward -- use `ColorConfig::getColorSpaceFromFilepath(...)` to set `oiio:ColorSpace`

> 4. Otherwise, attempt to use the colorInteropID with the color management system being used. Please see the [Color Interop ID recommendation](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE/edit?usp=sharing) for details on how to use the ID with OpenColorIO

The gist is:

  1. If `colorInteropID` contains a ":" (colon) -- i.e., if the value is _namespaced_.
     i.  Try searching the config for a color space that matches the full string ("{ns}:{name}")
     ii.   If that fails, split off the colon and everything before it from colorInteropID, and continue
  2. If `colorInteropID` matches the name or alias of a color space in the config, use it! 


> 5. If the color space identified by the colorInteropID cannot be used for some reason (e.g., it is not recognized by the user's OCIO config), then the application may fall back to some other mechanism of assigning a default, or ask the user to resolve the situation.

If we can't find an appropriate color space with just the user's OCIO config...
  1. Search an interchange reference config (`ocio://studio-config-latest` by default) for a color space that matches `colorInteropID`
  (Note -- we should implement a way for users to provide an optional "interchange config" that ostensibly contains color spaces / aliases for the colorInteropID we're trying to find -- e.g., a config provided by an external vendor with definitions for namespaced colorInteropID values).
  2. If not found, throw an exception (or, alternatively, treat as "unknown" and use getColorSpaceFromFilepath()...)
  3. Try to find an equivalent color space in the default config whose definition matches that of the color space identified by the `colorInteropID` in the interchange reference config -- OpenColorIO::Config::IdentifyBuitinColorSpace(...). In other words, use an "interchange reference" OCIO config known to provide matches for a given colorInteropID string to see if we can identify an definitionally-equivalent color space in the default ColorConfig to use as `oiio:ColorSpace`.
  4. If no definitionally equivalent color space can be identified in the default config, use `OpenColorIO::Config::GetProcessorFromConfigs` to attempt to create a transform from the "colorInteropID" color space found in the interchange reference config to either the default config's scene or display reference space -- if GetProcessorFromConfigs can ascertain a "bridge" between the two configs through a commonly shared color space, then use the resulting transform to define an ad-hoc color space named after the colorInteropID value, and the color space to the default config (in-memory); set `oiio:ColorSpace` = value of `colorInteropID`.

(@doug-walker, does all of this seem reasonable and in-line with your proposal?)


Also, I just wanted to draw your attention to a relevant bit from a previous PR of mine, pertaining to autocc stuff, just in case you weren't aware.

I'd adjusted `libOpenImageIO/color_ocio.cpp` to use `getColorSpaceFromFilepath`-style matching for `autocc`s; but to maintain OIIO-3.0's existing behavior and prevent surprises, instead of failing over to the config's Default rule / role like OCIO normally would, it fails over to the existing OIIO-1-style "parseColorSpaceFromString" mechanism that possibly returns empty strings (which turn into "lin_rec709"s elsewhere in the codebase):


check out color_ocio.cpp, lines 2079-2091:
```cpp
string_view
ColorConfig::getColorSpaceFromFilepath(string_view str) const
{
    if (getImpl() && getImpl()->config_) {
        std::string s(str);
        string_view r = getImpl()->config_->getColorSpaceFromFilepath(
            s.c_str());
        if (!getImpl()->config_->filepathOnlyMatchesDefaultRule(s.c_str()))
            return r;
    }
    // Fall back on parseColorSpaceFromString
    return parseColorSpaceFromString(str);  
}
```

So, I'd propose simplifying this to:

```cpp
string_view
ColorConfig::getColorSpaceFromFilepath(string_view str) const
{
    if (getImpl() && getImpl()->config_) {
        std::string s(str);
        string_view r = getImpl()->config_->getColorSpaceFromFilepath(
            s.c_str());
        return r;
    }
}
```
Which would cause OCIO to fail over to using the "default" color space, as designated by the config. 

This also means, if you shipped a default OIIO-flavored OCIO config, you could still have "lin_rec709" be the default failover (i.e., instead of "ACES2065-1", which is what the default would be with the standard Studio and CG ACES configs).

In this respect, it's not really a "mistake" to assume unlabeled EXRs are necessarily "lin_rec709" -- the mistake is in preventing the OCIO config from defining the default behavior, if that makes sense.






Sorry for the mouthful, Larry!

zachlewis:
> * Use the OCIO file naming rules, if that indicates a color space.
> 
> * Otherwise, leave it blank, don't mae any assumptions.

Also worth mentioning, OCIO's file naming rules will always indicate a color space.

lgritz:
> Also worth mentioning, OCIO's file naming rules will always indicate a color space.

Can you clarify? What does it return if there's no apparent color space name in the filename?

doug-walker:
> Can you clarify? What does it return if there's no apparent color space name in the filename?

@lgritz , All v2+ configs and most v1 configs define a default color space that is specified by the config author. Zach's proposed changes to OIIO's wrapper on getColorSpaceFromFilepath to return that rather than use parseColorSpaceFromString seem correct. Config authors may include the old parseColorSpaceFromString behavior in the File Rules, if they want, and this is already the default when using getColorSpaceFromFilepath with v1 configs. (Some users may not want this because it could create some unpredictable assignments based on random letter sequences in path names, so that is why it's not on by default in v2 configs.)  The part I'm not sure about is what config OIIO would use by default, if the user does not specify one. As Zach wrote, that seems like the correct place to impose a lin_rec709_scene type default and even the parseColorSpaceFromString usage.

@zachlewis , I think we're on the same page regarding what you wrote above. I will mention that we are going to add some helper functions to the OCIO API to simplify some of the steps.  And steps 3 and 4 would become more straight forward with the color space merge functionality.

It seems like one risk of assigning a lin_rec709_scene default is that this will get preserved through a sequence of operations and then written back out as if it is a known quantity instead of a guess. And that's how metadata becomes unreliable. So I can understand why the code fragment above is actually not using what the OCIO File Rules produce, if it is coming from the default rule.  But if you don't set anything, maybe that causes other problems for OIIO?  I wonder if there should be some way of indicating that a color space is provisional (i.e., a default guess) rather than a known quantity. Then at least OIIO would have something to use when needed, but it would not write that back out in new files?

zachlewis:
Thanks, Doug!

> The part I'm not sure about is what config OIIO would use by default, if the user does not specify one. As Zach wrote, that seems like the correct place to impose a lin_rec709_scene type default and even the parseColorSpaceFromString usage.

Currently, OIIO uses "ocio://default" if not otherwise specified, which I think is probably not the best default config for OIIO users for a variety of reasons, outside the scope of this. But to underscore one of Doug's points, `getColorSpaceFromFilepath`  only does the `parseColorSpaceFromString` thing if config authors _opt in_ to such behavior in their configs by specifying the `ColorSpaceNamePathSearch` rule. And `ocio://default` does not enable the Path Search rule.


> It seems like one risk of assigning a lin_rec709_scene default is that this will get preserved through a sequence of operations and then written back out as if it is a known quantity instead of a guess. And that's how metadata becomes unreliable. So I can understand why the code fragment above is actually not using what the OCIO File Rules produce, if it is coming from the default rule.

Damn good point. You're right, we should leave the wrapper as is -- this way, the config authors can use the "strictparsing" value to dictate whether OIIO throws an exception or just uses the default color space when an ImageBufAlgo that expects "oiio:ColorSpace" encounters an empty string. 

I think the idea of a "provisional" or "default guess" or "oiio:doNotWriteColorMetadata" flag is a good one. It could be set whenever the default color space is returned by the `getColorSpaceFromFilepath` wrapper (when setting the color space), and unset whenever `ImageSpec::set_colorspace` is used in other circumstances.

zachlewis:
> > Also worth mentioning, OCIO's file naming rules will always indicate a color space.
> 
> Can you clarify? What does it return if there's no apparent color space name in the filename?

So, to clarify, I think I may have misled you here. Currently, if `ColorConfig::getColorSpaceFromFilepath(...)` cannot otherwise identiy a color space name from the string input, it will either return an empty string (if the Config's strictparsing value == True) or the default color space name (if strictparsing == False). 

(Strict parsing tends to be enabled in just about all the configs I've seen "in the wild")

lgritz:
There's a lot to review above, but I'll just take little bites for now.

I think there are two interesting questions with respect to file path rules:

1. Is this something that should happen for every file, every plugin? (In this case, I mean at the EXRInput level?) Or should an ImageInput report *just* what's in the file, and only at a higher level like oiiotool if you use `--autocc` should it try to apply any filepath name logic?

2. What's a reasonable behavior if the filepath does not contain a color space? I tend to think that there should NOT be a default name, and that no name in the file path means it's unknown, don't guess. Especially in a non-studio environment where the config is probably the built-in one.

lgritz:
I would like to separate out some concerns here: Let's focus in this PR on the narrow task of what should be included in this single change that serves to undue the unwise OIIO-3.0 decision to mark exr files as lin_rec709 by default. But in #4787, we should have a more wide ranging discussion about the entire hitlist of behavior changes we desire for color management (color interop, CICP, OCIO 2.5 support, what does "oiio:ColorSpace" contain, etc.).

On the narrow "this PR" issue, I think I'm detecting the following action items from the review comments, please let me know if I've got this right:

* We should augment the logic of this PR to look for the acesImageContainer attribute, and if found, it overrides all other considerations and the color space is lin_ap0_scene. (Wait, what am I looking for? I don't see that listed in OpenEXR's ImfStandardAttributes.h.)
* If the color interop is "data", as many places as possible in OIIO should refuse to do color transformations, you have to assign a meaningful color space first. (This item, maybe, doesn't need to be in this PR because it's not really an exr issue.)
* Clarify what happens with the file path call -- I think that we should NOT assign a default space if the path has no color space name, and that should be treated as if it were "unknown."
* Historically, we have ignored Chromaticities. Should we or shouldn't we honor that if it matches known chromaticities, in the absence of the other rules taking precedence?

Also, as a reminder: OIIO's minimum OCIO currently is 2.2, and because our build is now able to auto-build OCIO and link it statically, there's really nothing preventing us from bumping to a higher OCIO (even up to 2.5, if the timing is right) for OIIO 3.1.

Is there already an in-progress release notes for OCIO 2.5? I'm not at all up to speed on what additional facilities we should be taking advantage of or what utilities have been added that will help any of this.

zachlewis:
Great questions!

> Is this something that should happen for every file, every plugin? (In this case, I mean at the EXRInput level?) Or should an ImageInput report just what's in the file, and only at a higher level like oiiotool if you use --autocc should it try to apply any filepath name logic

In my opinion, ImageInput should be as dumb as possible (unless there's a global "autocc" flag enabled); and otherwise, users will have to either use "--autocc" to set the color space automatically or "--iscolorspace" to set the color space manually for downstream operations; or otherwise specify the input color space when applying colorspace-conversion and display-view ImageBufAlgos.  

> What's a reasonable behavior if the filepath does not contain a color space? I tend to think that there should NOT be a default name, and that no name in the file path means it's unknown, don't guess. Especially in a non-studio environment where the config is probably the built-in one.

I agree. And that's pretty much what currently happens. 

Essentially, the failover behavior is determined by the config author -- if Config::isStrictParsingEnabled(), ColorConfig:getColorSpaceFromFilepath() returns an empty string; else, it returns the default color space.

Typically, OCIO configs' "strictparsing" fields are set to True, which means `ColorConfig::getColorSpaceFromFilepath(...)` returns empty strings if it can't find a match. Setting "strictparsing" to False in the config means the function returns the default color space instead of empty strings.

lgritz:
> In my opinion, ImageInput should be as dumb as possible

I agree. Perhaps ImageInput should only be responsible for taking what's actually in the file and turning it into metadata in the ImageSpec. That's it.

We should have some kind of utility function exposed that takes an ImageSpec, and a filename, and of course implicitly the active OCIO config, and is capable of assessing all of the clues therein to synthesize one of { an OCIO color space name from the active config, "data",  "unknown" }. That's what's acted upon by any IBA functions involving color transformations.

One or more utility functions that are used by the ImageOutput implementations that take the various clues in the ImageSpec and can figure out from that how to turn it into the file-specific ways of specifying color information (e.g. a CIF token if it can figure it out, or a CICF tuple if the output format supports that, etc.)


> Essentially, the failover behavior is determined by the config author -- if Config::isStrictParsingEnabled(), ColorConfig:getColorSpaceFromFilepath() returns an empty string; else, it returns the default color space.

Er, here I'm not sure. From OIIO's point of view, most random users will not have a config found and therefore rely on a built-in config. What do the built-in configs say is the default? If it's empty, then fine. But if not, I'm not sure that's the right behavior for OIIO.

> Typically, OCIO configs' "strictparsing" fields are set to True, which means 

Same follow-up: what is this for the built-in configs, which are the configs that most users in non-professionaly-color-managed situations will encounter.

doug-walker:
@lgritz , the configs built into the OCIO library currently set strictparsing to true because that is the default for the library (and the library always writes it), though there isn't really a strong argument about why that should be the case and it certainly could change in the future. In fact, that attribute is currently not used by the OCIO library except in the deprecated parseColorSpaceFromString function. In particular, the OCIO version of getColorSpaceFromFilepath ignores strictparsing and always return a color space. The filepathOnlyMatchesDefaultRule function is used to determine if that was from the default rule. So what Zach wrote above is specifically about OIIO's getColorSpaceFromFilepath.

Regarding some of the other points raised above:
- I agree that the low-level function should not try to assign a color space if it's not specifically in the file header. The use of file rules and etc. should happen at a higher level that are more under user control.
- I just looked at the EXR standard attributes header and was surprised that acesImageContainer is missing. It should be added, perhaps as part of [EXR PR 2075](https://github.com/AcademySoftwareFoundation/OpenEXR/pull/2075).  That is what SMPTE ST 2065-4 specifies. Though obviously, it doesn't have to be a standard attribute for OIIO to access it.
- If OIIO has historically ignored Chromaticities, this is probably not the time to start using it.
- OCIO 2.5 is still in progress and there is no good information yet about the upcoming API calls (there is [OCIO PR 2165](https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/2165), but that is not finished and there are more functions coming).

lgritz:
I amended the PR to adjust color space logic:

* Recognize "acesImageContainerFlag" as necessarily being lin_ap0_scene.
* Don't apply file naming rules at the EXRInput level.

doug-walker:
The new commit looks ok, but I want to clarify that the InteropID is technically the combination of a color space name and an optional namespace. (This was done in order to keep it to one attribute in the header rather than two.) Whether it's best to assign the entire interop ID to the OIIO color space is not clear to me. (Whether the entire thing should be treated as equivalent to a color space name is still a topic of discussion in OCIO & CIF.)

zachlewis:
> I amended the PR to adjust color space logic:
> 
>     * Recognize "acesImageContainerFlag" as necessarily being lin_ap0_scene.
> 
>     * Don't apply file naming rules at the EXRInput level.

Perfect!

Also, note #4791, requesting a way to write ST 2065-4-complaint EXRs (i.e., 16f uncompressed RGB/A EXRs with the "acesImageContainerFlag" and "chromaticities" attributes set). 

I agree with Doug, that we shouldn't do anything with "chromaticities" -- in fact, I'd recommend ignoring and erasing any "chromaticities" or "CICP" attributes found in the spec whenever an IBA that changes "oiio:ColorSpace" is applied. 


> Same follow-up: what is this for the built-in configs, which are the configs that most users in non-professionaly-color-managed situations will encounter.

Sorry, I should have spoken more clearly: 

"strictparsing" is set to True in the builtin configs; and it's also set to True when you create a new config with the API. So a config author would have to go out of their way to enable strict parsing. Historically, OIIO's use of "parseColorSpaceFromString" leads me to believe that  OIIO must have been "obeying" the Config's "strictparsing" parameter.

In other words, colorspace-aware IBAs that can't determine a valid input color space name will through an exception, _unless_ the config author has set `strictparsing: false`, indicating that they expect colorspace-aware IBAs to muscle through, using a possibly-incorrect-but-always-valid default color space as a last resort. 


> So what Zach wrote above is specifically about OIIO's getColorSpaceFromFilepath.

Yeah, I'm sorry if I've made things more confusing... OIIO's ColorConfig::getColorSpaceFromFilepath is a wrapper around OCIO's Config::getColorSpaceFromFilepath. Maybe it should be deprecated in favor of a better name? 

The intention was to create a function we could swap in for our use of "parseColorSpaceFromString". I didn't want to do anything too crazy in the original PR, but ultimately, I was gonna suggest having a bit (elsewhere) that automatically injects the ColorSpaceFromFilepath rule just before the Default rule if it isn't already specified in the active config's FileRules, and getting rid of the call to the deprecated "parseColorSpaceFromString" entirely in our wrapper; which would effectively ignore the "strictparsing" setting in the config, unless we specifically implement logic to do something with it. 

Anyway, let's decide what we want to do, moving forward:
1. Always failover to an empty string, potentially causing IBAs to throw exceptions (risk of false negatives)
2. Always failover to a default color space specified by the config, potentially causing IBAs to succeed in doing the wrong thing (risk of false positives)
3. Defer to the config's "strictparsing" value, slinging empty strings and throwing exceptions unless the config author goes out of their way to set "strictparsing" to false
4. Any of the above, but with an overloaded form of the function that also takes an argument for forcibly overriding the empty-string-vs-default-color-space-string behavior.

lgritz:
> The new commit looks ok, but I want to clarify that the InteropID is technically the combination of a color space name and an optional namespace. (This was done in order to keep it to one attribute in the header rather than two.) Whether it's best to assign the entire interop ID to the OIIO color space is not clear to me. (Whether the entire thing should be treated as equivalent to a color space name is still a topic of discussion in OCIO & CIF.)

Understood. It's a little academic since there are no OpenEXR files in the wild that have the InteropID, so I don't think we need to nail down what happens with the namespace quite yet. This is just setting the "oiio:ColorSpace" attribute which is internal only, so I think we can decide what it means. I'm tempted to sweep this under the rug for now, until the CIF documents are all finalized, OCIO 2.5 takes more shape, and OpenEXR merges its PRs.

lgritz:
I feel like there are some missing parts (out of scope for this PR and yet to be implemented):

`string deduce_colorspace(const ImageSpec&, string_view filename = {})` that roots through the ImageSpec (with filename as a fallback heuristic) to produce an OCIO color space name. It may know about a number of different possible metadata to figure it out, with some order of precedence. If "oiio:ColorSpace" is there, it takes precedence: it means somebody along the chain of custody has figured it out and made a declarative statement.

IBA functions that do color transformations will call deduce_colorspace() to know the input color space. They rely on OCIO for transformations, so only OCIO color spaces or configname:colorspace will have an effect. Those IBA functions will set oiio:ColorSpace in their output as the single source of truth, and erase any other contradictory attributes.

Then we will need some way to reverse the process upon output to files -- since some file types can have the OCIO names, others want CICP, etc.

Files that DON'T pass through IBA color-transforming functions just end up with all the metadata they started with, so an exr -> exr round trip doesn't lose any CICP or anything else.

Is that feeling like it's in the neighbourhood of a reasonable behavior?

lgritz:
@zachlewis Do you think this is ok for the limited scope of what we're trying to do in this PR?

zachlewis (approved):
This looks perfect!

> Is that feeling like it's in the neighbourhood of a reasonable behavior?

Yes, I'm seeing 100% eye-to-eye with you -- `deduce_colorspace` would do exactly the right thing at exactly the right time. The method should resolve "oiio:ColorSpace" to the "canonical name" for the color space for a given ColorConfig; and we could use it to pretty effortlessly "translate" between names used by the default (global) ColorConfig and custom IBA-specific ColorConfigs and back. 

> Then we will need some way to reverse the process upon output to files -- since some file types can have the OCIO names, others want CICP, etc.

Yes, exactly. And maybe this function could take into consideration the state of an internal "uncertainty" attribute -- i.e.,  refuse to propagate metadata if the attribute is tripped upstream (e.g., if ever `deduce_colorspace` returns a "default" color space because "strictparsing" is disabled; or if explicitly set by the user...). The function should also provide an option for translating "oiio:ColorSpace" to a given OCIO Config's "interopID" convention for the equivalent color space.

PR #4842, opened 2025-07-30
build: Propagate CMAKE_MSVC_RUNTIME_LIBRARY

lgritz:
* Allow CMAKE_MSVC_RUNTIME_LIBRARY to be initialized from an environment variable of the same name. (CMake doesn't do that on its own.)

* If set, propagate CMAKE_MSVC_RUNTIME_LIBRARY to any packages we auto-build within our build system.

### DIFF
--- a/src/cmake/compiler.cmake
+++ b/src/cmake/compiler.cmake
@@ -655,6 +655,17 @@ endif ()
 
 
 ###########################################################################
+# Windows: which MSVC runtime library should we use (to override default)?
+# Note that all dependencies need the same choice. We leave this as the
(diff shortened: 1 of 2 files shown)


lgritz:
Anybody with Windows experience want to give an approve? I need this in place to fix something on the OSL side.

lgritz:
> I got a LGTM offline from somebody who tried it and confirmed the fix, but isn't allowed to comment publicly. So merging.

Edit: oh crap, I posted this in the wrong ticket. Never mind.

lgritz:
I'm going to merge this by the end of the day if there are no objections, so that it can be incorporated into the release tomorrow.

PR #4843, opened 2025-07-30
fix(python): got strides wrong passing 2D numpy pixel array

lgritz:
When passing numpy arrays for pixel blocks (say, to ImageOutput.write_image), the sorting out of what the numpy arrays meant had an error for the 2D case. A typo (or maybe copy pasta) got the stride wrong.

This took a long time to be discovered because if there is more than one channel, it gets passed as a 3D numpy array (height, width, chans). So only for 1-channel images would it stumble across the bad code if it were passed as (height, width) with no dimension for channels.


lgritz:
To trigger the bug, it takes a single channel image that's specifically passed to write_image as only a 2D numpy array and not 3D.

PR #4844, opened 2025-07-31
build: C++23 support

lgritz:
* CI build as C++23 for bleeding edge test

* Fixes to disambiguate OIIO::print from std::print.  Apparently, at least for gcc14, std::print ends up in the global namespace and really messes with things.

* Change to the library that contains the implementation of C++23 stacktrace in gcc14 vs 12/13.


lgritz:
Can I get a review of this from somebody? It's been posted for a week now. It touches a lot of files, but almost all of the changes are simply renaming "print" to "OIIO::print" to avoid an compilation ambiguity with std::print.

grdanny (approved):
looks reasonable to me.

PR #4841, opened 2025-07-28
Add maketx flags to increase feature parity with txmake

scottdmilner:
## Description
- Bug fix with prman tile size for bumpslopes textures
- `--slopefilter`: maketx has used sobel, txmake uses centraldiff because sobel sometimes has ringing artifacts
- `--bumpinverts`: invert slopes in the s dir, txmake default behavior has this on
- `--bumpinvertt`: invert slopes in the t dir, txmake default behavior has this on
- `--bumpscale`: scale the strength of the resulting map
- `--bumprange`: support normal map values  on either [-1,1] or [0,1]
- `--keepaspect`: sets PixelAspectRatio metadata to match txmake `up` resize behavior.

`txmake -bumprough 1 0 0 0 0 0`: `maketx --bumpslopes --bumpinverts --bumpinvertt --slopefilter centraldiff`
`txmake -bumprough 2.5 0 0 1 0 0`: `maketx --bumpslopes --bumpscale 2.5 --bumpinvertt --slopefilter centraldiff`

## Tests

*In-progress*

## Checklist:

- [x] I have read the [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
- [x] I have updated the documentation, if applicable. (Check if there is no
  need to update the documentation, for example if this is a bug fix that
  doesn't change the API.)
- [ ] I have ensured that the change is tested somewhere in the testsuite
  (adding new test cases if necessary).
- [x] If I added or modified a C++ API call, I have also amended the
  corresponding Python bindings (and if altering ImageBufAlgo functions, also
  exposed the new functionality as oiiotool options).
- [x] My code follows the prevailing code style of this project. If I haven't
  already run clang-format before submitting, I definitely will look at the CI
  test that runs clang-format and fix anything that it highlights as being
  nonconforming.

PR #4845, opened 2025-08-01
test: Update ref output

lgritz:
PRs #4840 and #4841 crossed in flight with their tests passing individually, but they needed one small adjustment to reference output in order to to account for their interaction in main.


lgritz:
Because this doesn't change code and fixes a completely broken CI across most job variants, I will pre-emptively merge this as soon as it fully passes CI.

PR #4846, opened 2025-08-01
ci: Update ABI check standard

lgritz:
ABI changed slightly after #4827. (Allowed in main but requires an update of the reference.)

Also, bump to using the 2025 container for the ABI check.


lgritz:
Needed to fix CI, merging.

PR #4849, opened 2025-08-04
build: clean up obsolete logic for old compilers

lgritz:
* Enforce gcc, clang, and apple clang minimum versions.

* Raise advertised minimum clang version to 10.0. Error for clang < 5,
  which we know is the true minimum because earlier than that won't
  support C++17 which we require. But in truth, we haven't tested
  versions older than 10.0 for a long time, so we should not be
  implying we know older versions will work or that we will support
  them.

* Simplify cases no longer relevant for older gcc and clang.

* Remove unneeded gcc guard in simd.h.  I think it was important only
  for gcc 5! But it was also preventing the slightly faster construct
  from activating for clang. Seems to pass all our CI jobs with this
  simplified.


lgritz:
Any comments?

lgritz:
This has been in review without comment for 3 weeks. Merging.

PR #4850, opened 2025-08-05
Add ISO setting of the camera

TodicaIonut:
## Description

ISO setting of the camera

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1537,6 +1537,9 @@ The official OpenEXR site is http://www.openexr.com/.
    * - ``ExposureTime``
      - float
      - expTime
+   * - ``ISOSpeed``
+     - float
+     - isoSpeed
    * - ``FNumber``
      - float
      - aperture
--- a/src/doc/stdmetadata.rst
+++ b/src/doc/stdmetadata.rst
@@ -289,6 +289,10 @@ The following metadata items are specific to photos or captured images.
 
     The exposure time (in seconds) of the captured image.
 
+.. option:: "ISOSpeed" : float
+
+    The ISO speed of the camera when it captured the image.
+
 .. option:: "FNumber" : float
 
     The f/stop of the camera when it captured the image.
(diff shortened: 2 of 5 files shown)


lgritz:
Can you explain what the purpose of this PR is, what problem is being solved and what behavior change you are aiming for? I can't quite tell what the point is, since it's already possible for OpenEXR to store arbitrary named metadata.

JGoldstone:
Moreover isoSpeed has been in the implementation for years if not decades

//
// isoSpeed -- the ISO speed of the film or the ISO setting of the camera
// that was used to record the image
//

IMF_STD_ATTRIBUTE_DEF (isoSpeed, IsoSpeed, float)

lgritz:
OIIO already correctly reads and writes any of the arbitrary named metadata in an exr file (certainly for all the standard types like float and string, etc).

The only time we need special handling is for particular metadata items that are very important (and important to preserve when copying image from one file format to another), and that has DIFFERENT names or different types for each file formats. In those cases, we pick one canonical name and type and have the other formats automatically translate between the internal canonical name/type and the format-specific name.

An example of is this is that most formats have some kind of title, description, caption, or whatever, but they all call it something different. We chose the TIFF name for this, "ImageDescription", as our canonical name (somewhat arbitrarily, but mostly because 15 years ago, TIFF was widely used and long established). So in the openexr reader, we take the OpenEXR attribute "comments" and change it to "ImageDescription", and in the writer, we change "ImageDescription" to the standard OpenEXR attribute "comments." This basically lets you do things like

    oiiotool in.tif -o out.exr
    oiiotool in.exr -o out.tif

and the data will get transferred properly between the files, despite how the two formats use different nomenclature for it, and also for applications to be able to ask for a single name, "ImageDescription" to get this common data, regardless of file type, and without needing to know that it might have been different original names in different files.

lgritz:
To spell it out more concretely, when we identify one of these items for which we wish to create a single canonical cross-format name for:

* We choose a name used by one format, the one that we think is most descriptive, commonly used, or sometimes just default to the historical precedent of the name used by the oldest widely used format that picked a name for that common item.
* The format that already uses the name we want to use canonically _does not need to change_ since it's already using the "right" name.
* _Other formats_ may need translation from and to our chosen canonical name.

PR #4851, opened 2025-08-07
use get_plane2 introduced by libheif 1.20.2

toge:
## Description

libheif 1.20.2 introduces `get_plane2()` for compatibility. [ref](https://github.com/strukturag/libheif/issues/1566)

Because `get_plane()` has been marked as deprecated, we should use `get_plane2()` instead.

## Checklist:

  doesn't change the API.)
  nonconforming.


src/heif.imageio/heifinput.cpp
```diff
@@ -390,8 +390,13 @@ HeifInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
 #else
     int ystride = 0;
 #endif
+#if LIBHEIF_NUMERIC_VERSION >= MAKE_LIBHEIF_VERSION(1, 20, 2, 0)
+    const uint8_t* hdata = m_himage.get_plane2(heif_channel_interleaved,
+                                              &ystride);
```
lgritz:
```suggestion
                                               &ystride);
```
ever so slight formatting alignment issue

lgritz (approved):
LGTM! Thanks for the fix.

I tested it on my end as well -- and since on my laptop, Homebrew just upgraded to the new libheif which was broken, and this fixes it perfectly, I can verify directly that this is the right fix.

src/heif.imageio/heifinput.cpp
toge: Thanks!

lgritz:
Just out of curiosity, do you know why this was done on the libheif side? It looks like the old and new functions take the same arguments, so why make a new function and deprecate the old one instead of just changing the internal implementation of the one that existed?

kmilos:
See https://github.com/strukturag/libheif/issues/1566

toge:
@lgritz 
Thanks!

lgritz:
I'll do a backport immediately and a patch for 3.0.9.1 since people on Mac in particular might immediately run into this build break if they're getting their libheif dependency from Homebrew.

PR #4856, opened 2025-08-13
ci: Add a VFX Platform 2026 CI job

lgritz:
And a bunch of other minor updates:
* Change from 'testing' to regular containers
* Add forgotten 'export'
* Update default self-build pybind11 to 3.0.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,7 +294,7 @@ jobs:
           - desc: VFX2024 gcc11/C++17 py3.11 exr3.2 ocio2.3
             nametag: linux-vfx2024
             runner: ubuntu-latest
-            container: aswf/ci-osl:2024-clang17
+            container: aswf/ci-oiio:2024
             python_ver: "3.11"
             simd: "avx2,f16c"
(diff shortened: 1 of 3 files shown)


jfpanisset (approved):
Good to use the `aswf/` rather than `aswftesting/` images, it takes so long to build and release these images that I don't really want to chew up CPU minutes / hours / days releasing both, so in the future I may only be releasing the `aswf/` images.

Two small details:

- there are ci-oiio images for 2024 and 2025, including ones that are hot off the presses, so you could use these instead of the ci-osl images
- you are specifying pybind11 3.0.0 for your 2026 builds, but what's included in the ci-oiio 2026 (and most recent 2024/2025) images is 2.13.6, so I don't know if that means that OIIO will prefer downloading it's own 3.0.0 rather than using the pre-installed 2.13.6. For the next / "feature complete" set of 2026 images (when OpenVDB/OCIO/OpenEXR release their VFX 2026 versions) I will likely update to pybind11 3.0.0

jfpanisset (approved):
LGTM

PR #4857, opened 2025-08-13
build: fix build on NetBSD

0-wiz-0:
## Description

These patch fix the build on NetBSD 11.99.1. There were a couple problems:
- some missing `str*i*()` functions
- `isnan` etc. not being used from the proper `std::` namespace
- a missing header for `va_list`
- NetBSD also handles `long long` differently from `int64`
- 

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_compare.cpp
+++ b/src/libOpenImageIO/imagebufalgo_compare.cpp
@@ -21,8 +21,6 @@
 
 #include "imageio_pvt.h"
 
-using namespace std;
-
 OIIO_NAMESPACE_BEGIN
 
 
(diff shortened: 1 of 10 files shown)


src/libOpenImageIO/imagebufalgo_compare.cpp
```diff
@@ -21,6 +21,8 @@
 
 #include "imageio_pvt.h"
 
+using namespace std;
+
```
lgritz: I'm not super crazy about blanket 'using namespace std'. Can we just put `std::` in front of the places that need it?

0-wiz-0: Thanks for the review!
There were just a lot more in this file :-) I've used prefixes now.

lgritz: No worries. It would have been ok to do one function at a time for the whole file, too, like `using std::isnan;`, it's only the blanket pulling in of the *entire* `std::` namespace that I worried could lead to confusion later. But it's great as it is now.

lgritz: Oops, I think you need to re-clangformat that one file now.

0-wiz-0:
Thanks, and sorry for the second clang-format problem, I didn't expect the line break :-)
I looked over the CI failures, but they don't seem related to my change.

lgritz:
I think those failures, clearly unrelated, are just a random GitHub glitch.

PR #4859, opened 2025-08-15
ci: Lock down to ci-oiio container with correct llvm components

lgritz:


lgritz:
I think something else is still mis-configured. Not sure if you can access [this log](https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/16985839925/job/48154464407) but seeing errors like this, only on the aswf/ci-oiio:2024 containers:

```
-- Configuring done (10.1s)
CMake Error at /usr/local/lib/cmake/OpenColorIO/OpenColorIOTargets.cmake:61 (set_target_properties):
  The link interface of target "OpenColorIO::OpenColorIO" contains:

    MINIZIP::minizip-ng

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

  /usr/local/lib/cmake/OpenColorIO/OpenColorIOConfig.cmake:77 (include)
  src/cmake/dependency_utils.cmake:379 (find_package)
  src/cmake/externalpackages.cmake:139 (checked_find_package)
  CMakeLists.txt:213 (include)


CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_FMT_VERSION


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

PR #4860, opened 2025-08-16
feat: color space renaming to adhere to CIF conventions

lgritz:
This doesn't really change any of the color management algorithms per se, or change how we use OCIO. That will come later.

But this tries as best as we can to shift to CIF recommended interop names for color spaces.

* "srgb_rec709_display" as the canonical name for what we used to refer to as "sRGB".
* "lin_rec709_scene" as the canonical name for what we used to refer to as "lin_rec709" and several other names.
* "lin_ap1_scene" as the canonical name for ACEScg.
* Stop setting "oiio:Gamma" metadata (though we still accept it for back compatibility), instead using space names like "g22_rec709_display" and so on. We never came across files with truly arbitrary gamma.

If I've done it right, the old names will continue to be accepted as aliases.

I recognize that changing the output (when printing metadata) to the canonical names might be an inconvenience at first, but in the long run, I think it's better to comply with the conventions that will be used for USD, MaterialX, and OpenEXR.

I'm fine with the new wordy and pedantic names as what is stored underneath as the "oiio:ColorSpace" alias or stored in an OpenEXR file in the colorInteropID metadata. But I gotta say, from a UI/UX perspective, I do not feel good about replacing "srgb" and "acescg" as things that are easy to type and are known/understood by the vast majority of OIIO users with a change to these wonky, error-prone, hard-to-type names. So I'm very uncertain about changing oiiotool documentation, say, away from the short and widely understood aliases. Seeking feedback on what to do here!


lgritz:
After some consultation on the ocio slack channel, I changed to using certain _scene space names instead of _display. PR has been so amended.

src/doc/stdmetadata.rst
```diff
@@ -144,29 +144,24 @@ Color information
...
       configuration.
-    - `"lin_srgb"`, `"lin_rec709"` :  Color pixel values are known to be
-      linear and using sRGB/Rec709 color primaries. Note that `"linear"` is
-      treated as a synonym.
-    - `"sRGB"` :  Using standard sRGB response and primaries.
+    - `"lin_rec709_scene"`,  :  Color pixel values are known to be linear
+      scene-referred and using sRGB/Rec709 color primaries. Note that
+      `"lin_rec709"` is treated as a synonym.
+    - `"lin_ap1_scene"`, `"ACEScg"` :  ACEScg color space encoding.
+    - `"lin_ap0_scene"` :  ACES.
```
doug-walker:
```suggestion
    - `"lin_ap0_scene"` :  ACES2065-1, the recommended ACES space for interchange and archiving.
```

src/doc/stdmetadata.rst
```diff
@@ -144,29 +144,24 @@ Color information
...
+    - `"srgb_rec709_scene"` : Using standard sRGB response and primaries.
```
doug-walker:
```suggestion
    - `"srgb_rec709_scene"` : Using standard (piecewise) sRGB response and primaries.
```

src/doc/stdmetadata.rst
```diff
@@ -144,29 +144,24 @@ Color information
...
+       The token `"sRGB"` is treated as a synonym.
+    - `"g24_rec709_scene"` : Rec709/sRGB primaries, but using a response curve
+      corresponding to gamma 2.4.
```
doug-walker: The CIF Texture/CG color space list doesn't have the color space for g24_rec709_scene, though that is the string we would use if we add it. I didn't realize this was one of the standard OIIO color spaces, maybe we should add it to the CIF set? As things stand now, a namespace prefix would need to be added to use it as an Interop ID, unfortunately.

The string is used several places below, which is probably fine. When it comes time to write this into the new EXR attribute, that could add a namespace.

src/doc/stdmetadata.rst
```diff
@@ -144,29 +144,24 @@ Color information
...
+    - `"g22_rec709_scene"` : Rec709/sRGB primaries, but using a response curve
+      corresponding to gamma 2.2.
+    - `"g18_rec709_scene"` : Rec709/sRGB primaries, but using a response curve
+      corresponding to gamma 1.8.
     - `"Rec709"` :  Using standard Rec709 response and primaries.
```
doug-walker: It would be useful to clarify which transfer function this uses. The entry for g24_rec709_display in this document may help explain why:
https://docs.google.com/document/d/1MmBG4a3Dr6S6EO781WjK-xZW7QdHpuo-zd7wtMvG1Rs/edit?tab=t.0

Unlike most of the other ones, this one is less obvious whether it should be scene- or display-referred.

src/doc/stdmetadata.rst
```diff
@@ -144,29 +144,24 @@ Color information
...
-    - `"ACEScg"` :  ACEScg color space encoding.
     - `"AdobeRGB"` :  Adobe RGB color space.
```
doug-walker: The CIF lists have both a scene-referred and display-referred version of AdobeRGB. I'm not sure which would be more appropriate for OIIO to use.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -613,13 +620,13 @@ ColorConfig::Impl::classify_by_name(CSInfo& cs)
 
     // Set up some canonical names
     if (cs.flags() & CSInfo::is_srgb)
-        cs.canonical = "sRGB";
-    else if (cs.flags() & CSInfo::is_Rec709)
-        cs.canonical = "Rec709";
+        cs.canonical = "srgb_rec709_scene";
     else if (cs.flags() & CSInfo::is_lin_srgb)
-        cs.canonical = "lin_srgb";
+        cs.canonical = "lin_rec709scene";
```
doug-walker:
```suggestion
        cs.canonical = "lin_rec709_scene";
```

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -736,7 +744,7 @@ ColorConfig::Impl::identify_builtin_equivalents()
         if (CSInfo* cs = find(n)) {
             cs->setflag(CSInfo::is_lin_srgb | CSInfo::is_linear_response,
                         lin_srgb_alias);
-            DBG("Identified {} = builtin '{}'\n", "lin_srgb", cs->name);
+            DBG("Identified {} = builtin '{}'\n", "lin_rec709scene", cs->name);
```
doug-walker:
```suggestion
            DBG("Identified {} = builtin '{}'\n", "lin_rec709_scene", cs->name);
```

doug-walker (commented):
I think you're making the correct substitutions here for lin_rec709 and sRGB. For a few of the others, per the comments below, it's less obvious what would be best. Would be interesting to get Zach's thoughts.

src/doc/stdmetadata.rst
lgritz: It seems we documented this aspirationally, but we don't seem to use it anywhere. Maybe it's best if I just remove this.

But per your earlier advice, I would have guessed scene-referred? I think that was the conclusion -- assume everything is scene-referred unless told otherwise, or we know it's been tone mapped somehow.

src/libOpenImageIO/color_ocio.cpp
lgritz: oh, thanks for catching this typo

src/doc/stdmetadata.rst
lgritz: I have no idea. Does it help to know that the places we use it are for DPX, Cineon, and PPM files, which as I understand it called themselves Rec709 and not sRGB, and I had been using this transfer function (in case it helps you identify what's going on):

```
inline float
Rec709_to_linear(float x)
{
    if (x < 0.081f)
        return x * (1.0f / 4.5f);
    else
        return powf((x + 0.099f) * (1.0f / 1.099f), (1.0f / 0.45f));
}
```

I totally would believe I've been handling this wrong all along.

We may wish to tackle this set of issues separately from the rest of the PR, since it may implicate misunderstandings that have persisted for a long time and are unrelated to this simple set of renamings.

doug-walker: As I wrote in the document linked above, "Rec.709" (i.e., HD video) is complicated. ITU-R BT.709 is for a camera curve (OETF) and ITU-R BT.1886 is for the display curve (EOTF) and they are not inverses of one another. The OETF is essentially scene-referred, and the EOTF is display-referred.

Your formula above is for the OETF. There is no CIF equivalent for the OETF, although it does exist in the built-in Studio config.

In most cases, if I had to pick a default for "Rec.709" files, I would use the EOTF. This would be g24_rec709_display in the CIF list.

However, given that we've generally been choosing the scene-referred option in OIIO, and your existing implementation uses the OETF (scene-referred), it does seem defensible to continue as is.

I would update the definition as follows:
`- `"Rec709"` :  Using the ITU-R BT.709 camera response (OETF) and primaries.`

src/doc/stdmetadata.rst
doug-walker: I would remove it then, if it's not being used.

src/doc/stdmetadata.rst
lgritz: I can remove this, I'm not sure it has any important uses.

src/doc/stdmetadata.rst
lgritz: The important use case here is DPX files. Looking at this from Amendment 1:2012 to SMPTE 268M-2003:

dpxcodes

If I get a DPX file using code 6, what CIF (or built-in studio config) color space should I be saying that is?

src/doc/stdmetadata.rst
lgritz: I rewrote the section a bit and eliminated all but the most commonly used names.

src/doc/stdmetadata.rst
zachlewis: Tricky. I agree with Doug -- which is to say:
 - In practice, I'd most likely use "g24_rec709_display" as the default for decoding "Rec.709-encoded" images
 - But to be internally consistent with how we're treating "sRGB", it makes sense to do likewise for "Rec709", absent any other information.
 

 
In any case, I don't think it would hurt to list both "g24_rec709_display" and "rec709_camera" in the documentation.

zachlewis:
I think this looks good.

Important to note, for Reader plugins of integer-based formats and/or formats that don't explicitly contain a "colorInteropID" attribute -- (so, pretty much everything except OpenEXR) -- I think it's important to attempt to use the OCIO config's file rules (`Config.getFilenameFromPath(...)`) before failing over to a default "colorInteropID" attribute value. I'm pointing this out because according to the current proposal, where OpenEXRs are concerned, the "colorInteropID" value is always considered _before_ the filename. 

In other words, due to the ambiguity inherent in ascertaining "the right thing to do" with a given DPX or PNG or whathaveyou, we want to delegate, wherever possible, all decisions regarding color space identification to the OCIO config author; so we need to make sure that any of these hard-coded defaults, CIF-token or otherwise, are only ever considered as a last resort.

src/doc/stdmetadata.rst
zachlewis: > If I get a DPX file using code 6, what CIF (or built-in studio config) color space should I be saying that is?

Personally, I think DPX metadata is even less reliable than the EXR `chromaticities` attribute; and OIIO shouldn't attempt to infer anything about the encoding based on the transfer characteristics function.

lgritz:
> we need to make sure that any of these hard-coded defaults, CIF-token or otherwise, are only ever considered as a last resort.

Understood. I think that should maybe fall to a subsequent PR that actually messes with the color logic. The present one is, to the greatest extent possible, just trying to modernize the names we are using for things.

PR #4861, opened 2025-08-17
ci: Bump webp and openexr for "latest versions" test

lgritz:
* Test against alpha of openexr 3.4.
* Document that we build against several things we know work but hadn't updated INSTALL.md in a while.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -395,7 +395,7 @@ jobs:
             cxx_std: 20
             fmt_ver: 11.2.0
             opencolorio_ver: v2.4.2
-            openexr_ver: v3.3.4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -395,7 +395,7 @@ jobs:
             cxx_std: 20
             fmt_ver: 11.2.0
             opencolorio_ver: v2.4.2
-            openexr_ver: v3.3.4
+            openexr_ver: v3.4-alpha
```
jessey-git, line 398: Would be good to change the `desc` too to say exr3.4 I guess.

lgritz: yes, good point, will do

PR #4867, opened 2025-08-22
build: Some more minor wheel workflow changes after the py 3.9 bump

lgritz:


zachlewis:
Hmmm. I'm not sure how to interpret the CI failures here... the linker is detecting allegedly undefined symbols, but not consistently...


```
cd /var/folders/vk/nx37ffx50hv5djclhltc26vw0000gn/T/tmpwlx_i2by/build/src/libOpenImageIO && /usr/local/bin/cmake -E cmake_link_script CMakeFiles/OpenImageIO.dir/link.txt --verbose=1
    ld: Undefined symbols:
      OpenImageIO_v3_1_4_Imf__3_2_4::Chromaticities::Chromaticities(Imath_3_1::Vec2<float> const&, Imath_3_1::Vec2<float> const&, Imath_3_1::Vec2<float> const&, Imath_3_1::Vec2<float> const&), referenced from:
          OpenImageIO::v3_1_4::OpenEXROutput::put_parameter(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO::v3_1_4::TypeDesc, void const*, OpenImageIO_v3_1_4_Imf__3_2_4::Header&) in exroutput.cpp.o
```

lgritz:
Yeah, that's the same MacOS on Intel problem we're seeing on all branches.

lgritz:
This PR *started* as part of an attempt to fix the problem. Briefly, I thought "aha! maybe this is it, we didn't update everything after moving to python 3.9 minimum", but it doesn't help. Nonetheless, it's still legit stuff to update, so I spun it off into this PR. But the build failures remain.

PR #4870, opened 2025-08-28
build: ffmpeg 8 support

lgritz:
* Try a build against ffmpeg 8 -- seems to work out of the box, so document that we support that version in INSTALL.md.
* Augment FindFFmpeg.cmake to correctly identify 7.1 and 8.0 by the very non-intuitive versioning scheme of ffmpeg's headers.
* Add a FFMPEG_EXTRA_LIBRARIES variable to FindFFmpeg.cmake to allow us to easily inject extra library dependencies (coincidentally, I need this at work to force it to use static libraries).


lgritz:
Any objections? I need this.

lgritz:
I need this now and it's an incredibly minor, non-breaking, build-only change, so I'm going to pre-emptively merge it.

PR #4871, opened 2025-08-29
exr: OpenEXR 3.4 supports two compression types for HTJ2K

TodicaIonut:
## Description

`htj2k32` or `htj2k256` compression options.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/builtinplugins.rst
+++ b/src/doc/builtinplugins.rst
@@ -1541,7 +1541,7 @@ The official OpenEXR site is http://www.openexr.com/.
      - string
      - one of: ``"none"``, ``"rle"``, ``"zip"``, ``"zips"``, ``"piz"``,
        ``"pxr24"``, ``"b44"``, ``"b44a"``, ``"dwaa"``, ``"dwab"``, ``"htj2k256"`` or ``"htj2k32"``.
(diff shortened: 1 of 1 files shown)


src/doc/builtinplugins.rst
```diff
@@ -1540,8 +1540,8 @@ The official OpenEXR site is http://www.openexr.com/.
    * - ``compression``
      - string
      - one of: ``"none"``, ``"rle"``, ``"zip"``, ``"zips"``, ``"piz"``,
-       ``"pxr24"``, ``"b44"``, ``"b44a"``, ``"dwaa"``, ``"dwab"`` or ``"htj2k"``.
-       (``"htj2k"`` is only supported with OpenEXR 3.4 or later.)
+       ``"pxr24"``, ``"b44"``, ``"b44a"``, ``"dwaa"``, ``"dwab"``, ``"htj2k256"`` or ``"htj2k32"``.
+       (``"htj2k256"`` or ``"htj2k32"`` is only supported with OpenEXR 3.4 or later.)
```
lgritz:
```suggestion
       (``"htj2k256"`` and ``"htj2k32"`` are only supported with OpenEXR 3.4 or later.)
```

PR #4873, opened 2025-09-02
Look for boost headers for OpenVDBs older than 12

boberfly:
## Description

Look for boost headers for OpenVDBs older than 12. I was getting issues along the lines of this:
```
C:\Users\bober\packages\openvdb\11.0.0\platform-windows\arch-AMD64\python-3.11\include\openvdb\math\Math.h(13,10): error C1083: Cannot open include file: 'boost/numeric/conversio
n/conversion_traits.hpp': No such file or directory [C:\dev\rpks\openimageio\3.0.9.0\build\df31b1486e506952d10f9e8ae1f90b5a2144114d\_OpenImageIO\working\OpenImageIO-3.0.9.0\build
\src\libOpenImageIO\OpenImageIO.vcxproj]
```
Not too sure if my approach is the best but it fixed it for me on Windows MSVC v143/VS2022. I was building against OpenVDB 11 but I didn't test older versions. OpenVDB 12 I did test and it wasn't needed.

## Tests

Not too sure how to make one, it is more or less a compiling bugfix.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/modules/FindOpenVDB.cmake
+++ b/src/cmake/modules/FindOpenVDB.cmake
@@ -89,6 +89,16 @@ if (OpenVDB_FOUND)
         set_property(TARGET OpenVDB::openvdb APPEND PROPERTY
             IMPORTED_LOCATION "${OPENVDB_LIBRARIES}")
     endif ()
+
+    # Note: OpenVDB older than 12 needs Boost headers.
+    if (OpenVDB_VERSION VERSION_LESS 12)
+        find_package (Boost)
+        if (Boost_FOUND)
+            list(APPEND OPENVDB_INCLUDES ${Boost_INCLUDE_DIRS})
+        else()
+            unset(OpenVDB_FOUND)
+        endif()
+    endif()
 endif ()
 
 MARK_AS_ADVANCED(


src/cmake/externalpackages.cmake
```diff
@@ -182,6 +182,10 @@ checked_find_package (OpenVDB
                       VERSION_MIN  9.0
                       DEPS         TBB
                       DEFINITIONS  USE_OPENVDB=1)
+# Note: OpenVDB older than 12 needs Boost headers.
+if (OpenVDB_FOUND AND OpenVDB_VERSION LESS 12)
```
lgritz: should be `VERSION_LESS`

lgritz:
I'm tempted to think that the extra logic here should instead all be going inside FindOpenVDB.cmake.

src/cmake/modules/FindOpenVDB.cmake
```diff
@@ -89,6 +89,11 @@ if (OpenVDB_FOUND)
         set_property(TARGET OpenVDB::openvdb APPEND PROPERTY
             IMPORTED_LOCATION "${OPENVDB_LIBRARIES}")
     endif ()
+
+    # Note: OpenVDB older than 12 needs Boost headers.
+    if (OpenVDB_VERSION VERSION_LESS 12)
+        checked_find_package (Boost)
```
lgritz: I think that if this succeeds, you can just append Boost_INCLUDE_DIRS to OPENVDB_INCLUDES, and then the other change in openvdb.imageio/CMakeLists is not needed at all?

And should we bother to handle the case of if this find fails? The robust solution would be for a failure to find boost to actually set OpenVDB_FOUND to false again, maybe?

boberfly: I'm not sure if it's correct practice in CMake to just do an `unset(OpenVDB_FOUND)` but I did a quick test by removing `Boost_ROOT` and it did make it drop OpenVDB so it looks like that works.

lgritz: I mean, if we're really going to get picky about CMake best practices, it would involve not having FindOpenVDB at all, and the OpenVDB project's build should produce the correct exported cmake configs with all the right targets defined (which would include the boost include dirs without OIIO needing to figure it out at all).

lgritz:
Well, that kind of backfired, now it's failing a bunch of tests. I'm not quite sure why.

src/cmake/modules/FindOpenVDB.cmake
```diff
@@ -89,6 +89,16 @@ if (OpenVDB_FOUND)
...
```
lgritz:
```suggestion
        find_package (Boost)
```

I think it's failing because checked_find_package is not recursion-safe.

lgritz: Yeah, definitely. I can see how the checked_find_package is not at all safe to be called recursively.

So we either need to use find_package here instead of checked_find_package, or else move the boost logic back to externalpackages.cmake (sorry, maybe I steered you wrong).

boberfly: All good @lgritz I force-pushed the `find_package` approach and see what the CI will do - sorry I just got some time to get back to this, and if it doesn't work I'll put it back outside...

lgritz (approved):
LGTM to me now, sorry about the runaround

PR #4874, opened 2025-09-03
feat(oiiotool): allow easy splitting of subimages by name

lgritz:
Say you need to take a multi-part exr file and split it into separate files for each part. (Note: what OpenEXR calls "parts", OIIO calls "subimages.") You could do this

    oiiotool multipart.exr --sisplit -o:all=1 "part.%04d.exr"

But this would name the parts part.0001.exr, part.0002.exr, etc. Maybe you would prefer to split them into files that contain the subimage name rather than just the inscrutible index number (which may also not be consistent across files)? You could try this dance to loop over the subimages and write each one with a name extracted from its metadata:

    oiiotool multipart.exr --for i {TOP.SUBIMAGES} --dup --subimage {i} -o "subimage_{TOP.'oiio:subimagename'}.exr

That works, but that's a little awkward and hard to remember off the top of your head. Perhaps instead we would prefer a simpler version:

    oiiotool multipart.exr --sisplit -o:all=1 "part.{TOP.'oiio:subimagename'}.exr"

This PR makes that last command work. It actually only required a small change, which is that the function that handles the `-o` command needs to defer the evaluation of the expression subsitution in the filename to occur separately for each subimage being output, rather than just once for the whole `-o:all=1` (which would have baked the FIRST subimage name into all the output file names).

Docs are updated to give examples of how to do this.

Also fixed a silly typo in the test script.

### DIFF
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -868,11 +868,17 @@ Split a multi-image file into separate files
 --------------------------------------------
 
 Take a multi-image TIFF file, split into its constituent subimages and
-output each one to a different file, with names `sub0001.tif`,
-`sub0002.tif`, etc.::
+output each one to a different file, with names `sub.0001.tif`,
+`sub.0002.tif`, etc.::
 
-    oiiotool multi.tif -sisplit -o:all=1 sub%04d.tif
+    oiiotool multi.tif -sisplit -o:all=1 sub.%04d.tif
 
+Take a multi-image OpenEXR file (called "multi-part" in OpenEXR parlance),
+split into its constituent subimages and output each one to a different file,
+with names `sub.beauty.exr`, `sub.albedo.exr`, etc., using the name of each
+subimage according to its metadata::
+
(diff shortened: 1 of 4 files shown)


lgritz:
@christopherdavies This addresses something we discussed on a help ticket recently.

src/doc/oiiotool.rst
```diff
@@ -1498,17 +1504,25 @@ Writing images
         Output all images currently on the stack using a pattern.
         See further explanation below.
 
-    The `all=n` option causes *all* images on the image stack to be output,
-    with the filename argument used as a pattern assumed to contain a `%d`,
-    which will be substituted with the index of the image (beginning with
-    *n*). For example, to take a multi-image TIFF and extract all the
-    subimages and save them as separate files::
+    The `all=n` option causes *all* images on the image stack to be output. If
```
grdanny, line 1507: is it all=n? or all=1?

lgritz: it can be any number *n*. The number supplied is the beginning index, for the case where you use "%d" to substitute the number.

PR #4877, opened 2025-09-11
feat(openexr): Support for idManifest and deepImageState

lgritz:
Support for retrieving and setting the idManifest for deep image files with object ID channels. This is an OpenEXR feature where the image header can contian metadata that maps all the object ID channel integer codes to strings of the object names.

We report this in the ImageSpec as the metadata "openexr:compressedIDManifest", whose type is an array of uint8 (byte) values. It is encoded the same way it appears in the exr file itself: the first 8 bytes are a little endian uint64_t giving the *uncompressed* size of a serialized OpenEXR IDManifest object, and then starting at byte 8, the zip-compressed seriailized IDManifest. Therefore, the compressed block size is the size of the metadata blob, minus 8 bytes for the length. It is up to the caller to use the OpenEXR APIs to turn this compressed serialized IDManifest into a fully expanded IDManifest, if that's what they want to do.

It works for output, too, so it should copy the manifests correctly between files, even if the app doesn't know what to do with the binary data it contains.

Also added support for deepImageState, which we previously had ignored.  We report this as the metadata "openexr:deepImageState", a string, which if present has one of the values "messy", "sorted", "non_overlapping", or "tidy".


lgritz:
This has been sitting around for a while. Does anybody object or have feedback?

lgritz:
I will merge this by the weekend if there are no objections.

PR #4878, opened 2025-09-12
feat(webp): Support reading/writing the ICCProfile attribute

jessey-git:
## Description

Enable ICCProfile attribute support for WEBP for both input and output. Like other formats this support is passive; pixel data will not be altered by the mere presence of the ICCProfile attribute (i.e. it functions like other informational metadata only).

----
The WebP API is a bit odd when it comes to writing the ICCProfile. We must use the "mux" API set; the "advanced" API is apparently not advanced enough to add the metadata to the file. This wouldn't be too bad except that the mux API is at odds with how we write out the data. Before: When it comes time to do the file write, we would pass our own file writer into webp. The writes would go straight to disk. Now: If ICC profile data is necessary to write, we let webp use its own writer to write to memory. We then assemble the in-memory file and the ICC profile data together and pass that entire package to our writer as a final step. 

When no ICC profile data is used, the old write behavior is kept.

## Tests

Added `oiiotool` test variations which uses ICC profiles with WEBP.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/webp.imageio/webpoutput.cpp
+++ b/src/webp.imageio/webpoutput.cpp
@@ -216,6 +216,7 @@ WebpOutput::write_complete_data()
 
         ok = ioproxy()->write(assembly.bytes, assembly.size) == assembly.size;
 
+        WebPDataClear(&assembly);
         WebPMuxDelete(mux);
         WebPMemoryWriterClear(&wrt);
     }


lgritz:
The code looks fine, but do you by any chance have a test webp file with an ICC profile so we can verify that it's working?

jessey-git:
Yes, attached are 2 example .webp's with an ICC profile in each. The icc profile files are also included for cross checking.
[eamples.zip](https://github.com/user-attachments/files/22289623/eamples.zip)

lgritz:
I think what I meant was: should we add a validation to the testsuite?

jessey-git:
Ah yeah :)  I had a question about that since I couldn't find any tests for ICC, but I took another look and discovered there's already an `oiiotool` test that does this for JPG, so I added a WEBP variation to it.

PR #4879, opened 2025-09-12
fix: Switch to compile-commands

vvalderrv:
## Description

Fix SonarCloud analysis failures by switching the C/C++ configuration from the
deprecated `sonar.cfamily.build-wrapper-output` property to the current
`sonar.cfamily.compile-commands` property.

**Changes**
- Remove `sonar.cfamily.build-wrapper-output` from `sonar-project.properties`.
- In `.github/workflows/build-steps.yml`, pass  
  `--define sonar.cfamily.compile-commands="$BUILD_WRAPPER_OUT_DIR/compile_commands.json"`.

**Root cause**  
The SonarCloud config still used the deprecated `sonar.cfamily.build-wrapper-output`
property. The current CFamily analyzer expects a `compile_commands.json`
(produced by build-wrapper). Using the old property caused scanner failures and
blocked analysis.

**Validation**  
Regenerated `bw_output/compile_commands.json`, ran `sonar-scanner`, and
confirmed the scan completed and uploaded successfully.

## Tests

CI-only change. Verified locally by running the scanner against a fresh
build-wrapper capture.

## Checklist

- [x] I have read the
  [contribution guidelines](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md).
- [ ] I have updated the documentation, if applicable. *N/A — CI-only change.*
- [ ] I have ensured that the change is tested somewhere in the testsuite.
  *N/A — CI-only change.*
- [ ] If I added or modified a C++ API call, I have also amended the Python
  bindings. *N/A — not applicable.*


lgritz:
There is something going wrong in the sonar analysis job. That's on me to fix, I will look at it today! This failure is happening during the build, and is entirely unrelated to both your patch and the previously reported Sonar issues (which, when I filed the ticket, finished the build fine and failed on the actual sonar analysis step itself). Something else must have broken. Stand by, I will get that fixed and then we can see if your patch cleared up the other situation.

lgritz:
@vvalderrv 

I amended your PR with a fix that will let it get to the analysis step, avoiding the other spurious problem.

But you can see that it still crashes in the sonar analysis, as I originally reported:
https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/17681745315/job/50257265809

PR #4880, opened 2025-09-13
feat(heif): Read and write of CICP and bit depth 10 and 12

brechtvl:
## Description

* Read and write 10 and 12 bit images
* Read and write CICP

This makes it possible to write HDR images that are much smaller than PNGs, in a file format that is supported in all modern web browsers.

By setting the matrix coefficients, the library will perform RGB to YUV conversion on write. The YUV to RBG conversion was already happening on read automatically.

A future improvement would be a better default choice of matrix coefficients for writing. When no CICP is specified or when CICP is read from PNG, then no conversion to YUV will be performed and compression will not work as well. This is the same behavior as before. When there is support for conversion between CICP and display interop ID, a good default choice for matrix coefficients could be made as part of that.

## Tests

Tests for CICP and 10 bit were added.

## Checklist:

  doesn't change the API.)
  nonconforming.


brechtvl:
Example file: [nas_sole_mates.zip](https://github.com/user-attachments/files/22308389/nas_sole_mates.zip)

src/heif.imageio/heifinput.cpp
```diff
@@ -402,7 +449,19 @@ HeifInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+    if (m_spec.format == TypeUInt16) {
+        // Convert from 10 or 12 bits to 16 bits. Read little endian
+        // zero padded bits, and shift to scale up.
+        const uint8_t* in       = hdata;
+        uint16_t* out           = static_cast<uint16_t*>(data);
+        const int bitshift      = 16 - m_bitdepth;
+        const size_t num_values = m_spec.width * m_spec.nchannels;
+        for (size_t i = 0; i < num_values; i++, out++, in += 2) {
+            *out = uint16_t((in[1] << 8) | in[0]) << bitshift;
+        }
```
lgritz: Are you sure this is correct? If you're converting merely by shifting, that means that for 10 bits (for example), a fully saturated 0b1111111111 will turn into 16 bit 0b1111111111000000, but really you want it to be 0b1111111111111111, so that it's truly 1.0.

We have a utility in fmath.h called bit_range_convert that does this correctly, and you can see it used in several of the other readers like targa, rla, jpeg2000, etc.

brechtvl: Switched to using `bit_range_convert` now.

lgritz (approved):
LGTM. Is this ready to merge?

brechtvl:
Yes, it's ready.

PR #4881, opened 2025-09-13
ci: fix analysis workflow configuration

lgritz:
Something must have changed with the containers, it was failing on the build step, before ever getting to the actual sonar analysis.

Switch to the same container and OCIO version that we use for the regular 2024 era container CI.


lgritz:
Merging because it's just a CI change, and only to the analysis workflow, which is broken but this still is needed.

PR #4883, opened 2025-09-13
ci: better spread of libpng versions we test against

lgritz:
* The "oldest" didn't really test against the oldest we claim to support.
* The "latest versions" didn't really test against the latest versions.
* The "bleeding edge" didn't really test against libpng master.
* The auto-build was unnecessarily a few releases behind.
* Documentation touch-ups

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
             opencolorio_ver: v2.3.0
             pybind11_ver: v2.9.0
             setenvs: export FREETYPE_VERSION=VER-2-12-0
+                            BUILD_PNG_VERSION=1.6.30
           - desc: VP2022 clang13/C++17 py39 avx2 exr3.1 ocio2.3
             nametag: linux-vfx2022.clang13
             runner: ubuntu-latest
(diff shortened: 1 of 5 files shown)


lgritz:
Oh, I think I've done something bad here. Looks like I'm getting confusion between the libpng I'm building and the one already installed. Let me iterate on this a bit.

lgritz:
vfx2023 was some random glitch, worked fine when I just told it to rerun.
I think the sanitizer failure is my fault, fixing

lgritz:
Updated and fixed

lgritz:
There's an older version of libpng already on the system image, but as you pointed out, that was too old to support CICP, and so was not exercising this new code.

In my first stab at this PR, I just ran the build_libpng.bash script that we use for CI to build and install, thus ending up with a second version of the library. Because of inclulde paths and LD_LIBRARY_PATH ordering (which can be really hard to get right), it ended up using the headers of one but the library of the other and finding missing symbols. That's what the failures were.

But setting OpenImageIO_BUILD_LOCAL_DEPS=PNG gets us a new version in a totally different way: it runs our build system's "autobuild" approach for that package, which builds a static version and doesn't get confused about the system version.

jessey-git:
Gotcha, it was the detail about how this build option provides a static build that I was missing.

PR #4885, opened 2025-09-14
fix: FFmpeg sets zero oiio:BitsPerSample

brechtvl:
## Description

Many codecs don't provide `bits_per_raw_sample`. The bit depth of the luma channel is the closest equivalent to a single bit depth then, while the chroma channels may be subsampled.

Tested on basically all the video files on my computer, they all give a non-zero bit depth that looks correct now.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/ffmpeg.imageio/ffmpeginput.cpp
+++ b/src/ffmpeg.imageio/ffmpeginput.cpp
@@ -30,6 +30,7 @@ extern "C" {  // ffmpeg is a C api
 #endif
 
 #include <libavutil/imgutils.h>
+#include <libavutil/pixdesc.h>
 }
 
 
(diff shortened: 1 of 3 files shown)


lgritz:
In case it changes how you want to do this: It is not necessary to set "oiio:BitsPerSample" unless it differs from what is implied by ImageSpec.format. For example, if the standard type you are returning is uint16, but you want to say, "I'm conveying this data via 16 bit ints, and of course 65535 is interpreted as 1.0, but just you know, in the file, it was stored only 10 bits," then you set "oiio:BitsPerSample" to 10.

brechtvl:
In my tests this changes a bunch of cases from 0 bits to 8, 10 or 12 bits. So I think it is useful.

I could not set it if it matches the format, but other file formats seem to set it unconditionally.

lgritz:
> I could not set it if it matches the format, but other file formats seem to set it unconditionally.

If that's what all the others are doing, I think it's fine.

PR #4886, opened 2025-09-14
ci: disable macos-x86 wheels.

zachlewis:
Temporary workaround to unblock publishing of other wheels for 3.1 release


zachlewis:
Workaround for #4854. I'm still trying to find a proper solution in #4855, but if I don't find a solution before the release OIIO-3.1, feel free to merge this PR instead; and we'll release a hotfix with the macos-x86 wheels when we have one available.

.github/workflows/wheel.yml
```diff
@@ -227,57 +227,57 @@ jobs:
...
-      - uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
-        with:
-          name: cibw-wheels-${{ matrix.python }}
-          path: ./wheelhouse/*.whl
+  # macos:
+  #   name: Build wheels on macOS
+  #   runs-on: macos-13
+  #   if: |
+  #     github.event_name != 'schedule' ||
+  #     github.repository == 'AcademySoftwareFoundation/OpenImageIO'
```
lgritz, line 235: Just as an FYI, I think you could just change this to

```suggestion
    if: |
      0 && (github.event_name != 'schedule' ||
      github.repository == 'AcademySoftwareFoundation/OpenImageIO)'
```

and not need to comment out the whole thing.  Or even only do it if the branch name is something special, so it would only run for you while you're continuing to test, but not for anyone else.

zachlewis: ooohhh, smart... 

Unless you feel otherwise, I think I'd rather leave it as-is, just to make it super apparent what was changed, and that I intend to change it back; but I'm happy to do whatever you feel is best.

lgritz: No problem, I will merge as-is.

zachlewis: thanks for showing me this, btw -- I've been using this trick to quickly "solo" the macos-x86 wheels tasks to help me debug #4855

PR #4887, opened 2025-09-15
feat(raw): add thumbnail support to the raw input plugin

### LINKED ISSUE #4107: [FEATURE REQUEST] RAW: Ability to read the embedded thumbnails of raw files.
Most RAW-files have embedded thumbnails in them. It's currently not possible to access them with OIIO.

It would be super useful to be able to access them.

Using LibRaw directly, one can access them with the command line tool simple_dcraw.

**Additional context**
In LibRaw, you can get the highest resolution thumbnail with the command line tool `simple_dcraw -v -e raw_file.cr2`
or you can get all the thumbnails with `simple_dcraw -v -E raw_file.cr2`
So look at simple_dcraw for inspiration.

antond-weta:
This adds thumbnail support to the raw input plugin. 

Fixes #4107

This supports 3 of the 4 formats of thumbnails Libraw provides: BMP, JPEG, JPEGXL(untested). H265 is not supported, I don't know of any camera using it, and expect the code being significantly different from the other formats, given that the H265 is a movie container.

One thing I don't like about this implementation - I understand there is no way to query the number of thumbnails available in the file? We are expected to iterate, until `get_thumbnail()` returns false? If `get_thumbnail()` returns false not because there is no thumbnail for the requested index, but any other reason, we are not going to try reading past that index.

## Tests

I have tested manually with a bunch of raw images having either JPEG or BMP thumbnails. 

I have not tested the JPEGXL branch, as I'm not aware of any cameras using that for thumbnails. The code is identical to the JPEG branch though, so I'm not expecting any issues with that.

I couldn't see any unit tests around the thumbnail functionality at all. Should I add some? I assume we can't use `oiiotool` for this?

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -1710,7 +1710,7 @@ void
 _errorfmt(const RawInput* input, int subimage, const char* format,
           const Args&... args)
 {
-    std::string fmt = "Failed to unpack thumbnail at index "
+    std::string fmt = "Failed to extract thumbnail at index "
                       + std::to_string(subimage) + ": " + format + ".";
     input->errorfmt(fmt.c_str(), args...);
 }
@@ -1764,7 +1764,8 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
     }
 
     if (image_type.empty()) {
-        _errorfmt(this, subimage, "unknown image type %i", mem_thumb->type);
+        _errorfmt(this, subimage, "unknown image type %i",
+                  static_cast<int>(mem_thumb->type));
         return false;
     }
 
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -1718,6 +1718,12 @@ _errorfmt(const RawInput* input, int subimage, const char* format,
 bool
 RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
 {
+    if (m_processor == nullptr) {
+        _errorfmt(this, subimage,
+                  "ImageInput hasn't been initialised properly");
+        return false;
+    }
+
 #if LIBRAW_VERSION < LIBRAW_MAKE_VERSION(0, 21, 0)
     if (subimage > 0) {
         // Older versions of Libraw supported a single thumbnail per image.
@@ -1726,7 +1732,7 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
     }
     int errcode = m_processor->unpack_thumb();
     if (errcode != 0) {
-        if (errcode != LIBRAW_REQUEST_FOR_NONEXISTENT_THUMBNAIL)
+        if (errcode != LIBRAW_REQUEST_FOR_NONEXISTENT_IMAGE)
             _errorfmt(this, subimage, "unpack_thumb error");
         return false;
     }
@@ -1764,7 +1770,7 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
     }
 
     if (image_type.empty()) {
-        _errorfmt(this, subimage, "unknown image type %i",
+        _errorfmt(this, subimage, "unknown image type {}",
                   static_cast<int>(mem_thumb->type));
         return false;
     }
@@ -1782,8 +1788,8 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
     } else {
         auto image_input = OIIO::ImageInput::create(image_type, false);
         if (image_input == nullptr) {
-            _errorfmt(this, subimage, "OIIO::ImageInput::create(\"%s\") error",
-                      image_type.c_str());
+            _errorfmt(this, subimage, "OIIO::ImageInput::create(\{}\") error",
+                      image_type);
             return false;
         }
 
@@ -1791,8 +1797,8 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
         bool result = image_input->valid_file(&proxy);
         if (!result) {
             _errorfmt(this, subimage,
-                      "the thumbnail is not a valid image of type \"%s\"",
-                      image_type.c_str());
+                      "the thumbnail is not a valid image of type \"{}\"",
+                      image_type);
             return false;
         }
 
@@ -1815,8 +1821,8 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
         if (!result) {
             _errorfmt(
                 this, subimage,
-                "failed to initialise an ImageInput object of type \"%s\" with the thumbnail data",
-                image_type.c_str());
+                "failed to initialise an ImageInput object of type \"{}\" with the thumbnail data",
+                image_type);
             return false;
         }
     }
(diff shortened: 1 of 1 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -1702,4 +1705,75 @@ RawInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
     return true;
 }
 
+bool
+RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
+{
+#if LIBRAW_VERSION < LIBRAW_MAKE_VERSION(0, 21, 0)
+    if (subimage > 0)
+        return false;
```
soswow, line 1731: I can see you don't tell the user why it returns false in any of the cases. Maybe error logging if verbose mode? or I am not sure what is the best way to communicate reason of failure.

lgritz: The other (few) implementations of get_thumbnail set error messages if get_thumbnail failed to read from the file, or something of that nature, but just return false without setting an error message if there simply was no thumbnail available.

lgritz: I'm inclined to accept this PR as-is, since its error verbosity is not especially different than the other readers that can extract thumbnails. But if you think it would be helpful to have specific error messages for the different ways that a thumbnail can fail to be found, it's fine to submit a short follow-on PR that improves the errors (and maybe in that case, we should audit the other get_thumbnail implementations and ensure they are all doing a decent job of this).

If somebody does this, please note that some of the other get_thumbnail implementations do lots of things like this example from TGAInput::get_thumbnail:
```
    if (!ioread(&res, 2, 1))
        return false;
```

This is a case that should NOT be modified, because ioread (and ioseek) already issue error messages if they fail, so get_thumbnail can just return false straightaway in the case of one of those methods failing. We would only want to add error messages for cases where we return false due to tests that do not internally already handle errors.

lgritz (approved):
LGTM

Up to you whether to do a subsequent fix to give more detailed error messages.

lgritz:
> One thing I don't like about this implementation - I understand there is no way to query the number of thumbnails available in the file? We are expected to iterate, until get_thumbnail() returns false? If get_thumbnail() returns false not because there is no thumbnail for the requested index, but any other reason, we are not going to try reading past that index.

The unfortunate answer is that when get_thumbnail() was added, we really didn't consider that there might be more than one thumbnail per subimage. (Remember that a "subimage" is cases where there are multiple distinct images in a file, like frames of a movie or "parts" of an openexr file.) So currently, there simply is no way to ask how many thumbnails there are, or their resolutions, or to select among them. I guess if there are more than one, you should just pick one that seems like an appropriate preview resolution and return it?

I always thought of a thumbnail as a single, fairly low-res, preview image. I see that libraw and some raw formats support more flexibility, I guess, but the prior formats we added thumbnail support to did not have such a feature.

Maybe in the future, we could extend get_thumbnail to take an additional "thumbnail_index" parameter, and you could iterate though until you fail if you want to retrieve all of them. Or, images with thumbnails could set some metadata saying how many and their resolutions, and then the caller would know which indices are valid and wanted?

Or, another strategy for formats that truly contain a range of preview images of varying resolutions (including some that seem like real images and are not tiny previews) would be to instead treat them as MIP levels, which are designed to express multiple reduced resolution copies of a single image (and need not strictly differ by resolution factors of 2).

lgritz:
> I couldn't see any unit tests around the thumbnail functionality at all. Should I add some? I assume we can't use oiiotool for this?

testsuite/targa-thumbnail tests thumbnails for targa, using python and a call to get_thumbnail

I'm a little surprised to see that oiiotool has no thumbnail-extraction command, that seems like it would be useful. Though I also can understand how it would be a little awkward. Let's address that separately, I will file an issue for it.

lgritz:
Noting for the record here that there was an issue for raw thumbnails!

Fixes #4107

antond-weta:
I have added some error reporting last night.

> testsuite/targa-thumbnail tests thumbnails for targa, using python and a call to get_thumbnail

I didn't see this, thanks! Let me see if I can add some unit tests quickly.

src/raw.imageio/rawinput.cpp
```diff
@@ -1705,24 +1705,46 @@ RawInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
     return true;
 }
 
+template<typename... Args>
+void
+_errorfmt(const RawInput* input, int subimage, const char* format,
+          const Args&... args)
```
lgritz, line 1711: You could make this template function be a private templated method of RawInput, and then you wouldn't need to pass the input pointer every time?

lgritz: I think this might be over-engineered. Remember that these are all errors that are being generated by a call to get_thumbnail(). So the caller knows they were asking for a thumbnail and what subimage they were asking for, they just need to know why it failed. I think you can go simpler and shorter without this helper, if you wish. This is probably more than sufficient:

```
    bool RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
    {
        ...
        if (failure) {
            errorfmt("thumbnail failure: can't foo the bar");
            return false;
        }
        ...
    }
```

It's fine if you want the complex and wordy one. You're in charge of raw behavior as far as I'm concerned. But it's also fine to go simple by assuming the caller knows the full context of the request.

src/raw.imageio/rawinput.cpp
```diff
@@ -1764,7 +1764,8 @@ RawInput::get_thumbnail(ImageBuf& thumb, int subimage)
     }
 
     if (image_type.empty()) {
-        _errorfmt(this, subimage, "unknown image type %i", mem_thumb->type);
+        _errorfmt(this, subimage, "unknown image type %i",
```
lgritz: errorfmt uses the std::format syntax, not printf.

```suggestion
        _errorfmt(this, subimage, "unknown image type {}",
```

antond-weta:
Oh, in python bindings `get_thumbnail()` discards the boolean value the lib method returns, and just returns an ImageBuf object; so there are essentially 3 states:
- a thumbnail image returned in the ImageBuf, with no error message set;
- an ImageBuf with an error message returned if failed to read thumbnail;
- an uninitialised ImageBuf returned if no thumbnail for the requested index available.

This should be enough to make a unittest, but we probably should look at improving this logic in the bindings.

lgritz:
> Oh, in python bindings `get_thumbnail()` discards the boolean value the lib method returns, and just returns an ImageBuf object; so there are essentially 3 states:

Oh, that's kind of sloppy. I feel like maybe the failure cases should return None and set an error message in the ImageInput, not in the returned ImageBuf? Only return an ImageBuf if it's a valid image and there were no errors. Does that sound more Pythonic?

But in any case, this PR need not solve that existing problem. We can come back to it (if you don't intend to come back to it yourself, please file an issue so it's not forgotten).

As the prerequisite to merging this PR, I'm more concerned about the 3 failing test jobs, which appear to be about your path not working with the full range of libraw versions that we support -- looks like on some runners, we are getting a version of libraw that lacks a symbol you are using.

PR #4890, opened 2025-09-15
fix: gif output didn't handle FramesPerSecond attribute correctly

### LINKED ISSUE #3716: [BUG] Unsettable delay for animated GIFs w/proposed fix
**Describe the bug**
I am trying to reencode an animated GIF from an `IOMemReader` to an `IOVecOutput`.  All the code for copying the images and creating the output works except that I cannot set the delay (FPS, FramesPerSecond) in the output.  I've reviewed the code, and cannot see how I'm supposed to be able to set the delay without a change to the OIIO library.

Let me walk you through the library code's logic.  In gif.h the routine `GifWriteLzwImage` encodes the output using the argument delay.  `GifWriteLzwImage` is called from gif.h's `GifWriteFrame` routine which also gets delay from an argument.  `GifWriteFrame` is called by gifoutput.cpp's `GIFOutput::finish_subimage` which passes `GIFOutput::m_delay`.  `GIFOutput::m_delay` only gets set in a single place, `GIFOutput::open(const std::string&, int, const ImageSpec*)`.  This `open` routine is where I believe the issue to be.  `m_delay` is set using the attribute "FramesPerSecond" in `m_spec`.  You'll notice that `open` gets an `ImageSpec` as an argument.  The passed in `ImageSpec` is only ever used to initialize the `m_subimagespecs` array.  The `m_subimagespecs` array is never used for anything, and can be removed.  It's my belief that `m_spec`, which is used to create `m_delay`, is never initializable from outside the library.  From outside the library you're able to control the `specs` argument, but that's never actually used.  So I believe the fix is to use `specs` to initialize `m_spec`, and then get the "FramesPerSecond" attribute.

**To Reproduce**
1. Run code to take https://media.tenor.com/QWAo2echlVEAAAAC/cute.gif and reencode the animated GIF with the same 0.1 second delay.  That's a "FramesPerSecond" attribute of 100/10.

**Expected behavior**
It should be possible to use the OpenImageIO library to write an animated GIF with a non-default delay between frames.

**Evidence**
No matter what I do, the animated GIFs I write with OpenImageIO always encode with the default 1 seconds delay.  Hopefully my code review is evidence enough, but I'd appreciate any alternate way of viewing the fix.  Maybe there's another way to set `m_delay` or `m_spec` that I'm unaware of.

**Platform information:**
 - OIIO branch/version: master branch commit 1449b465000bf7c83e983cf870ca7b1f2966d446
 - OS: CentOS Stream 9
 - C++ compiler: gcc version 11.3.1 20221121
 - Any non-default build flags when you build OIIO: None

**IF YOU ALREADY HAVE A CODE FIX:**
Here's the fix that I've tested.  If there's an alternate solution, please let me know.  Thank you.
```
diff --git a/src/gif.imageio/gifoutput.cpp b/src/gif.imageio/gifoutput.cpp
index e033f221..14a8c0e4 100644
--- a/src/gif.imageio/gifoutput.cpp
+++ b/src/gif.imageio/gifoutput.cpp
@@ -83,7 +83,6 @@ private:
     int m_subimage;  // Current subimage index
     int m_nsubimages;
     bool m_pending_write;                    // Do we have an image buffered?
-    std::vector<ImageSpec> m_subimagespecs;  // Saved subimage specs
     GifWriter<Filesystem::IOProxy> m_gifwriter;
     std::vector<uint8_t> m_canvas;  // Image canvas, accumulating output
     int m_delay;
@@ -154,7 +153,9 @@ GIFOutput::open(const std::string& name, int subimages, const ImageSpec* specs)
     m_filename   = name;
     m_subimage   = 0;
     m_nsubimages = subimages;
-    m_subimagespecs.assign(specs, specs + subimages);
+    if (specs != nullptr) {
+        m_spec = *specs;
+    }
     float fps = m_spec.get_float_attribute("FramesPerSecond", 1.0f);
     m_delay   = (fps == 0.0f ? 0 : (int)(100.0f / fps));
```

lgritz:
The version of open that takes an array of subimage specs didn't use them properly for finding the FramesPerSecond hint (or the ioproxy!).

Fixes #3716

### DIFF
--- a/src/gif.imageio/gifoutput.cpp
+++ b/src/gif.imageio/gifoutput.cpp
@@ -155,10 +155,10 @@ GIFOutput::open(const std::string& name, int subimages, const ImageSpec* specs)
     m_subimage   = 0;
     m_nsubimages = subimages;
     m_subimagespecs.assign(specs, specs + subimages);
-    float fps = m_spec.get_float_attribute("FramesPerSecond", 1.0f);
+    float fps = specs[0].get_float_attribute("FramesPerSecond", 1.0f);
     m_delay   = (fps == 0.0f ? 0 : (int)(100.0f / fps));
 
-    ioproxy_retrieve_from_config(m_spec);
+    ioproxy_retrieve_from_config(specs[0]);
     if (!ioproxy_use_or_open(name))
         return false;
 


jessey-git:
I'm trying to validate the fix but I'm running into a few problems. Is the OIIO recipe of `oiiotool.exe "t:/render/foo_#.png" --siappendall --attrib FramesPerSecond 10.0 -o t:/anim-10fps.gif` correct? I'm not seeing the GIFOutput opened with the `AppendSubimage` mode. It's always `Create` and I end up with a single-image .gif at the very end.

Side quest: Using a file path like `t:\render\foo_#.png` is broken on windows because the `\` characters aren't escaped and so the regex fails to find the sequence of files at all :/

lgritz:
Here's a recipe I used to test, that didn't depend on any wildcard expansion:

```
oiiotool -pattern constant:color=1,0,0 100x100 3 -pattern constant:color=0,1,0 100x100 3 -pattern constant:color=0,0,1 100x100 3 --siappendall --attrib FramesPerSecond 50.0 -o anim.gif
```

lgritz:
Printing again with line breaks to make it clearer for readers:

```
oiiotool -pattern constant:color=1,0,0 100x100 3 \
         -pattern constant:color=0,1,0 100x100 3 \
         -pattern constant:color=0,0,1 100x100 3 \
         --siappendall --attrib FramesPerSecond 50.0 -o anim.gif
```

Obviously, try changing the 50 to different values.

lgritz:
And I found that just pointing a browser to anim.gif will display it properly with frame timing.

jessey-git:
Alright, yes, your examples also work for me. So the PR itself is good.

So are the docs wrong for the wildcard expansion? https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/doc/oiiotool.rst?plain=1#L816

lgritz:
The docs example is

```
oiiotool foo??.jpg --siappendall --attrib FramesPerSecond 10.0 -o anim.gif
```

That looks the same to me? Assuming that there are files matching foo??.jpg

Is this a Windows command line weirdness thing?

jessey-git:
I see the issue. That example depends on shell glob expansion, rather than say the OIIO wildcard stuff. I went directly to foo_#.png since ? didn't seem to work... Here's an example:
```
.\oiiotool.exe --debug -v t:/render/foo_????.png --siappendall --attrib FramesPerSecond 10.0 -o t:/anim-10fps.gif
oiiotool ERROR: read : File does not exist: "t:/render/foo_????.png"
Full command line was:
> oiiotool.exe --debug -v t:/render/foo_????.png --siappendall --attrib FramesPerSecond 10.0 -o t:/anim-10fps.gif
```

So in reality, the command line is actually: `oiiotool  <list of each file individually> --siappendall --attrib FramesPerSecond 10.0 -o anim.gif`  The following does work:
```
oiiotool.exe T:\render\foo_0001.png T:\render\foo_0002.png T:\render\foo_0003.png T:\render\foo_0004.png --siappendall --attrib FramesPerSecond 10.0 -o t:/anim-10fps.gif
```

jessey-git (approved):

lgritz:
I'm still amazed that in 2025, Windows shells aren't expected to do globbing.

Should we be making oiiotool do a glob of its command line arguments (on Windows), just to simplify being able to give example commands that will work everywhere?

PR #4891, opened 2025-09-16
api: Add a bits optional parameter to ImageSpec::set_format

lgritz:


lgritz:
Closing. This is flawed. Will re-submit when I have it worked out.

PR #4892, opened 2025-09-16
ci: Fix testsuite/heif ref output

lgritz:
PR #4880 checked in some output reference that contained version numbers, which passed in main but failed when backported to 3.1. This repairs it and makes the output suitable for any version that contains the patch.


lgritz:
Because this is causing CI to fail in 3.1 (or at least, preventing me from backporting patches and cutting a beta2), I intend to merge this, preemptively if I have to, as soon as it fully passes CI.

brechtvl (approved):

PR #4894, opened 2025-09-17
build: fix openjph target use

### LINKED ISSUE #4893: [BUILD] OpenJPH feature build issues
https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/333b63bac64b8280be4d376f4eda5a68dc5372f7/src/jpeg2000.imageio/CMakeLists.txt#L11-L16

is missing from the 3.0.10.1 tag so the feature is impossible to enable.

Even w/ that fixed, this recent change also has a side effect:

https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/v3.0.10.1/src/cmake/externalpackages.cmake#L179

`openjph-config.cmake` unfortunately does not set `OPENJPH_LIBRARIES` & friends for 'old-style' target linking, like it was the case w/ the custom local module. Only the _imported_ `openjph` target is available. Not even `OPENJPH_FOUND` is set, but `openjph_FOUND`.

I think you'll need to manually pull these out using `get_target_properties()` as currently the snippet above does nothing on the main branch.

Alternative: restore the local find module, but inside it look in config mode first and set variables. Then you also have a chance to fall back to older versions not shipping the CMake config.

lgritz:
PR #4875, switched to using openjph's exported cmake config instead of
our own FindOpenJPH.cmake, and in the process also renamed OpenJPH ->
openjph to follow their convention.

But I botched it, still using the old OPENJPH_LIBRARIES (etc.)
variables instead of fully switching to the correct targets exported
from their config.

This PR minimally fixes that error, re-enabling use of openjph.

Fixes #4893

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -278,7 +278,7 @@ macro (oiio_add_all_tests)
                     FOUNDVAR OPENJPEG_FOUND
                     IMAGEDIR oiio-images URL "Recent checkout of OpenImageIO-images")
     oiio_add_tests (htj2k
-                    FOUNDVAR OPENJPH_FOUND
+                    FOUNDVAR openjph_FOUND
                     IMAGEDIR oiio-images URL "Recent checkout of OpenImageIO-images")
     oiio_add_tests (jpeg2000-j2kp4files
                     FOUNDVAR OPENJPEG_FOUND
--- a/src/jpeg2000.imageio/CMakeLists.txt
+++ b/src/jpeg2000.imageio/CMakeLists.txt
@@ -8,10 +8,7 @@ if (OPENJPEG_FOUND)
     set(_jpeg2000_libs ${OPENJPEG_LIBRARIES})
     set(_jpeg2000_defs "USE_OPENJPEG")
 
-    if (OPENJPH_FOUND)
-        list(APPEND _jpeg2000_includes ${OPENJPH_INCLUDES})
-        list(APPEND _jpeg2000_lib_dirs ${OPENJPH_LIBRARY_DIRS})
-        list(APPEND _jpeg2000_libs ${OPENJPH_LIBRARIES})
+    if (openjph_FOUND)
         list(APPEND _jpeg2000_defs "USE_OPENJPH")
     endif()
 
@@ -19,6 +16,7 @@ if (OPENJPEG_FOUND)
         INCLUDE_DIRS ${_jpeg2000_includes}
         LINK_DIRECTORIES ${_jpeg2000_lib_dirs}
         LINK_LIBRARIES ${_jpeg2000_libs}
+                       $<TARGET_NAME_IF_EXISTS:openjph>
         DEFINITIONS ${_jpeg2000_defs}
     )
 else()


lgritz:
@kmilos Does this look right to you?

lgritz:
Anybody care to review this? It's the last piece I need to release a 3.1 beta2.

src/cmake/modules/FindOpenJPEG.cmake
```diff
@@ -174,6 +174,15 @@ if (OPENJPEG_FOUND)
   foreach (tmplib ${OpenJpeg_libvars})
     list (APPEND OPENJPEG_LIBRARIES ${${tmplib}})
   endforeach ()
+
+  if (NOT TARGET openjp2)
+      add_library(openjp2 UNKNOWN IMPORTED)
+      set_target_properties(openjp2 PROPERTIES
+          INTERFACE_INCLUDE_DIRECTORIES "${OPENJPEG_INCLUDES}")
+      set_property(TARGET openjp2 APPEND PROPERTY
+                   IMPORTED_LOCATION "${OPENJPEG_LIBRARIES}")
```
kmilos: I am not 100% sure this is sufficient - e.g. my `OpenJPEGTargets-release.cmake` config on Windows also has an `IMPORTED_IMPLIB` property for the stub to go w/ `IMPORTED_LOCATION` DLL...

I think it might be safer to do the minimal change just for the openjph target first (maybe [something like I suggested](https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4893#issuecomment-3303772262)), and only then refactor this later. This FindOpenJPEG module feels like overkill as OpenJPEG (at least v2) also ships CMake config files for a long time, maybe it should be checked first before setting stuff up manually...

lgritz: I should check if it's still the case, but last time I looked at using the OpenJPEG's own exported cmake configs, I think that they worked ok for the most recent version, but didn't exist or weren't reliable for some of the older versions that we still supported.

I can back off this a bit and propose a simpler, more direct fix, that doesn't mess with the OpenJPEG portions.

But before I commit to that, did you try this patch on your end and verify that there's something that doesn't work?

kmilos: > I can back off this a bit and propose a simpler, more direct fix, that doesn't mess with the OpenJPEG portions.

I would agree it's better to separate these.

> But before I commit to that, did you try this patch on your end and verify that there's something that doesn't work?

I haven't had the chance, sorry...

lgritz:
I've fully replaced this PR with a simpler version that just directly addresses the problem:

---

build: fix openjph target use

PR 4875, switched to using openjph's exported cmake config instead of
our own FindOpenJPH.cmake, and in the process also renamed OpenJPH ->
openjph to follow their convention.

But I botched it, still using the old OPENJPH_LIBRARIES (etc.)
variables instead of fully switching to the correct targets exported
from their config.

This PR minimally fixes that error, re-enabling use of openjph.

kmilos:
LGTM

PR #4898, opened 2025-09-20
ci: require all dependencies, with explicit exceptions

lgritz:
We recently had a problem where our use of a dependency was broken, and it had gone unnoticed because none of our CI jobs actually had the dependency present and tested it. This can easily happen because so many of our dependencies are "optional" -- our build will just give a warning and silently disable the functionality that would have been supported by the missing dependency.

We have long had build-time options `OpenImageIO_REQUIRED_DEPS`, to consider specific dependencies (or "all") required even if they would ordinarily be optional, and `OpenImageIO_OPTIONAL_DEPS`, to make exceptions. But we didn't use these in CI.

So this PR makes sets `OpenImageIO_REQUIRED_DEPS=all` to make all dependencies ostensibly required, and then list all exceptions explicitly. This should make it much more difficult in the future to make a mistake where use of a dependency is completely untested in our CI without our being aware of it. And it gives us a visible "hit list" of untested or under-tested dependencies to slowly whittle down.

Some changes that came long for the ride:

* checked_find_package: explicitly disabled packages are treated as optional.

* cuda_macros.cmake: Only look for CUDA on platforms that might conceivably have it (i.e., don't even look on Mac).

* Separate linux-aswf from linux-ubuntu into separate job groups, because that makes it easier to have shared commonalities of which dependencies they test.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -68,6 +68,10 @@ on:
         type: string
       nametag:
         type: string
+      required_deps:
+        type: string
+      optional_deps:
+        type: string
     secrets:
         PASSED_GITHUB_TOKEN:
           required: false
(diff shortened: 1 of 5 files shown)


lgritz:
Anybody have comments about this?

EmilDohne (approved):

PR #4902, opened 2025-09-23
fix: sonar scan guard

vvalderrv:
We recreated the SonarCloud project to clear the 404s. This commit fixes the workflow condition by comparing to the string '1', so the Sonar scan step runs and uploads results.

## Description
Updates the GHA workflow guard so the Sonar scan actually executes:
- Change `if: inputs.sonar == 1` → `if: inputs.sonar == '1'` because `inputs.sonar` is a string (from `workflow_call` inputs).

## Tests
No unit tests added (CI-only change). Validation: run the “Analysis” workflow and confirm SonarCloud receives results.

## Checklist:
- [ ] I have ensured that the change is tested somewhere in the testsuite.
- [ ] If I added or modified a C++ API call, I have also amended the corresponding Python bindings.

### DIFF
--- a/.github/workflows/build-steps.yml
+++ b/.github/workflows/build-steps.yml
@@ -177,7 +177,7 @@ jobs:
         if: inputs.coverage == '1'
         run: src/build-scripts/ci-coverage.bash
       - name: Sonar-scanner
-        if: inputs.sonar == 1
+        if: inputs.sonar == '1'
         env:
           GITHUB_TOKEN: ${{ secrets.PASSED_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


lgritz (commented):
I'm not sure why this matters. In the analysis.yml, it's clearly passed as `1`, not `'1'`, so isn't an integer comparison going to be fine?

lgritz:
I'm not sure this is necessary. It's been like this for a long time, and the sonar step runs just fine when it's supposed to. (It was crashing in the sonar code itself! But it was not failing to run this step of the workflow.)

lgritz (approved):
LGTM, even though I don't think it's necessary, its probably correct?

PR #4903, opened 2025-09-25
build: fix some build issues encountered on a musl libc system

omcaif:
## Description

Two small fixes to  compile on musl libc systems.

The first commit is about the use of `__WORDSIZE`, which is not a standard define and is defined another header in musl libc. I decided to fix this by following what was done in `libdnf` in [this commit](https://github.com/rpm-software-management/libdnf/pull/1159/commits/82a07dbe6e53fccfe3a8f8e0d951618e0aa72999). However, there are other ways, such as including `bits/reg.h` or `bits/user.h`, which define `__WORDSIZE` in musl libc, or using `__LP64__` to check for 64 bit instead (inspired by [Alpine Linux's patch](https://gitlab.alpinelinux.org/alpine/aports/-/blob/0f71e9504b59049e0ec1abb486ba455ab2650c83/community/openimageio/0001-fix-compile-error.patch)):

```c
#if defined(__GNUC__) && (__WORDSIZE == 64 || (defined(__LP64__) && __LP64__)) && !(defined(__APPLE__) && defined(__MACH__)) || defined(__NetBSD__)
```

The second commit is about the use of the `feenableexcept` function. This function is only defined by glibc. The original code assumed all Linux systems use glibc (checking `__linux__` define), so I just changed it to check for glibc directly (by checking the `__GLIBC__` define).

## Tests

I have only tested it on my musl libc system. I think just compiling it on different systems is enough of a test in this case...

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imageinout_test.cpp
+++ b/src/libOpenImageIO/imageinout_test.cpp
@@ -537,7 +537,7 @@ main(int argc, char* argv[])
     getargs(argc, argv);
 
     if (enable_fpe) {
-#if defined(__linux__)
+#if defined(__GLIBC__)
         fprintf(stderr, "Enable floating point exceptions.\n");
         feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
 #else


src/include/OpenImageIO/typedesc.h
```diff
@@ -409,7 +417,7 @@ template<> struct BaseTypeFromC<uint64_t> { static constexpr TypeDesc::BASETYPE
 template<> struct BaseTypeFromC<const uint64_t> { static constexpr TypeDesc::BASETYPE value = TypeDesc::UINT64; };
 template<> struct BaseTypeFromC<int64_t> { static constexpr TypeDesc::BASETYPE value = TypeDesc::INT64; };
 template<> struct BaseTypeFromC<const int64_t> { static constexpr TypeDesc::BASETYPE value = TypeDesc::INT64; };
-#if defined(__GNUC__) && __WORDSIZE == 64 && !(defined(__APPLE__) && defined(__MACH__)) || defined(__NetBSD__)
+#if defined(__GNUC__) && IS_64_BIT && !(defined(__APPLE__) && defined(__MACH__)) || defined(__NetBSD__)
```
lgritz: Since this is the only place it's used, how about just changing to

```suggestion
#if defined(__GNUC__) && (ULONG_MAX == 0xffffffffffffffff) && !(defined(__APPLE__) && defined(__MACH__)) || defined(__NetBSD__)
```

and then you don't actually need to define IS_64_BIT?

lgritz:
Is this intended as a DevDays project?

src/include/OpenImageIO/typedesc.h
omcaif: I'm okay with this change.

I was worried that the word size might be something like 128 bits in the future (hence the explicit check for both 64 bit and 32 bit, then erroring out otherwise), but now that I think about it, the compile will fail anyway in that case.

omcaif:
> Is this intended as a DevDays project?

No.

By the way, I am not sure how to respond on suggested changes. I just looked through the [contributing guidelines](CONTRIBUTING.md) and did not see anything about it.

Should I just use the GitHub UI to "sign off and commit suggestion"? Rebase? Add another commit manually to incorporate the change? Do I need to add a "Co-Author" tag?

Anyway, I have allowed maintainers to edit this PR, so feel free to make any edits you see fit.

lgritz:
> > Is this intended as a DevDays project?
> 
> No.

No problem either way, I just was trying to figure out whether or not to tag it as such.


> Should I just use the GitHub UI to "sign off and commit suggestion"? Rebase? Add another commit manually to incorporate the change? Do I need to add a "Co-Author" tag?
> 
> Anyway, I have allowed maintainers to edit this PR, so feel free to make any edits you see fit.

Any of the above. You can click and sign off on the change. Or you can do it yourself and push the additional commit. (Just make sure EVERY commit is done with `-s` or otherwise contains the DCO signoff in the commit comments.) You do not need to rebase again, for such a small change, when we expect that nothing has changed in main since you posted the PR that could make it a nontrivial merge or would interfere with the correctness of your patch.

You do not need a co-author tag for something this trivial.

omcaif:
Okay, thanks. I think I updated the commit now to address your comment.

lgritz (approved):
LGTM, thanks for the fixes

Issue #4649, opened 2025-02-24
JPEG-XL reads ICC profile but doesn't put it in the spec, doesn't write


lgritz:
Our JPEG-XL reader uses the jxl API to extract the ICC profile... but then it doesn't call decode_icc_profile to enter the information as attributes in the ImageSpec.

Also, the JPEG-XL writer doesn't write ICC profiles, even if one is in the ImageSpec supplied.

ManishSharma1609:
Hi @lgritz  I came across this issue and would like to work on resolving it.

lgritz:
Yes, please!

shanesmith-dwa:
Hi @lgritz, does this issue still needs work? I'm looking into it for dev days.

lgritz:
Yes, it looks like it's still an outstanding issue.

lgritz:
This was fixed by #4905

PR #4905, opened 2025-09-26
feat(jpeg-xl): ICC read and write for JPEG-XL files (issue 4649)

shanesmith-dwa:
## Description

This PR addresses [4649](https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4649) implementing ICC read and write support for JPEG-XL.

Fixes #4969

## Tests

The oiiotool-attribs testsuite was updated and tested.


lgritz (commented):
This looks great! Let's see how it comes out in the CI, but if that all passes I think it's good to go.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -538,6 +539,21 @@ JxlOutput::save_image(const void* data)
...
+    if (icc_profile_parameter != NULL) {
+        unsigned char* icc_profile
+            = (unsigned char*)icc_profile_parameter->data();
+        uint32_t length = icc_profile_parameter->type().size();
+        if (icc_profile && length) {
+            if (JXL_ENC_SUCCESS 
+                == JxlEncoderSetICCProfile(m_encoder.get(),
+                                           icc_profile,
+                                           length)) {}
+                errorfmt("JxlEncoderSetICCProfile failed\n");
```
lgritz: Is this correct? Did you mean to have an `else` clause that contains the errorfmt? Or should it be `JXL_ENC_SUCCESS != ...` and remove the `{}`?

src/jpegxl.imageio/jxlinput.cpp
```diff
@@ -21,12 +21,15 @@
...
 #include <OpenImageIO/tiffutils.h>
 
+#include <jxl/color_encoding.h>
 #include <jxl/decode.h>
 #include <jxl/decode_cxx.h>
 #include <jxl/resizable_parallel_runner_cxx.h>
 
 OIIO_PLUGIN_NAMESPACE_BEGIN
 
+#define ICC_PROFILE_ATTR "ICCProfile"
```
jessey-git: Is the define here, and the duplicated one in jxloutput, benefitting anything? I know TIFF happened to define it like this but generally we don't go through this ceremony for other attributes.

src/jpegxl.imageio/jxloutput.cpp
```diff
@@ -538,6 +539,21 @@ JxlOutput::save_image(const void* data)
         return false;
     }
 
+    // Write the ICC profile, if available
+    const ParamValue* icc_profile_parameter = m_spec.find_attribute(
+        ICC_PROFILE_ATTR);
+    if (icc_profile_parameter != NULL) {
```
jessey-git: nit: prefer to use `nullptr` in all new code and to match existing style already used in the file.

lgritz:
Your code looks great (though you can see there is a very minor formatting change needed to pass the clang-format test).

I think the reason so many CI tests are failing is that libjxl is an optional dependency -- if it's found, jxl support will be enabled, but if the library is not found, its not a hard error, it just disables jxl support. 

Some of the jobs in our test matrix have (or build) libjxl, but others do not. But the test cases you added are unconditional -- so of course they are not passing for the CI jobs where jxl is not supported by the build configuration.

I think the way to proceed is to make a NEW test, jxl, with just these commands pertaining to jxl. In testing.cmake, add that test but specify the `FOUNDVAR JXL_FOUND` (you can see how similar are done for the other format related tests). Does that make sense?

shanesmith-dwa:
Thanks @lgritz, that was helpful. I think I have made the necessary changes.

src/cmake/testing.cmake
```diff
@@ -284,6 +284,9 @@ macro (oiio_add_all_tests)
                     FOUNDVAR OPENJPEG_FOUND
                     IMAGEDIR j2kp4files_v1_5
                     URL http://www.itu.int/net/ITU-T/sigdb/speimage/ImageForm-s.aspx?val=10100803)
+    oiio_add_tests (jxl
+                    FOUNDVAR JXL_FOUND
+                    IMAGEDIR oiio-images URL "Recent checkout of OpenImageIO-images")
```
lgritz: I don't think this test is using anything in oiio-images, is it?

```suggestion
                    FOUNDVAR JXL_FOUND)
```

testsuite/jxl/run.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+# Test oiiotool's ability to add and delete attributes
+
```
lgritz: Can eliminate this line that was copied from a different test.

```suggestion
```

lgritz (changes requested):
I only have two extremely minor comments. If you agree with them, you should be able to apply my suggestions by clicking the buttons.

src/cmake/testing.cmake
shanesmith-dwa: It is using tahoe-tiny.tif from oiio-images

lgritz: Your run.py file says

```
command += oiiotool ("../common/tahoe-tiny.tif --iccread ref/test-jxl.icc -o tahoe-icc.jxl")
```

It is getting tahoe-tiny from testsuite/common, not from oiio-images. No?

shanesmith-dwa: True. I had thought it was from oiio-images

lgritz (approved):
LGTM!

PR #4907, opened 2025-09-26
feat(openexr): ACES Container hint for exr outputs

Glowies:
Implements #4791 

## Description
This PR introduces the `oiio:ACESContainer` hint for OpenEXR outputs.
The hint can take one of the following values: `none` (default), `strict`, or `relaxed`. If not `none`, the spec will be checked to see if it is compliant with the ACES Container format defined in [ST 2065-4](https://pub.smpte.org/doc/st2065-4/20130510-pub/st2065-4-2013.pdf). If it is, chromaticities will be set to the ACES AP0 ones, and the acesImageContainerFlag attribute will be set to 1. In `strict` mode, if the spec is non-compliant, the output will throw an error and avoid writing the image. While in `relaxed` mode, if the spec in non-compliant, only a warning will be printed and the attributes mentioned above will not be written to the spec.

## Tests
I've added several tests under `openexr-suite`. These use oiiotool to attempt writing several EXRs with the `oiio:ACESContainer` hint. 

## Checklist:

  doesn't change the API.)
  nonconforming.


src/doc/builtinplugins.rst
```diff
@@ -1643,6 +1643,18 @@ control aspects of the writing itself:
    * - Output Configuration Attribute
      - Type
      - Meaning
+   * - ``oiio:ACESContainer``
```
lgritz: Seeking other opinions, but I would call it "openexr:ACESContainer" since it is specific to the openexr plugin.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,71 @@ set_exr_threads()
 
 
 
+constexpr float ACES_AP0_chromaticities[8] = {
```
lgritz: static? Does constexpr also imply that?

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,71 @@ set_exr_threads()
...
+
+bool
+is_aces_container_compliant(const OIIO::ImageSpec& spec)
+{
+    // Check channels
+    std::vector<std::string> allowed_sets
+        = { "B,G,R",
+            "A,B,G,R"
+            "B,G,R,left.B,left.G,left.R",
+            "A,B,G,R,left.A,left.B,left.G,left.R" };
```
lgritz: OIIO conventions are that on the app side, the channel list always starts with R, G, B, A, in that order (if they exist), and the reader will rearrange channels as required by the file format. So EXROutput will never receive channels in B,G,R or A,B,G,R order.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,71 @@ set_exr_threads()
...
+
+
+
+bool
+process_aces_container(OIIO::ImageSpec& spec, std::string mode)
+{
+    bool is_compliant = is_aces_container_compliant(spec);
+
+    if (!is_compliant) {
+        std::cerr << "WARNING: Image spec is not ACES Container compliant\n";
```
lgritz: It's usually not desirable for a library to write to stderr (or stdout), since you have no idea what kind of app you're being called from. (For example, for Windows GUI apps, I think stderr goes nowhere.) We prefer to do all error reporting through calls to ImageInput::errorfmt() so that the calling app can retrieve them and respond in a way that's idiomatic for that app.

Unfortunately, we don't have a way currently to report warnings. We'll probably add that in the future, but for now, report the errors (the official way) and just leave a comment specifying where warning should happen, in the future when they are supported.

src/openexr.imageio/exroutput.cpp
```diff
@@ -812,6 +877,20 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
...
+    // Check ACES Container hint
+    std::string aces_mode = spec.get_string_attribute("oiio:ACESContainer",
+                                                      "none");
+    if (aces_mode == "strict" || aces_mode == "relaxed") {
+        bool should_panic = !process_aces_container(spec, aces_mode);
+
+        if (should_panic) {
+            const char* non_compliance_message
+                = "Cannot output non-compliant ACES Container in 'strict' mode.";
+            errorfmt(non_compliance_message);
```
lgritz: This is fine, but curious why you store in a variable instead of just passing the message to errorfmt?

testsuite/openexr-suite/run.py
```diff
@@ -50,3 +50,20 @@
 # Check writing overscan and negative range
 command += oiiotool("--create 64x64-16-16 3 -d half -o negoverscan.exr")
 command += info_command("negoverscan.exr", safematch=True)
+
+# Check ACES Container output for relaxed mode
+command += oiiotool("--create 4x4 3 -d half --compression none --ch B,G,R -sattrib oiio:ACESContainer relaxed -o relaxed-out.exr")
```
lgritz: I don't think you need to test B,G,R. No matter what order the channels are in a file, the exr reader will rearrange them to appear to be R,G,B,A first (if they exist), and no matter what order you start with, openexr itself will write it to the file in alphabetical order (this is an absurd thing for a file format to do, don't get me started). 

So I believe people understand that using --ch to reorder channels (without also renaming them) has no effect on exr files, and breaks just about everything else in OIIO (which assumes RGB without checking), so they don't do it. (But reordering and also renaming is a different matter, like "--ch R=B,G,B=R" actually does something useful.)

lgritz (commented):
Other than my fairly minor comments about attribute name and unnecessary work related to channel order, etc, this all looks right to me. I'll ask Anton for a review as well.

lgritz:
Would appreciate a review from @antond-weta as the person who filed the issue this is based on (as well as being more knowledgeable than me about ACES containers.)

lgritz:
@Glowies would you mind changing the description to say specifically 

    Closes #4907

or

    Fixes #4907 

those are the magic words, and I think anything else doesn't trigger GitHub to link the issue to be automatically closed by the merge of the PR.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,71 @@ set_exr_threads()
...
+    if (!is_compliant) {
+        std::cerr << "WARNING: Image spec is not ACES Container compliant\n";
+
+        // early out and return true iff in "relaxed" mode
+        // to indicate that the output can continue without
+        // throwing an error
+        return mode == "relaxed";
+    }
+
+    spec.attribute("chromaticities", OIIO::TypeDesc(OIIO::TypeDesc::FLOAT, 8),
```
antond-weta: I believe the chromaticities should also be set in the relaxed mode. 
In my original proposal the meaning of "relaxed" basically was: this file is not fully compliant, but still contains ACES AP0 data (the only hard requirement).

With the current implementation there is no difference in functionality between "none" and "relaxed", apart from the warning.

If the user knows the file won't be compliant (say, they need 32-bit floats written out), they will have to set the chromaticities manually, so there is not much point to pass "relaxed" at all.

antond-weta (commented):
Looks good!
I've raised a point about setting the chromaticities in the "relaxed" mode automatically.

src/doc/builtinplugins.rst
Glowies: Ah makes sense. I initially thought this could be extended to support other file types in the future, but it turns out that the ACES Container format is restricted to OpenEXR anyway!

src/openexr.imageio/exroutput.cpp
Glowies: As far as I can gather from a quick search online, `constexpr` implies `static` in most scenarios; however, it can still be beneficial to specify `static` so that the compiler has more information to go off of. I'll add the keyword here just to be safe.

JGoldstone:
The documentation reference should be changed to https://pub.smpte.org/pub/st2065-4/st2065-4-2023.pdf, as minor changes were made to the spec and the publication was updated.

On p. 5 there's a list of changes, and since they are short enough I'll just list them here:

- The definitions and usages of string types were clarified.
- The lensFirmwareVersion attribute was added.
- The usage of the aperture attribute was clarified.
- The definition of the timecodeRate attribute was clarified.
- The recommended supported size of data window to readers was increased.

Nothing in this changes the definition of an ACES container.

In the future (there is a five-year review of this document, only a couple years away), if the recent addition to OpenEXR of HTJ2K, a compression type that is **normatively specified** seems to be beneficial to the community, I could see that one particular compression type being allowed in ACES containers; I would lobby for it (if people are using it). But for now, compression remains disallowed, not because it's undesirable, but because ST 2065-4 is a normative international standard, and can only normatively reference other normatively specified international standards. At one point I did suggest a Google Summer of Code project to produce normative-quality documentation on piz, the DWA formats, etc. etc. and the output of which could be taken to SMPTE, but it was pointed out the phrase is "Summer of Code", not "Summer of Code and Doc".

Anyway, please update the reference to the more recent version of the normative definition of the ACES container.

lgritz:
@Glowies There are a variety of (fairly minor) suggestions for changes. I think that once you fix those few things, we'll be ready to merge this.

Glowies:
> @Glowies There are a variety of (fairly minor) suggestions for changes. I think that once you fix those few things, we'll be ready to merge this.
> 
> 

Yep, thank you for the suggestions everyone! Didn't get a chance over the weekend but I'll be addressing them tomorrow.

antond-weta:
I have another suggestion, sorry!

If the ImageBuf already contained the `acesImageContainerFlag` set, the writer should treat it the same way as if it had the `oiio:ACESContainer` hint set to "strict". 

Here is an example of a use case when this might be important:
Using oiitool to read an AcesContainer file presumably retains the `acesImageContainerFlag`, if the image is then modified in a way that it's no longer compliant with the AcesContainer standard, attempting to save as an exr should make the writer fail, unless the flag was explicitly removed by the user. 

This does change the existing behaviour, so I'm not sure if we want to make this change.

lgritz:
> If the ImageBuf already contained the `acesImageContainerFlag` set, the writer should treat it the same way as if it had the `oiio:ACESContainer` hint set to "strict".

Same as strict, or same as relaxed?

And now I'm wondering about inevitable confusion between the `acesImageContainerFlag` attribute and the `openexr:ACESContainer` hint. 

Maybe the presence of `acesImageContainerFlag` should be the one and only request to write THIS image to a file as an ACES container, and there is a separate (self-documentingly named) hint about the policy for how strict to be? And at that point, we should consider if the policy hint should be global (set by `OIIO:attribute()`) or per-file (set as a `config` hint)?

antond-weta:
> Same as strict, or same as relaxed?
Either/or. Rephrasing my comment to be more generic: in the absence of any hints, if the ImageBuf contains the acesImageContainerFlag, but is otherwise noncompliant, the writer should either fail, or remove the flag. The question is what behaviour we prefer. I generally prefer the tools not to make any decisions implicitly (despite wanting the proposed hint as an instruction to set the ACES chromaticities implicitly), so I would prefer "strict".

A similar check could be added regarding the colour space: if the ImageBuf contains colour space metadata, and it is different from AP0, the writer should fail in either mode. I have zero experience working with colour spaces in OIIO, so I actually don't have a strong opinion on this.

lgritz:
Let's be totally concrete about what behavior we want.

What specific combinations of 

* (a) metadata in the spec
* (b) config hints to EXRImage::open()
* (c) global OIIO::attribute() settings

do we want to trigger each of the following behaviors:

1. Non-ACES-container regular exr files
2. Fully compliant ACES container file
3. Error, no file written

And ensure that all combinations of a-c must deterministically result in one of behaviors 1-3.

antond-weta:
Here is my take:
```
bool global_aces_container_flag = get_global_aces_container_flag();
string aces_container_hint = get_aces_container_hint();
bool has_local_aces_container_flag = has_attribute("acesImageContainerFlag");
int local_aces_container_flag = get_attribute("acesImageContainerFlag", 0);

bool treat_as_aces_container = 
    (global_aces_container_flag == true) ||
    (aces_container_hint == "strict") ||
    (local_aces_container_flag == 1);

bool compliant = 
    check_compression_is_compliant() &&
    check_channels_are_compliant() &&
    check_chromaticities_missing_or_equal(AP0) &&
    check_colourspace_missing_or_equal(ACES2065-1) &&
    check_acesImageContainerFlag_missing_or_equal(1) &&
    check_colorInteropID_missing_or_equal("lin_ap0_scene");

if (treat_as_aces_container) {
    if (!compliant) error();
    else {
        if (has_local_aces_container_flag == false)
            set_attribute("acesImageContainerFlag", 1);
        set_chromaticities(AP0);
        write_file();
    }
}
else if (aces_container_hint == "relaxed") {
    if (compliant) {
        if (has_local_aces_container_flag == false)
            set_attribute("acesImageContainerFlag", 1);
        set_chromaticities(AP0);
        write_file();
    }
    else {
        if (has_local_aces_container_flag)
            remove_attribute("acesImageContainerFlag");
        set_chromaticities(AP0);
        set_attribute("colorInteropID", "lin_ap0_scene");
        issue_warnings();
        write_file();
    }
}
else {
    write_file();
}
```

JGoldstone:
Oh, one other thing I remember us clarifying in the recent 2065-4 revision: most but for some reason not all string attributes were prohibited from being blank. E.g. you can't have a lensMake that is "". That's something to add to the list of things to be checked

src/openexr.imageio/exroutput.cpp
Glowies: Ooh interesting, so should I be checking for the existence of these channel names instead of the exact order of them?

src/openexr.imageio/exroutput.cpp
Glowies: This was just left over from a previous commit where I was using the variable for both printing a warning and passing to errorfmt. I'll clean this up!

src/openexr.imageio/exroutput.cpp
lgritz: I'm saying that there is an exact order, but it's different than you wrote. It will be R,G,B or R,G,B,A or R,G,B,left.R,left.G,left.B or R,G,B,A,left.R,left.G,left.B,left.A. It won't be any other order once it passes through OIIO.

Glowies: I guess I'm a little confused about why the current tests are passing then. The oiiotool commands that I run with --ch B,G,R pass the compliance check, while the --ch R,G,B ones do not. Is it just that the reordering of the channels into R,G,B happens after the `spec_to_header` function?

testsuite/openexr-suite/run.py
lgritz: Well, now that I see your question, maybe the test should be the presence of those channels and no others, in any order, as  you suggested earlier? Yes, that sounds better to protect against these surprises, in case somebody does an unwise `--ch`.

Tedious explanation follows.

The things to remember are:
* OIIO readers will always rearrange channels to start with R,G,B,A if present. (And for what OpenEXR calls "layers", will rearrange within each layer as well.)
* LOTS of OIIO functionality simply assumes RGB[A] order, so people generally know not to rearrange channels (like --ch B,G,R) without also renaming them.
* OpenEXR demands alphabetical order, so no matter order channels are, the OpenEXR library (not really us) will sort them and end up writing A,B,G,R.

So now we can reason about a couple interesting phenomena.

**Why does `oiiotool in.exr --ch B,G,R -o out.exr` do nothing at all?** 

Because the --ch reorders the channels (without renaming them), and then -o will (thanks, openexr) write them in sorted order.

I assure you that `oiiotool in.exr --ch B,G,R --mulc 1,1,0 -o out.exr` will make the red channel black, not blue. But `oiiotool in.exr --ch R=B,G,B=R --mulc 1,1,0 -o out.exr` will make the blue channel black.

**Why does `oiiotool in.exr --ch B,G,R -o out.tif` swap the red and blue channels?**

Because TIFF files don't store channel names! The TIFF writer gets the channels in the order B,G,R and just write them in that order. There are no channel names to straighten it out, other than the order itself.

Glowies: Ah thank you for the explanation! This cleared things up for me. As you recommended, I've changed the implementation to just check for the existence of the correct channel names instead of checking for their exact order.

lgritz:
The CI failures that say something about ffmpeg not found are not related to this PR, just ignore them.

Glowies:
Hey @antond-weta!  I was reading through your pseudo-code and I think I will just follow what you're doing there. I was thinking about @lgritz's recommendation for a global flag and it sounds to me like the 'policy' attribute might be more suited as a global attribute. I would probably change the start of your pseudo-code to be this:
```
string global_aces_container_hint = get_global_aces_container_hint();
bool has_local_aces_container_flag = has_attribute("acesImageContainerFlag");
int local_aces_container_flag = get_attribute("acesImageContainerFlag", 0);
```
So I would use `global_aces_container_hint` instead of `aces_container_hint` and I would only rely on the existence of the `acesImageContainerFlag` to determine if the image should be treated as an ACES container.
Does that sound like a good approach to you?

antond-weta:
I'm not sure! Probably that would be enough, but you'll have to answer the Larry's question I was trying to address with my pseudo-code.

In my pseudo-code the meaning of the `global_aces_container_flag` was "treat every exr file as an AcesContainer", and the meaning of the local hint was "treat this file as an AcesContainer, either enforcing full compliance, or allowing some slack". So setting the global flag to "true" has the same effect as passing "oiio:ACESContainer" = "strict" to every writer. Note, with that logic the global setting just overrides one specific value of the local hint, we may want to decide proceeding with just adding the local hint for now, and adding the global setting later. 

If I understand the change you proposed correctly, you have the global hint meaning "treat all exr files as an AcesContainer, either enforcing full compliance, or allowing some slack", and set the "acesImageContainerFlag" to true on the images you want to be written as AcesContainer. But here is the catch, you are only supposed to set that flag on the images which are fully compliant, so the user will have to perform all the checks and fill in all required metadata, doing all the leg work the proposed hint was supposed to take care of; and the hint boils down to "on/off" state anyway. I can't see how the "relaxed" path would look like.

antond-weta:
Also note that if we are going to add a global setting, we should also think about what effect it should have on the exr reader as well. We may want to have two separate settings for reading and writing, which would be required in order to being able to convert between AcesContainer and regular exr files.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,86 @@ set_exr_threads()
...
+    0.0001f,  -0.077f,  // blue
+    0.32168f, 0.33767f  // white
+};
+
+
+
+bool
+is_spec_aces_container_channels_only(const OIIO::ImageSpec& spec)
+{
+    // Allowed channel sets
```
lgritz, line 309:
```suggestion
    // Note: this is constructing and comparing sets, so that channel order
    // doesn't matter.

    // Allowed channel sets
```

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,86 @@ set_exr_threads()
...
+    std::vector<std::set<std::string>> allowed_sets
```
lgritz: Can this be made `static const` so that these sets we're comparing are not constructed and destructed every time?

testsuite/openexr-suite/ref/out.txt
```diff
@@ -341,3 +341,30 @@ negoverscan.exr      :   64 x   64, 3 channel, half openexr
     screenWindowWidth: 1
     oiio:subimages: 1
     openexr:lineOrder: "increasingY"
+1
+
```
lgritz: This is weird. What's this `1` doing alone on this line?

lgritz: Oh, I think I understand this now. Will comment separately in the code where it happens.

testsuite/openexr-suite/run.py
```diff
@@ -50,3 +50,36 @@
 # Check writing overscan and negative range
 command += oiiotool("--create 64x64-16-16 3 -d half -o negoverscan.exr")
 command += info_command("negoverscan.exr", safematch=True)
+
+# Check ACES Container output for relaxed mode
+#
+# Valid ACES Container
+command += oiiotool("--create 4x4 3 -d half --compression none -sattrib openexr:ACESContainer relaxed -o relaxed-out.exr")
+command += oiiotool("relaxed-out.exr --echo {TOP[acesImageContainerFlag]}", failureok=True) # should give 1
```
lgritz: It's weird to have a `1` or just a blank line in the output with no explanation. Hard to tell what it's testing or what the output is supposed to be. I think we can make the output more self-documenting like this:

```suggestion
command += oiiotool("relaxed-out.exr --echo {TOP.filename} acesContainerFlag={TOP[acesImageContainerFlag]}", failureok=True) # should give 1
```
and similar for the other examples

lgritz:
I think we are experiencing some spec creep here.

In the interest of keeping this an easily implementable and reviewable unit, let's keep THIS PR limited to the following (note: I have added some suggestions I think are helpful):

* A single output hint, "openexr:ACESContainerPolicy" (how about that name?) for OpenEXR files.
* If the output image's ImageSpec doesn't have the metadata `acesImageContainerFlag` present and nonzero, the "openexr:ACESContainerPolicy" flag has no effect because the image is not intended to be an ACES container.
* If the output image's ImageSpec has metadata `acesImageContainerFlag` set to 1, then the user is trying to write an ACES image container. In that case, the new "openexr:ACESContainerPolicy" hint controls one of several behaviors:
    * `"none"` (default) : Write an image file, set the `acesImageContainerFlag` as instructed, hope the user knows what they are doing. It's on them to write a compliant file.
    * `"strict"` : the call to open() FAILS if any of the ACES container requirements are not compliant.
    * `"fix"` : non-compliant metadata or instructions will be FIXED to be compliant if at all possible -- compression will be set to "none", data type will be set to `half`, etc., and metadata that's not allowed in an ACES container will be dropped. If it's just impossible to turn what we were given into an ACES compatible container (for example, it doesn't have the right set of channels), then it's an error like "strict."
    * `"relaxed"` (unimplemented now, reserved for the future) : Unimplemented for now because we don't have a mechanism for warnings. We can decide down the road whether it behaves more like "`none` with warnings for noncompliance" or "`fix` with warnings for noncompliant things that can't be fixed".


IN THE FUTURE, we can consider additional PRs, including global options for things like setting the default policy differently, or requiring all exr outputs to be ACES containers.

antond-weta:
I'd still prefer to rely on the hint for determining whether the user intention was to write an AcesContainer file. When the hint's value is not "none", we are already implying AcesContainer, right?

Relying on acesImageContainerFlag to trigger the logic doesn't sit well with me for a few reasons:
- in the "relaxed" mode the flag needs to be removed if non-compliant. Setting the flag knowing that it will be removed doesn't seem very logical to me
- there is no way to distinguish whether the flag was set as a hint to the writer, or inherited from upstream
- it would be better to have all the checks and filling in the metadata in one place, splitting out acesImageContainerFlag means that the user needs to fill-in the AcesContainer metadata partially (the flag), but rely on the writer to fill in the rest (the chromaticities).

lgritz:
I see, that makes sense. Ultimately, I'm happy to go with you recommendation on all matters ACES container.

If I understand you correctly, you want to rely 100% on the hint for whether or not to write it as an aces container? Which means it needs to be proactively set when a file is being written?

So consider this:

    oiiotool acescontainer.exr -o out.exr

Should out.exr end up with the flag that makes it an ACES container?

antond-weta:
> So consider this:
> 
> oiiotool acescontainer.exr -o out.exr
> Should out.exr end up with the flag that makes it an ACES container?

Yes it should, assuming that `acescontainer.exr` is a valid AcesContainer file. 

In my pseudocode, if we don't want to do the global setting for now, we can just replace the top line with
`bool global_aces_container_flag = false;`
Then the presence of acesImageContainerFlag in the input file (we inherit it, but don't set manually!) should trigger the same logic as having the hint "openexr:ACESContainerPolicy" (I like this name) set to "strict". If the checks pass, the output is AcesContainer.

In other words, these (and only these?) cases should set the acesImageContainerFlag in the output file:
```
oiiotool valid_aces_container_file.exr -o out.exr
oiiotool aces_container_compliant_file.exr -sattrib openexr: ACESContainerPolicy relaxed -o out.exr
oiiotool aces_container_compliant_file.exr -sattrib openexr: ACESContainerPolicy strict -o out.exr
```

This case should set the ACES primaries and colorInteropID, but not acesImageContainerFlag:
`oiiotool not_fully_compliant_file.exr -sattrib openexr: ACESContainerPolicy relaxed -o out.exr`

Glowies:
Heya @antond-weta! I've done another pass on my implementation, and I tried to follow the logic in your pseudo-code - just without the global attribute. I also tried to converge some of the branching in your pseudo-code since it looked like there was some common behaviours on some of the branches. I made two assumptions there that I wanted to check with you:
* I saw that the colorInteropId attribute was only set in the branch where we're in "relaxed" mode. I assumed the same should be the case for the other branch as well (the case where we're in "strict" mode or we have the acesImageContainerFlag already set) 
* I also noticed that you were checking for the existence of the acesImageContainerFlag, and only assigned 1 to it if it *does not* exist. I removed that check, since I assume that the flag should be set to 1 on all cases where the image is compliant anyway.

Let me know if either of those assumptions are incorrect and I can fix them!

src/doc/builtinplugins.rst
```diff
@@ -1643,6 +1643,18 @@ control aspects of the writing itself:
...
+     - One of `none` (default), `strict`, or `relaxed`.
+       If not `none`, the spec will be checked to see if it is compliant
+       with the ACES Container format defined in `ST 2065-4`_. If it is,
+       `chromaticities` will be set to the ACES AP0 ones, and the
+       `acesImageContainerFlag` attribute will be set to 1.
+       In `strict` mode, if the spec is non-compliant, the output will
+       throw an error and avoid writing the image.
+       While in `relaxed` mode, if the spec in non-compliant, only a
+       warning will be printed and the attributes mentioned above will
+       *not* be written to the spec.
```
lgritz: Is this correct? We don't warn, right?

Glowies: Ah yes, this needs to be fixed with the recent changes. On it.

src/doc/builtinplugins.rst
```diff
@@ -1643,6 +1643,19 @@ control aspects of the writing itself:
...
+       If not `none`, the spec will be checked to see if it is compliant
+       with the ACES Container format defined in `ST 2065-4`_. If it is,
+       `chromaticities` will be set to the ACES AP0 ones, `colorInteropId`
+       will be set to 'lin_ap0_scene' and the `acesImageContainerFlag`
+       attribute will be set to 1.
+       In `strict` mode, if the spec is non-compliant, the output will
+       throw an error and avoid writing the image.
+       While in `relaxed` mode, if the spec is non-compliant, `chromaticities`
+       and `colorInteropId` will be set, but `acesImageContainerFlag`
+       will NOT.
```
lgritz, line 1658: I feel like this is not clear about what happens if the spec is passed with a colorInteropID that is set, but is not `lin_ap0_scene`. Are you saying that it will be changed to `lin_ap0_scene` (but pixels stay as they are)? Or is this something that will be rejected as non-compliant? Or does the behavior differ depending on the policy setting?

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,197 @@ set_exr_threads()
...
+        "reelName",
+        "storageMediaSerialNumber",
+    };
+
+    for (const auto& label : nonEmptyAttribs) {
+        const ParamValue* found = spec.find_attribute(label,
+                                                      OIIO::TypeDesc::STRING);
+        auto value              = spec.get_string_attribute(label, "");
+
+        if (found && value.empty()) {
```
lgritz: I know this isn't performance-critical code, but I think you are essentially doing a redundant search here. Can it be abbreviated to

```suggestion
        const ParamValue* found = spec.find_attribute(label,
                                                      OIIO::TypeDesc::STRING);
        if (found && (found->type() != TypeString || found->get_string(label).empty())) {
```

since found is already going to be pointing to the found entry? Also I added another failure case, which is if the named attrib is present but it's not a string type.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,197 @@ set_exr_threads()
...
+}
+
+
+
+bool
+is_aces_container_attributes_non_empty(const OIIO::ImageSpec& spec,
+                                       std::string& non_compliant_attr)
+{
+    // attributes in this list should NOT be empty if they exist
+    static const std::vector<std::string> nonEmptyAttribs = {
```
lgritz: I know I'm being totally pedantic, but doesn't this work just as well, without needing another allocation for the vector:

```suggestion
    static const std::string nonEmptyAttribs[] = {
```

I think the subsequent range-`for` still works unmodified.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,197 @@ set_exr_threads()
...
+
+    // Check attributes with exact values if they exist
+    if (spec.get_string_attribute("oiio:ColorSpace", ACES_AP0_colorInteropId)
+            != ACES_AP0_colorInteropId
+        || spec.get_string_attribute("colorInteropId", ACES_AP0_colorInteropId)
+               != ACES_AP0_colorInteropId) {
+        reason
+            = "Color space is not lin_ap0_scene as required for an ACES Container.";
+        return false;
+    }
```
lgritz, line 415: Do we want "no color space known" to pass? Or fail? The way to make "none specified" to fail is to not supply default values to the get_string_attribute calls.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,197 @@ set_exr_threads()
...
+        reason
+            = "EXR data type is not 'HALF' as required for an ACES Container.";
+        return false;
+    }
+
+    // Check compression
+    std::string compression = spec.get_string_attribute("compression", "zip");
+    if (compression != "none") {
+        reason = "Compression is not 'none' as required for an ACES Container.";
+        return false;
```
lgritz, line 395: What's the best policy here?

For something like "doesn't have the right channels", the only thing to do is fail.

But for "wrong data type" or "wrong compression", we could force compliance by just *fixing it*.

src/openexr.imageio/exroutput.cpp
```diff
@@ -290,6 +290,197 @@ set_exr_threads()
...
+    spec.attribute("chromaticities", OIIO::TypeDesc(OIIO::TypeDesc::FLOAT, 8),
+                   ACES_AP0_chromaticities);
+    spec.attribute("colorInteropId", ACES_AP0_colorInteropId);
+    spec.attribute("acesImageContainerFlag", 1);
+}
+
+
+
+bool
+process_aces_container(OIIO::ImageSpec& spec, std::string policy, int flag,
```
lgritz: What does flag mean here? Can we use a more descriptive name? And are there really multiple values that make sense, or should it simply be a `bool`?

src/openexr.imageio/exroutput.cpp
```diff
@@ -812,6 +1003,26 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
                                  Imf::LevelMode(m_levelmode),
                                  Imf::LevelRoundingMode(m_roundingmode)));
 
+    // Check ACES Container hint
+    int aces_container_flag = spec.get_int_attribute("acesImageContainerFlag",
```
lgritz, line 1008: Just as an FYI, the default is 0 even if you don't pass it.

lgritz:
Philosophically, the concerns I have boil down to these:

1. Should we have a mode where we repair the things that are easily fixed? Things like asking for a compression format that shouldn't be in an ACES container, forcing `half` type even if another is used, etc.
2. Should we reject an image that doesn't say what color space the pixels are in, or should we add the color space label (without actually knowing if it's right)?

src/openexr.imageio/exroutput.cpp
Glowies: Thank you so much! It looks like the `get_string` method takes in a 'maxsize' int instead of a string label. So, I just replaced the `label` that you were passing with just a 1 as the maxsize since we're only looking to check if the string is empty or not.

src/openexr.imageio/exroutput.cpp
Glowies: Ah yep, looks like that works with the for each loop there.

src/openexr.imageio/exroutput.cpp
Glowies: `flag` here holds the value of the acesImageContainerFlag attribute. I'll rename the parameter to explain that better!
Technically the acesImageContainerFlag attribute has to be an integer type, and should only have a value of 1 or it shouldn't exist at all, so I suppose this could be represented by a bool instead. However, I thought it would be best to keep this parameter the same type as the type of the attribute, just in case the ACES Container format specification is extended to allow values other than 1 in the future.

src/openexr.imageio/exroutput.cpp
Glowies: I think I'd like to get @antond-weta's opinions on that. I had assumed that we want the "no color space known" case to pass the compliance check. (ie. If there is no attribute indicating a color space, it passes the ACES Container compliance check, and we just continue processing the image assuming it is in linear AP0 space)

lgritz (approved):
I'm going to approve and merge this because I don't want it to languish any longer, and I think it's a definite step forward, although I believe we may wish to continue to iterate on this functionality in a couple of minor ways. 

In my mind, the biggest unanswered questions (which I hope we can discuss, definitively answer, and amend with subsequent PRs if necessary) are:

* Should we *require* a colorInteropID set to the correct color space in strict mode (or any mode), or else consider it an error? I think that now, having NO color space info is just accepted and assumed to be in the right space. But an ACES file not in the main ACES color space is totally out-of-spec and is a problem waiting to happen, so maybe "no color space news" is actually "bad news?"

* For any or all of the modes, should we *fix* things that are easily fixed, rather than make an ACES file that doesn't conform to spec? As an example, if an unsupported compression method is requested, should we just force an ACES-supported compression to ensure compliance?

zachlewis:
1. No. Per the CIR Interop ID proposal, we definitely should NOT set the colorInteropID here. 

2. We can enforce one of the lossless compression modes in non-strict mode if we'd like. Strict mode should be uncompressed.

zachlewis:
Oh, I misread item 1.

We should not prevent people from writing non-ACES stuff into the ACES container. Maybe we can log a warning with OCIO's logger, but we shouldn't do anything too contrived here.

PR #4908, opened 2025-09-26
Add type hints to Python docs

### LINKED ISSUE #4752: Python: add type annotations to all examples and documentation
Our minimum supported python version is 3.7 currently (soon to be 3.9, I suspect). Python 2.x is long gone for us, but lots of our python examples still haven't embraced the new syntax with [type annotations](https://docs.python.org/3/library/typing.html) that has been available since python 3.5.

There is PR #4692 in progress now to add stub files, and in the process, adds type annotations to most of the testsuite `python-*/src/test_*.py` files.

But that still leaves:
* All inline python examples in the docs themselves `src/doc/*.md` (note: rendered docs at https://docs.openimageio.org)
* The documentation examples that have been converted into tests and incorporated into docs by reference: `testsuite/docs-examples-python/src/docs-examples-*.py`
* Any other auxiliary scripts that we have, such as `testsuite/runtest.py`

This would be a great Dev Days project! (Especially for somebody who wants to do something for OIIO but knows more Python than C++ and maybe didn't think they had as much to offer.) I don't think any one person needs to do it ALL, this could be the project of multiple people, as long as you divide up the work -- each person can claim a few particular files, or a particular chapter of the docs, and just do that while others pick different non-conflicting parts.

cochang-dwa:
## Description

Closes #4752

Add Python type hints into documentation, in places where example code defines a function. That ended up being just `src/doc/pythonbindings.rst`. I also updated `testsuite/runtest.py` and the functions in `testsuite/docs-examples-python/src/docs-examples-*.py`.

While I was in `pythonbindings.rst`, I noticed a couple places with `py::method` (note the double colon). These lines didn't render in the docs, so I removed the extra colon to match other usage.

## Tests

I locally built the docs to make sure html looks good. I ran the tests with `ctest` and the tests ran, so I'm taking that as a sign that `runtest.py` still works.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/pythonbindings.rst
+++ b/src/doc/pythonbindings.rst
@@ -72,7 +72,7 @@ described in detail in Section :ref:`sec-typedesc`, is replicated for Python.
     These names are also exported to the `OpenImageIO` namespace.
 
 
-.. py::method:: TypeDesc.TypeDesc(typename='unknown')
+.. py:method:: TypeDesc.TypeDesc(typename='unknown')
 
     Construct a `TypeDesc` object the easy way: from a string description.
     If the type name is omitted, it will default to`UNKNOWN`.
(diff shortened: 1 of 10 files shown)


src/doc/pythonbindings.rst
```diff
@@ -72,7 +72,7 @@ described in detail in Section :ref:`sec-typedesc`, is replicated for Python.
     These names are also exported to the `OpenImageIO` namespace.
 
 
-.. py::method:: TypeDesc.TypeDesc(typename='unknown')
+.. py:method:: TypeDesc.TypeDesc(typename='unknown')
```
lgritz, line 75: oh, good catch on these!

lgritz (approved):
Looks good, thanks for the improvements! Merging.

cochang-dwa:
Awesome, thank you @lgritz!

PR #4909, opened 2025-09-26
fix(oiiotool): Ignore empty subimage(s) when calculating non-zero region

### LINKED ISSUE #4799: [BUG] issue with --trim multipart images that has parts with zero pixels
**Describe the bug**

If a multipart image has some parts/layers with zero pixels, --trim will trim to an inaccurate ROI.

It seems that those zero-pixel layers default to 1 pixel, causing the ROI union of all the layers to be innacurate.
nonzero_region_all_subimages(ImageRecRef A)
https://github.com/lgritz/OpenImageIO/blob/b22dcb38653e7bfe2ce9198e0dbce3d6962a2795/src/oiiotool/oiiotool.cpp#L3407

Image

The image above i illustrated what I expect to happen.
We want the display window to be strictly around non-zero valued pixels.
We can get rid of the empty layers and output a more accurate ROI, but there are cases we want to keep those layers i.e. for an image sequence where the empty layer will have some images later in the sequence.

```
OIIO 3.0.6.1 | unknown arch?
    Build compiler: MSVS 1943 | C++17/199711
Dependencies: DCMTK NONE, expat 2.6.3, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF NONE, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.23,
    Libheif NONE, libjpeg-turbo 3.0.4, LibRaw NONE, libuhdr 1.2.0, minizip-ng 4.0.7, OpenColorIO 2.4.1, OpenCV NONE, OpenEXR 3.2.4, OpenGL,
    OpenJPEG NONE, PNG 1.6.47, Ptex NONE, Ptex NONE, pybind11 2.13.6, pystring 1.1.4, Python3 3.9.13, Qt5 NONE, Qt6 NONE, Robinmap 1.4.0, TBB
    NONE, TIFF 4.6.0, WebP 1.4.0, yaml-cpp 0.8.0, ZLIB 1.3.1, ZLIB 1.3.1

```

**To Reproduce**

1. Have a multipart EXR, where all layers are non-zero pixels, and pixels are in the center of the image (far from the corners)
2. Run `oiiotool $INPUT --trim -o $OUTPUT`
3. Add a zero pixel layer to the multipart EXR.
4. Run oiiotool command again.
5. Compare the data window for both
6. You should see that when inserting a zero pixel layer, the data window will increase, because it includes one of the corners.

kookiCT:
Fixes #4799

## Description

When trimming an image that contains an empty subimage, the resulting ROI unexpectedly included the top-left pixel even if zero in all subimages.

To avoid this, empty ROIs are ignored when computing the overall non-zero region.
If all subimages are empty, the first subimage ROI is used and doctored to be 1 pixel.

## Tests

Two tests have been added to the oiiotool testsuite:
 - 1 to check the trimming of an empty image
 - 1 to check the trimming of a non-empty image with an empty subimage

## Checklist:

  doesn't change the API.)
  nonconforming.


lgritz (approved):
LGTM, thanks for the fix!

PR #4910, opened 2025-09-27
build: Update autobuild defaults for some dependencies

lgritz:
Raise the version we "autobuild" for some packages:
* OpenJPEG 2.5.2 -> 2.5.4
* pybind11 3.0.0 -> 3.0.1
* libtiff 4.6.0 -> 4.7.1
* webp 1.4.0 -> 1.6.0
* libjpeg-turbo 3.0.4 -> 3.1.2

And some minor debugging and other aids along the way:

* After a build, on linux, do a `ldd` to verify what libraries are linked dynamically for the logs.
* Enhance build_dependency_with_cmake macro to allow optional commands: GIT_COMMIT, GIT_SHALLOW, QUIET

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
             pybind11_ver: v2.9.0
             setenvs: export FREETYPE_VERSION=VER-2-12-0
                             BUILD_PNG_VERSION=1.6.30
+                            WebP_BUILD_VERSION=1.5.0
           - desc: VP2022 clang13/C++17 py39 avx2 exr3.1 ocio2.3
             nametag: linux-vfx2022.clang13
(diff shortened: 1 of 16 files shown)


lgritz:
I think this is correct, but in ci.yml, in the "all local builds" job, you'll want to remove the setting of env variable `OPENJPEG_VERSION=v2.5.4` because that will trigger the external script build_OpenJPEG.bash before the build. You don't want that, in this case you want to be sure you're forcing the autobuild one.

lgritz:
I'm investigating which of these bumps is the one that causes the wheel build to fail.

lgritz:
There were some CI failures in my first draft of this PR, but now they're all fixed, and it's ready for review.

grdanny (approved):
lgtm

PR #4911, opened 2025-09-27
cmake script to auto download and install the openjpeg dependency

grdanny:
## Description

sub task of : https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4387
for sep 2025 dev days.

Adding a autobuild cmake file for openjpeg.

Matching the latest version like in the yet unmerged: 
https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4910

## Tests

I built it locally - I can see it's downloading the files from github and building them.
I can run the tests.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/build_OpenJPEG.cmake
+++ b/src/cmake/build_OpenJPEG.cmake
@@ -1,0 +1,32 @@
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/Academ SoftwareFoundation/OpenImageIO
+
+set_cache (OpenJPEG_BUILD_VERSION 2.5.4 "OpenJPEG version for local builds")
+set (OpenJPEG_GIT_REPOSITORY "https://github.com/uclouvain/openjpeg.git")
+set (OpenJPEG_GIT_TAG "v${OpenJPEG_BUILD_VERSION}")
+set_cache (OpenJPEG_BUILD_SHARED_LIBS OFF #${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
+           DOC "Should a local OpenJPEG build, if necessary, build shared libraries" ADVANCED)
+
+string (MAKE_C_IDENTIFIER ${OpenJPEG_BUILD_VERSION} OpenJPEG_VERSION_IDENT)
+
+build_dependency_with_cmake(OpenJPEG
+    VERSION         ${OpenJPEG_BUILD_VERSION}
+    GIT_REPOSITORY  ${OpenJPEG_GIT_REPOSITORY}
+    GIT_TAG         ${OpenJPEG_GIT_TAG}
+    CMAKE_ARGS
+        -DBUILD_CODEC=OFF 
+
+
+    )
+# Set some things up that we'll need for a subsequent find_package to work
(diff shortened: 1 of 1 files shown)


src/cmake/build_OpenJPEG.cmake
```diff
@@ -0,0 +1,32 @@
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/Academ SoftwareFoundation/OpenImageIO
+
+set_cache (OpenJPEG_BUILD_VERSION 2.5.4 "OpenJPEG version for local builds")
+set (OpenJPEG_GIT_REPOSITORY "https://github.com/uclouvain/openjpeg.git")
+set (OpenJPEG_GIT_TAG "v${OpenJPEG_BUILD_VERSION}")
+set_cache (OpenJPEG_BUILD_SHARED_LIBS OFF #${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
```
lgritz: Was there a reason you forced OFF here?

lgritz: In fact, this may be why the Windows test is failing.

lgritz:
I think this is correct, but in ci.yml, in the "all local builds" job, you'll want to remove the setting of env variable OPENJPEG_VERSION=v2.5.4 because that will trigger the external script build_OpenJPEG.bash before the build. You don't want that, in this case you want to be sure you're forcing the autobuild one.

src/cmake/build_OpenJPEG.cmake
grdanny: I was "learning" a bit from the build_PNG.cmake (and a few others that had it off).
let me try to set it from the `LOCAL_BUILD_SHARED_LIBS_DEFAULT`

lgritz:
I pushed some fixes to the OpenJPEG build methodology. It was all about getting correct static libraries built and recognized.

grdanny:
I see, so if I'm running the bash script the does the OpenJpeg build and install I end up with these files:
```
oiio/ext/dist/lib64/pkgconfig/libopenjp2.pc
oiio/ext/dist/lib64/libopenjp2.a
oiio/ext/dist/lib64/libopenjp2.so
oiio/ext/dist/lib64/libopenjp2.so.7
oiio/ext/dist/lib64/libopenjp2.so.2.5.4
```
So basically we are telling the OpenImageIO build to prefer the "libopenjp2.a"?

Thanks for the help with this.

lgritz:
Aha, thanks for reminding me about the bash script. I need to adjust that, too.

lgritz:
I pushed one more amendment, to build_OpenJPEG.bash:

* Build it with position independent code always (lets us use the static library from a dynamic library).
* Allow a control to optionally disable building the dynamic libs.

That's for the bash script. The cmake auto-builder already does this (in my prior patch) and should default to the static library.

lgritz:
Wait, I think I need one more adjustment. Stay tuned.

lgritz (approved):
I think it all works now

grdanny:
Thanks for helping me take this to the finish line.

PR #4918, opened 2025-10-03
CHANGES and other misc docs changes from the 3.1 release

lgritz:
Forward-porting some docs touch-ups that were part of the 3.1 release staging.


lgritz:
Merging because this is not code and just catches main up to things that are already in the 3.1 release.

PR #4920, opened 2025-10-03
fix(api): Restore definition of OIIO_NAMESPACE_USING macro

lgritz:
Version 3.1 removed this definition, thinking it was safe because we didn't use it anywhere, didn't document it, and what it expands to (`using namespace OIIO`) isn't any longer than the macro name, so is pointless to use.

But downstream projects did use it. Oops.

### DIFF
--- a/src/include/OpenImageIO/oiioversion.h.in
+++ b/src/include/OpenImageIO/oiioversion.h.in
@@ -144,6 +144,7 @@ namespace OIIO = @PROJ_NAMESPACE@;
 // Macros to declare things in the current version's inline namespace.
 #define OIIO_NAMESPACE_BEGIN namespace @PROJ_NAMESPACE@ { inline namespace @PROJ_VERSION_NAMESPACE@ {
 #define OIIO_NAMESPACE_END } }
+#define OIIO_NAMESPACE_USING using namespace OIIO;
 #define OIIO_CURRENT_NAMESPACE @PROJ_NAMESPACE@::@PROJ_VERSION_NAMESPACE@
 
 #include <OpenImageIO/nsversions.h>


lgritz:
The thing I know is broken as a result of this is Blender, but @jessey-git says just leave it as-is and he'll fix on his end.

I will leave this PR up in case anybody else runs across this problem and wants it fixed. But if after a week I don't hear back from anybody saying yes they need it, I will close without merging.

mattyjams:
Hi @lgritz!

I actually just stumbled across this myself trying to build OpenUSD against the recently released 3.1.6.1, as it seems there are two usages of `OIIO_NAMESPACE_USING` in the current dev branch:

https://github.com/PixarAnimationStudios/OpenUSD/blob/a7c0637241a4fedce0338337bd7c63746c3d06eb/pxr/imaging/plugin/hioOiio/metadata.cpp#L21
https://github.com/PixarAnimationStudios/OpenUSD/blob/a7c0637241a4fedce0338337bd7c63746c3d06eb/pxr/imaging/plugin/hioOiio/oiioImage.cpp#L36

I can propose a PR to OpenUSD to change those to just `using namespace OIIO;` if that sounds right to you?

jessey-git:
Lol... I just posted on slack about this and yes please open a PR else I will do so tomorrow.

mattyjams:
Sounds good, the OpenUSD PR is here:
https://github.com/PixarAnimationStudios/OpenUSD/pull/3839

Thank you both!

lgritz:
Oh dear, that's two package impacted. I'll merge this and put out a patch release. So sorry for botching this.

PR #4921, opened 2025-10-06
build(gif): Add GIF library auto-build

### LINKED ISSUE #4387: [FEATURE REQUEST] Add self-building abilities for more dependencies

vangeliq:
## Description
Fixes #4387 

Added build_GIF.cmake module for GIF library dependency management. Since GIFLIB does not provide a CMakeLists.txt, I created a custom one that gets added to the cloned repository.

The flow is similar to other libraries with cmake support, so I added a conditional statement in the build_dependency_with_cmake macro that, if a cmakelists.txt path is provided, will add/replace existing CMakeLists.txt files on the cloned repository.

## Tests
I checked that all jobs were successful on the CI. Small side note that some jobs, like `VFX2024 clang/C++17 py3.11 exr3.2 ocio2.3 / clang++ c++17 py3.11`, already have GIFlib with the correct minimum version installed so the jobs pass without build_GIF being run. 

## Checklist:

- [X]  I have updated the documentation, if applicable. (Check if there is no
  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/build_GIF.cmake
+++ b/src/cmake/build_GIF.cmake
@@ -1,0 +1,39 @@
+# Copyright Contributors to the OpenImageIO project.
+# SPDX-License-Identifier: Apache-2.0
+# https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+######################################################################
+# GIF/GIFLIB
+# The original library does not have a CMake build system, so we
+# provide our own CMakeLists.txt template to build it.
+# See build_GIF_CMakeLists.txt for details.
+######################################################################
+
+set_cache (GIF_BUILD_VERSION "5.2.1" "GIFLIB version for local builds")
+super_set (GIF_BUILD_GIT_REPOSITORY "https://git.code.sf.net/p/giflib/code")
+super_set (GIF_BUILD_GIT_TAG "${GIF_BUILD_VERSION}")
+set_cache (GIF_BUILD_SHARED_LIBS ${LOCAL_BUILD_SHARED_LIBS_DEFAULT}
+           DOC "Should execute a local GIFLIB build; if necessary, build shared libraries" ADVANCED)
+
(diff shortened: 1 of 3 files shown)


lgritz (approved):
Holy cow, this is truly above and beyond the call of duty. When I wrote the issue, I don't think I realized that it lacked any cmake support, and I would not have expected (certainly not for a Dev Days level project) you to write a working CMake build system for *another project*.

Well done! This is great, thanks.

lgritz:
Great job, Valery!

PR #4932, opened 2025-10-17
ci: simplify ci workflow by using build-steps for old aswf containers, too

lgritz:
Trick learned from the rawtoaces CI workflow


lgritz:
This has been in review for over a week with no comments. Any objections?
Considering that it's a CI-only change, I will merge it if I don't hear anything by Monday morning.

PR #4933, opened 2025-10-17
fix(JXL): Correctly set Quality for JPEG XL

jonnyawsom3:
Distance in JPEG XL isn't linear, causing previous quality settings to scale between lossless and visually lossless, instead of lossless and the lowest quality.

Quality | libjxl Distance | Old OIIO Distance
|--------|------|------|
| 99 | 0.19 | 0.01 |
| 90  | 1.00 | 0.01 |
| 80  | 1.90 | 0.01 |
| 70  | 2.80 | 0.01 |
| 60  | 3.70 | 0.02 |
| 50  | 4.60 | 0.02 |
| 40  | 5.50 | 0.03 |
| 30  | 6.40 | 0.03 |
| 20  | 8.13 | 0.05 |
| 10  | 13.27 | 0.10 |
| 1   | 24.24 | 1.00 |
| 0   | 25.00 | Error |

Thankfully libjxl has a built in function to convert Quality to Distance, so I've swapped it in.
Unable to test it myself, but it's a simple fix that I ran by another libjxl dev too.


lgritz:
Looks reasonable to me, but I'm not a JXL expert.

@1div0 any thoughts?

1div0:
I will check++

Had free(time_t t) where t equals to 0

lgritz:
@1div0 Any chance to look at this? It's short, and it's certainly safe, I just seek a second opinion on whether it seems like a wise improvement to behavior.

lgritz:
@1div0 thoughts?

jonnyawsom3:
You can test the behaviour by encoding with OIIO without this PR and [cjxl](https://artifacts.lucaversari.it/libjxl/libjxl/latest/) at the same quality setting.
At Quality 1, OIIO gives a 267 KB file, cjxl gives a 35 KB file. At Quality 90, OIIO is 1.5 MB while cjxl is 267 KB.

1div0:
OK

lgritz:
@1div0 Does that mean "ok, I will take a look", or "I looked and I think this is ok?"

1div0:
@lgritz the second one.

The usage of

distance = JxlEncoderDistanceFromQuality(compqual.second);

is right.

lgritz:
Thanks, I appreciate the check from you.

lgritz (approved):

PR #4934, opened 2025-10-18
ci: if dnf install of ffmpeg fails, wait and try again

lgritz:
Lately, our retrieval of ffmpeg has been failing intermittently. Perhaps 5-10% of the time? Just enough that most CI runs have one job that fails and the others succeed (and sometimes 2 jobs fail, or all pass).

This patch is based on the hypothesis that the problem is truly transient, and that when it fails, maybe we can just pause 10 seconds and then try again, and it will usually succeed, cutting down on the effective failures rate. So far, so good?


lgritz:
Ugh, immediately ran into problem. Withdrawing and will fix.

PR #4937, opened 2025-10-20
ci(wheels): unbreak wheel release + other enhancements pt 1

zachlewis:
## Description

This PR fixes and improves a handful of wheels-related items:
 - Fix an issue where wheels produced with v3.1 releases were not getting uploaded to pypi
 - Update cibuildwheel to latest version (v3.2.1)
 - Remove the manylinux2014 builds, replace with manylinux_2_28 builds where applicable
 - Add CPython 3.14 wheels
 - Update macos-x86 runner to `macos-15-intel`
 - Update Python typestubs

## Notes
 - This PR ends support for pre-built wheels for glibc < 2.27 (CentOS-7 etc). However, the module can still be pip-installed locally, provided devtoolset-7+ (e.g., gcc-9+) is available. 
 - Unless we arrive at a more elegant solution, we must remember to change the conditional in the "upload_pypi" step for every major OIIO release.

### DIFF
--- a/src/cmake/build_OpenColorIO.cmake
+++ b/src/cmake/build_OpenColorIO.cmake
@@ -6,7 +6,7 @@
 # OpenColorIO by hand!
 ######################################################################
 
-set_cache (OpenColorIO_BUILD_VERSION 2.5.0 "OpenColorIO version for local builds")
+set_cache (OpenColorIO_BUILD_VERSION 2.4.2 "OpenColorIO version for local builds")
 set (OpenColorIO_GIT_REPOSITORY "https://github.com/AcademySoftwareFoundation/OpenColorIO")
 set (OpenColorIO_GIT_TAG "v${OpenColorIO_BUILD_VERSION}")
 set_cache (OpenColorIO_BUILD_SHARED_LIBS  OFF


lgritz:
You seem to be making a lot of adjustments. Let us know when it's ready to review.

.github/workflows/wheel.yml
```diff
@@ -220,33 +200,37 @@ jobs:
...
+              python: cp314-manylinux_aarch64
+              arch: aarch64
 
     steps:
       - name: Checkout repo
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Install Python
-        uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+        uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.
```
lgritz: Did you lose a character here in the comment?

.github/workflows/wheel.yml
```diff
@@ -308,7 +292,7 @@ jobs:
 
   macos:
     name: Build wheels on macOS
-    runs-on: macos-13
+    runs-on: macos-15-intel
```
lgritz, line 295: Oops, my bad, I totally didn't think about the wheels when I did this very change on the regular CI.

.github/workflows/wheel.yml
zachlewis: I did -- I'd intended to only add a line after this. Fixing now!

.github/workflows/wheel.yml
```diff
@@ -341,7 +328,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
         with:
-          python-version: '3.9'
+          python-version: 3.9
```
lgritz: FYI, the reason for the quotes is so that 3.10+ doesn't think it's floating point 3.1. I think it's not a bad idea to quote even when it's not necessary (3.7, 3.9, etc.) so that when somebody edits the line to bump the python version, they don't forget or aren't aware that some versions NEED to be quoted.

zachlewis:
This is ready for review -- I don't intend to make any more adjustments.

.github/workflows/wheel.yml
```diff
@@ -355,12 +342,12 @@ jobs:
         run: |
             brew install ninja ccache || true
         
-      - name: Remove brew OpenEXR/Imath
+      - name: Brew uninstall problematic dependencies
         run: |
-            brew uninstall --ignore-dependencies openexr imath || true
+            brew uninstall -f --ignore-dependencies openexr imath expat cmake || true
```
lgritz, line 347: Why uninstall cmake?

src/build-scripts/build_ccache.bash
```diff
@@ -16,7 +16,7 @@ echo "HOME=$HOME"
 echo "PWD=$PWD"
 echo "ARCH=$ARCH"
 
-CCACHE_PREBULT=${CCACHE_PREBULT:=1}
+CCACHE_PREBUILT=${CCACHE_PREBUILT:=1}
```
lgritz, line 19: Oh boy, that one is embarrassing

src/cmake/build_OpenColorIO.cmake
```diff
@@ -6,10 +6,10 @@
 # OpenColorIO by hand!
 ######################################################################
 
-set_cache (OpenColorIO_BUILD_VERSION 2.4.2 "OpenColorIO version for local builds")
+set_cache (OpenColorIO_BUILD_VERSION 2.5.0 "OpenColorIO version for local builds")
```
lgritz: This feels like an orthogonal decision to anything to do with the wheels, because it will change the version of OCIO that *anyone* will get for the auto-builder. Are we confident enough in 2.5.0 to make it the default?

.github/workflows/wheel.yml
zachlewis: Oh, good cool, I'll restore that too!
At an earlier point in this PR, I'd deleted the "python-version" line entirely, because as far as I'd known, we were only using it to force the python 3.8 wheels to build on macos-arm; and I ended up restoring it by hand when I discovered that github would whine about the line not being there. 
Anyway, adding those quotes back in!

lgritz:
Nice speed bump from not having to build ccache from source on Linux Intel. Is that just because you eliminated the jobs that involved the older glibc? Are there pre-built ccache we can use for the linux arm wheels?

src/cmake/build_OpenColorIO.cmake
zachlewis: Good point -- I had only been considering that users locally pip-installing (building) OIIO would want the same OCIO behavior as the available prebuilt wheels. It totally skipped my mind that this affects the default auto-building behavior everywhere!

I am confident enough in 2.5.0 to make it the default, in terms of stuff not breaking; but I agree with you that this change is way out of scope for this PR! I'll revert it here, and open a new PR just for this sucker so we can further discuss.

.github/workflows/wheel.yml
zachlewis: Because we want scikit-build-core to use whichever version of cmake scikit-build-core itself fetches; i.e., we want to be able to use skbuild settings (in pyproject.toml, or specified by the pip command line options, or set via sbuild env vars) to drive the cmake version used by the build.

Without this change, homebrew's cmake takes precedence over skbuild's cmake.

zachlewis:
> Is that just because you eliminated the jobs that involved the older glibc? 

I hadn't considered that, but... yeah, having eliminated the manylinux2014 (glibc-2.17) builds, we're able to use the pre-built ccache for the remaining manylinux_2_28 (glibc-2.28) builds for the time being.

> Are there pre-built ccache we can use for the linux arm wheels?

Not officially, as far as I can tell. The developers have only recently (as of ccache-4.12.0, I think) started to release aarch64-compatible pre-built binary distributions, and those seem to require a newer version of glibc than we'd like to support (w.r.t. current vfx reference platform recommendations). 

One option we could investigate is building ccache in a custom docker image based on the manylinux_2_28 image. 

Another option is to try switching from ccache to [sccache](https://github.com/mozilla/sccache), which seems to be a more purpose-built solution for what we're trying to do. 
(There are sccache binaries available for all platforms / archs we're interested in; the linux binaries are linked against musl, so there are no glibc issues, period).

src/cmake/build_OpenColorIO.cmake
lgritz: I feel like without a serious beta/RC cycle, OCIO is treating 2.5.0 as their release candidate and a future 2.5.1 as the one where stuff actually gets fixed. I think 2.5 hasn't actually had a chance to prove itself in production quite yet, so I'm cautious about making it the default. Anybody can override by setting either cmake or env variable OpenColorIO_BUILD_VERSION to 2.5.0, but the default should be playing it safe with a well-tested version we are 100% confident in.

zachlewis: My thinking was that we'd want the version of OCIO available on pypi to match what the version of OIIO available on pypi was built against. 

But I think that's a wise decision, to regard 2.5.0 as a release candidate, and I agree with you about holding off until 2.5.1 rolls out. Let's do that.

And if folks really want, they can locally build the OIIO bindings against OCIO 2.5 with:
`$ pip install OpenImageIO  --config-settings=cmake.define.OpenColorIO_BUILD_VERSION="2.5.0"`

zachlewis:
Thanks Larry! Seems good to go to me too.

PR #4940, opened 2025-10-26
fix: uninitialized value revealed by clang-21 warning

lgritz:


lgritz:
> nit: consider using nullptr rather than NULL but otherwise looks good.

The call is to a C stdlib function, I actually did look at it and ask myself if "nullptr" would always and forever be valid there. I guess so? But something made me wonder in the moment, so I used the C convention for NULL when calling the C function. Maybe that's overly cautious?

jessey-git:
I've honestly never thought too hard about using a different convention when calling c code from c++.  I believe nullptr will always be fine for pointers.  The only situation that's not equivalent is if the code used NULL to mean the integer literal 0 (e.g. `int foo = NULL;`). Replacing NULL with nullptr in those contexts would be a compiler error though.

lgritz:
ok, will fix

PR #4941, opened 2025-10-28
testing: Update ref images for heif

lgritz:
I think that in a recent checkin, we accidentally made some ref outputs for this test be identical, that were supposed to be for when libheif is built without avif support.


lgritz:
Any objections to my merging this? It's only reference output for testing, no code changes.

PR #4946, opened 2025-10-31
fix(exr): improve attribute translation rules

lgritz:
In exr output, we were doing attribute name substitutions using case-insensitive comparisions, whereas on the exr input side, we were using case-sensitive comparisons. This was leading to some strange behavior where certain attributes were not properly "round tripping" when reading and writing exr files, particularly when it involved attribute names that differed only in case from certain reserved/standard names.

Also be a little more flexible in adjusting exr attributes that have the wrong type, by being able to convert int or float to a value that's required to be a string.

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -1140,8 +1140,7 @@ OpenEXROutput::put_parameter(const std::string& name, TypeDesc type,
     TypeDesc exrtype  = TypeUnknown;
 
     for (const auto& e : exr_meta_translation) {
-        if (Strutil::iequals(xname, e.oiioname)
-            || (e.exrname && Strutil::iequals(xname, e.exrname))) {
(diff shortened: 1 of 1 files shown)


lgritz:
It would be helpful to get a fast review on this, I'd really like to include it in this weekend's set of releases. It addresses a problem that was reported at my work.

JGoldstone:
It feels as if attribute type coercion and adherence to the ACES container specification could be in conflict; I would hope we would never write a file that had the flag that says "I follow the ACES container spec" but that also had attribute types that didn't conform to the spec. I'm not sure what to do on the input side; I need to catch up on what's been proposed and, more than likely, already been merged.

lgritz:
What's the conclusion on this? Does anybody object to this particular fix?

lgritz:
No objections to this specific changes in two weeks -> merging.

PR #4951, opened 2025-11-09
test: image_span_test reduce benchmark load for debug and CI renders

lgritz:
Almost all the other unit tests do this, just adding it to image_span_test: automatic reduction in test load for CI and debug renders by doing fewer time trials. This cuts down on runtime for CI, but stil exercises the test code. Command line arguments allow you to override, in cases where you need really good timing fidelity even for a debug run.

### DIFF
--- a/src/libOpenImageIO/image_span_test.cpp
+++ b/src/libOpenImageIO/image_span_test.cpp
@@ -19,8 +19,7 @@
 using namespace OIIO;
 
 
-static int iterations = 0;  // 1000000;
-static int ntrials    = 5;
+static int ntrials = 5;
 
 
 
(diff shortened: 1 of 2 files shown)


lgritz:
Need this right away to address some CI issues. Will merge shortly if there are no objections. The code changes here are only to unit tests.

src/libOpenImageIO/image_span_test.cpp
```diff
@@ -18,6 +19,10 @@
 using namespace OIIO;
 
 
+static int iterations = 0;  // 1000000;
```
jessey-git: Looks like `iterations` is unused. It's only set, but never passed into something like `bench.iterations(...)`

src/libOpenImageIO/image_span_test.cpp
```diff
@@ -544,9 +554,37 @@ test_image_span_within_span()
...
 
+static void
+getargs(int argc, char* argv[])
+{
+    ArgParse ap;
+    ap.intro(
+          "image_span_test -- unit test and benchmarks for OpenImageIO/image_span.h\n" OIIO_INTRO_STRING)
+        .usage("image_span_test [options]");
+
+    ap.arg("--iterations %d", &iterations)
```
jessey-git: Funny stat: There's currently 14 instances of tests that use `--iters` and 14 instances that use `--iterations`.  This new addition here would tip the scale in favor of using the full `--iterations` name -- should *somebody* make them all the same and create a new PR for that?

[EDIT] Actually I miscounted, there's just 1 other test, simd_test.cpp that uses the full --iterations name.

src/libOpenImageIO/image_span_test.cpp
lgritz: fixing

src/libOpenImageIO/image_span_test.cpp
lgritz: Aha, fixing simd_test

PR #4952, opened 2025-11-10
fix: type warning in assertion in jpeg2000output.cpp

lgritz:


lgritz:
Need this right away to address some CI issues. Will merge shortly if there are no objections. The code change here only involves a single assertion in debug mode and can't change anything for people running release builds in production.

PR #4954, opened 2025-11-12
ci: fix broken ci, debug and static cases, bump some latest

lgritz:
Two Ubuntu-based cases (debug build and static build) started failing this past week, unrelated to us. I think the runners changed yet again, it seems to be having particular trouble because those tests were trying to use gcc-9. But there's no particular need for that anyway, it just reflects how long ago we last modified those tests.

So this spurred a grab bag of CI fixes:

* Move the debug and static linkage tests off of raw ubuntu runners and to using the ASWF docker VFX Platform 2025 containers. This makes them more modern and also more stable.

* Turns out the "static" test hadn't been working properly lately -- it was accidentally building dynamic libs, not static at all! Fix that, but then it uncovered some problems that are out of scope to fix in this PR. I will look into it separately, but in the mean time I have commented out the static build test to suppress the failures.

* Bump a few of the "latest" dependencies to the most recent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,7 +284,6 @@ jobs:
             simd: "avx2,f16c"
             fmt_ver: 10.1.1
             pybind11_ver: v2.12.0
-            benchmark: 1
             setenvs: export PUGIXML_VERSION=v1.14
             optional_deps_append: "LibRaw"
           - desc: VFX2024 clang/C++17 py3.11 exr3.2 ocio2.3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -313,6 +313,32 @@ jobs:
...
+            nametag: linux-vfx2025-debug
+            runner: ubuntu-latest
+            container: aswf/ci-oiio:2025
+            cxx_std: 17
+            build_type: Debug
+            python_ver: "3.11"
+            simd: "avx2,f16c"
+            fmt_ver: 11.2.0
+            pybind11_ver: v2.13.6
+            benchmark: 1
```
jessey-git: Just double-checking that `benchmark` was intended to be enabled here on the Debug config?

lgritz: um, no, thanks for catching. I think that's a cut and paste error.

lgritz: and I found a couple others I can eliminate

PR #4957, opened 2025-11-15
build: Disable LERC in libTIFF local build script

lji-ilm:
## Description

Recently, when building OIIO using `OpenImageIO_BUILD_LOCAL_DEPS=all` cmake option, I noticed a reference to `lerc` that fails during the linker stage, when compiling the simd tests:

```
[ 71%] Building CXX object src/libutil/CMakeFiles/simd_test.dir/simd_test.cpp.o
[ 71%] Linking CXX executable ../../bin/simd_test
/usr/bin/ld: ../../lib/libOpenImageIO.so.3.1.7: undefined reference to `lerc_decode'
/usr/bin/ld: ../../lib/libOpenImageIO.so.3.1.7: undefined reference to `lerc_encodeForVersion'
/usr/bin/ld: ../../lib/libOpenImageIO.so.3.1.7: undefined reference to `lerc_getBlobInfo'
collect2: error: ld returned 1 exit status
gmake[2]: *** [src/libutil/CMakeFiles/simd_test.dir/build.make:100: bin/simd_test] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2158: src/libutil/CMakeFiles/simd_test.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2
```
`lerc` is an image compression library from ESRI: https://github.com/Esri/lerc

A deeper dive of this reference to lerc shows that it is referred in `libtiff`. `libtiff`, however, did not seem to correctly export this dependency to OIIO when being built locally. 

If someone was building OIIO with a system-installed `libtiff`, this error is unlikely to happen.

By disabling `lerc` in `libTiff`, the local build scripts ( `OpenImageIO_BUILD_LOCAL_DEPS=all` ) runs without a problem.

## Tests

Tested build OIIO with  `OpenImageIO_BUILD_LOCAL_DEPS=all` with this patch and was successful in finishing the build.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/cmake/build_TIFF.cmake
+++ b/src/cmake/build_TIFF.cmake
@@ -44,6 +44,7 @@ build_dependency_with_cmake(TIFF
         -D lzma=OFF
         -D zstd=OFF
         -D jbig=OFF
+        -D lerc=OFF
         ${MORE_TIFF_CMAKE_ARGS}
     )
 


lgritz:
The icx/icc failures are unrelated to your changes.

PR #4958, opened 2025-11-16
ci: unbreak icc/icx CI

lgritz:
A few days ago, the icc and icx CI tests started failing, saying they were out of disk space while installing dependencies (specifically, the intel compilers themselves).

Now, the VFX Platform 2023 asf docker image I'm using hasn't changed, and neither has the years-old Intel rpms with the compilers. So maybe the runners themselves just changed again, in a way that adds some pre-installs and leaving us not quite enough disk space for everything else we need?

Asking for a few less things to get installed seems to free up just enough space. And for the icx one, I also swtiched to the 2025 container, which let me use a slightly newer icx version.

Also: Don't try to install giflib needlessly, which speeds up the setup time. And remove some dead code in my dependency setup script.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,30 +251,29 @@ jobs:
             fmt_ver: 7.1.3
             # icc MUST use this older FMT version
             pybind11_ver: v2.9.0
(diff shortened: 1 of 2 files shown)


lgritz:
Merging this because it fixes CI.

PR #4961, opened 2025-11-22
fix(exif): Support EXIF 3.0 tags

### LINKED ISSUE #4960: [FEATURE REQUEST] Access to metadata Exif.Photo.ImageTitle
I want to use Exif.Photo.ImageTitle as an auto-generated caption. However, it appears that OpenImageIO does not recognize this field.

Same API as to fetch ImageDescription

Using ImageDescription as substitute

**Additional context**
From https://exiv2.org/tags.html
0xa436 	42038 	Photo 	Exif.Photo.ImageTitle 	Ascii 	This tag records the title of the image.

lgritz:
We were supporting Exif 2.3, but in 2024, Exif 3.0 was published and it adds some more tags for us to recognize.

The only part I am not sure about here is that the ExifVersion we output is still marked as 2.3, not 3.0, because I'm worried that if we mark it as 3.0, older software reading the images will reject it. Let's see what happens. If downstream apps care, maybe we can bump to 3.0, or even bump to 3.0 only if any of the 3.0 tags are used (that is more involved).

Fixes #4960

### DIFF
--- a/testsuite/heif/ref/out-libheif1.12-orient.txt
+++ b/testsuite/heif/ref/out-libheif1.12-orient.txt
@@ -82,6 +82,7 @@ Reading ../oiio-images/heif/greyhounds-looking-for-a-table.heic
     Exif:ApertureValue: 2.52607 (f/2.4)
     Exif:BrightnessValue: 2.7506
     Exif:ColorSpace: 65535
+    Exif:CompositeImage: 2
     Exif:DateTimeDigitized: "2023:09:28 09:44:03"
     Exif:DateTimeOriginal: "2023:09:28 09:44:03"
     Exif:DigitalZoomRatio: 1.3057
--- a/testsuite/heif/ref/out-libheif1.4.txt
+++ b/testsuite/heif/ref/out-libheif1.4.txt
@@ -82,6 +82,7 @@ Reading ../oiio-images/heif/greyhounds-looking-for-a-table.heic
     Exif:ApertureValue: 2.52607 (f/2.4)
     Exif:BrightnessValue: 2.7506
     Exif:ColorSpace: 65535
+    Exif:CompositeImage: 2
     Exif:DateTimeDigitized: "2023:09:28 09:44:03"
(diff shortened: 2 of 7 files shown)


lgritz:
The MacOS failure is unrelated.

src/include/OpenImageIO/tiffutils.h
```diff
@@ -128,6 +136,16 @@ enum TIFFTAG {
...
     EXIF_GAMMA                      = 42240,
+    EXIF_IMAGETITLE                 = 42038,
+    EXIF_PHOTOGRAPHER               = 42039,
+    EXIF_IMAGEEDTOR                 = 42040,
+    EXIF_CAMERAFIRMWARE             = 42041,
+    EXIF_RAWDEVELOPINGSOFTWARE      = 42042,
+    EXIF_IMAGEEDITINGSOFTWARE       = 42043,
+    EXIF_METADATAEDITINGSOFTWARE    = 42044,
+    EXIF_COMPOSITEIMAGE             = 42080,
+    EXIF_SOURCEIMAGENUMBEROFCOMPOSITEIMAGE        = 42080,
```
jessey-git: Should be 42081

src/include/OpenImageIO/tiffutils.h
```diff
@@ -128,6 +136,16 @@ enum TIFFTAG {
...
+    EXIF_SOURCEIMAGEEXPOSURETIMESOFCOMPOSITEIMAGE = 42080,
```
jessey-git: Should be 42082

src/include/OpenImageIO/tiffutils.h
lgritz: Thanks, fixed

src/include/OpenImageIO/tiffutils.h
lgritz: Thanks, fixed

jessey-git (approved):

lgritz:
The MacOS issue seems to have gone away on its own. Yay, procrastination!

PR #4967, opened 2025-11-28
feat(exr): Write OpenEXR colorInteropID metadata based on oiio:ColorSpace

brechtvl:
## Description

If the colorspace exists and has an interop ID in an OCIO 2.5 config, use that.
Otherwise check if the colorspace is equivalent to a known color interop ID.

## Tests

Tests were added.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/openexr.imageio/exroutput.cpp
+++ b/src/openexr.imageio/exroutput.cpp
@@ -13,6 +13,7 @@
 #include <numeric>
 
 #include <OpenImageIO/Imath.h>
+#include <OpenImageIO/color.h>
 #include <OpenImageIO/platform.h>
 
 #include <OpenEXR/IlmThreadPool.h>
@@ -1024,6 +1025,15 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
                 non_compliance_reason);
             return false;
         }
+    }
+
+    // Set color interop ID from colorspace
+    if (spec.get_string_attribute("colorInteropID").empty()) {
+        const ColorConfig& colorconfig(ColorConfig::default_colorconfig());
+        string_view colorspace = spec.get_string_attribute("oiio:ColorSpace");
+        string_view interop_id = colorconfig.get_color_interop_id(colorspace);
+        if (!interop_id.empty())
+            spec.attribute("colorInteropID", interop_id);
     }
 
     // Deal with all other params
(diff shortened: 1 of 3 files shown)


brechtvl:
~~Draft because this depends on and includes the changes from #4964.~~

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -1997,6 +1997,170 @@ ColorConfig::parseColorSpaceFromString(string_view str) const
 }
 
 
+//////////////////////////////////////////////////////////////////////////
+//
+// Color Interop ID
+
+namespace {
+enum class CICPPrimaries : int {
```
zachlewis: I think we should have enums for Unspecified = 2 for  CICPPrimaries and CICPTransfer, too; and "Unknown" should map to 2-2-2. 

I also think, for the time being, we should have CICP 1,2 also map to "g24_rec709_display" (but not necessarily the other way around).

brechtvl:
It's suspicous that an EXR related test is failing on Windows-2022, but not sure how it could be caused by the changes in this PR.

lgritz:
The bleeding edge test is unrelated and I'm fixing it separately.

Not sure about the windows one. I'm rerunning the failed tests now. A lot of glitchy stuff sometimes happens on the GHA runners, it's always worth trying again when something that has consistently passed for a long time spuriously fails.

src/openexr.imageio/exroutput.cpp
```diff
@@ -1026,6 +1027,15 @@ OpenEXROutput::spec_to_header(ImageSpec& spec, int subimage,
...
     }
 
+    // Set color interop ID from colorspace
+    if (spec.get_string_attribute("colorInteropID").empty()) {
+        const ColorConfig& colorconfig(ColorConfig::default_colorconfig());
+        string_view colorspace = spec.get_string_attribute("oiio:ColorSpace");
+        string_view interop_id = colorconfig.get_color_interop_id(colorspace);
+        if (!interop_id.empty())
+            spec.attribute("colorInteropID", interop_id);
+    }
```
doug-walker, line 1037: Should it look for the color space name in the user's OCIO config before looking in the default config?

lgritz: > Should it look for the color space name in the user's OCIO config before looking in the default config?

That's what "default" means in this case. It is only the "built-in" config if there was no $OCIO set and none set through global OIIO attributes.

lgritz:
Is this one ready to merge?

brechtvl:
Yes, all my PRs that are not marked as draft are ready.

lgritz (approved):

PR #4968, opened 2025-11-28
feat(JXL): CICP read and write support for JPEG XL

brechtvl:
## Description

The JPEG XL color encoding metadata only supports a subset of CICP. So for
example `srgb_p3d65_display` and `pq_rec2020_display` are supported, but
`g26_xyzd65_display` is not.

Custom primaries, custom white point and arbitrary gamma could be used to
support more, but I didn't implement that.

## Tests

Tests for read and write were added.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/jpegxl.imageio/jxlinput.cpp
+++ b/src/jpegxl.imageio/jxlinput.cpp
@@ -16,6 +16,7 @@
 #include <cassert>
 #include <cstdio>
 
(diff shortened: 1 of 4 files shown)


brechtvl:
~~Draft because this depends on and includes the changes from #4964.~~

PR #4969, opened 2025-11-30
ci: Emergency fix change deprecated sonarqube action

lgritz:


lgritz:
There are several things unrelated broken, including other things about the analysis workflow, which now that I'm checking, I can see hasn't worked for a couple weeks.

But nonetheless, this PR prevents literally everything from being broken, so I'm going to emergency merge it and tackle the rest separately.

PR #4971, opened 2025-12-01
feat: Make color interop IDs work with older OpenColorIO ACES configs

brechtvl:
## Description

Add a legacy alias for color interop IDs, matching aliases that exist in the
OpenColorIO ACES configs back to version 1.0.0 (made for OpenColorIO 2.1).

Without this, only builds with OpenColorIO 2.5+ will correctly write color
metadata in various cases, for example when using `oiiotool` `--ociodisplay`
with a HDR or wide gamut display color space.

## Tests

Added test to verify if old OpenColorIO can find the interop ID for ACES2065-1 despite lacking the official interop ID in its default config.

## Checklist:

  doesn't change the API.)
  nonconforming.


brechtvl:
~~Draft because this depends on and includes the changes from https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4964.~~

lgritz (approved):
LGTM

zachlewis:
I'd like to propose an alternate approach -- i.e., a different PR I'd like to submit that provides a more robust mechanism for relating color interop IDs with config identities. Instead of mapping interop_id values to canonical names of color configs in previous ACES OCIO configs, I'm proposing that we compare the actual ColorConfig's color space definition to a known color space definition associated with the known CIF and OCIO-builtin interop IDs. 

Here's a python proof of concept:
https://gist.github.com/zachlewis/944d3c99b918565228013c1a05b78e6e

And it would be used like...
```python
cim = ColorInteropManager()
...
# Match a color space in the config to an alias, or the nearest definitional equivalent of a known interop id, if possible
color_space = cim.get_color_space_for_interop_id(my_color_config, spec["colorInteropID"])

...

# Get an interop_id for a color space in the config if defined; else, try to match the color space definition to a known CIF / OCIO-builtin interop_id.
color_interop_id = cim.get_interop_id_for_color_space(my_color_config, spec["oiio:ColorSpace"])
...

# Get an OCIO processor from an interop ID to a config color space; as long as the interop_id is an alias defined in the config, or one of the CIF / builtin-OCIO interop IDs, a processor will be created.
my_proc = cim._get_processor_from_interop_id(color_interop_id, iba_custom_color_config)

```

There are benefits / goals of my proposed approach:
  - Enable color-interop-id functionality in any config, not just the official / builtin ACES OCIO configs.
  - Implement the mechanisms described in the CIF [An ID for Color Interop](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE) manifesto
  - Help the default ColorConfig interact with an IBA-specific custom ColorConfig, and vice-versa
  - Help frame OpenImageIO's color management mechanisms in terms of "inner" (i.e. OCIO) versus "outer" (CICP, ICC, format-specific mechanisms, etc) mechanisms, glued together by "colorInteropID" (e.g., to defer the expensive parts of resolving interop_id --> color space name)
  - Provide a more robust mechanism for `ColorConfig.equivalent`; permits a way to test for equivalency across configs.
  - Work with OCIO-2.3+
  - Provide a means for using CIF / OCIO-builtin interop_id values interchangeably with color space names / aliases in the IBAs, even if they don't actually exist in the active config.
  
  
Detriments of my approach:
 - Comparing color space definitions is slower than comparing strings
 - Only mapped for known OCIO and CIF interop_ids; there are some color spaces in older ACES OCIO configs that have no mapping; will require communication with OCIO folks to decide on what to do
 - Some color spaces in some configs aren't mapping to interop_ids as expected. I'm looking into this.
 - There are ~500 lines of OCIO config that feels kinda nasty to paste into source; but when we raise the minimum to OCIO-2.5, we can pair this down significantly.

brechtvl:
I mainly did this because it was trivial. But if you want to implement another approach that's fine with me.

I am concerned about performance though. Would your approach require creating an optimized processor for every color space in the config, or is there a faster way to compare? Some numbers in a quick test creating a processor from scene linear to every color space in a config:
* Latest ACES studio config:  0.47s
* [Joegenco/PixelManager](https://github.com/Joegenco/PixelManager) config: 19.23s

lgritz:
When considering performance, the canonical use case that pops into my head is somebody writing a shell script that calls oiiotool on each of, say, 300 frames. Even 1 second of extra setup overhead for every oiiotool command that references color spaces seems like too much -- that alone would add 5 minutes to our loop. It could be a sizeable percentage of a simple operation like a file format change + color space transform, which is a total staple operation that productions rely on.

Yes, I know that it would be better to recode that shell script in Python using IBA functions directly, or to find a way to use a single oiiotool launch with wildcards, so that certain overheads would be incurred once per shot instead of once per frame. None of that changes the fact that people do this a lot, and we should try to not let this common practice be more wasteful than necessary.

Also, I'm not knocking either of these solutions or saying that there is any particular time cost that I would reject outright. I'm just saying: let's be mindful of performance in the loop-of-oiiotool-launches common use case and strive to reduce that overhead over time.

lgritz:
> Yes, I know that it would be better to recode that shell script

Even if users were perfect, another use case that concerns me is our own test suite, which contains hundreds of separate oiiotool launches. Adding seconds to each launch has a significant effect on how long our CI takes.

zachlewis:
> I am concerned about performance though. Would your approach require creating an optimized processor for every color space in the config, or is there a faster way to compare? Some numbers in a quick test creating a processor from scene linear to every color space in a config:
> 
>     * Latest ACES studio config:  0.47s
> 
>     * [Joegenco/PixelManager](https://github.com/Joegenco/PixelManager) config: 19.23s


Excellent, excellent question. I've been looking at the performance implications of different methods of resolving color spaces <--> interop_ids across configs. For example, using the OCIO-2.5 config-merging mechanisms is much faster than iteratively creating processors from one color space to every other in a config (in Python, in my case); but config-merging incurs a larger penalty at startup (or at first use of the resolution heuristics) that scales as a function of the length of the merged OCIO config, compared to iteratively testing each color space until a match is found. 

Currently, I'm providing three mechanisms for the resolution, mitigated by "strict" and "quick" options.

1. If "strict" is enabled, don't do any fancy heuristics -- just use the names and aliases defined in the given ColorConfig.
2. Try to find a matching color space with `Config.IdentifyBuiltinColorSpace` -- fast built-in mechanism that uses quick heuristics to ascertain a match
3. If the above method hasn't found a match, and if "quick" is disabled, apply a custom slower approximate-equivalency-testing method that iteratively creates draft-optimized processors and tests for a NoOp, returning the first match found.

(The approximate-equivalency-testing method in # 3 can be further optimized to avoid checking against every color space in the config)

((Also, we can cache individual calls, obviously))

> let's be mindful of performance in the loop-of-oiiotool-launches common use case and strive to reduce that overhead over time.

Absolutely, this is primarily what motivated my design, versus other approaches -- I'd like to incur the least amount of overhead for oiiotool uses, by means of a deferred color-interop-id --> color space resolution, and not doing any kind of config merging at startup (or at all, really).

lgritz:
> I'd like to propose an alternate approach

Is that imminently ready to go? Or do you think the right approach is to merge (and backport) this PR, even if at some point in the future we replace it, or replace it for forward-looking (3.2+) releases? (Remembering also that for forward-looking releases, we are also free to raise the OCIO minimum version.)

zachlewis:
>> I'd like to propose an alternate approach

> Is that imminently ready to go? 

The approach itself is ready to go -- the Python implementation is in the gist above. 

I haven't implemented in C++, as I'm not a strong C++ developer, and I wanted to give you guys a chance to have a look and see if you guys had any questions, comments, doubts, fears, concerns, or advice before I let LLM-autocompletes do fair amount of steering. (That being said, I haven't had too much difficulty making modest C++ contributions before, and I'm happy to muscle through without any AI if you'd prefer. Either way, I won't submit code I don't understand). In any case, it's not very much code.

> Or do you think the right approach is to merge (and backport) this PR, even if at some point in the future we replace it, or replace it for forward-looking (3.2+) releases? (Remembering also that for forward-looking releases, we are also free to raise the OCIO minimum version.)

Well, I'm mentioning all this here because I don't think it makes sense to merge this PR if we're gonna end up replacing it somewhat imminently, if that's the direction we decide to go. But it honestly doesn't make a difference to me, and I'm happy to do whatever's easiest for everyone else.

lgritz:
> Well, I'm mentioning all this here because I don't think it makes sense to merge this PR if we're gonna end up replacing it somewhat imminently,

I agree with that. I was calling out a few situations where it might make sense to merge despite an eventual replacement:
* If it's not really imminent, and is going to drag on for weeks.
* If we're not sure if it's going to be better, or might have perf issues, and won't know until the implementation is complete.
* If your solution is just going to be easier/simpler if it is accompanied by a bump in OCIO minimum, which we can only do in main.

> I'm happy to muscle through without any AI if you'd prefer. Either way, I won't submit code I don't understand)

ASWF doesn't have formal rules about this yet, though it's something we're trying to discuss in the TAC to have overall guidance. For myself, the main things I insist on at the moment are
* You need to understand and stand behind any code you submit, just as if you wrote it yourself. I don't want to hear "but the LLM..." if something breaks.
* Disclose (in the commit comment) what tool you used and how.
* Be careful with any generated from-scratch code, you are responsible for ensuring that it's not a nearly verbatim copy of something that came from code with an incompatible license. (I don't think you have to worry about that when using it for language-to-language translation of something you authored, or other touch-up work on existing code.)

And I am curious to hear your experience with it! I go in with an open mind and am trying to learn more about where it's really useful.

zachlewis:
How about this... if I don't have a mergable PR ready or very very near ready by the beginning of the second week of January, let's merge this PR. I'm pretty sure I can get something in before before then, but no matter what, I'd like to get the PRs Brecht has submitted so far in and backported in time for the early Feb release. 

 
> If your solution is just going to be easier/simpler if it is accompanied by a bump in OCIO minimum, which we can only do in main.

Naaaaah. The main difference between OCIO-2.5 and 2.3 will be whether or not we can access a color space's defined interop_id value. For full support (and a properly working "strict" mode), folks will have to build with OCIO-2.5; otherwise, the proposed method will rely on equivalency heuristics. 

Beyond that, the code will be marginally simpler to maintain once we bump to OCIO-2.5, and there may be faster methods we can utilize if 2.5 is available; but, otherwise, what we have will work very nicely with 2.3. 

Re: performance: I'm confident that...
 - The oiiotool startup time difference will be negligible -- essentially nonexistent if we defer the instantiation of a "ColorInteropManager" class to its first actual use. It'll take as long to instantiate as it does to load the ocio://studio-config-latest, so we're talking not many milliseconds.
 - The hits will only occur when attempting to retrieve a matching color space for a given interop_id if the particular config does not define equivalent aliases or color space definitions. In other words, if OIIO is choking with a custom config for a given interop_id, improving performance is simply a matter of editing the custom config. 
 - We can make iterative improvements to the equivalency-testing heuristics. I've already experimented with some improvements, but I'd like to keep the initial PR as simple as I can.
 
Also, there are "strict" and "quick" flags which will help keep performance as fast as can be, at the cost of potentially failing to infer equivalent interop_ids <--> color spaces. 

> And I am curious to hear your experience with it! I go in with an open mind and am trying to learn more about where it's really useful.

Aye-aye! Yeah, given what you've written on Slack / 3DPro, I had an idea of how you'd feel about this. I figured a code-to-code translation of stuff I'd solely written would be about as safe as could be, but I wanted to make sure we had that conversation before I opened that can of worms...

brechtvl:
`IdentifyBuiltinColorSpace` also loops over color spaces in the config and creates processors, but it's faster because it excludes colorspaces that use e.g. 3D LUT files. I would expect that's where most of the overhead is in the PixelManager case, and I guess even when "strict" and "quick" are off such heuristics would still make sense. Not sure where performance would end up exactly.

Two things in the `IdentifyBuiltinColorSpace` heuristics that may not work well:
* It only checks active color spaces, but in earlier ACES configs the display colorspaces are inactive.
* It excludes any color space defined with a `ColorSpaceTransform` because it may contain hidden expensive transforms, but almost all color spaces in the PixelManager config use it.

zachlewis:
> `IdentifyBuiltinColorSpace` also loops over color spaces in the config and creates processors, but it's faster because it excludes colorspaces that use e.g. 3D LUT files. I would expect that's where most of the overhead is in the PixelManager case, and I guess even when "strict" and "quick" are off such heuristics would still make sense. Not sure where performance would end up exactly.


Yeah, the 3D LUT thing was one of the improvements for the "slow" method I was thinking about. Another is a quick check to see if we're trying to match linear color spaces; and another is to see if we can limit the search to a common set of color spaces that share an "encoding" value (if defined).


> Two things in the `IdentifyBuiltinColorSpace` heuristics that may not work well:
> 
> * It only checks active color spaces, but in earlier ACES configs the display colorspaces are inactive.

Indeed. That's one of the reasons for the "slow" method.

> * It excludes any color space defined with a `ColorSpaceTransform` because it may contain hidden expensive transforms,

I did not realize this...! with that in mind, I think I can make improvements to the default "interop-identities" config.

That being said, I think I can improve on the heuristic by permitting ColorSpaceTransforms, so long as neither the source nor the destination matches a name or alias of a colorspace that implements a 3D LUT (or contains a "$" in the value). We could certainly prioritize "simpler" color spaces over more complex ones.

lgritz:
> How about this... if I don't have a mergable PR ready or very very near ready by the beginning of the second week of January, let's merge this PR.

If @brechtvl is okay with this deadline, it's fine with me. But if he needs the Jan 1 patch release of 3.1 to include this, then I think we should merge and then replace later.

lgritz:
BTW, I believe that if you rebase on the current main, it should fix the broken CI job variants.

brechtvl:
I have no particular deadline for these color management changes, or need for them to land in a particular version. Just contributing because it seems like the right thing longer term.

zachlewis:
>  Just contributing because it seems like the right thing longer term.

And it couldn't be more appreciated :)
Thank you so much...

lgritz:
Where are we with this, @zachlewis ?

As a reminder, there was this:

> How about this... if I don't have a mergable PR ready or very very near ready by the beginning of the second week of January, let's merge this PR. I'm pretty sure I can get something in before before then, but no matter what, I'd like to get the PRs Brecht has submitted so far in and backported in time for the early Feb release.

zachlewis:
Making very good progress! I'll have a draft PR in on Monday, in time for this week's OCIO /  OIIO / CIF meetings. 

After a few minutes with Copilot, I decided it would be a better use of everyone's time if I bit the bullet and took time to do things the old-fashioned way, to better understand `color_ocio.cpp` (and C++). Upshot is, I revised my approach a little to keep everything private to ColorConfig; and now I'm working on improving the matching algorithm and keeping stuff performant with super heavy configs.

lgritz:
Human brain FTW!

lgritz:
Where are we on this, @zachlewis ?

zachlewis:
I apologize for the late reply!

I've had something I'd like to submit for a little while, but there's too much going on in my branch for a single PR; I've gone back and done a little refactoring to make stuff easier to submit and review and migrate back to OCIO in the future, and I'm breaking up what I've done into what will be two or three PRs. I'm gonna try to wrap those up this weekend.

brechtvl:
Not sure what the status is of the alternative solution, but updating this to latest main.

It's not really a blocker for my other work, and probably not many people are mixing older OpenColorIO configs with new OpenImageIO in practice. It just would make things work better in those cases.

lgritz:
@brechtvl I think you were not present at the last OIIO TSC meeting two weeks ago, but @zachlewis laid out that he has a series of color related PRs almost ready to unleash, so I want to be cautious about not doing merges where you two are going to stomp on each other's work or are working on things redundantly. Let's try to get his eyes on these and some more guidance about what specifically he has coming.

lgritz:
Sorry, @brechtvl and @zachlewis, I kind of lost track of the status of this one. Is this one that should be merged? Or is it superseded by things Zach is working on? Independent of the other recently merged color management changes, or likely to interfere?

brechtvl:
There was no agreement to merge this PR so far.

Personally I would recommend to merge this, since I feel it's trivial and would not interfere, and the timeline for landing Zach's more ambitious changes is unclear to me. Since we're actively testing against older OpenColorIO versions it would be nice if the color interop functionality worked fully with them.

But it will probably be superseded by Zach's changes, assuming that approach will have good enough performance. It would just be a secondary way to identify a matching color space.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2308,94 +2308,98 @@ enum class CICPRange : int {
...
-    { "srgb_p3d65_scene", CICPPrimaries::P3D65, CICPTransfer::sRGB,
-      CICPMatrix::BT709 },
-    { "g22_adobergb_scene" },
-    { "data" },
-    { "unknown" },
+    { "lin_ap1_scene", "lin_ap1" },
+    { "lin_ap0_scene", "lin_ap0" },
+    { "lin_rec709_scene", "lin_rec709", CICPPrimaries::Rec709,
+      CICPTransfer::Linear, CICPMatrix::BT709 },
+    { "lin_p3d65_scene", "lin_p3d65", CICPPrimaries::P3D65,
```
zachlewis, line 2346: - lin_rec709_display, lin_p3d65_display, and llin_rec2020_display are now official CIF colorInteropIDs, color spaces, and should probably be matched to the CICP enums. 
- (also, we should use NCL, not CL, for the matrix; constant-luminance is rarely supported.anywhere afaik)
- the xyz color spaces should use a CICPMatrix of 0, just like RGB --  not unspecified.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2308,94 +2308,98 @@ enum class CICPRange : int {
...
+      CICPTransfer::PQ, CICPMatrix::Rec2020_NCL },
+    { "pq_rec2020_display", "rec2100_pq_display", CICPPrimaries::Rec2020,
+      CICPTransfer::PQ, CICPMatrix::Rec2020_NCL },
+    { "hlg_rec2020_display", "rec2100_hlg_display", CICPPrimaries::Rec2020,
+      CICPTransfer::HLG, CICPMatrix::Rec2020_NCL },
     // No CICP mapping to keep previous behavior unchanged, as Gamma 2.2
     // display is more likely meant to be written as sRGB. On read the
     // scene referred interop ID will be used.
-    { "g22_rec709_display",
+    { "g22_rec709_display", nullptr
```
zachlewis, line 2388: I think for 3.2, we should switch to explicitly writing g22_rec709_display with the gamma-2.2 CICP trc=4, and explicitly reading it in-kind.
Which isn't to say we shouldn't allow for a way to impose that discontinuity -- I just don't think it should be our default behavior, because it's tricky to reason about. Things are a lot easier to troubleshoot if OIIO's default mode is to write g22 as trc=4, and srgb as trc=13 explicitly -- and from that baseline, it's easier for config authors to bake in something _other than_ what's advertised by CICP in order to elicit surround compensation and/or people having different ideas about what an "sRGB display" is. I can go more into this later. But for now, I wanted to see if you were cool with that, or if we should circle back to this outside of this PR?

zachlewis (approved):
Absolutely merge this! I apologize, I should have been more clear earlier. 

(I have a couple of nitpicky things noted for some edge cases, but nothing show-stopping)

No worries about stuff superseding other stuff  -- everything I'm doing is just another way of doing what Brecht is doing; and everything I'm doing is entirely under-the-hood, and can be swapped into place when the time comes. But as far as this PR is concerned, this does what it needs to do, and closes the reader-to-writer end-to-end flow of different types of color metadata supported by different kinds of plugins and formats <--> color Interop ID string representation <--> ocio 2.3+ implementation! This is magnificent, and very satisfying to have, and I should have merged this long ago. 


This is great, Brecht -- I appreciate you switching to using the display-referred CIF colorInteropIDs for the CICP permutations.

lgritz:
I think what I'd like to do is merge this now, and Brecht can address the "nits" in subsequent work (either rolling it into something big, or just tiny focused PRs to fix those specific things before the larger overhauls are ready).

Sound good?

zachlewis:
sounds good! I'm happy to care of my nits in future PRs :)

src/libOpenImageIO/color_ocio.cpp
brechtvl: Make sense to me. I'm assuming you will be the one to submit the PR for this given your last comment (I'm happy to do it otherwise).

zachlewis: Nah, I'll take care of it; I just wanted to make sure you won't be surprised by the behavior change, because I'm probably going to tuck the changes into other CICP related functionality, and frame OIIO's "default" CICP read / write policies around a 1:1 alignment with what CIF recommends, at least as a starting point; and from there, we can open the door to distinguishing OIIO's default behavior to CIF's, if the community feels like it; and it'll also make it easier to represent policy decisions in terms of custom color spaces and file rules and interop-ids and string aliases that can be hermetically injected into most OCIO config graphs... which, I hope, will help make things easier for contributors (and Larry) down the line, from an ease of implementation and tech-debt perspective!)

src/libOpenImageIO/color_ocio.cpp
brechtvl: Fine with me if you want to change this.

PR #4972, opened 2025-12-01
feat: Support display color space interop IDs in I/O

brechtvl:
## Description

* Display referred interop IDs are now chosen over scene referred ones by
  default on file read. This more closely matches the standards for CICP, ICC
  and other color metadata. Most images will now have `oiio:ColorSpace` set to
  `srgb_rec709_display` rather than `srgb_rec709_scene` on file read.

* To change the behavior to prefer scene default (as may be appropriate for
   textures), set `oiio:ImageStateDefault` to `scene` in `config` for
   `ImageInput.open` or with `oiiotool --iconfig`.

* There is also a new `ImageCache` attribute `image_state_default`. The default
  value here is also `display` as the `ImageCache` is used through OIIO for various
  purposes, but specifically for texture systems `scene` may be a better choice.

* Both `srgb_rec709_scene` and `srgb_rec709_display` are now (distinct) built-in
  color spaces. The built-in `sRGB` color space name remains an alias of
  `srgb_rec709_scene`.

  Documentation was updated to discourage the use of `sRGB` in favor of
  the more specific interop IDs.

* Previously only `srgb_rec709_scene` was recognized as sRGB for file metadata
  and display, now `srgb_rec709_display` and `g22_rec709_display` are treated as
  sRGB as well.

  The reason for `g22_rec709_display` behavior is that this type of display is
  often used to correct for the discrepancy where images are encoded as sRGB but
  usually decoded as gamma 2.2 by the physical display. By encoding it as gamma
  2.2 and claiming it's sRGB the transfer functions cancel out exactly.

* `g24_rec709_display` is now recognized as having gamma 2.4, and the name
  `g24_rec709_scene` was replaced with `ocio:g24_rec709_scene` since the
  former is not an official interop ID.

* This change also fixes an issue where tga and rla were incorrectly using
  `g22_rec709` and `g18_rec709` without `_scene` suffix, and not handling
  `g24_rec709`.

* `cio:itu709_rec709_scene` and `ocio:lin_ciexyzd65_display` were added to complete
  the list of interop IDs in OCIO configs that match a CICP.

* `ColorConfig` now includes inactive colorspaces. In older ACES configs (like
  v2.1.0 bundled with OpenColorIO 2.3), the display color spaces are inactive.
  But we now heavily rely on them.

## Tests

* Many test outputs were changed due to the new display referred default.
* Added test for `g22_rec709_display` behavior.
* Added test for `color:prefer_image_state`.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/doc/imagebufalgo.rst
+++ b/src/doc/imagebufalgo.rst
@@ -2849,12 +2849,14 @@ Color space conversion
        .. code-tab:: c++
 
           ImageBuf Src ("tahoe.jpg");
-          ImageBuf Dst = ImageBufAlgo::colorconvert (Src, "sRGB", "acescg", true);
+          ImageBuf Dst = ImageBufAlgo::colorconvert (Src, "srgb_rec709_scene",
+                                                     "acescg", true);
 
        .. code-tab:: py
 
           Src = ImageBuf("tahoe.jpg")
-          Dst = ImageBufAlgo.colorconvert (Src, "sRGB", "acescg", True)
+          Dst = ImageBufAlgo.colorconvert (Src, "srgb_rec709_scene", "acescg",
+                                           True)
 
        .. code-tab:: bash oiiotool
 
@@ -2950,14 +2952,15 @@ Color space conversion
        .. code-tab:: c++
 
           ImageBuf Src ("tahoe.exr");
-          ImageBuf Dst = ImageBufAlgo::ociodisplay (Src, "sRGB", "Film", "lnf",
-                                                    "", true, "SHOT", "pe0012");
+          ImageBuf Dst = ImageBufAlgo::ociodisplay (Src, "srgb_rec709_scene",
+                                                    "Film", "lnf", "", true,
+                                                    "SHOT", "pe0012");
 
        .. code-tab:: py
 
           Src = ImageBuf("tahoe.jpg")
-          Dst = ImageBufAlgo.ociodisplay (Src, "sRGB", "Film", "lnf",
-                                          "", True, "SHOT", "pe0012")
+          Dst = ImageBufAlgo.ociodisplay (Src, "srgb_rec709_scene", "Film",
+                                          "lnf", "", True, "SHOT", "pe0012")
 
        .. code-tab:: bash oiiotool
 
--- a/src/doc/pythonbindings.rst
+++ b/src/doc/pythonbindings.rst
@@ -3610,7 +3610,8 @@ Color manipulation
     .. code-block:: python
 
         Src = ImageBuf ("tahoe.jpg")
-        Dst = ImageBufAlgo.colorconvert (Src, "sRGB", "scene_linear")
+        Dst = ImageBufAlgo.colorconvert (Src, "srgb_rec709_scene",
+                                         "scene_linear")
 
 
 
@@ -3660,7 +3661,7 @@ Color manipulation
     .. code-block:: python
 
         Src = ImageBuf ("tahoe.exr")
-        Dst = ImageBufAlgo.ociodisplay (Src, "sRGB", "Film", "lnf",
+        Dst = ImageBufAlgo.ociodisplay (Src, "srgb_rec709_scene", "Film", "lnf",
                                   context_key="SHOT", context_value="pe0012")
 
 
--- a/src/doc/stdmetadata.rst
+++ b/src/doc/stdmetadata.rst
@@ -149,9 +149,11 @@ Color information
     - `"lin_ap1_scene"`, `"ACEScg"` :  ACEScg color space encoding.
     - `"lin_ap0_scene"` :  ACES2065-1, the recommended ACES space for
       interchange and archiving.
-    - `"srgb_rec709_scene"` : Using standard (piecewise) sRGB response and
+    - `"srgb_rec709_display"` : Using standard (piecewise) sRGB response and
       primaries. The token `"sRGB"` is treated as a synonym.
-    - `"g22_rec709_scene"` : Rec709/sRGB primaries, but using a response curve
+    - `"srgb_rec709_scene"` : Same response and primaries as
+      `"srgb_rec709_display"` but for scene referred images.
+    - `"g22_rec709_display"` : Rec709/sRGB primaries, but using a response curve
       corresponding to gamma 2.2.
 
     Additionally, `"scene_linear"` is a role that is appropriate for color
--- a/src/doc/stdmetadata.rst
+++ b/src/doc/stdmetadata.rst
@@ -150,7 +150,8 @@ Color information
     - `"lin_ap0_scene"` :  ACES2065-1, the recommended ACES space for
       interchange and archiving.
     - `"srgb_rec709_display"` : Using standard (piecewise) sRGB response and
-      primaries. The token `"sRGB"` is treated as a synonym.
+      primaries. The token `"sRGB"` is treated as a synonym, but it is
+      recommended to use the more specific interop ID.
     - `"srgb_rec709_scene"` : Same response and primaries as
       `"srgb_rec709_display"` but for scene referred images.
     - `"g22_rec709_display"` : Rec709/sRGB primaries, but using a response curve
--- a/src/dpx.imageio/dpxoutput.cpp
+++ b/src/dpx.imageio/dpxoutput.cpp
@@ -18,6 +18,8 @@
 #include <OpenImageIO/strutil.h>
 #include <OpenImageIO/typedesc.h>
 
+#include "imageio_pvt.h"
+
 OIIO_PLUGIN_NAMESPACE_BEGIN
 
 
@@ -436,18 +438,14 @@ DPXOutput::prep_subimage(int s, bool allocate)
     m_desc = get_image_descriptor();
 
     // transfer function
-    const ColorConfig& colorconfig = ColorConfig::default_colorconfig();
-    std::string colorspace = spec_s.get_string_attribute("oiio:ColorSpace", "");
-    if (colorconfig.equivalent(colorspace, "lin_rec709_scene"))
-        m_transfer = dpx::kLinear;
-    else if (colorconfig.equivalent(colorspace, "srgb_rec709_scene"))
+    const float gamma = pvt::get_colorspace_rec709_gamma(spec_s);
+    if (pvt::is_colorspace_srgb(spec_s, false))
         m_transfer = dpx::kITUR709;
-    else if (colorconfig.equivalent(colorspace, "g22_rec709_scene")
-             || colorconfig.equivalent(colorspace, "g24_rec709_scene")
-             || colorconfig.equivalent(colorspace, "g18_rec709_scene")
-             || Strutil::istarts_with(colorspace, "Gamma"))
+    else if (gamma == 1.0f)
+        m_transfer = dpx::kLinear;
+    else if (gamma != 0.0f)
         m_transfer = dpx::kUserDefined;
-    else if (colorconfig.equivalent(colorspace, "KodakLog"))
+    else if (spec_s.get_string_attribute("oiio:ColorSpace") == "KodakLog")
         m_transfer = dpx::kLogarithmic;
     else {
         std::string dpxtransfer = spec_s.get_string_attribute("dpx:Transfer",
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -55,8 +55,6 @@ static int disable_builtin_configs = Strutil::stoi(
     Sysutil::getenv("OIIO_DISABLE_BUILTIN_OCIO_CONFIGS"));
 static OCIO::ConstConfigRcPtr ocio_current_config;
 
-static const ustring scene_us("scene");
-
 
 const ColorConfig&
 ColorConfig::default_colorconfig()
@@ -2198,13 +2196,13 @@ ColorConfig::get_color_interop_id(string_view colorspace) const
 
 string_view
 ColorConfig::get_color_interop_id(const int cicp[4],
-                                  const string_view prefer_image_state) const
+                                  string_view image_state_default) const
 {
     string_view other_interop_id = "";
     for (const ColorInteropID& interop : color_interop_ids) {
         if (interop.has_cicp && interop.cicp[0] == cicp[0]
             && interop.cicp[1] == cicp[1]) {
-            if (!Strutil::ends_with(interop.interop_id, prefer_image_state)) {
+            if (!Strutil::ends_with(interop.interop_id, image_state_default)) {
                 if (other_interop_id.empty()) {
                     other_interop_id = interop.interop_id;
                 }
@@ -2841,6 +2839,20 @@ ImageBufAlgo::colorconvert(span<float> color, const ColorProcessor* processor,
 void
 ColorConfig::set_colorspace(ImageSpec& spec, string_view colorspace) const
 {
+    OIIO::set_colorspace(spec, colorspace);
+}
+
+
+void
+ColorConfig::set_colorspace_rec709_gamma(ImageSpec& spec, float gamma) const
+{
+    OIIO::pvt::set_colorspace_rec709_gamma(spec, gamma, string_view());
+}
+
+
+void
+set_colorspace(ImageSpec& spec, string_view colorspace)
+{
     // If we're not changing color space, don't mess with anything
     string_view oldspace = spec.get_string_attribute("oiio:ColorSpace");
     if (oldspace.size() && colorspace.size() && oldspace == colorspace)
@@ -2864,45 +2876,6 @@ ColorConfig::set_colorspace(ImageSpec& spec, string_view colorspace) const
     spec.erase_attribute("oiio:Gamma");
 }
 
-
-
-void
-ColorConfig::set_colorspace_rec709_gamma(ImageSpec& spec, float gamma) const
-{
-    gamma = std::round(gamma * 100.0f) / 100.0f;
-    if (fabsf(gamma - 1.0f) <= 0.01f) {
-        set_colorspace(spec, "lin_rec709_scene");
-    } else if (fabsf(gamma - 1.8f) <= 0.01f) {
-        set_colorspace(spec, "g18_rec709_scene");
-        spec.attribute("oiio:Gamma", 1.8f);
-    } else if (fabsf(gamma - 2.2f) <= 0.01f) {
-        set_colorspace(spec,
-                       (OIIO::get_string_attribute("color:prefer_image_state")
-                        == scene_us)
-                           ? "g22_rec709_scene"
-                           : "g22_rec709_display");
-        spec.attribute("oiio:Gamma", 2.2f);
-    } else if (fabsf(gamma - 2.4f) <= 0.01f) {
-        set_colorspace(spec,
-                       (OIIO::get_string_attribute("color:prefer_image_state")
-                        == scene_us)
-                           ? "ocio:g24_rec709_scene"
-                           : "g24_rec709_display");
-        spec.attribute("oiio:Gamma", 2.4f);
-    } else {
-        set_colorspace(spec, Strutil::fmt::format("g{}_rec709_display",
-                                                  std::lround(gamma * 10.0f)));
-        spec.attribute("oiio:Gamma", gamma);
-    }
-}
-
-
-void
-set_colorspace(ImageSpec& spec, string_view colorspace)
-{
-    ColorConfig::default_colorconfig().set_colorspace(spec, colorspace);
-}
-
 void
 set_colorspace_rec709_gamma(ImageSpec& spec, float gamma)
 {
@@ -2914,35 +2887,45 @@ OIIO_NAMESPACE_3_1_END
 OIIO_NAMESPACE_BEGIN
 
 void
-pvt::set_colorspace_srgb(ImageSpec& spec, bool erase_other_attributes)
+pvt::set_colorspace_rec709_gamma(ImageSpec& spec, float gamma,
+                                 string_view image_state_default)
 {
-    string_view srgb_colorspace
-        = (OIIO::get_string_attribute("color:prefer_image_state")
-           == OIIO::v3_1::scene_us)
-              ? "srgb_rec709_scene"
-              : "srgb_rec709_display";
+    gamma = std::round(gamma * 100.0f) / 100.0f;
+    if (fabsf(gamma - 1.0f) <= 0.01f) {
+        set_colorspace(spec, "lin_rec709_scene");
+    } else if (fabsf(gamma - 1.8f) <= 0.01f) {
+        set_colorspace(spec, "g18_rec709_scene");
+        spec.attribute("oiio:Gamma", 1.8f);
+    } else if (fabsf(gamma - 2.2f) <= 0.01f) {
+        set_colorspace(spec, (image_state_default == "scene")
+                                 ? "g22_rec709_scene"
+                                 : "g22_rec709_display");
+        spec.attribute("oiio:Gamma", 2.2f);
+    } else if (fabsf(gamma - 2.4f) <= 0.01f) {
+        set_colorspace(spec, (image_state_default == "scene")
+                                 ? "ocio:g24_rec709_scene"
+                                 : "g24_rec709_display");
+        spec.attribute("oiio:Gamma", 2.4f);
+    } else {
+        set_colorspace(spec, Strutil::fmt::format("g{}_rec709_display",
+                                                  std::lround(gamma * 10.0f)));
+        spec.attribute("oiio:Gamma", gamma);
+    }
+}
+
+
+void
+pvt::set_colorspace_srgb(ImageSpec& spec, string_view image_state_default,
+                         bool erase_other_attributes)
+{
+    string_view srgb_colorspace = (image_state_default == "scene")
+                                      ? "srgb_rec709_scene"
+                                      : "srgb_rec709_display";
     if (erase_other_attributes) {
         spec.set_colorspace(srgb_colorspace);
     } else {
         spec.attribute("oiio:ColorSpace", srgb_colorspace);
     }
-}
-
-bool
-pvt::is_colorspace_srgb(const ImageSpec& spec, bool default_to_srgb)
-{
-    string_view colorspace = spec.get_string_attribute("oiio:ColorSpace");
-    if (default_to_srgb && colorspace.empty()) {
-        return true;
-    }
-
-    const ColorConfig& colorconfig(ColorConfig::default_colorconfig());
-    string_view interop_id = colorconfig.get_color_interop_id(colorspace);
-
-    // See the interop table above for why g22_rec709_display is treated as sRGB
-    return (interop_id == "srgb_rec709_scene"
-            || interop_id == "srgb_rec709_display"
-            || interop_id == "g22_rec709_display");
 }
 
 float
@@ -2978,6 +2961,23 @@ pvt::get_colorspace_rec709_gamma(const ImageSpec& spec)
     return spec.get_float_attribute("oiio:Gamma", 0.0f);
 }
 
+bool
+pvt::is_colorspace_srgb(const ImageSpec& spec, bool default_to_srgb)
+{
+    string_view colorspace = spec.get_string_attribute("oiio:ColorSpace");
+    if (default_to_srgb && colorspace.empty()) {
+        return true;
+    }
+
+    const ColorConfig& colorconfig(ColorConfig::default_colorconfig());
+    string_view interop_id = colorconfig.get_color_interop_id(colorspace);
+
+    // See the interop table above for why g22_rec709_display is treated as sRGB
+    return (interop_id == "srgb_rec709_scene"
+            || interop_id == "srgb_rec709_display"
+            || interop_id == "g22_rec709_display");
+}
+
 std::vector<uint8_t>
 pvt::get_colorspace_icc_profile(const ImageSpec& spec, bool /*from_colorspace*/)
 {
@@ -2991,12 +2991,13 @@ pvt::get_colorspace_icc_profile(const ImageSpec& spec, bool /*from_colorspace*/)
 }
 
 void
-pvt::set_colorspace_cicp(ImageSpec& spec, const int cicp[4])
+pvt::set_colorspace_cicp(ImageSpec& spec, const int cicp[4],
+                         string_view image_state_default)
 {
     spec.attribute("CICP", TypeDesc(TypeDesc::INT, 4), cicp);
     const ColorConfig& colorconfig(ColorConfig::default_colorconfig());
-    string_view interop_id = colorconfig.get_color_interop_id(
-        cicp, OIIO::get_string_attribute("color:prefer_image_state"));
+    string_view interop_id
+        = colorconfig.get_color_interop_id(cicp, image_state_default);
     if (!interop_id.empty())
         spec.attribute("oiio:ColorSpace", interop_id);
 }
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -2100,7 +2100,8 @@ struct ColorInteropID {
 // Mapping between color interop ID and CICP, based on Color Interop Forum
 // recommendations.
 constexpr ColorInteropID color_interop_ids[] = {
-    // Display referred interop IDs.
+    // Display referred interop IDs first so they are the default in automatic.
+    // conversion from CICP to interop ID.
     { "srgb_rec709_display", CICPPrimaries::Rec709, CICPTransfer::sRGB,
       CICPMatrix::BT709 },
     // Not all software interprets this CICP the same, see the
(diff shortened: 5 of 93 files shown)


brechtvl:
Draft because:
* ~~This depends on and includes the changes from #4964.~~

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2864,11 +2870,14 @@ pvt::get_colorspace_rec709_gamma(const ImageSpec& spec)
         || colorconfig.equivalent(colorspace, "scene_linear")
         || interop_id == "lin_rec709_scene")
         return 1.0f;
+    // See the interop table above for why g22_rec709_display is not treated as gamma
     else if (interop_id == "g22_rec709_scene")
         return 2.2f;
-    // Note g24_rec709_scene is not a standard interop ID
-    else if (colorconfig.equivalent(colorspace, "g24_rec709_scene"))
+    // Note g24_rec709_scene is not a standard interop
+    else if (colorconfig.equivalent(colorspace, "g24_rec709_scene")
```
zachlewis: Would you mind changing this to `ocio:g24_rec709_scene`?  This matches the interop_id used by the "Gamma 2.4 Encoded Rec.709" scene color space or whatever it's called in the latest studio config.

lgritz: Is g24_rec709_scene not a real interop tag?

lgritz: So here I am torn about this prefixing. Does OCIO recognize "ocio:g24_rec709_scene", or does this mean that something elsewhere in OIIO we need to look for, and strip out, the prefixes any time it needs to actually call OCIO? That seems pretty ugly, too. 

Convince me otherwise, but I'm inclined to think that the "oiio:ColorSpace" attrib can have any unadorned OCIO color space name, as long as it's either in the current config or the built-in studio config. Or am I thinking of it wrong?

doug-walker: The interop ID description [is here](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE/edit?usp=sharing). The "g24_rec709_scene" name is not on the [CIF texture ID list](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/01_TextureAssetColorSpaces/TextureAssetColorSpaces.md#summary-table--overview-of-the-recommendations). Therefore it needs a namespace prefix. The prefix we use in the OCIO configs for ACES is "ocio". You would never strip that out, it's part of the interop ID. Zach is trying to keep OIIO in sync with the usage in the released OCIO configs.

zachlewis: `ocio:g24_rec709_scene` matches an alias for, and the interop_id of, a color space that exists in the latest (v4.0) OCIO Studio config. 


Let me back up -- 

The "Studio Config" contains a superset of the color spaces contained by the "CG Config"; and the "CG Config" is what we're using as OIIO's default ("ocio://default"). 

Each of the 65ish color spaces in OCIO-2.5's `ocio://studio-config-latest` defines a unique interop_id whose value matches an alias for that color space. If the color space definition matches the identity of one of the standard CIF tokens, the interop_id value will match the corresponding CIF string. All other Color Spaces use a similar convention, but are prefixed with an `ocio` "namespace". In turn, these "ocio:"-prefixed interop_id values are reserved values for referencing color spaces defined in the builtin OCIO configs (or color spaces in other OCIO configs that are definitionally equivalent to those in the builtin configs.


Two examples snipped from `studio-config-v4.0.0_aces-v2.0_ocio-v2.5`:

```
  - !<ColorSpace>
    name: Gamma 2.2 Encoded Rec.709
    aliases: [g22_encoded_rec709, g22_rec709, Utility - Gamma 2.2 - Rec.709 - Texture, Gamma 2.2 Rec.709 - Texture, g22_rec709_tx, g22_rec709_scene]
    interop_id: g22_rec709_scene
    family: Utility
    equalitygroup: ""
    bitdepth: 32f
    description: Convert ACES2065-1 to 2.2 gamma-corrected Rec.709 primaries, D65 white point
    isdata: false
    categories: [file-io, texture]
    encoding: sdr-video
    allocation: uniform
    from_scene_reference: !<GroupTransform>
      name: AP0 to Gamma 2.2 Encoded Rec.709
      children:
        - !<MatrixTransform> {matrix: [2.52168618674388, -1.13413098823972, -0.387555198504164, 0, -0.276479914229922, 1.37271908766826, -0.096239173438334, 0, -0.0153780649660342, -0.152975335867399, 1.16835340083343, 0, 0, 0, 0, 1]}
        - !<ExponentTransform> {value: 2.2, style: pass_thru, direction: inverse}

  - !<ColorSpace>
    name: Gamma 2.4 Encoded Rec.709
    aliases: [g24_encoded_rec709, g24_rec709, rec709_display, Utility - Rec.709 - Display, Gamma 2.4 Rec.709 - Texture, g24_rec709_tx, ocio:g24_rec709_scene]
    interop_id: ocio:g24_rec709_scene
    family: Utility
    equalitygroup: ""
    bitdepth: 32f
    description: Convert ACES2065-1 to 2.4 gamma-corrected Rec.709 primaries, D65 white point
    isdata: false
    categories: [file-io, texture]
    encoding: sdr-video
    allocation: uniform
    from_scene_reference: !<GroupTransform>
      name: AP0 to Gamma 2.4 Encoded Rec.709
      children:
        - !<MatrixTransform> {matrix: [2.52168618674388, -1.13413098823972, -0.387555198504164, 0, -0.276479914229922, 1.37271908766826, -0.096239173438334, 0, -0.0153780649660342, -0.152975335867399, 1.16835340083343, 0, 0, 0, 0, 1]}
        - !<ExponentTransform> {value: 2.4, style: pass_thru, direction: inverse}
```

So, the "Gamma 2.2 Encoded Rec.709" color space can be referenced by its alias, "g22_rec709_scene", which matches both, the color space's "interop_id" value, as well as the name of the CIF token that describes this color space's mathematical "identity" and image state. 

Meanwhile, "Gamma 2.4 Encoded Rec.709" has the interop_id (and a matching alias) "ocio:g24_rec709_scene", because there is no CIF token that matches both the color space's identity and image state. 

Elsewhere in the config, under the `display_colorspaces` section, there's this:

```
  - !<ColorSpace>
    name: Rec.1886 Rec.709 - Display
    aliases: [rec1886_rec709_display, g24_rec709_display]
    interop_id: g24_rec709_display
    family: Display
    equalitygroup: ""
    bitdepth: 32f
    description: Convert CIE XYZ (D65 white) to Rec.1886/Rec.709, mirror neg. values
    isdata: false
    categories: [file-io]
    encoding: sdr-video
    interchange:
      amf_transform_ids: |
        urn:ampas:aces:transformId:v2.0:InvOutput.Academy.Rec709-D65_100nit_in_Rec709-D65_BT1886.a2.v1
        urn:ampas:aces:transformId:v2.0:Output.Academy.Rec709-D65_100nit_in_Rec709-D65_BT1886.a2.v1
    allocation: uniform
    from_display_reference: !<BuiltinTransform> {style: DISPLAY - CIE-XYZ-D65_to_REC.1886-REC.709 - MIRROR NEGS}
```

The Display ColorSpace named "Rec.1886 Rec.709 - Display" matches both the mathematical identity and image state described by the "g24_rec709_display" CIF token, and uses that value accordingly for its interop_id; and, like all the other color spaces in the config, each color space's interop_id value is also an alias of the color space itself. 

---

There are certain rules by which valid `colorInteropId` attributes / OCIO `interop_id` values must abide:
 - All interop_ids _other_ than the standard CIF tokens _must_ be prefixed with a _namespace_.
 - Namespaced interop_ids are formatted: "ns:interop_id_name" (the namespace is separated from the rest of the interop_id by a single colon).
  - Interop ids can only contain certain characters (regex: `[a-zA-Z-_~/*#%^+()[\]{}|:]+`)
  - interop ids can only contain, at most, a single ":" value.
  
---

> So here I am torn about this prefixing. Does OCIO recognize "oiio:g24_rec709_scene", or does this mean that something elsewhere in OIIO we need to look for, and strip out, the prefixes any time it needs to actually call OCIO? That seems pretty ugly, too.

We've briefly outlined guidelines in one of the CIF specs. I'll have to check the wording, but I believe the current heuristic is essentially:
 1. First check to see if the full string matches a name or alias of a color space in the current config.
 2. Else, if the full string matches the name or alias of a color space in "ocio://studio-config-latest", try to find and use a definitionally equivalent color space in the current config; and if no equivalent color space exists, either merge the color space from the builtin config into the current config, or use Config::getProcessor to transform from the specified color space in the bulitin config to a color space in the current config, depending on what you're trying to do
 
 
And then, in the future, there will be more complex mechanisms for searching for color spaces in external configs, given a namespaced interop_id (which is all stuff I'm hoping and assuming OpenColorIO itself will manage one day...)

> Convince me othearwise, but I'm inclined to think that the "oiio:ColorSpace" attrib can have any unadorned OCIO color space name, as long as it's either in the current config or the built-in studio config. Or am I thinking of it wrong?

I think it's probably okay if we incorporate functionality to allow users to assign builtin config names and aliases to "oiio:ColorSpace"; but I think the OIIO codebase should endeavor to avoid naively assigning hardcoded color space names to "oiio:ColorSpace" directly (although roles are okay). Partly because this is generally just a bad practice when developing with OCIO that provides opportunities for fail conditions, and partly because we want to make it explicit that our use of these hardcoded values assumes (or requires) downstream handling to ensure that IBAs are equipped to make sense of the value, given the ColorConfig the IBA is expected to use. 

I recommend that, in the codebase, we instead use the "colorInteropId" attribute to hold such "untranslated" / "config-aganostic" interop_id strings; and we use our TBD `interpret_metadata_as_color_space()` function as needed either to turn the "colorInteropId" value into an "oiio:ColorSpace" value valid for the ColorConfig that the IBA is using (i.e., by finding a matching color space for the string, as described above, making ad-hoc edits to the ColorConfigs to merge missing color spaces from a builtin config into the config used by the IBA); or, alternatively, instead of any merging of color spaces into configs, we simply create a cross-config transform from the builtin color space indicated by the "colorInteropId" value to the appropriate reference space in the IBA's active ColorConfig (using OCIO's uber-overloaded `Config::getProcessor(...)`), and then concatenate that cross-config transform to the transform that the IBA would normally create when using the reference space as the IBA's "src" color space.  

Man, I hope what I just typed made sense. 

Upshot is, whether using "colorInteropId" to store interop_id values that need special handling, or implementing logic to handle cases where the value of "oiio:ColorSpace" might not refer to a color space that actually exists in a given ColorConfig, we pretty much have to apply the same handling heuristics; and I just think it's cleaner and more explicit if we use the "colorInteropId" attribute internally the same way we would use it when reading EXRs.

doug-walker: > Else, strip the namespace prefix and the ":", and see if the remaining string matches a name or alias of a color space in the current config.

That was indeed at one point in the interop ID document I link to above. However, it was decided to remove it in the interest of making things simpler. As the document stands now, there is no search without the namespace part of the ID.

lgritz: Is the built-in studio config a strict superset of the default OCIO cg config these days? Should we make the studio config be the one that OIIO uses as a default / backstop? I'm not sure if there are performance implications to that. 

But the reason why the studio config is not the default is only because a couple years ago when I was looking at it, the main thing the studio config added was cameras that I thought "naive" users (those not working in a production environment with its own config) would not need. But if it has evolved to the point where a bunch of color spaces that people might encounter in files found in the wild, but that are are not pro-camera-specific (i.e. stuff that any old CICP tags might implicate) are in the studio config but not on the default cg config, maybe we should re-evaluate the choice?

doug-walker: Yes, the Studio is a strict superset of the CG config. But I think all the color spaces we've been discussing are already in the CG config?  Your original thinking on which config to use still seems correct to me, though as I think Zach mentioned somewhere it might be useful to customize the CG config for OIIO specific needs.

zachlewis: Welp, I'd like to imbue the ColorConvert IBA with the ability to "do the right thing" given a known "builtin interop_id" (i.e., the set of interop_id values associated  with each Studio config color space), regardless of whether the current ColorConfig actually contains a color space matching that name or associated definition. 

In other words, the entire set of Studio config color spaces will always be accessible to ColorConvert (except in rare cases where OCIO's heuristics cannot ascertain a way to bridge the active ColorConfig and the builtin Studio config). This means there's a reserved set of ColorConfig-agnostic strings that will reliably work with ColorConvert:

```python
import PyOpenColorIO as ocio

def get_builtin_interop_ids() -> list[str]:
  return [i.getInteropID() for i in ocio.Config.CreateFromFile("ocio://studio-config-latest").getColorSpaces()]
```
```python
print(sorted(get_builtin_interop_ids(), key = lambda s: s.startswith("ocio:")))
```
```
['srgb_rec709_display', 'g22_rec709_display', 'srgb_p3d65_display', 'srgbe_p3d65_display', 'g26_p3d65_display', 'g24_rec709_display', 'hlg_rec2020_display', 'pq_rec2020_display', 'pq_p3d65_display', 'lin_ap0_scene', 'lin_ap1_scene', 'srgb_rec709_scene', 'g18_rec709_scene', 'g22_rec709_scene', 'srgb_p3d65_scene', 'g22_adobergb_scene', 'srgb_ap1_scene', 'g22_ap1_scene', 'lin_adobergb_scene', 'lin_p3d65_scene', 'lin_rec2020_scene', 'lin_rec709_scene', 'data', 'ocio:acescc_ap1_scene', 'ocio:acescct_ap1_scene', 'ocio:adx10_apd_scene', 'ocio:adx16_apd_scene', 'ocio:applelog_rec2020_scene', 'ocio:arrilogc3_awg3_scene', 'ocio:lin_awg3_scene', 'ocio:arrilogc4_awg4_scene', 'ocio:lin_awg4_scene', 'ocio:bmdfilm5_wg5_scene', 'ocio:lin_bmdwg5_scene', 'ocio:davinci_dwg_scene', 'ocio:lin_dwg_scene', 'ocio:canonlog2_cgamutd55_scene', 'ocio:lin_cgamutd55_scene', 'ocio:canonlog3_cgamutd55_scene', 'ocio:djilog_dgamut_scene', 'ocio:lin_dgamut_scene', 'ocio:vlog_vgamut_scene', 'ocio:lin_vgamut_scene', 'ocio:redlog3g10_rwg_scene', 'ocio:lin_rwg_scene', 'ocio:slog3_sgamut3_scene', 'ocio:slog3_sgamut3cine_scene', 'ocio:slog3_sgamut3venice_scene', 'ocio:slog3_sgamut3cinevenice_scene', 'ocio:lin_sgamut3_scene', 'ocio:lin_sgamut3cine_scene', 'ocio:lin_sgamut3venice_scene', 'ocio:lin_sgamut3cinevenice_scene', 'ocio:g24_rec709_scene', 'ocio:itu709_rec709_scene']
```

lgritz: Indeed, I'm in favor of a general heuristic of "if it's not in the active config, try the broadest built-in config we have, and it's only an error if it still can't figure it out at that point."

brechtvl: I have changed it to `ocio:g24_rec709_scene`. But also made further changes to prefer display colorspaces, which means it will be `g24_rec709_display` by default.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2139,15 +2196,23 @@ ColorConfig::get_color_interop_id(string_view colorspace) const
...
+                if (other_interop_id.empty()) {
+                    other_interop_id = interop.interop_id;
+                }
+            } else {
+                return interop.interop_id;
+            }
         }
     }
-    return "";
+    return other_interop_id;
```
doug-walker, line 2214: Was other_interop_id properly initialized?

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2057,6 +2100,43 @@ struct ColorInteropID {
...
+      CICPMatrix::Unspecified },
+    { "pq_xyzd65_display", CICPPrimaries::XYZD65, CICPTransfer::PQ,
+      CICPMatrix::Unspecified },
+    // OpenColorIO interop IDs.
+    { "ocio:lin_ciexyzd65_display", CICPPrimaries::XYZD65, CICPTransfer::Linear,
+      CICPMatrix::Unspecified },
+
     // Scene referred interop IDs first so they are the default in automatic
     // conversion from CICP to interop ID. Some are not display color spaces
     // at all, but can be represented by CICP anyway.
```
doug-walker: Scene IDs are no longer first.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -581,14 +597,18 @@ ColorConfig::Impl::classify_by_name(CSInfo& cs)
...
-        || Strutil::iequals(cs.name, "srgb_tx")
-        || Strutil::iequals(cs.name, "srgb_texture")
-        || Strutil::iequals(cs.name, "srgb texture")
-        || Strutil::iequals(cs.name, "srgb_rec709_scene")
-        || Strutil::iequals(cs.name, "sRGB - Texture")
+    //
+    if (Strutil::iequals(cs.name, "srgb_rec709_display")
+        || Strutil::iequals(cs.name, "srgb_display")
+        || Strutil::iequals(cs.name, "sRGB - Display")
         || Strutil::iequals(cs.name, "sRGB")) {
```
doug-walker: Is moving "sRGB" to srgb_rec709_display going to be disruptive? Just wondering if it should stay as srgb_rec709_scene, or be based on prefer_image_state, as it does not seem directly related to CICP.

doug-walker (commented):
I did a quick pass through this one. I'm still an OIIO novice, but it seems like there are a lot of nice improvements in here.

src/libOpenImageIO/color_ocio.cpp
lgritz: I guess the question is: when a naive user asks to convert something to or from "sRGB", do they mean srgb_rec709_display or srgb_rec709_scene?

(I still struggle to understand under which practical, in-the-wild situations those would yield different pixel values, except for expert users who have defined an OCIO config that has an srgb color space with tone mapping, right?)

brechtvl: I think in general using `sRGB` should not be recommended anymore, and I could update the documentation for that. With that in mind, keeping it as `srgb_rec709_scene` would give better backwards compatibility for something that we are phasing out. The reason I set it to `srgb_rec709_display` was to match the default interpretation of sRGB image files. I'm happy to change it if backwards compatibility is preferred for this case.

---

My understanding is that with the ACES and other well designed configs, there is no difference for direct conversions between color spaces for `srgb_rec709_display` and `srgb_rec709_scene`.

The difference between these spaces is the default view transform for commands like `--ociodisplay`. For scene referred it will apply e.g. the ACES view transform, while for display referred it will effectively skip it and do a direct conversion to the chosen display color space.

These assumptions break with configs that:
* Incorrectly set an interop ID as an alias on a color space that does not match the CIF recommendation.
* Set `default_view_transform` (the connection between scene and display spaces, not the default view transform mentioned above) to something that's arguably invalid.

src/libOpenImageIO/color_ocio.cpp
brechtvl: I believe the default `string_view` constructor is semantically equivalent to an empty string, but I can add `string_view other_interop_id = "";` to make it more clear.

src/libOpenImageIO/color_ocio.cpp
doug-walker: Larry, the concern applies even when using the current built-in ACES configs. Brecht, I mostly agree with what you wrote but will add the caveat that I think it's perfectly reasonable for a config to use something other than Un-tone-mapped as the default_view_transform (I think it would still be considered well-designed). And I would add a third bullet point to your list:

The current built-in ACES configs contain the viewing_rules, which are intended to modify the default way images are displayed on a monitor, based on their color space. So for example, if an app wants to generate a thumbnail or preview for an image in a given color space, they would call:
`getDefaultView(const char * display, const char * colorspaceName)`. The view that is returned will differ between a color space of srgb_rec709_scene and srgb_rec709_display. The former will get an ACES Output Transform (because it is scene-referred) whereas the latter will get the equivalent of Un-tone-mapped (which would better match what someone would see in Photoshop).

doug-walker: But just to be clear, I have no objection to Brecht's updates to "sRGB" in the latest commit. I don't have a strong opinion on what is the preferred interpretation for that legacy name. I just wanted to provide some more context about what the practical difference is between the options.

brechtvl: > Brecht, I mostly agree with what you wrote but will add the caveat that I think it's perfectly reasonable for a config to use something other than Un-tone-mapped as the default_view_transform (I think it would still be considered well-designed).

Ah, from discussion in https://github.com/AcademySoftwareFoundation/OpenImageIO/discussions/4787#discussioncomment-15167195 I assumed such a config is considered invalid. I guess if you modify `default_view_transform` then using `oiiotool` and `maketx` correctly gets harder, but there's not much we can do about that.

brechtvl:
Regarding the current status for this PR. I'm not currently planning further changes. There's more refactoring that could done, for deferred color space determination and more unified code. But that's probably best left for another PR?

I think the main decisions to be taken are:
* [x] Defaut to scene or display referred? (Consensus seems to be display.)
* [x] Should `sRGB` be `srgb_rec709_scene` (backwards compatibility) or `srgb_rec709_display` (consistency with default image load) or even be controlled by `color:prefer_image_state`? (Consensus is `srgb_rec709_scene`)
* [ ] Write CICP and ICC even for scene referred color spaces? (In my opinion, yes)
* [ ] Is the special `g22_rec709_display` handling a good thing? Or too confusing or undesirable for users with a Gamma 2.2 workflow? Or we won't have a good answer soon, so remove the exception and defer the decision? (Unsure)

src/libOpenImageIO/color_ocio.cpp
lgritz: > I think in general using sRGB should not be recommended anymore,

Oh yes, 100% agreed. The only thing we should "recommend" are the official CIF names, or to use something they know is in their facility's OCIO config.

The sole concern here is what legacy aliases we accept and what they mean. Lots of scripts probably exist that use "sRGB", and it's really difficult to track down all the places anybody in a facility may have used it. In particular, there are probably many instances of things like

    oiiotool production.exr --tocolorspace sRGB -o web.jpg

and we would like that to continue working. I don't really care which CIF tag we decide that means, as long as web.jpg keeps making the same pixels.

brechtvl: The most backwards compatible is `sRGB` = `srgb_rec709_scene`, so I'll change it back to that. (Even if in your example it would be wrong, but wrong the same way as before, and only for certain OCIO configs that are probably uncommon.)

I'll write some example commands in the comments for posteriority.

zachlewis: I think this is the right approach, aligning "sRGB" to "srgb_rec709_scene". We'd probably want to discourage use of "sRGB".

> Ah, from discussion in https://github.com/AcademySoftwareFoundation/OpenImageIO/discussions/4787#discussioncomment-15167195 I assumed such a config is considered invalid. I guess if you modify default_view_transform then using oiiotool and maketx correctly gets harder, but there's not much we can do about that.

I apologize for the confusion... at some point, I deleted a few hundred words going into detail about `default_view_transform`, because I thought it would be most straightforward to frame `default_view_transform` as an immutable implementation detail determined by the color specialist responsible for defining the color spaces for the config.

In any case, if an OCIO config author is setting `default_view_transform` incorrectly, it'll likely be obvious in that the config will either be invalid (i.e., if a view name is specified instead of a view_transform name), or display-views would (very likely) look very obviously incorrect. `oiiotool` and `maketx` would suffer the same problems other applications would suffer with that OCIO config. 

There's rarely a reason for users, and never a reason for applications, to modify the value of `default_view_transform`, once set by the color specialist responsible for the config. @doug-walker , is that fair?

doug-walker: Here is more detail about the default_view_transform. OCIO has two primary methods of converting between color spaces: ColorSpaceTransform (which is for converting between color spaces of the same image state) and DisplayViewTransform (which is for converting between color spaces of different image states). However, those transforms try to be forgiving and not throw an exception if someone tries to use, e.g. a ColorSpaceTransform to convert from a scene-referred to a display-referred color space. In those scenarios, a View Transform must be inserted to allow the processing. That's the purpose of the default_view_transform. (It is not the default view, for which it is sometimes mistaken due to the similarity in the naming.)

It could plausibly be any View Transform in the config. For example, someone could make it the same as the default view (i.e., the first entry in the active_views) and that would be a perfectly fine choice. That would likely even be a better choice in some industries, but Un-tone-mapped makes the most sense for CG since it makes srgb_rec709_scene and srgb_rec709_display behave the same in most, though not all, circumstances.

I agree that this is something that is best decided by the config author and should not be modified by applications or end-users.

But I will stress that if the color space tag is correctly assigned, and the application is using the appropriate type of transform for the task at hand (ColorSpace vs. DisplayView), then the default_view_transform is never used. It is a fallback.

brechtvl:
I have made changes so preferring scene or display referred image state can be controlled per `ImageInput` and `ImageCache`, as I think that's the control you'd need in practice in multithreaded applications that may handle different types of images.

This required changes in many files though, so it's a bit ugly but I could not see a nice way to do it without breaking the API significantly. If deferred color space resolution is added at some point this can probably be simplified.

src/doc/imagebufalgo.rst
```diff
@@ -2950,14 +2952,15 @@ Color space conversion
        .. code-tab:: c++
 
           ImageBuf Src ("tahoe.exr");
-          ImageBuf Dst = ImageBufAlgo::ociodisplay (Src, "sRGB", "Film", "lnf",
-                                                    "", true, "SHOT", "pe0012");
+          ImageBuf Dst = ImageBufAlgo::ociodisplay (Src, "srgb_rec709_scene",
+                                                    "Film", "lnf", "", true,
+                                                    "SHOT", "pe0012");
```
zachlewis, line 2957: In this example, "sRGB" is actually the name of the display, not the source color space, so it doesn't make sense to replace with "srgb_rec709_scene" here... however, I wouldn't mind replacing "lnf"  (the source color space) with "srgb_rec709_scene". 
```suggestion
          ImageBuf Dst = ImageBufAlgo::ociodisplay (Src, "sRGB",
                                                    "Film", "srgb_rec709_scene", "", true,
                                                    "SHOT", "pe0012");
```

src/doc/imagebufalgo.rst
```diff
@@ -2950,14 +2952,15 @@ Color space conversion
...
 
        .. code-tab:: py
 
           Src = ImageBuf("tahoe.jpg")
-          Dst = ImageBufAlgo.ociodisplay (Src, "sRGB", "Film", "lnf",
-                                          "", True, "SHOT", "pe0012")
+          Dst = ImageBufAlgo.ociodisplay (Src, "srgb_rec709_scene", "Film",
+                                          "lnf", "", True, "SHOT", "pe0012")
```
zachlewis, line 2963: Same as above, "sRGB" is supposed to be the name of the display.

This suggestion is a twist on the above; it resolves the default view for the default display when the input color space ~= "srgb_rec709_scene"
```suggestion
          Dst = ImageBufAlgo.ociodisplay (Src, "", "",
                                          "srgb_rec709_scene", "", True, "SHOT", "pe0012")
```

src/doc/pythonbindings.rst
```diff
@@ -3660,7 +3661,7 @@ Color manipulation
     .. code-block:: python
 
         Src = ImageBuf ("tahoe.exr")
-        Dst = ImageBufAlgo.ociodisplay (Src, "sRGB", "Film", "lnf",
+        Dst = ImageBufAlgo.ociodisplay (Src, "srgb_rec709_scene", "Film", "lnf",
```
zachlewis, line 3664: Another twist: resolve the default view for the "sRGB" display when the input color space ~= "srgb_rec709_scene"
```suggestion
        Dst = ImageBufAlgo.ociodisplay (Src, "sRGB", "", "srgb_rec709_scene",
```

src/dpx.imageio/dpxoutput.cpp
```diff
@@ -436,18 +438,14 @@ DPXOutput::prep_subimage(int s, bool allocate)
...
-    else if (colorconfig.equivalent(colorspace, "g22_rec709_scene")
-             || colorconfig.equivalent(colorspace, "g24_rec709_scene")
-             || colorconfig.equivalent(colorspace, "g18_rec709_scene")
-             || Strutil::istarts_with(colorspace, "Gamma"))
+    else if (gamma == 1.0f)
+        m_transfer = dpx::kLinear;
+    else if (gamma != 0.0f)
         m_transfer = dpx::kUserDefined;
-    else if (colorconfig.equivalent(colorspace, "KodakLog"))
+    else if (spec_s.get_string_attribute("oiio:ColorSpace") == "KodakLog")
```
zachlewis, line 448: This doesn't seem very reliable. I'm fine with leaving the behavior intact for this PR -- I'd also be fine with removing this clause entirely for now. 

In the future, we could use the `oiio:ColorSpace` color space's `encoding` attribute to map to various dpx:Transfer constants. (DPX metadata isn't very reliable in my experience, beyond distinguishing "log" from "video" encodings)

src/doc/imagebufalgo.rst
```diff
@@ -2849,12 +2849,14 @@ Color space conversion
...
           ImageBuf Src ("tahoe.jpg");
-          ImageBuf Dst = ImageBufAlgo::colorconvert (Src, "sRGB", "acescg", true);
+          ImageBuf Dst = ImageBufAlgo::colorconvert (Src, "srgb_rec709_scene",
+                                                     "acescg", true);
 
        .. code-tab:: py
 
           Src = ImageBuf("tahoe.jpg")
-          Dst = ImageBufAlgo.colorconvert (Src, "sRGB", "acescg", True)
+          Dst = ImageBufAlgo.colorconvert (Src, "srgb_rec709_scene", "acescg",
```
zachlewis, line 2858:
```suggestion
          Dst = ImageBufAlgo.colorconvert (Src, "srgb_rec709_scene", "lin_ap1_scene",
```
(config-agnostic behavior)

src/doc/stdmetadata.rst
```diff
@@ -149,9 +149,12 @@ Color information
     - `"lin_ap1_scene"`, `"ACEScg"` :  ACEScg color space encoding.
     - `"lin_ap0_scene"` :  ACES2065-1, the recommended ACES space for
       interchange and archiving.
-    - `"srgb_rec709_scene"` : Using standard (piecewise) sRGB response and
-      primaries. The token `"sRGB"` is treated as a synonym.
-    - `"g22_rec709_scene"` : Rec709/sRGB primaries, but using a response curve
+    - `"srgb_rec709_display"` : Using standard (piecewise) sRGB response and
+      primaries. The token `"sRGB"` is treated as a synonym, but it is
+      recommended to use the more specific interop ID.
```
zachlewis, line 154: I'd like to deprecate treating "sRGB" as a synonym for either "srgb_rec709_display" or "srgb_rec709_scene"

zachlewis (commented):
Thanks for your patience, Brecht! 
I apologize, I started this review three weeks ago; I didn't realize I hadn't submitted my approval! The only stuff I have comments on pertain to documentation. 

I think I understand what you're trying to do with `oiio:imageStateDefault` etc -- if I'm understanding correctly, a use case might be a file-browser UI for picking scene-referred diffuse texture maps; and the goal is to prevent downstream handlers from automatically applying an inverse display-view transform instead of a simple color space transform when applying a deferred conversion to the rendering space. That's a totally reasonable use case, and I think exposing an attribute for controlling this on the spec-level is a pretty good way to handle this. 

Maybe, in the future, we can use the same attribute to convey information about a specific OCIO View to use by default for inverse display-view transforms.

Anyway, thanks for your work on this, and, again, I apologize for my super-slow response.

brechtvl:
Sorry for the late reply. Looking at this again after a while, I'm not really happy where this PR went. Particularly:
* The `image_state_default` stuff doesn't feel future proof and more of intermediate solution until there is more control over resolving colorspace.
* Changing the default from scene to display referred is quite significant breaking change. And it's not obvious to me that it's even the right one given that OpenImageIO is used a lot as a texturing system where I think display referred is the wrong default for textures.

I tried to make something more future proof by unifying color space resolve. But that brings additional complexity and breaking changes, given that the file formats are quite inconsistent in their handling of color spaces now.

Given that there is not much time before 3.2 is stable, my plan is to peel off a few simpler smaller changes to improve interop and docs (taking into account the pending review comments here).  For the bigger changes, I'll post a WIP branch with some question about the best direction later.

brechtvl:
Closing this one, I'm keeping track of the changes that replace this in #4980.

PR #4973, opened 2025-12-04
fix(webp): Move write_complete_data() to close() for proper scanline …

pmady:
…order support

The WebP output plugin incorrectly assumed that scanlines would be written in sequential order, triggering the final write when y == height - 1. This violated the write_scanline API contract which allows scanlines to be written in any order (the plugin even advertises 'random_access' support).

Changes:
- Remove premature write_complete_data() call from write_scanline()
- Move final encoding/writing to close() where it belongs
- Add m_image_complete flag to track write state
- Move buffer cleanup to after write_complete_data() in close()

Fixes the issue where writing scanlines out of order would result in incomplete or corrupted WebP files.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/webp.imageio/webpoutput.cpp
+++ b/src/webp.imageio/webpoutput.cpp
@@ -238,11 +238,6 @@ WebpOutput::write_scanline(int y, int z, TypeDesc format, const void* data,
     std::vector<uint8_t> scratch;
     data = to_native_scanline(format, data, xstride, scratch, m_dither, y, z);
     memcpy(&m_uncompressed_image[y * m_scanline_size], data, m_scanline_size);
-
-    /* If this was the final scanline, we are done. */
-    if (y == m_spec.height - 1) {
-        return write_complete_data();
-    }
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/webp.imageio/webpoutput.cpp
```diff
@@ -271,9 +268,16 @@ WebpOutput::close()
...
         ok &= write_scanlines(m_spec.y, m_spec.y + m_spec.height, 0,
                               m_spec.format, &m_uncompressed_image[0]);
-        std::vector<uint8_t>().swap(m_uncompressed_image);
     }
 
+    // Write the complete image data on close, not during write_scanline.
+    // This allows scanlines to be written in any order.
+    if (ok && !m_image_complete && m_uncompressed_image.size()) {
+        ok = write_complete_data();
+        m_image_complete = true;
```
jessey-git: Is the `m_image_complete` field necessary at all? This close() method will return early if we are not in the opened state so a double-close should already be guarded I think?

pmady: You'true The ioproxy_opened() check at line 262 already guards against double-close, and 
init() is called at line 282 which would reset state anyway. The m_image_complete flag is unnecessary. removing the unnecessary m_image_complete field. The ioproxy_opened() check at the start of 
close() already prevents double-close scenarios. Thanks

lgritz:
The CI failures with messages about filling the disk are unrelated to this PR. If you rebase on the current main (or cherry pick #4974 into this PR), it will contain the fix and hopefully everything else will pass.

pmady:
Rebased fix/webp-write-scanline-order on upstream/main which now includes #4974 (the disk space fix). Force pushed to 9dad1962

PR #4974, opened 2025-12-05
ci: Address tight disk space on GHA runners

lgritz:
- Mount the host's root under /host/root so it's visible in the containers.
- Remove stuff we don't need to free lots of disk space.
- While we're at it, arrange for the commands (copied from OSL) that allow the containers to see a CPU, when available.

This is freeing 20+ GB!

### DIFF
--- a/src/build-scripts/gh-installdeps.bash
+++ b/src/build-scripts/gh-installdeps.bash
@@ -9,14 +9,12 @@
 set -ex
 
 # Make extra space on the runners
-if [[ -e /host/root ]] ; then
-    df -h .
-    : ${CRUFT_TO_REMOVE:="/host/root/usr/share/dotnet /host/root/usr/local/lib/android /host/root/usr/local/.ghcup"}
-    time rm -rf $CRUFT_TO_REMOVE &
-    sleep 3
-    # rather than block, delete in background, but give it a few secs to start
-    # clearing things out before moving on.
-fi
+df -h .
+time rm -rf /usr/local/lib/android /host/root/usr/local/lib/android &
+sleep 3
+# rather than block, delete in background, but give it a few secs to start
+# clearing things out before moving on.
+# Other candidates, if we need it: /usr/share/dotnet /usr/local/.ghcup
 
 
 #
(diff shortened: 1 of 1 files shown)


.github/workflows/build-steps.yml
```diff
@@ -95,6 +95,10 @@ jobs:
     container:
       image: ${{ inputs.container }}
       volumes: ${{ fromJson( inputs.container_volumes ) }}
+      options: -v /:/host/root ${{ (contains(inputs.runner, 'gpu') && '-e NVIDIA_DRIVER_CAPABILITIES=compute,graphics,utility --gpus all') || '-e A=x' }}
+      # Extra options:
```
jfpanisset, line 99: Not used this conditional syntax before, I'm guessing that the `|| '-e A=x'` part just sets an innocuous env var `A=x` since you can't have an empty "else" part of the conditional?

jfpanisset (approved):
LGTM, hope it works for you!

jfpanisset:
Also note that on a system where the container image store is on a different filesystem than the root filesystem, running `df -h .` will show the amount of space on that device, whereas `df -h /host/root` will show the space on the root filesystem.

Of course we only care about freeing up disk space on the same filesystem as where the container image is stored, and this is a Ubuntu GHA runner specific hack anyway.

lgritz:
I think they're the same, but I fixed anyway.

.github/workflows/build-steps.yml
lgritz: I think so. That part is just ported from OSL and is what I needed to make it work.

jfpanisset:
I was able to shave around 1.3GB from the images and have re-released ci-oiio 2024.5 / 2025.4 / 2026.2 and here are the results:

```
                Before   Before    After    After
                Docker   On Disk    Docker    On Disk
                Hub                          Hub
- 2024.5 5.46GB / 12.5GB -> 4.09GB / 11.0GB
- 2025.4 5.67GB / 13.0GB -> 4.19GB / 11.4GB
- 2026.2 7.01GB / 15.6GB -> 5.32GB / 13.8GB
```

This should hopefully allow the `ci-oiio:2026` image to get pulled an expanded on a free runner with 14GB of free disk space.

lgritz:
Thanks, @jfpanisset. I believe I will need the same for ci-osl

jfpanisset:
All 2024.5 / 2025.4 / 2026.2 images have been re-released with the spurious Conan download cache purged.

PR #4975, opened 2025-12-08
ci: optimize install_homebrew_deps by coalescing installs

lgritz:
There seems to be a couple seconds overhead for each "brew install". Refactor install_homebrew_deps.bash to first assemble the list of all packages to install, and then do a single `brew install` for the whole list.

Also remove mention of packages that are pre-installed on the Mac GHA runners or that we don't need.

In the process, by using the `OIIO_BREW_INSTALL_PACKAGES` variable and only setting it if empty, we now have the ability to fully override the list of packages that brew will install on a per-job basis.


lgritz:
No objections for nearly a week + CI change only = merge

PR #4976, opened 2025-12-09
fix(idiff): Fail comparison when images have different dimensions

### LINKED ISSUE #4948: [BUG] idiff's perceptual comparison passes images with different dimensions and contents
**Describe the bug**

Running idiff's perceptual difference test on two images with different dimension and obvious content differences passes the test.

These are the images being compared:

Image

Image

OIIO 3.1.7.0 | MacOS/x86_64
    Build compiler: Apple clang 16.0 | C++17/201703
Dependencies: BZip2 1.0.8, FFmpeg 8.0, fmt 12.1.0, Freetype 2.14.1, GIF 5.2.2, Imath 3.2.2, JXL
    0.11.1, Libheif 1.20.2, libjpeg-turbo 3.1.2, LibRaw 0.21.4, libuhdr NONE, OpenColorIO 2.5.0,
    OpenEXR 3.4.2, openjph 0.24.5, OpenVDB NONE, PNG 1.6.50, pugixml 1.15, pybind11 3.0.1, Python3
    3.14.0, Robinmap 1.4.0, TBB 2022.3.0, TIFF 4.7.1, WebP 1.6.0, ZLIB 1.3.1

(Installed and upgraded to the latest version of OpenImageIO with homebrew)

**To Reproduce**

1. Download and extract the images from this ZIP file: 
[images.zip](https://github.com/user-attachments/files/23289778/images.zip)
2. In the directory with the two images, run `idiff -p -abs -o diff.jpg nav_buttons_new.jpg nav_buttons.jpg`

It outputs 

> Comparing "nav_buttons_new.jpg" and "nav_buttons.jpg"
> PASS

even though the images are clearly very different. This is the diff that is generates:

Image

Passing `-od` to the command makes it not generate any diff because the comparison passes.
The expected result would be that the comparison fails.

pmady:
Previously, idiff would not explicitly fail when comparing images with different dimensions. Both numeric and perceptual comparisons use roi_union which silently handles dimension differences by comparing against black pixels for non-overlapping regions.

This fix adds an explicit dimension check before any comparison to ensure images with different sizes or channel counts fail with ErrDifferentSize (exit code 3) and print a clear error message.

Fixes #4948

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_yee.cpp
+++ b/src/libOpenImageIO/imagebufalgo_yee.cpp
@@ -52,7 +52,7 @@ public:
         if (lev >= PYRAMID_MAX_LEVELS)
             return 0.0f;
         else
-            return level[lev].getchannel(x, y, 0, 1);
+            return level[lev].getchannel(x, y, 0, 0);
     }
 
 #if 0 /* unused */
@@ -65,7 +65,7 @@ public:
     float operator()(int x, int y, int lev) const
     {
         OIIO_DASSERT(lev < PYRAMID_MAX_LEVELS);
-        return level[lev].getchannel(x, y, 0, 1);
+        return level[lev].getchannel(x, y, 0, 0);
     }
 #endif
 
--- a/testsuite/diff/ref/out-fmt6.txt
+++ b/testsuite/diff/ref/out-fmt6.txt
@@ -16,7 +16,7 @@ Computing diff of "img1.exr" vs "img2.exr"
   121 pixels (2.95%) over 1e-06
 FAILURE
 Computing perceptual diff of "img1.exr" vs "img2.exr"
-  Max error  = 1.0 @ (5, 17, R)  values are 0.1, 0.1, 0.1 vs 0.1, 0.6, 0.1
+  Max error  = 10.0 @ (5, 17, R)  values are 0.1, 0.1, 0.1 vs 0.1, 0.6, 0.1
   121 pixels (2.95%) failed the perceptual test
 FAILURE
 Computing perceptual diff of "img1.exr" vs "img1.exr"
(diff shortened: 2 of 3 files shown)


lgritz:
It's very common (especially for openexr files, which support an explicit "display window" separately from a "pixel data window") for applications to "shrink wrap" the images by reducing the pixel data window to only encompass the non-zero part of the image. By definition, pixels outside the pixel window are simply black, and it would break a lot of workflows to suddenly start failing comparisons of shrink-wrapped image files. 

We should not consider images to differ in contents merely on the basis of their windows being different. We should only consider if the pixel values are different (and honoring the precept that pixels outside the data window exist, but are 0 in all channels).

I think the real problem of Issue #4948 is perhaps different than the initial diagnoss. Note that it works fine when not using the `-p` flag. So it has probably is a case of ImageBufAlgo::compare_Yee in particular misbehaving under those circumstances.

lgritz:
Um... you can see that all of the CI is failing, and it's because our testsuite specifically tests this case of ensuring that two images are considered the same even if they have different crop windows, provided that they are always 0 in the areas where the two images do not overlap their windows.

So I feel compelled to ask: Before submitting the PR, did you not either run the tests locally on your machine, or at least check the results of the CI run that happened when you pushed your changes to your fork of OIIO?

I understand that sometimes we're in a rush and stuff slips through the cracks, but it's really always worth the time to pause after you push to your fork, check the CI results there, and then only submit the PR if it passes all tests at that stage.

pmady:
Sure when i ran the CI worked, let me try locally and test again.

lgritz:
This doesn't make sense to me. There were only 2 reference outputs for this test before -- and one exists only to account for slightly different formatting for old fmt 6.0. Why should your fix require 7 additional reference outputs? That sounds fishy to me.

src/libOpenImageIO/imagebufalgo_yee.cpp
```diff
@@ -52,7 +52,7 @@ class GaussianPyramid {
         if (lev >= PYRAMID_MAX_LEVELS)
             return 0.0f;
         else
-            return level[lev].getchannel(x, y, 0, 1);
+            return level[lev].getchannel(x, y, 0, 0);
```
lgritz, line 55: It was this all along? Weird. Why did this ever work?

pmady: "It 'worked' in the sense that it didn't crash - getchannel() likely returns 0 for out-of-bounds channels. Since both pyramids returned the same incorrect value (0), the computed differences were always 0, making images appear identical. This explains why perceptual comparison was passing for clearly different images in #4948."

PR #4977, opened 2025-12-10
Speedup to detect the existence of files on Windows

wingfiring:
## Description
On Windows, `Filesystem::exists` forward the call to `std::filesystem::exists`, and finally calls `_std_fs_get_stats` . `_std_fs_get_stats`  is slow if the argument is a netwok path.

## Tests
No, performance related.

## Checklist:

  doesn't change the API.)
  nonconforming.

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -330,8 +330,15 @@ Filesystem::path_is_absolute(string_view path, bool dot_is_absolute)
 bool
 Filesystem::exists(string_view path) noexcept
 {
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


lgritz:
This looks fine to me. Does anybody with Windows experience wish to comment?

nrusch:
I'm certainly not a Win32 expert, but this seems correct based on my understanding and some spelunking in the MSVC STL source (which, in hindsight, is not an activity I would recommend).

lgritz (approved):

lgritz:
LGTM. The bleeding edge CI test failure is entirely unrelated and happening on all branches. I'm investigating separately.

PR #4987, opened 2025-12-29
docs: fix missing docs for `OIIO:attribute()` and `OIIO::getattribute()`

lgritz:
Rearrangements in 3.1 dropped the list of recognized attributes and failed to document the span varieties. We fix and also reword a lot of the descriptions for clarity and uniformity.

### DIFF
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -215,49 +215,12 @@ These helper functions are not part of any other OpenImageIO class, they
 just exist in the OpenImageIO namespace as general utilities. (See
 :ref:`sec-pythonmiscapi` for the corresponding Python bindings.)
 
-.. doxygenfunction:: OIIO::attribute(string_view, TypeDesc, const void *)
-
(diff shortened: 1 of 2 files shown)


jessey-git (approved):
Ah, alright, it was the `@defgroup` setup that was missed. Good to know.

lgritz:
> Ah, alright, it was the `@defgroup` setup that was missed. Good to know.

That's not quite it, actually.

The previous organization was that there were several varieties of `attribute()`. In the header, the first one had the overall long explanation, including the list of all the recognized attributes. The other ones had short explanations of how they differed. In the docs, each one was referenced explicitly, pulling in its attendant bit of documentation.

What really happened is that in the header, I made the new span-based version the "flagship" one with the full explanation, but I neglected to reference it in the docs, so the long description disappeared.

I could have fixed by just adding refs to the new functions to the docs, as I originally meant to. But while I was there, I took the opportunity to surround the whole collection with a group marker, and then include the lot of them with a single reference to the group, rather than need to refer to each function variant individually.

PR #4988, opened 2025-12-29
deps: Test against libraw 0.21.5

lgritz:


lgritz:
In anticipation of the imminent monthly patch release, I'm preemptively merging this, since it only changes CI test cases and not code.

PR #4994, opened 2026-01-07
Add Highway SIMD acceleration to ImageBufAlgo [add, sub, mul, div, mad, resample]

ssh4net:
Optional SIMD optimizations for selected ImageBufAlgo operations using the Google Highway library: • add/sub
• mul/div
• mad
• resample
Adds CMake and build system support, new implementation helpers, and developer documentation.

_Code mostly wrote using frontier Opus4.5 and Codex GPT5.2 High models with a strict rules._

### Checklist:

  behavior.
  testsuite**.
  on the GitHub runners.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ else ()
 endif ()
 option (${PROJ_NAME}_BUILD_TOOLS "Build the command-line tools" ON)
 option (${PROJ_NAME}_BUILD_TESTS "Build the unit tests" ON)
-option (OIIO_USE_HWY "Enable experimental Google Highway SIMD optimizations (if Highway is available)" OFF)
+set_option (OIIO_USE_HWY "Enable experimental Google Highway SIMD optimizations (if Highway is available)" OFF)
 set (OIIO_LIBNAME_SUFFIX "" CACHE STRING
      "Optional name appended to ${PROJECT_NAME} libraries that are built")
 option (BUILD_OIIOUTIL_ONLY "If ON, will build *only* libOpenImageIO_Util" OFF)
--- a/src/doc/imageioapi.rst
+++ b/src/doc/imageioapi.rst
@@ -378,7 +378,10 @@ inside the source code.
     image formats (uint8, uint16, int8, int16, uint32, int32, etc.) and provide
     additional speedup for scale-invariant operations (add, sub, min, max,
     absdiff) that can operate directly on integer data without float conversion.
-    (This was added in OpenImageIO 3.1.)
+
+    This was added in OpenImageIO 3.2.
+
+    NOTE: This is currently disabled by default while we are testing!
 
 .. cpp:var:: OPENIMAGEIO_PYTHON_LOAD_DLLS_FROM_PATH
 
--- a/src/libOpenImageIO/imageio.cpp
+++ b/src/libOpenImageIO/imageio.cpp
@@ -53,7 +53,14 @@ int png_linear_premult(0);
 int tiff_half(0);
 int tiff_multithread(1);
 int dds_bc5normal(0);
-int enable_hwy(1);  // Enable Google Highway SIMD optimizations by default
+#if OIIO_USE_HWY
+// hwy enabled at build time, but allow env var to override at runtime While
+// we're still testing, default to disabled. We intend to change the default
+// to be enabled in time for the release of 3.2.
+int enable_hwy = Strutil::stoi(Sysutil::getenv("OPENIMAGEIO_ENABLE_HWY", "0"));
+#else
+int enable_hwy = 0;  // Not enabled at build time
+#endif
 int limit_channels(1024);
 int limit_imagesize_MB(std::min(32 * 1024,
                                 int(Sysutil::physical_memory() >> 20)));
@@ -407,7 +414,9 @@ attribute(string_view name, TypeDesc type, const void* val)
         return true;
     }
     if (name == "enable_hwy" && type == TypeInt) {
+#if OIIO_USE_HWY
         enable_hwy = *(const int*)val;
+#endif
         return true;
     }
     if (name == "limits:channels" && type == TypeInt) {


lgritz:
I suspect you used LLM for some of this? Which is fine, but I think you should document in the PR description (commit comment) which tool you used and for what parts.

CMakeLists.txt
```diff
@@ -112,6 +112,20 @@ else ()
...
+#
+# Back-compat: honor -DUSE_HWY=OFF by mapping it to OIIO_USE_HWY if the latter
+# was not explicitly provided.
+if (DEFINED USE_HWY AND NOT DEFINED OIIO_USE_HWY)
+    set (OIIO_USE_HWY ${USE_HWY} CACHE BOOL
+         "Enable Google Highway SIMD optimizations (if Highway is available)" FORCE)
+else ()
+    option (OIIO_USE_HWY "Enable Google Highway SIMD optimizations (if Highway is available)" ON)
+endif ()
+
```
lgritz: Seems unnecessarily complex. Maybe just replace this whole thing with

```suggestion
option (OIIO_USE_HWY "Enable experimental Google Highway SIMD optimizations (if Highway is available)" OFF)
```

Also note that I changed the default to OFF, and described it as experimental, which I think is better at this point, in case there is an unforeseen issue.

src/cmake/externalpackages.cmake
```diff
@@ -225,6 +225,9 @@ if (USE_QT AND OPENGL_FOUND)
 endif ()
 
 
+# Google Highway for SIMD (optional optimization)
+checked_find_package (hwy ENABLE ${OIIO_USE_HWY})
```
lgritz: Curiously, the checked_find_package accepts an ENABLE parameter but doesn't use it! (That's my fault, I guess.) I made another suggestion about the top level enabling variable, so I would recommend changing this to




```suggestion
if (OIIO_USE_HWY)
    checked_find_package (hwy)
endif ()
```

src/include/OpenImageIO/platform.h
```diff
@@ -39,6 +39,7 @@
 #endif
 
 #ifdef _MSC_VER
+#    include <malloc.h>  // for alloca
```
lgritz: This doesn't make sense to me. We already use alloca extensively without trouble, why did this need to be added now?

src/include/OpenImageIO/platform.h
```diff
@@ -305,8 +306,15 @@
...
-#if defined(__GNUC__)
+#if defined(__has_include)
+#    if __has_include(<alloca.h>)
+#        include <alloca.h>  // for alloca (when available)
+#    endif
+#endif
+#if defined(__GNUC__) || defined(__clang__)
 #    define OIIO_ALLOCA(type, size) (assert(size < (1<<20)), (size) != 0 ? ((type*)__builtin_alloca((size) * sizeof(type))) : nullptr)
+#elif defined(_MSC_VER)
+#    define OIIO_ALLOCA(type, size) (assert(size < (1<<20)), (size) != 0 ? ((type*)_alloca((size) * sizeof(type))) : nullptr)
```
lgritz: Why? I think this definition is identical to the `#else` clause below, so the MS case would already be handled properly, right?

src/libOpenImageIO/imagebufalgo_addsub.cpp
```diff
@@ -10,6 +10,10 @@
 #include <iostream>
 #include <limits>
 
+#if defined(_WIN32)
+#    include <malloc.h>  // for alloca
+#endif
+
```
lgritz: why? It always worked before

src/libOpenImageIO/imagebufalgo_addsub.cpp
```diff
@@ -58,6 +67,298 @@ add_impl(ImageBuf& R, const ImageBuf& A, cspan<float> b, ROI roi, int nthreads)
 
 
 
+#if defined(OIIO_USE_HWY) && OIIO_USE_HWY
```
lgritz: On the cmake side, you see to *always* define OIIO_USE_HWY. So these can just be


```suggestion
#if OIIO_USE_HWY
```

Or else, if you want cmake to only add the definition at all if enabled, then `#ifdef` is what should be in the cpp.

src/libOpenImageIO/imagebufalgo_addsub.cpp
```diff
@@ -58,6 +67,298 @@ add_impl(ImageBuf& R, const ImageBuf& A, cspan<float> b, ROI roi, int nthreads)
...
+        }
+    });
+    return true;
+}
+
+template<class Rtype, class Atype, class Btype>
+static bool
+add_impl_hwy(ImageBuf& R, const ImageBuf& A, const ImageBuf& B, ROI roi,
+             int nthreads)
+{
```
lgritz, line 85: I haven't done a line-by-line comparison, but it seems to me that the only difference between add_impl_hwy, sub_impl_hwy, and mul_impl_hwy is likely going to be 

```
[](auto d, auto a, auto b) { return hn::Add(a, b); }
```

versus that one lambda changing for Sub and Mul.

I would love for even the initial commit to reduce this whole thing to a shared hwy_binary_perpixel_op() template that takes the lambda housing the op kernel as a templated parameter.

src/libOpenImageIO/imagebufalgo_addsub.cpp
```diff
@@ -58,6 +67,298 @@ add_impl(ImageBuf& R, const ImageBuf& A, cspan<float> b, ROI roi, int nthreads)
...
+                                                   return hn::Add(a, b);
+                                               });
+            } else {
+                // Process pixel by pixel (scalar fallback for strided channels)
+                for (int x = roi.xbegin; x < roi.xend; ++x) {
+                    Rtype* r_ptr = ChannelPtr<Rtype>(Rv, x, y, roi.chbegin);
+                    const Atype* a_ptr = ChannelPtr<Atype>(Av, x, y,
+                                                           roi.chbegin);
+                    const Btype* b_ptr = ChannelPtr<Btype>(Bv, x, y,
+                                                           roi.chbegin);
```
lgritz: I think we should benchmark the strided case and see how it compares to the contiguous case and the full scalar fallback that we've always had. If there is no big speed gain, I would be in favor of eliminating this whole clause and let non-contiguous strides use the old scalar path, then there is much less template expansion for hwy in the cases where there is not a large gain to be had. Note that this means that the "to hwy or not to hwy" test would need to test contiguity in addition to just localpixels().

src/libOpenImageIO/imagebufalgo_addsub.cpp
ssh4net: My fault, forgot cleanup.

lgritz:
@ssh4net It's been a while since this PR has been updated, but after your last push, it's failing to build. Can you please rebase on main, fix so it passes CI, and ensure that there is a DCO sign-off on each commit? I would like to proceed with this in some form.

ssh4net:
@lgritz sure! Give me a bit of time. I have added some fixes based on discussion above, but not verified fully yet, and switched to other projects 😅
Will check, rebase and push soon.

ssh4net:
@lgritz, please check. lateast code. It now should support SIMD path for RGB ROI in case of RGBA inputs.

MAD-Specific Note
mad can legitimately differ from the scalar a*b + c by tiny amounts when HWY uses FMA (MulAdd), because fused vs non-fused evaluation changes rounding. In normalized integer formats that often show up as exactly 1 LSB (e.g., uint16 max error 1/65535 = 1.52588e-05), and for uint32 it can reflect float32 precision limits after scaling/rounding.

test tool in the archive
[hwy_test.zip](https://github.com/user-attachments/files/25519021/hwy_test.zip)

And I see CL errors in this repo when in my repo all of them run successfully.

1div0:
Small error in the maximum error  1/65536 instead of the 1/65535. This is really MAD.

lgritz:
@ssh4net I'm sorry, I think I didn't realize how big a change you made after my last comments. I actually came back to this today thinking that maybe I'd pitch in and do some of the refactoring that I suggested, but... you had already done it! I really like what you've done here, it's much simpler at the IBA call sites than last time I viewed your earlier drafts.

There are just a handful of tests failing, and I'm looking into it.

I will update you soon, but I wanted you to know that this is still a very active PR and I do expect us to merge this with only a few other minor changes.

ssh4net:
@lgritz sure, no problem.

lgritz:
Would you mind if I squashed these, rebased on main, fix the conflicts, and force push the results so we have a clean single commit to build on? I know why the tests are failing. (At least some of them.)

ssh4net:
@lgritz sure, do what you think is needed.

lgritz:
I squashed, rebased (fixing conflicts), and pushed, so now this is directly versus the current main.

I made the following additions so far, amended to your work:

* Make the build-time cmake option OIIO_USE_HWY able to be overridden by environment (by using `set_option()`)
    
* Let env variable OPENIMAGEIO_ENABLE_HWY override runtime hwy enabling. (It was documented to do so, but not yet implemented.)
    
* Clarify that hwy support will be added for 3.2, not 3.1. And that for right now, it defaults to OFF until we have worked out all the bugs and gain more confidence in it. The intention is for it to be enabled by default in time for the release of OIIO 3.2.
    
* When hwy is disabled at build time, don't allow attribute() to appear to enable it.

Note in particular that I've changed the default behavior to
1. Build with support if hwy is present AND it is enabled at build time with `-DOIIO_USE_HWY=ON`.
2. Enable runtime support only if either (a) env variable OPENIMAGEIO_ENABLE_HWY is nonzero, or (b) `OIIO::attribute("enable_hwy")` is set by the program to nonzero.

I believe that makes it totally safe, even with any current flaws, because people need to take action to build it, and also to run it even if it was enabled at build time. The intent, of course, is to switch both of those defaults to "enable" by the time we ship 3.2, but we have some more work and fixes to do first. 

I think that with this, I would like to propose that we merge what we have here (assuming that CI passes), and then I will have some additional fixes coming shortly. But then with this merged, all the changes from here on out should be small and easy to review.

lgritz:
Like I said, the default as I have it here now is that it's disabled (making bugs harmless until we fix them).

I think that after merging, the remaining tasks are as follows:

1. There are some test failures currently when it's enabled, and I think I have identified at least one of the causes: the test for when we can use the hwy code paths (versus when we fall back to the old pixel-by-pixel scalar code) is not quite right. You test for the images all being local pixels and for those pixels to be contiguous across a scanline, but that's not enough! It turns out that also all of the input and output ImageBufs need to have their valid pixel data regions fully overlap the ROI that we are operating on, because the hwy code paths are not considering pixels outside the data windows. Because the ImageBuf::Iterator knows how to handle this (automatically pointing to black pixel data when "off the edge"), the logic in the old pixel loops didn't need to explicitly handle this, so it was probably not apparent to you looking at the old code that it was a condition we needed to account for.

   I have a fix cooking for that, though I also need to think about if there are any other hidden conditions that haven't occurred to either of us yet.

2. The next thing I want to do (and I have a plan for how to do it) is set up the testsuite and the CI to be sure we are testing both scalar and hwy code.

3. At that point, and if there are no apparent failures left, we can think about switching the build and runtime defaults to enable hwy.

4. It would be great to have "self-builders" set up for hwy so that we can use it even if not found already installed on the system.

5. I want to re-benchmark everything, and make sure that we're only enabling hwy for the particular cases where it seems to speed things up, on the various platforms we test on. We want it enabled by default, and to be confident that it only kicks in for cases where it will improve performance. 

   As part of this, I think there will be some cases where it's slow and we probably should expect it. There may also be cases where we *think* it should be faster, but it's not, and we want to take the time to investigate that and make sure there isn't some simple reason that we can fix.

At that point, I think we an return to looking at other IBA functions and trying to figure out which cases have benefits to hwy code paths and enhance them one by one. (With some attention to a priority that makes sense -- a combination of which look most promising for SIMD speedups and which are the most commonly used operations.)

But I'm gating this all on your approving the small number of changes from me that I attached here; I don't want to merge them into your PR without you approving it. If we are both satisfied with what we have here for the moment, then I will merge and move on to the other items listed above.

1div0:
I will take another look at the code this week.

ssh4net:
@lgritz, from what I see, there are only a couple of lines changed from my commit, and mainly about enabling/disabling hwy, so I don't see any issues there. And trust your next plan ideas.

lgritz:
Great. I am merging this as an initial experimental feature and a foundation on which to continue developing it into a major feature of OIIO (with subsequent improvements coming in smaller PRs that should be easy to review).

PR #4995, opened 2026-01-08
admin: Refine PR template to give more visual separation

lgritz:


lgritz:
Merging, this is not code related

PR #5004, opened 2026-01-12
fix: several bug fixes related to internal use of image_span

lgritz:
* ImageBuf internal buffer span lacked correct chansize. The internal `m_bufspan` is an `image_span<byte>`, and as such, it needs to remember the size of the original data type. Otherwise, there's a cascade of potential errors when it thinks that the individual values are byte sized.

* In both ImageInput and ImageOutput, several sanity checks of image_span size versus expectations were incorrect. They were only checking if the total byte sizes matched expectations, but they are allowed to disagree when you consider type conversions (in which case, it's the total number of values that need to match, not the total byte sizes.

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -150,10 +150,11 @@ public:
                      stride_t ystride = AutoStride,
                      stride_t zstride = AutoStride)
     {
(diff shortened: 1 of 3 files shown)


lgritz:
Any comments about these important bug fixes?

lgritz:
No objections -> merging.

PR #5007, opened 2026-01-14
test: Add new ref image for jpeg test

lgritz:
Needed for some systems after the changes of #4963.

Ever so slightly different LSB somewhere makes the hashes not match, but it's a correct visual result.


lgritz:
I need this right away, it's just an additional ref image that changes no code and doesn't change tests on platforms where they already passed, so I'm intending to merge it as soon as CI passes.

PR #5011, opened 2026-01-18
api(ImageBuf): IB::localpixels_as_[writable_]byte_image_span

lgritz:
Utilities to ask the IB for the local pixels as an untyped span of bytes. This is a "safe" alternative to `localpixels()`, which just returned a single pointer, instead returning an image_span that understands the sizes and strides of the buffer.

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -1377,7 +1377,7 @@ public:
     /// Return an `image_span<std::byte>` giving the extent and layout of
     /// "local" pixel memory, if they are fully in RAM and not backed by an
     /// ImageCache, and it is a writable IB, or an empty span otherwise.
-    image_span<std::byte> localpixels_as_writable_byte_image_span() const;
+    image_span<std::byte> localpixels_as_writable_byte_image_span();
 
     /// Pixel-to-pixel stride within the localpixels memory.
     stride_t pixel_stride() const;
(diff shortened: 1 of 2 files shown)


lgritz:
Note: the bleeding edge test failure is not from this PR, it's due to a new commit in main of pybind11 that is causing things to crash. I have reported it.

lgritz:
Any comments?

src/include/OpenImageIO/imagebuf.h
```diff
@@ -1369,6 +1369,16 @@ class OIIO_API ImageBuf {
     void* localpixels();
     const void* localpixels() const;
 
+    /// Return an `image_span<const std::byte>` giving the extent and layout
+    /// of "local" pixel memory, if they are fully in RAM and not backed by an
+    /// ImageCache, or an empty span otherwise.
```
EmilDohne, line 1374: [consider] It may be clearer to indicate that the return of this can be inferred from `ImageBuf::storage`. So for example:
```
Return an .. if ImageBuf::storage() is Storage::MEMORY`, otherwise ...
```

Apokogies if I botched that ayntax but hopefully the intention comes across

lgritz: I think I'm going to keep it as-is, because the storage test is actually a little more complicated than that. There are multiple storage() types that have addressable pixels, and multiple (at some point in the future, if not now) that do not. So I think keeping vague "if they are fully in RAM" is the distinction I'm trying to make.

PR #5012, opened 2026-01-19
build: Use libheif exported config if available

lgritz:
Needed to change the Libheif::Libheif target we set up ages ago to the "heif" name that the package's exported config uses.

Also, need to give the warning about static libraries occur any time we are using static libheif, even if we didn't set LIBSTATIC to try to force all static libraries.


lgritz:
I need this. Any objections?

lgritz:
Pre-emptively merging, I need this fix. It won't be in any releases until Feb 1, so if you spot something unwise here after the merge or encounter any resulting trouble yourself, let me know and we'll fix it.

PR #5014, opened 2026-01-26
api: IBA::make_texture now honors "maketx:threads" hint

### LINKED ISSUE #4254: [FEATURE REQUEST] OIIO::ImageBufAlgo::make_texture doesn't take an nthreads argument
https://openimageio.readthedocs.io/en/stable/imagebufalgo.html says:
`All ImageBufAlgo functions take an optional nthreads parameter that signifies the maximum number of threads to use to parallelize the operation. The default value for nthreads is 0, which signifies that the number of thread should be the OIIO global default set by OIIO::attribute() (see Section [Global Attributes](https://openimageio.readthedocs.io/en/stable/imageioapi.html#sec-globalattribs)), which itself defaults to be the detected level of hardware concurrency (number of cores available).`

But this is not true for make_texture. In the renderer my company uses, we do texture conversions automatically on scene load, concurrently with using multiple OpenImageIO clients via the shared TextureSystem. As such, we are beholden to whatever another client may have set nthreads to globally when calling make_texture.

How hard would it be to add an nthreads argument to make_texture, and why doesn't it already have one when the docs say it should do?

Thanks!

Texture conversion thread usage being separable from other client usage of OIIO.

An nthreads argument for make_texture.

Running maketx from the command line with an nthreads comment in a sys shell from C++ (urgh, no thanks).

**Additional context**
I could make a PR, but before I do that, I'm wondering if there was a good reason this wasn't the case already, to prevent me "finding out the hard way".

lgritz:
Nearly all IBA functions take an optional parameter controlling the threading. But make_texture() did not, so there was no way to control its thread usage (it would always use the default number of threads).

Without changing the call signature or ABI, this patch merely makes make_texture honor any "maketx:threads" hint passed in the config ImageSpec that it already takes to convey all sorts of controls over the texture creation process. Then this value is passed to any IBA functions, use of parallel_image, and anything else in the implementation of make_texture that would end up using the thread pool.

Fixes #4254

### DIFF
--- a/src/libOpenImageIO/maketexture.cpp
+++ b/src/libOpenImageIO/maketexture.cpp
@@ -929,7 +929,8 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
                         std::swap(small, sharp);
                     }
                     if (do_highlight_compensation) {
-                        ImageBufAlgo::rangeexpand(*small, *small);
+                        ImageBufAlgo::rangeexpand(*small, *small, false, ROI(),
+                                                  nthreads);
                         ImageBufAlgo::clamp(*small, *small, 0.0f,
                                             std::numeric_limits<float>::max(),
                                             true, ROI(), nthreads);
@@ -1452,7 +1453,7 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
         std::shared_ptr<ImageBuf> newsrc(new ImageBuf(src->spec()));
         ImageBufAlgo::channels(*newsrc, *src, src->nchannels() - 1,
                                cspan<int>(), cspan<float>(),
-                               cspan<std::string>(), true);
+                               cspan<std::string>(), true, nthreads);
         std::swap(src, newsrc);  // N.B. the old src will delete
     }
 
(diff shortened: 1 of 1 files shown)


src/libOpenImageIO/maketexture.cpp
```diff
@@ -926,13 +932,14 @@ write_mipmap(ImageBufAlgo::MakeTextureMode mode, std::shared_ptr<ImageBuf>& img,
                         ImageBufAlgo::rangeexpand(*small, *small);
```
jessey-git: Should `rangeexpand` also be modified?

src/libOpenImageIO/maketexture.cpp
```diff
@@ -1455,14 +1466,14 @@ make_texture_impl(ImageBufAlgo::MakeTextureMode mode, const ImageBuf* input,
...
-        && ImageBufAlgo::isMonochrome(*src)) {
+        && ImageBufAlgo::isMonochrome(*src, 0.0f, ROI(), nthreads)) {
         if (verbose)
             OIIO::print(
                 outstream,
                 "  Monochrome image detected. Converting to single channel texture.\n");
         std::shared_ptr<ImageBuf> newsrc(new ImageBuf(src->spec()));
         ImageBufAlgo::channels(*newsrc, *src, 1, cspan<int>(), cspan<float>(),
-                               cspan<std::string>(), true);
+                               cspan<std::string>(), true, nthreads);
```
jessey-git, line 1477: Should the `channels` call on line 1453 above this be modified too?

src/libOpenImageIO/maketexture.cpp
lgritz: yes.

hmmm, I really thought I was careful about search for every "ImageBufAlgo::" call. I will double check it all.

src/libOpenImageIO/maketexture.cpp
lgritz: Yes. Updated.

PR #5022, opened 2026-01-30
build(ocio): bump build ver to 2.5.1

zachlewis:
Makes OpenColorIO 2.5.1 the default version when building OCIO locally. This will cause the released python wheels to build against the currently-latest version of OCIO.

### DIFF
--- a/src/cmake/build_OpenColorIO.cmake
+++ b/src/cmake/build_OpenColorIO.cmake
@@ -6,7 +6,7 @@
 # OpenColorIO by hand!
 ######################################################################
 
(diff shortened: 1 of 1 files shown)


zachlewis:
I do not believe the CI failures are related to anything I've changed in this PR. 

I see there's a python-colorconfig failure, but it seems to be the same bindings-related failure that's affecting other tests. 
The actual output of test_colorconfig.py is correctly identified as consistent with out-ocio25.txt

zachlewis:
(that said, there's no rush to get this PR in before the next release, if it's causing problems.)

lgritz:
No, the CI failure is a known issue with top-of-tree pybind11.

PR #5024, opened 2026-01-31
ci: lock bleeding edge to pybind11 latest version

lgritz:
There's something in pybind11 master at the moment that is crashing in its destructors. It's been causing our "bleeding edge" test to fail for over a week now. I'm tired of our test failing, so I'm locking down to the last known working version. Will check back periodically and return to testing against pybind11 master after they have fixed it.


lgritz:
Merging this preemptively, since it demonstrably fixes broken CI and does not change code or builds for anybody else.

PR #5029, opened 2026-02-02
cleanup: remove left over tile emulation code for various formats

jessey-git:
### Description

Since the OpenImageIO 2.5 series, when calls to `check_open` were added, any format that did not declare support for "tiles" would immediately fail to open. But many of the formats which attempted to emulate tiles, by buffering the contents and writing it all as scanlines at the end, were not updated. All of the tile emulation code for these formats is effectively dead-code and untested.

Remove the tile emulation code from these formats.

An example of what the failure currently looks like:
```python
>>> out = oiio.ImageOutput.create("test.png")
>>> spec = oiio.ImageSpec(64, 64, 3, 'uint8')
>>> spec.tile_width = 64
>>> out.open("test.png", spec)
False

>>> out.geterror()
'png does not support tiled images'
```

### Tests

None were impacted.

### Checklist:
  behavior.
  testsuite**.
  on the GitHub runners.)

### DIFF
--- a/src/bmp.imageio/bmpoutput.cpp
+++ b/src/bmp.imageio/bmpoutput.cpp
@@ -27,9 +27,6 @@ public:
     bool close(void) override;
     bool write_scanline(int y, int z, TypeDesc format, const void* data,
                         stride_t xstride) override;
-    bool write_tile(int x, int y, int z, TypeDesc format, const void* data,
-                    stride_t xstride, stride_t ystride,
-                    stride_t zstride) override;
 
 private:
     int64_t m_padded_scanline_size;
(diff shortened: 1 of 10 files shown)


lgritz (approved):
LGTM

lgritz:
This is predicated on the following observations:
1. Some format writers emulated tile output (by buffering) even if they were inherently scanline on disk, but neglected all along to advertise this feature through the API by returning true from `supports("tiles")`.
2. A change made a couple years ago to OIIO 2.5 inadvertently made it so that requests to write tiles to those formats that didn't do `supports("tiles")` would fail to `open()` at all.
3. Nobody ever noticed this or complained.

So we are inferring that nobody has ever depended on this "scanline file that emulates the tile interface" feature, or they would have noticed that it stopped working.

Therefore, rather than repairing this, Jesse is just ripping out that emulation code entirely.
Removing code that nobody uses is better than maintaining that code indefinitely!

PR #5030, opened 2026-02-05
fix(bmp): detect corrupt files where palette doesn't match bpp

lgritz:
Extra protections for corrupted BMP files that claim to be palette images, but have a BPP that doesn't support palette images. Also an extra guard around accessing the palette array if it is empty.

Add an extra test case for this kind of corruption.


jessey-git (approved):
Confirmed that this does fix the issue here.

PR #5031, opened 2026-02-06
testing: Add testsuite/heif ref output for libheif 1.21 + avif support

lgritz:


lgritz:
Assuming this passes CI, I'm planning to merge it immediately because I need it right away. It can't break anything, it's just adding a new reference output for a test.

PR #5033, opened 2026-02-07
fix: address fmath.h warning with ispow2

lgritz:
I was seeing warnings with instantiation of the ispow2 function template for unsigned type, where the `x >= 0` clause is meaningless. Use a constexpr if to eliminate that pointless test for unsigned types.


lgritz:
Been op for a couple weeks. Last chance. Seem ok?

PR #5034, opened 2026-02-07
build: self-builder logic fixes for deep vs shallow clones

lgritz:
* cmake utility build_dependency_with_cmake was unconditionally doing a shallow clone and using `clone -b`, but that only works if it's got a branch or tag name, not if it has a commit hash. So change the logic so it does a shallow clone only if GIT_TAG is specified but GIT_COMMIT is not.
* pybind11 self-builder is modified to allow a git commit override.


lgritz:
2 weeks with no comments. I will merge this at the end of the day if nobody objects.

PR #5035, opened 2026-02-09
fix(tiff): Fix TIFF output crash for multi-count Exif metadata

### LINKED ISSUE #5023: bug: OpenImageIO crashes when generating TIFF with extra attributes.
**Describe the bug**
I found a bug where OpenImageIO crashes when trying to generate TIFF files from input with extra attributes. Using the .exr file in [image_extra_attributes.zip](https://github.com/user-attachments/files/24974270/image_extra_attributes.zip) should yield a silent crash in oiiotool:
```
 oiiotool image_extra_attributes.exr -o out.tif 
```

From what I've gathered from my own debugging, this looks like an issue with how extra attributes (e.g. EXIF data) are handled and might be a memory allocation issue.

```
OIIO 3.1.9.0 | unknown arch?
    Build compiler: MSVS 1938 | C++17/199711

Dependencies: DCMTK NONE, expat 2.6.3, FFmpeg NONE, fmt 10.2.1, Freetype
    2.13.2, GIF 5.2.1, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.23, Libheif   
    NONE, libjpeg-turbo 3.1.2, LibRaw NONE, libuhdr 1.4.0, minizip-ng 4.0.10,      
    OpenColorIO 2.4.2, OpenCV NONE, OpenEXR 3.3.5, OpenGL, OpenJPEG 2.5.4,
    openjph NONE, PNG 1.6.50, Ptex NONE, Ptex NONE, pybind11 3.0.1, pystring       
    NONE, Python3 3.11.9, Qt5 NONE, Qt6 NONE, Robinmap 1.4.0, TBB NONE, TIFF       
    4.7.1, WebP 1.6.0, yaml-cpp 0.8.0, ZLIB 1.3.1, ZLIB 1.3.1
```

**To Reproduce**
See above description.

lgritz:
Fixes #5023

This was crashing when writing TIFF information that was supposed to be arrays of more than one rational, but in fact was provided as a single value, it was reading past the end of a memory array.

I noticed that this whole region needs a cleanup, this is not the only problem. But a full overhaul seems too risky to backport, so my strategy is as follows:

* THIS fix first, which I will backport right away to 3.0 and 3.1.

* I will then submit a separate PR (already implemented and tested) that is a much more complete fix and overhaul of this portion of the code (and other places). That will get merged into main when approved.

* After the second PR is merged, I'll hold it in main for a while to test its safety, and then decide if it seems ok to backport to 3.1 (but definitely not 3.0).

### DIFF
--- a/src/tiff.imageio/tiffoutput.cpp
+++ b/src/tiff.imageio/tiffoutput.cpp
@@ -1141,16 +1141,16 @@ TIFFOutput::write_exif_data()
             if (tifftype == TIFF_ASCII) {
                 ok = TIFFSetField(m_tif, tag, *(char**)p.data());
             } else if ((tifftype == TIFF_SHORT || tifftype == TIFF_LONG)
-                       && p.type() == TypeDesc::SHORT) {
+                       && p.type() == TypeDesc::SHORT && count == 1) {
                 ok = TIFFSetField(m_tif, tag, (int)*(short*)p.data());
             } else if ((tifftype == TIFF_SHORT || tifftype == TIFF_LONG)
-                       && p.type() == TypeDesc::INT) {
+                       && p.type() == TypeDesc::INT && count == 1) {
                 ok = TIFFSetField(m_tif, tag, *(int*)p.data());
             } else if ((tifftype == TIFF_RATIONAL || tifftype == TIFF_SRATIONAL)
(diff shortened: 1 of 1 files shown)


lgritz:
The attribute is completely wrong in the exr file to begin with, so there's not much we can do.

PR #5036, opened 2026-02-09
fix(tiff): Improve TIFF robustness for non-matching tag/metadata types

lgritz:
This is a more comprehensive fix for issues discovered in PR #5035.

The original problem reported in Issue #5023 was a crash when writing TIFF information that was supposed to be arrays of more than one rational, it was reading past the end of a memory array.

#5035 is a minimal, immediate fix to address the crashes. But in the process, I saw a number of ways in which we were dropping metadata on the floor when the types didn't exactly match, but that we *could* handle with automatic conversion.

The new cases that we handle with this PR are:

* Exif RESOLUTIONUNIT tag is a short, but by convention we store it by the name as a string in OIIO metadata, so we need to convert back to a code (we did so for the main TIFF metadata, but not for Exif in TIFF).
* Handle Exif "version" and "flashpixversion" metadata which have unusual encoding in TIFF files (they are 4-character strings, but must be stored in a TIFF tag of type BYTES, not as the usual type ASCII that most strings use.
* Handle things that TIFF insists are ASCII but that come to us as metadata that's strings. Easy -- our `ParamValue.get_string()` automatically converts ther things like ints or floats into string representation.
* Much more flexibility in automatically converting among the signed and unsigned, 16 and 32 bit, integer types when the metadata in our ImageSpec is integer but not the specific type of integer that TIFF/Exif thinks it should be.

This doesn't appear to change the results of anything in our testsuite, but it's possible that some non-TIFF-to-TIFF image conversions that contain Exif data may now do certain type conversions properly instead of just silently dropping the metadata that had non-matching (but reasonably valid) types.

Additionally, to do this nicely, I ended up adding a new TypeURational alias in typedesc.h (similar to TypeRational, but the case where both numerator and denominator are unsigned ints).

And also fixed a random comment typo I noticed in tiffinput.cpp.

### DIFF
--- a/src/tiff.imageio/tiffoutput.cpp
+++ b/src/tiff.imageio/tiffoutput.cpp
@@ -1150,15 +1150,21 @@ TIFFOutput::write_exif_data()
                     ok = TIFFSetField(m_tif, TIFFTAG_RESOLUTIONUNIT, r);
                 }
                 handled = true;
-            } else if ((tag == EXIF_EXIFVERSION || tag == EXIF_FLASHPIXVERSION)
-                       && p.type() == TypeString) {
-                // These tags are a 4-byte array of chars, but we
-                // allow users to set it as a string. Convert it if needed.
-                std::string version = p.get_string();
-                if (version.size() >= 4) {
-                    ok = TIFFSetField(m_tif, tag, version.c_str());
+            } else if (tag == EXIF_EXIFVERSION || tag == EXIF_FLASHPIXVERSION) {
+                if (p.type() == TypeString) {
+                    // These tags are a 4-byte array of chars, but we
+                    // allow users to set it as a string. Convert it if needed.
+                    std::string version = p.get_string();
+                    if (version.size() >= 4) {
+                        ok = TIFFSetField(m_tif, tag, version.c_str());
+                    }
+                    handled = true;
+                } else if (p.type() == TypeInt) {
+                    std::string s = Strutil::fmt::format("{:04}", p.get_int());
+                    if (s.size() == 4)
+                        ok = TIFFSetField(m_tif, tag, s.c_str());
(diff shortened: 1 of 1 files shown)


jessey-git:
The original bug report now produces the following log output. The first line is expected but I'm unsure why Exif:ExifVersion is being reported?
```
Unhandled EXIF Exif:LensSpecification (float) / tag 42034 tifftype 5 count 4
Unhandled EXIF exif:ExifVersion (int) / tag 36864 tifftype 7 count 1
```

The original attribute looks like: `exif:ExifVersion: 221`

lgritz:
DEBUG builds (but not Release) print that message when a known Exif attribute is encountered but it can't figure out what to do. This is supposed to help us catch the set of unhandled cases. Currently, nothing in the testsuite triggers this.

The file in the bug report is deeply flawed. The Exif "LensSpecification" is supposed to be composed of 4 rational (or float, on our side) numbers. But the exr file in question just gives a single float.

But I do think we can probably handle the ExifVersion, let me fix that.

jessey-git:
Code looks good now and `exiftool` shows the (corrected) version on the converted .tif:
`Exif Version                    : 0221`

But somehow `oiiotool --info -v` does not list it. What might be causing that?

lgritz:
> Code looks good now and `exiftool` shows the (corrected) version on the converted .tif: `Exif Version : 0221`
> 
> But somehow `oiiotool --info -v` does not list it. What might be causing that?

I'm investigating. I think there are some deficiencies on the TIFF reading side as well. Since that's a separate code path and set of issues than the present PR that addresses the writing side of things, I'd like to merge this PR if the write looks correct, and then I will make a separate PR for fixes to come on the reading side. Sound ok?

PR #5039, opened 2026-02-10
build: Raise fmt auto-build version to 12.1, handle Windows flags

lgritz:
Bump the version of 'fmt' library that we download and build (if not found) from 10.2 to 12.1.

Some other touch-ups in build_fmt.cmake.

Also, we have seen that recent fmt versions will fail to compile on MSVC unless using the `/utf-8` compiler flag, so ensure that is used and also passed on to other clients of libOpenImageIO_Util (which expose templates using those headers).

### DIFF
--- a/src/cmake/build_fmt.cmake
+++ b/src/cmake/build_fmt.cmake
@@ -6,7 +6,7 @@
 # fmt by hand!
 ######################################################################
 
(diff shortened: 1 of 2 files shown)


lgritz:
My intent is to backport this to 3.1 for sure, and maybe to 3.0 as well.

PR #5040, opened 2026-02-11
build: Remove support for deprecated Intel icc compiler

lgritz:
Intel icc is deprecated and hasn't had a release for a few years. It's holding us back, both by making us work around an ever growing number of icc bugs and limitation that will never be fixed, as well as not allowing us to upgrade minimum versions of certain dependencies, because icc can't correctly compile newer versions (as an example, it cannot use a 'fmt' library newer than the oldest we support, 7.0).

So it's time to thank icc for its service and put it on the ice floe for the polar bears to eat. This is of course in main (future 3.2), and will not be backported to release branches, since we never stop support of a dependency or toolchain of existing releases. People requiring icc for whatever reason may keep using OIIO 3.1 or older.

We will continue to support and test icx, the fully supported Intel LLVM-based compiler.

### DIFF
--- a/src/include/OpenImageIO/benchmark.h
+++ b/src/include/OpenImageIO/benchmark.h
@@ -15,7 +15,7 @@
 #include <OpenImageIO/timer.h>
 
 
-#if (((OIIO_GNUC_VERSION && NDEBUG) || OIIO_CLANG_VERSION >= 30500 || OIIO_APPLE_CLANG_VERSION >= 70000 || defined(__INTEL_COMPILER)  || defined(__INTEL_LLVM_COMPILER)) \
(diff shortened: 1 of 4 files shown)


lgritz:
Any objections to dropping icc support, or does anybody see anything unwise about the code I'm removing in this PR?

zachlewis:
No objections from me. If icc is still seeing regular industry use, I imagine we'll hear about it in the issues / PRs.

lgritz:
icc is considered deprecated by its maker, Intel. I really hope nobody is still depending on it. But if they are, they are by definition stuck on a several-years-old toolchain, so I figure they ought to be able to make do with OIIO 3.1 or earlier as well.

zachlewis:
Hey, OIIO 3.1 isn't too shabby. I say, let 'em suffer.

lgritz:
Last chance for comments. I will merge this at the end of the week if nobody objects.

PR #5041, opened 2026-02-11
build(deps): Raise minimum fmt library version to 9.0

lgritz:
The previous minimum, 7.0, dated from mid-2020.

We are raising now (in main / future 3.2 only) to 9.0, which dates from mid-2022, so we're still supporting several versions and/or years back.

Because this changes minimum dependency versions, it will NOT be backported to release branches (3.1 or earlier).

I had to remove the CI test variant for icc, because ancient icc can't correctly build newer versions of fmt, it seems. There is a separate PR to simply drop icc from our list of supported compilers.

If anybody wants to argue for pulling the minimum up even farther (say, to fmt 10.0, released in 2023, so still supporting 3 years back), which would simplify even more places, I would consider it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
             cxx_std: 17
             python_ver: 3.9
             simd: "avx2,f16c"
-            fmt_ver: 8.1.1
+            fmt_ver: 9.0.0
             opencolorio_ver: v2.3.0
             pybind11_ver: v2.9.0
             setenvs: export FREETYPE_VERSION=VER-2-12-0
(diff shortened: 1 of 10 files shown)


lgritz:
I'm not going to merge this until after #5040 is approved and merged, because raising the minimum fmt to 9.0 will, I believe, break under icc. (But not a worry once we officially drop icc support.)

But in the mean time, does anybody object for any other reason to raising the fmt minimum to 9.0 for future (3.2 and beyond) releases of OIIO?

lgritz:
I will merge this shortly after #5040 is merged if there are no specific objections to raising our fmtlib requirement.

PR #5042, opened 2026-02-12
ci: Turn off nightly analysis and wheel workflows for user forks

lgritz:


brechtvl:
Thanks!

brechtvl:
I don't think this actually works, `if` seems not supported there but silently ignored?

Maybe this approach is better
https://hugovk.dev/blog/2023/til-how-to-disable-cron-for-github-forks/

```
 jobs:
   test:
+    if: ${{ github.event.repository.fork == false || github.event_name != 'schedule' }}
     runs-on: ubuntu-latest
```

lgritz:
> I don't think this actually works, `if` seems not supported there but silently ignored?
> 
> Maybe this approach is better https://hugovk.dev/blog/2023/til-how-to-disable-cron-for-github-forks/
> 
> ```
>  jobs:
>    test:
> +    if: ${{ github.event.repository.fork == false || github.event_name != 'schedule' }}
>      runs-on: ubuntu-latest
> ```

Oh, that's MUCH better!

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ on:
   schedule:
     # Full nightly build
     - cron: "0 8 * * *"
-      if: github.repository == 'AcademySoftwareFoundation/OpenImageIO'
+      if: github.event.repository.fork == false
```
brechtvl: To be clear, I don't think `if` is supported here.

If I search for ".if" on this page, it's only for jobs and job steps.
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax

lgritz: yeah, Fixing that too and just posted another update

.github/workflows/ci.yml
```diff
@@ -42,7 +41,7 @@ concurrency:
 jobs:
 
   aswf-old:
-    if: ${{ ! contains(github.ref, 'windows-only') && ! contains(github.ref, 'macos-only') }}
+    if: ${{ (github.event.repository.fork == false || github.event_name != 'schedule') && ! contains(github.ref, 'windows-only') && ! contains(github.ref, 'macos-only')  && (github.event.repository.fork == false || github.event_name != 'schedule') }}
```
brechtvl: This is checking the same thing twice.

lgritz:
I'll merge after CI finishes (which of course, does not test whether I've done it right, only whether I've completely broken the workflow scripts).

I guess we'll find out by tomorrow morning if it really worked -- by seeing if our forks did any real work overnight.

lgritz:
Looking at jobs for my fork, it does look like it skipped (well, started but immediately terminated) all of the nightly cron jobs on my fork. So I think this worked!

PR #5050, opened 2026-02-18
feat(tiff): Support GPS fields, and other metadata enhancements

### LINKED ISSUE #5049: bug: OpenImageIO is not able to write GPS tags to tiff files
**Describe the bug**
When writing an image file with ImageOutput, if the extra_attribs contains any GPS exif data (for example, `GPS:Latitude`) they will not be written to the file.

2.4.14 but the problem appears to exist in the latest version as well by code inspection

**To Reproduce**
easiest is to take a jpg with GPS tags and run oiiotool in.jpg -o out.tif

**Evidence**
example jpg file with GPS tags, example oiiotool output tif without GPS tags, example tif with GPS tags written by other means:
Image
[image_2026-02-16-140503.tif](https://github.com/user-attachments/files/25349729/image_2026-02-16-140503.tif)
[out.tif](https://github.com/user-attachments/files/25349728/out.tif)

- Example input: If the problem only happens with certain image files, please
  attach the smallest image you can make that reproduces the problem.

please just go straight to making a [pull request](https://github.com/AcademySoftwareFoundation/OpenImageIO/pulls).

lgritz:
Most visible change: The TIFF reader and writer now correctly handle GPS tags.

Fixes #5049

But along the way, and to get it completely right, a fairly extensive refactoring of our TIFF tag handling was needed. The libtiff behavior around different tags is extremely convoluted, and there were a bunch of cases we either didn't handle correctly, or were dropping on the floor. So there are also some other non-GPS tags that we missed all along but now read properly.

Note that the support for separate IFDs for GPS tags is only supported via the API in libtiff 4.2 and newer, so this is disabled when bulding against older libtiff versions. (Our support of old libtiff versions extends extraordinarily far back in time.)

### DIFF
--- a/src/tiff.imageio/tiffoutput.cpp
+++ b/src/tiff.imageio/tiffoutput.cpp
@@ -1257,13 +1257,13 @@ TIFFOutput::write_extra_tag_directory(string_view tag_set_name)
                 // floats. But actually, libtiff uses C conventions, so we actually
                 // pass doubles.
                 if (count == 1) {
-                    ok = TIFFSetField(m_tif, tag, p.get_float());
+                    ok      = TIFFSetField(m_tif, tag, p.get_float());
                     handled = true;
                 } else if (count > 1) {
                     auto vals = OIIO_ALLOCA_SPAN(double, count);
                     for (int i = 0; i < count; ++i)
                         vals[i] = p.get_float_indexed(i);
-                    ok = TIFFSetField(m_tif, tag, vals.data());
+                    ok      = TIFFSetField(m_tif, tag, vals.data());
                     handled = true;
                 }
             }


lgritz:
The failures on the wheel builds are unrelated to this PR and are happening in main. We are investigating that separately, and it should not inhibit review and acceptance of this PR.

lgritz:
@kjellpeterson-shotover Are you able to tell if this patch satisfies your requirements as described in #5049 ?

kjellpeterson-shotover:
Hi, sorry for the wait on that. I couldn't manage to include the PR in our project and test that way, but based on oiiotool built from the PR it seems like it should do what I was asking for. Thanks!

lgritz:
Thanks for reporting back!
Will merge.

PR #5051, opened 2026-02-18
Bugfix for libheif detection

irieger:
### Description

When the LIBHEIF_LIBRARIES refers to targets (like happening when conan packaging provides the targets), an empty string from get_filename_component will fail as it is not counted as an argument if used without quotes.

Adding the quote lets the code pass without an error. It will not display the warning, but it provides at least a fix for letting it build. (In the conan package we patch that check out at the moment to get it to build: https://github.com/conan-io/conan-center-index/blob/master/recipes/openimageio/all/patches/3.1.10.0-conan-fixes.patch)

I tried to extract the underlying information but haven't come that far. For example some debugging code I tried:
```
        if (TARGET ${_libeheif_library})
            get_property(libeheif_type TARGET _libeheif_library PROPERTY TYPE)
            get_target_property(interface_library "${_libeheif_library}" INTERFACE_LINK_LIBRARIES)
            message("Get the actual libraries of the target ${_libeheif_library}: interface_library")
            message("  Type of library: ${libeheif_type}")
        endif()
```

Such commands extract a bit more info, as the library type is only INTERFACE_LIBRARY and not shared or static, that didn't help much. Also at least for usages like conan, where there are prepared helper tools to get all license information etc. So I'm not too deep into CMake for that specific stuff, so just at least suggesting to use the quotes to not break it if CMake can't extract the library file ending. Could also switch to display another warning saying it MIGHT link to libheif with its broken license.

### Tests

The target to have it build with conan is achieved. Tried different experiments, including the current one locally via changing the patch accordingly.

### Checklist:

- [ ] **I have read the guidelines** on [contributions](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md) and [code review procedures](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/dev/CodeReview.md).
  behavior.
  testsuite**.
  on the GitHub runners.)
- [ ] **My code follows the prevailing code style of this project** and I


lgritz:
Superseded by #5053

PR #5054, opened 2026-02-20
jxl: Extending JXL CICP support to include P3 / color primaries 12

shanesmith-dwa:
### Description

I tested out the JPEG XL CICP support and noticed that color primaries 12 was not supported. This pull request is looking to extend P3 support for color primaries 12.
Note: color primaries 11 uses the DCI white point and color primaries 12 uses the D65 white point.

The JxlPrimaries enum only covers P3 primaries as value 11 and not 12. See, 
https://github.com/libjxl/libjxl/blob/main/lib/include/jxl/color_encoding.h#L55-L75
Further code is therefore required to account for this on read and write.

### Tests

Tests for read and write of color primaries 11 and 12 were added.

### Checklist:

  behavior.
  testsuite**.
  on the GitHub runners.)

### DIFF
--- a/src/jpegxl.imageio/jxlinput.cpp
+++ b/src/jpegxl.imageio/jxlinput.cpp
@@ -383,9 +383,15 @@ JxlInput::open(const std::string& name, ImageSpec& newspec)
     if (have_color_encoding && color_encoding.primaries != JXL_PRIMARIES_CUSTOM
         && color_encoding.white_point != JXL_WHITE_POINT_CUSTOM
         && color_encoding.transfer_function != JXL_TRANSFER_FUNCTION_GAMMA) {
-        const int cicp[4] = { color_encoding.primaries,
-                              color_encoding.transfer_function, 0 /* RGB */,
-                              1 /* Full range */ };
+        int color_primaries = color_encoding.primaries;
+        // JxlPrimaries enum only covers P3 primaries as value 11 and not 12
+        // but CICP has separate code values based on white point.
+        if (color_primaries == JXL_PRIMARIES_P3
+            && color_encoding.white_point == JXL_WHITE_POINT_D65) {
+            color_primaries = 12;
+        }
+        const int cicp[4] = { color_primaries, color_encoding.transfer_function,
+                              0 /* RGB */, 1 /* Full range */ };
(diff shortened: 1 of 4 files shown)


zachlewis:
@brechtvl , any thoughts?

I don't really know enough about the JXL API to know if this is sufficient; and I'm not sure if the fact (I think?) that the API only enumerates D65-based color space encodings is an oversight -- I think it may suggest something about the limits of the intended use of CICP codes. JXL may expect folks to either convert to or chromatically adapt to P3D65, in order to use CICP codes, and perhaps use JXL's custom chromaticities API elsewhere?

Maybe this is something that is better implemented upstream first? 

Although I don't mind, in theory, supporting non-d65 display color spaces in OIIO, it does require that we make certain assumptions of the mastering display creative whitepoint used for data tagged CICP.Primaries == 11, and/or provide a means for users to override the default behavior. In particular, we'd tend to assume a D65 creative whitepoint, and interpret with a color space that represents a "D65-as-DCI" encoding, meaning, a D65-based encoding that's been adjusted to maintain intended appearance when viewed though a DCI-P3-calibrated projector. It's also possible that the creative whitepoint is something other than D65 -- the D60 ACES whitepoint is a far more likely candidate than the DCI-P3 whitepoint itself (which is greenish and ugly).

May I ask what your use case is, @shanesmith-dwa?

shanesmith-dwa:
The JXL API detaches the color primaries from the white point which is why reconciling to CICP is tricky. 

The libjxl white point enum is a little higher up in the same file:
https://github.com/libjxl/libjxl/blob/main/lib/include/jxl/color_encoding.h#L35-L53
This does cover D65, custom, Illuminant E and DCI. 

Here are some other references to libjxl's handling of the reconciliation:
https://github.com/libjxl/libjxl/blob/main/tools/wasm_demo/jxl_decompressor.cc#L44-L50
https://github.com/libjxl/libjxl/blob/main/lib/extras/enc/apng.cc#L211-L218

I am looking for JXL CICP 12,13,0,1 support.

zachlewis:
Alright, fair enough. We'll address mapping CICP 12,13,0,1 to an internal color space in a separate commit.

PR #5055, opened 2026-02-20
fix(oiiotool): make sure `oiiotool --compression` does expression substitution

lgritz:
`oiiotool --compression XXX` just used the argument as a literal, and wasn't able to use expression substitution for the value. That made it behave differently than the supposedly synonymous `--attrib compression XXX` in the case that the argument was an expression rather than a literal value.


lgritz:
Any comments?

PR #5056, opened 2026-02-21
fix(oiiotool): better type understanding with -i:ch= and other cleanup

lgritz:
Background: oiiotool is pretty good at remembering the data type of the first file it reads in a command, and ensuring that the output preserves the same types, despite the fact that internally, it tends to keep everything in float. (As a further reminder: -d can override the heuristics that make this work.)

But there was an exception: when `-i` was used with the optinal modifier `-i:ch=...` to select a subset of channels or reorder channels upon read, the underlying call to IBA::channels would cause it to not have that information in the "nativespec" at the right time that it was collected.

This PR primarily ensures that the `-i:ch=...` preserves the information about which channels had which data types, despite the channel shuffle.

Some other bits and pieces along the way:

* A new `TypeDesc::basetype_merge()` variety that take a `cspan<TypeDesc>`.
* A new utility `TypeDesc::all_types_equal(cspan<TypeDesc>)`
* Cleanup in IBA::channels() to more correctly compute the common type of the channel subset that you actually specify.


lgritz:
2 weeks, no objections. Will merge by the end of the week.

PR #5058, opened 2026-02-22
fix(python): Address new pybind11 float/int auto-conversion behavior

lgritz:
It's not in any pybind11 tagged release yet, but in its master, a change has been introduced that causes it to allow functions with float paramters to automatically match int arguments. This is fatal for us, as we have set up several cases of methods that overload on int vs float, and with this change, the float one gets called when an int is passed. Pybind11 docs say that the `.noconvert()` modifier on the argument will do the trick, but I'm finding that it doesn't help for overloaded function names.

The solution is to find these cases, change our declaration to a generic py::object, and then sort out what type was passed with py::isinstance. It seems to mainly affect `attribute()` style calls.

In the process, minor changes to the pybind11 auto-builder allow it to take a git commit hash directly. Also, I took the opportunity to remove a few reference outputs that are no longer needed because they were python2-specific.

Meanwhile, just in case anybody had been paying attention, the problem in pybind11 master that had caused us to temporarily stop CI testing against master seems to have been resolved. So re-enable the bleeding edge test to use pybind11 master (which will fail if not for this PR here), and bump the "latest version" tests to the latest pybind11 3.0.2.

I diagnosed the problem myself, wrote the solution for one class and verified that it solved the issues with that class. Then I used Claude Code to make the analogous set of changes to the other classes that needed them and I reviewed that it was correct.

### DIFF
--- a/src/python/py_oiio.h
+++ b/src/python/py_oiio.h
@@ -521,6 +521,8 @@ attribute_onearg(T& myobj, string_view name, const py::object& obj)
         myobj.attribute(name, int(obj.template cast<py::int_>()));
     else if (py::isinstance<py::str>(obj))
         myobj.attribute(name, std::string(obj.template cast<py::str>()));
+    else if (py::isinstance<py::bytes>(obj))
+        myobj.attribute(name, std::string(obj.template cast<py::bytes>()));
     else
         throw py::type_error("attribute() value must be int, float, or str");
 }


lgritz:
Any comments on this? It's been up for a week.

lgritz:
No objections -> merging

PR #5059, opened 2026-02-23
ci: add MacOS 26 (ARM) to the CI lineup

lgritz:


lgritz:
Merging because nobody objected and all this does is add a new CI test case (no code changes).

PR #5060, opened 2026-02-23
ci: Add CI test for MSVS 2026

lgritz:
And some minor fixes discovered along the way.


lgritz:
The wheel test failures are entirely unrelated

src/build-scripts/build_opencolorio.bash
```diff
@@ -20,7 +20,7 @@ OPENCOLORIO_BUILD_DIR=${OPENCOLORIO_BUILD_DIR:=${LOCAL_DEPS_DIR}/OpenColorIO-bui
 OPENCOLORIO_INSTALL_DIR=${OPENCOLORIO_INSTALL_DIR:=${LOCAL_DEPS_DIR}/dist}
 OPENCOLORIO_BUILD_TYPE=${OPENCOLORIO_BUILD_TYPE:=Release}
 OPENCOLORIO_CXX=${OPENCOLORIO_CXX:=g++}
-OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:="-Wno-unused-function -Wno-deprecated-declarations -Wno-cast-qual -Wno-write-strings"}
+# OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:="-Wno-unused-function -Wno-deprecated-declarations -Wno-cast-qual -Wno-write-strings"}
```
jessey-git: Was this change intentional?

lgritz: It was intentinal, though it turned out to be tangential to the rest of the PR.

At one stage of developing this PR, the OCIO self-building was not working on the new test variant for MSVS 2026 (that turned out to be the misspelling of "expat" as "EXPAT" in build_expat.cmake -- and do not ask me "how did that ever work" because I have no idea).

So while the self-building wasn't working, I tried to have Windows use the build_opencolorio.bash script to build OCIO before our build, but I found that those `-W` flags were not accepted by MSVS (in retrospect -- DUH), so I tried commenting that line out so that it would work for Windows, and to see if it still worked for Linux/Mac, on the theory that maybe those warning suppressions were added for a long-ago OCIO version and were no longer needed. And that turned out to be the case! No CI variant failed without that line, so I left it commented out.

Then I returned to figuring out why the self-builder wouldn't work, and found the EXPAT spelling problem, so in the end, the Windows CI didn't actually need to run the build_opencolorio.bash script. But it's fixed on Windows now anyway.

I guess I left the line commented out, but not removed, in case we shortly discovered a case where it was needed, as a reminder that we can customize it at that spot. But if you think it's better to remove the line, I can do that.

jessey-git: Maybe add a comment right above it saying that it can be used as a customization spot if absolutely necessary? I never quite know how to reason about perfectly working code that's been commented out so having the comment would help I think.

lgritz: I used a slightly different construct that leaves it in (so it's self-documenting), but doesn't assign it anything.

```
OPENCOLORIO_CXX_FLAGS=${OPENCOLORIO_CXX_FLAGS:=}
```

PR #5081, opened 2026-03-12
fix(jpeg): Improved safety and error reporting for jpeg and iptc

lgritz:
* `decide_iptc_iim()` (semi-public) : change from taking a pointer & length to a string_view, and internal safety checks.

* `jpeg_decode_iptc()` (non-public): change from taking an unlengthed pointer (!) to a string_view, and internal safety checks.

* JpgInput::open(): a number of extra checks for memory bounds validity.


jessey-git (changes requested):
Current changes look ok again, but there's a few obvious interger overflows left when dealing with larger cymk images. The calculations on lines  658, 661, and 691 can overflow given the (1 << 16) max width and height of incoming images.

lgritz:
Updated with several more fixes including your suggestions and some of those by Copilot.

jessey-git (approved):
Changes seem alright to me.

greg-churchillnavigation:
Is there a way to ignore IPTC blocks deemed corrupt so we can load an image regardless?
I'm seeing some old .JPG files exported by Affinity Photo no longer load. They return error `"Corrupted IPTC data"` from `JpgInput::jpeg_decode_iptc`.

lgritz:
Can you send me one?

lgritz:
@greg-churchillnavigation  Which version of OIIO are you using? I think maybe I already fixed this.

greg-churchillnavigation:
I believe I'm using v3.1.12.0-f108fca93f my jpeginput.cpp matches this PR.
I now see main has changed.
Here's an image that fails (attempting attachment 'test_bad'.jpg)
test_bad

lgritz:
It's broken as you describe in 3.1.12.0, but it's been fixed by 3.1.13.1.

lgritz:
It was fixed by https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5140

greg-churchillnavigation:
Thank you!

PR #5084, opened 2026-03-12
feat: Nanobind for python bindings (first steps -- pybind11 still working)

soswow:
First PR on the way to a migration from pybind11 to nanobind.

New env flag introduced `OIIO_PYTHON_BINDINGS_BACKEND` = `pybind11` or `nanobind` or `both`

> Select which Python binding backend(s) to configure. `both` keeps the existing pybind11 module and also builds the experimental nanobind module.

When it is nanobine one is build it is in `PyOpenImageIONanobindExperimental` target.

```
cmake --fresh -S . -B build \
  -DUSE_CCACHE=OFF \
  -Dfmt_DIR=/opt/homebrew/lib/cmake/fmt \
  -DOpenColorIO_DIR=/opt/homebrew/lib/cmake/OpenColorIO \
  -DOIIO_INTERNALIZE_FMT=ON \
  -DOIIO_PYTHON_BINDINGS_BACKEND=both
```

### Checklist:

  behavior.
  testsuite**.
  on the GitHub runners.)

This code contribution was assisted by Cursor/Composer-2


lgritz:
@soswow This is looking good! 

The only real concern I have at this stage is that rather than the logic for running the nanobind tests living in runtest.py, I would much rather that it all be in testing.cmake.

You can see (circa testing.cmake:197) how for each of the texture tests, it actually adds them to the list a second time with different arguments to make the same test run again with in batch texturing mode (and modifying the name to append ".batch" to make it look like a separate test). I think this is how it should work for pybind11 vs nanobind as well. There should be minimal logic changes necessary in runtest.py.

lgritz:
@soswow Ping. I proposed some changes to the testing strategy, but otherwise would like to merge what you have as a great first step to the nanobind transition.

src/python-nanobind/py_imagespec.cpp
```diff
@@ -0,0 +1,301 @@
+// Copyright Contributors to the OpenImageIO project.
+// SPDX-License-Identifier: Apache-2.0
+// https://github.com/AcademySoftwareFoundation/OpenImageIO
+
+#include "py_oiio.h"
+
+namespace {
+
+OIIO_NAMESPACE_USING
```
jessey-git, line 9: We try not to use this macro any longer. Just do a `using namespace OIIO;` here and in the other files.

src/python-nanobind/py_paramvalue.cpp
```diff
@@ -0,0 +1,229 @@
...
+        .def("__init__", [](ParamValue* self, const std::string& name,
+                            TypeDesc type, int nvalues,
+                            ParamValue::Interp interp, nb::handle obj) {
+            new (self) ParamValue(
+                paramvalue_from_pyobject(name, type, nvalues, interp, obj));
+        });
+
+    nb::class_<ParamValueList>(m, "ParamValueList")
+        .def(nb::init<>())
+        .def("__getitem__",
```
jessey-git, line 157: The `__getitem__` calls in the old pybind11 use `return_value_policy::reference_internal`. Do we have to do something equivalent here in some form?

src/python-nanobind/py_imagespec.cpp
```diff
@@ -0,0 +1,301 @@
...
+                if (Strutil::iequals(verbose, "brief"))
+                    verb = ImageSpec::SerialBrief;
+                else if (Strutil::iequals(verbose, "detailed"))
+                    verb = ImageSpec::SerialDetailed;
+                else if (Strutil::iequals(verbose, "detailedhuman"))
+                    verb = ImageSpec::SerialDetailedHuman;
+                return std::string(spec.serialize(fmt, verb));
+            },
+            "format"_a = "text", "verbose"_a = "detailed")
+        .def(
```
jessey-git, line 267: The old pybind11 has a few other functions right before this and right after `serialize`. Seems like the following 4 were somehow missed?
```
        .def("to_xml",
             [](const ImageSpec& spec) { return PY_STR(spec.to_xml()); })
        .def("from_xml", &ImageSpec::from_xml)
        .def(
            "valid_tile_range",
            [](ImageSpec& self, int xbegin, int xend, int ybegin, int yend,
               int zbegin, int zend) {
                return self.valid_tile_range(xbegin, xend, ybegin, yend, zbegin,
                                             zend);
            },
            "xbegin"_a, "xend"_a, "ybegin"_a, "yend"_a, "zbegin"_a = 0,
            "zend"_a = 1)
        .def("copy_dimensions", &ImageSpec::copy_dimensions, "other"_a)
```

src/python-nanobind/py_paramvalue.cpp
soswow: After digging a bit into this (I am still not 100% expert in this) I can answert your question - no, we don't **have to** but for parity, we can. If I understood it correctly, in pybind11 it was required for runtime to keeps the ParamValueList instance alive at least as long as that returned wrapper exists (so you do not drop the list on the Python side and leave a wrapper pointing at freed vector storage).
Without it here it will be making new copies. Whether we **must** change it depends on whether we want parity (and avoid silent extra copies) vs accepting copy semantics as an intentional nanobind default.
For now, I will add it.

soswow: For second override (with string parameter) there is no need for that policy since there is no talk about returning a referenced value and new python object is created regardless.

src/python-nanobind/py_imagespec.cpp
soswow: yeah. fair. I am pretty sure it was minimal impl. to satisfy existing tests to pass. 

I will add those in (And couple more things that got left behind) and see if I Can have regression tests that would have failed if these things are not added.

soswow:
@jessey-git thx for taking a look. See last 2 commits.

src/python-nanobind/py_imagespec.cpp
jessey-git: Thanks for adding the test coverage! The files looked very complete so it seemed like the 4 were accidentally forgotten and it would have been more difficult to notice later on I think. I think, for future nanobind changes that might come in more than 1 piece, any APIs remaining should be clearly marked as a TODO so it's more obvious we can't quite ship it yet.

... Hmm, just thinking out loud: I wonder if we could add coverage for this actually. We could dump the API surface with another test that does something similar to:
```python
import pprint
pprint.pprint(dir(OpenImageIO.ImageSpec))
pprint.pprint(dir(OpenImageIO.ImageBuf))
```
And that way we would have a good comparison against the new nanobind API surface and it would clearly show the differences.  Unfortunately that would mean having to completely implement each class in a single PR...

jessey-git (approved):
I'm approving from my side though do fixup that clang-format issue from the last commit.

soswow:
I wonder if we want those tests to run with nanobind backend in CI? It only happens when `OIIO_PYTHON_BINDINGS_BACKEND` is `both` or `nanobind`, which it is not in CI.
I mean it obv. should at some point. The Q is - should it be this PR?

src/python-nanobind/py_imagespec.cpp
soswow: how about using existing stub at `src/python/stubs/OpenImageIO/__init__.pyi` for that?

src/python-nanobind/py_paramvalue.cpp
lgritz: I think that in this case, we'd like to avoid an unnecessary copy since a ParamValue could, in theory, be holding something arbitrarily large.

INSTALL.md
```diff
@@ -42,6 +42,11 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
      * Python >= 3.9 (tested through 3.13).
      * pybind11 >= 2.7 (tested through 3.0)
      * NumPy (tested through 2.4.4)
+     * For the experimental nanobind migration backend:
+     * NumPy (tested through 2.2.4)
+     * For the nanobind (WIP) migration backend used in source/CMake builds:
+        * nanobind discoverable by CMake, or installed in the active Python
+          environment so `python -m nanobind --cmake_dir` works
```
lgritz: Is this a little messed up? You have two "for the nanobind..." headers, and numpy ends up listed twice.

lgritz:
These all look fine to me. I made a minor comment or two that you can address in the next round if you wish. This is more or less what I pictured the nanobind bindings looking like. 

Let me know if you are ready for me to merge what you have so far, and then we'll do the subsequent rounds on top of it.

Again, does not have to be in this PR, but if not then soon: I think you should check in some kind of design checklist (could be in docs/dev? or as comments at the top of one of the important headers like py_oiio.h?) that enumerates what's done so far and what's left to do, that we can be checking off and adding to with subsequent PRs so it's easy to tell from a glance what is the current status as well as what jobs might be done in parallel by different people, in case anybody is in a position to help.

lgritz:
Naive question -- Do any of our CI job variants test the nanobind path?

soswow:
> Naive question -- Do any of our CI job variants test the nanobind path?

yeah, as I mentioned in [above message](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5084#issuecomment-4323717591) - I don't think so. I feel like it is rather important

soswow:
@lgritz 
> I made a minor comment or two

which is it? I think I've addressed those made [on march 19](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5084#issuecomment-4086347586)

INSTALL.md
soswow: yeah. will fix that now

soswow: fixed

src/build-scripts/ci-startup.bash
```diff
@@ -36,7 +36,7 @@ export PYTHONPATH=$OpenImageIO_ROOT/lib/python${PYTHON_VERSION}/site-packages:$P
 export COMPILER=${COMPILER:=gcc}
 export CC=${CC:=gcc}
 export CXX=${CXX:=g++}
-export OpenImageIO_CI=1
+export OpenImageIO_CI=true
```
soswow: not sure what happened here. Checking.

soswow: reverted.

soswow:
> I think you should check in some kind of design checklist (could be in docs/dev? or as comments at the top of one of the important headers like py_oiio.h?) that enumerates what's done so far and what's left to do, that we can be checking off and adding to with subsequent PRs so it's easy to tell from a glance what is the current status as well as what jobs might be done in parallel by different people, in case anybody is in a position to help.

Check out `src/python-nanobind/MIGRATION_STATUS.md` what do you think?

src/libutil/hash_test.cpp
```diff
@@ -216,7 +215,7 @@ main(int argc, char* argv[])
 
     // fill data with random values so we can hash it a bunch of different ways
     std::mt19937 rng(42);
-    data.resize(round_to_multiple(iterations, 4) / sizeof(data[0]), 0);
+    data.resize(iterations / sizeof(data[0]), 0);
```
soswow: prob. missed merge mistake. investigating.

soswow: yeap. reverted that too.

src/cmake/testing.cmake
```diff
@@ -55,8 +56,12 @@ set(OIIO_TESTSUITE_IMAGEDIR "${PROJECT_BINARY_DIR}/testsuite" CACHE PATH
 # The optional ENVIRONMENT is a list of environment variables to set for the
 # test.
 #
+# The optional ENVIRONMENT_MODIFICATION is a list of environment variable
```
soswow: This feature apperently requires CMake higher then 3.18.2 which is one in this project.

soswow: I've made some changes not to have dependency on this feature so it will work on 3.18.2

soswow:
@jessey-git Unfortunately I've found some more divergent / missed / gaps. Feel free to see latest commits. I've squished (I believe before I Started making new changes) and other commits are all new stuff.

lgritz:
What version does it need? This is for main only, I'm not averse to raising the cmake minimum for 3.2, depending on how recent it needs.



On April 27, 2026 2:19:48 a.m. PDT, Aleksandr Motsjonov ***@***.***> wrote:
>@soswow commented on this pull request.
>
>
>
>> @@ -55,8 +56,12 @@ set(OIIO_TESTSUITE_IMAGEDIR "${PROJECT_BINARY_DIR}/testsuite" CACHE PATH
> # The optional ENVIRONMENT is a list of environment variables to set for the
> # test.
> #
>+# The optional ENVIRONMENT_MODIFICATION is a list of environment variable
>
> This feature apperently requires CMake higher then 3.18.2 which is one in this project.
>
>-- 
>https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5084#pullrequestreview-4179658675
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
--
Larry Gritz
***@***.***

src/cmake/testing.cmake
lgritz: That feature needed 3.22. Something to consider for the OIIO 3.2 release. Requires checking on other projects in the ecosystem and what they need. If lots of others require that or newer, maybe it's time to bump our minimum again. I'm writing this more as a reminder to myself, not as instructions for you.

lgritz:
Thanks for the migration status document. Looks good.

The failing tests (which hit an assertion) are the only thing left to fix before I think we should merge and then build on top of it. I'd like to stop the scope creep on this initial PR as soon as we can. 

After the merge, I believe the very highest priority is rigging it so that we are testing the nanobind path for at least some of the CI test (ideally at least one variant for each OS platform). Then we can move forward with confidence, knowing that downstream users who try building with nanobind will at least never have broken builds.

soswow:
ok folks, take a look at two last commits.
- one is about fixing build on windows
- other contains actual bugs (that's why pybind is touched) that got revieled when I've added more test cases. And also I had to add some more migration into py_oiio.h to make tests pass.

lgritz (approved):

lgritz:
All tests passing, looks good to me as a first step to nanobind.
Does anybody think this shouldn't get merged today?

soswow:
I don't mind. Happy to start looking into making nanobind tests run in CI next

PR #5106, opened 2026-03-22
fix(xmp): Correctly parse XMP with self-closing elements

### LINKED ISSUE #5021: bug: OpenImageIO fails to parse valid XMP when rdf:Description is self-closing
Hi,

I found that OpenImageIO fails to extract XMP metadata from some valid images
when the XMP uses a self-closing `<rdf:Description ... />` element.

Example XMP :
```
<rdf:Description
    xmlns:Camera="xxxx/"
    Camera:CameraModel="xxxx"
    Camera:Pitch="-59/10"
    Camera:Roll="-27/10"
    Camera:Yaw="-459/10"
    ... />
```
This is valid XML, but OpenImageIO's `decode_xmp()` currently only looks for:

    <rdf:Description> ... </rdf:Description>

and therefore does not enter the parsing loop if the tag is self-closing.

Example repro JPEG with self-closing rdf:Description XMP:

https://github.com/user-attachments/assets/30c0f3ec-322e-4f26-8450-e0371e15f1ca

lgritz:
decode_xmp() only searched for `</rdf:Description>` end tags, so self-closing `<rdf:Description ... />` elements (valid XML) were silently skipped. Now detect both `/>` and `</rdf:Description>` endings, plus simplify quite a bit generally, by rewriting decode_xmp to use pugixml more thoroughly to do the whole parse job instead of trying to hunt for tags ourselves.

Fixes #5021

Assisted-by: Claude Code / Opus 4.6

### DIFF
--- a/src/libOpenImageIO/xmp.cpp
+++ b/src/libOpenImageIO/xmp.cpp
@@ -517,54 +517,35 @@ decode_xmp(string_view xml, ImageSpec& spec)
 #endif
     if (!xml.length())
         return true;
-    for (size_t pos = 0;;) {
-        // Find the start of the next rdf:Description element
-        size_t startpos = xml.find("<rdf:Description", pos);
-        if (startpos == std::string::npos)
-            break;
-
-        // Determine end: look for both </rdf:Description> and self-closing />
-        size_t endclose = xml.find("</rdf:Description>", startpos);
-        size_t endself  = xml.find("/>", startpos + 16);
-
-        size_t endpos;
-        if (endclose != std::string::npos) {
-            // Has a closing tag
-            endpos = endclose + 18;  // length of "</rdf:Description>"
-        } else if (endself != std::string::npos) {
-            // Self-closing <rdf:Description ... />
-            endpos = endself + 2;  // length of "/>"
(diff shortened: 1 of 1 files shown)


jessey-git:
Question: If the incoming data is really XML, why not load it entirely into pugi first and then use their API to process each rdf:Description element?

Right now I think the code is still broken? If you provide it this xml fragment:
```
<rdf:Description />
<rdf:Description></rdf:Description>
```
Then the code will find the following: `<rdf:Description /><rdf:Description></rdf:Description>`

But if you give it:
```
<rdf:Description></rdf:Description>
<rdf:Description />
```
It will then find 2 separate tags and process them individually:
```
<rdf:Description></rdf:Description> and then
<rdf:Description />
```

lgritz:
Sorry, that last comment crossed my threshold of "too complicated to think about right now" when it came, and I kept postponing until now.

Great point, and when I tried it, it was a pretty straightforward refactor that greatly simplified the decode_xml function. Thanks for the suggestion. PR updated with this change.

jessey-git (approved):
Nice simplification!

PR #5135, opened 2026-04-09
ImageBuf: `contiguous_scanline()` + fixes to internal span

lgritz:
Add a new ImageBuf method: `contiguous_scanline()`, which reveals if each scanline is contiguous in the ImageBuf's internal memory (no padding or unusual spacing between channels or pixels). It's analogous to the existing `contiguous()`, but the latter also tests contiguity in y and z.

Additional internal fixes:

* Always and only set m_bufspan via calls to `set_bufspan()`, so it's consistent. In particular, it ensures that we correctly set the channel size, which some call sites previously did not do correctly for untyped "byte" buffers.

* Re-evaluate the contiguous flags whenever resetting m_bufspan.

* contiguous() calls need to call validate_pixels().

  An IB from a file that hasn't been read yet into memory yet isn't considered contiguous. But once it's read, it will be (if it's got the strides right, etc.). So when somebody asks if it's contiguous, what does that mean? They are asking because they are about to do something that requires memory contiguity (meaning, it needs to be IN memory), just like if they call localpixels(). So just like localpixels(), we need to call validate_pixels() to assure that if the pixels are destined for the local buffer, it's done now if not yet.

### DIFF
--- a/src/include/OpenImageIO/imagebuf.h
+++ b/src/include/OpenImageIO/imagebuf.h
@@ -1386,13 +1386,22 @@ public:
     /// Z plane stride within the localpixels memory.
     stride_t z_stride() const;
 
-    /// Is this an in-memory buffer with the data layout "contiguous", i.e.,
+    /// Is this an in-memory buffer with the data layout "fully contiguous",
+    /// i.e.,
     /// ```
     ///     pixel_stride == nchannels * pixeltype().size()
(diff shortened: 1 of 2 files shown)


lgritz:
The Mac CI failures appear to be totally unrelated to this PR.

lgritz:
Any comments or concerns?

PR #5136, opened 2026-04-10
testing: hwy testing strategies

lgritz:
1. Some tests are repeated, run once with hwy off and once with hwy on.  The "on" versions append ".hwy" to the test name and set the OPENIMAGEIO_ENABLE_HWY=1 environment variable to force it on while running the test. The advantage to this approach is that the test itself doesn't need to change at all, we just run it separately in each mode.

   In this PR, I do this with oiiotool, oiiotool-xform, and docs-examples-cpp.

   For now, while hwy is not enabled by default, it's doing these extra things to enable hwy. Eventually, if/when hwy is enabled by default, we'll flip the sense and use the env variable to turn it off.

2. Some tests do certain operations both ways when hwy support was enabled at build time. This is more convenient for certain unit test executables, where perhaps we want to benchmark both ways and have those show up right next to each other within the run.

   In this PR, I do this with the imagebufalgo_test unit test.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ else ()
 endif ()
 option (${PROJ_NAME}_BUILD_TOOLS "Build the command-line tools" ON)
 option (${PROJ_NAME}_BUILD_TESTS "Build the unit tests" ON)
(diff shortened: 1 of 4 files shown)


ssh4net:
@lgritz looks good for me.

PR #5137, opened 2026-04-10
int: Some minor cleanup of the hwy related code

lgritz:
* imageio.h: Document 'enable_hwy' global option.

* imagebufalgo_hwy_pvt.h: Remove RoiNChannels -- it's the same as roi.nchannels()

* Macros: we can simply test OIIO_USE_HWY, since it's always defined; there is no need to also test defined(OIIO_USE_HWY).


ssh4net:
that one also good :)

PR #5138, opened 2026-04-10
fix: incorrect test for when it's safe to use hwy code path

lgritz:
We didn't have a complete enough test for when it was safe to hwy on a
buffer. Encapsulate in new function HwySupports so that all the logic
lives in one place and doesn't have to be replicated everywhere. The
more complete test now checks for:

- contiguous scanlines
- right number of channels (only indirectly checked)
- 2D image, not 3D volume (not previously checked)
- pixel data window fully contains the ROI (not previously checked)
- correct data type (should always be true, but was not checked)

This eliminates some test failures, which were specifically cases
where the ROI being operated on did not overlap with the data area of
the ImageBufs, and the Highway related code did not handle those in
the way that the old ImageBuf::Iterator transparently handled running
off the edge and various wrap modes.

(This incorporates PR #5135, which is a prerequisite. I will merge that
PR first and then rebase this PR on top of it, so that the two sets of
changes stay as separate commits. You need not review the imagebuf.{cpp,h}
portions in this PR, they will be reviewed in the other one.)

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_addsub.cpp
+++ b/src/libOpenImageIO/imagebufalgo_addsub.cpp
@@ -138,39 +138,29 @@ static bool
 add_impl(ImageBuf& R, const ImageBuf& A, const ImageBuf& B, ROI roi,
          int nthreads)
 {
-#if defined(OIIO_USE_HWY) && OIIO_USE_HWY
-    if (OIIO::pvt::enable_hwy && R.localpixels() && A.localpixels()
-        && B.localpixels()) {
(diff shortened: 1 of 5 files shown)


ssh4net:
@lgritz, as we discussed, it looks like there is no regression in this code, and just not all possible use cases, which will be added later, are covered.

lgritz:
@ssh4net, do the other two hwy related PRs also look ok to you?

ssh4net:
@lgritz yes, clean and nice

PR #5139, opened 2026-04-10
ci: Unbreak Mac CI by adding jpeg-xl install

lgritz:
It must have been among the homebrew packages in the base Mac installation on GHA, until just a couple days ago?


lgritz:
Merging without review since this is CI-only and fixes CI that is broken in all other PRs because of this issue.

PR #5140, opened 2026-04-11
fix(jpeg): be a little more flexible with corrupt IPTC blocks

lgritz:
Recent PR #5081 added more detection of corruptions in (among other places) the IPTC blocks in a JPEG file.

But we maybe got a little too aggressive, since there appear to be some jpg files out there that have strange (maybe technically incorrect?) IPTC blocks but are otherwise fine.

Anyway, I noticed that for ICC profiles in JPEG files, we use the OIIO attribute "imageinput:strict" to decide whether a localized corruption in the ICC profile ONLY causes the ICC profile to be skipped (reading the pixels just fine), or a stricter mode where anything suspicious is grounds to stop reading the file entirely and declare it corrupt.

So in this patch, I'm using the same control for the analogous purpose when reading the IPTC blocks. By default, just skip the IPTC block if we don't know how to decode it. But if in strict mode, make a full error that will cause us to stop reading the file.

With this change, I needed to update the jpeg-corrupt test to force strict node, and update some some ref outputs.


lgritz:
I need this at work, if any has time to give it a review

aconty (approved):
LGTM

PR #5141, opened 2026-04-11
fix(sgi): Better detection of corrupt RLE info that could overflow

lgritz:
* Change DASSERTs (which were correct, but did not enforce anything in release builds) into real checks that will return error messages and flags if the corruptions are found.
* Switch to using spans for extra checks if anything goes wrong.
* Use check_open.
* Check that various header items are valid enum values.


jessey-git (approved):
Looks good and tests out well locally.

PR #5142, opened 2026-04-12
fix(softimage): Prevent buffer overruns from corrupt RLE

lgritz:
* Check for buffer overruns.
* Use check_open.
* Check for invalid bit depths.


lgritz:
This is a week old. Any comments or objections?

lgritz:
I just realized, I already have a set of changes written that I was waiting for this PR to be merged before submitting, and they include all the suggestions that Copilot just mentioned. 

Since this has gone 10 days without object, I'm going to merge it and then submit my follow-on.

PR #5144, opened 2026-04-14
testing: Fix ref output for jpeg-corrupt tet

lgritz:


lgritz:
Merging without review because it fixes broken CI

PR #5145, opened 2026-04-15
int: Turn on OIIO_NODISCARD_ERROR_ENABLE for internal builds

lgritz:
Rearrange the automatic setting of `OIIO_NODISCARD_ERROR_ENABLE` so that for external (not OIIO itself) builds, we default to enabling for OIIO >= 3.3, and disabling for older and current versions. But now, when building OIIO itself, it is ENABLED regardless of version.

So this means that as we add OIIO_NODISCARD_ERROR annotations to functions, we MUST fix any calls to those functions where we ourselves don't check their return values, and can't accidentally add new breaking code.

Also, had to fix a bunch of spots that were missed in the original PR, places where we didn't patch up our sloppy uses of what was annotated in that patch. I'm not sure I understand how these could have been missed in that last PR, since that PR did fix other uses of the annotated functions.

### DIFF
--- a/src/doc/Doxyfile
+++ b/src/doc/Doxyfile
@@ -2202,6 +2202,7 @@ PREDEFINED             = DOXYGEN_SHOULD_SKIP_THIS \
                          OIIO_CONST_FUNC= \
                          OIIO_MAYBE_UNUSED= \
                          OIIO_NODISCARD:= \
+                         OIIO_NODISCARD_ERROR:= \
                          OIIO_DEPRECATED(x)=[[deprecated]] \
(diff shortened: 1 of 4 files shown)


lgritz:
@mvanhorn does this look right to you?

mvanhorn:
Looks right. Flipping the OIIO_INTERNAL-gated default to always-enable is the right call. It means NODISCARD_ERROR annotations land with their own call-site fixes in the same PR instead of sneaking past our own build.

The test-fix pattern reads clean: OIIO_ASSERT(ok) where correctness matters, (void) in imagespeed_test.cpp so the timing harness doesn't assert inside the hot loop.

PR #5146, opened 2026-04-17
build: Fix Makefile wrapper to properly quote test regex

lgritz:
This fixes a longstanding issue where saying

    make test TEST="foo|bar"

would be malformed, instead of running tests that matched either foo or bar (misinterpreting the `|` as a pipe). The problem was just not quoting it properly.


lgritz:
No objections for a week + harmless + probably only used by me --> merge

PR #5147, opened 2026-04-17
oiiotool: Add --experimental command line argument.

lgritz:
Introducing this as a way to allow us to have experimental oiiotool features or behaviors that are enabled only if this flag is also present.

It's a way of allowing features to be turned on for feedback before we're sure it's in a form that can be a permanent, supported feature, thus allowing it to undergo further change that is exempt from our usual rules about what may change in which versions. By forcing users to also employ this flag to enable the trial behavior, there's no way that they can inadvertently come to depend on a temporary behavior and be surprised if it changes later.


grdanny (approved):
looks reasonable!

PR #5148, opened 2026-04-18
deps: Update to the R3D  SDK version 9.2.0

1div0:
Only version has been changed and one line removed.


lgritz (changes requested):
LGTM

There's a one-line fix needed to make it conform to clang-format that you need to fix before we can merge.

lgritz (approved):

PR #5150, opened 2026-04-20
ci: Try to reduce costs and timeouts for expensive tests in debug mode

lgritz:
* Fix setting and response of OIIO_CI -- it was less functional before than I had thought because we didn't actually define the symbol for compilation.

* More aggressive cost reduction when combining CI + Debug builds, reduce resolution of some benchmarks, etc., in addition to reducing iterations as we had done before.

* Mark the most expensive unit tests with a cost, so CTest knows to start them first. (more explanation below)

* Add Testing/Temporary/CTestCostData.txt to the cache-save paths -- that preserves CTest's understanding of how much each test costs so it can schedule it more intellegently for the next run.


lgritz:
Will merge by the end of the weekend if there are no objections.

PR #5154, opened 2026-04-20
feat(IBA): Add FLIP perceptual image difference metric

lgritz:
Implements the FLIP algorithm from Andersson et al. 2020, rewritten with OIIO idioms (no source code from the BSD-licensed reference FLIP.h is directly incorporated). Included are C++ and Python API for ImageBufAlgo::FLIP_diff(), and `oiiotool --flipdiff` command.  The basic operation is to compare two images and produce a per-pixel error map that conveys perceptual difference to human observers.

I won't explain it all here; see the extensive comments in imagebufalgo.h, imagebufalgo_flip.cpp, oiiotool.cpp, imagebufalgo.rst, oiiotool.rst.

There are some important changes (especially for how we expose this via oiiotool) versus the way the original NVIDIA reference implementation's command line tool worked. Please read the extensive comments at the top of imagebufalgo_flip.cpp for details.

This is a preliminary, experimental implementation. It's hidden behind an `experimental` namespace (and the oiiotool command requires use of the `--experimental` argument) to emphasize that it may change and is not yet considered part of OIIO's public API, and thus is exempt from our usual strict rules about breaking backward compatibility. Try it out and give feedback, but do not rely on this yet!

Assisted-by: Claude Code / sonnet-4.6 + opus-4.6

I used Claude Code for the inital stab at transforming the NVIDIA reference implementation into OIIO idiomatic equivalents. But to be honest, that got me over the hump of the blank page, but I rewrote most of it bit by bit as I continued to refine my design ideas for how it should work and be exposed to users.


lgritz:
Maybe @grdanny or @chellmuth can give this a review?

I know that the implementation is complex, but remember: (a) it is still in flux with many planned improvements, so there is no point over-scrutinizing code that is still changing; (b) the tests confirm that it matches the numerical results of the NVIDIA reference implementation, so we know it is correct. I'm more concerned with any big complaints you have about the APIs or overall design and vision of this feature, before we merge and let people start playing with it.

chellmuth (commented):
From a high level, looks pretty solid to me!

lgritz:
I'm going to take that as an approval so we can merge. It's experimental and still under development, so if we spot anything later that we missed, fixes can be folded in with the subsequent PRs that will continue to advance this feature.

PR #5155, opened 2026-04-22
fix(softimage): More hardening against corrupted input files

lgritz:
* Fix incorrect return check for fread in read_pixels_pure_run_length.
* Detect and report error for unsupported packet type.
* Fix possible infinite loop from malformed RLE packet with a zero run-count.
* Check for valid y range in read_native_scanline.


lgritz:
Release approaching, no objections for a week -> merging tomorrow if I don't hear anything to the contrary

PR #5156, opened 2026-04-22
fix(softimage): Various fixes for malformed channel packets

lgritz:
* Fix bugs when channel order isn't strictly R,G,B,A by using a `m_channel_map`.
* Detect malformed file with too many channel packets.
* Gracefully handle duplicate channel bits across packets by not thinking that there are duplicate channels.
* Guard against channel packets that contain no channels.

Assisted-by: Claude Code / Opus 4.6
(mostly for analysis/identification of the potential problems; touching the code only for the out-of-order channels fix)


lgritz:
Release approaching, no objections for a week -> merging tomorrow if I don't hear anything to the contrary

PR #5160, opened 2026-04-23
fix(tiff): care with missing rowsperstrip

lgritz:
According to the TIFF spec, TIFF files with a missing rowsperstrip should assume the whole image is one strip.

Somewhat related (found in debugging): add a DASSERT to help debug if fmath.h's round_to_multiple() is given an invalid denominator.

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -1309,8 +1309,13 @@ TIFFInput::readspec(bool read_meta)
     m_rowsperstrip = -1;
     if (!m_spec.tile_width) {
         TIFFGetField(m_tif, TIFFTAG_ROWSPERSTRIP, &m_rowsperstrip);
-        if (m_rowsperstrip > 0)
+        if (m_rowsperstrip > 0) {
+            // Only set the attrib if a legit value was found in the file
             m_spec.attribute("tiff:RowsPerStrip", m_rowsperstrip);
+        } else {
+            // Default if not found is "one strip for the whole image"
+            m_rowsperstrip = m_spec.height;
+        }
     }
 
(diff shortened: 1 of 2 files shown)


src/tiff.imageio/tiffinput.cpp
```diff
@@ -1309,8 +1309,13 @@ TIFFInput::readspec(bool read_meta)
     m_rowsperstrip = -1;
     if (!m_spec.tile_width) {
         TIFFGetField(m_tif, TIFFTAG_ROWSPERSTRIP, &m_rowsperstrip);
-        if (m_rowsperstrip > 0)
+        if (m_rowsperstrip > 0) {
```
jessey-git, line 1312: I think we should also cap the m_rowsperstrip to the height of the image in all cases since downstream computations and memory allocations use the value too.

lgritz: Good suggestion. Done.

jessey-git: Hmm, is there a reason not to clamp the value before setting the tiff:RowsPerStrip attribute value too?

lgritz: Yes, two reasons:
(1) The attribute should reflect what's actually in the file (for the sake of `iinfo -v` and the like), and there's nothing invalid about having it larger than the number of scanlines (just like happens for the last strip any time the strip size does not evenly divide into the height);
(2) I want it to make its way to the output image and be set there as well, so that something like `iconvert in.tif out.tif` will be as close as possible to `cp in.tif out.tif` and not change TIFF tag values that are perfectly legal.

jessey-git: Gotcha

jessey-git (approved):

PR #5164, opened 2026-04-23
fix: Relax OpenColorIO exception handling

lgritz:
In PR #5114, just a few weeks ago, I was trying to catch every possible OCIO exception. That was a good idea, but I was over-eager and treated every OCIO exception as if something really bad happened. I should have recognized that OCIO enjoys throwing exceptions for things where I would have just used an empty return value to indicate a simple failure to do what was asked.

So this patch is revising the approach as follows:

* Exceptions thrown because it couldn't read a config properly (it was missing or broken), or couldn't produce a ColorProcessor when asked -- those are real problems that we should treat as serious errors.

* Exceptions thrown because it can't satisfy a simple request (like "what is the name of the color space for this role" in a case where the role you passed is not known by the config) -- we should catch those exceptions, but it's not an "error" for OIIO, we just return an empty string because there is no color space for that role.

Also beefed up some of the documentation of our ColorConfig class methods to describe what happens with certain kinds of failures.

### DIFF
--- a/src/include/OpenImageIO/color.h
+++ b/src/include/OpenImageIO/color.h
@@ -121,11 +121,12 @@ public:
     int getColorSpaceIndex(string_view name) const;
 
     /// Get the name of the color space representing the named role,
-    /// or NULL if none could be identified.
+    /// or nullptr if none could be identified.
     const char* getColorSpaceNameByRole(string_view role) const;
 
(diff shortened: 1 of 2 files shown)


lgritz:
@meimchu seem ok?

zachlewis:
This seems like the right way to go, to me...!

zachlewis (approved):

PR #5167, opened 2026-04-28
dicom: enforce reasonable resolution limits to guard against corrupt file

lgritz:


lgritz:
Any comments on this one?

src/dicom.imageio/dicominput.cpp
```diff
@@ -231,6 +231,8 @@ DICOMInput::seek_subimage(int subimage, int miplevel)
 
     m_spec = ImageSpec(m_img->getWidth(), m_img->getHeight(), nchannels,
                        format);
+    if (!check_open(m_spec, { 0, 1 << 20, 0, 1 << 20, 0, 1 << 16, 0, 1 << 16 }))
```
jessey-git, line 234: That's a very big image... 62bits required if utilizing the entire space there :)  I wasn't able to find the format limits going through their docs but this seems reasonable until we're told otherwise.

jessey-git (approved):
Seems alright given my comment below.

src/dicom.imageio/dicominput.cpp
lgritz: These are the limits on individual dimensions. There's a separate limit on the total (w x h x d x chans x channelsize). It won't really allocate for a 1M x 1M image unless the app raises that limit via an attribute, but why not let it do 1M x 16? DICOM is for scientific and medical data, so it may have weirder shapes than the kinds of images that we generally use.

jessey-git: Yeah, still protected by "imagesize_MB" setting, it was just a general "Wow that would be a big image, I hope the doc never needs something like that for me."

PR #5171, opened 2026-04-29
fix: immediate fix to PR 5154, uncheked return

lgritz:
PR #5154 crossed streams with another that turned on more warnings for unchecked errors. The check isn't necessary here, though.


lgritz:
I'm about to get on a plane and am not sure if I'll have internet access. If this passes CI, can somebody with the ability please merge it when it's done? Thanks.

jessey-git (approved):

PR #5153, opened 2026-04-20
fix(rla): Harden against corrupted files

lgritz:
1. Harden against buffer overruns from malformed RLE blocks. The simple fix is the `if (count > ...)` clauses on lines 503 & 510, but I also took the opportunity to rewrite decode_rle_span using spans rather than raw pointers.

Fixes #5152

2. Clean up seeking methodology to be more robust.

Fixes #5159

PR #5172, opened 2026-04-29
fix(rla): immediate fix for PR #5153 error

lgritz:
Caught in review, but I didn't see the comment before I merged.


lgritz:
To make it pass CI, this incorporates the fixes that are being reviewed separately in #5171. Let's be sure to merge 5171 BEFORE merging this one to keep them straight.

I'm about to get on a plane, not sure if I'll have internet. In the interest of not having a broken main, can somebody with merge permissions please keep an eye on the CI here (and 5171) and merge 5171, and THEN this one, assuming that both CI runs pass ok? Thanks.

jessey-git (approved):

PR #5173, opened 2026-04-30
feat(iconvert): allow -o for convenience

lgritz:
iconvert was the very first command line utility for OIIO, and its syntax was just `iconvert [options] infile outfile`. YEARS later, oiiotool came along, superceding iconvert with 100x the functionality, and also differed by requiring `-o` to specify output file.

iconvert still exists for testing and examples -- iconvert uses just plain old ImageInput + ImageOutput straightforwardly, whereas oiiotool is a very complex app that's ImageBuf powered. It's often useful for debugging to test the ImageInput/ImageOutput layer without any extra complication, or to directly see how going through an ImageBuf might differ.

The fact that oiiotool requires `-o` to specify an output file, whereas iconvert considers that an error, is a minor annoyance when switching back and forth while debugging, and it's really bugging me.

So here it is, a tiny patch to let iconvert silently accept -o, or not, and do the obviously right thing. A little quality of life gift to myself.

### DIFF
--- a/src/doc/iconvert.rst
+++ b/src/doc/iconvert.rst
@@ -19,7 +19,8 @@ certain metadata to the image.
 
 The `iconvert` utility is invoked as follows:
 
-    `iconvert` *optiions input output*
+    `iconvert` *options input output*
+    `iconvert` *options input* `-o` *output*
 
 Where *input* and *output* name the input image and desired output filename.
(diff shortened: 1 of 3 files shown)


lgritz:
Any comments? Will merge over the weekend if nobody objects.

PR #5174, opened 2026-04-30
fix(jpeg): more correctly handle metadata marker bounds

ssh4net:
### Description

Fixes bounds checks for malformed JPEG APP1 Exif and APP2 ICC metadata
markers.

Before this change, APP1 markers were treated as Exif when they started with
`"Exif"` and had at least 4 bytes. The reader then skipped 6 bytes for the
`Exif\0\0` prefix, so short payloads such as `Exif` or `Exif\0` could advance
past the saved marker buffer and underflow the remaining length passed to the
Exif parser.

The APP2 ICC path also used `strcmp()` on raw marker bytes before proving the
marker was long enough or NUL-terminated. Short APP2 markers could therefore be
read past the saved marker buffer, and the code later accessed sequence bytes
and subtracted `ICC_HEADER_SIZE` without first checking the marker length.

This is reachable by normal OpenImageIO image-open paths, including tools such
as `iinfo`, `iconvert`, and `oiiotool`, when opening a malformed JPEG.

The fix adds bounded marker-prefix helpers:
- APP1 Exif is accepted only when the full `Exif\0\0` prefix is present.
- APP2 ICC is accepted only when the marker is at least `ICC_HEADER_SIZE` bytes
  and matches `ICC_PROFILE\0` using bounded comparison.

### Tests

Added coverage to `testsuite/jpeg-corrupt` for short APP1/APP2 metadata
markers:
- APP1 Exif payload lengths 4 and 5
- APP2 ICC payload lengths 11, 12, and 13

Validated locally with:

`ctest --test-dir build-jpeg-only -R '^jpeg-corrupt$' --output-on-failure`

### Checklist:

  and if I used AI coding assistants, I have an `Assisted-by: Codex GPT 5.5xHigh
  behavior.
  on the GitHub runners.)


testsuite/jpeg-corrupt/run.py
```diff
@@ -25,5 +28,13 @@
 # This file has a corrupted ICC profile block
 command += info_command ("src/corrupt-icc-4552.jpg", safematch=True)
 
+# These files have short APP1/APP2 metadata marker payloads that used to be
+# read past their saved-marker buffers before being ignored.
+command += run_app(oiiotool("short-exif-app1-len4.jpg --echo short-exif-app1-len4-ok"))
```
lgritz: I'm not so sure that this is testing anything. oiiotool will read lazily. Saying "oiiotool filename" but then not having any subsequent commands that NEED either the metadata or the pixel data might not even open the file.

Either of these will cause a full read of the file, but without much extra work:

```
iinfo -a -v --hash filename.jpg 
iconvert in.jpg out.null
```

For iinfo, it prints (and therefore has to read) the metadata by default, and the `--hash` will cause it to need all the pixel values (and also do the hash computation on it).

For iconvert, calling the output file "*.null" will use the "null" driver that doesn't actually write anything (no true I/O, no leftover file), but goes through all the motions including reading the full input file.

lgritz: iconvert-to-null is probably the least wasteful way to force a full read of a file. It will make calls to a "writer", but it's a fake writer that doesn't do anything.

ssh4net: Fixed

testsuite/jpeg-corrupt/run.py
```diff
@@ -8,6 +8,17 @@
 failureok = 1
 redirect = ' >> out.txt 2>&1 '
 
+command += oiiotool("--create 1x1 3 -d uint8 -o base-short-marker.jpg")
+command += run_app(pythonbin + " src/make-short-marker-jpegs.py", silent=True)
+
+
+def iconvert_to_null(filename, success_marker):
+    output = filename.rsplit(".", 1)[0] + ".null"
```
lgritz: Since null files never get written, there is not a reason to carefully construct the filename based on the input, just always call it "out.null" and be done with it. 

I don't think you need the iconvert_to_null function at all. I would modify runtest.py's `iconvert()` function to take a new optional argument `success_message` and if it's non-empty, have it append `"&& echo" + success_message` (much in the way that it appends `|| true` if you set failureok), then for the test you can just say

```
command += iconvert("foo.jpg out.null", successmessage="foo-ok")
```

lgritz (changes requested):
LGTM.

I made a suggestion for simplifying the test logic that boils down to:
* 2-line extension to runtest.py's `iconvert` function.
* No custom logic in this one test directory's run.py
* Just use `command += iconvert(...)` with the new option

Other than that, this is ready to go, thanks for the fix.

lgritz (approved):
LGTM

PR #5175, opened 2026-04-30
fix(webp): fix exif chunk bounds handling

ssh4net:
### Description

Fixes bounds handling for WebP EXIF metadata chunks.

Before this change, the WebP reader always skipped the first 6 bytes of an EXIF
chunk before calling `decode_exif()`:

```cpp
bytes + 6
size - 6
```

That assumed every WebP EXIF chunk was at least 6 bytes long and always started
with the optional Exif\0\0 prefix. A malformed WebP file with a shorter EXIF
chunk could therefore advance the pointer past the chunk payload and underflow
the remaining size passed to the EXIF parser.

This is reachable through normal OpenImageIO image-open paths, including tools
such as iinfo, iconvert, and oiiotool, when opening a malformed WebP file.

The fix changes the WebP reader to treat the EXIF chunk as a bounded byte span.
Before calling decode_exif(), it now verifies that the payload contains enough
bytes for a TIFF header, either at the start of the chunk or after an optional
Exif\0\0 prefix. The WebP reader no longer performs an unconditional 6-byte
skip; the shared EXIF decoder already handles the optional prefix safely.

### Tests

Added WebP regression coverage for malformed short EXIF chunks.

The testsuite/webp test now generates WebP files with EXIF chunk payload
lengths:

- 0 bytes
- 4 bytes: Exif
- 5 bytes: Exif\0
- 6 bytes: Exif\0\0
- 13 bytes: Exif\0\0 plus a truncated TIFF header

Each generated file is opened through oiiotool to verify that malformed EXIF
metadata does not crash or read past the chunk buffer.

Validated locally with:

python3 -m py_compile testsuite/webp/src/make-short-exif-webp.py
git diff --check
cmake --build build-webp --target oiiotool -j 8

Also ran direct oiiotool smoke tests for all generated malformed WebP inputs.

### Checklist:

  and if I used AI coding assistants, I have an `Assisted-by: Codex GPT5.5xHigh`
  behavior.
  on the GitHub runners.)


testsuite/webp/run.py
```diff
@@ -14,3 +14,23 @@
 
 command += oiiotool ("--create 64x64 4 -o rgba.webp")
 command += info_command ("rgba.webp", "--iconfig oiio:UnassociatedAlpha 1", safematch=True)
+
+
+def iconvert_to_null(filename, success_marker):
+    output = filename.rsplit(".", 1)[0] + ".null"
+    app = "(" + oiio_app("iconvert") + filename + " " + output
+    app += " && echo " + success_marker + ")"
+    return run_app(app)
```
lgritz: As I suggested in the other review, I think if this utility is showing up in multiple tests' testsuite/*/run.py file, it's probably better to just add it once by extending the `iconvert()` function in runtest.py to take an extra optional parameter that can do this.

lgritz (approved):
LGTM!

PR #5176, opened 2026-04-30
ci: Run both tests pybind11 and nanobind in CI

soswow:
Follow up to https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5084

Here we make at last one OS to run both tests

Assisted-by: Cursor/Composer-2


soswow:
works

image

lgritz (approved):
LGTM

PR #5178, opened 2026-05-03
Fix/softimage channel bounds

ssh4net:
### Description

  Fixes malformed Softimage PIC channel packet handling.

  Before this change, the reader counted the number of channels in the file, but decoded pixels using fixed channel indexes: R=0, G=1, B=2, A=3. A malformed PIC could advertise only an alpha channel, producing a one-channel ImageSpec, while decode still wrote to channel index 3. That could write past the scanline
  buffer in iconvert, iinfo --stats, or other callers that read pixel data from an untrusted .pic.

  After this change, Softimage channel packets are validated during open() before the output spec is accepted. The reader now rejects empty, duplicate, unsupported, mixed-depth, and sparse channel packet maps instead of accepting a layout that decode cannot safely write into the declared channel count.

  ### Tests

  Added regression coverage to testsuite/softimage.

  The test now generates small malformed PIC files with alpha-only sparse channel packets for the uncompressed, pure-RLE, and mixed-RLE decode paths. Each file is passed through iconvert to out.null, and the test records a rejection marker when the malformed input is rejected.
### Checklist:

  and if I used AI coding assistants, I have an `Assisted-by: Codex GPT5.5 xHigh`
  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/softimage.imageio/softimageinput.cpp
+++ b/src/softimage.imageio/softimageinput.cpp
@@ -114,6 +114,7 @@ SoftimageInput::open(const std::string& name, ImageSpec& spec)
 
     // Get the ChannelPackets
     ChannelPacket curPacket;
+    int bits_per_channel = 0;
     std::vector<std::string> encodings;
     do {
         // Read the next packet into curPacket and store it off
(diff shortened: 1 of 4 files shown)


lgritz:
Can you rebase this one on top of current main? I think it now conflicts slightly with the other I just merged.

lgritz:
Don't want to let this keep languishing. There appears to be something going on with the Windows jobs, they are producing different output?

PR #5179, opened 2026-05-04
ci: need ref output for libheif 1.17

lgritz:
This is mostly to get it to pass when the GHA runners can't add-apt-repository ppa:strukturag when it's glitchy down.  It can fall back to an older libheif 1.17 that's already on those runners, but needs a matching reference output for that version. This is it.


lgritz:
I'm going to merge this immediately because it's holding up overdue releases.

The CI is not fully going to pass anyway, because this only fixes SOME tests (the ones where it libheif can't be installed from the repo, but an older one already exists on those runner configurations, so it only needs a matching reference output). There are other runner configs where there is not a within-rang libheif installed at all, and it's failing to connect to the repo. I will fix that separately.

lgritz:
#5180 fixes the other case, where some runners have no libheif at all, and we can't connect to the server to get one

PR #5180, opened 2026-05-04
ci: Temporary fix -- make libheif optional for Ubuntu jobs

lgritz:
This will get us through the great Canonical attack of May 2026, where it seems that `add-apt-repository ppa:strukturag` is still unreliable even as we are now multiple days overdue to get our releases out.

We will revert this after everything is fixed. I don't like it silently passing when we think it's testing heif files but it's really not, any longer than we have to be in that state.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -353,7 +353,7 @@ jobs:
       # Override required_deps to be 'all' and explicitly list as optional
       # only the ones we are intentionally not testing for those jobs.
       required_deps: ${{ matrix.required_deps || 'all' }}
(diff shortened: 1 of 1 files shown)


lgritz:
I'm going to merge this immediately after CI finishes because it's holding up overdue releases.

The CI is not fully going to pass anyway, because this only fixes SOME tests (the ones where some runners have no libheif at all, and we can't connect to the server to get one).

Other jobs are on runners where it libheif can't be installed from the repo, but an older one already exists on those runner configurations, so it only needed a matching reference output. That is fixed separately by PR #5179.

I swear that the combination of these two gets us to a point where ALL BUT ONE job variant pass. The only holdout is the "bleeding edge" test, where it's still failing for a different reason -- somehow it's still having trouble connecting to yet another server repo where it gets gcc15 from. But we are used to allowing a failing "bleeding edge" CI test to pass, so I'm not going to rig it to pass, we'll just wait out the crisis on that one and do our releases anyway.

PR #5190, opened 2026-05-09
int: Add some more debugging help for OCIO exceptions

lgritz:
In PR #5164, we changed to not be quite as aggressive in treating OCIO exceptions as real errors. All still good with that, but sometimes it is helpful to know exactly what OCIO wants to tell us, for debugging.

In this PR, we allow our color debug mode (triggered when environment variable OIIO_DEBUG_COLOR is set to nonzero) to print debugging messages that reveal the exception messages. It's a debugging aid for developers, no users will want this. In ordinary operations, we really do want to just ignore those exceptions. But if something is misbehaving and we don't know why, it may be helpful to know which of the exceptions is being caught and what the message was.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -446,7 +446,8 @@ ColorConfig::Impl::inventory()
                     scene_linear_alias = lin->getName();
                 return;  // If any non-"raw" spaces were defined, we're done
             }
-        } catch (...) {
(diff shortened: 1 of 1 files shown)


lgritz:
@zachlewis This look ok to you?

zachlewis (approved):
I think this is a great idea! As you say, OCIO raises a lot of exceptions... I was wondering how to gracefully handle this with OIIO, and of course this is the most intuitive way. Nicely done. LGTM.

PR #5191, opened 2026-05-10
api(ColorConfig): add isData, fix python isColorSpaceLinear.

lgritz:
* Add `ColorConfig::isData(name)` method to C++ and Python APIs.
* Fix `ColorConfig.isColorSpaceLinear()` that was missing from the Python bindings. (Tests added as well.)

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -610,7 +610,8 @@ ColorConfig::Impl::classify_by_name(CSInfo& cs)
                    ACEScg_alias);
     } else if (Strutil::iequals(cs.name, "Rec709")) {
         cs.setflag(CSInfo::is_Rec709, Rec709_alias);
-    } else if (Strutil::iequals(cs.name, "Raw")) {
+    } else if (config_
+               && Strutil::iequals(cs.name, config_->getCanonicalName("data"))) {
         cs.setflag(CSInfo::is_data);
     }
 #ifdef OIIO_SITE_spi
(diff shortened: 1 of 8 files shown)


lgritz:
@zachlewis Does this look good to you?

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -607,6 +610,8 @@ ColorConfig::Impl::classify_by_name(CSInfo& cs)
                    ACEScg_alias);
     } else if (Strutil::iequals(cs.name, "Rec709")) {
         cs.setflag(CSInfo::is_Rec709, Rec709_alias);
+    } else if (Strutil::iequals(cs.name, "Raw")) {
```
zachlewis: I'm not in favor of these string-based assumptions.  Aren't we catching these cases with OCIO's API in a subsequent classification stage anyway?

In any case, if you do want to have a fast path to selecting color spaces we know are data spaces, without having to actually inspect them, you can try looking up the "data" role, which is MUCH more likely to point to a data color space than assuming the current config has a color space named "Raw", and that "Raw" is a data color space. 

You can try Config::getCanonicalColorSpaceName("data"), which will either return the name of a color space that can be mapped to the "data" colorInteropID (or any other data color space needs); or it will return something falsey if the "data" role isn't set (I think it usually is -- IIRC, configs fail to validate otherwise).

zachlewis (changes requested):
I think this is the right direction for sure; would prefer you check the "data" role instead of assuming "Raw" always points to a data space. Other than that, LGTM!

src/libOpenImageIO/color_ocio.cpp
lgritz: I see your point here, and amended the PR with a change that uses getCanonicalColorSpaceName rather than hard-coding the name.

PR #5193, opened 2026-05-13
feat: allow freeing of ustring table for debugging cases

### LINKED ISSUE #3913: memory leak in ustring
ustring cpp
```
    ustring::TableRep* make_rep(string_view str, size_t hash)
    {
        char* repmem = pool_alloc(sizeof(ustring::TableRep) + str.length() + 1);
        return new (repmem) ustring::TableRep(str, hash);
    }
```
```

        ustring::TableRep* rep = make_rep(str, hash);
        entries[pos]           = rep;
```
all entries are not freed.  The only call I see is

`        free(entries);
`

lgritz:
This patch optionally frees the contents of the ustring table as the process is exiting. It's triggered one of two ways:

- `OIIO::attribute("ustring:cleanup", 1);`
- Environment variable `OIIO_USTRING_CLEANUP=1`

If neither is used, the teardown won't happen, thus avoiding any pointless work when it should be exiting the process.

The purpose of this is that when using certain debugging tools, the ustring table (which only grows and never frees) looks like a memory leak, even though it's by design. Someimes, it's useful to make it not appear ot leak, and you are happy for a large table to spend time pointlessly freeing all the data.

Fixes #3913

Assisted-by: Claude Code / Sonnet 4.6

### DIFF
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -764,10 +764,6 @@ OIIO_UTIL_API int oiio_ustring_cleanup = Strutil::stoi(
 static int ustring_cleanup_atexit_registered = []() {
     std::atexit([]() {
         if (pvt::oiio_ustring_cleanup) {
-#ifndef NDEBUG
-            OIIO::print("ustring: freeing table resources ({} bytes)\n",
-                        v3_1::ustring_table().get_memory_usage());
-#endif
             v3_1::ustring_table().free_resources();
             v3_1::reverse_map().clear();
         }
--- a/src/libutil/ustring.cpp
+++ b/src/libutil/ustring.cpp
@@ -15,6 +15,24 @@
 #include <OpenImageIO/unordered_map_concurrent.h>
 #include <OpenImageIO/ustring.h>
 
+
+
+OIIO_NAMESPACE_BEGIN
+namespace pvt {
+
+// If nonzero, the ustring table will be freed at process exit. This is off by
+// default because cleanup is unnecessary (the OS reclaims the memory) and can
+// add measurable time at exit for large tables. Enable it when using valgrind
+// or other leak detectors to suppress false positives. Settable via
+// OIIO::attribute("ustring:cleanup",1) or the OIIO_USTRING_CLEANUP
+// environment variable.
+OIIO_UTIL_API int oiio_ustring_cleanup = Strutil::stoi(
+    Sysutil::getenv("OIIO_USTRING_CLEANUP"));
+
+}  // namespace pvt
+OIIO_NAMESPACE_END
+
+
 OIIO_NAMESPACE_3_1_BEGIN
 
 // Use rw spin locks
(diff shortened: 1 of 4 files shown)


jreichel-nvidia:
Thanks for the quick turnaround! But the allocations for entries and pool in `clear()` are still leaking. Are these necessary? Maybe set to `nullptr`, rename `clear()` to `destroy()`, and add some documentation about the purpose/consequences of that method?

lgritz:
I updated with some tightening up of the allocation logic, including shrink_to_fit on the vectors.

Also, I added a print message when it does the free (for Debug builds only, and only if you have turned on the ustring cleanup), maybe you can verify that it's really running the cleanup.

If it's printing that and still complaining about leaks, is it able to tell you which allocation it is that's still never freed?

jreichel-nvidia:
The cleanup code is running. It still shows that the `calloc` for `entries` in `ustring.cpp:73` is leaking. It no longer shows a leak for `pool`, although I don't understand why your 2nd commit would change the behavior there ...

Edit: the print message is just for now for debugging the PR? I'd prefer if it would not print anything later, even not in Debug builds.

lgritz:
I understand why. Stay tuned, will have an amendment coming soon.

lgritz:
@jreichel-nvidia Amendment pushed. Any better?

jreichel-nvidia:
Yes, now it works! What about the debug output? Would it be possible to get rid of that?

lgritz:
> Yes, now it works! What about the debug output? Would it be possible to get rid of that?

Absolutely! That was only there to help you figure out for sure if the destructor was being called. Will update.

lgritz:
Wait, I screwed this up. By overzealously using unique_ptr, I've actually converted the code to doing the deallocations I once wished to avoid. I need to rework this again, sorry.

lgritz:
I took another pass at this, simplifying quite a bit (the whole atexit was not necessary, I could have done it in the destructors all along) and also ensuring that most of the deallocations are skipped when the special flag is not set.

Sorry to make you revisit this, @jreichel-nvidia, but can you try once more to check that what I have now still fixes all the false-positive leaks you were detecting? Thanks.

jreichel-nvidia:
Tests with 3ddc703 are good!

lgritz:
@fpsunflower I think you were the last person to overhaul how this hash table worked. Does this look ok to you?

src/libutil/ustring.cpp
```diff
@@ -146,31 +179,36 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
+
+        // Copy each entry from old into the new, recomputing the hash because
+        // the mask has changd.
         for (size_t i = 0; to_copy != 0; i++) {
-            if (entries[i] == 0)
+            if (old_entries[i] == nullptr)
                 continue;
-            size_t pos = entries[i]->hashed & new_mask, dist = 0;
+            // i is old position, pos will be new position
+            size_t pos = old_entries[i]->hashed & mask, dist = 0;
```
fpsunflower, line 198: Don't you want to keep using `new_mask` here? Where is the mask updated?

I think it makes more sense to keep the update of the mask in sync with the update of the number of entries.

src/libutil/ustring.cpp
```diff
@@ -187,29 +228,62 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
-    ustring::TableRep** entries;
-    size_t num_entries = 0;
-    char* pool;
-    size_t pool_offset = 0;
-    size_t memory_usage;
+    // Allocate one more standard POOL_SIZE block for `pool`
+    void allocate_pool_block()
+    {
+        memory_usage += POOL_SIZE;
+        pool        = new char[POOL_SIZE];
```
fpsunflower, line 247: I would either use `make_unique` here if you want to use `std::unique_ptr` or stick to malloc and add a `free` in the destructor and then you wouldn't have to do any work at all (like before) in the "let it leak" case.

I think this code is low level and self-contained enough that `unique_ptr` doesn't buy you much and just adds compilation overhead.

src/libutil/ustring.cpp
```diff
@@ -187,29 +228,62 @@ template<unsigned BASE_CAPACITY, unsigned POOL_SIZE> struct TableRepMap {
...
+        // case where the internal std::string aliases the pool chars.
+        for (auto& e : entries) {
+            if (e) {
+                e->~TableRep();
+                e = nullptr;
+            }
+        }
+    }
+
+    void resize_entries(size_t newsize)
```
fpsunflower, line 264: If I'm reading this right, this method is only ever called from the constructor - so does it really need to be split as its own method?

lgritz:
Alternate, simpler stab at this posted as PR #5213

lgritz:
Closing, subsumed by #5213.

PR #5194, opened 2026-05-14
fix(color mgmt): More sensible OCIO file rules

lgritz:
I recently discovered that OCIO's built-in configs' "file rules" that come into play when using getColorSpaceFromFilepath actually do not incorporate the search for color space name embedded in the filename into the logic, and moreover, assume that files ending in ".exr" are ACES2065-1, of all things.  This seems an unfortunate choice because in the real world, almost no exr files people will encounter actually use the ap0 primaries.

This PR does the following to shore up this and related edge cases:

* For OCIO built-in configs, fully replace the existing file rules with a blank slate that just uses the path parsing rule. Everything else is categorized as "unknown". (Though many of our format readers already do something better, setting the metadata if they actually know what the file is.)

* Recognize "unknown" as equivalent to having an empty color space, an error if you pass such a color space name to IBA color transform functions.

* For all IBA color conversion, if the input space is not color managed, be sure NOT TO mark the output color space as the original `to` request! Input raw means output is also raw.

### DIFF
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -833,13 +833,17 @@ fix_config_file_rules(OCIO::ConfigRcPtr& config)
         DBG("  rule {}/{}: pat='{}' ext='{}' -> \"{}\"\n", i, rules->getName(i),
             rules->getRegex(i), rules->getExtension(i),
             rules->getColorSpace(i));
+#if 0
+        // If we wanted to doctor just the exr rule, here's how:
         if (Strutil::iequals(rules->getExtension(i), "exr")) {
             // Change the rule for exr extension, if it exists, to "unknown".
             // Make no assumptions. OCIO's built-in configs think it should be
             // ACES2065-1, which is almost never right.
             rules->setColorSpace(i, "unknown");
             DBG("    changed cs to \"{}\"\n", rules->getColorSpace(i));
-        } else if (!strcmp(rules->getName(i), "Default")) {
+        } else
+#endif
+        if (!strcmp(rules->getName(i), "Default")) {
             // Default rule or one that matches everything -- for OIIO, we
             // just want to change this to unknown. We made decisions about
             // default per-file-format color space decisions in the individual
@@ -2402,20 +2406,6 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
     if (!colorconfig)
         colorconfig = &ColorConfig::default_colorconfig();
 
-    if (colorconfig->isData(from) || colorconfig->equivalent(from, "raw")) {
-        // If the input color space is not color managed, the transformation
-        // is meaningless, just set 'to' to the same space so that the whole
-        // thing is a no-op, including continuing to think it's in the raw
-        // space.
-        to = from;
-    } else if (colorconfig->isData(to) || colorconfig->equivalent(to, "raw")) {
-        // If the output color space is not color managed, the transformation
-        // is meaningless, just make it look like it's already in the 'from'
-        // space. Note that this DOES change the apparent color space label
-        // to the requested raw output space.
-        from = to;
-    }
-
     ColorProcessorHandle processor
         = colorconfig->createColorProcessor(colorconfig->resolve(from),
                                             colorconfig->resolve(to),
@@ -2433,7 +2423,10 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
     logtime.stop(-1);  // transition to other colorconvert
     bool ok = colorconvert(dst, src, processor.get(), unpremult, roi, nthreads);
     if (ok) {
+        // Coming from a non-color space preserves the original space
         // DBG("done, setting output colorspace to {}\n", to);
+        if (colorconfig->isData(from))
+            to = from;
         dst.specmod().set_colorspace(to);
     }
     return ok;
--- a/src/libOpenImageIO/color_ocio.cpp
+++ b/src/libOpenImageIO/color_ocio.cpp
@@ -201,8 +201,8 @@ struct CSInfo {
 // Hidden implementation of ColorConfig
 class ColorConfig::Impl {
 public:
-    OCIO::ConstConfigRcPtr config_;
-    OCIO::ConstConfigRcPtr builtinconfig_;
+    OCIO::ConfigRcPtr config_;
+    OCIO::ConfigRcPtr builtinconfig_;
 
 private:
     std::vector<CSInfo> colorspaces;
@@ -815,6 +815,49 @@ ColorConfig::~ColorConfig() {}
 
 
 
+// OIIO doctoring of OCIO configs for different default file rules. Currently,
+// we only do this for built-in configs.
+static void
+fix_config_file_rules(OCIO::ConfigRcPtr& config)
+{
+    OIIO_CONTRACT_ASSERT(config);
+    DBG("Fixing up rules:\n");
+#if 1
+    // Just start with a clean slate
+    auto rules = OCIO::FileRules::Create();
+#else
+    // Alternate universe: Start with the existing rules
+    auto rules = config->getFileRules()->createEditableCopy();
+#endif
+    for (size_t i = 0, e = rules->getNumEntries(); i != e; ++i) {
+        DBG("  rule {}/{}: pat='{}' ext='{}' -> \"{}\"\n", i, rules->getName(i),
+            rules->getRegex(i), rules->getExtension(i),
+            rules->getColorSpace(i));
+        if (Strutil::iequals(rules->getExtension(i), "exr")) {
+            // Change the rule for exr extension, if it exists, to "unknown".
+            // Make no assumptions. OCIO's built-in configs think it should be
+            // ACES2065-1, which is almost never right.
+            rules->setColorSpace(i, "unknown");
+            DBG("    changed cs to \"{}\"\n", rules->getColorSpace(i));
+        } else if (!strcmp(rules->getName(i), "Default")) {
+            // Default rule or one that matches everything -- for OIIO, we
+            // just want to change this to unknown. We made decisions about
+            // default per-file-format color space decisions in the individual
+            // readers. We don't even consider file extension to be reliable
+            // evidence of the file type.
+            rules->setColorSpace(i, "unknown");
+            DBG("    changed cs to \"{}\"\n", rules->getColorSpace(i));
+        }
+    }
+
+    // But make the path search rule (look for the right-most color space name
+    // embedded in the path) have precedence over file naming rules.
+    rules->insertPathSearchRule(0);
+    config->setFileRules(rules);
+}
+
+
+
 bool
 ColorConfig::Impl::init(string_view filename)
 {
@@ -825,7 +868,10 @@ ColorConfig::Impl::init(string_view filename)
     OCIO::SetLoggingLevel(OCIO::LOGGING_LEVEL_NONE);
 
     try {
-        builtinconfig_ = OCIO::Config::CreateFromFile("ocio://default");
+        auto cfg = OCIO::Config::CreateFromFile("ocio://default");
+        OIIO_CONTRACT_ASSERT(cfg);
+        builtinconfig_ = cfg->createEditableCopy();
+        fix_config_file_rules(builtinconfig_);
     } catch (OCIO::Exception& e) {
         error("Error making OCIO built-in config: {}", e.what());
     }
@@ -841,10 +887,13 @@ ColorConfig::Impl::init(string_view filename)
     } else {
         // Either filename passed, or taken from $OCIO, and it seems to exist
         try {
-            config_ = OCIO::Config::CreateFromFile(
-                std::string(filename).c_str());
             configname(filename);
-            m_config_is_built_in = Strutil::istarts_with(filename, "ocio://");
+            auto cfg = OCIO::Config::CreateFromFile(
+                std::string(filename).c_str());
+            if (cfg)
+                config_ = cfg->createEditableCopy();
+            if (config_ && Strutil::istarts_with(filename, "ocio://"))
+                fix_config_file_rules(config_);
         } catch (OCIO::Exception& e) {
             error("Error reading OCIO config \"{}\": {}", filename, e.what());
         } catch (...) {
@@ -2344,27 +2393,41 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
         from = src.spec().get_string_attribute("oiio:Colorspace",
                                                "scene_linear");
     }
-    if (from.empty() || to.empty()) {
-        dst.errorfmt("Unknown color space name");
+    if (from.empty() || from == "unknown" || to.empty() || to == "unknown") {
+        dst.errorfmt("Unknown color space name (from=\"{}\", to=\"{}\")", from,
+                     to);
         return false;
     }
-    ColorProcessorHandle processor;
-    {
-        if (!colorconfig)
-            colorconfig = &ColorConfig::default_colorconfig();
-        processor
-            = colorconfig->createColorProcessor(colorconfig->resolve(from),
-                                                colorconfig->resolve(to),
-                                                context_key, context_value);
-        if (!processor) {
-            if (colorconfig->has_error())
-                dst.errorfmt("{}", colorconfig->geterror());
-            else
-                dst.errorfmt(
-                    "Could not construct the color transform {} -> {} (unknown error)",
-                    from, to);
-            return false;
-        }
+
+    if (!colorconfig)
+        colorconfig = &ColorConfig::default_colorconfig();
+
+    if (colorconfig->isData(from) || colorconfig->equivalent(from, "raw")) {
+        // If the input color space is not color managed, the transformation
+        // is meaningless, just set 'to' to the same space so that the whole
+        // thing is a no-op, including continuing to think it's in the raw
+        // space.
+        to = from;
+    } else if (colorconfig->isData(to) || colorconfig->equivalent(to, "raw")) {
+        // If the output color space is not color managed, the transformation
+        // is meaningless, just make it look like it's already in the 'from'
+        // space. Note that this DOES change the apparent color space label
+        // to the requested raw output space.
+        from = to;
+    }
+
+    ColorProcessorHandle processor
+        = colorconfig->createColorProcessor(colorconfig->resolve(from),
+                                            colorconfig->resolve(to),
+                                            context_key, context_value);
+    if (!processor) {
+        if (colorconfig->has_error())
+            dst.errorfmt("{}", colorconfig->geterror());
+        else
+            dst.errorfmt(
+                "Could not construct the color transform {} -> {} (unknown error)",
+                from, to);
+        return false;
     }
 
     logtime.stop(-1);  // transition to other colorconvert
--- a/src/oiiotool/oiiotool.cpp
+++ b/src/oiiotool/oiiotool.cpp
@@ -5566,6 +5566,8 @@ input_file(Oiiotool& ot, cspan<const char*> argv)
             // Try to deduce the color space it's in
             std::string colorspace(
                 ot.colorconfig().getColorSpaceFromFilepath(filename, "", true));
+            if (colorspace == "unknown")
+                colorspace.clear();
             if (colorspace.size() && ot.debug)
                 OIIO::print("  From {}, we deduce color space \"{}\"\n",
                             filename, colorspace);
--- a/testsuite/oiiotool-color/ref/out-ocio2.3.txt
+++ b/testsuite/oiiotool-color/ref/out-ocio2.3.txt
@@ -4,11 +4,11 @@ Full command line was:
 nope: 0.5,0.5,0.5 half
 raw: 0.5,0.5,0.5 half
 acescg: 0.5,0.5,0.5 half
-srgb_tx: 0.213989,0.213989,0.213989 half
+srgb_tx: 0.5,0.5,0.5 half
 nope: 0.5,0.5,0.5 half
 raw: 0.5,0.5,0.5 half
 acescg: 0.5,0.5,0.5 half
-srgb_tx: 0.735352,0.735352,0.735352 half
+srgb_tx: 0.5,0.5,0.5 half
 Comparing "colormap-inferno.tif" and "ref/colormap-inferno.tif"
 PASS
 Comparing "colormap-custom.tif" and "ref/colormap-custom.tif"
--- a/testsuite/oiiotool-color/run.py
+++ b/testsuite/oiiotool-color/run.py
@@ -7,6 +7,9 @@
 
 import os
 
+redirect = " >> out.txt 2>&1 "
+
+
 # print("ociover =", ociover)
 
 # Make test pattern with increasing intensity left to right, decreasing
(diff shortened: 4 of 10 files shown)


src/libOpenImageIO/color_ocio.cpp
```diff
@@ -815,6 +815,49 @@ ColorConfig::~ColorConfig() {}
...
+#endif
+    for (size_t i = 0, e = rules->getNumEntries(); i != e; ++i) {
+        DBG("  rule {}/{}: pat='{}' ext='{}' -> \"{}\"\n", i, rules->getName(i),
+            rules->getRegex(i), rules->getExtension(i),
+            rules->getColorSpace(i));
+        if (Strutil::iequals(rules->getExtension(i), "exr")) {
+            // Change the rule for exr extension, if it exists, to "unknown".
+            // Make no assumptions. OCIO's built-in configs think it should be
+            // ACES2065-1, which is almost never right.
+            rules->setColorSpace(i, "unknown");
```
zachlewis, line 842: If you want to use the same color space for both the "exr" rule and the default rule, why not just remove the "exr" rule entirely?

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -815,6 +815,49 @@ ColorConfig::~ColorConfig() {}
...
+            DBG("    changed cs to \"{}\"\n", rules->getColorSpace(i));
+        } else if (!strcmp(rules->getName(i), "Default")) {
+            // Default rule or one that matches everything -- for OIIO, we
+            // just want to change this to unknown. We made decisions about
+            // default per-file-format color space decisions in the individual
+            // readers. We don't even consider file extension to be reliable
+            // evidence of the file type.
```
zachlewis, line 851: Interesting. In my mind, OCIO FileRules are a better place to store OIIO's format-specific colorspace defaults than scattered across the code base, because it keeps the coupling of extension and color space name in sync and unambiguous and easily to update in one place. But I guess it's kind of moot if the defaults we're using for non-EXR containers are always either "srgb_rec709_scene" or, very occasionally, "ocio:itu709_rec709_scene". 

Anyway, whether the container defaults live in individual readers / writers or in an OIIO default config is entirely up to you -- as long as the strings you're hardcoding are known CIF or OCIO colorInteropIDs, everything should be gravy.

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2344,27 +2393,41 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
...
-                dst.errorfmt(
-                    "Could not construct the color transform {} -> {} (unknown error)",
-                    from, to);
-            return false;
-        }
+
+    if (!colorconfig)
+        colorconfig = &ColorConfig::default_colorconfig();
+
+    if (colorconfig->isData(from) || colorconfig->equivalent(from, "raw")) {
```
zachlewis: I think this is doing too much, and possibly not doing what you want.

- If colorconfig->isData(from) == True, then any IBA that references that color space will always (passively) elicit a no-op processor automatically, so there really isn't any reason to do anything else. Manually setting the input and output color space values to the same string is confusing and difficult to debug, because it's not immediately clear where or why that would happen; whereas if we can clearly see that either a source or destination color space argument is a data color space, then we know where and why a no-op is produced, and where specifically to apply a fix, if that's not the desired behavior. This helps us clearly pinpoint exactly where and why unexpected no-ops are elicited. 

- If colorconfig->isData(from) == False, but colorconfig->equivalent(from, "raw") == True, then it means a "raw" color space exists in the config, but it is NOT a data color space; and the subsequent assumptions for equivalent (and thus non-data) color spaces would be incorrect. 

(In any case, it's far more robust to compare against "data" instead of "raw", since "data" is a canonical (and required!) role to a color space that elicits a no-op processor. The OCIO API provides a str enum for ROLE_DATA; whereas "raw" is just some string that happens to be the name of a data color space in the builtin OCIO configs, as they exist today). 


Upshot is... I think lines 2405-2417 are unnecessary and kind of serve to obscure what's actually happening. If either "to" or "from" is a data color space, OCIO will elicit a no-op processor, and that's all there is to it.

HOWEVER... something like this would be helpful _after_ a no-op Processor is successfully applied, just before line 2437, when determining what to update the ImageSpec `oiio:ColorSpace` attribute to -- to me, it makes sense to persist whatever the "from" value resolves to. If the "from" color space was a data color space, then no matter what the "to" color space was, nothing happened, and we may as well continue tagging "oiio:ColorSpace" as the data color space. Else, if the "to" color space was a data color space, then -- once again -- nothing happened, and "oiio:ColorSpace" may as well reflect that the output color space encoding is effectively the same as the input. 

This is slightly different from the logic you have below... I hope what I'm saying makes sense. By always persisting the "from" value if either "to" or "from" is a data color space, we're persisting the best information we have about the image state, and causes subsequent IBAs to behave predictably and consistently. If we're converting _to_ a data space, and thus the processor is a NoOp, we don't want to tag the output color space as that data space, because downstream IBAs would lose the ability to meaningfully infer valid transforms from `oiio:ColorSpace`.

src/libOpenImageIO/color_ocio.cpp
zachlewis, line 2430:
```suggestion
   if (colorconfig->isData(to) || colorconfig->isData(from)) {
            dst.specmod().set_colorspace(from);
        } else {
            dst.specmod().set_colorspace(to);
        }
```

src/libOpenImageIO/color_ocio.cpp
```diff
@@ -2344,27 +2393,41 @@ ImageBufAlgo::colorconvert(ImageBuf& dst, const ImageBuf& src, string_view from,
...
+        // If the input color space is not color managed, the transformation
+        // is meaningless, just set 'to' to the same space so that the whole
+        // thing is a no-op, including continuing to think it's in the raw
+        // space.
+        to = from;
+    } else if (colorconfig->isData(to) || colorconfig->equivalent(to, "raw")) {
```
zachlewis:
```suggestion
    } else if (colorconfig->isData(to)) {
```

src/oiiotool/oiiotool.cpp
```diff
@@ -5566,6 +5566,8 @@ input_file(Oiiotool& ot, cspan<const char*> argv)
             // Try to deduce the color space it's in
             std::string colorspace(
                 ot.colorconfig().getColorSpaceFromFilepath(filename, "", true));
+            if (colorspace == "unknown")
+                colorspace.clear();
```
zachlewis, line 5570: tl;dr --  I think we should check if the config recognizes "unknown" as a valid color space; if not, we should set `colorspace` to an empty string if strict_parsing is enabled in the config; otherwise, use the default role. 

We need to make sure we're not interfering with config authors who want to define custom behavior for handling cases where `colorInteropID` == "unknown". For example, a config author might create a do-nothing color space called "unknown" in order to capture and route images so tagged through a special "error" default view, so that when the view is enabled in image browsers / thumbnail previewers (or oiitool's default behavior when empty strings are provided as arguments for `--ociodisplay`), it's extremely easy to distinguish media explicitly tagged as "unknown" from untagged media and color managed media.

We need clarity from CIF about this -- it isn't clear how hosts should handle cases where `colorInteropID` == "unknown", and "unknown" is a valid color space in the config. I think the safest way to convert `colorInteropID` strings --> OCIO color space names is for hosts to interpret "unknown" as a signal to try the config's FileRules with the file path, i that information is available, and fail over to the "unknown" color space, instead of the "default" color space. 

There's a subtle but important implication for OIIO here: `colorInteropID` and `oiio:ColorSpace` metadata communicate similar but distinct information -- `colorInteropID` is _prescriptive_ of behavior ("use the CIF guidelines to translate 'srgb_rec709_scene' into something your color management framework can reason about), while `oiio:ColorSpace` is _descriptive_ of internal state of the ImageBuf, as identified and understood by OIIO. In practice, this means, 

 - If both `colorInteropID` and `oiio:ColorSpace` are unset:  OIIO should go ahead and try to divine meaning from whatever other information is available (other color metadata, embedded ICC profiles, file paths...); if all else fails, return an empty string if strict_parsing is enabled in the config; otherwise, fail over to the default color space.
 - if  `colorInteropID` == "unknown" and `oiio:ColorSpace` is unset: same as above, but fail over to the "unknown" color space if it happens to exist before returning either the default color space or an empty string.
 - If `oiio:ColorSpace` == "unknown": OIIO has already tried and failed to resolve the color space, and has failed over to the "exit state" color space as intended by config authors who want to elicit specific behavior for failing to identify explicitly "unknown" `colorInteropID` metadata.

zachlewis (changes requested):
I've been giving this PR *a lot* of thought over the past few days, and... I'm not thrilled about this. I think this conflates the notions of `colorInteropID` metadata, OCIO color space names, and the internal state of OIIO's color space resolution heuristics.

I think I understand your intentions, but I'd like to dig into this a bit more with you to see if we can come up with a more robust solution. I need to continue thinking about this.

src/libOpenImageIO/color_ocio.cpp
lgritz: Well, considering that I also clear the rules entirely, I think you could classify this whole clause as a memo to myself about HOW to change it, rather than being a consequential change at the moment. At the moment, this non-functional code marks both "exr" and "everything" as unknown, but I could imagine (and did experiment with) having them be different, such as assuming that exr files were scene_linear. And we may go back to that, I don't know.

src/libOpenImageIO/color_ocio.cpp
lgritz: 2nd part first: Yes, the intent is that all hardcoded color space names that OIIO assigns will be CIF tokens.

1st part: OIIO doesn't really consider file names, even extensions, to be authoritative when reading. It's at most an initial guess. Some formats have *many* extensions that are sanctioned, some facilities or apps have custom extensions (like how many places use .tx for TIFF textures that are properly tiled and MIPmapped, though the TIFF standard never mentions this extension), and sometimes files are just plain mislabeled.

src/libOpenImageIO/color_ocio.cpp
lgritz: Can I trust that isData() will be correct for all configs, and in all versions of OCIO that we support? There will never be a "raw" space that isn't known to be data? Or can there be? What would that mean?

src/libOpenImageIO/color_ocio.cpp
lgritz: I think I see what you are saying here. But I'm not sure you're right about the whole thing. Surely
```
    transform("data", "something else")
```
should mark the output as `from` ("data") because it hasn't really been transformed, and certainly at this point is not color managed. But I think I was feeling that
```
    transform("something else", "data")
```
should consider the output to be data. It's not changing the pixel values, but it is trying to label it as no longer color managed, and we should respect that.

So I can agree with removing that earlier pre-transform checking of data/etc, but I think this post-setting of the color space should be only:

    if (olorconfig->isData(from)) {
        dst.specmod().set_colorspace(from);
    } else {
        dst.specmod().set_colorspace(to);
    }

Does that seem reasonable?

src/oiiotool/oiiotool.cpp
lgritz: There are a couple interpretations possible, so let's get more explicit.

**"colorInteropID"**

This is either (a) what was actually found in the file as metadata with this name, or possibly (b) set by OIIO when if we know for sure what the color space is AND it corresponds to a known CIF token.

Distinguishing these depends on our answer to the following: For a file format like DPX that is unaware of CIF and does not have metadata called "colorInteropID", but nonetheless has differently named fields, which may at least sometimes specify a CIF-known colorspace, should we set this metadata in the ImageSpec (when it corresponds to a known token) or not?

**"oiio:ColorSpace"**

This continues to be an internal-only (i.e. *never written to files as metadata with this name*) OIIO-specific hint about what we think the pixels are. This can be a CIF token, but I'm not sure it needs to be restricted to that.

When I set "oiio:ColorSpace" to "unknown", that's for OIIO to communicate with itself.

Though I see what you mean about a case where a config has a known color space called "unknown". I'm just not sure if this is a real situation that we should care about, just like there could be a color space called "lin_rec709_scene" that is really something else, or a trick to handle that color space in a special way.

src/libOpenImageIO/color_ocio.cpp
zachlewis: > Can I trust that isData() will be correct for all configs, and in all versions of OCIO that we support? 

Yep!

> There will never be a "raw" space that isn't known to be data? Or can there be? What would that mean?

"Raw" just happens to be a string that matches the name of a data color space historically found in ACES configs. It doesn't have any fixed meaning, and there's no guarantee that a given config will have a "raw" space, or that color spaces named "raw" are necessarily data spaces. 

(Many folks, myself included, are confounded by the ACES configs' use of the word "raw", as the term is overloaded in this domain; and would seem to imply that OCIO can do something meaningful with "camera raw" data, when, in fact, it's quite the opposite)

Checking a color space's "isData" attribute is by far the most reliable way to find out if the color space's "isData" attribute is set :)

lgritz:
Updated based on some of your suggestions.

src/libOpenImageIO/color_ocio.cpp
zachlewis: Hmmm. Interesting. I'm not sure that I agree that applying a conversion to a data space is necessarily the same thing as declaring an intent to treat the data as explicitly unmanaged... but now I'm not sure that I disagree, either. In my mind, I was assuming folks would use --iscolorspace or otherwise adjust oiio:ColorSpace with the API to explicitly communicate that intent downstream, as I normally think of applying conversions to or from data colorspaces as side-effect-free noops... 

but you're right, it's perfectly reasonable to assume conversions to a data space suffice as a declaration of intent. You've convinced me!

src/libOpenImageIO/color_ocio.cpp
zachlewis: Whoops, I hadn't realized that this isn't even implemented..

> I think you could classify this whole clause as a memo to myself about HOW to change it, rather than being a consequential change at the moment. 

I appreciate that, it's helpful to see how you're thinking things through. I hope you don't mind my long-assed commentary. And I apologize if I came across as implying you were doing the wrong thing here by setting the EXR and default rules to the same thing -- I was just curious to see if you were thinking about treating responses triggered by the EXR rule differently from the default rule, and whether you had considered using the API to  I think that would provide a way for you to do what I think you're trying to do, without having to patch a sentinel string value into the file rules. 

Indeed, this is a tricky situation. At first, I was going to emphatically suggest switching to using the `scene_linear` role, but I've come around to fully grasping your concerns with the default config causing `oiioltool foo.exr -o bar.exr` to persist unverified, potentially incorrect metadata. I definitely think that it's not a good idea to infer otherwise nonexistent `colorInteropID` metadata automatically from OIIO's default ColorConfig's default file rule unless the user is actually invoking some aspect of OCIO / OIIO's color management. 

At the same time, I'm uneasy about using empty strings to intentionally elicit error states, if we can possibly avoid doing so. Something I try to keep in mind when I'm thinking about hacking around OCIO's designed behavior is, how are other OCIO hosts in the pipeline likely to behave under similar circumstances? Do we risk surprising users or stepping on config authors who expect all OCIO hosts to behave more or less the same way? Do the benefits of doing stuff differently outweigh the tech-debt costs working around the way OCIO "wants to" behave? Am I burdening others by injecting additional complexity into an already-complex system? Usually, I come around, one way or another, to aligning what I'm trying to do to what OCIO already does, to whatever degree possible. 

So, if we consider OCIO v2 FileRules behavior to be the lowest common denominator, what do we gain by allowing filepath parsers to return empty strings, given that OCIO has deprecated such v1-style behavior?

That's not to say we _shouldn't_ allow empty-string responses -- I just mean, if we decide to go that route, we  should have pretty good reasons for doing so; and we should work to contain the blast radius of what we break and how we break it.


One last thing: if we do want to emit empty strings, we should strongly consider using the config's strict_parsing attribute to allow config authors opt in to potentially breaking behavior.

lgritz: I find your long-assed commentary extremely valuable, thank you for it. I really need this to help me think it through.

I'm really thinking about this as two separate problems:

1. Short term and back-portable to 3.1: I need to immediately fix the really awful issues of (a) autocc with default configs unexpectedly considering exr to be ap0, and (b) raw/unmanaged color spaces of inputs getting laundered through IBA::colorconvert to look like a transform happened and they are in the destination space, when they are actually not. 

2. Long-term, hopefully ironed out to be a solid 3.2 feature: a more holistic, consistent approach to figuring out what's managed, unmanaged, and unknown, how each should be handled, and adhering to CIF as much as possible.

This PR is only meant to address part 1, be immediately deployable, and not make anything worse than it currently is. It doesn't need to fix all possible problems, and anything that did would probably be too much of a behavior change to backport to 3.1 anyway.

lgritz: > One last thing: if we do want to emit empty strings, we should strongly consider using the config's strict_parsing attribute to allow config authors opt in to potentially breaking behavior.

I have to admit that I'm not really sure what that does or what else might break or change if I do it.

I'm still thinking that every suggestion needs to be considered in terms of whether it will "change/fix things that users will clearly consider broken now if they notice it" or "change the way it *ought* to be done in the future, when we can alter behavior as much as we want at a version boundary."

I guess I also have divided loyalties. Maybe 10% of the time I'm thinking "what does the color scientist / config author want oiio to do", but the other 90% of the time I'm primarily putting myself in the shoes of the end user who is directly using oiiotool in battle and wanting consistent and intuitive behavior regardless of what kooky config they might be stuck with.

lgritz:
Are my latest changes closer? What else do you think is important to change for *this* PR that I intend to backport to 3.1 -- in which I want to minimally change behavior, ideally only fixing things that are very broken, and leaving everything else in place that people have somehow accepted up until now? Then we can also make further changes bound for 3.2 that take bigger steps toward where we eventually want to be.

src/oiiotool/oiiotool.cpp
zachlewis: > Though I see what you mean about a case where a config has a known color space called "unknown". I'm just not sure if this is a real situation that we should care about, just like there could be a color space called "lin_rec709_scene" that is really something else, or a trick to handle that color space in a special way.

I 100% agree that it would _vastly_ simplify things if all around if we weren't concerned with color spaces named "unknown" -- pretty much all of my concerns pertain to the implications of writing or not writing "unknown" colorInteropID metadata for pipelines that use OCIO to imbue special handling of "unknown" metadata. Unfortunately, unless the Color Interop Forum guides against it, or OpenColorIO disallows it, I don't think it's safe to assume config authors _aren't_ relying on color spaces named "unknown" to steer "explicit unknown" colorInteropID behavior. Given the unique behavior of the "unknown" CIID compared to the others, I don't think it's a stretch to presume that OCIO config authors might complement the "unknown" CIID's unique  behavior with similarly uniquely behaving "unknown" OCIO color spaces. For better or worse, the current CIF guidelines coupled with OCIO's architecture makes it possible to orchestrate pretty complex (and fancy!) behaviors pertaining to reading and writing explicitly "unknown" `colorInteropID` metadata. 

Is this a real situation we should care about? I dunno. I'd argue, per [Hyrum's Law](https://www.hyrumslaw.com/) that the situation is real enough to warrant considerable thought and caution at this stage, to make sure we aren't building around assumptions that prove difficult to back out of should they prove false. On the other hand, I'm not sure if I'm trying to protect for workflows that the Color Interop Forum is consciously trying to permit. I don't think too many other people have given this as much thought as I have; and, as you know by now, I am batshit insane, so I could definitely be fretting about nothing. 

I would love for OIIO to support such workflows almost as much as I would love CIF / OCIO to explicitly disallow naming color spaces "unknown", so OIIO wouldn't have to worry about supporting such workflows.

zachlewis:
> Are my latest changes closer? What else do you think is important to change for _this_ PR that I intend to backport to 3.1 -- in which I want to minimally change behavior, ideally only fixing things that are very broken, and leaving everything else in place that people have somehow accepted up until now? Then we can also make further changes bound for 3.2 that take bigger steps toward where we eventually want to be.

I definitely think so. Having voiced my concerns and discussed stuff with you, I'm pretty comfortable with where things are at now. However, I'd would like to see how tomorrow's CIF meeting shakes out before I approve this PR, on the off-chance we have a moment to discuss the "unknown" OCIO Color Space thing; although I imagine we won't really have time to get into it with the other stuff on the agenda. I will see if other folks think it's worth adjusting the wording in the EXR guidelines to clarify whether or not EXR readers should attempt to concern themselves with OCIO color spaces named "unknown".

(Either way, if I were to ask you to change anything, I would probably ask that you use "oiio:unknown" instead of "unknown" for the purposes of OIIO communicating with itself, at least for the time being)

src/oiiotool/oiiotool.cpp
lgritz: Well, there is a sense in which my use of "unknown" in this PR is arbitrary. It's a signal from OIIO to other parts of OIIO, really. It will never get saved to a file, because... the ONLY color space "name" that's saved to a file is colorInteropID to exr files, and we only do that if it's a recognized CIF token.

So we could pick a different name, something that would never be found in a config. Would that help you sleep better at night? Are there any characters that have always been disallowed in OCIO color space names?

lgritz:
> (Either way, if I were to ask you to change anything, I would probably ask that you use "oiio:unknown" instead of "unknown" for the purposes of OIIO communicating with itself, at least for the time being)

Ah, I hadn't read this yet when I suggested using a special name, maybe even containing a character that is not allowed (or that no reasonable person would use) in an OCIO color space name. Clearly we are operating on the same wavelength!

lgritz:
It's a work holiday in Canada tomorrow, but I'll try to drop in on the CIF meeting if you think that would be helpful. I'd love for CIF to have a way to officially represent "unknown" as separate from "data" or "explicitly not color managed."

lgritz:
So what is the verdict, @zachlewis, now that the meeting has happened?

Should we declare that a color space called "undefined" is so unlikely/unwise that we shouldn't care? That seemed to be the vibe I got from the meeting -- that CIF already thinks "unknown" means "will be handled by the app in some extra-judicial manner" and any configs with that name for a color space are going to collide badly with CIF conventions anyway and we are the least of their problems.

Or should we pick some other word, add a special character, or a way of mangling it that is less likely to interfere? Like calling it "`<unknown>`" or "`#UNKNOWN`" or something else silly? (Please, not "oiio:unknown", that breaks my heart since we're already calling the metadata "oiio:ColorSpace" and it oiio:feels oiio:ridiculous oiio:to oiio:do oiio:both.

Anything else we should revise here at this time?

doug-walker:
@lgritz, I'll admit that, after reading the above thread, I don't fully understand why the ID "unknown" is causing such a problem, but I do want to chime in with two observations:
1. It is perfectly legit for an OCIO config to contain a color space named "unknown". This is a valid way for a config author to provide a color space to be used to handle those scenarios. In addition, it provides a way for artists to ask an application to use that ID when writing files where the actual color space is either unknown to the artist or not present in the config.
2. That said, config authors must acknowledge the possibility that an application (e.g., oiiotool) could intercept the ID "unknown" and handle it some other way (e.g. using OCIO File Rules) rather than simply assigning such a color space, if it exists.

lgritz:
@doug-walker If a config has a color space named "unknown", and it really means "unknown", then I think we aren't causing trouble by using it in OIIO (that is, consider it not color managed and in some contexts issue errors when using it as input to functions that require color managed pixels). I'm not convinced that there is a realistic situation where a config calls something "unknown" and expects the surrounding app/library to treat it like it's color managed and necessarily pass it back to OCIO. I think?

Maybe Zach and I got into a momentary worry that somebody has a color space called "unknown" and yet it would mess things up for OIIO to handle that specially without passing it along to the OCIO transform. But now somehow I'm less worried about it again.

zachlewis:
I provided a more detailed response re: my thoughts on "unknown" colorInteropID metadata here: https://github.com/AcademySoftwareFoundation/ColorInterop/pull/13#issuecomment-4491206770

But as for this PR... LGTM! This does what you need it to do for now, and it shouldn't confuse, contradict, or interfere with efforts to align with the CIF guidelines. Feel free to merge at your leisure (suit), Larry. 
(sorry)


> I'm not convinced that there is a realistic situation where a config calls something "unknown" and expects the surrounding app/library to treat it like it's color managed and necessarily pass it back to OCIO. I think?

Maybe. Ideally, I would like for OCIO-enabled applications to behave as consistently here as they do in other areas; but at minimum, I expect OCIO-enabled applications using the same config not to interfere with each other, to whatever degree that's possible.

> Or should we pick some other word, add a special character, or a way of mangling it that is less likely to interfere? Like calling it "<unknown>" or "#UNKNOWN" or something else silly? (Please, not "oiio:unknown", that breaks my heart since we're already calling the metadata "oiio:ColorSpace" and it oiio:feels oiio:ridiculous oiio:to oiio:do oiio:both.

Heh -- honestly, go for whatever you'd like. Personally, I like <using> <these>, but it's your call. I was advising using anything other than "unknown" because I didn't anyone to get in the habit of thinking of `oiio:ColorSpace` == "unknown" and `colorInteropID` == "unknown" mean the same thing unless I was sure it would be absolutely safe to always make that assumption for all configs, across all color-y parts of the code base; because that assumption does more or less hold for other possible `colorInteropID` values.

lgritz:
I think you meant to say you prefer `"<unknown>"` but it rendered above as `""` because, html. 😄 

> I didn't anyone to get in the habit of thinking of oiio:ColorSpace == "unknown" and colorInteropID == "unknown" mean the same thing unless

Yeah, yeah, but it sure felt like at the CIF meeting, there was agreement that CIF/OCIO didn't want to be overly prescriptive and was acknowledging that different apps will have different approaches to dealing with unknown interop ID. For example, an interactive app might pop up a dialog and as the user for guidance, whereas a scripting API does not have that option and has do something else (even if it's just terminating). I'm not sure what we're doing in this PR is at all outside that assumption of a range of responses.

Issue #4781, opened 2025-05-24
[FEATURE REQUEST] Add [[nodiscard]] to image read/write methods


EmilDohne:
This is not so much a feature request but more of an open discussion on whether all of the `read_` and `write_` methods should have a `[[nodiscard]]` attribute added (since C++17) to avoid users not catching whether the read/write was successful or not (caught myself doing this).

This would obviously break ABI and backwards compatibility so I'm happy to hear opinions!

lgritz:
I like the idea in general, but you are right, if not applied carefully, it will effectively break downstream projects in an annoying way.

I don't feel like we have to wait for 4.0 like we would with truly breaking API changes. Sometimes with no help from us, the compilers or the C++ version change and that makes warnings or deprecations that didn't used to be there. This feels more like that situation. But we still need to stage it carefully.

Maybe the right way to do it is something like this:

We already have OIIO_NODISCARD defined in platform.h. But I think what we need is a secondary definition like this:


    #if OIIO_INTERNAL || OIIO_WARN_FOR_NODISCARD
        #define OIIO_API_NODISCARD OIIO_NODISCARD
    #else
        #define OIIO_API_NODISCARD
    #endif

    ...
    OIIO_API_NODISCARD bool public_function();

and so we would get `[[nodiscard]]` warnings for our own codebase, but not for downstream users, unless they do a

    #define OIIO_WARN_FOR_NODISCARD 1

before including OIIO headers, or `-DOIIO_WARN_NODISCARD=1` in their build system. This lets people easily find out if they are ignoring return values, but without breaking the builds of people not trying to do so. Then in a future release (3.2 maybe? or later?) change to

    #if OIIO_INTERNAL && !OIIO_WARN_FOR_NODISCARD

which flips the sense to "warn by default, but give a way to easily turn off just that warning by `#define OIIO_WARN_FOR_NODISCARD 0` before including OIIO headers.

For *NEW* API calls that didn't exist before (so can't break existing code), we can add use OIIO_NODISCARD (or `[[nodiscard]]`) straightaway if we feel strongly it's a result that shouldn't be ignored.

EmilDohne:
Sorry for only getting back so late to this, had some other things to take care of. I think thats an excellent solution and I'll look to create a PR for this sometime next week (adding it to imageinput at first to slowly phase it in like you're doing with image_span)

lgritz:
I think there are judgment calls to make as well. 

[...]

PR #5196, opened 2026-05-14
api: apply OIIO_NODISCARD to color.h

luna-y-kim:
### Description

Adds `OIIO_NODISCARD` and `OIIO_NODISCARD_ERROR` to `color.h`.
Deprecated functions are left unannotated.

Part of #4781 

Assisted-by: Claude / Opus 4.7

### Tests

No new tests, only attributes are added.

### Checklist:

- [ ] (N/A) **I have updated the documentation** if my PR adds features or changes
  behavior.
  on the GitHub runners.)
- [ ] (N/A) If I added or modified a public C++ API call, I have also amended the

### DIFF
--- a/src/include/OpenImageIO/color.h
+++ b/src/include/OpenImageIO/color.h
@@ -41,8 +41,8 @@ class OIIO_API ColorProcessor {
 public:
     ColorProcessor() {};
     virtual ~ColorProcessor(void) {};
-    virtual bool isNoOp() const { return false; }
-    virtual bool hasChannelCrosstalk() const { return false; }
+    OIIO_NODISCARD virtual bool isNoOp() const { return false; }
+    OIIO_NODISCARD virtual bool hasChannelCrosstalk() const { return false; }
 
     // Convert an array/image of color values. The strides are the distance,
     // in bytes, between subsequent color channels, pixels, and scanlines.
@@ -93,11 +93,11 @@ public:
     ///
     /// Multiple calls to this are potentially expensive. A ColorConfig
     /// should usually be shared by an app for its entire runtime.
-    bool reset(string_view filename = "");
+    OIIO_NODISCARD_ERROR bool reset(string_view filename = "");
 
     /// Has an error string occurred?
     /// (This will not affect the error state.)
-    bool has_error() const;
+    OIIO_NODISCARD bool has_error() const;
 
     /// DEPRECATED(2.4), old name for has_error().
     OIIO_DEPRECATED("Use has_error()")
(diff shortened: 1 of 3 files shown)


lgritz:
Hi, @luna-y-kim ! The changes in color.h look good to me, but (assuming that you built it and tested it on your side before submitting) I think you got bit in CI by the fact that not all compilers (or versions thereof) actually catch using deprecated functions and issue the error that would stop the build. But others do! So what's happening in the failed tests is that it's finding all the places *in our own codebase* where we inadvertently failed to check the return results of a function. You'll need to track down those spots and do some appropriate error handling (or, if it really is a case where the result doesn't matter, dispose of it sanely).

lgritz:
It looks like gcc is not catching the errors, but all the CI jobs that use clang do find them. So if you're using gcc on your own machine, that might be why you didn't see any problems when you did a build.

If you're up for addressing this as well, it seems that we're actually disabling this in src/cmake/compiler.cmake, line 188 (`-Wno-unused-result`). I think eliminating this line would make the discarded result warnings do the right thing for gcc... but I'm not sure what else it might break! There must be some reason why we put it there originally, but I don't remember why, and maybe it's something that isn't even relevant any more. 

If you're up for also solving that mystery...

luna-y-kim:
Hi, thanks for the explanation. I found all 11 failures are from the same spot `ColorConfig::ColorConfig` in `color_ocio.cpp:810`. As a first try, I've added a cast (void) because a constructor can't return the result. However, I see other spots also have the same problem. I'll try to track down those spots and do error handling. 

Once I've handled all the discarded-result call sites, I might look at removing `-Wno-unused-result` from `compiler.cmake:188` to see if it's still needed or if it's safe to remove.

lgritz:
Sounds good. I really don't know what will happen when you remove the -Wno-unused-result. Will be interesting to see if there's still a relevant reason why that is there. It appears to have been there for many, many years, and that code has been through enough some movement and refactoring, so I'm not exactly sure when it was first introduced or why.

lgritz:
If you are wondering whether to finish the work tonight, or meet the rest of us at the pub, definitely choose the pub! Fixing the last couple errors in the next couple days is fine.

lgritz:
The failure for the "VFX2023" test is unrelated to your PR and is being fixed separately. If everything else is ok, we will not hold up the merge just because of that.

luna-y-kim:
Sounds good to me. I'll explore the `-Wno-unused-result`​ in `compiler.cmake:188​` separately from this PR so this one can merge.

lgritz (approved):
LGTM! Thanks for the patch, this looks great.

PR #5200, opened 2026-05-15
ci: Fix broken CI by locking down aswf container for 2023

lgritz:
Our CI job that builds against VFX Platform 2023 via docker container aswf/ci-osl:2023-clang15 broke after some dependency changes in a recent aswf docker container update. It's not worth trying to "fix" an old year, just lock down to the previous version of the container that is known to work.


jfpanisset:
LGTM, I'll spend some time over the weekend to figure out exactly why yaml-cpp isn't building with the new image.

Would be interesting to see if this could move to ci-oiio:2023

lgritz:
The ci-oiio container also seems broken for me. So I'm going to merge what is in this PR, which works an is fine.

PR #5201, opened 2026-05-16
api: add OIIO_NODISCARD_ERROR to ImageOutput methods

rose413:
### Description

- Annotate all ImageOutput write methods (write_image, write_scanline, write_scanlines, write_tile, write_tiles, write_rectangle, write_deep_scanlines, write_deep_tiles, write_deep_image), as well as open, close, and copy_image, with OIIO_NODISCARD_ERROR.
- Also fix the handful of internal callsites (imagebuf.cpp, maketexture.cpp, py_imagebuf.cpp, iconvert.cpp, imagebufalgo_test.cpp, and imagespeed_test.cpp) that were previously ignoring the return value of close(), either by propagating the error or by casting to void where the error path is already being handled.
- Assisted-by: Claude Sonnet 4.6
- Reference: #4781 

### Tests

- tested against unit_imageinout, unit_imagebuf, unit_imagebufalgo

### Checklist:

- [X] **I have read the guidelines** on [contributions](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/CONTRIBUTING.md) and [code review procedures](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/dev/CodeReview.md).
- [X] **I have read the [Policy on AI Coding Assistants](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/dev/AI_Policy.md)**
  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/iconvert/iconvert.cpp
+++ b/src/iconvert/iconvert.cpp
@@ -494,7 +494,9 @@ convert_file(const std::string& in_filename, const std::string& out_filename)
         } while (ok && in->seek_subimage(subimage, miplevel));
     }
 
-    out->close();
+    if (!out->close()) {
+        errorfmt("Error closing \"{}\" : {}", out_filename, out->geterror());
+    }
     in->close();
 
     // Figure out a time for the input file -- either one supplied by
--- a/src/iconvert/iconvert.cpp
+++ b/src/iconvert/iconvert.cpp
@@ -496,6 +496,7 @@ convert_file(const std::string& in_filename, const std::string& out_filename)
 
     if (!out->close()) {
         errorfmt("Error closing \"{}\" : {}", out_filename, out->geterror());
+        ok = false;
     }
     in->close();
 
--- a/src/libOpenImageIO/imagebufalgo_test.cpp
+++ b/src/libOpenImageIO/imagebufalgo_test.cpp
@@ -1743,7 +1743,7 @@ test_demosaic()
 
     if (write_files) {
         auto imageOutput = OIIO::ImageOutput::create("exr");
-        bool ok = imageOutput->open("source.exr", src_image.spec());
+        bool ok          = imageOutput->open("source.exr", src_image.spec());
         OIIO_ASSERT(ok);
         src_image.write(imageOutput.get());
     }
--- a/src/libOpenImageIO/imagebufalgo_test.cpp
+++ b/src/libOpenImageIO/imagebufalgo_test.cpp
@@ -129,9 +129,7 @@ filled_image(cspan<float> value, int width = 4, int height = 4,
 // the value array).
 inline ImageBuf
 filled_image(cspan<float> value, TypeDesc dtype)
-{
-    return filled_image(value, 4, 4, dtype);
-}
+{ return filled_image(value, 4, 4, dtype); }
 
 
 
@@ -1255,9 +1253,9 @@ test_opencv()
         std::cout << loaded_image.geterror() << 'n';
         return;
     }
-    auto cv_image = cv::Mat {};
+    auto cv_image = cv::Mat { };
     try {
-        bool ok = OIIO::ImageBufAlgo::to_OpenCV(cv_image, loaded_image, {}, 1);
+        bool ok = OIIO::ImageBufAlgo::to_OpenCV(cv_image, loaded_image, { }, 1);
         OIIO_CHECK_ASSERT(ok);
         if (!ok) {
             std::cout << "Error when converting: " << OIIO::geterror() << '\n';
@@ -1406,7 +1404,7 @@ chan_reverse(span<float> d, cspan<float> s)
 // Functor to reverse channels
 class ChannelReverser {
 public:
-    ChannelReverser() {}
+    ChannelReverser() { }
     bool operator()(span<float> d, cspan<float> s)
     {
         for (size_t c = 0, nc = size_t(d.size()); c < nc; ++c)
@@ -1448,7 +1446,7 @@ test_simple_perpixel()
         // Test with lambda, including variable capture
         float bias = 0.0;  // Force capture of this variable
         result     = ImageBufAlgo::perpixel_op(src, [&](span<float> d,
-                                                    cspan<float> s) {
+                                                        cspan<float> s) {
             for (size_t c = 0, nc = size_t(d.size()); c < nc; ++c)
                 d[c] = s[nc - 1 - c] + bias;
             return true;
--- a/src/libOpenImageIO/imagebufalgo_test.cpp
+++ b/src/libOpenImageIO/imagebufalgo_test.cpp
@@ -1613,7 +1613,7 @@ test_demosaic_algo(const ImageBuf& src_image, const ImageBuf& mosaiced_image,
         std::string ext  = type.is_floating_point() ? "exr" : "png";
         std::string path = file_name + "_" + test_name + "." + ext;
         auto imageOutput = ImageOutput::create(ext);
-        imageOutput->open(path, demosaiced_image.spec());
+        (void)imageOutput->open(path, demosaiced_image.spec());
         demosaiced_image.write(imageOutput.get());
     }
 }
@@ -1650,7 +1650,7 @@ test_demosaic(const DemosaicTestConfig& config, const ImageBuf& src_image,
                 std::string path = file_name + "_src." + ext;
 
                 auto imageOutput = ImageOutput::create(ext);
-                imageOutput->open(path, mosaiced_image.spec());
+                (void)imageOutput->open(path, mosaiced_image.spec());
                 mosaiced_image.write(imageOutput.get());
             }
 
@@ -1741,7 +1741,7 @@ test_demosaic()
 
     if (write_files) {
         auto imageOutput = OIIO::ImageOutput::create("exr");
-        imageOutput->open("source.exr", src_image.spec());
+        (void)imageOutput->open("source.exr", src_image.spec());
         src_image.write(imageOutput.get());
     }
 
--- a/src/libOpenImageIO/imagebufalgo_test.cpp
+++ b/src/libOpenImageIO/imagebufalgo_test.cpp
@@ -129,7 +129,9 @@ filled_image(cspan<float> value, int width = 4, int height = 4,
 // the value array).
 inline ImageBuf
 filled_image(cspan<float> value, TypeDesc dtype)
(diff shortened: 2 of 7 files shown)


src/libOpenImageIO/imagespeed_test.cpp
```diff
@@ -187,7 +187,7 @@ time_write_image()
     OIIO_ASSERT(out);
     bool ok = out->open(output_filename, outspec);
     OIIO_ASSERT(ok);
-    out->write_image(bufspec.format, &buffer[0]);
+    (void)out->write_image(bufspec.format, &buffer[0]);
```
lgritz: Maybe?

```suggestion
    ok = out->write_image(bufspec.format, &buffer[0]);
    OIIO_ASSERT(ok);
```

src/libOpenImageIO/imagespeed_test.cpp
```diff
@@ -203,7 +203,7 @@ time_write_scanline_at_a_time()
     size_t pixelsize         = outspec.nchannels * sizeof(float);
     imagesize_t scanlinesize = outspec.width * pixelsize;
     for (int y = 0; y < outspec.height; ++y) {
-        out->write_scanline(y + outspec.y, outspec.z, bufspec.format,
+        (void)out->write_scanline(y + outspec.y, outspec.z, bufspec.format,
```
lgritz: Maybe


```
    bool ok = true;
    for (int y = 0; y < outspec.height && ok; ++y) {
        ok = out->write_scanline(y + outspec.y, outspec.z, bufspec.format,
                                 &buffer[scanlinesize * y]);
    }
```

src/libOpenImageIO/imagespeed_test.cpp
```diff
@@ -268,11 +268,11 @@ time_write_tiles_row_at_a_time()
     imagesize_t scanlinesize = outspec.width * pixelsize;
     for (int z = 0; z < outspec.depth; z += outspec.tile_depth) {
         for (int y = 0; y < outspec.height; y += outspec.tile_height) {
-            out->write_tiles(outspec.x, outspec.x + outspec.width,
-                             y + outspec.y, y + outspec.y + outspec.tile_height,
-                             z + outspec.z, z + outspec.z + outspec.tile_depth,
-                             bufspec.format, &buffer[scanlinesize * y],
-                             pixelsize /*xstride*/, scanlinesize /*ystride*/);
+            (void)out->write_tiles(outspec.x, outspec.x + outspec.width,
```
lgritz: here, too

src/libOpenImageIO/imagespeed_test.cpp
```diff
@@ -245,10 +245,10 @@ time_write_tile_at_a_time()
     for (int z = 0; z < outspec.depth; z += outspec.tile_depth) {
         for (int y = 0; y < outspec.height; y += outspec.tile_height) {
             for (int x = 0; x < outspec.width; x += outspec.tile_width) {
-                out->write_tile(x + outspec.x, y + outspec.y, z + outspec.z,
-                                bufspec.format,
-                                &buffer[scanlinesize * y + pixelsize * x],
-                                pixelsize, scanlinesize, planesize);
+                (void)out->write_tile(x + outspec.x, y + outspec.y, z + outspec.z,
```
lgritz: Same here, maybe more like

```
bool ok = true;
for (int z = 0; z < outspec.depth && ok; ...) {
    for (int y = 0; y < outspec.height && ok; ...) {
        for (int x = 0; y < outspec.width && ok; ...) {
            ok = out->write_tile(...)
```

I guess I'm superstitious about ignoring the return code, that's what we're trying to be caching. Even though this is a test... what better time to catch a problem then when we're testing?

lgritz:
The failure for the "VFX2023" test is unrelated to your PR and is being fixed separately. If everything else is ok, we will not hold up the merge just because of that.

lgritz (changes requested):
This is looking really good.

I made some suggestions -- I think a few of the places in tests where you ignore the results, probably it would be better if we checked, and exit the loops early if there is a problem.

Oh, I guess in my explanations, I showed how to detect and exit the loop, but then didn't handle it. For tests, you don't need nice error recovery, but maybe just do a `OIIO_CONTRACT_ASSERT(ok)` after the loop?

src/libOpenImageIO/imagespeed_test.cpp
lgritz: I wrote OIIO_ASSERT, but probably `OIIO_CONTRACT_ASSERT(ok)` is the "more modern" idiom we're aiming for most of the time.

lgritz:
I have a feeling that you did a reformat of code you weren't really trying to change, and you must have a different version of clang-format locally than we are using in CI.

rose413:
What would you recommend I do to make sure I’m using the same clang-format version as CI?

lgritz:
> What would you recommend I do to make sure I’m using the same clang-format version as CI?

I don't, really. We tend to lock down to a clang-format version in CI for a few years at a time to make a stable standard, so that we're not constantly reformatting with every clang-format update. But of course, that means that for those of us who keep our own machines fairly up to date, we can drift from the CI standard. (Of course you COULD install the clang-format 17 we are using in CI but it's probably not worth the trouble.)

So first, I never run a clang-format of the full code base on my machine (for this reason -- it will change too many things I wasn't even working on). I use VSCode and so usually I'll just highlight/select the function I altered, or sometimes just the lines I changed, and ask it to reformat that selection only, rather than the whole file or the whole codebase. That at least limits any changes to just the sections I was intending to edit.

When my CI fails the clang-format test, I just look at the logs for that run, which in the "clang-format" section, will show the diffs between the actual code and the CI-formatted version, as you [can see here](https://github.com/AcademySoftwareFoundation/OpenImageIO/actions/runs/25976806688/job/76361003856?pr=5201). USUALLY it only differs on a couple lines and I just edit those by hand. 

It's usually easy to see exactly what to do from the diffs in the log. But when it's more extensive or hard to do by hand, I check the downloadable artifacts (click "Summary" on the upper left of the CI dashboard, then scroll all the way to the bottom, where you'll see downloads for each job, and one will be called clang-format). That saved artifact is a zip file of the source code with proper formatting, so you can extract the files you need formatted precisely the way the CI thought it should be done.

I'm not sure what tools you have on your end, but what I'd do first if I were you is try to revert all the places in your last commit that reformatted parts of the code you didn't intend to change at all. Then you will only have the small number of differences in the code you really did modify. That should be manageable by hand or by downloading the changed files.

Also, if you just can't figure it out, I am happy to fix the formatting for you on my end, and amend your PR with the fixes

lgritz:
Oh, one more thing -- it looks like at least one of your commit did not use `git commit -s` to affix signed-off-by line that certifies your acceptance of the DCO.

lgritz:
Aha, looks like you got all the formatting fixed!

src/iconvert/iconvert.cpp
```diff
@@ -494,7 +494,9 @@ convert_file(const std::string& in_filename, const std::string& out_filename)
         } while (ok && in->seek_subimage(subimage, miplevel));
     }
 
-    out->close();
+    if (!out->close()) {
+        errorfmt("Error closing \"{}\" : {}", out_filename, out->geterror());
+    }
```
lgritz, line 500: I think you should also set ok=false so that the function's return value indicates that something went wrong and that we don't overwrite the real file (at this point we are writing to a temporary file, and will, later i the function, replace the destination file only if everything appeared to go right).

```suggestion
        errorfmt("Error closing \"{}\" : {}", out_filename, out->geterror());
        ok = false;
    }
```

src/libOpenImageIO/imagespeed_test.cpp
```diff
@@ -202,9 +203,10 @@ time_write_scanline_at_a_time()
 
     size_t pixelsize         = outspec.nchannels * sizeof(float);
     imagesize_t scanlinesize = outspec.width * pixelsize;
+    bool ok                  = true;
     for (int y = 0; y < outspec.height; ++y) {
```
lgritz: You assign the result to ok, but ok isn't actually used anywhere. I think that it would be better if you had an OIIO_ASSERT(ok) (or OIIO_CONTRACT_ASSERT, I guess) after the loop. And the loop should probably terminate as soon as there's an error, say:

    for (int y = 0; y < outspec.height && ok; ++y)

This file is a bunch of unit tests, so it's not important to issue meaningful error messages with errorfmt or anything like that. But it is a test, so we should at least have an assertion to let us know that it's not really successfully testing the thing that we think it's testing.

I'm only writing this comment here, but I think all of the changes in this file should follow a similar pattern.

lgritz (changes requested):
We are very close. Please see my comments about at least asserting success for the changes to the tests.

src/libOpenImageIO/imagebufalgo_test.cpp
```diff
@@ -1741,7 +1743,7 @@ test_demosaic()
 
     if (write_files) {
         auto imageOutput = OIIO::ImageOutput::create("exr");
-        imageOutput->open("source.exr", src_image.spec());
+        (void)imageOutput->open("source.exr", src_image.spec());
```
lgritz: Just one more that I can see, otherwise I think you got them all.

```suggestion
        bool ok = imageOutput->open("source.exr", src_image.spec());
        OIIO_ASSERT(ok);
```

lgritz (changes requested):
Just one more tiny spot. I think you can just click the button to accept this change if you agree with it.

src/libOpenImageIO/imagebufalgo_test.cpp
```diff
@@ -1741,7 +1743,8 @@ test_demosaic()
 
     if (write_files) {
         auto imageOutput = OIIO::ImageOutput::create("exr");
-        imageOutput->open("source.exr", src_image.spec());
+        bool ok = imageOutput->open("source.exr", src_image.spec());
```
lgritz: My bad

```suggestion
        bool ok          = imageOutput->open("source.exr", src_image.spec());
```

lgritz (approved):
I pushed a small formatting fix to the line that I myself broke when I suggested it. I believe we have crossed the finish line. 

LGTM! Thanks, Irene, this looks great. Congrats on your first PR merge in this project!

rose413:
Thank you so much for the feedback and helpful suggestions! I really appreciate your help!

PR #5203, opened 2026-05-16
fix(pnm,jxl): Prevent PNM/JXL readers from loading arbitrarily large non-image files

maxwelliverson:
##### PNM
The PNM reader did not provide an optimized `valid_file` implementation. While `PNMInput::open` would perform basic validation of the PNM header, it would only do this after having read the entire contents of the file into memory. Besides the obvious risks/issues with having `open` unconditionally read arbitrarily large files into memory, this also meant that any call to `valid_file` would necessarily do the same (as the default implementation delegates to `open` when ioproxy support is present).

To fix these issues, the following changes were made:
* Provide an efficient implementation of `valid_file` for `PNMInput` that only loads the header, and then validates magic number/dimensions.
* In `PNMInput::open`, first load only the header to memory, and subsequently, only load the rest of the image data if the header is parsed successfully, and the data contained within is valid.
* Added a rough limit to header read size of 1KiB.
* Added a rough limit to full file read size of 1GiB.

Note that, for now, if the file size exceeds the 1GiB limit, the read is simply truncated to 1GiB, rather than failing altogether.

##### JPEGXL
The JXL reader already provided an efficient `valid_file` override, implemented by validating the 128 byte JXL signature. The signature, however, was not being validated in `JxlInput::open` before attempting to decode the file.
The process of JXL decoding appears to read the entire contents of the file to memory (at least for some non-JXL inputs). Taken in combination, this means that when `JXLInput::open` was called on arbitrarily large non-JXL inputs, the entire file was being read into memory before `open` would fail.

As a simple fix for this, `JxlInput::open` now checks the result of `valid_file` before attempting any decoding.

### Tests

I did not add a new testsuite case for this, but did test on my end that for both PNM and JPEGXL, large invalid files are no longer being read into memory, while valid images (of their respective types) are still read/opened successfully.

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/pnm.imageio/pnminput.cpp
+++ b/src/pnm.imageio/pnminput.cpp
@@ -74,6 +74,12 @@ private:
     bool read_file_scanline(void* data, int y);
     bool read_file_header();
 
+    static string_view read_header_to_buffer(std::vector<char>& buffer,
+                                             Filesystem::IOProxy* io);
+    static string_view append_remainder_to_buffer(std::vector<char>& buffer,
+                                                  Filesystem::IOProxy* io,
+                                                  string_view remaining);
+
     template<typename T> bool nextVal(T& val);
 
     template<class T>
@@ -111,43 +117,6 @@ static const imagesize_t max_pnm_header_size = 1024;
 // 1GiB rough limit on file size to avoid loading arbitrarily
 // large files into memory
 static const imagesize_t max_pnm_file_size = 1024 * 1024 * 1024;
-
-
-
-// Read only enough of the file to contain the header (at momst 1KB) into buffer
-static string_view
-read_header_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io)
-{
-    imagesize_t header_size = std::min(static_cast<imagesize_t>(io->size()),
-                                       max_pnm_header_size);
-    buffer.resize(header_size);
-    io->pread(buffer.data(), header_size, 0);
-    return string_view(buffer.data(), buffer.size());
-}
-
-
-
-// buffer contains at most the first 1K of the file. At this point, we know
-// the file seems valid. Read the rest in, appending to what we have, and
-// return the adjusted string_view of the contents.
-static string_view
-append_remainder_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io,
-                           string_view remaining)
-{
-    // Assume we've already read the header into buffer
-    imagesize_t header_size    = buffer.size();
-    imagesize_t full_size      = std::min(static_cast<imagesize_t>(io->size()),
-                                          max_pnm_file_size);
-    ptrdiff_t remaining_offset = remaining.data() - buffer.data();
-
-    buffer.resize(full_size);
-    io->pread(buffer.data() + header_size, full_size - header_size,
-              header_size);
-    
-    string_view result { buffer.data(), buffer.size() };
-    result.remove_prefix(remaining_offset);
-    return result;
-}
 
 
 
@@ -434,6 +403,45 @@ PNMInput::read_file_header()
     }
     m_spec.set_colorspace("Rec709");
     return true;
+}
+
+
+
+// Read only enough of the file to contain the header (at momst 1KB) into buffer
+string_view
+PNMInput::read_header_to_buffer(std::vector<char>& buffer,
+                                Filesystem::IOProxy* io)
+{
+    imagesize_t header_size = std::min(static_cast<imagesize_t>(io->size()),
+                                       max_pnm_header_size);
+    buffer.resize(header_size);
+    io->pread(buffer.data(), header_size, 0);
+    return string_view(buffer.data(), buffer.size());
+}
+
+
+
+// buffer contains at most the first 1K of the file. At this point, we know
+// the file seems valid. Read the rest in, appending to what we have, and
+// return the adjusted string_view of the contents.
+string_view
+PNMInput::append_remainder_to_buffer(std::vector<char>& buffer,
+                                     Filesystem::IOProxy* io,
+                                     string_view remaining)
+{
+    // Assume we've already read the header into buffer
+    imagesize_t header_size    = buffer.size();
+    imagesize_t full_size      = std::min(static_cast<imagesize_t>(io->size()),
+                                          max_pnm_file_size);
+    ptrdiff_t remaining_offset = remaining.data() - buffer.data();
+
+    buffer.resize(full_size);
+    io->pread(buffer.data() + header_size, full_size - header_size,
+              header_size);
+    
+    string_view result { buffer.data(), buffer.size() };
+    result.remove_prefix(remaining_offset);
+    return result;
 }
 
 
--- a/src/pnm.imageio/pnminput.cpp
+++ b/src/pnm.imageio/pnminput.cpp
@@ -5,6 +5,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <fstream>
+#include <optional>
 #include <string>
 
 #include <OpenImageIO/filesystem.h>
@@ -23,6 +24,13 @@ OIIO_PLUGIN_NAMESPACE_BEGIN
 // http://netpbm.sourceforge.net/doc/pam.html  (base format)
 //
 
+enum PNMType { P1, P2, P3, P4, P5, P6, Pf, PF };
+
+
+
+using PNMBasicInfo = std::tuple<PNMType, int, int>; // type, width, height
+
+
 
 class PNMInput final : public ImageInput {
 public:
@@ -40,10 +48,9 @@ public:
     int current_subimage(void) const override { return 0; }
     bool read_native_scanline(int subimage, int miplevel, int y, int z,
                               void* data) override;
+    bool valid_file (Filesystem::IOProxy* ioproxy) const override;
 
 private:
-    enum PNMType { P1, P2, P3, P4, P5, P6, Pf, PF };
-
     PNMType m_pnm_type;
     int m_max_val;
     float m_scaling_factor;
@@ -63,17 +70,7 @@ private:
     bool read_file_scanline(void* data, int y);
     bool read_file_header();
 
-    void skipComments()
-    {
-        while (m_remaining.size() && Strutil::parse_char(m_remaining, '#'))
-            Strutil::parse_line(m_remaining);
-    }
-
-    template<typename T> bool nextVal(T& val)
-    {
-        skipComments();
-        return Strutil::parse_value(m_remaining, val);
-    }
+    template<typename T> bool nextVal(T& val);
 
     template<class T>
     bool ascii_to_raw(T* write, imagesize_t nvals, T max, bool invert = false);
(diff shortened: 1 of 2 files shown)


maxwelliverson:
Running into issues at the moment with easycla; following the "please click here to be authorized" link above takes me to a page that starts to load something before quickly going all white. Checking the console indicates that there's possibly some authentication error of some kind?
I've enabled and authorized the github integration, ensured that third party cookies are enabled in case needed for this service to work, tried on different browsers on different hosts, all the same....

Help I've been unpersoned by the Linux Foundation

lgritz:
@maxwelliverson I checked, you're definitely on the list for this project. Are you sure that the email you used (your home gmail) on the commits is the same as what GitHub itself thinks you use? If so, then try the "Please click here to be authorized" again? Maybe paste that link into an incognito window just in case your normal browser has some cookie or plugin that's interfering with the site somehow?

If all that fails, then we'll poke LF for help. Sometimes something get wedged.

src/pnm.imageio/pnminput.cpp
```diff
@@ -104,6 +101,67 @@ OIIO_EXPORT const char* pnm_input_extensions[] = { "ppm", "pgm", "pbm",
...
+// large files into memory
+static const imagesize_t max_pnm_file_size = 1024*1024*1024;
+
+
+
+static string_view
+read_header_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io)
+{
+    // couch std::min in parentheses to work around annoying windows.h defs
+    imagesize_t header_size = (std::min)(io->size(), max_pnm_header_size);
```
lgritz: I feel like the fact that we don't do this anywhere else indicates that it's not necessary here. 

I think std::min needs `#include <algorithm>`.
Does that help?

The thing is, I'm not sure I understand how windows.h would come into play here. It's not included in this file, or in any of our headers.

src/pnm.imageio/pnminput.cpp
```diff
@@ -23,6 +24,13 @@ OIIO_PLUGIN_NAMESPACE_BEGIN
 // http://netpbm.sourceforge.net/doc/pam.html  (base format)
 //
 
+enum PNMType { P1, P2, P3, P4, P5, P6, Pf, PF };
+
+
+
+using PNMBasicInfo = std::tuple<PNMType, int, int>; // type, width, height
```
lgritz: Is there a reason not to just define a tiny helper struct that's more self-documenting? Like
```
struct PNMBasicInfo {
    PNMtype type;
    int width;
    int height;
};
```
You could also add an `Err` value to the enum, and then you wouldn't need `optional<>` at all because the struct would tell you it's invalid because of an error.

src/pnm.imageio/pnminput.cpp
maxwelliverson: Mmm noted, I can remove then. This is mostly just an old abundance-of-caution habit of mine; I didn't add it for a reason specific to the context of OIIO other than not being super familiar with the codebase as a whole and being wary of surprise windows.h inclusions somewhere down the chain that I wouldn't spot immediately due to working on Linux.

src/pnm.imageio/pnminput.cpp
```diff
@@ -104,6 +101,67 @@ OIIO_EXPORT const char* pnm_input_extensions[] = { "ppm", "pgm", "pbm",
...
+static const imagesize_t max_pnm_header_size = 1024;
+
+// 1GiB rough limit on file size to avoid loading arbitrarily
+// large files into memory
+static const imagesize_t max_pnm_file_size = 1024*1024*1024;
+
+
+
+static string_view
+read_header_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io)
```
lgritz: Let's add a comment:

```suggestion
// Read only enough of the file to contain the header (at most 1KB) into buffer
static string_view
read_header_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io)
```

src/pnm.imageio/pnminput.cpp
```diff
@@ -104,6 +101,67 @@ OIIO_EXPORT const char* pnm_input_extensions[] = { "ppm", "pgm", "pbm",
...
+    buffer.resize(header_size);
+    io->pread(buffer.data(), header_size, 0);
+    return string_view(buffer.data(), buffer.size());
+}
+
+
+
+static string_view
+read_image_data_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io,
+                          string_view remaining)
```
lgritz: Command and slight renaming suggestion:

```suggestion
// buffer contains at most the first 1K of the file. At this point, we know
// the file seems valid. Read the rest in, appending to what we have, and
// return the adjusted string_view of the conents.
static string_view
append_remainder_to_buffer(std::vector<char>& buffer, Filesystem::IOProxy* io,
                          string_view remaining)
```

maxwelliverson:
>  Are you sure that the email you used (your home gmail) on the commits is the same as what GitHub itself thinks you use?

Yup definitely, this is the primary email associated with this account.

> Maybe paste that link into an incognito window just in case your normal browser has some cookie or plugin that's interfering with the site somehow?

Just tried this and it actually got me to a non-blank page, which then prompts me to choose whether to proceed as an individual contributor or a corporate contributor. Unfortunately, it then also warns me that the site needs cookies to work and is likely to not be functional in a private window, and indeed the "proceed" buttons are greyed-out and disabled....

Given this seems like some issue with browser state, I'm going to try purging some caches/disabling plugins to see if that does the trick.

src/pnm.imageio/pnminput.cpp
lgritz: I'm also curious why you didn't make this a method of PNMInput. Would that simplify anything? Including making it so that you didn't need to pull those other methods out of the class and turn them into standalone functions when they needed no other changes?

src/pnm.imageio/pnminput.cpp
maxwelliverson: The deed is done.

lgritz:
> Just tried this and it actually got me to a non-blank page, which then prompts me to choose whether to proceed as an individual contributor or a corporate contributor. 

I can't tell if you're just telling me how far you got, or expressing that you don't know which to choose. Choose corporate.

src/pnm.imageio/pnminput.cpp
```diff
@@ -412,4 +496,37 @@ PNMInput::read_native_scanline(int subimage, int miplevel, int y, int z,
     return true;
 }
 
+
+
+bool
+PNMInput::valid_file(Filesystem::IOProxy* ioproxy) const
+{
```
lgritz: This is pretty nit-picky, and I know it doesn't matter to the compiler, but even though this is the latest addition to this file, I think it would make more logical sense for the reader of this file if valid_file appeared right before open(), rather than at the end of the file.

maxwelliverson:
> > Just tried this and it actually got me to a non-blank page, which then prompts me to choose whether to proceed as an individual contributor or a corporate contributor.
> 
> I can't tell if you're just telling me how far you got, or expressing that you don't know which to choose. Choose corporate.

Sorry, I should have been more clear there, I do know that I need to choose corporate. When I was saying that the "proceed" buttons were greyed out/disabled, I meant that those two buttons specifically were disabled, ie. I'm unable to choose either

src/pnm.imageio/pnminput.cpp
maxwelliverson: No real strong reason other than a personal affinity for the following construct
```
if (auto result = returns_optional_tuple()) {
    // unpack *result
}
```

But I think you're right that the self documenting nature of a simple struct makes more sense here.

src/pnm.imageio/pnminput.cpp
maxwelliverson: This was something I was conscious of, but hadn't managed to convince myself 100% one way or the other.
My personal preference for code organization tends towards something along the lines of "a function should only be a (possibly static) method if it requires access to class internals, or if it forms part of the interface of the class". For this reason, I made `read_header_into_buffer` a free function (as the buffer in question would not necessarily be `m_file_contents`). Then, for parity with `read_header_into_buffer`, `read_image_data_to_buffer` was also made a free function with a nearly identical declaration. It could, however, just as easily have been a private method, and perhaps does make more sense as such, as it is only ever called on the private data members of the class anyways...

lgritz: I think your instinct is mostly correct, and for fresh code I would also lean toward free functions for things that don't need class internals. Though the counter-weight to that is a desire to use methods as a kind of scoping -- if it's never called by anyone but methods of this one class, keep it in the most local scope possible. 

I don't feel strongly about it. It doesn't *need* access to class internals, but it is *operating* on data members (as it's used, in the one and only place it's called). Do whatever you think makes the code simplest and easiest to read.

src/pnm.imageio/pnminput.cpp
maxwelliverson: I've made a change to swap out the tuple for a struct, but before I commit/push, I wanted to ask if there was any particular reason we want to avoid using `std::optional`? I personally think it does make for clearer code than having `type` double as an error indicator, but I can appreciate there may be reasons to not want to introduce it (eg. avoid reliance on c++17-features-not-named-string_view)

src/pnm.imageio/pnminput.cpp
maxwelliverson: Nothing wrong with a little delousing, this makes perfect sense :)
'tis done.

maxwelliverson:
Finally managed to get the CLA sorted out! It seemed to have been some weird issue with my linux foundation account in general; going to any linux foundation login page gave me the same blank screen no matter which device/browser I tried from.

The only thing that eventually worked was.... downloading a brand new browser I had never used before and using _that_

src/pnm.imageio/pnminput.cpp
maxwelliverson: This has been added :)

src/pnm.imageio/pnminput.cpp
lgritz: I think it's mostly historical. std::optional is a C++17 feature, and it's only fairly recently (OIIO 3.0) that we raised our minimum supported C++ to 17. So it really just hasn't come up, since all the major APIs were in place already. I guess that I, in particular, just still haven't worked it into my standard toolbox.

src/pnm.imageio/pnminput.cpp
maxwelliverson: I think that makes a lot of sense. I just pushed another commit that I think strikes a decent balance: both functions are now static methods of `PNMInput`, but are otherwise unchanged. I also moved their definitions further down so they're just before `valid_file` and `open`.

maxwelliverson:
Argh, apologies. 
For some reason, running `clang-format` locally appears to not be matching the output of the `clang-format` CI test, hence the numerous post-hoc tiny commits to fix small formatting issues. I promise I'm normally better than this ahaha

I've tried the make target, running manually on the specific files with `--style file`, running the `run-clang-format.bash` build script etc., but everything appears to produce (the same) different results (its flagging formatting issues in like 80% of source/header files across the whole repo, so something is definitely wrong).
My best guess is that it has something to do with that fact that I had updated my local llvm suite a week or so ago to the most recent release, 22.1.5? Either that or I have something else weird on my system messing with defaults... Oy vey. I'll continue investigating on my end, but hopefully that last commit should do it...

src/pnm.imageio/pnminput.cpp
maxwelliverson: Gotcha gotcha, that makes sense. Would you like me to remove its use then? Sorry I just wanted to be clear about what was expected/preferred here.

lgritz:
clang-format has an annoying habit of changing quite a bit (in non-back-compatible ways) from one release to the next. And so we don't have constant churn with every new clang-format release, we tend to lock down to a particular version for CI for a couple years at a time, and so our individual machines may drift form that.

See my explanation I just wrote minutes ago here:
https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5201#issuecomment-4484624578

lgritz (approved):
LGTM! Thanks for the quick fix on this!

I'm marking this as a Dev Days contribution. Why not? It came in within 24 hours of DD, and you are a first-time contributor to the repo.

PR #5204, opened 2026-05-17
ci: Remove jobs using the aswf-docker 2022 containers

lgritz:
In main (working toward the late-2026/2027 releases) these 2022 containers are both outside the promised 3 year support window, and also are using CentOS 7-based containers that are essentially irrelevant now that all the studios have moved to RHEL9 (and Rocky/Alma) compatible OS distros. So there is really no need to continue CI testing on them for branches aimed at the future.

Note that we still have Ubuntu-based "oldest gcc" and "oldest clang" jobs, so dropping this does not actually impact our ability to test against the oldest gcc, C++, and python that we claim to support.


lgritz:
This seem ok to everyone? Does anybody feel like we would be losing testing coverage that we still need? (Reminder: in main / OIIO 3.2 only, not losing this coverage from supported release branches.)

lgritz:
Over a week, no objections, CI-only change -> merge

PR #5205, opened 2026-05-17
build: Improve error message about mismatching git tags

lgritz:
This message is triggered by the check that the tag of an auto-build dependency matches the SHA hash we use as a cross-check. This patch is just amending the error message to reveal the way to override the check, so people (like me) don't have to go hunting for it if they are purposely trying to build from a nonstandard commit of a dependency.


lgritz:
This has gone a week without objections. All it does is add helpful information to a build error message. So I am merging.

PR #5209, opened 2026-05-20
feat(oiiotool): flexible offset notation with commas

### LINKED ISSUE #5207: bug: oiiotool --offset syntax is not good with expressions
On the Slack channel, Liam Collod wrote:

I am working with oiiotool and I have an issue with expressions. I am setting the --origin which relies on the x,y being prefixed by their sign. However both x and y are expression which means their result may be positive or negative without knowing in advance.

* `--origin +{exprX}+{exprY}` : If I add the + prefix and exprY result is negative, oiiotool will complain. 
* `--origin -{exprX}-{exprY}` If do the inverse with a - prefix and exprY is negative, oiiotool will complain too. 
* `--origin {exprX}{exprY}` And I don't add a sign prefix positive expression will fail  if exprY is positive.

How do I handle this ?

The only solution I found is to store the sign in an user variable using oiiotool --if conditions but that looks to be a quite complicated solution. (`--if {exprx>0} --set xsign "+" --else --set xsign "" --endif --origin {xsign}{exprX}...`)

lgritz:
Fixes #5207

Sometimes the `WxH+X+Y` and `+X+Y` notation (borrowed from X11!) can be a little awkward. This patch extends the accepted syntax to allow commas to separate the offset components, for example also accepting `WxY,X,Y` and `X,Y`.

This affects the commands

    --create
    --crop
    --cut
    --fit
    --fullsize
    --origin
    --originoffset
    --paste
    --pattern
    --printstats
    --resize

### DIFF
--- a/src/doc/oiiotool.rst
+++ b/src/doc/oiiotool.rst
@@ -2503,7 +2503,7 @@ current top image.
 
         --create 1920x1080 3         # RGB with w=1920, h=1080, x=0, y=0
         --create 1024x768+100+0 4    # RGBA with w=1024, h=768, x=100, y=0
-        --create 1024x768,100,0 4    # same, usign comma notation
+        --create 1024x768,100,0 4    # same, using comma notation
         --create:type=uint8 1920x1080 3  # RGB, store internally as uint8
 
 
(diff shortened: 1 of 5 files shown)


lgritz:
Tomorrow will be a week since this was posted. If there are no objections or requests for change by then, I will merge it.

PR #5210, opened 2026-05-20
fix: Remove redundant declarations

lgritz:
This is an emergency fix on top of #5201.

PR 5201 passed CI completely.  Then I merged it, and now it fails. How did that happen?

OOOOOOH. The imagespeed_test.cpp is not compiled at all for CI! Only for local builds. So a build break in it will not be caught in CI. OK, fixing that, too. It doesn't run in CI, it's not that kind of test, it's a special benchmarking test meant to run locally and needs to be told a specific image file to test with.  But at least build it for CI so we can't inadvertently break it again.


lgritz:
Merging immediately to fix badly broken build, sorry.

PR #5211, opened 2026-05-20
fix(cmake): Make the cmake config files always pull in the required libraries

DarkDefender:
Without these changes, we could get link time errors as openimageio doesn't pull in the libraries correctly when using modern cmake targets. It would work when libraries have been installed globally on the system. But I don't think we should rely on the linker automagically finding the libraries in `/usr/lib` for example.

Especially because we can make this a lot better by warning/erroring out in the cmake setup step when we can't find all required libraries.

I added the libraries that I saw show up in `OpenImageIOTargets-release.cmake` under the `OpenImageIO::OpenImageIO` target properties on my system.

### Minimal test example

Create a text project with this C++ file called `main.cxx`:

``` cpp

int main() {
  OIIO::ImageBuf image("test.tiff");
  image.write("output.exr");
  return 0;
}
```
And this `CMakeLists.txt` file:

``` cmake
cmake_minimum_required(VERSION 3.23)

project(test_oiio_proj)

set(Imath_ROOT ${CMAKE_SOURCE_DIR}/libs/imath)
set(OpenEXR_ROOT ${CMAKE_SOURCE_DIR}/libs/openexr)
set(openjph_ROOT ${CMAKE_SOURCE_DIR}/libs/openjph)
set(TBB_ROOT ${CMAKE_SOURCE_DIR}/libs/tbb)
set(OpenColorIO_ROOT ${CMAKE_SOURCE_DIR}/libs/opencolorio)
set(OpenImageIO_ROOT ${CMAKE_SOURCE_DIR}/libs/openimageio)
find_package(OpenImageIO REQUIRED)

add_executable(test_oiio)

target_sources(test_oiio
  PRIVATE
    main.cxx
)

target_link_libraries(test_oiio OpenImageIO::OpenImageIO)
```

**ENSURE** that you do not have OpenColorIO or OpenEXR installed globally on your system.

Make sure that you have `git-lfs` installed and configured on your system and then clone the precompiled linux libraries from blender into a `libs` directory next to the other files:
`git clone https://projects.blender.org/blender/lib-linux_x64.git libs`

Now if you run cmake and then try to build you will get a lot of errors like the ones below at link time:
```
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.0/ld: libOpenImageIO.so.3.1.13: undefined reference to `ojph::get_error()'
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.0/ld: libOpenImageIO.so.3.1.13: undefined reference to `Imf_3_4::Header::dwaCompressionLevel()
/usr/x86_64-pc-linux-gnu/binutils-bin/2.46.0/ld: libOpenImageIO.so.3.1.13: undefined reference to `exr_get_user_data'

```

Go and edit the `libs/openimageio/lib/cmake/OpenImageIO/OpenImageIOConfig.cmake` file and add the following on line 42:

``` cmake

find_dependency(OpenColorIO)
find_dependency(OpenEXR)
if (TRUE)
    find_dependency(TBB)
endif ()
```

Clean out the cmake build files and rerun cmake and try to build again.

Notice that the build finishes without any linker errors.

I don't know if there is any easy way to add tests for this in the test suite, so that is why I left that box unchecked.

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/cmake/Config.cmake.in
+++ b/src/cmake/Config.cmake.in
@@ -16,26 +16,29 @@ if (NOT @fmt_LOCAL_BUILD@ AND NOT @OIIO_INTERNALIZE_FMT@)
     find_dependency(fmt)
 endif ()
 
+find_dependency(OpenColorIO)
+find_dependency(OpenEXR)
+if (@OIIO_TBB@)
+    find_dependency(TBB)
+endif ()
+if (@JPEG_FOUND@)
+    find_dependency(JPEG)
+endif()
+if (@WebP_FOUND@)
+    find_dependency(WebP)
+endif()
+
 if (NOT @BUILD_SHARED_LIBS@)
     # This is required in static library builds, as e.g. PNG::PNG appears among
     # INTERFACE_LINK_LIBRARIES. If the project does not know about PNG target, it will cause
     # configuration error about unknown targets being linked in.
     find_dependency(TIFF)
-    find_dependency(OpenColorIO)
-    if (@JPEG_FOUND@)
-        find_dependency(JPEG)
-    endif()
     if (@PNG_FOUND@)
         find_dependency(PNG)
     endif()
     if (@DCMTK_FOUND@)
         find_dependency(DCMTK)
     endif()
-    # The following have the same problem except that INTERFACE_LINK_LIBRARIES use
-    # TARGET_NAME_IF_EXISTS, so the error only happens on link time.
-    if (@OIIO_TBB@)
-        find_dependency(TBB)
-    endif ()
 endif ()
 
 # Compute the installation prefix relative to this file. Note that cmake files are installed


lgritz:
I'm a little confused about the circumstances we're talking about.

These libraries are not exposed in the public headers of OIIO, so projects downstream do not need to know where to find the include files, etc.

The theory was that for static library libOpenImageIO.a, static libs don't inherently know about their link needs, so the downstream app using OIIO needs to know about those targets in order to know to link the libraries to the final app.

For dynamic library libOpenImageIO.so, it knows where which libraries it in turn needs, so our belief was that this takes care of it all and the downstream didn't need to know about the targets. So we only did the `find_dependency` when building static libs.

But you're saying: no, libOpenImageIO.so might know it needs to link against libtbb.so (say), but it doesn't know where to find it if it's not in a standard place. So even for dynamic libOpenImageIO, it would be better for the exported cmake config to establish and use the targets for all the dependencies.

OK, I can buy that. But in that case, I think the solution is to just get rid of the `if (NOT @BUILD_SHARED_LIBS@)` test itself, rather than to move only some of the dependencies out of that `if` clause. There's probably no point moving just the couple libraries you have in nonstandard places, but leaving others inside that `if`. Better to remove the `if` and do them all unconditionally. Or am I misunderstanding?

lgritz:
Oh, and it's probably worse than that, all the CI is failing.

How does this happen? Did you not run CI at all prior to submitting the PR?

DarkDefender:
> Oh, and it's probably worse than that, all the CI is failing.
> 
> How does this happen? Did you not run CI at all prior to submitting the PR?

Not all of the CI pipeline is failing. Out of all 30 checks 8 is failing. If I read the logs correctly it seems like it fails because it can't find some image files? I looked at the contribution guidelines and there it says that if the majority for the CI tests pass (which I think 22 passing out of 30 is the majority), then it is OK to submit a PR and figure out what is wrong in the PR.

I'm not really sure why this cmake change would make the tests fail they way they are. I would think that there would be at least some cmake output hinting that it can't find the required libraries if that were the case. Do you have any clue what is going on?

> But you're saying: no, libOpenImageIO.so might know it needs to link against libtbb.so (say), but it doesn't know where to find it if it's not in a standard place. So even for dynamic libOpenImageIO, it would be better for the exported cmake config to establish and use the targets for all the dependencies.

Some of the libraries have exposed symbols in the `libOpenImageIO.so` library yes. So you could in theory only link to OpenImageIO but then use openexr functions or definitions directly. Because of this, the linker needs to know where these are defined at link time.

 > These libraries are not exposed in the public headers of OIIO, so projects downstream do not need to know where to find the include files, etc.
 
 Ah, I see! I thought this was on purpose. I guess then perhaps the correct solution is to not expose these symbols in the OpenImageIO library and only expose OpenImageIO specific APIs?

DarkDefender:
You can check the exported symbols on linux with `nm -D libOpenImageIO.so`.

For example you can see that the libOpenImageIO.so does export OpenEXR functions:
```
 nm -D libOpenImageIO.so | grep exr
                 U exr_attr_get_envmap
                 U exr_decoding_choose_default_routines
                 U exr_decoding_destroy
                 U exr_decoding_initialize
                 U exr_decoding_run
                 U exr_decoding_update
                 U exr_finish
                 U exr_get_attribute_by_index
                 U exr_get_attribute_count
                 U exr_get_channels
                 U exr_get_compression
                 U exr_get_count
                 U exr_get_data_window
                 U exr_get_default_error_message
                 U exr_get_display_window
                 U exr_get_error_code_as_string
                 U exr_get_level_sizes
                 U exr_get_name
                 U exr_get_scanlines_per_chunk
                 U exr_get_storage
                 U exr_get_tile_descriptor
                 U exr_get_tile_levels
                 U exr_get_user_data
                 U exr_read_scanline_chunk_info
                 U exr_read_tile_chunk_info
                 U exr_start_read
                 U exr_test_file_header
 ```
 
 There is also opencolorio, imath, and other library symbols in there.

**Comment by lgritz** [comment]
> ```
> nm -D libOpenImageIO.so | grep exr
>                   U exr_attr_get_envmap
>                   U exr_decoding_choose_default_routines
> ```

`U` means "undefined." These are not symbols that the OpenImageIO library is exposing or providing; they are the symbols that the library needs to get dynamically from other libraries at run time. There is not a way to hide that, and there is also not a way that downstream projects can use it.

**Comment by lgritz** [comment]
>  If I read the logs correctly it seems like it fails because it can't find some image files?

No, that's just the downstream symptom.

The tests that are failing are cmake-consumer and docs-examples-cpp. These are the two tests that try to build something as if it were a downstream project, and thus consuming the exported cmake config that you altered. (Whereas all the 200+ other tests are just using the OIIO build components directly, *within* the OIIO build system so to speak, rather than acting like a separate project that doesn't have access to our own build scripts.)

If you download the saved artifacts from those jobs, it also contains the build logs from those secondary builds, and for example cmake-consumer looks like:

```
-- The CXX compiler identification is GNU 12.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/ccache/g++-12 - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building consumer  - Release
CMake Error at /usr/local/share/cmake-3.31/Modules/CMakeFindDependencyMacro.cmake:76 (find_package):
  By not providing "FindOpenColorIO.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "OpenColorIO", but CMake did not find one.

  Could not find a package configuration file provided by "OpenColorIO" with
  any of the following names:

    OpenColorIOConfig.cmake
    opencolorio-config.cmake

  Add the installation prefix of "OpenColorIO" to CMAKE_PREFIX_PATH or set
  "OpenColorIO_DIR" to a directory containing one of the above files.  If
  "OpenColorIO" provides a separate development package or SDK, be sure it
  has been installed.
  /home/runner/work/OpenImageIO/OpenImageIO/dist/lib/cmake/OpenImageIO/OpenImageIOConfig.cmake:43 (find_dependency)
  CMakeLists.txt:22 (find_package)
```

So we can make sure the export contains `find_dependency(OpenColorIO)`, but if the downstream process doesn't know where to find OpenColorIO's cmake exports, it will be a build error for the downstream project, **even though no OpenImageIO .h files contain any references to OpenColorIO's API**.


> Some of the libraries have exposed symbols in the `libOpenImageIO.so` library yes.

I believe they do not. They only have recorded OpenImageIO's *need* for those symbols to be resolved. They are not exported symbols.

> So you could in theory only link to OpenImageIO but then use openexr functions or definitions directly.
> ...
> Ah, I see! I thought this was on purpose. I guess then perhaps the correct solution is to not expose these symbols in the OpenImageIO library and only expose OpenImageIO specific APIs?

We already do that.

I think you cannot use that from downstream apps, because there are no function declarations for those OpenEXR functions anywhere in OIIO's header files, and OpenImageIO library doesn't provide the implementations.

But the fact that a downstream project can't use those definitions directly is not an important detail here, because it's still true that libOpenImageIO needs those symbols to be resolved at runtime so that IT can call the functions internally. It needs to be found at load time in order to execute. 

The libOpenImageIO.so library itself knows it needs libOpenColorIO.so (whereas libOpenImageIO.a, as a static library, would not), but it still assumes that there are enough clues at runtime to FIND that library.

This is the function of the `LD_LIBRARY_PATH` environment variable.

I believe that is the key here. I think that changing the cmake exports is not really going to fully solve the problem (and may create new ones). I think the real solution is that if you are going to store required libraries in a nonstandard place, you should expect to need to put those nonstandard locations in your `LD_LIBRARY_PATH`.

lgritz:
@DarkDefender Any comments on this?

DarkDefender:
@lgritz  sorry for the late reply!
I wanted to do some more tests on my side before I got back to you.

> If you download the saved artifacts from those jobs, it also contains the build logs from those secondary builds, and for example cmake-consumer looks like:

Right, thanks for pointing this out. From what I can tell on the linux side, the VFX2024 and VFX2023 CI workers fails because:

1. The opencolorio cmake files are not in a standard location. (I managed to work around it by adding cmake hints to `OpenColorIO_LOCAL_INSTALL_DIR`).
2. TBB doesn't seem to have any .cmake files shipped.
3. On VFX2023, the WebP .cmake files are missing too.

On the more recent VFX platform releases, all the .cmake files for the libraries seems to be installed, so it works.

Don't know if you feel it is worth making sure these work by having cmake be able to find them. I'll go into why I proposed these changes in the next section.

> I believe that is the key here. I think that changing the cmake exports is not really going to fully solve the problem (and may create new ones). I think the real solution is that if you are going to store required libraries in a nonstandard place, you should expect to need to put those nonstandard locations in your LD_LIBRARY_PATH.>

I think that would make sense if the libraries are staying in those non standard locations at runtime. However this is not true for the Blender build system (where I first ran into this linking issue). There the libraries are in their own folders during compile and link time. At runtime, they are put into a library folder that the compiled binary can find.

The Blender build system uses the cmake `_ROOT` hints where it can to try to ensure that it does not link against any system wide libraries that might have been installed (when using the precompiled blender libraries).

To me, I feel that being able to rely on cmake these `_ROOT` hints for the libraries that provides their own .cmake files is really nice. Because then, regardless if you are using static or shared libraries, you will get nice cmake errors when configuring that certain required libraries are not found, wrong version etc.

On the Blender side, the issue with not pulling in all the libraries has been worked around by manually calling `find_package` on the required libraries before calling `find_package` on openimageio. As OpenImageIO was the only library in Blender that had to do this for, I thought it would be good to try to fix it upstream.

If Blender didn't manually call `find_package` for some of the OpenImageIO dependencies, then the required link time library paths wouldn't properly be pulled in for the modern `OpenImageIO::OpenImageIO` cmake target. For example, the `OpenEXR::OpenEXR` target wouldn't exist when cmake parses the `OpenImageIOTargets-relwithdebinfo.cmake` file, and thus wouldn't know to pull in the library path to it at link time.

Of course as you stated, you could work around this with `LD_LIBRARY_PATH`, but doesn't that kinda defeat the purpose of the modern cmake targets if it can't pull in the required libraries?

Feels a bit weird to me as well to have this mostly work with static libraries, but not shared libraries, as it is currently.

lgritz:
@jessey-git Do you have any insight here?
I'm not familiar with the Blender build setup.
And I also admit that I may just not know the right way to make our cmake exports robust to things being in nonstandard locations, what the expectations and current best set of idioms are.

jessey-git:
Sorry, I'm not great with cmake, or builds in general, and don't quite understand the issue as presented myself. 

I can chip in with concrete details if needed but I don't want to derail with unnecessary details since I don't know where the problem first originates.

kmilos:
> I think that changing the cmake exports is not really going to fully solve the problem (and may create new ones). 

I agree. Removing `if (NOT @BUILD_SHARED_LIBS@)` would mean one would have to install `*-devel` packages for **_all_** OIIO dependencies just to build some app using OIIO, even when linking to it dynamically. Shouldn't be the case.

brechtvl:
From what I can tell the CMake convention is not to find packages for private dependencies with `BUILD_SHARED_LIBS`. And so I think the changes in this PR should not be needed.

For context, the classic `ld` linker is pedantic and needs indirect library dependencies at link time (newer linkers like `gold`, `lld` and `mold`  do not). If those indirect library dependencies are in standard locations like `/usr/lib`, or a link directory that was specified for another reason, the linker is also able to find them. Additionally, the `libOpenImageIO.so` `RUNPATH` should contain the directories to find them, which is probably the way this works in e.g. GitHub CI.

Where it gets tricky is if you build libraries on one system, and then copy them to another directory or another system and the absolute paths in `RUNPATH` break. But that's a problem to solve on the Blender side.

lgritz:
I'm going to close this, because I think we agree that this is probably not the correct change.

But if further investigation reveals that we do have a problem on the OIIO side, or even simply that there is something we could do to make the solution easier Blender, please do not hesitate to suggest something and we can revisit this.

PR #5212, opened 2026-05-23
testing: new ref image for libheif 1.22

lgritz:
New release of libheif, which continues to live up to its habit of changing pixel values ever so slightly with each release.


lgritz:
No comments, just adding a new ref output -> merge

PR #5218, opened 2026-05-30
api: add OIIO_NODISCARD_ERROR to Imageio.h

zoomhunter2010:
### Description
 
Add OIIO_NODISCARD_ERROR to make sure callers check their bool return values

Assisted by: Claude Sonnet 4.6
Reference: https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4781

### Tests

No new test suite case added

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/include/OpenImageIO/imageio.h
+++ b/src/include/OpenImageIO/imageio.h
@@ -456,7 +456,7 @@ public:
     /// whole image.  If this returns false, the image is much too big
     /// to allocate and read all at once, so client apps beware and check
     /// these routines for overflows!
-    bool size_t_safe() const noexcept {
+    OIIO_NODISCARD_ERROR bool size_t_safe() const noexcept {
         const imagesize_t big = std::numeric_limits<size_t>::max();
         return image_bytes() < big && scanline_bytes() < big &&
             tile_bytes() < big;
@@ -775,7 +775,7 @@ public:
     /// Helper function to verify that the given pixel range exactly covers a
     /// set of 2D tiles.  Also returns false if the spec indicates that the
     /// image isn't tiled at all.
-    bool valid_tile_range (int xbegin, int xend, int ybegin, int yend) noexcept {
+    OIIO_NODISCARD_ERROR bool valid_tile_range (int xbegin, int xend, int ybegin, int yend) noexcept {
         return (tile_width &&
                 ((xbegin-x) % tile_width)  == 0 &&
                 ((ybegin-y) % tile_height) == 0 &&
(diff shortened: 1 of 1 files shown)


src/include/OpenImageIO/imageio.h
```diff
@@ -456,7 +456,7 @@ class OIIO_API ImageSpec {
     /// whole image.  If this returns false, the image is much too big
     /// to allocate and read all at once, so client apps beware and check
     /// these routines for overflows!
-    bool size_t_safe() const noexcept {
+    OIIO_NODISCARD_ERROR bool size_t_safe() const noexcept {
```
lgritz: This isn't an error code, it's the actual result. This is a case where we should use `OIIO_NODISCARD`.

lgritz (changes requested):
All of the changes you made used `OIIO_NODISCARD_ERROR`, but none of the functions annotated return error codes. Those are all cases where they really do return a value, and so so should use `OIIO_NODISCARD`.

Also, while it's ok to just change a few functions, I'm wonder how and why you chose just these three in particular, instead of, say, choosing one class and annotating all the appropriate member functions in that class?

src/include/OpenImageIO/imageio.h
zoomhunter2010: To answer your question, I'm a part of the same team as rose413 and PandorasPallet from codeday, so I was only taking on a few functions from the ImageSpec class to annotate since I am relatively new to GitHub and C++. I just chose a few that I understood best and felt I could make the changes to. Thank you for your help and feedback.

lgritz: Ah, I see. That's all fine.

lgritz (approved):
LGTM! Thanks for the patch.

lgritz:
Can I ask a question without it seeming accusatory? I'm always trying to learn more about how people are using coding assistants in practice.

Your PR said "Assisted by: Claude Sonnet 4.6", and considering that your patch changed only 3 lines (and, um, not entirely correctly the first time), I'm really curious in what way you used the tool. What did you ask it to do, and do you feel like it saved you any time for this task?

zoomhunter2010:
For me, I try not to let it do much of the work for me. I try to use it to explain things to me or help me skim through the larger files. I do believe it's a bad habit because it isn't always right, and it can just make stuff up. I think for this task, it didn't end up saving me time.

lgritz:
I think that's 100% the right attitude.

Our request to document LLM use with "assisted-by" is for if you are having it directly help you write the code you submit. If you're just using it to explain existing code to you, find things in a big codebase, answer general questions about programming, or even review the code you're written before you submit a PR, I don't think you need to bother with the "assisted-by."

PR #5219, opened 2026-06-01
ci: Fixes to build_opencolorio.bash script

lgritz:
It was throwing things off to set CMAKE_CXX_COMPILER in this way, wrecking the ability to just run as a utility from the command line without it being set.  It was not necessary, as far as I can tell.


lgritz:
This has been posted for a week. Any objections?

PR #5224, opened 2026-06-02
api: span and string_view get extra contract asserts per std

lgritz:
According to C++26, some newly required std container hardening for std::span and std::string_view were not reflected in our own analogous types. In particular, some more precondition checking was required for span::subspan, first, last, and string_view::remove_prefix and remove_suffix in order to match the expectations of true parity with with the std versions.

This is important because some day, we will probably discard our own string_view and span for the std ones (especially when we are using compilers in which they are fully hardened).

I found a couple spots where we used remove_prefix and remove_suffix that assumed the range would be clamped, as it is in our version, rather than contractually require them to be valid ranges as in std. This would eventually fail if replaced by std::string_view, so fixing now.

### DIFF
--- a/src/include/OpenImageIO/span.h
+++ b/src/include/OpenImageIO/span.h
@@ -154,40 +154,51 @@ public:
     /// Assignment copies the pointer and length, not the data.
     constexpr span& operator= (const span &copy) = default;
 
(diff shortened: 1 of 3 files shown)


lgritz:
This has been posted for a week. Any objections?

PR #5225, opened 2026-06-03
build: Support gcc-16 and test it in CI

lgritz:
Bump the "bleeding edge" variant to gcc-16, and the "latest releases" variant to gcc-15.

One minor fix needed to make gcc16 happy -- it is better than earlier versions at discovering variables whose values are not used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,12 +298,12 @@ jobs:
                             EXTRA_DEP_PACKAGES="clang-format-17"
                             CLANG_FORMAT_EXE=clang-format-17
 
-          - desc: latest releases gcc13 C++20 py3.12 avx2 exr3.4 ocio2.4
+          - desc: latest releases gcc15 C++20 py3.12 avx2 exr3.4 ocio2.4
             nametag: linux-latest-releases
             oiio_python_bindings_backend: both
             runner: ubuntu-24.04
-            cc_compiler: gcc-13
-            cxx_compiler: g++-13
+            # cc_compiler: gcc-13
+            # cxx_compiler: g++-13
             cxx_std: 20
             fmt_ver: 12.1.0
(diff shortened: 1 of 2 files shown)


lgritz:
This is an extremely minor change that allows us to build with gcc-16. Will merge tomorrow if there are no objections by then.

.github/workflows/ci.yml
```diff
@@ -298,12 +298,12 @@ jobs:
...
 
-          - desc: latest releases gcc13 C++20 py3.12 avx2 exr3.4 ocio2.4
+          - desc: latest releases gcc15 C++20 py3.12 avx2 exr3.4 ocio2.4
             nametag: linux-latest-releases
             oiio_python_bindings_backend: both
             runner: ubuntu-24.04
-            cc_compiler: gcc-13
-            cxx_compiler: g++-13
+            # cc_compiler: gcc-13
+            # cxx_compiler: g++-13
```
jessey-git, line 306: Any reason to leave these commented rather than remove them altogether?

I have to say, setting the compiler version is very confusing. Some tasks like this just use `gcc_action_ver` but other tasks end up setting both `cc_compiler` as well as using `gcc_action_ver`.  Should those other ones be cleaned up to only use gcc_action_ver?

lgritz: I can delete those. In a way, I guess the commented out line is kind of a note to myself that reminds me that setting cxx_compiler is not what works here. But that's not a very good excuse.

It should probably all be cleaned up. It's been growing by accretion, and we have different methods for different versions of gcc, and I admit that it is confusing. I wanted this PR to just be a tiny adjustment to ensure we tested gcc-16, but an overhaul is definitely called for. Maybe as a follow-up?

jessey-git: Followup is fine I think

jessey-git (approved):

PR #5226, opened 2026-06-04
docs: Clarify plugin.h explanations by saying their Unix-like equivalents

lgritz:
Mostly for the sake of bots (but also humans) reading plugin.h and trying to figure out exactly what each function does, just plainly state which Unix-like dlopen/dlsym/etc function it is the equivalent of.


lgritz:
Comments-only change, will merge tomorrow if there are no objections by then.

PR #5228, opened 2026-06-09
GPU texture system prototype

aconty:
### Description

Implemented as a standalone testsuite prototype, without modifying the core OpenImageIO library, this work demonstrates the proposed data model and runtime design in a working executable that already produces images.

The prototype validates the host/device architecture end to end: request-driven texture lookup, residency updates, retry flow, filtering behavior, and output generation. It is intentionally scoped to the testsuite so we can iterate quickly on API shape, data structures, and execution strategy before promoting pieces into library code.

More details in README.md, started from SPEC.md. Designed and written by me, assisted by gpt-5.3-codex in copilot.

Nobody wanted to jump into this task, including myself. So I wrote a prototype SPEC for AI to do it. Since I didn't like where it was going I took full control. of it and ended up enjoying the job a lot. Came up with a couple of patterns that I like a lot. My aim was simplicity and clarity. Hope you people like it.

### Tests

The prototype itself is a test. This doesn't interfere with the rest of the library.

### Checklist:

  behavior.
  on the GitHub runners.)


aconty:
Here is the README, btw https://github.com/imageworks/OpenImageIO/blob/gputex/testsuite/texture-device/README.md

testsuite/texture-device/tagged_ptr.h
```diff
@@ -0,0 +1,171 @@
...
+    return OIIO::Strutil::strhash64(std::strlen(s), s);
+}
+
+template<size_t N>
+inline constexpr uint64_t
+ptrtag(const char (&s)[N])
+{
+    static_assert(N > 1, "tag literal must be non-empty");
+    return OIIO::Strutil::strhash64(N - 1, s);
+}
```
lgritz: Does the string_view overload of Strutil::strhash64 subsume both of these?

testsuite/texture-device/vector_lite.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <array>
+#include <cassert>
+#include <cstddef>
+#include <utility>
+
+namespace texture_device {
+
+template<class T, size_t N> struct vector_lite : public std::array<T, N> {
+    using Base = std::array<T, N>;
```
lgritz, line 16: C++26 includes a new [`inplace_vector<>`](https://en.cppreference.com/cpp/container/inplace_vector) that I think is the same as this. If so, we should adopt their nomenclature and as much of the semantics as makes sense.

I'm a little suspicious of it being is-a std::array versus has-a. Is it just to save a few lines by not having to repeat `operator[]` and the like? Are we inadvertently inheriting anything from std::array that we wouldn't want here?

testsuite/texture-device/vector_lite.h
```diff
@@ -0,0 +1,52 @@
...
+
+    size_t size() const { return m_size; }
+    size_t capacity() const { return N; }
+    bool empty() const { return m_size == 0; }
+
+    void clear() { m_size = 0; }
+
+    void push_back(const T& value)
+    {
+        assert(m_size < N);
```
lgritz: I would prefer OIIO_CONTRACT_ASSERT these days.

testsuite/texture-device/texture_loader.h
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+private:
+    struct TextureMetadata {
+        int width  = 0;
+        int height = 0;
+    };
+
+    std::vector<std::string> m_texture_paths;
+    std::unordered_map<std::string, TextureMetadata> m_metadata_cache;
```
lgritz, line 70: Would these be better as ustring? (asking; I do not yet have a strong opinion)

testsuite/texture-device/texture_loader.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    if (spec.tile_width != int(TileRecord::kTileWidth)
+        || spec.tile_height != int(TileRecord::kTileHeight)) {
+        in->close();
+        return loader_error("unsupported tile size for " + filename + " got "
+                            + std::to_string(spec.tile_width) + "x"
+                            + std::to_string(spec.tile_height) + " expected "
+                            + std::to_string(TileRecord::kTileWidth) + "x"
+                            + std::to_string(TileRecord::kTileHeight));
```
lgritz: Strutil::format is probably more idiomatic for OIIO than `string::operator+` (and almost certainly much faster). If we want to be really fancy, we can make loader_error itself format input instead of passing an already formatted string.

lgritz (approved):
I see where you're going with this, and LGTM. The threshold is pretty low since this is just a testsuite experiment for now, so by no means does anything need to be perfect or final.

I made a number of comments, but really it's nothing that we can't fix up as we continue to work on it, if you are inclined to address any of it in subsequent PRs rather than now.

I would like to discuss it at the TSC meeting today before we merge it, to make sure nobody has any reservations.

One overarching comment/question -- would it be a big simplification to the scheme (and would it be feasible and performant on today's HW) to use unified/managed memory rather than separate allocations and explicit copies?

lgritz:
Needless to say, there are a whole bunch of things here that are idiomatically different from TextureSystem (in implementation as well as in use). I'm noting that here only to say that as we eventually make this all part of TS, we'll probably want to revisit which should become more like the other, or if TS should support multiple usage idioms. 

For example, should TS, even on the CPU, have a set of texture lookup calls that have a "texture not resident" result rather than being obligated to load it within the texture lookup? And a "here are a set of tiles we need, go get them all as efficiently as possible" function. It would be interesting to know how that approach to texture compares to what we're doing now, maybe we'd find it to be the method of choice no matter which HW you're running on.

testsuite/texture-device/vector_lite.h
aconty: Will do that change.

testsuite/texture-device/vector_lite.h
aconty: Sure

testsuite/texture-device/texture_loader.h
aconty: Indeed

aconty:
Your unified memory support comment got me thinking. I'm going to spin this a bit more because I think I can do that and also simplify.

The API design, yes, probably better to discuss that live.

aconty:
I updated this prototype with unified memory support

testsuite/texture-device/arena.cpp
```diff
@@ -0,0 +1,377 @@
...
+    {
+        if (allocated.empty())
+            return;
+
+        std::fprintf(stderr,
+                     "texture-device: %s leak check failed (%zu allocations)\n",
+                     owner, allocated.size());
+        for (const auto& it : allocated) {
+            const auto& rec = it.second;
+            std::fprintf(stderr, "  leak ptr=%p bytes=%zu purpose=%s\n",
```
lgritz: I would prefer use of OSL::print (which is an alias for OIIO::print and uses the same style as std::print/std::format) and no more use of old printf style.

lgritz:
Hey, @aconty, there were some other suggestions made on your draft other than the unified memory which you addressed.

Did you want to tackle any of those in this PR, or did you prefer to merge what we have and then come back to the other suggestions in subsequent rounds?

lgritz:
Note for onlookers: Alex told me offline that he's going to make some fixes before we merge.

aconty:
I believe I addressed everything.

testsuite/texture-device/closed_hashmap.h
```diff
@@ -0,0 +1,372 @@
...
+
+#define OPT_CONSTRUCT(condition) \
+    template<bool __C = condition, typename = std::enable_if_t<__C>>
+
+#define OPT_FIELD(condition, field_type) \
+    std::conditional_t<condition, field_type, std::monostate>
+
+template<class Key, class Value, class Hash, typename Arena,
+         typename ManagedArena = NullArena>
+class ClosedHashMap {
```
lgritz, line 28: Not gonna hold this PR up, but to put somewhere on the list of things to do in the future: Maybe this file (i.e. the ClosedHashMap class) can get some comments explaining its principles of operation and when one would use it? I can imagine a future in which this class gets promoted out of just the private texture_device enclave if there are wider uses for it.

testsuite/texture-device/vector_lite.h
lgritz: Let's remember that this is still on the to-do list.

lgritz:
I'm going to merge this so that we can continue to refine it with smaller PRs. (At some point it gets hard to look at and understand what specifically has changed.)

There are still some issues on the to-do list, but we can tackle them as this code matures and migrates from just a prototype implemented as a testsuite item to being an actual feature in the library.

PR #5230, opened 2026-06-12
build: Support C++26, and test it in CI in the bleeding edge test

lgritz:
Bump the "bleeding edge" test to C++26 (from 23), and the "latest releases" Linux test to C++23 (from 20). There are still other variants that test C++20 including anything based on the VFX2026 platform.

Just one tiny change needed to compile in C++26 mode, related to a deprecated std type_traits template.

### DIFF
--- a/src/include/OpenImageIO/strongparam.h
+++ b/src/include/OpenImageIO/strongparam.h
@@ -107,7 +107,9 @@ template<typename Tag, typename Basetype> struct StrongParam {
 
 private:
     Basetype m_val;
-    static_assert(std::is_trivial<Basetype>::value, "Need trivial type");
+    static_assert(std::is_trivially_copyable_v<Basetype>
+                      && std::is_trivially_copyable_v<Basetype>,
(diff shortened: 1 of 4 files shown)


src/include/OpenImageIO/strongparam.h
```diff
@@ -107,7 +107,9 @@ template<typename Tag, typename Basetype> struct StrongParam {
 
 private:
     Basetype m_val;
-    static_assert(std::is_trivial<Basetype>::value, "Need trivial type");
+    static_assert(std::is_trivially_copyable_v<Basetype>
+                      && std::is_trivially_copyable_v<Basetype>,
```
kmilos: Copypasta?

lgritz: um... must be

lgritz: updated

lgritz:
Any objections?

PR #5232, opened 2026-06-13
admin: Update security alerts and clarify security policy

lgritz:
Also make our CONTRIBUTING guide link to, and be consistent with,
our SECURITY.md policy file.

Follow OpenEXR's lead by explaining briefly that we only consider
something to be a security issue if it can plausibly be triggered by
ordinary inputs coming from untrusted parties. For example, a
maliciously crafted input image may be a vulnerability, but incorrect
use of an API call is not, because only trusted parties should have
direct access to the APIs.


grdanny (approved):
lgtm

PR #5234, opened 2026-06-13
cleanup: Get rid of some old files that have no use

lgritz:
python_tests was from 2010 and either was checked in inadvertently, or has long, long been supplanted by our real testsuite feature.

The various TRAVIS_SKIP_LSAN file used to mark certain tests as skippable when running the leak sanitizer... but we long ago removed the code in runtest.py that recognized its presence and took action. So it has served no purpose for quite some time.


grdanny (approved):
other than the ci fail (which was cause by me - sorry)
this lgtm

PR #5235, opened 2026-06-14
deps: auto-build pystring update to v1.2.0

lgritz:


aconty (approved):
Funny, I was fixing this in my own PR. AI fix?

lgritz:
What happened here is an interesting sequence:

A while back we put in a security fix where packages OIIO pulls down for CI need a commit hash, not just a tag or branch (to combat supply chain vulnerabilities involving setting tags of dependencies to malicious commits). Our build system checks that the tag it wants still has the hash we say it should.

Then we needed a fix in pystring that wasn't yet in a tagged release, so I said to use "master" and gave the commit hash of the current TOT that had Danny's fix. My intention was to switch it to a real tag as soon as the release happened.

But I forgot.

And all was fine, because main still pointed to that same commit whose hash I specified. Until yesterday, when Danny made merged more PRs and made a new release -- to pystring "master" was no longer the hash we said, oops.

Issue #4888, opened 2025-09-15
future feature: how to deal with multiple thumbnails?


lgritz:
ImageInput and ImageBuf() have `get_thumbnail()` methods. But they assume that there are 0 or 1 thumbnails per subimage. But some raw camera formats (and libraw) supports multiple reduced-resolution versions of the main image. We don't particularly have a way for an app to find out that there are multiple thumbnails, what their resolutions are, or to pick a specific one.

Maybe one possible route for the future is to extend the get_thumbnail() method to take a thumbnail index, to select which one is chosen.

Note that this would be a breaking change to the ImageInput API, so could not be backported to existing releases. (And in main, it would require making a v3_2 namespace version of ImageInput.)

jinhgkim:
Relevant discussion about choosing among multiple thumbnails: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5236#issuecomment-4713205618

jinhgkim:
A quick investigation shows:

|                    | 1 thumbnail     | Many thumbnails     |
| ------------------ | ------------------- | ------------------- |
| **1 subimage**     | TGA, JPEG | RAW     |
| **Many subimages** | EXR (1 thumb/subimage), PSD (1 thumb/file)    | TIFF, HEIF     |

which shows that some image formats can have many subimages, and each can carry its own set of reduced-resolution thumbnails.

### Scope:

This covers the **read** side: enumerating the thumbnails in a file and picking one.

### API Proposal:
```cpp
// ---- ImageInput ----------------------------------------------------
virtual int       nthumbnails(int subimage = 0) { return 0; }
virtual ImageSpec thumbnail_spec(int subimage = 0, int index = 0);
virtual bool      get_thumbnail(ImageBuf& thumb, int subimage, int index = 0);

// ---- ImageBuf ------------------------------------------------------
int       nthumbnails() const;              // from specs cached at init_spec()
ImageSpec thumbnail_spec(int index = 0) const;   // cached, no decode
std::shared_ptr<ImageBuf> get_thumbnail(int index = 0) const;  // lazy decode
```

The idea is to mirror the existing `spec_dimensions()` to query data for thumbnail selection without decoding.

### Example flow:
```text
  APP                    ImageBuf                       ImageInput
  ───                    ────────                       ──────────

  ImageBuf b(file)
      │   nothing happens yet — ImageBuf is lazy
      │
  b.nthumbnails() ─────► init_spec()   ◄── runs once, automatically
[...]

PR #5236, opened 2026-06-15
feat(oiiotool): Add thumbnail get/set commands and fix TGA thumbnail I/O

### LINKED ISSUE #4889: [FEATURE REQUEST] oiiotool thumbnail extraction
Not sure how this oversight happened, but there currently not an oiiotool command that extracts a thumbnail from an image. That would be handy.

For that matter, oiiotool also doesn't make it easy to create or add a thumbnail and save the resulting image. That is trickier, though, requires a bit more design maybe.

jinhgkim:
### Description

Adds `oiiotool` commands to extract and attach embedded thumbnails, and fixes TGA postage-stamp read/write bugs they exposed. 

Fixes #4889.

Note that thumbnail support is uneven across formats: `get_thumbnail` is implemented for PSD, camera RAW, and TGA, but `set_thumbnail` only for TGA. So you can read thumbnails from several formats, yet TGA is the only format that writes an embedded thumbnail today. e.g., `input.exr ... --thumbnail-set -o out.exr` silently produces an EXR with no embedded thumbnail.

- Features:
  - `oiiotool`: add `--thumbnail-get`
  - `oiiotool`: add `-i:get_thumbnail=1`
  - `oiiotool`: add `--thumbnail-set`
  
- Fixes:
  - targa thumbnail `write`: encode thumbnail like the main image (bottom-up, deassociated alpha, BGR(A))
  - targa thumbnail `read`: apply `associateAlpha()`
 

### Tests
New tests in `oiiotool-thumbnail-get`, `oiiotool-thumbnail-set`, `imagebuf_test`

### Checklist:

  behavior.
  on the GitHub runners.)


jinhgkim:
During the CI run, a TGA thumbnail test failed because I added `associateAlpha()` logic in this branch. My reasoning is that if we apply `associateAlpha()` in `TGAInput::readimg()`, we should apply the same operation when reading thumbnails.

Let me know what you think.

Also, fyi, I have broken down the commits into logical steps with some details in the commit message.

src/doc/oiiotool.rst
```diff
@@ -2339,6 +2351,50 @@ current top image.
...
+    Because this replaces the top image, use ``--dup`` beforehand if you also
+    want to keep the original.
+
+    Optional appended modifiers include:
+
+    `:fail=` *int* (default: 1)
+      If 1, it is an error if the image has no embedded thumbnail.
+      If 0, an empty (0x0) image is pushed in its place instead, so a batch
+      script can continue; guard any subsequent output (see the example
+      below), since writing the empty image is itself an error.
```
lgritz: Two things to consider here.

1. I'm not sure how resilient anything downstream will be to a literal 0x0 image. 
2. Does it make more sense for a "handled failure" to be either (a) just leave the original full-size image on the stack, or (b) perform the equivalent of a --resize or --resample to a reasonable thumbnail-like resolution?

src/oiiotool/oiiotool.cpp
```diff
@@ -7393,6 +7507,12 @@ Oiiotool::getargs(int argc, char* argv[])
     ap.arg("--flatten")
       .help("Flatten deep image to non-deep")
       .OTACTION(action_flatten);
+    ap.arg("--get-thumbnail")
+      .help("Extract an embedded thumbnail (options: fail=, index=)")
+      .OTACTION(action_get_thumbnail);
+    ap.arg("--set-thumbnail")
+      .help("Attach the top image as the thumbnail of the image below it")
+      .OTACTION(action_set_thumbnail);
```
lgritz: I'm not sure I feel strongly about this, but want to throw it out there in case it resonates with you. Would you prefer "--thumbnail-get" and "--thumbnail-set" for the sake of ensuring that they are always next to each other in an alphabetical listing of commands?

We purposely picked --mulc so that it would be next to --mul in alphabetical listings for similar reason.

src/targa.imageio/targainput.cpp
```diff
@@ -706,6 +635,95 @@ associateAlpha(T* data, int64_t size, int channels, int alpha_channel,
 
 
 
+bool
+TGAInput::get_thumbnail(ImageBuf& thumb, int subimage)
+{
+    if (m_ofs_thumb <= 0)
+        return false;  // no thumbnail info
+
```
lgritz, line 671: Can I ask why this function moved position in the file? That makes it hard to tell form the diffs exactly what changed. Or did it only move?

lgritz (commented):
This more or less LGTM. I didn't click approve only because I want to give you the chance to respond to some of the comments I left, none of which are strictly required, but maybe you'll particularly want to do some of those things.

lgritz:
Just another suggestion I thought of while reviewing this, so I'll mention it here. Though you need not add it, maybe it's an item for a follow-up, or just to put on the list of extensions to consider for some day (maybe only after `index=` works?).

Do you think it would be useful to have a `:requested=WxH` hint that, if more than one thumbnail were present, would pick the one closest to what you ask for? You could even imagine a few variants that extend it, like `:requested=<WxH` means the largest thumbnail  that fits in WxH, `:requested=>WxH` means "the smallest thumbnail at least that size" and `:requested=WxH!` means "I really need that size exactly, pick the best but also resize/fit if you need to."

src/oiiotool/oiiotool.cpp
jinhgkim: That makes sense to me. Addressed in 470541cef0ef9bfbd3ed2352500bf93892bd9c1e.

src/doc/oiiotool.rst
jinhgkim: Addressed in c0563674dd466415e47738ae643ebf706623b302

src/targa.imageio/targainput.cpp
jinhgkim: It has been relocated below `associateAlpha()` so it can call it without a
forward declaration. Also, added these lines: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5236/changes/4bfe52948c10f9c083e00fb396a11e73081c68ba#diff-9b6bbf7aa737fc8081351ed0ede0476a6b903a780e55f1b9f92f7ea006f46549R726-R745 which basically copies what `readimg()` does in the same file.

jinhgkim:
@lgritz, sorry for the delayed response. Hopefully this is still fresh enough that you don't need to re-review from scratch. (I've pushed three new commits)

> Just another suggestion I thought of while reviewing this, so I'll mention it here. Though you need not add it, maybe it's an item for a follow-up, or just to put on the list of extensions to consider for some day (maybe only after `index=` works?).
> 
> Do you think it would be useful to have a `:requested=WxH` hint that, if more than one thumbnail were present, would pick the one closest to what you ask for? You could even imagine a few variants that extend it, like `:requested=<WxH` means the largest thumbnail that fits in WxH, `:requested=>WxH` means "the smallest thumbnail at least that size" and `:requested=WxH!` means "I really need that size exactly, pick the best but also resize/fit if you need to."

Yes, that would be useful. It might also be worth allowing a single number, `:requested=200` meaning "the thumbnail whose shorter/longer side is closest to 200."

lgritz:
@jinhgkim make sure you see #5334 and weigh in on any issues there.

src/doc/oiiotool.rst
```diff
@@ -1232,6 +1232,18 @@ Reading images
         to following the input with a `--ch` command, except that by integrating
         into the `-i`, it potentially can avoid the I/O of the unneeded
         channels.
+      `:get_thumbnail=` *int*
+        If nonzero, read the file's embedded thumbnail instead of its main
+        image (equivalent to a following `--thumbnail-get`). The `:fail=` and
+        `:index=` modifiers are forwarded (e.g. `-i:get_thumbnail=1:fail=0`),
```
lgritz, line 1238: I'm not sure I understand what "index" does here?

src/doc/oiiotool.rst
```diff
@@ -2339,6 +2351,50 @@ current top image.
...
+      box, preserving aspect ratio and never enlarging, so a 512x384 image
+      yields a 256x192 thumbnail. Generating reads the image's pixels;
+      retrieving an embedded thumbnail does not.
+
+    `:index=` *int* (default: 0)
+      Selects which embedded thumbnail to retrieve, for formats that can
+      store more than one (such as some camera raw formats). Currently only
+      the primary thumbnail (`index=0`, the default) is available; a nonzero
+      value is an error until multiple-thumbnail support is added (see issue
+      #4888).
```
lgritz, line 2375: Ah, this is the "index". Just to clarify, are you talking about the subimage index (if it's a multi-subimage image at the top of the stack), or are you talking about the index of the thumbnail in an image that allows for multiple thumbnails?

Assuming you mean the multiple thumbnail case, are you just reserving a notation here, or do you have an idea for implementation? As far as I know, we have not outfitted any of the get_thumbnail() methods (of ImageInput or ImageBuf) with a way to specify one of potentially multiple thumbnails. We -- perhaps shortsightedly -- assumed that there would be only one, but in the case of camera raw images, that seems to not be the right assumption.

We don't have to solve it now; it is ok to just reserve a notation and try to figure out down the road how to actually implement it in the lower levels.

lgritz: @jinhgkim If this is not functional yet, I think it deserves a note here that the index modifier is not expected to work and is merely reserved for future expansion.

lgritz:
@jinhgkim I think this is about ready to go. I just had one outstanding comment, which is that if the `:index=` modifier is not currently functional, that should be noted in the docs.

src/doc/oiiotool.rst
jinhgkim: > Currently only the primary thumbnail (`index=0`, the default) is available; a nonzero
      value is an error until multiple-thumbnail support is added (see issue
      #4888).
      
 Is this good or should I make it more obvious?

lgritz: Oh, sorry, I think I did not read carefully. It's fine as it is.

lgritz (approved):
LGTM, thanks for this PR. And timely, Anton can build on this for his work with raw.

PR #5246, opened 2026-06-20
deps: New PugiXML min 1.11, rely on cmake config now, bump vendored version

lgritz:
Upgrade the embedded/vendored pugixml to 1.16, also the default version that we use for CI with build_pugixml.bash.

Raise PugiXML minimum to from 1.8 to 1.11 (late 2020, over 5 years ago).

Rely completely on pugixml exported cmake config, eliminate our Findpugixml.cmake. This is possible because we raised our minimum pugixml to 1.11, which is the point where its exported cmake configs were reliable.


lgritz:
Any objections? (Bumps pugixml minimum, but only for main / 3.2)

PR #5247, opened 2026-06-20
build: openjph auto-build and better exception handling

lgritz:
Add scripts for openjph auto-build when not found (and OpenImageIO_BUILD_MISSING_DEPS specifies it, or "all"). Build openjph 0.30.1 when missing. 

Since we can build it when missing, take away the various missing openjph allowances from our CI; auto-build when needed for all variants.

More thorough exception catching for openjph in our jpeg2000input.cpp.

When I compile openjph with clang (including icx), I get some runtime exceptions.  Not when I build it with gcc! I don't have the time or inclination to investigate why, so for our CI, rig it to build openjph with gcc even when I'm building OIIO with clang/icx.


lgritz:
Opinions?

lgritz:
I will merge this on Monday if there are no objections or comments.

PR #5249, opened 2026-06-20
fix(cineon): more robust to invalid numbers of channels & bit depth

lgritz:
Guard against corrupted images, which if they said they had more than the 8 channels allowed by Cineon format, would overrun array bounds while reading the file. And check for valid bit depths.

Also opportunistically fix a bug I spotted where a red channel was incrementing the count of greyscale channels rather than the count of red channels -- which must have been a copy and paste error from literally 15 years ago. (But those counters are only used for channel naming, and only if there is more than one greyscale or red channel in the file. So I doubt anybody has ever hit this bug, or ever would have in the future.)


lgritz:
withdrawing, there is more to fix

PR #5251, opened 2026-06-21
fix(iff): Fix allocation bug when reading 16 bit RGBA + float z

lgritz:
We were allocating a temporary buffer big enough for the RGBA but neglecting the z. Only for the 16 bit case.


lgritz:
It's been a week without comment, merging this bug fix.

PR #5252, opened 2026-06-21
api(deepdata): widen merge_deep_pixels srcpixel to int64_t

### LINKED ISSUE #5242: bug: `srcpixel` param of `DeepData::merge_deep_pixels` should be `int64_t` (breaks ABI)
Found during the nodiscard attribute task.
There are a couple of places that need follow-up to #2363. 
The `srcpixel` param of `DeepData::merge_deep_pixels` should be widened from `int` to `int64_t`, but it breaks compatibility (fails the ABI check), so it is left as `int` for now.
Another `int` to `int64_t` change is included in #5241.

```cpp
void merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel);
```

- [deepdata.h:191](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c62e7ca9110c6921f86d3dac10a309e802262f12/src/include/OpenImageIO/deepdata.h#L191)
- [deepdata.cpp:1179](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/c62e7ca9110c6921f86d3dac10a309e802262f12/src/libOpenImageIO/deepdata.cpp#L1179)

luna-y-kim:
### Description
To avoid breaking ABI, the old int overload is kept, but its declaration is hidden behind `#ifdef OIIO_INTERNAL`. Downstream code only sees the int64_t version, but old binaries still work. The forwarding wrapper handles it until the next first-digit version changes.

Also, the Python binding is adjusted using a wrapper to resolve the overload ambiguity. While adding the wrapper, I noticed other bindings still took `int pixel`, so widened them to `int64_t` for consistency.

Follow-up to #2363
Fixes #5242

### Tests
N/A

### Checklist:

  behavior.
- [ ] **(N/A)** **I am sure that this PR's changes are tested in the testsuite**.
  on the GitHub runners.)

### DIFF
--- a/src/include/OpenImageIO/deepdata.h
+++ b/src/include/OpenImageIO/deepdata.h
@@ -188,7 +188,13 @@ public:
     /// Merge the samples of `src`'s pixel into this `DeepData`'s pixel.
     /// Return `true` if ok, `false` if the operation could not be
     /// performed.
+    void merge_deep_pixels(int64_t pixel, const DeepData& src,
+                           int64_t srcpixel);
+
+#ifdef OIIO_INTERNAL
+    // DEPRECATED(3.2): use the version with int64_t
     void merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel);
+#endif
 
     /// Return the z depth at which the pixel reaches full opacity.
     float opaque_z(int64_t pixel) const;
--- a/src/libOpenImageIO/deepdata.cpp
+++ b/src/libOpenImageIO/deepdata.cpp
@@ -1167,7 +1167,8 @@ DeepData::merge_overlaps(int64_t pixel)
 
 
 void
-DeepData::merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel)
+DeepData::merge_deep_pixels(int64_t pixel, const DeepData& src,
+                            int64_t srcpixel)
 {
     int srcsamples = src.samples(srcpixel);
     if (srcsamples == 0)
@@ -1201,6 +1202,15 @@ DeepData::merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel)
 
     // Now merge the overlaps
     merge_overlaps(pixel);
+}
+
+
+
+// DEPRECATED(3.2): use the version with int64_t
+void
+DeepData::merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel)
+{
+    merge_deep_pixels(pixel, src, int64_t(srcpixel));
 }
 
 
--- a/src/python/py_deepdata.cpp
+++ b/src/python/py_deepdata.cpp
@@ -8,7 +8,7 @@ namespace PyOpenImageIO {
 
 
 void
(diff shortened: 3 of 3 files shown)


src/include/OpenImageIO/deepdata.h
```diff
@@ -188,7 +188,8 @@ class OIIO_API DeepData {
     /// Merge the samples of `src`'s pixel into this `DeepData`'s pixel.
     /// Return `true` if ok, `false` if the operation could not be
     /// performed.
-    void merge_deep_pixels(int64_t pixel, const DeepData& src, int srcpixel);
+    void merge_deep_pixels(int64_t pixel, const DeepData& src,
+                           int64_t srcpixel);
```
lgritz, line 192: Can you avoid the ABI break by just not removing the old one at all? Just add the new one, and change the implementation of the old one to call the new one. Leave a comment on the old one indicating that we intend to deprecate it. So the idiom is like this:

```
    /// full comments
    void func(int64_t);

    // DEPRECATED(3.2): use the version with int64_t
    void func(int);
```

does that work? Or do you just get warnings everywhere about ambiguous arguments?

lgritz: Wait, I think this is even better:

```
    /// full comments
    void func(int64_t);

#ifdef OIIO_INTERNAL
    // DEPRECATED(3.2): use the version with int64_t
    void func(int);
#endif
```

This hides the declaration from external/downstream clients of OIIO when they recompile. They will only see the int64_t version, which if they pass an int32, will just transparently accept it. So starting from their next recompile, they will start using the new function, without having to do anything.

Meanwhile, *internally* we still declare and implement the old function. Which means that ABI is preserved, the library exports all the same symbols it used to, and an existing program that calls the old function will continue to link properly against a new copy of libOpenImageIO.

When we eventually have OIIO 4.0, we'll chase down all those "DEPRECATED" comments and finally remove the functions entirely (because we don't guarantee any compatibility at first-digit version changes).

lgritz: (I should point out that it's only because the change is specifically int -> int64_t, which is a widening that is implicitly accepted by the compiler without warning, that we are able to do this trick and something like `func(my_int)` will just work and know to call `func(int64_t)`. If we were changing int to uint64_t, or doing some other change that wouldn't just implicitly cast, we'd be in a more complex situation.)

luna-y-kim: So it silently rebinds downstream code to the new int64_t overload with no changes on their side! (Of course because this is the `int` -> `int64_t` case as you mentioned.) Thanks, it's a cool tip!

> does that work? Or do you just get warnings everywhere about ambiguous arguments?

It works! No ambiguity at the call sites. The Python binding needed a small change due to the ambiguous overload. I fixed it with a wrapper as the other overloaded functions already do.

lgritz (approved):
Yeah, looks like it works great!
LGTM.

lgritz:
I'm going to rebase this and watch the CI before I merge, just in case it interferes in any way with the other deep related patch I just merged from you.

PR #5255, opened 2026-06-26
docs: fix typo in bitcast docs

lgritz:


lgritz:
Only changing an assertion message, merging.

PR #5258, opened 2026-06-26
fix(rla): better detection of corrupted RLE encoding

lgritz:
Addressing a buffer overrun when the RLE count exceeds the number of encoded bytes left to read from.


lgritz:
This one has been sitting around for a while. Objections?

lgritz:
Fully two weeks without comment -> merging

PR #5259, opened 2026-06-26
fix(psd): identify corruptions of layer resolutions, eof in strings

lgritz:
* Find corruptions by validating layer resolutions.
* Find corruptions when eof is hit in the middle of a string.


lgritz:
Getting stale. Any opinions or objections?

lgritz:
No objections for two weeks -> merging

PR #5263, opened 2026-06-27
build: fix certain clang varieties warning about giflib headers

lgritz:
Assisted-by: Claude Code / sonnet-4.6


lgritz:
This has gone without comment for well over a week, and is only a build warning fix. Merging.

PR #5264, opened 2026-06-27
fix(ico): Better error checking for PNG-in-ICO files

lgritz:


jessey-git:
I'm a little worried about this one.  Seems that many programs out there have no issue opening larger .ico files, at least with the few I've tried (and when using the PNG embedding). That probably means there's a de-facto expectation that such things are "supported".

lgritz:
OK, that's a fair observation. I wasn't aware of that.

I'll give it full PNG limits, then.

lgritz:
Amended

jessey-git (approved):

PR #5265, opened 2026-06-27
ptex: More comprehensive valid_file and validation of header

lgritz:
Previously, we did not have a PtexInput::valid_file() method, and our open() method just directly called PtexTexture::open(), which is really bad with validity checks and can crash or OOM with corrupted files.

But Ptex has a fixed 64 byte header (cheap to read and inspect) that has lots of fields that let us more accurately check for reasonable file validity. Therefore, we both implement valid_file, and also have open() do these checks (rejecting many forms of corruption) before turning the job over to the Ptex library itself.

### DIFF
--- a/src/ptex.imageio/ptexinput.cpp
+++ b/src/ptex.imageio/ptexinput.cpp
@@ -5,6 +5,7 @@
 #include <Ptexture.h>
 
 #include <OpenImageIO/dassert.h>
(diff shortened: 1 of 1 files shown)


lgritz:
Amended with a much more comprehensive approach, and updated the title and description.

lgritz:
Over a week. Any comments?

PR #5266, opened 2026-06-27
webp: Improve valid_file to check the header for magic words

lgritz:
This is MUCH faster than asking libwebp to open the file!

### DIFF
--- a/src/webp.imageio/webpinput.cpp
+++ b/src/webp.imageio/webpinput.cpp
@@ -15,6 +15,15 @@ OIIO_PLUGIN_NAMESPACE_BEGIN
 namespace webp_pvt {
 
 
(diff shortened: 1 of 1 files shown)


jessey-git:
Here we are giving a 64byte header to WebpGetInfo to parse through. Is that so much slower than doing a 12byte read here? Is WebP going off the rails validating that 64bytes?

lgritz:
I found at least some cases where this kept webp from going off the rails.

jessey-git (approved):
Sounds like an issue that would be interesting to file against libwebp. Extreme resource usage for small inputs can be classified as a security issue sometimes.  But I also can't look that closely right now either.  So this change is fine, just a little unfortunate we can't use someone else's code to validate.

PR #5267, opened 2026-06-27
perf: less redundant opening when files fail

lgritz:
When creating an ImageInput, if the initial guess (based on the file extension) fails to open, we try all the others in turn.

This tries to make that a little cheaper by not doing a full open() but instead first checking for valid_file, which in theory is a much less expensive approximate check.


jessey-git (approved):
Yeah, the original code here always struck me as a odd to only do this for the do_open case.

PR #5268, opened 2026-06-27
fix(iff): detect corrupt chunk sizes, flags, channel configs

lgritz:
Detect and reject corrupt chunk sizes, flags, and channel configurations in the header.


lgritz:
I have amended this PR to handle an additional related case.

mikaelsundell (approved):
LGTM. Z-only Maya depth files may exist and could be handled in a separate PR with a real .z sample and dedicated tests. Let me see what I can do about that.

lgritz:
Did we support it before?

If we never supported it properly, let's let this stand and we'll come back to it.
But if this PR is breaking z-only support, I can try to rework the PR.

mikaelsundell:
My code was broken, I suggest we move on, let me prepare a new PR
on top of this to properly deal with .z iff files?

lgritz:
OK, but actually, maybe wait until I get some more PRs submitted. I have more iff ones in flight, so let's let the security bug fixes settle down before adding support for things that never worked. If it didn't work before and nobody complained, there can't be much of a rush.

mikaelsundell:
Thats cool, it may have worked but never properly tested or validated. Need to create a sample file in Maya so no stress 👍🏻

lgritz:
CVE-2026-63419

PR #5270, opened 2026-06-28
fix(jpeg2000): reject corrupt component geometry and subsampling

lgritz:
A fuzzed JPEG2000 caused a heap-buffer-overflow (caught by ASan) when copy_scanline read past the end of openjpeg's decoded component buffer. The file's image canvas was 32x32, but one of its three components had a horizontal subsampling factor dx=249 (larger than the entire image). OIIO derives the ImageSpec size from the union of each component's x0+w*dx / y0+h*dy window, so that bogus component inflated the spec to 249x32. copy_scanline then iterated x over the inflated width while indexing a component whose data array only held 1 column, walking off the end. The per-sample guard was also broken: it compared the component row index against reference-grid offsets (wrong units) and used `>` instead of `>=` on the column.

Two fixes:

1. Detect the corruption at open(): the JPEG2000 canvas (x1-x0, y1-y0) is the authoritative image size, and a component's subsampling factor must be nonzero and no larger than the canvas. Reject files that violate this instead of decoding a nonsensical, inflated image.

2. Harden copy_scanline by bounds-checking each access against the actual component array geometry and skip the channel (zero-fill) when comp.data is null or a subsampling factor is zero, which also closes a latent divide-by-zero on comp.dx/comp.dy.

### DIFF
--- a/src/jpeg2000.imageio/jpeg2000input.cpp
+++ b/src/jpeg2000.imageio/jpeg2000input.cpp
@@ -581,11 +581,33 @@ Jpeg2000Input::open(const std::string& name, ImageSpec& p_spec)
         return false;
     }
 
+    // The JPEG2000 canvas (x0,y0)-(x1,y1) is the authoritative image size.
+    // Each component samples that canvas at its own subsampling factor.
+    const unsigned int canvas_w = (m_image->x1 > m_image->x0)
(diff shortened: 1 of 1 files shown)


lgritz:
Any comments?

lgritz:
Two weeks, no comments -> merging

PR #5271, opened 2026-06-28
fix(dpx): detect corrupt userbuf size, overflow guard

lgritz:
- Detect corrupted userbuf size.
- Protection against integer multiply overflow in various places in the DPX reading code.
- Also fix memory load alignment issue. Didn't seem to be a problem in practice, but the sanitizers weren't happy on some platforms.

Assisted-by: Claude Code / sonnet-4.6


jessey-git (approved):
Looks good.

PR #5272, opened 2026-06-28
fix(build): fix EMBEDPLUGINS=0, make many build options env-override

lgritz:
Change several build-time options from `option()` to our `set_option()`, which mostly has the effect of allowing them to be overridden by environment variables of the same name, which is really handy for CI variants.

In the process, realized that we had some real problems with handling of EMBEDPLUGINS=0, it probably hasn't worked properly for years and went unnoticed because the one CI job variant that was trying to test EMBEDPLUGINS=0 actually *was not* because it was trying to do it by setting the environment variable, unaware that it could not be used to set the option until this patch. So there is a bit of cmake and linkage fixing that needed to be done to get it squared away.


lgritz:
Any concerns here?

PR #5273, opened 2026-06-28
webp: Improve valid_file to check the header for magic words (5266)

lgritz:


jessey-git:
What is the context behind this change? The current call to WebPGetInfo is a more complete check I think, albeit a tiny bit slower. Is this mainly for performance?

jessey-git:
Oh, this is a duplicate of PR #5266

lgritz:
Oops, did I submit this twice?

lgritz:
Yeah, looks like I doubled this one up. Will close this one.

PR #5274, opened 2026-06-28
fix(bmp): validate scanline file position before reading

lgritz:
read_native_scanline computed the scanline's file position as m_bmp_header.offset + y * m_padded_scanline_size and seeked there without checking it against the file size. A corrupt header (bogus pixel-data offset or scanline size) can place that position far outside the file.


jessey-git (approved):
Seems sane to do.

PR #5275, opened 2026-06-28
fix(raw): apply LibRaw memory cap before unpack to prevent OOM

lgritz:
open_raw set imgdata.rawparams.max_raw_memory_mb *after* calling m_processor->unpack(), so LibRaw's allocation cap never protected the unpack step -- which is exactly where LibRaw allocates the raw image buffer. A corrupt raw file claiming huge dimensions could therefore allocate unbounded and OOM (showing up on the fuzzer as a 0-length crash artifact from an out-of-memory trip rather than a normal reproducer).

Move the max_raw_memory_mb assignment to immediately after open_file succeeds, before unpack(), so the cap (default 2048 MB) actually bounds the allocation. LibRaw then fails the file cleanly instead of OOMing.

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -431,6 +431,13 @@ RawInput::open_raw(bool unpack, bool process, const std::string& name,
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


grdanny (approved):
looks reasonable to me

PR #5277, opened 2026-06-29
build(tiff): recommend libtiff 4.5+ for security fixes

lgritz:
Circular IFD chain loop detection (identified by for CVE-2022-3970) was fixed in libtiff by MR 386
(https://gitlab.com/libtiff/libtiff/-/merge_requests/386), released in libtiff 4.5.0.

Raise the recommended minimum from 4.2 to 4.5 so users on older installations get a build-time advisory. The hard minimum stays at 4.0 (for now) to preserve compatibility.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -92,8 +92,8 @@ checked_find_package (libuhdr
 
 checked_find_package (TIFF REQUIRED
                       VERSION_MIN 4.0
-                      RECOMMEND_MIN 4.2
-                      RECOMMEND_MIN_REASON "4.2 for GPS support")
+                      RECOMMEND_MIN 4.5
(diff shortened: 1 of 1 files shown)


jessey-git (approved):
I have no qualms about gently nudging the ecosystem forward, so +1 from me.

I was going to suggest the same for `libheif` but there's literally no good version to reference except maybe the very latest one, which isn't too useful for us here.

lgritz:
I wanted to backport this patch, so I only wanted to raise the recommended min.

I think in a follow-up I will raise the hard min, but I wasn't yet sure how high, and also that would not be backportable to the current release branches.

lgritz:
Before it's time to branch and beta 3.2 (which is really only in another month or so), we'll do a double check across all the dependencies and see what we can raise.

PR #5278, opened 2026-06-29
ci: Address disabled old node support on aswf-old jobs for OIIO 3.0

lgritz:
THIS IS A 3.0 PATCH ONLY, backporting components from 3.1 to cover yet another GitHub change.


lgritz:
I will merge this immediately, since it doesn't change code -- it is only changing CI, and quite obviously taking it from failing to passing.

PR #5279, opened 2026-06-29
fix(sgi): size RLE offset tables by spec dims to prevent OOB on corrupt headers

lgritz:
read_offset_tables() sized the RLE start/length tables from the raw header (`ysize * zsize`), but read_native_scanline() indexes them by the spec dimensions (`height * nchannels`, derived from the image's dimension field). A corrupt header (e.g. dimension=1 with zsize=0) makes `ysize*zsize` = 0 so the tables are empty, while indexing still uses `height*nchannels` = 1 -- a null-pointer dereference and crash.

Size the tables by `m_spec.height * m_spec.nchannels`, the same quantity used to index them, so the two can never desync. For well-formed files this equals ysize*zsize, so behavior is unchanged; a corrupt file now fails cleanly when the (too-short) table read hits EOF. Also guard against non-positive table sizes, use vector::data() instead of `&v[0]` (UB on empty), and fix an off-by-one in the scanline bounds check (y == height would index start_tab[-1]).

Add the fuzzer reproducer as testsuite/sgi/src/broken02.sgi.


jessey-git (approved):
Looks good.

PR #5283, opened 2026-07-01
fix(cineon): validate bit depth against libcineon's supported set

lgritz:
CineonInput::open() accepted any per-channel bit depth in [1,32], but libcineon's ComponentDataSize()/ComponentByteCount() only recognize {8,10,12,16,32,64} and assert(0) on anything else. A crafted file with an unsupported bit depth (e.g. 26) passed the OIIO-side check and then aborted deep inside the vendored library instead of failing cleanly.

Tighten the check to libcineon's actual whitelist so bogus bit depths are rejected with a proper errorfmt() before ever reaching the vendored code. Adds a regression fixture (broken_bitdepth2.cin) to the cineon testsuite.


grdanny (approved):
looks reasonable to me.

lgritz:
CVE-2026-63638

PR #5284, opened 2026-07-02
fix(iff): reject implausible image dimensions before allocating tile buffer

lgritz:
TBHD width/height/channel fields were never cross-checked against how much tile data the file actually contains, so a corrupt header could claim an arbitrarily large image (e.g. 16k x 16k) backed by only a few hundred bytes of real data. readimg() would then resize() a multi-gigabyte buffer for a tiny malformed file, hanging (or very slowly succeeding) instead of failing -- a resource-exhaustion DoS on untrusted input, found via fuzzing.

Bound the claimed image size against how much tile data remains in the file, generously allowing for RLE's worst-case 128x expansion ratio, and fail fast if it's implausible.

### DIFF
--- a/src/iff.imageio/iffinput.cpp
+++ b/src/iff.imageio/iffinput.cpp
@@ -3,6 +3,8 @@
 // https://github.com/AcademySoftwareFoundation/OpenImageIO
 #include "iff_pvt.h"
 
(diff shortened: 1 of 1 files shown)


grdanny (approved):
looks reasonable to me

PR #5285, opened 2026-07-02
fix(iff): add resolution validity checking

lgritz:


grdanny (approved):
64k sounds like a reasonable limit here...

PR #5286, opened 2026-07-02
fix(dds): reject corrupt bytes-per-pixel before it forces a huge allocation

lgritz:
A DDS file with a garbage pixel-format bpp field (e.g. fmt.bpp ~2.5e9) yields a native bytes-per-pixel (m_Bpp) of hundreds of millions. internal_readimg() then does `new uint8_t[width * m_Bpp]`, a multi-GB allocation that OOMs. The core resolution/imagesize limits do not catch this: they only see the small output ImageSpec, not the native pixel size, and the existing fmt.bpp check is gated on RGB/LUM/YUV/ALPHA flags that a corrupt header can simply omit.

Validate m_Bpp for uncompressed formats right after it is computed in open(): the largest real uncompressed DDS/DXGI pixel is 128-bit (16 bytes), so anything outside [1,16] is a corrupt header and is rejected before any allocation. The check is confined to uncompressed formats because m_Bpp is neither set nor used for compressed (block) formats, which size their data via GetStorageRequirements.


jessey-git (approved):
Looks good.

PR #5288, opened 2026-07-02
fix(psd): detect implausibly large ICC, Exif, XMP before allocating

lgritz:
10MB was conservatively chosen as big enough that anything larger is definitely bogus for an ICC, Exif, or XMP block in a PSD file.

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -1250,6 +1250,17 @@ PSDInput::load_resource_1036(uint32_t length)
 bool
 PSDInput::load_resource_1039(uint32_t length)
 {
+    if (length > 10 * 1024 * 1024) {
+        if (OIIO::get_int_attribute("imageinput:strict")) {
+            errorfmt(
+                "Implausibly large ICC profile ({} bytes), presumed corrupted file\n",
+                length);
+            return false;
+        } else {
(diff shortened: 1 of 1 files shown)


lgritz:
I wrote this code with my own hands and brain (hence, no tool disclosure), but I did ask Claude what a reasonable limit was, and got this hilarious conversation / cautionary tale:

Claude:
> OpenImageIO's current limit (since you work on it) is 50 MB, which I believe was chosen as a round number well past any real profile while still catching runaway reads from corrupt data.
> If the context is defense against malicious or corrupt input, 10 MB is the number I'd use and could defend — anything beyond that is either a bug or an attack.

LG:
> why do you say OIIO's limit for an ICC profile is 50MB?

Claude:
> You're right to challenge that — I stated it with false confidence. I don't actually know what OIIO's limit is; I was fabricating a specific number. I shouldn't have done that, especially framed as something "you'd know."

jessey-git (approved):
Looks good, though 10mb seems high still. But I don't have a good corpus of real files to check what might be better limit either.

lgritz:
It's really hard to tell how big a legit one could ever be. Maybe 10MB is too generous, but it's still small enough that it's not going to lead to OOM or lead to some long pause initializing an array or whatever.

PR #5289, opened 2026-07-02
fix(ico,png): avoid leaking error string across png_chunk_error longjmp

lgritz:
PngReadCallback passed input->geterror(false).c_str() to png_chunk_error() when a PNG read failed. png_chunk_error() does not return -- it longjmp's back to libpng's setjmp point, which skips C++ destructors, so the temporary std::string returned by geterror() was never freed. The fuzzer surfaced this as a small, intermittent leak (a few KB across hundreds of thousands of runs).

Copy the message into a plain stack buffer (no destructor) before calling png_chunk_error, so nothing heap-allocated is alive across the longjmp.

### DIFF
--- a/src/ico.imageio/icoinput.cpp
+++ b/src/ico.imageio/icoinput.cpp
@@ -84,7 +84,13 @@ private:
         OIIO_DASSERT(icoinput);
         if (!icoinput->ioread(data, length)) {
             icoinput->m_err = true;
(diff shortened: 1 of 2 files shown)


jessey-git (approved):
Seems fine.  I'll note that libpng has an (AFAICS undocumented and unexposed) #define that limits the passed in error message to 196(yes 6) characters. But from reading their code it's fine if our buffer is larger, the message will just be truncated if necessary (not that our error message is likely to get that large).

PR #5292, opened 2026-07-03
fix(gif): avoid int32 overflow in palette-split pixel-count math

lgritz:
GifSplitPalette() computed `numPixels * (splitElt - firstElt)` as a 32-bit int before dividing. For any image over roughly 16.9M pixels (e.g. 4117x4117), that intermediate multiplication overflows INT_MAX and is undefined behavior; in practice it produced a bogus negative/garbage split count that could drive later indexing into `image` out of its bounds.

Do the multiply in 64 bits and narrow back afterward -- the quotient is always <= numPixels, so the narrowing itself is safe.

### DIFF
--- a/src/gif.imageio/gif.h
+++ b/src/gif.imageio/gif.h
@@ -340,7 +340,8 @@ void GifSplitPalette(uint8_t* image, int numPixels, int firstElt, int lastElt, i
     if(bRange > gRange) splitCom = 2;
     if(rRange > bRange && rRange > gRange) splitCom = 0;
 
-    int subPixelsA = numPixels * (splitElt - firstElt) / (lastElt - firstElt);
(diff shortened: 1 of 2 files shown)


jessey-git:
Given the max dimensions of 32768x32768x4, there are 4 other int overflows related to gif input, might as well do them in the same PR:
For `gif.h`
- The `GIF_MALLOC` calculation inside GifBegin
- The `imageSize` calculation inside GifMakePalette

For `gifinput.cpp`
- The `memcpy` calculation inside GIFInput::read_native_scanline
- The `idx` calculation inside GIFInput::read_subimage_data

lgritz:
So amended.

jessey-git (approved):
Looks good, but watch out for clang-format.

PR #5293, opened 2026-07-03
fix(targa): reject implausible image dimensions before allocating

lgritz:
An 18-byte TGA file consisting of just a header (no pixel data) can claim an arbitrary width/height/bpp -- e.g. a 1472x12000 24bpp image -- and readimg() would allocate the full ~53MB buffer and only discover the truncation once the first read hits EOF. Worse, the resulting partially-formed ImageSpec could still propagate to a downstream writer before the read failure was acted on.

Cross-check the header's width/height/bpp against how many bytes of pixel data the file could plausibly contain before committing to the allocation, mirroring the existing pattern in iff.imageio and jpeg2000.imageio. The check compares against the *on-disk* pixel size (palette/16-bit storage is smaller per pixel than the decoded, unpacked ImageSpec) and allows up to 128x expansion for RLE-compressed images (a 1-byte run-length count can encode up to 128 output pixels).

Updated testsuite/targa/ref/out.txt: several existing corrupt-file regression cases (crash1.tga, crash6.tga, crash3952.tga) are now caught earlier with clearer messages, and two real oiio-images TGA files (UCM8.TGA, UTC16.TGA) that use palette/16-bit storage smaller than their decoded size continue to read correctly.

### DIFF
--- a/src/targa.imageio/targainput.cpp
+++ b/src/targa.imageio/targainput.cpp
@@ -254,6 +254,34 @@ TGAInput::open(const std::string& name, ImageSpec& newspec)
     if (m_tga.type >= TYPE_PALETTED_RLE)
         m_spec.attribute("compression", "rle");
 
+    // The width/height/bpp fields are not cross-checked against how much
+    // pixel data the file actually contains, so a corrupt header can claim
+    // an arbitrarily large image (e.g. thousands of pixels square) backed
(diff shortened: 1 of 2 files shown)


jessey-git (approved):
Should be safe to do.

lgritz:
CVE-2026-65969

PR #5294, opened 2026-07-03
fix(tiff): avoid use-after-scope in multithreaded scanline reads

lgritz:
read_native_scanlines declared task_set before the locals its queued decompression lambdas capture (ok, compressed_scratch). On the leftover-scanline early-return path those locals were destroyed before ~task_set() waited for the queued work, so workers could write through a dangling ok and hand a freed compressed buffer to zlib.

Declare task_set after the captured locals so its destructor wait runs while they are still alive, and propagate async decompression failures via return ok instead of unconditionally returning true.

### DIFF
--- a/src/tiff.imageio/tiffinput.cpp
+++ b/src/tiff.imageio/tiffinput.cpp
@@ -2232,7 +2232,6 @@ TIFFInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
     // Make room for, and read the raw (still compressed) strips. As each
     // one is read, kick off the decompress and any other extras, to execute
     // in parallel.
-    task_set tasks(pool);
     bool ok        = true;  // failed compression will stash a false here
     int y          = ybegin;
     size_t ystride = m_spec.scanline_bytes(true);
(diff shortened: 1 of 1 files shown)


src/tiff.imageio/tiffinput.cpp
```diff
@@ -2330,13 +2335,17 @@ TIFFInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
     // If we have left over scanlines, read them serially
     m_next_scanline = y - m_spec.y;
     for (; y < yend; ++y) {
-        bool ok = read_native_scanline_locked(subimage, miplevel, y, data);
-        if (!ok)
-            return false;
+        // Note: a false return here leaves the loop (and function) before the
+        // explicit tasks.wait() below; ~task_set() still waits, and the
+        // declaration order above ensures the captured locals outlive it.
```
jessey-git: This comment is a bit confusing if reading this code without the benefit of this PR: "... a false return here" -- what false return? Yes, there was a "return false" before the PR but not now, and the phrase "false return" seems wrong somehow.  Is this a comment from the LLM as it modified the code?

lgritz: It was. Yeah, I see why it's awkward. It just means the "return false" right above. I will try to clean it up.

lgritz: Amended.

Thanks for calling these out. I'm scrambling so much on these vulnerability reports that I was not being as diligent as I should about rewriting everything, including comments, as I would have done them. The LLM is very good at finding the causes of these crashes and plugging the holes, but it does tend to make comments be a little too documentary about the specific conditions that led to the fix, instead of just replacing bad code with good and letting it stand mostly on its own.

jessey-git (approved):
Much better :)

lgritz:
CVE-2026-65970

PR #5295, opened 2026-07-04
fix(exr): use rectangle row stride for partial edge tile reads

lgritz:
read_native_tiles() on a tiled EXR whose dimensions aren't a multiple of the tile size wrote past the caller's buffer when reading an edge tile range: both readers used a row stride padded up to a whole number of tiles (nxtiles*tile_width) instead of the caller's actual rectangle width, overrunning the buffer whenever the last tile column is partial.

Fixed in both exrinput_c.cpp (Core reader) and exrinput.cpp (classic reader) to use the requested rectangle width as the destination row stride instead. exrinput.cpp additionally needs the pre-clamp width specifically, since read_native_tile() forwards a tile-aligned xend past the image edge for ordinary full-tile reads.

Adds testsuite/openexr-partialtile, a compiled regression test that calls read_native_tiles() directly for both readers -- the only path that exercises this code -- and checks pixels and buffer bounds.

### DIFF
--- a/src/cmake/testing.cmake
+++ b/src/cmake/testing.cmake
@@ -407,6 +407,9 @@ macro (oiio_add_all_tests)
                         IMAGEDIR openexr-images
                         URL http://github.com/AcademySoftwareFoundation/openexr-images)
     endif ()
+    # Regression test (compiles its own helper and generates its own image)
+    # for a partial edge-tile heap overflow in the OpenEXR readers.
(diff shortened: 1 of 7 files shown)


lgritz:
Any comments or concerns about this one?

lgritz:
Nearly two weeks old, no objections -> merging.

lgritz:
CVE-2026-63422

PR #5296, opened 2026-07-04
fix(tiff): fix heap overflow unpacking sub-8-bit contig CMYK samples

lgritz:
We had missed a case where bit_convert needed outputs to go to the scratch buffer: PHOTOMETRIC_SEPARATED, but not requesting raw color. (Because the real number of channels differ from what we are presenting to the caller.)


jessey-git (approved):
Looks ok to me. The `m_photometric == PHOTOMETRIC_SEPARATED && !m_raw_color` conditional is used in several other places already.

lgritz:
CVE-2026-67549

PR #5298, opened 2026-07-06
fix(dpx): fix heap overflow in 1-channel 10-bit filled scanline swap

lgritz:
Read10bitFilled() unpacks a scanline of 10-bit "filled" (Method A/B) samples backwards, and for 1-channel images applies a workaround that swaps the first and third datum of each group of 3 packed samples (so column order comes out right). The swap didn't check that a full group of 3 remained in bounds. Guard the swap so it only fires when a complete group of 3 remains.

Add a regression test in testsuite/dpx.


lgritz:
Any comments on this small fix?

lgritz:
No objections -> merging

PR #5299, opened 2026-07-06
fix(gif): guard canvas index overflow

br0nzu:
This handles malformed GIF files whose logical canvas dimensions can overflow legacy int-indexed canvas calculations.

The patch rejects oversized canvas dimensions before scanline/canvas indexing and adds a regression test for the malformed oversized-canvas case.


testsuite/gif/run.py
lgritz, line 9:
```suggestion

redirect = " >> out.txt 2>&1 "

files = ["gif_animation.gif", "gif_oiio_logo_with_alpha.gif",
```

I think that will capture the error output and fix your failing CI tests.

lgritz: I'm not sure if that diff came out exactly right, but I just mean to add that `redirect =` line.

lgritz (commented):
The code looks good! The tests are failing only because the output to stderr is not captured. (Does that mean you are running on Windows when you tested it yourself?)

lgritz:
Tip: In your own fork of OIIO, you can edit settings to enable Actions workflows on your repo. Then you can push a proposed fix to a branch of your fork, and CI will automatically run and you can iterate as needed to fix any problems. Then, you can post the PR and know that the version we see will pass CI right away.

br0nzu:
Thanks, fixed.

I added `redirect = " >> out.txt 2>&1 "` to `testsuite/gif/run.py` as suggested. I rechecked the corrupt-GIF regression locally; stderr is now captured in `out.txt`, and the existing `ref/out.txt` already matches the captured output.

lgritz (approved):
LGTM

PR #5300, opened 2026-07-06
fix(imageio): ignore invalid RowsPerStrip chunking

br0nzu:
This treats non-positive tiff:RowsPerStrip metadata as absent before using it to align generic scanline read/write chunks. That avoids feeding invalid chunking metadata into round_to_multiple while preserving normal positive RowsPerStrip behavior.

The patch adds regression coverage for an OpenEXR file carrying RowsPerStrip=0 through iinfo --hash, iinfo --stats, and an oiiotool conversion path.


lgritz:
Where does the negative rps come from in the first place?

The only place that is ever set is in tiffinput.cpp:

```
    if (!m_spec.tile_width) {
        TIFFGetField(m_tif, TIFFTAG_ROWSPERSTRIP, &m_rowsperstrip);
        if (m_rowsperstrip > 0) {
            // Only set the attrib if a legit value was found in the file
            m_spec.attribute("tiff:RowsPerStrip", m_rowsperstrip);
            m_rowsperstrip = std::min(m_rowsperstrip, m_spec.height);
        } else {
            // Default if not found is "one strip for the whole image"
            m_rowsperstrip = m_spec.height;
        }
    }
```

So it looks like it will only be set if it's positive.

OH, I see... it has to come in through an OpenEXR that cleverly includes metadata called "tiff:RowsPerStrip" set to a bad value.

testsuite/iinfo/run.py
```diff
@@ -30,3 +30,6 @@
 # Info for subimages and mips
 command += info_command ("--stats src/subimage.tif", info_program="iinfo")
 command += info_command ("--stats src/mip.tif", info_program="iinfo")
+command += run_app (pythonbin + " src/check-rowsperstrip-zero.py "
+                    + oiio_app("iinfo") + " " + oiio_app("oiiotool"),
+                    silent=True)
```
lgritz: I'm wondering if we really need an extra .py file (code to maintain) instead of just directly running the two iinfo and one oiiotool tests here like we would for any other test? Are all of them necessary, or is any one enough to test that this issue has not regressed?

And I'm wondering about whether the test for this fix really fits best here in the "iinfo" test (it's not really testing an iinfo feature in particular) or if there is a different testsuite directory where it makes more sense.

lgritz:
I took the liberty of pushing an amendment to your PR that makes a separate test (and somewhat simpler) header of putting it into the iinfo test. If you think that is ok, I am prepared to merge it.

br0nzu:
LGTM

lgritz (approved):

PR #5301, opened 2026-07-06
fix(ffmpeg): align swscale output buffers

br0nzu:
This aligns FFmpeg/libswscale destination image buffers and validates the buffer-size, frame-fill, and swscale return paths before decoded pixels are consumed.

The change prevents narrow high-bit-depth VP9/GBR inputs from writing past OIIO's destination buffer during swscale conversion, and adds a regression case for the width-1 high-bit-depth FFmpeg path.


lgritz (commented):
Code looks good!

Needs clang-format fix, just in one isolated spot

br0nzu:
Thanks, fixed.

lgritz (approved):
LGTM

PR #5302, opened 2026-07-07
ci: Add vfx2027 container tests

lgritz:
One gcc and one clang based test in the new containers.


jfpanisset (approved):
LGTM, hopefully it builds fine.

I want to tackle adding libheif soon so OIIO can exercise one more dependency.

PR #5305, opened 2026-07-07
fix(jxl): guard corrupt basic-info dimensions early

lgritz:
Run check_open on JXL basic-info dimensions/channels before decode output allocation to prevent oversized allocation paths on malformed input.

Also update ref output for libjxl 0.12.0.

Assisted-by: GitHub Copilot / GPT-5.3-Codex


src/jpegxl.imageio/jxlinput.cpp
```diff
@@ -283,6 +283,16 @@ JxlInput::open(const std::string& name, ImageSpec& newspec)
...
             m_channels = info.num_color_channels + info.num_extra_channels;
+
+            // Validate dimensions/channel count before the decoder proceeds
+            // further and may attempt huge allocations on corrupt inputs.
+            ImageSpec earlyspec(info.xsize, info.ysize, m_channels,
+                                m_data_type);
+            if (!check_open(earlyspec, { 0, (1 << 30) - 1, 0, (1 << 30) - 1, 0,
+                                         1, 0, 4099 })) {
+                return false;
+            }
```
jessey-git, line 294: Do we need the other check_open call below this still? It seems like every file is going to hit this JXL_DEC_BASIC_INFO conditional block and fill in the `info` struct.

lgritz: Quite right. I made an amendment.

jessey-git (approved):
Looks good to me.

PR #5306, opened 2026-07-08
fix(psd): avoid signed overflow computing layer/mask extents

lgritz:
read_layer_record computed a layer's width/height as std::abs((int)right - (int)left). The rectangle coordinates are signed 32-bit values (layers can have negative offsets), but were stored in unsigned (uint32_t) struct fields, and subtracting two ints near the extremes overflows -- undefined behavior that UBSan flagged on a fuzzed file (1615681375 - -1605526113 cannot be represented in type 'int').

Retype Layer's and MaskData's top/left/bottom/right fields as int32_t, matching what they actually hold, instead of reinterpreting bits at the point of use. Widen to int64_t before subtracting so the computation can't overflow; the absolute difference of two int32 values always fits in uint32, and validate_resolution still rejects implausibly large extents.

The mask data rect (used for layer masks) had the identical std::abs((int)right - (int)left) overflow hazard in load_layer_channel and gets the same fix.

### DIFF
--- a/src/psd.imageio/psdinput.cpp
+++ b/src/psd.imageio/psdinput.cpp
@@ -143,7 +143,7 @@ private:
     };
 
     struct Layer {
-        uint32_t top, left, bottom, right;
+        int32_t top, left, bottom, right;
         uint32_t width, height;
         uint16_t channel_count;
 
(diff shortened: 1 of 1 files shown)


lgritz:
Comments or concerns?

lgritz:
No objections -> merging

PR #5307, opened 2026-07-08
fix(psd): guard row interleave bounds on corrupt data

lgritz:
Validate channel-row byte availability before interleaving scanlines in PSD reads, and fail gracefully with an error instead of risking out-of-bounds access on malformed files.

Assisted-by: GitHub Copilot / GPT-5.3-Codex


grdanny (approved):
makes sense to me!

lgritz:
CVE-2026-63420

PR #5308, opened 2026-07-09
fix(python): update getattribute one-arg test for unified bindings

lgritz:
PRs #5260 and #5254 interact: #5254 exposed the default type arg on the module getattribute (`"type"_a = TypeUnknown`), so a one-arg call now returns None instead of raising TypeError. #5260's test still expected the old pybind behavior. Update the test and ref to the intended new behavior.


lgritz:
Because this fixes a horribly broken CI across the board, I will merge it immediately if CI fully passes.

PR #5309, opened 2026-07-09
fix(softimage): support channel packets with differing bit depths

lgritz:
A Softimage PIC file stores a chain of channel packets, each carrying its own bit depth (8 or 16). The reader assumed one uniform per-channel size for the whole image, so a file mixing depths (e.g. 16-bit R with 8-bit G) wrote past the end of the scanline buffer.

Fix: always expose one uniform pixel format, the widest depth among the file's packets, and promote narrower channels to it on read via exact bit replication (v*257 for 8->16), which preserves the normalized value exactly. This keeps the pixel layout naturally aligned; representing channels with their true differing formats instead would pack them at non-uniform, sometimes misaligned byte offsets.


lgritz:
Any comments on this one?

lgritz:
Over a week, no comments -> merging

PR #5310, opened 2026-07-10
feat(python): Expand dual-backend support in Python bindings

soswow:
Bring remaining Python modules onto the shared pybind11/nanobind sources,
simplify helpers (buffer request, make_numpy_array shapes, module attrs),
and add OIIO_PY_PROP_RW_NONE for None-assignable properties.
Add regression coverage for 0-d ImageBuf buffers, get_bytes_attribute
default conversion (str/bytes/None), and related colorconfig/imagebuf
paths. Update MIGRATION_STATUS to reflect full dual-backend coverage.

Assisted-by: Cursor / Grok


lgritz (approved):
LGTM!

Let's merge it and see what happens!

PR #5312, opened 2026-07-11
fix(raw): reject decompression-bomb / corrupt headers before unpack

lgritz:
Add a pre-unpack guard that rejects when the declared raw data size dwarfs the actual file size by more than 10000x (real raw sensor data never compresses anywhere near that ratio), plus a check_open() call enforcing OIIO's standard global resolution/imagesize limits and LibRaw's own per-dimension and per-image channel caps, matching the pattern already used in other format readers.


lgritz:
Any objections?

PR #5313, opened 2026-07-12
build(cmake): guard libtiff's missing Deflate target

### LINKED ISSUE #4439: [BUG] Incremental builds fail: unknown "Deflate::Deflate" target
**Describe the bug**

When I have OIIO find and build missing TIFF and libdeflate libraries once, there's no problem. However, if I use that same build directory for successive builds, the build system is able to find TIFF along with the other dependencies OIIO had built previously, as expected; but the build fails very shortly after, due to the TiffTargets.cmake 

building with CMake-3.29+
```
OIIO 2.6.6.0dev | Linux/ARM
    Build compiler: gcc 10.2 | C++17/201703
    HW features enabled at build: neon
Dependencies: BZip2 NONE, DCMTK NONE, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF NONE, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.20, Libheif NONE, libjpeg-turbo 3.0.3, LibRaw NONE, OpenColorIO 2.3.2, OpenCV
    NONE, OpenEXR 3.2.4, OpenGL, OpenJPEG NONE, PNG 1.6.44, Ptex NONE, Ptex NONE, pybind11 2.13.6, Python3 3.9.20, Qt5 NONE, Qt6 NONE, Robinmap 1.3.0, TBB NONE, TIFF 4.6.0, WebP NONE, yaml-cpp 0.8.0, ZLIB 1.2.7
Running on 20 cores 31.4GB 
    No compute specific hardware enabled.
```
```
OIIO 2.6.3.0dev | MacOS/ARM
    HW features enabled at build: neon
Dependencies: BZip2 1.0.8, DCMTK NONE, FFmpeg NONE, fmt 10.2.1, Freetype 2.13.2, GIF NONE, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.20, Libheif
    NONE, libjpeg-turbo 3.0.3, LibRaw NONE, OpenColorIO 2.3.2, OpenCV NONE, OpenEXR 3.2.4, OpenGL, OpenJPEG NONE, PNG 1.6.43, Ptex NONE, Ptex NONE, pybind11
    2.13.5, Python3 3.9.16, Qt5 NONE, Qt6 NONE, Robinmap 1.3.0, TBB NONE, TIFF 4.6.0, WebP NONE, ZLIB 1.2.12
```

**To Reproduce**

In an environment that does not have TIFF available: 
Build OIIO twice in a row, using the same build directory

1. Build OIIO using a persistent build directory, like `/tmp/build_oiio`, building all missing dependencies. The build should complete successfully, and TIFF libraries should live under `/tmp/build_oiio/deps/dist/lib`.
2. Do the exact same thing again: build OIIO using the same build directory.

**Evidence**

After the first build completes successfully, the second build finds the freshly-made TIFF and libdeflate under `<build_dir>/deps/dist/lib` and elicits the following error:

```
  CMake Error at /tmp/build_oiio/deps/dist/lib/cmake/tiff/TiffTargets.cmake:61 (set_target_properties):
    The link interface of target "TIFF::tiff" contains:

      Deflate::Deflate

    but the target was not found.  Possible reasons include:

      * There is a typo in the target name.
      * A find_package call is missing for an IMPORTED target.
      * An ALIAS target is missing.

    /tmp/build_oiio/deps/dist/lib/cmake/tiff/TiffConfig.cmake:24 (include)
    /opt/_internal/pipx/venvs/cmake/lib/python3.10/site-packages/cmake/data/share/cmake-3.29/Modules/FindTIFF.cmake:103 (find_package)
    src/cmake/dependency_utils.cmake:374 (find_package)
    src/cmake/externalpackages.cmake:88 (checked_find_package)
    CMakeLists.txt:178 (include)
```

# 🔍 
If I'm understanding what's happening, the Deflate::Deflate alias is only created [if OIIO is building TIFF](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/fe9f7f8451415d8daef920e76f3c16423b1efe7d/src/cmake/build_TIFF.cmake#L15-L18). This isn't a problem if the build system immediately finds and uses the TIFF libraries it just built; but I guess on subsequent builds, `checked_find_package(TIFF)` immediately finds TIFF from the first build, and never has the opportunity to invoke anything in `build_TIFF.cmake`. 

This is a deceptively tricky issue, I think. I ran into this a few weeks ago, and pretty quickly discovered it wasn't as simple as lifting the `checked_find_package(libdeflate ...)` and subsequent `alias_library_if_not_exists` command to `externalpackages.cmake`, before the check for TIFF, although I can't remember what goes wrong, specifically -- just that I knew I was pretty immediately out of my depth.

# 🤷 
I've been working around this by providing `-DOpenImageIO_BUILD_LOCAL_DEPS=TIFF` to CMake, forcing OIIO to always build TIFF. This works fine.

zachlewis:
## Description

Fixes configuration failures when OIIO finds a previously autobuilt static libtiff package whose exported `TIFF::tiff` target references `Deflate::Deflate` without importing or defining that target.

Before TIFF discovery, this adds a narrow compatibility guard that quietly loads libdeflate's config package and aliases its static or shared target as `Deflate::Deflate` when necessary.

Fixes #4439.

## Root cause

libtiff links static TIFF builds against `Deflate::Deflate`, causing the installed `TIFF::tiff` target to export `$<LINK_ONLY:Deflate::Deflate>`.

However, libtiff's installed `tiff-config.cmake` loads that exported target without importing its dependencies. The corresponding upstream template still contains `# TODO: import dependencies`.

Upstream report: https://gitlab.com/libtiff/libtiff/-/work_items/871

OIIO encounters this during `checked_find_package(TIFF)`. CMake's `FindTIFF` module first attempts libtiff's config package, which exposes the malformed export. This is not specific to CMake 4.3: I reproduced it with CMake 3.29.6 and 4.3.1. Issue #4439 likewise originally reported CMake 3.29+ on Linux and macOS rather than a Homebrew-specific failure.

## Why fix this in OIIO?

The underlying package-export defect belongs in libtiff, but an upstream fix will not repair already-installed libtiff releases or OIIO dependency caches. OIIO also builds the affected static libtiff configuration itself. 

This also breaks OIIO's ability to self-build (some) dependencies that [in]directly depend on libtiff or deflate. (Not for any dependencies for which we currently have self-building recipes so far; but this has blocked me in the past from submitting PRs for some of the more complex (sub)dependencies (libheif comes to mind; maybe jxl...? can't remember, it was more than a year ago when I last looked into this)

The guard is intentionally defensive:

- it runs only when `Deflate::Deflate` does not already exist;
- libdeflate is searched quietly in config mode;
- it does not make libdeflate a new required dependency;
- it does not change TIFF's config-mode or module-mode selection;
- it does not hide an actually missing library if a TIFF export requires one.

## Testing

Tested on:

- macOS 26.3.1, Apple Silicon
- AppleClang 21.0.0
- Homebrew 6.0.9
- libtiff 4.7.1
- libdeflate 1.23
- CMake 4.3.1 and 3.29.6

Verification performed:

- reproduced the original generation failure before the change;
- fresh config-mode configure with CMake 4.3.1 against the defective static TIFF export;
- build completed successfully: 126/126 targets;
- module-mode configure with Homebrew libtiff and libdeflate discovery explicitly disabled;
- configure with CMake 3.29.6 against the same defective static TIFF export.

Steps taken to verify:

- Built the wheel twice with the same persistent `SKBUILD_BUILD_DIR`, `IGNORE_HOMEBREWED_DEPS=1`, and without forcing a local TIFF rebuild. The first build autobuilt static TIFF 4.7.1 and libdeflate 1.23; the second found the cached TIFF package and completed without the previous `Deflate::Deflate` error.

### No robots worth speaking of were harmed in the making of this PR

Assisted-by: Codex / GPT-5

Codex was used to trace the CMake package-discovery chain, inspect the libtiff exports, reproduce the failure, prepare the compatibility guard, and run the verification described above.

### DIFF
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -91,7 +91,7 @@ checked_find_package (libuhdr
                       VERSION_MIN 1.3)
 
 # Static libtiff configs may reference this target without importing it.
-# https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4439
+# https://gitlab.com/libtiff/libtiff/-/work_items/871
 if (NOT TARGET Deflate::Deflate)
     find_package (libdeflate CONFIG QUIET)
     alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_static)
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -90,6 +90,14 @@ endif ()
 checked_find_package (libuhdr
                       VERSION_MIN 1.3)
 
+# Static libtiff configs may reference this target without importing it.
+# https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4439
+if (NOT TARGET Deflate::Deflate)
+    find_package (libdeflate CONFIG QUIET)
+    alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_static)
+    alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_shared)
+endif ()
+
 checked_find_package (TIFF REQUIRED
                       VERSION_MIN 4.0
                       RECOMMEND_MIN 4.5
--- a/src/cmake/externalpackages.cmake
+++ b/src/cmake/externalpackages.cmake
@@ -90,10 +90,14 @@ endif ()
 checked_find_package (libuhdr
                       VERSION_MIN 1.3)
 
-# Static libtiff configs may reference this target without importing it.
-# https://gitlab.com/libtiff/libtiff/-/work_items/871
+# Static libtiff configs may reference Deflate::Deflate without importing it
+# (https://gitlab.com/libtiff/libtiff/-/work_items/871), so libdeflate must be
+# located before TIFF discovery. In particular, a previously auto-built static
+# TIFF rediscovered from the local deps cache needs this; the libdeflate found
+# during build_TIFF.cmake does not carry over to later reconfigures.
 if (NOT TARGET Deflate::Deflate)
-    find_package (libdeflate CONFIG QUIET)
+    checked_find_package (libdeflate
+                          VERSION_MIN 1.18)
     alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_static)
     alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_shared)
 endif ()
--- a/src/cmake/Config.cmake.in
+++ b/src/cmake/Config.cmake.in
@@ -20,6 +20,16 @@ if (NOT @BUILD_SHARED_LIBS@)
     # This is required in static library builds, as e.g. PNG::PNG appears among
     # INTERFACE_LINK_LIBRARIES. If the project does not know about PNG target, it will cause
     # configuration error about unknown targets being linked in.
+    # Static libtiff configs may reference this target without importing it.
+    # https://gitlab.com/libtiff/libtiff/-/work_items/871
+    if (NOT TARGET Deflate::Deflate)
+        find_package (libdeflate CONFIG QUIET)
+        if (TARGET libdeflate::libdeflate_static)
+            add_library (Deflate::Deflate ALIAS libdeflate::libdeflate_static)
+        elseif (TARGET libdeflate::libdeflate_shared)
+            add_library (Deflate::Deflate ALIAS libdeflate::libdeflate_shared)
+        endif ()
+    endif ()
     find_dependency(TIFF)
     find_dependency(OpenColorIO)
     if (@JPEG_FOUND@)


src/cmake/externalpackages.cmake
```diff
@@ -90,6 +90,14 @@ endif ()
 checked_find_package (libuhdr
                       VERSION_MIN 1.3)
 
+# Static libtiff configs may reference this target without importing it.
+# https://gitlab.com/libtiff/libtiff/-/work_items/871
+if (NOT TARGET Deflate::Deflate)
+    find_package (libdeflate CONFIG QUIET)
+    alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_static)
+    alias_library_if_not_exists (Deflate::Deflate libdeflate::libdeflate_shared)
+endif ()
```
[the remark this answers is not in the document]
zachlewis: Verified and fixed. I reproduced the same Deflate::Deflate failure by installing a static OpenImageIO build and configuring a downstream consumer against its OpenImageIOConfig.cmake. The installed config now creates the guarded libdeflate alias before find_dependency(TIFF). After the change, the consumer advances past the Deflate failure. The test also exposed separate pre-existing static-package gaps involving CMath::CMath and GIF::GIF, which are outside the scope of #4439.

src/cmake/externalpackages.cmake
```diff
@@ -90,6 +90,14 @@ endif ()
 checked_find_package (libuhdr
                       VERSION_MIN 1.3)
 
+# Static libtiff configs may reference this target without importing it.
+# https://gitlab.com/libtiff/libtiff/-/work_items/871
+if (NOT TARGET Deflate::Deflate)
+    find_package (libdeflate CONFIG QUIET)
```
lgritz: Internally, we use checked_find_package, which has lots of bells and whistles, including that it will be included in the build log report of which dependencies were found at which version.

zachlewis: There's something about when + how checked_find_package does what it does that makes it difficult for the config that libtiff sets up that fails to survive a rebuild, when the locally-built libtiff is rediscovered... I don't really understand enough about what's happening behind the scenes to grok exactly what's going on. But the only thing that seems to work is forcing tiff to always build locally (when rebuilding using an existing build dir persisting previously built deps).

lgritz: so... what do we do here? I really don't want to set a precedent of having some packages use a bare find_package, which bypasses the nice version report, as well as preclude using the auto-builders.

zachlewis: No, I understand -- I'll sic a robot on this to better understand why this is happening and how to fix in checked_find_package -- in a worst case scenario, we might have to do some contrived patching of libtiff's find module for deflate, or patch deflate itself. I went through discovering all of this a long time ago, and now it's all coming back to me... this was very confusing, and I still don't have the language to describe what I'm experiencing concisely. Hopefully robots can help with that, too. 

In the meantime, no pressure -- let's not merge until we have something that works with checked_find_package. I'll dig into this as I have cycles, and/or maybe libtiff folks will address this upstream. I'll mark this PR as a draft for now.

lgritz: I have no doubt that there's room for improvement of checked_find_package. Among other things, the auto-build works well when the dependency is not found at all, but has some sharp edges if another copy (particular one that was rejected for being older than our required minimum) is elsewhere on the system, sometimes it can leak into the components we are trying to build and use locally. I suspect there are various ways my cobbled-together cmake code in that part of the code has some weaknesses.

lgritz: @zachlewis Any update on this?

zachlewis: Got to the bottom of it. Two things compose:

1. libtiff's installed config exports `TIFF::tiff` referencing `Deflate::Deflate` but never imports it (their template still says `# TODO: import dependencies`).
2. Since CMake 3.29, `FindTIFF` probes `find_package(Tiff CONFIG QUIET)` first, so even our module-mode `checked_find_package(TIFF)` loads that config when one is on the prefix path — and the local deps dist dir always is.

On a fresh run, `build_TIFF.cmake` defines `Deflate::Deflate` before the export loads. On reconfigure, the cached TIFF is *found*, the recipe is skipped, nothing defines the target, generate fails. The prerequisite lived inside the recipe that gets skipped exactly when its output is rediscovered. (Hence `BUILD_LOCAL_DEPS=TIFF` "working" — it re-runs the recipe.)

Per your point, the guard is now `checked_find_package (libdeflate VERSION_MIN 1.18)` before TIFF discovery — dependency report and auto-builders both work through it (with system packages hidden, it auto-built libdeflate itself). The consumer-side guard in the installed OIIO config keeps plain `find_package`, since `checked_find_package` doesn't exist there. Verified the fresh-build → reconfigure cycle fails on main and passes here, on CMake 4.3.1 and 3.29.6; healthy system-libtiff configures unaffected.

Proof it unblocks real work: I have a working branch adding a local-build recipe for libjxl (plus submodule support in the dep builder). With this fix it builds and links; on main it can't even complete an incremental reconfigure. That's the loop that kept blocking me from adding recipes for the complex dependencies.

Filing separately: the installed static-OIIO config also can't stand alone for `CMath::CMath`/`GIF::GIF`; and your "sharp edges" comment confirmed — dep child builds can resolve system packages the parent is ignoring (`CMAKE_IGNORE_PATH` doesn't cover config searches; a Homebrew Imath 3.2 got into a local OpenEXR build and broke the link against our Imath 3.1.10). One-line fix via `CMAKE_IGNORE_PREFIX_PATH`, small PR coming.

CI is green on the updated branch, so I've marked this ready for review.

lgritz (approved):
LGTM.

Thanks for the fixes, I am happier with this being checked_find_package so that it shows up properly in the build report.

PR #5320, opened 2026-07-14
admin: very minor fixes to README & SECURITY

lgritz:


lgritz:
Important typo fixes, but text files only, can't break code -> merging

PR #5322, opened 2026-07-15
fix: beef up Exif error detection

lgritz:
One small improvement in how decode_exif detects corruption, to ensure that the initial header examination doesn't read past the array bounds -- by starting off with a check that the exif blob's size is at least a big as a TIFFHeader, otherwise parsing it is pointless.

Then the rest of the changes are finding places where readers call decode_exif but ignore the results, effectively ignoring corrupt Exif blobs. That is in fact the default intended behavior (if the rest of the image file can be read properly), but we are supposed to be checking the global "imageinput:strict" attribute and considering it a hard error when in strict mode. There were several places where we didn't handle that case, in the JPEG, PNG, WebP, and Heif readers.

### DIFF
--- a/src/jpeg.imageio/jpeginput.cpp
+++ b/src/jpeg.imageio/jpeginput.cpp
@@ -296,8 +296,13 @@ JpgInput::open(const std::string& name, ImageSpec& newspec)
         if (is_exif_marker(m)) {
             // The block starts with "Exif\0\0", so skip 6 bytes to get
             // to the start of the actual Exif data TIFF directory
-            decode_exif(string_view((char*)m->data + 6, m->data_length - 6),
-                        m_spec);
+            bool ok = decode_exif(string_view((char*)m->data + 6,
+                                              m->data_length - 6),
+                                  m_spec);
+            if (!ok && OIIO::get_int_attribute("imageinput:strict")) {
+                errorfmt("Corrupted Exif data");
+                return false;
+            }
         } else if (m->marker == (JPEG_APP0 + 1) && m->data_length >= 28
                    && !strncmp((const char*)m->data,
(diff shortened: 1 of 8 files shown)


src/jpeg.imageio/jpeginput.cpp
```diff
@@ -296,8 +296,13 @@ JpgInput::open(const std::string& name, ImageSpec& newspec)
         if (is_exif_marker(m)) {
             // The block starts with "Exif\0\0", so skip 6 bytes to get
             // to the start of the actual Exif data TIFF directory
-            decode_exif(string_view((char*)m->data + 6, m->data_length - 6),
-                        m_spec);
+            bool ok = decode_exif(string_view((char*)m->data + 6,
+                                              m->data_length - 6),
+                                  m_spec);
+            if (!ok && OIIO::get_int_attribute("imageinput:strict")) {
+                errorfmt("Corrupted Exif data");
```
jessey-git: nit: Can we pick and use a common phrase to use when `decode_exif` fails? This PR uses both "Corrupted Exif data" and "Could not decode Exif".  I don't have an opinion on which of these 2 is best, just that there should really be just 1 :)

lgritz: Ah, I thought you might catch that. I was trying to be a little lazy and just keep the same messages that were already used in each file rather than try to harmonize them (and being slightly afraid that once I started to pull on that thread, the sweater might unravel as I found 50 more places to change :-)

Let me take a pass at that.

lgritz: Amended to have the same error message for all failed Exif decodes

jessey-git (approved):
All good from my side.

PR #5325, opened 2026-07-16
fix(imagebuf): fix data race that could corrupt multithreaded reads

### LINKED ISSUE #5324: bug: computePixelHashSHA1 is unstable when multithreaded
**Description**

The `ImageBufAlgo.computePixelHashSHA1` function supports `blocksize` and `nthreads` parameters to allow the hash to be computed in parallel. Unfortunately when parallelised the results are unstable (presumably because the threads can return results in a non-deterministic order).

Consider the following script

```python
import OpenImageIO as oiio

bs=64
nthreads=2
for i in range(1, 10):
    print(oiio.ImageBufAlgo.computePixelHashSHA1(oiio.ImageBuf(sys.argv[1]), blocksize=bs, nthreads=nthreads))
```

Running using [`BrightRings.exr`](https://openexr.com/en/latest/test_images/TestImages/BrightRings.html) yields a variety of hashes

```sh
$ python id.py BrightRings.exr
170C7F4BBD658CBDAFE3EF24F024A1EA1A2FD950
CA255D0D716E9EC4F8E9A1D23B48343B8250E626
3B4B2F827B7BA179F4294E95E37A6325F4B5DD06
52E7BA6AA88F9BE3609610015673DF37321CAEEF
20B1D7DB81FE35F942B1584ACCA2459687F7CC77
9E29AB1F28825024E74C72118F5EC846F908DFE7
3B4B2F827B7BA179F4294E95E37A6325F4B5DD06
7F8997A4372B5DCC7B31BAB2307C1C23456E94F1
3B4B2F827B7BA179F4294E95E37A6325F4B5DD06
```

Reducing the number of threads to 1 yields a consistent result `3B4B2F827B7BA179F4294E95E37A6325F4B5DD06`.

I've only tried this on MacOS.

Please run `oiiotool --buildinfo` and paste the output here.

```
OIIO 3.1.15.0 | MacOS/ARM
    HW features enabled at build: neon
Dependencies: BZip2 1.0.8, DCMTK NONE, expat 2.6.3, FFmpeg NONE, fmt 12.1.0, Freetype 2.14.1, GIF 5.2.1, Imath 3.1.10, JPEG 80, JXL NONE, libdeflate 1.23, Libheif NONE, libjpeg-turbo 3.1.2, LibRaw NONE, libuhdr 1.4.0, minizip-ng 4.0.10, OpenColorIO 2.5.1, OpenCV
    NONE, OpenEXR 3.3.5, OpenGL, OpenJPEG 2.5.4, openjph NONE, PNG 1.6.50, Ptex NONE, Ptex NONE, pybind11 3.0.1, pystring NONE, Python3 3.10.11, Qt5 NONE, Qt6 NONE, Robinmap 1.4.0, TBB NONE, TIFF 4.7.1, WebP 1.6.0, yaml-cpp 0.8.0, ZLIB 1.2.12, ZLIB 1.2.12
```

This is a standard build from PyPi.

**To Reproduce**

Call `oiio.ImageBufAlgo.computePixelHashSHA1` on an image with `blocksize>0` and `nthreads>1`

The hash will vary.

lgritz:
m_pixels_valid was set true by realloc() as soon as the pixel buffer was allocated, before read() actually filled it with file data. Since validate_pixels() checks the flag lock-free before falling back to the mutex, a second thread (e.g. another computePixelHashSHA1 block) could see it as true and read a still-empty buffer. init_spec() had the same issue with m_spec_valid, set before all spec fields were written.

Move both flags to be set only after their data is fully populated, and make them (plus m_pixels_read, read via the public ImageBuf::pixels_read() with no lock) atomic so their lock-free readers get a proper memory barrier.

Fixes #5324

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -3,6 +3,7 @@
 // https://github.com/AcademySoftwareFoundation/OpenImageIO
 
 
+#include <atomic>
 #include <iostream>
 #include <memory>
 #include <regex>
@@ -348,10 +349,13 @@ private:
     image_span<std::byte> m_bufspan;   // Bounded buffer for local pixels
     typedef std::recursive_mutex mutex_t;
     typedef std::unique_lock<mutex_t> lock_t;
-    mutable mutex_t m_mutex;              // Thread safety for this ImageBuf
-    mutable bool m_spec_valid   = false;  // Is the spec valid
-    mutable bool m_pixels_valid = false;  // Image is valid
-    mutable bool m_pixels_read = false;  // Is file already in the local pixels?
+    mutable mutex_t m_mutex;  // Thread safety for this ImageBuf
+    // N.B. These three flags are atomic because each is read lock-free by
+    // some caller. The atomicity guarantees that a thread seeing `true` also
+    // sees all the data (spec fields, pixel buffer) correctly.
+    mutable std::atomic<bool> m_spec_valid   = false;  // spec is valid?
+    mutable std::atomic<bool> m_pixels_valid = false;  // Image is valid?
+    mutable std::atomic<bool> m_pixels_read  = false;  // local pixels loaded?
     bool m_readonly            = true;   // The bufspan is read-only
     bool m_badfile             = false;  // File not found
     float m_pixelaspect        = 1.0f;   // Pixel aspect ratio of the image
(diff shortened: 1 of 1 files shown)


lgritz:
The Mac CI failures are unrelated to this PR -- looks like GHA Mac runners updated their OpenCV and it's not compatible with what we're looking for.

lgritz:
Comments?

lgritz:
No objections -> merging

PR #5326, opened 2026-07-16
build(opencv): support OpenCV 5, add standalone header smoke test

lgritz:
Find OpenCV via its exported CMake config (PREFER_CONFIG) so OpenCV 5 is located correctly.

Modify our imagebufalgo_opencv.h to work with OpenCV 5. But how to test whether we did it correctly?

We never had an actual test for the OpenCV interop. So set one up.


grdanny (approved):
lgtm

PR #5327, opened 2026-07-17
fix(jpeg2000): guard against oversized and decompression-bomb headers before decode

lgritz:
check_open() validated the resolution only after opj_decode() had already run, so a corrupt or malicious header could make openjpeg spend a long time and allocate a huge buffer before OIIO ever rejected the file.

Add two pre-decode guards right after opj_read_header():

1. Build a provisional ImageSpec from the header's canvas size and per-component precision (channel count can still grow later from palette expansion, but that only makes the image larger, so the header-based estimate is a safe lower bound), and run it through check_open() before calling opj_decode().

2. Add a decompression-bomb check that rejects when the claimed uncompressed size both exceeds 1 GB and dwarfs the actual file size by more than 10000x -- a ratio far beyond anything a real JPEG2000 codestream produces.

The existing post-decode check_open() call stays in place as the authoritative check once the final channel count is known.

Also drop dead commented-out debug std::cout lines and an unused chantypes vector left over from earlier work.


lgritz:
Any comments or objections on this one?

PR #5330, opened 2026-07-18
ci: Test against and document the latest dependency versions

lgritz:


lgritz:
Just bumping some CI versions, no objections -> merging

PR #5331, opened 2026-07-20
feat(iba): Remove 'experimental' guards around FLIP_diff

lgritz:
Approaching release, we haven't had anybody request breaking change to the FLIP_diff APIs, so I think that in main we can remove the 'experimental' namespace guards. It's possible that we'll add features or controls to the behavior, but probably not break the ABI.


lgritz:
Any objections to this?

lgritz:
2 weeks, no objections -> merging.

This only goes in main, so it just means that for 3.2, this will no longer be hidden behind the "experimental" namespace.

PR #5332, opened 2026-07-22
Fix vcpkg install command syntax in INSTALL.md

BrianHanke:
I was looking into Windows build options and it seems vcpkg is a great choice now that it offers a more current version.

Just a very minor PR here: the syntax is `vcpkg install openimageio[tools]` (no space).


lgritz (approved):
LGTM, thanks for the fix!

PR #5333, opened 2026-07-23
feat(jpeg): add thumbnail functionality to jpeg reader

antond-weta:
### Description

This adds the thumbnail functionality to the jpeg reader by implementing get_thumbnail().

### Tests

I could not find any unit tests related to get_thumbnail(), so I implemented a test in `python-thumbnail-jpeg`. The test is done in python because we don't expose the thumbnails in oiiotool. The test is pretty basic, it just checks the thumbnail dimensions for now. Happy to add more unit tests if needed.

I am also going to submit another PR soon fixing the thumbnails in raw reader, that will add `python-thumbnail-raw`, hence the test name.

### Checklist:

  behavior.
  on the GitHub runners.)


src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
     return decode_iptc_iim(buf, m_spec);
 }
 
+void
+JpgInput::scan_for_thumbnail(cspan<uint8_t> exif)
+{
+    try {
```
antond-weta: This code performs a basic scan for the subimage tag in the exif, it doesn't check if the found bit pattern is not a part of any other exif data. We may want to move this functionality to exif.cpp, but that may affect any other readers relying on exif.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+                                     image_spec.format, thumb.localpixels());
+
+    if (!result) {
+        errorfmt(
+            "failed to initialise an ImageInput object of type \"jpeg\" with the thumbnail data");
+        image_input->close();
+        return false;
+    }
+
+    image_input->close();
```
antond-weta, line 920: Without this I was getting address sanitiser errors in the image_input's destructor. I couldn't understand the error, and don't really think that it is related to the change.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+            m_spec.attribute("thumbnail_nchannels", numchan);
+        }
+
+    } catch (const std::exception& e) {
+        errorfmt("ERROR scanning for thumbnail: {}", e.what());
+    }
+}
+
+bool
+JpgInput::get_thumbnail(ImageBuf& thumb, int subimage)
```
antond-weta, line 873: The get_thumbnail() methods in the TGA input and ImageBuf contain lock guards. Does this method need to be thread safe?

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+    if (m_thumbnail_data.empty()) {
+        errorfmt("No thumbnail found");
+        return false;
+    }
+
+    auto image_input = OIIO::ImageInput::create("jpeg", false);
+    if (image_input == nullptr) {
+        errorfmt("OIIO::ImageInput::create(\"jpeg\") error");
+        return false;
+    }
```
lgritz, line 889: If create() fails, it's supposed to leave an error message in OIIO::geterror().

```suggestion
    if (image_input == nullptr) {
        std::string errmsg = OIIO::geterror();
        errorfmt("{}", errmsg.size() ? errmsg.c_str() : "unknown error");
        return false;
    }
```

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+    ImageSpec temp_spec, image_spec;
+    Filesystem::IOProxy* pp = &proxy;
+    temp_spec.attribute("oiio:ioproxy", TypeDesc::PTR, &pp);
+
+    result = image_input->open("", image_spec, temp_spec);
+    if (!result) {
+        errorfmt(
+            "failed to initialise an ImageInput object with the thumbnail data");
+        return false;
+    }
```
lgritz, line 907: I think that if get_thumbnail, it's sufficient to just return false. But if you really want to return an error message, what you should be returning is `image_input->geterror()`.

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+
+    result = image_input->read_image(0, 0, 0, image_spec.nchannels,
+                                     image_spec.format, thumb.localpixels());
+
+    if (!result) {
+        errorfmt(
+            "failed to initialise an ImageInput object of type \"jpeg\" with the thumbnail data");
+        image_input->close();
+        return false;
+    }
```
lgritz, line 918: Here, too. Either just return false with no error, or do
```
errorfmt("{}", image_input->geterror());
```

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+        if (eoi_ptr) {
+            m_thumbnail_data = cspan<uint8_t>(soi_ptr, eoi_ptr);
+            m_spec.attribute("thumbnail_width", width);
+            m_spec.attribute("thumbnail_height", height);
+            m_spec.attribute("thumbnail_nchannels", numchan);
+        }
+
+    } catch (const std::exception& e) {
+        errorfmt("ERROR scanning for thumbnail: {}", e.what());
+    }
```
lgritz: Why the try/catch here? Is there anything called inside that will actually throw an exception?

src/jpeg.imageio/jpeginput.cpp
```diff
@@ -791,4 +793,140 @@ JpgInput::jpeg_decode_iptc(string_view buf)
...
+}
+
+bool
+JpgInput::get_thumbnail(ImageBuf& thumb, int subimage)
+{
+    if (!m_decomp_create) {
+        errorfmt("ImageInput hasn't been initialised properly");
```
lgritz: I truly hate to pick on this, but in this code base we try to uniformly use the American spelling conventions: "initialized"

jinhgkim:
> I could not find any unit tests related to get_thumbnail(), so I implemented a test in python-thumbnail-jpeg. The test is  done in python because we don't expose the thumbnails in oiiotool. The test is pretty basic, it just checks the thumbnail  dimensions for now. Happy to add more unit tests if needed.

Hello, I just stumbled on this while reviewing #5334. I wanted to let you know that I have PR #5236 up which exposes thumbnails in `oiiotool`, with tests in `oiiotool-thumbnail-get`. Once it's merged, I think it would make more sense to add the JPEG case there instead, since `python-*` tests are usually reserved for testing the Python bindings. I could be wrong, but just wanted to mention it here.

lgritz:
That's a good point from @jinhgkim.
Why don't I merge her PR first (as soon as she has addressed the last minor question/request I had), then @antond-weta, you can rebase on top of that, and maybe you'll find er oiiotool thumbnail get and set tools helpful? Though it's also fine if you want to leave your tests as python, there's nothing wrong with that.

antond-weta:
Sure thing! I can update both PRs to use oiiotool for testing. The raw side of thing would still require a separate test, as it is based on an optional component, but the jpeg side can live in `oiiotool-thumbnail-get`

src/jpeg.imageio/jpeginput.cpp
antond-weta: done

src/jpeg.imageio/jpeginput.cpp
antond-weta: done

src/jpeg.imageio/jpeginput.cpp
antond-weta: removed

src/jpeg.imageio/jpeginput.cpp
antond-weta: done

antond-weta:
I've changed the unit test to use the new functionality in oiiotool. I have kept it separate in `jpeg-thumbnail`, happy to move it over to `oiio-thumbnail-get` if we are ok with adding a dependency on `oiio-images` there.

lgritz (approved):
LGTM

Issue #5229, opened 2026-06-11
feat: ultraHDR jpeg writer plugin


filmkorn:
**Is your feature request related to a problem? Please describe.**
There's no simple method (that I know of) of converting HDR content from floating point formats like EXR to widely web compatible images.

**Describe the solution you'd like**
Write support for UltraHDR JPEGs.

**Describe alternatives you've considered**
ImageMagic

**Additional context**

mugulmd:
@lgritz I don't know if that's a feature you're interested in having, but if so then I would gladly work on it :) (I was the one to initially add support for UltraHDR in `oiio`, it was a while ago but I still remember how things work)

lgritz:
@mugulmd -- absolutely, any format that we read, we would welcome a writer.

mugulmd:
Great, I'm on it!

Issue #4980, opened 2025-12-16
feat: Color Interop IDs and related color space improvements


brechtvl:
Discussion is happening in #4787 and various pull requests. This is an issue to keep track of concrete tasks and design decisions (not as another place for discussion).

### Color Interop IDs

* [x] Scene referred color interop IDs support in I/O
* [ ] #4972
  * [ ] Prefer display referred interop IDs over scene referred on read
  * [ ] Special exception for Gamma 2.2 display?
* [ ] #4971
* [ ] Do or do not write interop IDs for scene referred?
* [ ] Default to scene referred interop IDs for texture system and maketx?
  * [ ] Preference as `config` instead of global attribute, for multithreaded apps (part of #4972)
  * [ ] Change make maketx defaults?
  * [ ] Change texture system defaults? But how, `ImageCache` is used for other things.
* [ ] Changes needed for namespaced interop IDs
* [ ] Look up interop IDs in default config if current config does not have them

### Color Space Metadata
* OpenEXR `colorInteropID`
  * [x] Read support
  * [x] #4967
* CICP
  * [x] #4964
  * [x] PNG support
  * [x] HEIF support
  * [x] #4968
    * [ ] Handle more interop IDs through custom primaries, white point and gamma.
* ICC profiles
  * [ ] Use ICC profile associated with color spaces in OCIO 2.5 configs in image writing
  * [ ] Bundle ICC profiles for display interop IDs and automatically write them to file?
  * [ ] Read `cicpTag` from ICC profile, or other ways to detect interop ID from ICC profile?
  * [ ] Determine interop ID from ICC profile by comparing OCIO transforms?
* [ ] Deferred and unified resolution of color interop ID from metadata

### HDR Metadata

* [ ] Mastering Display Color Volume metadata read and write
* [ ] HDR reference white metadata read and write (SMPTE ST 2094-50)

PR #5335, opened 2026-07-24
feat(jpeg): support Ultra HDR output

mugulmd:
Assisted-by: `claude-code` / `opus-5`

### Description

Related to https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/5229.

In its current state, OpenImageIO only supports *reading* Ultra HDR images, and does not support *writing* such images.

This feature has been requested because it enables converting an HDR image from specialized formats such as OpenEXR to widely-supported and easily displayable/exchangeable JPEG files.

`libultrahdr` (the codec we use for Ultra HDR) has plenty of options but we only use a restricted subset for simplicity, which leads to the following constraints:
- incoming image data must be half or float
- `oiio:ColorSpace` must be one of `lin_rec709_scene`, `lin_p3d65_scene`, `lin_rec2020_scene` because it needs a corresponding "Ultra HDR color gamut".

Also, in order to trigger the Ultra HDR output path, the incoming image must have the `jpeg:ultrahdr` metadata set to any non-zero integer.

### Tests

3 tests have been added to `jpeg-ultrahdr`:
- synthesize an Ultra HDR image, write it, then ensure reading it back succeeds
- ensure writing an Ultra HDR image with integer data fails
- ensure writing an Ultra HDR image with an unsupported color space fails.

### Checklist:

- [X] **I have read the [Policy on AI Coding Assistants](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/docs/dev/AI_Policy.md)**
  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/jpeg.imageio/jpegoutput.cpp
+++ b/src/jpeg.imageio/jpegoutput.cpp
@@ -9,6 +9,7 @@
 
 #include <OpenImageIO/filesystem.h>
 #include <OpenImageIO/fmath.h>
+#include <OpenImageIO/half.h>
 #include <OpenImageIO/imageio.h>
 #include <OpenImageIO/tiffutils.h>
 
@@ -65,10 +66,28 @@ private:
     unsigned long m_outsize = 0;
 #endif
 
+#if defined(USE_UHDR)
+    // Ultra HDR output. When m_write_uhdr is true, we bypass the libjpeg
+    // compressor entirely: scanlines are accumulated into m_uhdr_pixels (a
+    // full-frame packed RGBA half-float buffer) during write_scanline(), and
+    // the actual encode via libultrahdr happens in close(). libultrahdr needs
+    // the whole raw image at once, so we cannot stream it.
+    bool m_write_uhdr = false;
+    std::vector<half> m_uhdr_pixels;  // Full-frame packed RGBA, 4 halfs/pixel
+    uhdr_color_gamut_t m_uhdr_cg = UHDR_CG_UNSPECIFIED;
+    int m_uhdr_quality           = 95;
+#endif
+
     void init(void)
     {
         m_copy_coeffs       = NULL;
         m_copy_decompressor = NULL;
+#if defined(USE_UHDR)
+        m_write_uhdr = false;
+        m_uhdr_pixels.clear();
+        m_uhdr_cg      = UHDR_CG_UNSPECIFIED;
+        m_uhdr_quality = 95;
+#endif
         ioproxy_clear();
         clear_outbuffer();
     }
@@ -96,6 +115,18 @@ private:
     // Read the XResolution/YResolution and PixelAspectRatio metadata, store
     // in density fields m_cinfo.X_density,Y_density.
     void resmeta_to_density();
+
+#if defined(USE_UHDR)
+    // Validate the spec for Ultra HDR output, infer the color gamut, and
+    // allocate the full-frame pixel buffer. Called from open() when the
+    // "jpeg:ultrahdr" attribute is set. Returns false (with an error set) if
+    // the spec is not suitable for Ultra HDR encoding.
+    bool setup_uhdr_output();
+
+    // Encode the accumulated m_uhdr_pixels buffer with libultrahdr and write
+    // the resulting JPEG stream to the IOProxy. Called from close().
+    bool encode_uhdr();
+#endif
 };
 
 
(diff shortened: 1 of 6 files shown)


lgritz:
Hi, @mugulmd, just a reminder that this is still marked as a draft, needs a DCO, and has some clang-format fixes. I think that the "hobbled" failure is unrelated and has been separately fixed, should be fixed if you rebase this on top of the current main.

mugulmd:
Hi @lgritz yep sorry I just reached the stage of "this works on my machine, let's make a draft PR and finish this later" and I've been procrastinating a little bit :sweat_smile: I have some spare time in the days to come, I'll finish this properly!

lgritz:
Can you rebase on top of current main to fix the apparent conflict in buildinplugins.rst?

I wonder if the conflict is related to the several places where you seem to have changed whitespace, probably something your editor or pre-commit hooks did automatically?

mugulmd:
@lgritz all done! I found the culprit: Zed (which I use as editor) automatically removes trailing whitespaces on save. Should be good now 🙂

src/jpeg.imageio/jpegoutput.cpp
```diff
@@ -34,7 +35,11 @@ class JpgOutput final : public ImageOutput {
     const char* format_name(void) const override { return "jpeg"; }
     int supports(string_view feature) const override
     {
-        return (feature == "exif" || feature == "iptc" || feature == "ioproxy");
+        return (feature == "exif" || feature == "iptc" || feature == "ioproxy"
+#if defined(USE_UHDR)
+                || feature == "uhdr"
```
lgritz: I'm not crazy about "uhdr" here as a jpeg-specific query -- the supports() is supposed to be standard tokens that could apply to lots of formats. But I think it totally makes sense for an app to want to know if a given file format can support HDR images, so I would be in favor of:
* this being renamed to "hdr"
* documented in imageio.h where we explain what all the accepted ImageOutput::supports() tokens are
* added to other formats that support HDR output

src/jpeg.imageio/jpegoutput.cpp
```diff
@@ -458,6 +511,129 @@ JpgOutput::resmeta_to_density()
...
+{
+    // libultrahdr's half-float HDR intent requires linear pixel data. Only
+    // float-based OIIO formats can carry HDR values, so reject integer specs.
+    if (m_spec.format != TypeDesc::HALF && m_spec.format != TypeDesc::FLOAT) {
+        errorfmt(
+            "JPEG Ultra HDR output requires a half or float pixel data type, "
+            "not \"{}\"",
+            m_spec.format);
+        return false;
+    }
```
lgritz: I wonder if the right thing here, when handed a request to output a data type that does not require HDR, it just to silently fall back to writing a normal JPEG, rather than a hard error?

I think the usual OIIO way is "the app doesn't necessarily have to know details of the file format, they make their request, and we try to fulfill it as best as we can, only failing if there is no way to make a good guess about 'what they probably want.'" I think all the writers are robust to being asked to write data types they don't support, they just fall back to a type that they do support that minimizes loss of precision versus the original request. This is what allows `oiiotool in.fmt1 -o out.fmt2` to work in a common sense way without really caring about whether every data type that fmt1 supports is also supported in fmt2 and forcing the user or calling app to have to handle every kind of mismatch. If uint8 data is read from a jpeg and written to exr, we just silently write `half`, we don't refuse just because exr doesn't support uint8.

lgritz (commented):
This is looking really good!

Two minor design issues I commented on for us to discuss and possibly alter, but other than those, this all is great.

src/jpeg.imageio/jpegoutput.cpp
mugulmd: Actually I preferred to remove this for now as this is out-of-scope for this PR. I'll be happy to implement this in a follow-up PR though :)

src/jpeg.imageio/jpegoutput.cpp
mugulmd: Completely agreed, I got used to write "agressive" code because dealing with errors makes everything much easier to debug as a developer, but this is not ideal for user experience.
I updated the code so that any failure to setup the Ultra HDR output falls back to regular JPEG :)
(this means failure due to data type AND failures due to color space)

lgritz (approved):
LGTM!

lgritz:
Merged!

As food for thought, I wonder what it would need to take (technically? culturally?) before having to set the "jpeg:uhdr" hint attribute on each output starts to seem like a burden?

Is it worth having a global `OIIO::attribute("jpeg:uhdr", ..)` that sets a default once, say in an app, without it needing to be set for each file you write?

Will it ever be wise to have the default be "always write a jpeg with uhdr if handed half or float data and libultrahdr was used at build time?" IIRC, a reader not aware of uhdr will simply read such a file in LDR mode and still get a usable image, right?

mugulmd:
That's exactly right! So in theory it should be always safe to shoot for uhdr because:
- writer falls back to regular JPEG if anything goes wrong
- reader falls back to LDR mode if it does not support HDR.

However it seems that `libultrahdr` brought quite a bunch of new features in the recent versions, including some breaking changes and the possibility to encode Ultra HDR in other file formats (namely WEBP and AVIF), so this may require a bit more API design thought.

I'm going on vacation but after that I can definitely draft up a complete overview of what `libultrahdr` supports, how it could/should map to `oiio`, and some proposal for an API design that makes the user's life as easy as possible.

lgritz:
OIIO 3.1 added the ability to read UHDR in JPEG. Now for 3.2 we have the ability to write it to JPEG -- when asked. That's already a step forward, and lets us evaluate it, and we can always change future versions to do that by default as well as to add the capabilities to other formats. There's no reason to rush further changes. Enjoy your vacation!

brechtvl:
Nice work! Writing Ultra HDR images is something Blender users have requested, though I think mainly if it supports supplying separate SDR and HDR graded images, so that unlike other image formats there is more control over how it displays on SDR and HDR monitors. And the HDR would probably be in PQ or HLG.

I added to the list in #4980 as a personal reminder, and might look into it one day if no one does before me, though probably not anytime soon.

PR #5336, opened 2026-07-24
fix(tiff): hardening: prevent OOB read, overflow protection

lgritz:
In the raw-strip read path of read_native_scanlines(), a failed TIFFReadRawStrip() returns csize < 0.  Bail out of the strip loop for this case before running the decompress, matching the existing tiled path (read_native_tiles already breaks on this condition), and guard the leftover-scanline loop with ok so it doesn't run after the bail.

Also size scanline scratch with 64-bit arithmetic, for overflow protection.


lgritz:
This is the first PR in a new initiative I'm taking to proactively look file format by file format and try to uncover potential bugs before the fuzzers or the bad guys even find them. So expect another series of PRs. I will try to organize them into one per format.

lgritz:
Near as I can tell, the "hobbled" test failures are unrelated to this PR and are happening (nondeterministically) on several branches/PRs. I think it's some kind of GH configuration thing, am looking at it separately.

lgritz:
Any objection to or comments about this very small code change?

lgritz:
No objections -> merging

PR #5337, opened 2026-07-25
fix(rla): better detection of corrupted RLE encoding

lgritz:
Addressing a buffer overrun when the RLE count exceeds the number of encoded bytes left to read from.


lgritz:
I'm very confused by what's going on here. This sure looks like an exact duplicate of #5258 -- look at it's diffs! -- and GH says that PR was merged. And yet... the code isn't in main or in the history. I do not understand what is going on. GitHub glitch?

Since I *thought* this was already merged, and it had sat in review for a couple weeks, I intend to merge this right away if it passes CI. (Since we all believed it already had been merged.)

PR #5338, opened 2026-07-26
test(fuzz): changes to fuzz corpus seeding for better coverage

lgritz:
The old rule copied only the 5 smallest files (<=100 KB) per format into the run corpus. This is really biased and favors degenerate inputs (decompression bombs, truncated headers) over larger valid files as the basis to mutate around to find subtle bugs. So we make these changes:

* populate_corpora.py: drop the MAX_FILES count cap; gather every example from each format's testsuite examples; raise the per-file size cap from 100 KB to 5 MB.
* ci-fuzztest.bash: gather seeds into a scratch seed_dir, then run `oiio_fuzz_image -merge=1` to fold only the coverage-increasing subset into the cached run corpus before the timed session.

### DIFF
--- a/docs/dev/fuzzing.md
+++ b/docs/dev/fuzzing.md
@@ -116,50 +116,52 @@ merging the fix.
 
 ## Adding seeds for a new format
 
+Seeds are **not** hand-curated into `src/fuzz/corpora/` per format. At fuzz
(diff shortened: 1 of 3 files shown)


lgritz:
I will merge this over the weekend if there are no objections by then. It doesn't change any code, it just improves our strategy for the fuzzing runs.

PR #5339, opened 2026-07-26
fix(hdr): address conflict between hdr and raw with .hdr files

lgritz:
There is a confluence of unfortunate events:

- The .hdr extension is used by both Radiance hdr files, and actual raw files from Hasselblad cameras (rare).
- When EMBEDPLUGINS=0, the order that the plugins are declared are such that an .hdr file will get opened by the raw reader first.
- Certain old versions of libraw will crash when trying to open files it thinks are invalid (like Radiance hdr files).

This is causing our CI to fail the hdr test (because of the raw reader!) ever since we merged the fix that made EMBEDPLUGINS=0 really work (it had been subtly broken before). It only fails the "hobbled" test, because that's the only one with EMBEDPLUGINS=0.

The fix here is to remove the very-rarely-used .hdr extenion from the RAW extension list. If we really do encounter an Hasselblad raw file, it just means that the Radiance HDR reader will try it first, fail (safely), and then as long as try_all_readers is still enabled (it is the default), the raw reader will still get a crack at it.

Assisted-by: Copilot / Opus 4.8 (but only for investigation).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -534,7 +534,7 @@ jobs:
             # Use the oldest supported versions of required dependencies, and
             # disable most optional dependencies and features (no SSE or
             # OpenCV, don't embed plugins).
-            nametag: linux-disabled-ubuntu
+            nametag: linux-hobbled-ubuntu
             runner: ubuntu-22.04
             cc_compiler: gcc-9
             cxx_compiler: g++-9
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -238,12 +238,17 @@ raw_input_imageio_create()
     return new RawInput;
 }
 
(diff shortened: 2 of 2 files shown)


lgritz:
@antond-weta I tagged you here because you work extensively with libraw. Does this seem sane to you? Is the assumption correct that the .hdr extension for Hasselblad camera raw files are extremely rare compared to the old Radiance files with the same extension, and that to avoid the problem of ambiguous .hdr files which will crash if a Radiance file is given first to certain older versions of libraw, it's a small price to pay to force the Radiance hdr reader to always get first crack at it?

antond-weta (approved):
I've never heard of `.hdr` raw files from Hasselblad, or any other camera. Quick googling also didn't show anything of value. I've seen `.fff` and `.3FR` files from Hasselblad, but not `.hdr`.

I think this change is reasonable.

PR #5340, opened 2026-07-27
fix: convert integer channels to OpenEXR instead of reinterpreting bytes

lgritz:
Converting an image whose channels have differing native integer depths (e.g. a Softimage PIC with a 16-bit R and an 8-bit G channel) to OpenEXR produced garbage pixels: the uint16/uint8 bytes were copied straight into the half channels and reinterpreted rather than rescaled. Three coupled issues, all rooted in treating "output supports channelformats" as "output stores these exact per-channel formats":

- exroutput: spec_to_header forced spec.format to one of EXR's three stored types but left spec.channelformats naming the original (uint16/uint8) types, so the post-open spec misdescribed what was actually stored. Remap channelformats through the same mapping (and drop the list when all channels collapse to one type) so the spec is consistent with the half/float/uint the file really holds.

- ImageOutput::copy_image: the native (no-conversion) fast path was taken whenever the input had channelformats. Only take it when the output will store each channel in the input's native format; otherwise fall back to the converting path.

- iconvert: its by-hand copy path (used whenever the input has mixed channel formats) requested a native per-channel transfer unconditionally. Apply the same check against the opened output's stored formats and, when they differ, transfer through the output's format so values are rescaled.

Regression test added to testsuite/softimage: converting the mixed 16/8-bit file to EXR now yields correctly normalized values.

Assisted-by: Claude Code / Opus 4.8

### DIFF
--- a/src/iconvert/iconvert.cpp
+++ b/src/iconvert/iconvert.cpp
@@ -473,16 +473,37 @@ convert_file(const std::string& in_filename, const std::string& out_filename)
                         "iconvert ERROR copying \"{}\" to \"{}\" :\n\t{}\n",
                         in_filename, out_filename, out->geterror());
             } else {
-                // Need to do it by hand for some reason.  Future expansion in which
-                // only a subset of channels are copied, or some such.
-                std::vector<char> pixels((size_t)outspec.image_bytes(true));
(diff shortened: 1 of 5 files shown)


lgritz:
This has been sitting for over a week with no objections. I will merge it at the end of the day today if nobody says no.

PR #5342, opened 2026-07-27
fix: use check_open(), check_compression_ratio() in all readers lacking them

### LINKED ISSUE #3974: [FEATURE REQUEST] Add check_open validity plausibiity checks to format readers

lgritz:
Apply check_open() and check_compression_ratio() consistently across all the format readers that previously lacked them. This helps ensure that a small, malformed file cannot drive a large pixel allocation before any pixel data is read, or other chaos that might result from the headers claiming resolutions that the file formats themselves (or common sense) disallow. 

Also add a new validity test to check_open: verify that tile sizes are non-negative and not larger than the largest allowed image size for that format.

Readers guarded here: cineon, dds, dpx, fits, gif, hdr, heif, ico, iff, jpeg, null, openexr (both the C++ and C-API readers), png, pnm, rla, sgi, softimage, webp, zfile.

Apologies for making changes to many formats in one commit, but since the changes to the readers are short and are nearly identical, it seemed like less reviewer burden to do it all in one PR rather than breaking it up per-format.

Also adds a whole bunch of self-contained decompression-bomb / extent regression tests (targa, png-bomb, openexr-bomb, sgi, dpx, cineon, iff, webp, hdr, ico, softimage, pnm) and registers the new seed corpora with the fuzz harness.

Fixes #3974


lgritz:
I'd really like to get eyes on this and an approval. It's been posted for a week, and I have quite a few other patches that are not submitted PRs yet because they need to be on top of this (because independently, they produce different error output than when this is also present, so submitting them separately without this is a merge nightmare for the reference test output).

The payoff is high -- this patch catches a lot of possible corruptions across 16 different formats, for most of them only changing a handful of lines of code, in a very similar way for each format.

src/null.imageio/nullimageio.cpp
jinhgkim, line 297:
```suggestion
            int nchans = Strutil::from_string<int>(a.second);
            if (nchans < 0 || nchans > 1024) {
                errorfmt("Invalid channel count {}", a.second);
                return false;
            }
            m_topspec.nchannels = nchans;
```
Channel names are built during `m_topspec.default_channel_names();`, the very next line, before `check_open()` runs. I think we should check the channel count here to avoid "foo.null?CHANNELS=-1". What do you think?

lgritz: Hmmm, good point. Or do you think it's better to just make default_channel_names() safe by putting inside it a check for nchannels > 0? (I think the only harmful thing in it for negative channels is the call to reserve().)

jinhgkim: Yup, that's good too. Don't we wanna check the upperbound as well though? What if `nchannels` is huge and we hang in the push_back loop that loops over the nchannels.

lgritz: ooh, yeah. That's an interesting dilemma. Because, check_open() is always called after the ImageSpec is created, and the IS construction itself does an allocation based on number of channels, that creates a bit of a chicken and egg problem. I guess the sensible thing is to put some sort of check in there.

Let me look at doing that and I will amend the PR.

lgritz: To prevent scope creep of this PR, I think what I want to do is create a separate, small PR for the default_channel_names fix, and so let's just sweep that under the rug for this PR and know that it will be separately fixed. Sound ok?

jinhgkim: Sounds great.

lgritz:
Is this ready to go? Anybody care to approve?

jinhgkim (approved):
Seems ready to me

PR #5343, opened 2026-07-27
fix: ImageBuf error message with too few format arguments aborts

lgritz:
While fixing something entirely different, CC spotted this bug.


grdanny (approved):
lgtm

PR #5344, opened 2026-07-27
fix(pnm): use 64-bit scanline math to avoid integer overflow

lgritz:
Also hoisted an invariant computation of the number of bytes read per scanline out of the loop.


grdanny (approved):
looks good to me

PR #5345, opened 2026-07-27
docs: never defined the alias for OIIO_DEPRECATED_EXTERNAL

lgritz:


lgritz:
A week with no objections, very minor docs-only change --> merging

PR #5346, opened 2026-07-29
python: nanobind tidying, auto-build, CI testing

lgritz:
A collection of changes related to our switch from pybind11 to nanobind.  None of these change the binding code itself! Mostly related to build and testing odds and ends.

Sorry for the large number of things mixed together; each change individually either does nothing or doesn't work on its own, they are a collection of things that are all needed to fully enable working nanobind across all our CI tests.

* CI: Build and test both python back ends! Except for Windows, only do nanobind there, since it's really slow to build the pybind11 bindings.

* auto-builder for nanobind if not found
  - src/cmake/build_nanobind.cmake, mirroring build_pybind11.cmake
  - With auto-build of nanobind, no need to use pip to install nanobind.
  - Take care to share robin-map with OIIO core (both OIIO and nanobind need robin-map).

* CMakeLists.txt : Change OIIO_PYTHON_BINDINGS_BACKEND initialization to use set_cache() to allow env variables to easily control it.

* Testing :
  - Fixes that make the tests run properly when only nanobind is built (it previously worked for "both" and of course for "pybind11").
  - Add tests to the nanobind set that were overlooked: filters, docs-examples-python, and the hwy-enabled python-imagebufalgo variant.
  - Add missing PYTHONPATH to python-imagebufalgo.hwy and openexr-copy tests, which relied on the ambient shell already having it set (true in CI via ci-startup.bash, not for a plain local ctest run).
  - Semi-related cleanup: Don't have docs-examples-python depend on docs-examples-cpp having already run to create its ref symlink; create it ourselves if missing. Fixes failures when running testing subset that includes docs-examples-python but not docs-examples-cpp.

* pythonutils.cmake: comment out a now-redundant find_package call for nanobind, since externalpackages.cmake already resolves it before this macro runs.

* Fix discover_nanobind_cmake_dir(): it was a macro, so its return could abort the entire including file (externalpackages.cmake) rather than just itself, whenever a stale cached nanobind_DIR was present -- silently skipping every dependency check after the Python section. Convert to function.

* Robin-map:
  - Raise robinmap minimum to 1.3, as needed by nanobind.
  - Remove our bespoke FindRobinmap.cmake module in favor of using modern robinmap's exported cmake configs (including cleanup of the changed target names).

* Suppress nanobind reference leak warnings for python 3.9, allow for 3.10+.  We were only getting these for 3.9, not newer Python, so discussion with Wenzel concluded that it's probably just an old python cleanup problem and not worth tracking down unless we see it surface with a newer Python release.


lgritz:
@soswow I would love your review of this, since it builds directly on your work.

soswow:
> @soswow I would love your review of this, since it builds directly on your work.

by the end of my today 👍

lgritz:
This is in a sense not directly related to the present PR, but people using both OIIO and OSL may be interested in https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/2142 that is relevant to how we're going to migrate the projects safely to nanobind, since mixing and matching an OIIO and OSL where one uses pybind11 and the other uses nanobind will fail if you use any OSL API calls that involve types imported from OIIO (luckily, there is only one such call!).

lgritz:
@soswow Ping. No rush, but also... I would like to incorporate this as soon as we can so that we are fully testing both bindings in CI.

soswow:
my bad. looking at it right now.

src/build-scripts/ci-startup.bash
```diff
@@ -53,6 +53,9 @@ export TESTSUITE_CLEANUP_ON_SUCCESS=${TESTSUITE_CLEANUP_ON_SUCCESS:=1}
 # For CI, default to building missing dependencies automatically
 export OpenImageIO_BUILD_MISSING_DEPS=${OpenImageIO_BUILD_MISSING_DEPS:=all}
 
+# Build both set of python bindings in main/3.2 for now
+export OIIO_PYTHON_BINDINGS_BACKEND=both
```
soswow: Do you want to use it as a default thought?

```suggestion
export OIIO_PYTHON_BINDINGS_BACKEND=${OIIO_PYTHON_BINDINGS_BACKEND:=both}
```

Otherwise value from, say `.github/workflows/ci.yml:705`
```
          - desc: Windows-2022 VS2022
            runner: windows-2022
            nametag: windows-2022
            generator: "Visual Studio 17 2022"
            python_ver: "3.12"
            oiio_python_bindings_backend: nanobind
            ctest_test_timeout: "240"
            setenvs: export OPENIMAGEIO_PYTHON_LOAD_DLLS_FROM_PATH=1
```
will be ignored.

.github/workflows/ci.yml
```diff
@@ -704,14 +702,15 @@ jobs:
...
             python_ver: "3.12"
+            oiio_python_bindings_backend: nanobind
             ctest_test_timeout: "240"
             setenvs: export OPENIMAGEIO_PYTHON_LOAD_DLLS_FROM_PATH=1
           - desc: Windows-2025 VS2026
             runner: windows-2025-vs2026
             nametag: windows-2025-vs2026
             generator: "Visual Studio 18 2026"
             python_ver: "3.12"
+            oiio_python_bindings_backend: both
```
soswow, line 713: Just to confirm - you said in PR description that we don't want to build both for windows builds. But we do both here.

src/python/py_oiio.cpp
```diff
@@ -591,6 +591,15 @@ OIIO_DECLARE_NB_MODULE(OpenImageIO)
 {
     m.doc() = "OpenImageIO nanobind bindings.";
 
+#    if PY_VERSION_HEX < 0x030a0000 /* less then 3.10 */
```
soswow:
```suggestion
#    if PY_VERSION_HEX < 0x030a0000 /* less than 3.10 */
```

soswow:
`src/build-scripts/ci-requirements-nanobind.txt` can be killed now, since we don't do `pip install` any more.
Also, we can update `INSTALL.md` part:
```
nanobind discoverable by CMake, or installed in the active Python 
environment so `python -m nanobind --cmake_dir` works
```
to something else, like
```
* [nanobind](https://github.com/wjakob/nanobind) >= 2.8 (tested through 2.13). 
If not found at build time, this will be automatically downloaded and built 
(same pattern as pybind11 / fmt). A system or pip install is still fine if you 
already have one discoverable by CMake (`nanobind_DIR`, or `python -m nanobind --cmake_dir`).
```

src/cmake/pythonutils.cmake
```diff
@@ -239,12 +249,13 @@ macro (setup_python_module_nanobind)
...
-        find_package (nanobind CONFIG REQUIRED)
     endif ()
 
     nanobind_add_module(${target_name} ${lib_SOURCES})
-    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND TARGET nanobind-static)
-        target_compile_options (nanobind-static PRIVATE -Wno-error=format-nonliteral)
+    if (TARGET nanobind-static AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang"
+                                    OR CMAKE_CXX_COMPILER_ID MATCHES "Apple"
+                                    OR CMAKE_CXX_COMPILER_ID MATCHES "IntelLLVM"))
+        target_compile_options (nanobind-static PUBLIC -Wno-format-nonliteral)
```
soswow: Do we really want it to be PUBLIC?
would `target_compile_options (nanobind-static PRIVATE -Wno-error=format-nonliteral)` do?
I am not good at this. Asking LLMs to explain this to me, but I understand this might hide (not even warning) actual warnings even in non-third-party code?

soswow (approved):
LGTM. up to you if you want to act on any of my comments.

src/cmake/pythonutils.cmake
lgritz: I found PUBLIC necessary, because the warning is encountered in the header, so it needs to be suppressed both in the build of nanobind itself (that's what PRIVATE would cover), and also in our files where we include the nanobind headers. Note that this declaration only suppresses the warning within our python bindings (the only part that uses the nanobind-static target), not across the whole codebase.

But I will double check one more time before merging, to see if there is a more localized way to suppress the warning. I'm not super worried about it. The fact that we weren't getting warned before means that we aren't doing it in our code currently -- anywhere -- which makes it seems unlikely that we will start abusing it now. I assume this is something nanobind will eventually fix on its end, but in the mean time, it's really hard to control for things that 3rd party libraries do that happen to warn at the high warning levels we prefer.

.github/workflows/ci.yml
lgritz: I think I updated the code subsequently. I will edit the commit message when I merge.

src/cmake/pythonutils.cmake
lgritz: Maybe I spoke too soon on this! I tried changing this to PRIVATE and it passes CI. I'm not sure what changed, I really thought I remembered that not working before. So I will make this fix, too.

lgritz:
I made all the changes suggested, including changing that warning to PRIVATE (now I'm really not sure how I concluded it needed to be public to work), so I am going to merge.

PR #5347, opened 2026-07-30
docs: correct Doxygen style comments in IOProxy

lgritz:


lgritz:
This has gone for several days without objection, and it is a comment-only change, so can't possibly break anything. Merging.

PR #5348, opened 2026-07-30
fix(psd): harden reader against oversized/bomb headers and ZIP-layer overflow

lgritz:
Add open-time hardening to the PSD reader:

- check_open() on the composite spec enforces the global `limits:*` (channels/resolution/imagesize) policy in addition to the existing per-dimension header validation.
- check_compression_ratio() rejects a tiny file that declares a huge uncompressed composite before any pixel read (decompression bomb).
- Size the ZIP / ZIP-prediction layer-channel decompression buffer with 64-bit arithmetic. `width*height*(depth/8)` was computed in uint32 and could wrap for large PSB layer dimensions, undersizing the buffer while read_channel_row() computes the row offset in 64-bit -> out-of-bounds read. Both now agree.

Add a decompression-bomb regression fixture (testsuite/psd); it is picked up as a fuzz seed automatically via FORMAT_SOURCES. Valid files read byte-identically.


src/psd.imageio/psdinput.cpp
jinhgkim, line 2501: Wouldn't this assert fail for `PSB` since it still uses `uint32_t` to compute, but `dest.size()` is now in 64-bit? Should we cast these to `size_t` as well?

lgritz: I think you are correct. Will fix.

lgritz: Actually, it was easier (and maybe caught another couple problems) to change the width and height parameters themselves to uint64_t. Pushed an update.

src/psd.imageio/psdinput.cpp
```diff
@@ -634,6 +634,19 @@ PSDInput::open(const std::string& name, ImageSpec& newspec)
...
+    // Enforce open-time extent/policy limits and a decompression-bomb guard on
+    // the composite image before any pixel read is attempted. Per-dimension
+    // sanity is already done in validate_header(); check_open() adds the global
+    // limits:* policy (channels/resolution/imagesize) and check_compression_ratio
+    // rejects a tiny file that declares a huge uncompressed composite.
+    if (ok && !m_specs.empty()) {
+        if (!check_open(m_specs[0], { 0, 300000, 0, 300000, 0, 1, 0, 56 })
+            || !check_compression_ratio(m_specs[0], ioproxy()->size())) {
+            close();
+            return false;
```
jinhgkim: Hmm, one question: do we want this guard here, or before `load_layers()`? My read is that we decompress the file in `load_layer_channel()`, which runs during `load_layers()`, and we want to avoid that. Correct me if I'm wrong!

lgritz: That's a good point. I'm not sure if we can do it before load_layers, but I'm really eyeing load_image_data as something we certainly want to do only after all the cheap early-outs. Let me take a look at all the various load_* functions and see how early I can push the check_open/check_compression_ratio, versus which of these read functions are needed for the information that's crucial to the check functionality.

lgritz: After some investigation, I think this can be arranged to happen right after load_header() and set_type_desc(), and before any of the other load_* functions get a chance to act on absurd header values. Update pushed, thanks for noticing this and making the suggestion!

Code review FTW!

jinhgkim: Glad it helped. I had to deal with a bunch of PSD/PSB decompression bombs in the past, so it caught my eye haha

lgritz:
How are we on this? Are there any further comments? Anybody want to give it an approval?

jinhgkim:
LGTM on my end.

PR #5350, opened 2026-07-30
fix(IBA): reorient() cases for transpose and transverse

### LINKED ISSUE #4661: [BUG] ImageBufAlgo::reorient() wrong orientation when src orientation is 5 & 7
**Describe the bug**

When src orientation is 5, ImageBufAlgo::reorient perform the pixel transformation by first rotate in 270 degree increments, then flop. This will result in a wrong orientation.
    ///     AC ------------> CD -----> DC
    ///     BD   rotate270   AB   flop   BA
```
    case 5:
        ok = ImageBufAlgo::rotate270(tmp, src, ROI(), nthreads);
        if (ok)
            ok = ImageBufAlgo::flop(dst, tmp, ROI(), nthreads);
```

The correct way should be first rotate in 90 degree increments, then flop:
```
    ///     AC ----------> BA -----> AB
    ///     BD   rotate90  DC  flop  CD
```

Also the correct way to transform from 7->1 should be
```
    ///     DB -----> CA -----------> AB
    ///     CA  flip  DB   rotate270  CD
```

2.5.11.

Also please tell us if there was anything unusual about your environment or
nonstandard build options you used.

**To Reproduce**

1. Do this...
2. Then this...
3. Then THIS happens (reproduce the exact error message if you can)
4. Whereas I expected this other specific thing to happen instead.

If the problem occurs in your C++ or Python code that uses the OIIO APIs, can
you also reproduce the problem using oiiotool? If so, please describe the
exact command line that reproduces the problem. (Being able to reproduce the
problem using only OIIO components makes it a lot easier for the developers
investigate and makes it clear it's not your application's fault.)

**Evidence**

- Error messages (paste them here exactly)
- Screenshots (if helpful)
- Example input: If the problem only happens with certain image files, please
  attach the smallest image you can make that reproduces the problem.

please just go straight to making a [pull request](https://github.com/AcademySoftwareFoundation/OpenImageIO/pulls).

luna-y-kim:
Orientations 5 and 7 were swapped: case 5 produced transverse and case 7 produced transpose. Fix both cases.

This adds reorient() tests covering all 8 values to the oiiotool-xform testsuite, and updates the python-imagebufalgo reorient() test to use the new test image for consistency.

Also, add orientation example images to documentation.

Fixes #4661

### Checklist:

  behavior.
  on the GitHub runners.)
- **(N/A)** If I added or modified a public C++ API call, I have also amended the


src/doc/stdmetadata.rst
```diff
@@ -89,9 +89,9 @@ Display hints
      3   rotated :math:`180^\circ` (bottom to top, right to left)
      4   flipped vertically (bottom to top, left to right)
      5   transposed (left to right, top to bottom)
-     6   rotated :math:`90^\circ` clockwise (right to left, top to bottom)
+     6   rotated :math:`90^\circ` counter-clockwise (right to left, top to bottom)
```
lgritz: Are you sure this description is correct?

If this is default orientation 1:
```
    0--->
    1--->
    2--->
```
and this is orientation 6, rows right to left and columns top to bottom:
```
   2 1 0
   | | |
   v v v
```
Wouldn't you say that the second pattern is rotated clockwise from the default?

lgritz:
The orientation codes are so inherently confusing to think about in words.
I really think that ultimately, we should illustrate this with a picture as well as the words (not necessarily on you to do in this PR).

I also think, if we don't have it, we really need a testsuite verification of reorient for EACH of the 8 values. I guess we don't, or the code change you made would either be failing tests, or be accompanied by updated reference images? I have to admit I'm afraid to accept this change in behavior without a visual test that proves that the new code definitely does the right thing.

luna-y-kim:
> The orientation codes are so inherently confusing to think about in words.
> I really think that ultimately, we should illustrate this with a picture as well as the words (not necessarily on you to do in this PR).

Yes, the wording is super confusing. A visual illustration sounds good. That'd be helpful.

Also, the intro paragraph in the display hints docs does define the ordering convention, but still it requires some eyeballing for me. For values like 6 and 8, I couldn't tell which part of '(right to left, top to bottom)' described what.

lgritz:
It feels like the right order of operations is:

1. Create an orientation=0 image that is small but clearly and visually shows the orientation. Then transform it to each of the 7 other orientations. We look at them and agree that we are all confident that all 8 exactly match what the TIFF spec says the orientation values mean.
2. Make a testsuite test that, for each of those 8, does a reorient() and ensures that the result exactly matches the orientation 0 we started with.
3. Use that testsuite to validate this PR and merge it.
4. Use those same images to produce a figure that augments the description of orientation in the stdmetadata.rst.

Without the visual example (and multiple people looking it over), I honestly don't trust myself to get this right just using the words and my head. It's really easy to make mistakes when reasoning about it.

lgritz:
Maybe something that looks like this:

```
+------------------------+
| O -------------------> |
| |               cols   |
| |                      |
| |r                     |
| |o      [LOGO]         |
| |w                     |
| |s                     |
| |                      |
| v                (1,1) |
+------------------------+

**Comment by lgritz** [comment]
oh no, did I already mix up rows and columns?

**Comment by lgritz** [comment]
It's even confusing visually :-)

Does "cols" mean "this line is an example of a column" or does it mean "column indices increase in this direction?"

**Comment by lgritz** [comment]
Better?

```
+------------------------+
| O ---row 0-----------> |
| |                width |
| |                      |
| |c                     |
| |o      [LOGO]         |
| |l                     |
| |0                     |
| |                      |
| v height         (1,1) |
+------------------------+
```

**Comment by lgritz** [comment]
I want to make something that nobody could possibly look at and still be confused. :-)

**Comment by luna-y-kim** [comment]
> Does "cols" mean "this line is an example of a column" or does it mean "column indices increase in this direction?"

That is exactly what confused me. 

```
                   Top
        +------------------------+
        | <--- col 0 --------- O |
        |                      | |
        |                      r |
        |                      o |
Left    |                      w |    Right
        |                      0 |
        |                      | |
        |                      | |
        |                      ↓ |
 (1,1)  +------------------------+
                 Bottom
```
An example of Orientation 6.
In addition to your sketch, top/bottom/left/right tags could be helpful.

**Comment by luna-y-kim** [comment]
> It feels like the right order of operations is:
> 
> 1. Create an orientation=0 image that is small but clearly and visually shows the orientation. Then transform it to each of the 7 other orientations. We look at them and agree that we are all confident that all 8 exactly match what the TIFF spec says the orientation values mean.
> 2. Make a testsuite test that, for each of those 8, does a reorient() and ensures that the result exactly matches the orientation 0 we started with.
> 3. Use that testsuite to validate this PR and merge it.
> 4. Use those same images to produce a figure that augments the description of orientation in the stdmetadata.rst.
> 
> Without the visual example (and multiple people looking it over), I honestly don't trust myself to get this right just using the words and my head. It's really easy to make mistakes when reasoning about it.

I think I should've done this before opening the PR. I thought I understood it correctly. Lesson learned. I'm going to work on step 1 and share soon!

**Comment by lgritz** [comment]
> In addition to your sketch, top/bottom/left/right tags could be helpful.

But not IN the image. If we're showing how each one should display, top is always top, etc. but the contents of the image rotate.

OH, I just realized I described the process incorrectly. (See? Confusing!) I think the correct description is:

1. Create a single small image that clearly shows origin, direction of rows and columns, etc., and make 7 more copies with the same pixel order but each of the other 7 "Orientation" metadata values. Shown in iv or some other viewer that tries to honor Orientation, we should clearly see each one in a different orientation and agree that they match what the TIFF spec describes.
2. Make a testsuite test that, for each of those 8 images, does a reorient() that results in an image with Orientation 0, but the pixels have actually been changed so that it matches what we saw in the viewer in (1). (The correct versions of this should be checked in as reference images.)
3. Use that testsuite to validate this PR and merge it.
4. Use those same reference images to produce a figure that augments the description of orientation in the stdmetadata.rst.

**Comment by luna-y-kim** [comment]
Screenshot_all_orient

Here is the screenshot of the 8 images. I created an image, duplicated it 7 times, and set the Orientation metadata to each value on the duplicates.

Interestingly, Apple Preview's inspector shows orientation 6 as "rotated 90° CCW" and 8 as "rotated 90° CW" (in the screenshot). I tried hard to wrap my head around it, and I think both wordings "rotated 90° CW" (OIIO) and "rotated 90° CCW" (Preview app) are right and are saying the same thing in different words. Preview app describes how the encoded data is rotated relative to the intended display (that is, in software that doesn't honor orientation metadata, the image appears to be rotated 90° CCW), and OIIO describes how it should be displayed (rotated 90° CW for 6, the configuration suggested by the Orientation metadata). Am I understanding the OIIO part right?

**Comment by lgritz** [comment]
Love it!

Now I need to stare at these with the TIFF spec open and convince myself that every one is correct.

The nomenclature is really hard. If the starting image is

```
A B
C D
```

And you asked me what this image was:

```
C A
D B
```

I would say that second image is "rotated 90 degrees clockwise" (relative to the starting image). Does that help?

**Comment by luna-y-kim** [comment]
OH, YES!! Super helpful. Thanks! The "what this image was" framing is the best part.

**Comment by lgritz** [comment]
For reference, the meaning of the orientation values, from the 1992 TIFF 6.0 spec:

TIFF-orientation

**Comment by lgritz** [comment]
I'm very confused by Apple's preview describing orientation 6 as "rotated counter-clockwise". That feels like the opposite of how I conceptualize it. I will look for other explanations of orientation in the other file format specs. (They all ultimately borrowed TIFF's meaning of the flags, but they might explain it in different ways. I notice that TIFF doesn't say anything about rotation, it just says which sides of the view are the 0th row and column.)

**Comment by lgritz** [comment]
I am convinced that the images you posted (that is, as displayed by Apple Preview) match the descriptions in the TIFF spec.

Now we just need to verify that OIIO is also doing the right thing when displaying with iv, and that reorient() works correctly in all cases (i.e. the reoriented result should LOOK just like those apple images, but say that they are orientation 1).

And then we need to clean up our own documentation to be less confusing/ambiguous.

**Comment by luna-y-kim** [comment]
Yep, sounds great! I'm going to work on the testsuite so that the results of reorient() look just the same as those intended orientations (the images in the screenshot), but with all images having Orientation=1 in the spec.

Also, I'll check the display in iv (haven't tried it yet, so I'd like to explore).

**Comment by luna-y-kim** [comment]
iv displays all 8 as expected, and for reference, here is a screenshot of the new reference images for the testsuite.

testsuite_ref_images

**Comment by luna-y-kim** [comment]
### Feedback wanted: (Feedback from anyone is welcome!)

For docs, would something like this image work? It makes sense to me, but I'm not sure if it reads as clearly to others.

Screenshot_docs

----------
Also, maybe a header like the one below would help?

Screenshot_docs2

**Comment by lgritz** [comment]
I'm not sure if it's easy given the documentation system, but I was imagining that the picture for each could be incorporated directly into a single table.  Like

    +---+-----------+-------------+
    |   |           |             |
    | 1 | [picture] | Explanation |
    |   |           |             |
    +---+-----------+-------------+
    |   |           |             |
    | 2 | [picture] | Explanation |
    |   |           |             |
    +---+-----------+-------------+


Also, given how much we have been confused by those textual descriptions, I wonder if it would be better to directly adopt the descriptions used in the TIFF spec. I need to check the various image file format specs (not just TIFF) that incorporate this encoding to see if any of them do a better job with explanations. I'm not sure I remember where I got the wording that we have now.

**Comment by luna-y-kim** [comment]
You mean something like this?

Screenshot 2026-08-05 at 11 43 34 PM

If this looks like the right direction, I'll update this PR with just this layout change for now, and leave the wording as is until you have a chance to check the other format specs. No rush on my side either way!

**Comment by luna-y-kim** [comment]
Or, it also works without the box, like this:
Screenshot 2026-08-05 at 11 50 57 PM

**Comment by lgritz** [comment]
Yes, I like that.

**Comment by luna-y-kim** [comment]
Great! Just to be sure: the version with the box (first screenshot) or without (second)? (Personally, the second one looks cleaner!)

And about the "right to left, top to bottom" style of wording (if that is what you meant): a very similar phrasing appears in `CineonHeader.h` and `DPXHeader.h`, I ended up reading those files while trying to understand the wording for this PR.

**Comment by lgritz** [comment]
I also prefer the second one.

**Comment by lgritz** [comment]
> And about the "right to left, top to bottom" style of wording (if that is what you meant): a very similar phrasing appears in `CineonHeader.h` and `DPXHeader.h`, I ended up reading those files while trying to understand the wording for this PR.

I'm not sure what I like best here, but I note that the wording from the original TIFF spec (which I posted above a few messages) is at least unambiguous, saying things like "the 0th row is the visual top of the image and the 0th column is the visual left side of the image." I'm thinking now that this is a lot better than talking about rotations, which are harder to discern which direction they mean (rotate memory to get appearance or rotate appearance to get the memory layout).

We should also probably explain, right before the table, that "row/column" refers to the order that the pixels are arranged *in memory*: starting with scanline/row 0, and within each scanline, starting with column 0. The orientation merely describes how that data is intended to be shown *visually* on a display.

It's tricky, so easy to get confused, and there is a real tension between wanting the explanations to be unambiguous and not open to misunderstanding, but not wanting to end with a long word salad that nobody understands.

**Comment by luna-y-kim** [comment]
> "row/column" refers to the order that the pixels are arranged _in memory_

I really like the "in memory" wording here. It clears up the vagueness of "0th row", and it also implicitly explains how software that doesn't honor Orientation will show the image. 

In addition to this part, maybe worth stating that "this means value 1 is displayed in memory order"? It can give readers a concrete anchor for what 'in memory' means in their mental model. (...Could this be part of a long word salad? or redundant?)

> The orientation merely describes how that data is intended to be shown _visually_ on a display.

When I read the TIFF spec for the first time, I was also confused by the wording "the visual right" together with "0th row" because neither term was defined clearly. So I think "_intended_ to be shown visually _on a display_" is great here.

**Comment by lgritz** [comment]
> In addition to this part, maybe worth stating that "this means value 1 is displayed in memory order"? 

I think maybe this lands us right back into ambiguity, since there is no particular reason to think a display will show the pixel with the lowest memory address at the visual "top". Many display and windowing systems over the years have considered the coordinate origin to be at the bottom of the screen.

**Comment by luna-y-kim** [comment]
Ah, right! I didn't think of bottom-origin coordinate systems. Thanks for pointing that out! There's more to consider than I thought.

File: testsuite/oiiotool-xform/run.py [hunk 2]
```diff
@@ -164,7 +167,8 @@ def make_test_pattern1 (filename, xres=288, yres=216) :
             "flipflop.tif", "flipflop-crop.tif",
             "rotate90.tif", "rotate90-crop.tif",
             "rotate270.tif", "rotate270-crop.tif",
-            "reorient1.tif", "reorient2.tif", "reorient3.tif",
+            "reorient1.tif", "reorient2.tif", "reorient3.tif", "reorient4.tif",
+            "reorient5.tif", "reorient6.tif", "reorient7.tif", "reorient8.tif",
```
lgritz, line 171: I'm slightly worried about the ways this can fail, since in order to support multiple references when they differ per-platform, we consider a test to pass if an output we are checking matches **any** of the reference images. So if orientation 2 were to break and started making the image that should be orientation 4 instead, I think this would still pass?

Here is an idea, what do you think?

After you make the reorient?.tif images, assemble them into one big image using `oiiotool --mosaic`, and compare THAT (not the individual reoriented images) to a reference image. Then any individual reorient images being wrong will certainly be a failure and can't pass by coincidentally matching a different reorientation.

luna-y-kim:
> I'm slightly worried about the ways this can fail, since in order to support multiple references when they differ per-platform, we consider a test to pass if an output we are checking matches **any** of the reference images.

If I understood correctly, I think this is what you're worrying about:

(testsuite/runtest.py:353-358)
```
# Check one output file against reference images in a list of reference
# directories. For each directory, it will first check for a match under
# the identical name, and if that fails, it will look for alternatives of
# the form "basename-*.ext" (or ANY match in the ref directory, if anymatch
# is True).
def checkref (name: str, refdirlist: list[str]) -> tuple[bool, str]:
```

The alternatives pattern is built from the whole name including its number (`reorient2-*.tif`), so no variant pattern can match `reorient2.tif` against another orientation's reference such as `reorient4.tif`. Only `anymatch = True` can.

But, as you worried, the image check itself did falsely pass in Test2 (with the name `reorient2-*.tif`) and Test4 (with `anymatch = True`) below. However, it was caught because out.txt didn't match with ref/out.txt, so the issue doesn't slip through.

Therefore, the current test is working. But now I'm worrying that I'm missing something!

**Your mosaic idea is cool. It would be harder to break if someone changes things later. I'm happy to switch to the mosaic version if you'd prefer it!**


----------
### Tests below are just for reference:

#### Test1:
* _Correct_ operation in `reorient()`
* _Removed_ `ref/reorient2.tif` file (so no file to compare in ref directory)

Test result:
```
comparisons are ['ref/reorient1.tif']
comparing reorient1.tif to ref/reorient1.tif
PASS: reorient1.tif matches ref/reorient1.tif
comparisons are ['ref/reorient2.tif']
NO MATCH for reorient2.tif
FAIL reorient2.tif
comparisons are ['ref/reorient3.tif']
comparing reorient3.tif to ref/reorient3.tif
PASS: reorient3.tif matches ref/reorient3.tif
```

----------

#### Test2:
* _Wrong_ operation in `reorient()`
* _Removed_ `ref/reorient2.tif` file
* `ref/reorient2-rotate180.tif` exists which is the same as `reorient4.tif`

(src/libOpenImageIO/imagebufalgo_orient.cpp:362)
changing case 2 to be flip instead of flop in `ImageBufAlgo::reorient()`:
```
// correct
// case 2: ok = ImageBufAlgo::flop(dst, src, ROI(), nthreads); break;

// test with wrong transform
case 2: ok = ImageBufAlgo::flip(dst, src, ROI(), nthreads); break;
```

Test result:
```
Diff was:
-------
--- out.txt
+++ ref/out.txt
@@ -82,7 +82,7 @@
 PASS
 Comparing "reorient1.tif" and "ref/reorient1.tif"
 PASS
-Comparing "reorient2.tif" and "ref/reorient2-rotate180.tif"
+Comparing "reorient2.tif" and "ref/reorient2.tif"
 PASS
 Comparing "reorient3.tif" and "ref/reorient3.tif"
 PASS
```

-> The image check itself did falsely pass, but it was caught in `out.txt` diff.

----------

#### Test3:
* _Wrong_ operation in `reorient()` (same as Test2)
* _Present and correct_ `ref/reorient2.tif` file

Test result:
```
Comparing "reorient1.tif" and "ref/reorient1.tif"
PASS
Comparing "reorient2.tif" and "ref/reorient2.tif"
240 x 160, 4 channels
  Mean error = 0.0527273
  RMS error = 0.189596
  Peak SNR = 14.4434
  Max error  = 1 @ (103, 12, R)  values are 1, 1, 1, 1 vs 0, 0, 0, 1
  6132 pixels (16%) over 0.008
  7922 pixels (20.6%) over 0.004
FAILURE
Comparing "reorient3.tif" and "ref/reorient3.tif"
PASS
```

----------

#### Test4:
* _Wrong_ operation in `reorient()` (same as Test2)
* _Removed_ `ref/reorient2.tif` file
* `anymatch = True`

Test result:
```
Diff was:
-------
--- out.txt
+++ ref/out.txt
@@ -82,7 +82,7 @@
 PASS
 Comparing "reorient1.tif" and "ref/reorient1.tif"
 PASS
-Comparing "reorient2.tif" and "ref/reorient4.tif"
+Comparing "reorient2.tif" and "ref/reorient2.tif"
 PASS
 Comparing "reorient3.tif" and "ref/reorient3.tif"
 PASS
```

-> The image check itself did falsely pass, but it was caught in `out.txt` diff.

----------

lgritz:
Ah, ok, so I guess the naming pattern is enough in this case. Great!

lgritz (approved):
LGTM!

lgritz:
Great work, in addition to the bug fixes, this is a big improvement in the documentation and the testing.

luna-y-kim:
Thank you for walking me through it. I learned a lot!

PR #5354, opened 2026-07-31
fix(filesystem): Use 64-bit file offsets on Windows

### LINKED ISSUE #5351: bug: Filesystem::fseek truncates offsets above 2 GiB in MinGW builds
**Describe the bug**

`Filesystem::fseek()` fails for file offsets above 2 GiB when OpenImageIO is built with MinGW-w64.

The function accepts an `int64_t` offset, but the implementation uses `_fseeki64()` only when `_MSC_VER` is defined:

https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/libutil/filesystem.cpp#L550-L570

```cpp
int
Filesystem::fseek(FILE* file, int64_t offset, int whence)
{
#ifdef _MSC_VER
    return _fseeki64(file, __int64(offset), whence);
#else
    return fseeko(file, offset, whence);
#endif
}
```

MinGW targets Windows but does not define `_MSC_VER`, so it uses the `fseeko()` branch. In the official MSYS2 MINGW64 build, this results in offsets above `INT32_MAX` failing.

This causes valid image files containing data beyond the 2 GiB boundary to be reported as corrupted.

This behavior also appears to conflict with the intent of PR #2399, which introduced these wrappers to provide consistently 64-bit file offsets:

https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/2399

The affected binary comes from the official MSYS2 package:

```text
mingw-w64-x86_64-openimageio 2.5.18.0-1
```

Output of `oiiotool --buildinfo`:

```text
OIIO 2.5.18.0 | Windows/x86_64
    Build compiler: gcc 14.2 | C++17/201703
Dependencies: Boost 1.87.0, BZip2 1.0.8, DCMTK 3.6.8, FFmpeg 7.0, fmt 11.1.3, Freetype 2.13.3, GIF 5.2.2, Libheif 1.19.5, libjpeg-turbo 3.1.0, LibRaw 0.21.3, OpenColorIO 2.4.1, OpenCV 4.11.0, OpenEXR 3.3.2, OpenGL, OpenJPEG 2.5, OpenVDB 12.0.0, PNG 1.6.46, Ptex 2.4, pugixml 1.15, pybind11 2.13.6, Python 3.12.8, Qt6 6.8.1, Robinmap, TBB 2022.0.0, TIFF 4.7.0, WebP 1.5.0, ZLIB 1.3.1
```

There were no custom build options or unusual environment settings. This is the binary distributed by the official MSYS2 MINGW64 repository.

The corresponding MSYS2 build recipe is available here:

https://github.com/msys2/MINGW-packages/blob/19060ba7d9a29b66200eb758299aee244754a923/mingw-w64-openimageio/PKGBUILD

**To reproduce**

The issue occurs with a valid uncompressed scanline OpenEXR image with the following properties:

```text
resolution:   16384 x 16384
channels:     Y, A
channel type: float
compression:  none
file size:    2147746125 bytes
```

The uncompressed pixel payload is exactly 2 GiB, so the final scanline chunks are located beyond the signed 32-bit file offset limit.

Run:

```powershell
$env:OPENIMAGEIO_OPTIONS='openexr:core=0'

C:\msys64\mingw64\bin\oiiotool.exe `
    --stats `
    C:\path\to\large-uncompressed.exr
```

The error is:

```text
(EXR_ERR_READ_IO) Unable to seek to desired offset 2147483965: File input failed.

oiiotool ERROR: stats : Failed OpenEXR read: Unable to query scanline information
```

With the default OpenEXR Core path, the same failure is reported as:

```text
Some scanline chunks were missing or corrupted
```

`oiiotool --info` succeeds because it reads the header without reading the pixel chunks located beyond 2 GiB.

**Expected behavior**

The file should be read successfully.

The same file is read successfully by:

- OpenEXR `exrmetrics` using OpenEXR 3.3.2;
- DJV 3.4.2 on Windows.

`exrmetrics` reads the complete image without reporting missing or corrupted chunks.

I can provide the reproducing EXR file if needed, although its size makes it impractical to attach directly to the issue.

**Suggested fix**

Since both MSVC and MinGW-w64 target Windows and provide `_fseeki64()` and `_ftelli64()`, these functions could be selected based on the target platform rather than only the compiler:

```cpp
int
Filesystem::fseek(FILE* file, int64_t offset, int whence)
{
#if defined(_WIN32)
    return _fseeki64(file, static_cast<__int64>(offset), whence);
#else
    return fseeko(file, static_cast<off_t>(offset), whence);
#endif
}

int64_t
Filesystem::ftell(FILE* file)
{
#if defined(_WIN32)
    return static_cast<int64_t>(_ftelli64(file));
#else
    return static_cast<int64_t>(ftello(file));
#endif
}
```

Alternatively, MinGW could be handled explicitly:

```cpp
#if defined(_MSC_VER) || defined(__MINGW32__)
```

Checking `_WIN32` may better reflect that this is a Windows platform distinction rather than a compiler distinction.

Maxi741cv:
### Description

Use `_fseeki64()` and `_ftelli64()` on all Windows builds. MinGW does not
define `_MSC_VER`, so it previously used `fseeko()` and `ftello()`, which can
have 32-bit offsets.

Fixes #5351.

### Tests

- `clang-format --dry-run --Werror src/libutil/filesystem.cpp`
- `git diff --check`
- Verified that `oiiotool --stats` reads the reproducing EXR successfully

### DIFF
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -552,7 +552,7 @@ int
 Filesystem::fseek(FILE* file, int64_t offset, int whence)
 {
 #ifdef _WIN32
-    return _fseeki64(file, __int64(offset), whence);
+    return _fseeki64(file, static_cast<__int64>(offset), whence);
 #else
     return fseeko(file, offset, whence);
 #endif
--- a/src/libutil/filesystem.cpp
+++ b/src/libutil/filesystem.cpp
@@ -551,7 +551,7 @@ Filesystem::fopen(string_view path, string_view mode)
 int
 Filesystem::fseek(FILE* file, int64_t offset, int whence)
 {
-#ifdef _MSC_VER
+#ifdef _WIN32
     return _fseeki64(file, __int64(offset), whence);
 #else
     return fseeko(file, offset, whence);
@@ -563,7 +563,7 @@ Filesystem::fseek(FILE* file, int64_t offset, int whence)
 int64_t
 Filesystem::ftell(FILE* file)
 {
-#ifdef _MSC_VER
+#ifdef _WIN32
     return _ftelli64(file);
 #else
     return ftello(file);


lgritz (approved):
LGTM, thanks for the fix

PR #5355, opened 2026-07-31
fix(IBA): detect iterator read errors in orientation functions

luna-y-kim:
Check the `ImageBuf::ConstIterator` error flag for IC images in orientation functions (flip, flop, rotate90/180/270, and transpose). Previously, such errors were silently ignored and the functions returned true. Now they return false if `ImageBuf::retile()` indicates an error during the read.
    
In `IBA::transpose()`, the error flag is a `std::atomic<bool>` because the multithreading helper `IBA::parallel_image()` is called.
    
Follow-up to #3233
    

### Checklist:

  behavior.
  on the GitHub runners.)
- [ ] **(N/A)** If I added or modified a public C++ API call, I have also amended the

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_orient.cpp
+++ b/src/libOpenImageIO/imagebufalgo_orient.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // https://github.com/AcademySoftwareFoundation/OpenImageIO
 
+#include <atomic>
 #include <cmath>
 #include <iostream>
 
@@ -33,7 +34,7 @@ flip_(ImageBuf& dst, const ImageBuf& src, ROI dst_roi, int /*nthreads*/)
         for (int c = dst_roi.chbegin; c < dst_roi.chend; ++c)
             d[c] = s[c];
     }
-    return true;
+    return !s.has_error();
 }
 
 
@@ -82,7 +83,7 @@ flop_(ImageBuf& dst, const ImageBuf& src, ROI dst_roi, int /*nthreads*/)
         for (int c = dst_roi.chbegin; c < dst_roi.chend; ++c)
             d[c] = s[c];
     }
-    return true;
+    return !s.has_error();
 }
 
 
@@ -153,7 +154,7 @@ rotate90_(ImageBuf& dst, const ImageBuf& src, ROI dst_roi, int /*nthreads*/)
         for (int c = dst_roi.chbegin; c < dst_roi.chend; ++c)
             d[c] = s[c];
     }
-    return true;
+    return !s.has_error();
 }
 
 
@@ -216,7 +217,7 @@ rotate180_(ImageBuf& dst, const ImageBuf& src, ROI dst_roi, int /*nthreads*/)
         for (int c = dst_roi.chbegin; c < dst_roi.chend; ++c)
             d[c] = s[c];
     }
-    return true;
+    return !s.has_error();
 }
 
 
@@ -267,7 +268,7 @@ rotate270_(ImageBuf& dst, const ImageBuf& src, ROI dst_roi, int /*nthreads*/)
         for (int c = dst_roi.chbegin; c < dst_roi.chend; ++c)
             d[c] = s[c];
     }
-    return true;
+    return !s.has_error();
 }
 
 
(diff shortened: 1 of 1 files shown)


src/libOpenImageIO/imagebufalgo_orient.cpp
```diff
@@ -410,8 +412,10 @@ transpose_(ImageBuf& dst, const ImageBuf& src, ROI roi, int nthreads)
             for (int c = roi.chbegin; c < roi.chend; ++c)
                 d[c] = s[c];
         }
+        if (s.has_error())
+            ok = false;
```
lgritz, line 416: This is fine, but just as a note (and please correct me if you think I'm wrong here): I don't think this needs to be atomic. It starts as true, the threads can only set it to false if they change it at all, it's never read until execution is single threaded again and beyond what has to be a synchronization barrier at the end of the threaded loop. So if two threads happen to execute `ok = false` at the same time, it doesn't really matter which one happens first or second, it still gets set to false.

In contrast, if it were an int and the threads were incrementing it (reading, adding one, then writing the result), in that case you would have a real race condition that would require an atomic operation.

luna-y-kim: I totally agree that multiple threads happening to execute `ok = false` at the same time doesn't matter for returning the correct value at the end, and the value is only read after the multithreaded work is done. Also in practice, I agree nothing can go wrong here, but as far as I understand, concurrent writes to the same non-atomic object is still a data race, even if all the writes are identical, like `ok = false` here. So my reasoning was it's good practice to use atomic (although it'd be super rare here, and harmless even if it occurred). In addition, it matches what `copy_pixels_impl` in imagebuf.cpp does. (Actually, that is where I got the idea that `IteratorBase` has the `m_readerror` member.)

I think this is more like a design matter, so I'm happy to change it to a plain bool if you prefer.

lgritz: Short answer: Technically, doing this without the atomic is C++ undefined behavior, so let's leave it. It's not enough of a performance-critical situation to try to split hairs about whether it's truly necessary.

Longer answer:

On both x86_64 and ARM/AArch64, a simple aligned store (a plain 'mov' instruction) is already atomic. You only need the bus lock for read-modify-write situations, or if the value is unaligned or longer than register length, and therefore might, underneath, require two actual memory writes to complete (obviously never true for a 1-byte bool).

Because all threads are writing the same value and none are reading it (until the return, which is outside the parallel loop and therefore after a memory barrier), I think this is not technically a data race.

But that's from the perspective of the **hardware**. It's still C++ UB, so it is a data race as far as the compiler is concerned, so we'll play this by the book.

lgritz (approved):
LGTM

src/libOpenImageIO/imagebufalgo_orient.cpp
luna-y-kim: Thanks for the explanation. The hardware perspective is interesting, I should definitely dig into it later.

One small follow-up (just to confirm my understanding as I'm new to this): 
It's write only (not read-modify-write), and a 1-byte bool store is a single step (half-write won't happen). So nothing can happen in the middle of the write, and that is why you said "a simple aligned store is already atomic" at the hardware level. Did I understand that correctly?

lgritz: Right. From the hardware perspective, as long as the value is register sized and aligned (as obviously a bool would have to be), just the plain write is already as atomic as it gets in terms of how memory is actually interacted with.

As you know, if it's fundamentally read-modify-write (RMW) such as an increment operation, then you have a race, because the sequence for threads A and B could be

    A: read (1, say)
    B: read (also 1)
    A: increment (now A's register is 2)
    B: increment (also 2)
    A: write the 2
    B: also write the 2

so you end up with 2, but if they hadn't been overlapping in time, it would be 3 because there were, in theory, two independent increments of the same variable.

RMW can sneak up on you sometimes. For example, if you have bitwise operations like

    struct S {
        int a : 4;
        int b : 4;
    }

Then `S.a = 1` looks like a pure write, but it is actually a RMW because it has to fetch the word, modify only certain bits, and then write it again.

lgritz: The bottom line is that you were right all along: std::atomic is the right way to do this from the C++ perspective. It's only if this were a highly-contended atomic in performance-critical code would we be looking for ways to bend the rules for the sake of avoiding the bus lock.

luna-y-kim: > It's only if this were a highly-contended atomic in performance-critical code would we be looking for ways to bend the rules for the sake of avoiding the bus lock.

Because when an atomic is highly contended, the RMW operations on the same atomic object have to take turns (waiting in line), so it can become a bottleneck!

Thank you again for taking the time to explain this, a new world just unlocked! I'll pay it forward with more contributions. 😉

PR #5357, opened 2026-08-01
fix(thumb): don't write `thumbnail_*` metadata to formats that can't hold a thumbnail

jinhgkim:
### Description

Noticed a thumbnail metadata leak in `iconvert` when converting from a format that supports thumbnails to one that doesn't. The `thumbnail_*` attributes ride along in the spec and get serialized, so the file claims a thumbnail it doesn't have.

`oiiotool` already stripped these, so I moved the stripping into `ImageOutput::check_open()` where it covers every writer and caller.

Also stopped `ImageBuf` treating those attributes as proof a thumbnail exists. Both read paths now take `m_has_thumbnail` from `get_thumbnail()`'s return. The ImageCache-backed path was ignoring the return entirely, so it and the direct path disagreed on the same file.

### Optional Read
One thought I had while working on this: currently we check whether a thumbnail exists by looking at its width and height in ImageBuf, which I think is unreliable. `rawinput` even hacks around this guard, since `twidth`/`theight` aren't always populated by libraw when the file does contain a thumbnail - something I only learned recently! I feel we need a more reliable guard: something like the `nsubimages` mechanism, but an `nthumbnails()` we could check instead. That would break ABI though, so we could save it for 3.3. Just wanted to mention it so I don't forget.

### Tests

Added `test_thumbnail_attribs()` that writes a spec claiming a thumbnail for every format that can't embed one and asserts it's absent on read back.

Assisted-by: Claude Code / Opus 5

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/libOpenImageIO/imagebuf.cpp
+++ b/src/libOpenImageIO/imagebuf.cpp
@@ -1207,8 +1207,11 @@ ImageBufImpl::init_spec(string_view filename, int subimage, int miplevel,
         if (m_spec["thumbnail_width"].get<int>()
             && m_spec["thumbnail_height"].get<int>()) {
             m_thumbnail.reset(new ImageBuf);
-            m_imagecache->get_thumbnail(m_name, *m_thumbnail, subimage);
-            m_has_thumbnail = true;
+            m_has_thumbnail = m_imagecache->get_thumbnail(m_name, *m_thumbnail,
+                                                          subimage);
+            // Don't keep a thumbnail buffer we failed to fill in.
+            if (!m_has_thumbnail)
+                m_thumbnail.reset();
         }
 
         // Subtlety: m_nativespec will have the true formats of the file, but
@@ -1291,6 +1294,9 @@ ImageBufImpl::init_spec(string_view filename, int subimage, int miplevel,
             m_thumbnail.reset(new ImageBuf);
             m_has_thumbnail = input->get_thumbnail(*m_thumbnail.get(),
                                                    subimage);
+            // Don't keep a thumbnail buffer we failed to fill in.
+            if (!m_has_thumbnail)
+                m_thumbnail.reset();
         }
 
         m_current_subimage = subimage;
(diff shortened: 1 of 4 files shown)


lgritz:
Most file formats do not support arbitrary named metadata, and for those, these orphaned metadata items without real thumbnails will be harmless.

I think you catch the cases of using an ImageBuf, and oiiotool, but I wonder if we should also try to catch cases where by using the APIs directly.

If somehow those attributes leak into the ImageSpec written to an output file that doesn't support thumbnails but DOES support arbitrary metadata, should we try to catch it? Are there any formats we support that have that combination (no thumb, yes named metadata), and somewhere in the ImageOutput subclass for those formats themselves, also squash the orphaned metadata?

I agree with you that eventually, we'd like all the thumbnail metadata to disappear and be replaced by a real API that can handle multiple thumbnails (among other improvements).

jinhgkim:
> I think you catch the cases of using an ImageBuf, and oiiotool, but I wonder if we should also try to catch cases where by using the APIs directly.

The stripping is now in `ImageOutput::check_open()`, where it covers every writer and caller. Would this suffice?

> If somehow those attributes leak into the ImageSpec written to an output file that doesn't support thumbnails but DOES support arbitrary metadata, should we try to catch it? Are there any formats we support that have that combination (no thumb, yes named metadata), and somewhere in the ImageOutput subclass for those formats themselves, also squash the orphaned metadata?

I think EXR doesn't support thumbnails (yet) but does support arbitrary metadata.

lgritz:
Yeah, check_open in ImageOutput should suffice, modulo the issue of whether every format writer actually uses it! (Much like my recent attempt to add the input check_open to all the readers that didn't yet have it added, we will need to do the same, most likely.)

I believe the OpenEXR *file format* does support thumbnails (do I remember that right?), but it's possible that they are so rarely encountered in practice that we never got around to adding them to the OIIO exr reader/writer.

jinhgkim:
Yup, OpenEXR does support thumbnails - one per subimage. Not sure if they are used in practice.

lgritz (approved):
LGTM

PR #5360, opened 2026-08-02
build: fix declaration mismatch for unity builds on Windows

lgritz:
Windows Unity builds might combine exrinput.cpp and exrinput_c.cpp into one translation unit. The forward declaration in exrinput.cpp lacked OIIO_EXPORT while the definition in exrinput_c.cpp has it, causing MSVC error. I'm not sure why this never failed before today!


grdanny (approved):
so we are just matching this with the actual definition that's in exrinput_c?
sounds reasonable...

lgritz:
> so we are just matching this with the actual definition that's in exrinput_c? 

Yes, exactly. This macro turns into nothing on non-Windows platforms. But on Windows, it needs to match.

PR #5361, opened 2026-08-03
build(deps): keep Homebrew out of local dependency child builds

zachlewis:
## Motivation

`IGNORE_HOMEBREWED_DEPS=ON` prunes the Homebrew prefixes from the search paths and sets `CMAKE_IGNORE_PATH`, but `CMAKE_IGNORE_PATH` is not honored by config-package searches — and the local dependency child builds launched by `build_dependency_with_cmake` can therefore quietly resolve Homebrew packages that the parent build is deliberately ignoring.

Concrete failure: with Homebrew Imath 3.2 installed, a local OpenEXR dependency build resolved `Imath_DIR=/opt/homebrew/lib/cmake/Imath` while OpenImageIO itself compiled against the locally built Imath 3.1.10. Imath's versioned namespace is baked into every mangled symbol, so the mismatch (`Imath_3_2` vs `Imath_3_1`) surfaced as unresolved symbols when linking libOpenImageIO. (This is an instance of the auto-build "sharp edges" discussed on #5313.)

## What this does

- In the `IGNORE_HOMEBREWED_DEPS` block, also append the Homebrew prefixes to `CMAKE_IGNORE_PREFIX_PATH` (CMake ≥ 3.23), which config-package searches do honor.
- Forward `CMAKE_IGNORE_PREFIX_PATH` to dependency child builds alongside the existing `CMAKE_IGNORE_PATH` forwarding.

No behavior change when `IGNORE_HOMEBREWED_DEPS` is off and `CMAKE_IGNORE_PREFIX_PATH` is unset — both hunks are conditional on the variable being non-empty.

## Testing

On macOS (arm64, Homebrew Imath 3.2.2 present), `IGNORE_HOMEBREWED_DEPS=ON` + `OpenImageIO_BUILD_MISSING_DEPS=all`:

- before: OpenEXR child build cached `Imath_DIR=/opt/homebrew/lib/cmake/Imath`; libOpenImageIO failed to link with `Imath_3_2`/`Imath_3_1` symbol mismatches;
- after: child build resolves the local deps-dist Imath and libOpenImageIO links clean.

Assisted-by: Claude Code (Fable 5)


lgritz:
This is still marked draft. Is it ready for final review?

zachlewis:
I'm not sure. I think it's doing the right thing, which is causing something to break elsewhere; but I've been submitting draft PRs for dependencies in a certain order so I can make sure that the libraries linkers link when linkers link linked libraries link linked libraries... uh... its a buffalo buffalo buffalo world out there.  


I'm not sure if I'm ready for you to review things that are breaking  in #5362, but I'm pretty sure this is breaking things that should break correctly.

lgritz (approved):
OK, I guess? I can't say I honestly understand exactly how this was broken before, but it passes CI, so I'll take your word for me.

PR #5362, opened 2026-08-03
build(deps): add local build recipe for JPEG XL (libjxl)

zachlewis:
## Motivation

libjxl has no local-build recipe, so `OpenImageIO_BUILD_MISSING_DEPS` can't supply the JPEG XL plugin's dependency the way it can for TIFF, PNG, OpenColorIO, etc. Adding one was historically blocked by #4439: once a static libtiff lands in the deps cache, no incremental reconfigure of that build dir can complete, which kills the iteration loop that recipe development (and persistent-build-dir wheel building) requires. #5313 fixed that, and this recipe is the follow-through.

## What this does

- Adds a `GIT_SUBMODULES` argument to `build_dependency_with_cmake`: shallow-initializes the named submodule paths after the main checkout. The submodule commits are the gitlinks pinned by the verified superproject commit, so they inherit the existing tag/commit verification.
- Adds `src/cmake/build_JXL.cmake`: builds libjxl 0.12.0 (pinned tag + commit) with its vendored brotli/highway/skcms submodules. Tools, tests, docs, and optional integrations are disabled; color management uses the bundled skcms, so no system lcms2 is needed. Like the other recipes, it follows `LOCAL_BUILD_SHARED_LIBS_DEFAULT` (static, except on Windows).
- `FindJXL`: a static libjxl doesn't carry its dependencies, so when `libjxl.a` is found, the `jxl_cms`, brotli, and `hwy` archives installed beside it are linked too. A shared libjxl links shared `jxl_cms` and brotli libraries, so the recipe sets its rpath to its own directory (`$ORIGIN` / `@loader_path`) and installs the brotli libraries alongside it.
- Clears the cached `JXL_THREADS_LIBRARY` before the refind. The generic refind cleanup doesn't cover that name, so a rejected system `libjxl_threads` (e.g. Ubuntu's 0.7) otherwise stays on the link line and drags the old libjxl in with it.

## Stacking note

The last commit is #5361 (needed to link on Homebrew machines with a mismatched Imath). Happy to drop it here once that lands.

No HEIF/AVIF recipes: a Libheif/libde265/aom chain works in testing, but shipping locally built HEVC support needs license/patent provisions worked out first.

## Testing

- macOS arm64 wheel, built like the CI wheels (`LINKSTATIC=1`, `IGNORE_HOMEBREWED_DEPS=1`, all dependencies built locally): JXL 0.12.0 is built and linked statically, `delocate-wheel` succeeds with no JXL/brotli dylibs to bundle, and the installed wheel round-trips a `.jxl` image. An incremental reconfigure of the same build dir rediscovers JXL without rebuilding.
- Ubuntu 24.04 with the distro's libjxl 0.7 installed: 0.7 is rejected, 0.12.0 is built locally and re-found, and `oiiotool` links against it (no system libjxl in `ldd`) and round-trips a `.jxl` image.
- Ubuntu 22.04 with the distro's brotli 1.0.9 installed and no system libjxl: builds and links clean, `oiiotool` round-trips a `.jxl` image.
- Shared on Linux/macOS (`-DJXL_BUILD_SHARED_LIBS=ON`, opt-in): both Ubuntu setups above link and round-trip, with `libjxl`, `jxl_cms`, and brotli resolved from the local deps dir (`RUNPATH [$ORIGIN]`); the macOS wheel bundles them and passes `delocate-wheel`.
- CI: all jobs pass. On Windows (shared by default) both the VS2022 build and the wheels build JXL 0.12.0 locally and install the brotli DLLs alongside it; the "all local builds" job and the Linux/macOS wheels exercise the static path.

Assisted-by: Claude Code (Fable 5, Opus 5)
Assisted-by: GitHub Copilot Autofix


lgritz:
I pushed one more minor change on top of yours -- removing "JXL" from the list of optional dependencies in the ci.yml recipes, to test that when it was considered required, that your auto-builder would for-sure work. It did!

lgritz (approved):
LGTM!

Will merge assuming it passes CI after my minor amendment.

zachlewis:
Nice! I've got a few more of these up my sleeve that I've been meaning to contribute after the color stuff. 

Thanks again for doing whatever it was that you did that fixed the problems I was having with the build system + previously-built TIFf / Deflate -- that was a real blocker for me when I started putting recipes together for some of the heavier dependencies (libheif etc), but I've since revived lots of that work... next step is formalizing and implementing a way to opt in to auto-building static gpl dependencies, and/or building and packaging shared libraries for certain dependencies in our wheels builds...

PR #5364, opened 2026-08-03
fix(dicom): null-check DCMTK pixel-data accessors before use

lgritz:
DicomImage::getInterData() and getOutputData() can return null when a corrupt DICOM's frame fails to decode, but their results were used without checking.

### DIFF
--- a/src/dicom.imageio/dicominput.cpp
+++ b/src/dicom.imageio/dicominput.cpp
@@ -147,8 +147,7 @@ DICOMInput::seek_subimage(int subimage, int miplevel)
 
     if (subimage < m_subimage) {
         // Want an earlier subimage, Easier to close and start again
-        close();
-        m_subimage = -1;
+        close();  // note: resets m_subimge to -1
     }
 
     // Open if it's not already opened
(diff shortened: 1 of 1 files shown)


src/dicom.imageio/dicominput.cpp
jinhgkim: nit: redundant, already set in close()

src/dicom.imageio/dicominput.cpp
```diff
@@ -183,7 +183,12 @@ DICOMInput::seek_subimage(int subimage, int miplevel)
...
     }
 
-    m_dipixel             = m_img->getInterData();
+    m_dipixel = m_img->getInterData();
+    if (!m_dipixel) {
+        errorfmt("Unable to read pixel data from DICOM file {}", m_filename);
+        m_img.reset();
+        return false;
+    }
     m_internal_data       = (const char*)m_dipixel->getData();
```
jinhgkim: nit: this assignment can be removed since it's being overwritten by line 204.

jinhgkim (approved):
LGTM, just a few nits I encountered while reviewing...

src/dicom.imageio/dicominput.cpp
lgritz: quite right

src/dicom.imageio/dicominput.cpp
lgritz: Thanks for those patches, which were not lines changed by this PR, but have been there for many years!

PR #5365, opened 2026-08-03
fix(ptex): guard null face-data returns; add malformed-header tests

lgritz:
PtexTexture::getData() and PtexFaceData::getTile() return null when the Ptex reader enters an error state during lazy per-face reads (a corrupt or truncated per-face block whose size the 64-byte header validator cannot catch). seek_subimage() and read_native_tile() dereferenced these unchecked, crashing on such files. Null-check both and hard-error cleanly, with release() cleanup on the tile path.

Add malformed-header regression fixtures (truncated, bad magic, inconsistent level info, metadata/level-data bombs, absurd face count) plus a generator; run.py reads each expecting clean rejection. testsuite/ptex/src already feeds the fuzzer corpus.


lgritz:
No objections for a week, bot had nothing to add -> merging

PR #5366, opened 2026-08-03
fix(gif): reject a short graphics-control extension block

lgritz:
read_gif_extension() read ext[1..4] from the graphics-control block without checking ext[0], the sub-block byte count. A corrupt file can declare fewer than the 4 bytes a well-formed block carries, so bail before reading past the giflib-provided buffer.


lgritz:
Simple fix, no comments for a week, bot examines without making any suggestions -> merge.

PR #5368, opened 2026-08-04
int: improve ImageSpec::default_channel_names -- nchannels safety

lgritz:
Since ImageSpec::default_channel_names does an allocation (`channelnames.reserve(nchannels)`), it seems prudent to do a validity check on nchannels first.

This is similar in spirit to check_open, but of course this happens during construction of the ImageSpec, whereas check_open can't do its magic until after the ImageSpec already exists and has done this allocation.

I also took the opportunity to change a change a push_back to an emplace_back, possibly saving an occasional std::string allocation.

### DIFF
--- a/src/libOpenImageIO/formatspec.cpp
+++ b/src/libOpenImageIO/formatspec.cpp
@@ -197,7 +197,7 @@ ImageSpec::default_channel_names() noexcept
     channelnames.clear();
     alpha_channel = -1;
     z_channel     = -1;
-    if (nchannels < 1 || nchannels >= limit_channels)
+    if (nchannels < 1 || (limit_channels && nchannels > limit_channels))
         return;  // early out for invalid channel counts
     channelnames.reserve(nchannels);
     if (nchannels == 1) {  // Special case: 1-channel is named "Y"
--- a/src/libOpenImageIO/formatspec.cpp
+++ b/src/libOpenImageIO/formatspec.cpp
@@ -195,9 +195,11 @@ void
 ImageSpec::default_channel_names() noexcept
 {
     channelnames.clear();
-    channelnames.reserve(nchannels);
     alpha_channel = -1;
     z_channel     = -1;
+    if (nchannels < 1 || nchannels >= limit_channels)
+        return;  // early out for invalid channel counts
+    channelnames.reserve(nchannels);
     if (nchannels == 1) {  // Special case: 1-channel is named "Y"
         channelnames.emplace_back("Y");
         return;
@@ -214,7 +216,7 @@ ImageSpec::default_channel_names() noexcept
         alpha_channel = 3;
     }
     for (int c = 4; c < nchannels; ++c)
-        channelnames.push_back(Strutil::fmt::format("channel{}", c));
+        channelnames.emplace_back(Strutil::fmt::format("channel{}", c));
 }
 
 


src/libOpenImageIO/formatspec.cpp
```diff
@@ -195,9 +195,11 @@ void
 ImageSpec::default_channel_names() noexcept
 {
     channelnames.clear();
-    channelnames.reserve(nchannels);
     alpha_channel = -1;
     z_channel     = -1;
+    if (nchannels < 1 || nchannels >= limit_channels)
```
jinhgkim:
```suggestion
    if (nchannels < 1 || (limit_channels && nchannels > limit_channels))
```
limit_channels = 0 means there's no limit to the nchannels so we want this additional check I suppose. Mimicking [check_open](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/libOpenImageIO/imageinput.cpp#L1691-L1696). Also, I don't know if we want to add an error message and reset nchannels before returning.

jinhgkim: Hmm, with my suggested change, it seems we are repeating the same channel count check - one in `default_channel_names` and one in `check_open`. Do you think it would be better to validate the dimensions before building the imagespec, so that we can get rid of this guard, keep default_channel_names() simple, and know the data is valid by the time we build the spec?

So, open() -> lightweight dimension check -> build ImageSpec -> check open / compression ratio ...

I think the dilemma we have is:
when we open() an image -> we want to check_open() and do all sorts of dimensions checks beforehand -> but check_open needs an ImageSpec() -> so we build IS -> but IS internally calls default_channel_names() -> we end up building channel names before the channel count check ... and ...
only after nchannels strings have been allocated do we finally check nchannels in check_open.

We could consider something like,
```cpp
// validate image dimensions and channel count
bool check_dimensions (int width, int height, int depth, int nchannels, ROI range, uint64_t flags);

bool check_open() {
    if (!check_dimensions())
       return false;
    // validate the rest that needs more data from ImageSpec
}
```
but I also think that adds unnecessary complexity to the code.

I'm not sure how strict we want to be about this. Is an invalid nchannels really a thing, though? I think the only outlier is nullreader.

lgritz: I think it's fine to do the comparison twice. It's a couple int compares per opened file.

If we try to slice it too thin to avoid any redundant computation at all, I worry that we could be somehow be less robust to ways we haven't thought of that it could be modified between construction and the call to check_open().

In fact, remember that this fix is ONLY for default_channel_names(). There's nothing stopping a reader from creating a baseline ImageSpec, then based on more detailed header information changing the channel count before calling check_open.

I see what you mean about check_dimensions. Yeah, that might have been a good way to make the tests a little more modular, but I'm not sure it's worth changing things now. But something to keep in mind if we continue to find more edge cases that make doing everything in check_open awkward or redundant.

lgritz: Thanks for noticing the "limit_channels == 0" edge case, that had slipped my mind.

jinhgkim: Thanks for clarifying, I was overthinking this.

lgritz:
How is this looking to you now, @jinhgkim?

jinhgkim:
looks good to me, thanks

lgritz:
OK, I'll consider that an approval

jinhgkim:
Just so you know, on the reviewer side, there’s no way to approve a PR after the initial review unless the PR author explicitly re-requests it.

lgritz:
Really? If you go to the "files changed" tab, you don't see a green box for "submit review" where you can click approve? (On a PR that's still open, I mean.)

jinhgkim:
I'm blind, my bad

PR #5370, opened 2026-08-05
fix(fits): validate the NAXIS count

lgritz:
Reject a NAXIS value outside the FITS-permitted 0-999 range before using it to size m_naxis, so a negative or huge count cannot wrap or blow up the resize. Also fail cleanly on an unsupported BITPIX (UNKNOWN format) for a non-empty image rather than proceeding with a zero-sized pixel type.

Also add a test case for the (already added in a prior PR) decompression bomb check.


lgritz:
Any comments on this? Seems pretty straightforward and not able to break anything. 
I appear to have some subsequent FITS fixes coming later, so I wanted to get this merged first.

jinhgkim (approved):
Seems straightforward indeed

PR #5372, opened 2026-08-06
fix(zfile): check gzread returns for truncated or corrupt input

lgritz:
The reader ignored gzread()'s return value when reading the header and each scanline, so a truncated or corrupt (gzip) file was processed with whatever happened to be in the buffer. Verify the requested number of bytes was actually read and error out otherwise.

Also add a test for the decompression bomb checks that have already been added in a prior PR.


lgritz:
Straightforward small fix -- check return codes for errors that we neglected to check before.
Any comments or objections?

brechtvl (approved):

PR #5376, opened 2026-08-06
fix(jpeg): more jpeg detection of corruptions

lgritz:
Add tests for jpeg detection of decompression bombs, both for regular
and progressive JPEGs.

Improvement: check_open() and check_compression_ratio() ran after
jpeg_start_decompress()/jpeg_read_coefficients(), which is too late
for progressive JPEGs. Move the checks before those calls.


testsuite/jpeg-corrupt/run.py
```diff
@@ -45,3 +45,9 @@
 
 # This file had corrupted IPTC data
 command += oiiotool("-oiioattrib imageinput:strict 1 -info -v src/corrupt-iptc-8011.jpg")
+
+# This file's SOF0 header declares a ~12 GB image (65500x65500x3) but holds
```
jinhgkim, line 49: Sorry, I must have missed this in the previous PR (the one with a bunch of check_open guards), but JPEG reader actually allocates the buffer before the `check_open/check_compression_ratio` guards. https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/jpeg.imageio/jpeginput.cpp#L267-L286

Did a quick research on the SOF0 header (baseline), and the reason why this test passes (catches the error correctly) is because at buffer alloc time it only allocates a small buffer, sized by width, so height never comes into it - so it actually only uses `65500` instead of `65500x65500`. But for SOF2 (progressive), it allocates the full res before we ever reach the guards. 

I may have some of the details wrong since I just came across this, but I think it'd be beneficial to add a progressive-encoded JPEG as another test case? It might also be worth repositioning the guards to before the call to `jpeg_start_decompress()` if the new test is actually failing.

lgritz: Yeah, that makes sense. If I don't have any progressive jpegs as decompression bomb test cases, I would not have discovered that this is too late. Will fix, thanks for the tip.

jinhgkim: wouldn't this be caught by the fuzzer eventually?

lgritz: Probably, yeah. But may as well fix it now.

lgritz: I just meant, lack of an example in progressive mode is why in practice I didn't discover the problem in the last batch of changes that tried to solve decompression bombs.

lgritz: Updated code and the description

jinhgkim (approved):

PR #5377, opened 2026-08-07
fix(oiiotool): don't dereference specs of a failed read

lgritz:
Add "null reader" test cases for invalid procedural specifiers (negative or huge CHANNELS, degenerate RES, negative TILE).

That was all I was originally trying to do, but it crashed oiiotool under ASan.

The null reader claims supports("procedural"), so input_file() skips the missing-file path and proceeds to read even when the query string is invalid. The read then fails, leaving the ImageRec with no subimages, and Oiiotool::error() only aborts the ArgParse loop -- it does not return from input_file(). Two sites then dereferenced the nonexistent spec:

* Oiiotool::read() bound `(*img)().nativespec()` before testing the read's return value, indexing an empty m_subimages.
* input_file() ignored the read result entirely and dereferenced curimg->nativespec(), which is null when there are no subimages.

Both read wrappers now report the error and return false if the read failed or produced no subimages, and input_file() breaks out on failure.

Assisted-by: GitHub Copilot / Claude Opus 4.8 (for help generating the new null test cases)

Assisted-by: Claude Code / Claude Opus 5 (for help diagnosing why it was crashing oiiotool)

### DIFF
--- a/testsuite/null/ref/out.txt
+++ b/testsuite/null/ref/out.txt
@@ -36,3 +36,18 @@ bar.null?RES=128x128&CHANNELS=3&TILE=64x64&TEX=1&TYPE=uint16&PIXEL=0.25,0.5,1 : 
       Constant Color: 16384.00 32768.00 65535.00 (of 65535)
       Monochrome: No
 Writing outtile.null
+oiiotool ERROR: read : file not found: "bad.null?CHANNELS=-1"
+Full command line was:
+> oiiotool --info bad.null?CHANNELS=-1
+oiiotool ERROR: read : file not found: "bad.null?CHANNELS=2000000000"
+Full command line was:
+> oiiotool --info bad.null?CHANNELS=2000000000
+oiiotool ERROR: read : file not found: "bad.null?RES=-4x-4"
+Full command line was:
+> oiiotool --info bad.null?RES=-4x-4
+oiiotool ERROR: read : file not found: "bad.null?RES=0x0"
+Full command line was:
+> oiiotool --info bad.null?RES=0x0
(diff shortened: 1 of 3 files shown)


testsuite/null/ref/out.txt
```diff
@@ -36,3 +36,18 @@ bar.null?RES=128x128&CHANNELS=3&TILE=64x64&TEX=1&TYPE=uint16&PIXEL=0.25,0.5,1 :
       Constant Color: 16384.00 32768.00 65535.00 (of 65535)
       Monochrome: No
 Writing outtile.null
+oiiotool ERROR: read : file not found: "bad.null?CHANNELS=-1"
+Full command line was:
+> oiiotool --info bad.null?CHANNELS=-1
+oiiotool ERROR: read : file not found: "bad.null?CHANNELS=2000000000"
+Full command line was:
```
jinhgkim, line 43: Question: Are these "file not found" error messages match what we expect for the corrupted files? These files should be caught in `check_open` and I recall we had something more informative for bad inputs.

lgritz: Yes, I had noticed that, too, and essentially it's a known oiiotool problem (not unique to this test, nor related to the 'null' reader). I'd like to address that as a separate follow-on PR (which will probably change this, and maybe other, test output, but that's fine).

lgritz: So let's evaluate this PR strictly in terms of "does it catch more errors that can happen with the null reader", and then there will be a separate PR that attacks the issue of "sometimes oiiotool loses specific error messages from a reader and just prints a generic 'can't open'".

jinhgkim: Alright, sounds good.

jinhgkim (approved):

PR #5380, opened 2026-08-08
fix(oiiotool): don't discard the reader's specific error message

lgritz:
In oiiotool, a failed read, where no reader could successfully open the file, was reporting a generic "file not found" even when the format reader had explained exactly what was wrong. For example,

    oiiotool ERROR: read : file not found: "bad.null?RES=-4x-4"

where iinfo, going straight through ImageInput, reported the real reason:

    null image resolution must be at least 1x1, but the file specified
    -4x-4x1, 4 chans. Possible corrupt input?

The message survives the "try every reader" logic; ImageInput::create() stashes the best-guess plugin's complaint and returns it if all other readers also fail to open. But it was thrown away when ImageRec::read() and read_nativespec() checked get_image_info()'s return value but never called ImageCache::geterror(), substituting a fixed string.  They now retrieve it, falling back to the generic text only if it is empty.

Also fix ImageCacheFile::open(), which took the trust_file_extensions shortcut on exactly the wrong branch and passed the still-empty `fmt` to Filesystem::extension() instead of the filename. With trust_file_extensions set, every REST-ful name failed with "ImageInput::create() called with no filename", clobbering the reader's message before the cache ever saw it.

This replaces some test output where the error messages have improved.

### DIFF
--- a/src/libtexture/imagecache.cpp
+++ b/src/libtexture/imagecache.cpp
@@ -767,8 +767,8 @@ ImageCacheFile::open(ImageCachePerThreadInfo* thread_info)
         // an unnecessary file open.
         std::string fmt;
         if (m_imagecache.trust_file_extensions()
-            && m_filename.find('?') != m_filename.npos)
-            fmt = OIIO::Filesystem::extension(fmt, false);
+            && m_filename.find('?') == m_filename.npos)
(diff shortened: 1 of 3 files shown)


jinhgkim (approved):
Looks good to me!

PR #5383, opened 2026-08-09
deps: Raise LibRaw recommended min to 0.21.3 (hard min still 0.20.0)

### LINKED ISSUE #5382: build: raise minimum LibRaw to 0.21.2
## Summary

Please consider raising the minimum LibRaw version from **0.20** to **0.21.2**.

LibRaw **CVE-2023-1729** (heap buffer overflow in `raw2image` / legacy `color4_image` path) is fixed in **0.21.2**.  
OpenImageIO currently accepts LibRaw ≥ 0.20 (`INSTALL.md` / `#4217`) and passes through `raw:half_size`. With a vulnerable LibRaw (e.g. 0.21.1), enabling half-size decoding can trigger that upstream bug:

```bash
oiiotool --iconfig raw:half_size 1 --info <poc.dng>

lgritz:
Fixes #5382

We don't want to raise the true min this high, since it's less than 3 years old. But we can raise the *recommended* minimum to trigger a build-time warning reminding people that newer versions have security fixes they might want.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -53,7 +53,7 @@ NEW or CHANGED MINIMUM dependencies since the last major release are **bold**.
  * If you want support for PNG files:
      * libPNG >= 1.6.0 (tested though 1.6.58)
  * If you want support for camera "RAW" formats:
-     * LibRaw >= 0.20 (tested though 0.22.2 and master)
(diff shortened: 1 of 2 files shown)


antond-weta (approved):
Looks good

PR #5385, opened 2026-08-10
fix(fits): read/write color FITS image using the correct convention

### LINKED ISSUE #5384: bug: Wrong channel count when reading FITS image
**The Bug**

A program calling `OIIO::ImageInput::Open` on an image file and then printing the image dimensions and channel count reports a channel count equal to 1 for a 2-D, 3-channel FITS color image, but a channel count equal to 3 or 4 for randomly selected JPG, TIFF, or PNG color images.  I expected OIIO to report a channel count of 3 for the FITS file.

If I allocate memory for `height`×`width`×`nchannels` data values then that is adequate for the JPG, TIFF, or PNG color images but insufficient for the FITS color image.  If I pretend that `nchannels=3` for the FITS file and allocate memory and read the image data on that basis then the resulting color image looks fine and I don't run into a memory violation error, so OIIO reads 3 values per pixel, not just 1 (as the value of `nchannels` suggests).

The FITS header of the FITS file begins with (newlines inserted for clarity):

```
SIMPLE  =                    T / file does conform to FITS standard             
BITPIX  =                  -32 / number of bits per data pixel                  
NAXIS   =                    3 / number of data axes                            
NAXIS1  =                  321 / length of data axis 1                          
NAXIS2  =                  300 / length of data axis 2                          
NAXIS3  =                    3 / length of data axis 3                          
EXTEND  =                    T / FITS dataset may contain extensions            
```

```
OIIO 3.1.15.0 | Linux/x86_64
    Build compiler: gcc 16.1 | C++17/201703
Dependencies: BZip2 1.0.8, DCMTK 3.6.9, FFmpeg NONE, fmt 11.2.0, Freetype 2.14.3, GIF 6.1.3, Imath 3.1.12, JXL
    0.11.2, Libheif 1.21.2, libjpeg-turbo 3.1.3, LibRaw 0.22.1, libuhdr 1.4.0, OpenColorIO 2.4.2, OpenCV 4.13.0,
    OpenEXR 3.2.4, OpenGL, OpenJPEG 2.5.4, openjph 0.25.3, OpenVDB 13.0.0, PNG 1.6.58, Ptex 2.5, pugixml 1.16,
    pybind11 3.0.4, Python3 3.14.6, Qt6 6.11.1, Robinmap, TBB 2020.3, TIFF 4.7.1, WebP 1.6.0, ZLIB 1.3.1
```

I built the source using `g++ -std=c++23 try.cc -o try -lOpenImageIO -lOpenImageIO_Util -lc`.

**To Reproduce**

`iinfo path-to-the-fits-file` yields: 

```
path-to-the-fits-file :  321 x  300 x    3, 1 channel, volume float fits
```

where `iinfo path-to-a-png-file` yields

```
path-to-a-png-file : 1422 x  804, 4 channel, uint8 png
```

so it appears that OIIO doesn't see the FITS contents as a 3-channel 2-D image but as a 1-channel 3-D volume, or perhaps as a sequence of 3 1-channel 2-D images.  But https://openimageio.readthedocs.io/en/latest/builtinplugins.html says

```
Currently, OpenImageIO only supports 2D FITS data (images), not 3D (volume) data
```

1. Compile the attach _try.cc_
2. Run it on the attached FITS image: `try small.fits`.  The output is

```
File         : small.fits
Channel count: 1
Width        : 321
Height       : 300
```

NOTE: The "Paste, drop, or click to add files" feature of the "Create new issue" web page keeps failing to upload my FITS file, even when I've wrapped it in a tgz file.  It also fails to upload my C++ file.  When I try to upload it with its original name "try.cc" then I get a complaint that the file type isn't supported, and when I try to upload it after renaming it to "try.cpp" then I get 

```
<!-- Failed to upload "try.cpp" -->
```

inserted into the text field where I'm writing this.  I'll try again after creating the issue.

lgritz:
FITS has no formal notion of color channels, just numbered axes, and OIIO's guess for a 3D color image (NAXIS1 = nchannels, interleaved) was wrong: research while investigating issue #5384 turned up no real-world FITS software using that layout. The universal convention is NAXIS3 = nchannels, one full-resolution plane per channel, the same layout astronomical software uses for spectral and Stokes cubes.

Reader and writer now both use only that convention. The old NAXIS1-interleaved guess is dropped entirely rather than kept for compatibility: nothing but OIIO itself ever produced or consumed it, and it silently misread genuine external color cubes as volumes, which was the actual bug in #5384. Such old files now read as a plain grayscale volume instead.

We feel comfortable with that choice because (a) no software but OIIO wrote the other convention, (b) no other software could have properly read or displayed the incorrect files OIIO was writing, yet nobody ever pointed out the error or complained in any way until #5384. I am interpreting that as very strong circumstantial evidence that there are none of these wrong image OIIO produced that people still need to read with the wrong conventions.

Also fix stale, meaningless channel names, and add real names where the header supports it (STOKES axis codes, or informal FILTERn/BANDn keywords). If there are multiple channels and none of these conventions are apparent, default to just calling it R, G, B. Document the convention and channel-naming rules in builtinplugins.rst.

For the rarer NAXIS=4 (volume+color) case, apply the same pattern: the channel axis is NAXIS4, the slowest-varying one, not NAXIS1. There's no known real-world convention to confirm this against like there was for NAXIS=3, but it's the natural extension of the same reasoning.

Fixes #5384


src/doc/builtinplugins.rst
```diff
@@ -522,10 +522,33 @@ found at: http://fits.gsfc.nasa.gov/
...
+
+This is only a heuristic -- FITS has no way to formally declare an axis as
+"color" -- so a genuinely non-color NAXIS3 that happens to be <= 4 long can be
+misclassified as color.
+
+Channel names default to the usual OpenImageIO conventions (``R,G,B``,
+``R,G,B,A``, ``Y``, etc.), improved when the header says more: if the
+channel axis is a FITS WCS ``STOKES`` axis, channels are named from the
+standard Stokes codes (``I``, ``Q``, ``U``, ``V``, ``RR``, ``LL``, ``RL``,
+``LR``, ``pp``, ``qq``, ``pq``, ``qp``); otherwise, if every plane has a
```
jinhgkim: STOKES use XX YY XY YX instead of pp qq pq qp according to
https://casacore.github.io/casacore/Stokes_8h_source.html

src/fits.imageio/fits_pvt.h
```diff
@@ -114,6 +114,21 @@ class FitsInput final : public ImageInput {
...
     std::string convert_date(const std::string& date);
+
+    // Give m_spec.channelnames sensible values: generic defaults (R,G,B,...),
+    // improved with real per-plane names when the header tells us what the
+    // channel axis actually is (a STOKES polarization axis, or informal
+    // FILTERn/BANDn keywords).
+    void assign_channel_names();
+
+    // True if multi-channel pixel data is stored as separate per-channel
+    // blocks -- one full width x height (x depth, for NAXIS=4) block per
```
jinhgkim: `width x height (x depth, for NAXIS=4)` reads a bit confusing (could be just me). At a first glance, i thought it was a typo for `z` depth. maybe we can rewrite it as the following: `width x height (or w x h x depth, for NAXIS=4)`. But again, it could just be me!

src/fits.imageio/fitsinput.cpp
```diff
@@ -100,16 +102,46 @@ FitsInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+        if (n != m_spec.scanline_bytes()) {
+            if (feof(m_fd))
+                errorfmt("Hit end of file unexpectedly (offset={}, scanline {})",
+                         ftell(m_fd), y);
+            else
+                errorfmt("read error");
+            return false;  // Read failed
+        }
+    } else {
+        // Channels are stored as separate contiguous width x height (x depth,
```
jinhgkim: same comment as above

src/doc/builtinplugins.rst
```diff
@@ -522,10 +522,33 @@ found at: http://fits.gsfc.nasa.gov/
...
+FITS files can store arbitrary N-dimensional data arrays (``NAXIS`` /
+``NAXISn``), but say nothing about which axes, if any, are "color."
+OpenImageIO infers a role for each axis from ``NAXIS`` (trailing axes of
+length 1 are dropped first):
+
+* ``NAXIS`` = 1: a single row (width = NAXIS1, height = 1).
+* ``NAXIS`` = 2: an ordinary 2D grayscale image (width = NAXIS1, height = NAXIS2).
+* ``NAXIS`` = 3: if NAXIS3 <= 4, a color image, stored as one full-resolution
+  plane per channel (width = NAXIS1, height = NAXIS2, NAXIS3 = nchannels).
+  Otherwise, if NAXIS3 is more than 3, the file is interpreted as a grayscale
```
jinhgkim: Do you mean `if NAXIS3 is more than **4**`?

src/fits.imageio/fitsinput.cpp
```diff
@@ -416,6 +456,89 @@ FitsInput::add_to_spec(const std::string& keyname, const std::string& value)
...
+    if (m_naxes == 3)
+        axis = 3;
+    else if (m_naxes == 4)
+        axis = 4;  // volume+color: channel is the slowest-varying axis
+    if (!axis)
+        return;
+
+    // The FITS WCS standard defines a STOKES axis (polarization) with fixed
+    // integer codes per plane: 1=I, 2=Q, 3=U, 4=V, -1=RR, -2=LL, -3=RL,
+    // -4=LR, -5=pp, -6=qq, -7=pq, -8=qp. If that's what this axis is, decode
```
jinhgkim: -5=XX, -6=YY, and so on... https://casacore.github.io/casacore/Stokes_8h_source.html

src/fits.imageio/fitsinput.cpp
```diff
@@ -416,6 +456,89 @@ FitsInput::add_to_spec(const std::string& keyname, const std::string& value)
...
+    // the per-plane values via the standard linear WCS mapping and use the
+    // Stokes names as channel names.
+    std::string ctype = m_spec.get_string_attribute(
+        Strutil::format("Ctype{}", axis));
+    if (Strutil::iequals(ctype, "STOKES")) {
+        // Standard FITS WCS linear axis mapping and Stokes codes, see:
+        // https://fits.gsfc.nasa.gov/standard40/fits_standard40aa.pdf
```
jinhgkim, line 502: Oh, this doc you linked also says -5=XX etc

src/fits.imageio/fitsinput.cpp
```diff
@@ -100,16 +102,46 @@ FitsInput::read_native_scanline(int subimage, int miplevel, int y, int /*z*/,
...
+        }
+    } else {
+        // Channels are stored as separate contiguous width x height (x depth,
+        // for NAXIS=4) blocks rather than interleaved, so we must gather one
+        // row from each block and interleave them into the scanline buffer.
+        size_t comp_size   = m_spec.format.size();
+        size_t row_bytes   = size_t(m_spec.width) * comp_size;
+        size_t plane_bytes = row_bytes * size_t(m_spec.height)
+                             * size_t(m_spec.depth);
+        long row_off = (m_spec.height - y) * long(row_bytes);
```
jinhgkim: Urm, I don't think this takes into account the case of NAXIS=4, where depth exists.
I think, it should be something more like:

```cpp
size_t zplane_bytes = row_bytes * size_t(m_spec.height);
size_t plane_bytes = zplane_bytes * size_t(m_spec.depth);
long row_off = m_spec.depth * long(zplane_bytes) + (m_spec.height - y) * long(row_bytes);
```

If you agree, we need more changes down below accordingly...

jinhgkim (changes requested):

src/fits.imageio/fitsinput.cpp
lgritz: Yes, quite right. Fixing.

src/fits.imageio/fits_pvt.h
lgritz: Ah, I see the point of confusion, yes. Of course I was thinking x = "times". Will clarify.

src/fits.imageio/fitsinput.cpp
lgritz: Oh, right. That's always been broken in this way (the `z` coordinate is even commented out so it can't be used). 

Which is the right approach, do you think?
1. Fix so "volume" files work.
2. Stop supporting volumetric FITS files at all.

Circumstantial argument for choice 2: What you point out here shows that we were *never* able to read volumetric FITS files, so it can't break anybody to drop support, and arguably, nobody should be using OIIO to read FITS files for anything but 2D images. (We do support volumetric images for a few formats such as OpenVDB where they might realistically be used as 3D textures, but FITS files seem unlikely to be used in this way and are ill-suited for it, mostly because they are not tiled.)

To be honest, choice 1 is less work for me right now, but choice 2 might be better in the long run by not promising support for something that we don't think anybody will use this library for.

jinhgkim: Do it once, do it right -- choice 2! (only if you have capacity)

jinhgkim: Well I don't know if it's worth it to go with choice 2 cuz it seems like FITS is rarely used for the OIIO users.

lgritz: Wait, which way are you saying?

1 is "fix the couple of lines needed to read FITS volumes correctly"

2 is "nobody will ever need OIIO + volumetric FITS, so drop the support for 3D FITS files and get rid of the related code"

jinhgkim: I initially chose choice 2 and flipped my mind to choice 1, mid-thinking. If you have time, choice 2 sounds better.

jinhgkim (approved):
Looks good to me!

PR #5389, opened 2026-08-10
deps: Raise LibRaw minimum to 0.21.0

lgritz:
This LibRaw 0.21.0 release dates from Dec 2022, so will be nearly 4 years old by the time we release 3.2. The old/current enforced minimum of 0.20.0 dates from mid-2020, so is pretty old.

Raising the minimum to 0.21.0 allows us to get rid of much `#if` clutter that catered to older versions.

Change in main (OIIO 3.2) only, since we don't raise dependency minimums within minor releases.


antond-weta (approved):
Nice!

PR #5393, opened 2026-08-10
fix(IBA): keep nearest resample off the Highway path

wingfiring:
### Description

resample_hwy() ignores its `interpolate` argument and always bilerps, but resample_() routed nearest sampling to it whenever Highway was enabled, so ImageBufAlgo::resample(dst, src, false) returned filtered pixels instead of the source pixel the caller asked for. Measured against a 128x128 nearest resample of testsuite/common/grid.tif, 40.9% of pixels were wrong, with a maximum error of 0.608.

Latent today because enable_hwy defaults to 0, but it becomes a live behavior change as soon as Highway is turned on by default.

Also fixes an MSVC build failure in the same subsystem. The two hwy_*_perpixel_op_rgba_rgb_roi() lambdas derived a tag type with decltype(d), where d is captured from the enclosing function; MSVC yields a reference type there, and D::template Rebind<T> cannot be named through a reference. Naming the tag directly drops the dependency on how decltype treats a captured entity, and says what the tag is rather than what it matches.

### Tests

Case added: test_resample_hwy_nearest

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/libOpenImageIO/imagebufalgo_hwy_pvt.h
+++ b/src/libOpenImageIO/imagebufalgo_hwy_pvt.h
@@ -1100,7 +1100,7 @@ hwy_binary_perpixel_op_rgba_rgb_roi(ImageBuf& R, const ImageBuf& A,
                 (void)ba;
 
                 using DstLaneT = HwyLaneTypeT<Rtype>;
-                auto d_dstlane = hn::Rebind<DstLaneT, decltype(d)>();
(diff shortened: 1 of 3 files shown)


lgritz (approved):
LGTM
(and I'm going to assume that @ssh4net's 👍 is his approval, since I'm definitely not a hwy expert)

PR #5394, opened 2026-08-10
deps: Raise cmake minimum to 3.23.0

lgritz:
This is still 5 years old, but versus the prior 3.18 minimum, it gets us presets, better support for Apple Silicon, CUDA, C++23, among other things.

There's not much simplification this lets us do right away (like, no catering to old cmake that we can get rid of), but it opens the door to future PRs where we can confidently use cmake features that were introduced between 3.19 and 3.22.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -481,7 +481,7 @@ jobs:
             python_ver: "3.9"
             python_action_ver: "3.9"
             gcc_action_ver: 9
-            setenvs: export  CMAKE_VERSION=3.22.0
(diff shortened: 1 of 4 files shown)


jessey-git (approved):
I think 3.22 is fine. I quickly looked at some other dependencies like Alembic and OpenUSD:

USD 25.05 or greater (May 2025) already requires 3.26.
USD 26.08 or greater (July 2026) requires 3.27.
Alembic 1.8.11 or greater (March 2026) then you need at least 3.29.

So chances are the ecosystem has access to the more recent 3.x versions already.

PR #5398, opened 2026-08-13
fix(ImageInput): guarantee close()-on-failure for all open() methods

lgritz:
If ImageInput::open() encounters an error and returns false, is the object expected to be left in the same state as a newly constructed, never-opened instance (other than having an error that can be retrieved), as if close() had been called? Or is it up to the caller to do the close()?  We were never quite clear on this, and some plugins have any open() failures do a close() before returning, but others did not.

Clean this up once and for all by documenting ImageInput::open() to make clear that the contract is that a failed open() will close the file and needs no further cleanup action on the part of the caller.

Audit all readers and ensure that in every instance that an open() method returns false to indicate an error, close() is also called.

Note that:
- The error message is stil retrievable from the closed ImageInput.
- If an IOProxy was passed to the ImageInput to use, that proxy is still open and able to be used for further reads to the same file.

Co-authored by:	Brecht Van Lommel <brecht@blender.org>

### DIFF
--- a/src/cineon.imageio/cineoninput.cpp
+++ b/src/cineon.imageio/cineoninput.cpp
@@ -88,6 +88,7 @@ CineonInput::open(const std::string& name, ImageSpec& newspec)
     m_cin.SetInStream(m_stream);
     if (!m_cin.ReadHeader()) {
         errorfmt("Could not read header");
+        close();
         return false;
     }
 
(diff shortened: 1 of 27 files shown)


brechtvl (commented):
The changes look good to me, but they appear to be incomplete.

Here's a patch made with Claude assistance to find more missing cases. It includes some drive-by fixes and simplifications that seemed relevant to include.
[missed_close_and_other_fixes.patch](https://github.com/user-attachments/files/31081240/missed_close_and_other_fixes.patch)

brechtvl (approved):

PR #5399, opened 2026-08-13
fix(exif): harden the shared Exif decoder against malformed input

lgritz:
Audit of the shared Exif decoder, used by JPEG, PNG, Raw, HEIF, PSD, and WebP. Turned up several issues that could be trouble for corrupt, malicious, or unusual Exif blocks:

* Recursive IFD directory-count overread: the recursive ExifIFD/GPSIFD and Interoperability-IFD branches in read_exif_tag only checked that the target offset was < buffer size before reading a 2-byte directory count there, so an offset of `buffer_size - 1` would run one byte past the end of the blob.

* Unknown-type size sentinel: tiff_data_size() reports an unrecognized TIFF data type as size_t(-1). Then dataspan(), dataptr() and tiff_dir_data() multiplied that by the entry's count and fed the result to their bounds checks. Now tiff_data_size knows the type (which also makes Exif 3.0 UTF-8 tags decode correctly rather than being dropped or crashing), and a new pvt::dirdata_size() rejects the sentinel so that unknown types don't wrap the arithmetic. Also switch to 64 bit math for computing those offsets.

* IFD depth: ifd_offsets_seen stops an IFD chain from looping, but nothing stopped it from being arbitrarily deep and overflowing the stack. Cap the depth at at 32 levels, more than enough for any real file.

* Change some OIIO_ALLOCA to OIIO_ALLOCATE_STACK_OR_HEAP, to guarantee not allocating things too big to fit on the stack.

* The directory-entry bounds test used >= where it wanted >, quietly discarding a legitimate final entry that ended exactly at the blob end, and the TIFFHeader was read through a cast that assumed alignment.

* Add jpeg-corrupt and png-damaged tests (and scripts that generate those odd test cases).


src/libOpenImageIO/exif.cpp
```diff
@@ -161,6 +161,8 @@ tiff_data_size(TIFFDataType tifftype)
     static size_t sizes[]    = { 0, 1, 1, 2, 4, 8, 1, 1, 2, 4, 8, 4, 8, 4 };
     const int num_data_sizes = sizeof(sizes) / sizeof(*sizes);
     int dir_index            = bitcast<int>(tifftype);
+    if (dir_index == EXIF_UTF8_TYPE)
+        return 1;  // Exif 3.0 UTF-8 string: one byte per element
```
brechtvl, line 165: I think we need to handle `EXIF_UTF8_TYPE` in `tiff_datatype_to_typedesc` below as well.

```
    if (bitcast<int>(tifftype) == EXIF_UTF8_TYPE) {
        return TypeString;
    }
```

Because `rawinput.cpp` uses these functions together. Not a new issue though, without that the metadata will just continue to be skipped as before. Just more obvious now.

brechtvl (approved):
Difficult to verify if the entire Exif decoding is safe now, but at least the changes look correct to me.

src/libOpenImageIO/exif.cpp
lgritz: Good catch; updating PR.

brechtvl (approved):

PR #5404, opened 2026-08-15
fix(simd): NEON vfloat4::load(values,n) read past the end of the buffer

lgritz:
The NEON branch loaded all 4 lanes with vld1q_f32 and then zeroed the ones it didn't want, so a partial load of n<4 elements read up to 12 bytes past the caller's buffer. ASan flags it as a heap-buffer-overflow of size 16. The SSE branch gets this right by switching on n before loading.

Within OIIO the over-read has been harmless, because imagebuf.cpp and imagecache.cpp deliberately pad their allocations by OIIO_SIMD_MAX_SIZE_BYTES to permit exactly this. But simd.h is a public header, and a downstream caller has no reason to expect a 4-element load to touch 16 bytes.

Also, n==0 fell through to `default:` after the load without zeroing anything, returning the over-read garbage instead of zeros.

### DIFF
--- a/src/include/OpenImageIO/simd.h
+++ b/src/include/OpenImageIO/simd.h
@@ -6701,18 +6701,25 @@ OIIO_FORCEINLINE void vfloat4::load (const float *values, int n) {
         break;
     }
 #elif OIIO_SIMD_NEON
-    //switch (n) {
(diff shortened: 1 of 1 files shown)


jinhgkim (approved):
Looks good to me

PR #5416, opened 2026-08-20
fix(openvdb): validate grid extents, guard int overflow, error checking

lgritz:
Auditing of the openvdb reader looking for potential bugs and security issues turned up a few miscellaneous things to fix.

- Add a check_open() call to validate the image extents that come from the grid's file_bbox_min/file_bbox_max metadata.

  Note: There is deliberately no check_compression_ratio() added; since VDB is a sparse format, there is no relationship between a volume's maximum spatial extent and its expected minimum file size.

- Switch to 64 bit arithmetic for the dense-window math.

- Several error paths were also hardened. read_native_tile() dereferenced `gridPtrCast<>` without a null check. It also ran the OpenVDB tree walk with no exception guard. open() had no catch-all, no null check on readGrid(), and did not reset state on failure. A file with no representable grid returned a bare false with no error message.

- OpenVDB builds exception messages from strings whose length comes from the file. So exception text is now clamped to 1 KB.

- open() also resets state unconditionally on entry now.  The old reset check never fired. The member it checked was never assigned. So a second open() call used to append to the layer list instead of starting fresh.

- New testsuite fixtures cover these cases and are wired into the fuzz corpus.


testsuite/openvdb-damaged/ref/out.txt
```diff
@@ -0,0 +1,21 @@
+oiiotool ERROR: read : "src/bad-bbox-huge.vdb": openvdb image resolution may not exceed 1048576x1048576x1048576, but the file appears to be 1000000009x1000000009x1000000009. Possible corrupt input?
+Full command line was:
+> oiiotool --info -v src/bad-bbox-huge.vdb
+oiiotool ERROR: read : "src/bad-bbox-bomb.vdb": Uncompressed image size 30521698304.0 MB exceeds the 32768 MB limit.
```
brechtvl: This test fails on my machine:

```
-oiiotool ERROR: read : "src/bad-bbox-bomb.vdb": Uncompressed image size 30521698304.0 MB exceeds the 31929 MB limit.
+oiiotool ERROR: read : "src/bad-bbox-bomb.vdb": Uncompressed image size 30521698304.0 MB exceeds the 32768 MB limit.
```

I think you need to set an explicit `-oiioattrib limits:imagesize_MB` like the jxl and targa tests do, since it's automatically determined by system memory otherwise.

It passes on CI because that has:
```
export OPENIMAGEIO_OPTIONS="limits:imagesize_MB=32768,${OPENIMAGEIO_OPTIONS}"
```

testsuite/openvdb-damaged/src/make_malformed_vdb.cpp
```diff
@@ -0,0 +1,142 @@
...
+    {
+        FloatGrid::Ptr g = FloatGrid::create(0.0f);
+        g->setName("empty");
+        write_grid(p("bad-bbox-empty.vdb"), g);
+    }
+
+    // A file that ends in the middle of the grid descriptor. Truncating any
+    // deeper trips an unbounded length-prefixed string read inside OpenVDB
+    // that costs ~12 GB of RSS, which is no good for CI or the fuzzer; see
+    // the openvdb entries in docs/dev/format-hardening-deferrals.md.
```
brechtvl: `format-hardening-deferrals.md` doesn't seem to exist in the repository. Not sure if it's a local file that is not meant to be committed or if it will come later.

testsuite/openvdb-damaged/ref/out.txt
lgritz: Ah, thanks. Yes, I've run into a couple of those myself as well. Because the error message says what the limit was, you get a different error message text on the low-memory GitHub runners than you might on other machines. The solution, as you point out, is to give an explicit low limit that makes everybody fail with the same message. I've done this in some other areas, but missed this one.

testsuite/openvdb-damaged/src/make_malformed_vdb.cpp
lgritz: Oh, sorry, I will clean up that comment. I have a bunch of "spec" files that governed how this whole long (nearly 2 month) process has crawled its way through the different readers. My intent was to merge the spec files and checklists at the very end, after the last of the code patches has been merged, to preserve the rationale and procedure for this whole set of patches, and to avoid churn of those files that were being incrementally updated with each patch. I certainly did not mean to leave comments in the code that referred to those files.

(In case you're wondering, I've been having great luck with [spec-kit](https://github.com/github/spec-kit) when working on these larger tasks and audits with the help of the the coding agents. It makes it much more thorough and predictable, and helpfully leaves behind readable spec documents that can become part of the developer documentation to understand the planning and implementation behind large tasks.)

brechtvl (approved):

PR #5419, opened 2026-08-21
fix(jpeg2000): enable ICC profile write

### LINKED ISSUE #4608: ICC profile in JPEG2000
I also ran into the inability to save the ICC profile in JPEG 2000 while creating the [KDE plugin](https://invent.kde.org/frameworks/kimageformats/-/merge_requests/324).

I wrote a patch that allows me to [save it](https://github.com/uclouvain/openjpeg/pull/1574)...

luna-y-kim:
ICC profile writing was disabled in #1452 due to assertion failures. That was fixed on the OpenJPEG side in v2.5.4 (uclouvain/openjpeg#1574), which is also our current build default.

The ICC profile write code is compiled only when building against OpenJPEG >= 2.5.4. The double-free problem in `Jpeg2000Output::close()` is addressed by clearing the `icc_profile_buf` pointer (which points at memory owned by `m_icc_profile`) before `opj_image_destroy()` frees it.

Also add a testsuite case. Now the testsuite has two reference outputs: `out.txt` (no ICC, older OpenJPEG) and `out-icc.txt` (with ICC). Since `runtest.py::checkref()` accepts either as alternatives, the text comparison alone cannot catch a breakage. Thus, the extracted .icc file is additionally compared only on OpenJPEG >= 2.5.4, as the real guard.

Fixes #4608

### Checklist:

- **(N/A)** **I have updated the documentation** if my PR adds features or changes
  behavior.
  on the GitHub runners.)
- **(N/A)** If I added or modified a public C++ API call, I have also amended the


src/jpeg2000.imageio/jpeg2000output.cpp
```diff
@@ -66,7 +66,7 @@ class Jpeg2000Output final : public ImageOutput {
     bool m_convert_alpha;  //< Do we deassociate alpha?
     std::vector<unsigned char> m_tilebuffer;
     std::vector<unsigned char> m_scratch;
-#if 0
+#if OIIO_OPJ_VERSION >= 20504
```
lgritz: Since it's harmless to have m_icc_profile present even if not used, and it will only be used for it's a sufficiently new version (guarded at line 487), maybe it will be simpler to just remove the guard entirely here?

src/jpeg2000.imageio/jpeg2000output.cpp
```diff
@@ -84,7 +84,7 @@ class Jpeg2000Output final : public ImageOutput {
         m_codec         = NULL;
         m_stream        = NULL;
         m_convert_alpha = true;
-#if 0
+#if OIIO_OPJ_VERSION >= 20504
```
lgritz: Here, too, I think the simplest approach is to remove the `#if` and `#else`, just leaving the clear() as it is. It's harmless now, and nobody needs to go back and change it again some day when we're always using 2.5.4 or newer.

testsuite/jpeg2000/run.py
```diff
@@ -16,3 +16,28 @@
...
+    for elem in liblist.split(";"):
+        if elem.startswith("jpeg2000"):
+            try:
+                major, minor, build = (int(x) for x in elem.split()[-1].split("."))
+                return (major * 10000) + (minor * 100) + build
+            except Exception:
+                pass
+    return -1
+
+# Test for writing jpeg2000 with icc profile
```
lgritz: Maybe just amend this comment to explain that this test case is only added for openjpeg >= 2.5.4, because that's when it becomes reliable to implement this feature.

lgritz (commented):
EXTREMELY minor suggestions, but otherwise LGTM

src/jpeg2000.imageio/jpeg2000output.cpp
luna-y-kim: Yep, I agree on fewer #ifs, less work for later, since it's harmless. I saw the USE_OPENJPH guards around a few members, so I thought it might be better practice at the time. Updated!

testsuite/jpeg2000/run.py
luna-y-kim: Right, updated!

lgritz (approved):
LGTM

PR #5424, opened 2026-08-25
fix(raw): fixes cropping when the image is rotated

antond-weta:
### Description

This fixes incorrectly positioned crop when the image is rotated.

### Tests

I have not added any unit tests specifically for this fix. I'm planning to add some tests later on using generated DNG images, but I doubt that I will get them in time for the v3.2 release. Happy to delay this fix until the tests are ready, if we prefer.

Tested manually with images from several cameras.

### Checklist:

  behavior.
  on the GitHub runners.)

### DIFF
--- a/src/raw.imageio/rawinput.cpp
+++ b/src/raw.imageio/rawinput.cpp
@@ -815,8 +815,8 @@ RawInput::open_raw(bool unpack, bool process, const std::string& name,
         ushort crop_top    = 0;
         ushort crop_width  = 0;
         ushort crop_height = 0;
-        ushort left_margin = 0;
-        ushort top_margin  = 0;
+        ushort left_margin = m_processor->imgdata.sizes.left_margin;
+        ushort top_margin  = m_processor->imgdata.sizes.top_margin;
 
         auto p = config.find_attribute("raw:cropbox");
         if (p) {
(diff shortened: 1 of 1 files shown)


src/raw.imageio/rawinput.cpp
```diff
@@ -828,54 +828,75 @@ RawInput::open_raw(bool unpack, bool process, const std::string& name,
...
+        if (crop_width == 0 || crop_height == 0) {
+            const auto& raw_inset_crops
+                = m_processor->imgdata.sizes.raw_inset_crops[0];
+            if (raw_inset_crops.cwidth != 0) {
+                crop_left   = raw_inset_crops.cleft;
+                crop_top    = raw_inset_crops.ctop;
+                crop_width  = raw_inset_crops.cwidth;
+                crop_height = raw_inset_crops.cheight;
+            }
+        }
```
jinhgkim, line 852: Correct me if I'm wrong, but this seems to diverge from what the doc says [here](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/main/src/doc/builtinplugins.rst?plain=1#L2293-L2294): "The default cropping can be disabled by setting the cropbox to zero size."

We now get the camera crop even when the cropbox is zero sized.

src/raw.imageio/rawinput.cpp
```diff
@@ -828,54 +828,75 @@ RawInput::open_raw(bool unpack, bool process, const std::string& name,
...
+
         if (crop_width > 0 && crop_height > 0) {
-            ushort image_width  = m_processor->imgdata.sizes.width;
-            ushort image_height = m_processor->imgdata.sizes.height;
+            const auto S         = m_processor->imgdata.sizes;
```
jinhgkim, line 855: Perhaps the intent was `const auto&`?

jinhgkim (commented):
A couple things I thought were worth mentioning

src/raw.imageio/rawinput.cpp
antond-weta: good catch, fixed

src/raw.imageio/rawinput.cpp
antond-weta: fixed

testsuite/raw/run.py
```diff
@@ -67,4 +67,16 @@
     command += oiiotool ("-iconfig raw:Demosaic none --stats src/" + f,
                          failureok = True)
 
+# Check the crop size and position in all 4 orientations.
+for rotation in [0, 90, 180, 270]:
+    command += oiiotool (
+        "--iconfig raw:Demosaic none " +
+        "-i src/" + "crop-36x32_" + str(rotation) + ".dng " +
+        "--echo \"________________\" " +
+        "--echo \"ROTATION " + str(rotation) + " info:\" " +
```
lgritz, line 76: I like what you've done here

lgritz (approved):
LGTM, thanks for the cleanup

PR #5425, opened 2026-08-26
cleanup: Minor libraw related cleanup

lgritz:
* Rename _errorfmt to thumb_errorfmt for clarity -- this function is only used for thumbnail related errors, and should not be easily confused with the regular errorfmt.

* Get rid of no-longer-needed libraw test outputs that were output reference for old ibraw versions we no longer support.


antond-weta (approved):
nice!

PR #5428, opened 2026-08-27
fix(int): OpenColorIO exception safety improvements

lgritz:
We had recently done a around to be sure that all OCIO calls were wrapped in try/catch (since OCIO uses exceptions, and we do not).  But mostly we were catching `OCIO::Exception` and that was ok for an exception purposely thrown by OCIO, but what about an exception thrown by a system library call that OCIO called and failed to catch? It wouldn't be an OCIO::Exception, it would be a std::exception.

So catch those instead. OCIO::Exception derive from std::runtime_error, which derives from std::exception, so nothing will be missed because of this... but more things might get caught.


chellmuth (approved):
LGTM!

PR #5429, opened 2026-08-27
Annotate spin_mutex and spin_rw_mutex for ThreadSanitizer

NathanGray-ChurchillNavigation:
### Problem

Building with `-fsanitize=thread` reports a data race inside `spin_mutex::lock`:

```
WARNING: ThreadSanitizer: data race
  Read of size 1 at 0x... by thread T2:
    #0 OpenImageIO::v3_1::spin_mutex::lock() OpenImageIO/thread.h:224:22
    #1 OpenImageIO::v3_1::spin_mutex::lock_guard::lock_guard(...) OpenImageIO/thread.h:253:18
```

Line 224 is the `OIIO_THREAD_ALLOW_DCLP` spin, `while (*(volatile bool*)&m_locked)` —
a non-atomic read of a word other threads write atomically. tsan cannot tell the
surrounding construct is a lock, because it is built on `std::atomic_flag` rather
than a mutex tsan recognises. `OIIO_THREAD_ALLOW_DCLP=0` avoids the report at the
cost of the fast path; the only other option downstream is suppressing every
`OpenImageIO::` frame, which also hides real bugs in the calling code.

### Change

`__tsan_mutex_*` annotations (`sanitizer/tsan_interface.h`) on the acquire and
release paths of `spin_mutex` and `spin_rw_mutex`. `try_lock` uses
`__tsan_mutex_try_lock` / `__tsan_mutex_try_lock_failed` so a failed attempt is not
recorded as an acquisition; `lock()` needs nothing of its own, since it spins on
`try_lock`.

Annotating an address makes tsan treat it as a mutex rather than data, so line 224
is no longer reported and the fast path is left exactly as written. This changes no
behaviour, only what tsan is told.

All of it sits behind `OIIO_TSAN_ANNOTATE_LOCKS`, detected via `__SANITIZE_THREAD__`
or `__has_feature(thread_sanitizer)`, expanding to `((void)0)` otherwise. No new
include and no effect on non-tsan builds.

### Testing

Two threads contending a `spin_lock` under `clang++ -std=c++20 -fsanitize=thread`:
reported before the change, clean after. The header also compiles with
`-fsyntax-only` both with and without the sanitizer. Happy to add the reproducer as
a test if you want one in-tree.

### DIFF
--- a/src/include/OpenImageIO/thread.h
+++ b/src/include/OpenImageIO/thread.h
@@ -41,6 +41,45 @@
 #endif
 
 
+// ThreadSanitizer cannot tell that the spin locks below are locks: they are
+// built out of atomics, so it never sees the happens-before edges they
+// establish and reports races on the data they protect. The __tsan_mutex_*
+// hooks tell it. Compiled away unless the TU is built with tsan.
+#if defined(__SANITIZE_THREAD__)
+#    define OIIO_TSAN_ANNOTATE_LOCKS 1
+#elif defined(__has_feature)
+#    if __has_feature(thread_sanitizer)
+#        define OIIO_TSAN_ANNOTATE_LOCKS 1
+#    endif
+#endif
+#ifndef OIIO_TSAN_ANNOTATE_LOCKS
+#    define OIIO_TSAN_ANNOTATE_LOCKS 0
+#endif
+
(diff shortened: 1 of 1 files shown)


lgritz (approved):
LGTM, thanks for this!

PR #5430, opened 2026-08-28
python: Switch default python bindings to nanobind

lgritz:
Change the default OIIO_PYTHON_BINDINGS_BACKEND to nanobind for the 3.2 release.

Shuffle the distribution in CI tests around -- building against old years tests pybind11, building against new tests mostly nanobind, sometimes both. Make the "bleeding edge" test against nanobind main, make "latest releases" test against nanobind 3.0.0.

The autobuild of nanobind still sticks with 2.13.0 by default, since the newer 3.0 requires a minimum python that's higher than our own minimum python.

### DIFF
--- a/src/cmake/build_nanobind.cmake
+++ b/src/cmake/build_nanobind.cmake
@@ -17,7 +17,7 @@
 
 # NOTE: We are sticking with autobuild of nanobind defaulting to 2.13.0,
 # because nanobind 3.0+ requiers a minimum python of 3.10, whereas we
-# still support python 3.0.
+# still support python 3.9.
 set_cache (nanobind_BUILD_VERSION 2.13.0 "nanobind version for local builds")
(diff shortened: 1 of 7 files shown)


src/cmake/build_nanobind.cmake
```diff
@@ -15,6 +15,9 @@
 # the (unusually light) local build performed below.
 ######################################################################
 
+# NOTE: We are sticking with autobuild of nanobind defaulting to 2.13.0,
+# because nanobind 3.0+ requiers a minimum python of 3.10, whereas we
+# still support python 3.0.
```
zachlewis:
```suggestion
# still support python 3.9.
```

zachlewis (approved):
Exciting...! 

Our current python stub generation tooling will probably (well, definitely) need some love at some point, but nanobind really is quite nice!

PR #5438, opened 2026-08-30
deps: Bump DCMTK minimum to 3.6.2

lgritz:
This is still many many years old, but it's the only jump (from 3.6.1) that gives us any code simplification on our side, to there's no benefit to us of insisting on an even newer minimum version.


jessey-git (approved):
Yeah, this seems perfectly fine.

PR #5448, opened 2026-09-05
fix(heif): raw reads of irot-rotated files: transposed dims/OOB read + inverted irot direction

TomRoyls:
Two bugs in the HEIF/HEIC reader's raw-read path, found while debugging a deterministic SIGSEGV in COLMAP's point color extraction (COLMAP passes `oiio:reorient=0` in `Bitmap::Read`). Both trigger only for files that carry `irot`/`imir` transformation properties (e.g. Apple iPhone photos) read with the `oiio:reorient=0` configuration hint.

## Bug 1: transposed dimensions → out-of-bounds read (crash)

`HeifInput::seek_subimage` decodes with `heif_decoding_options::ignore_transformations = !m_reorient`. With `oiio:reorient=0`, libheif therefore returns the image in **stored, unrotated** orientation, and the spec assigned from the decoded image already describes that plane geometry. The orientation block then swapped `m_spec.width/height` for orientations ≥ 5, on the assumption that "libheif supplies oriented width & height". That premise holds for `heif_image_handle_get_width()` (the *handle*) but not for `heif_image_get_width()` on an image decoded with `ignore_transformations=1`, which reports the **stored** dimensions.

After the swap the spec disagrees with the plane (for an 8064×6048 stored portrait: spec claims 6048 wide while the plane is 8064 wide with only 6048 rows). `read_scanlines` then asks for more scanlines than the plane has, and `read_native_scanline`'s `memcpy` reads up to ~48 MB past the end of the plane. Whether it segfaults depends on heap layout; in COLMAP's incremental mapper it does, deterministically at the first rotated image whose colors get extracted. Non-crashing callers silently get heap garbage in the tail rows — COLMAP feature extraction reads the same way, so features of rotated images were computed over out-of-bounds memory.

Fix: don't swap the spec dimensions — they already match the plane. Instead, report the stored orientation via the `Orientation` attribute so raw-read callers can account for the rotation themselves. The default reoriented (`oiio:reorient=1`) path is unchanged.

Minimal reproducer (crashes on main; clean after the fix):

```cpp
OIIO::ImageSpec config;
config["oiio:reorient"] = 0;
auto in = OIIO::ImageInput::open("IMG_1307.HEIC", &config);
// spec reports 6048x8064; actual plane is 8064x6048
std::vector<unsigned char> buf(in->spec().width * in->spec().height * 3);
in->read_image(0, 0, 0, 3, OIIO::TypeDesc::UINT8, buf.data());  // SIGSEGV
```

## Bug 2: irot orientation computed in the wrong direction

The `irot` property value is the **anticlockwise** presentation rotation (ISO/IEC 23008-12), but the composition loop advanced the orientation by `rot/90` **clockwise** quarter turns — the opposite direction. For `irot=270` the computed orientation was 8 instead of 6, and for `irot=90` it was 6 instead of 8.

Fix: advance by `(4 - rot/90) % 4` clockwise quarter turns.

This was mostly invisible until now because the default (auto-reorient) decode returns oriented pixels and resets `Orientation`, so only the informational `oiio:OriginalOrientation` attribute carried the wrong value; the raw-read `Orientation` attribute introduced by the previous commit makes it matter.

### Why the reference outputs change

Pixel truth, using `oiio-images/heif/greyhounds-looking-for-a-table.heic` (irot=90, no imir):

```
$ oiiotool --iconfig oiio:reorient 0 greyhounds.heic -o raw.tif
$ oiiotool greyhounds.heic -o oriented.tif
$ oiiotool raw.tif --rotate270 -o ccw.tif && oiiotool ccw.tif oriented.tif --diff   # PASS
$ oiiotool raw.tif --rotate90 -o cw.tif   && oiiotool cw.tif  oriented.tif --diff  # FAILURE
```

Rotating the raw image 90° **CCW** reproduces the auto-reoriented pixels exactly, i.e. orientation **8** is correct for irot=90. This matches libheif's own semantics, stated in its source and enforced by its unit tests: *"Box_irot stores rotation in counter-clockwise degrees; EXIF labels rotations clockwise"* (`libheif/tests/mini_box.cc`), with `irot(90) == heif_orientation_rotate_270_cw` (≡ EXIF 8).

The value 6 recorded in most `testsuite/heif/ref/out-libheif*.txt` variants was the output of the inverted computation, never pixel-verified; notably `ref/out-libheif1.12-orient.txt` — added in #4262 alongside the comment that the different result on old libheif was "not worth tracking down" — already recorded 8 (that build takes the pre-1.16 path that reads the file's EXIF tag, which agrees with the pixel truth and libheif's semantics). The second commit corrects that one line in the affected variants.

## Test coverage

New `testsuite/heif` section with a committed fixture `ref/rotated-90cw.heic`: the asymmetric orientation pattern from `testsuite/oiiotool-xform/src/orientation1.tif` (per review — a symmetric checker cannot distinguish CW from CCW mixups), converted with libheif's `heif-enc --rotate-cw 90` since the OIIO writer doesn't emit `irot`. The stored image is 160×240 with irot=270.

- `--info -v --iconfig oiio:reorient 0` must report the stored dims (160×240) plus `Orientation: 6` — on main it reports transposed 240×160 and no orientation;
- consistency check: `--rotate90` of the raw read vs. the auto-reoriented decode → `--diff` must PASS (FAILURE on main). This is, as far as I can tell, the first pixel-level verification of the heif reader's orientation handling in the suite.

The new assertions go to a dedicated `out-reorient.txt` (per the review discussion: a single libheif-version-independent ref instead of another family of version variants). Volatile metadata (`DateTime|Software|ImageHistory|CICP`) is filtered from the info dumps; the remaining lines (dims, channel list, Orientation, ColorSpace, diff result) are stable across the libheif versions used by CI (1.19.x and 1.23.x).

## Verification

- The reproducer above crashes on main (matches the reported COLMAP crash: PC in `memcpy`, caller `read_native_scanline`, `rdx=18144`) and decodes cleanly with the fix; spec 8064×6048 matches the plane.
- Non-rotated HEIC control: unchanged behavior.
- Full COLMAP 4.2.0 incremental reconstruction over a 151-image iPhone HEIC dataset that previously crashed deterministically right after "Registering image #21": completes (exit 0) with the patched library.
- heifinput.cpp is unchanged between 3.1.16 and 3.1.17 and the swap logic predates the 10/12-bit rewrite, so all 3.1.x releases built against libheif ≥ 1.16 are affected.

Assisted-by: Sisyphus (OhMyOpenCode) / GLM (zai-coding-plan/glm-5.3)


testsuite/heif/run.py
```diff
@@ -32,5 +32,20 @@
 command += oiiotool("--pattern fill:color=0.5,0.5,0.5 47x31 3 -o odd-size.avif")
 command += info_command("odd-size.avif", safematch=True)
 
+# Test reading an irot-transformed image with oiio:reorient=0, which asks
+# for the raw stored pixels without applying the transformation. The spec
+# must describe the stored (unrotated) plane, the orientation must be
+# reported via the "Orientation" attribute, and rotating the raw pixels
+# back must reproduce the default auto-reoriented read exactly.
+rotated = make_relpath(os.path.join(imagedir, "rotated-90cw.heic"))
+reorient_redirect = " >> out-reorient.txt "
```
lgritz: We don't usually split the test output into multiple files. It's ok to just keep it all as one file and update the references.

But also ok to keep it like this. Libheif does seem to have a lot of reference output because it changes so much from version to version, so maybe in this case if these tests do not have a lot of version-to-version variation, it makes sense to split it out and have a single ref output.

TomRoyls: Kept as a separate `out-reorient.txt` for exactly that reason: the assertions it contains (stored dims, `Orientation` value, diff PASS/FAIL) don't vary with the libheif version, so a single ref replaces what would otherwise become yet another family of version variants.

lgritz: Can I ask for one tiny change, then? Since this is so unusual (maybe the only instance yet of having the console output split into two different saved outputs to compare), let's make an extremely clear delineation of the set of commands going to a different destination, and not have to repeat the use of reorient_redirect that can lead to cut-and-paste errors in the future.

Perhaps the following structure would be a way to make it maximally clear?

```
###############
# redirect this section of commands to out-reorient.txt
redirect_save = redirect
redirect = " >> out-reorient.txt "
outputs += [ "out-reorient.txt" ]

command += oiiotool("...")
command += oiiotool("...")

redirect = redirect_save
###############
```

And, I'm not necessarily asking you to do the following in this PR, but if we *really* wanted to make it clear and easy and set this up as an idiom to use elsewhere, maybe we'd prefer

```
###############
# redirect this section of commands to out-reorient.txt
redirect_push("out-reorient.txt")
command += oiiotool("...")
command += oiiotool("...")
redirect_pop()
###############
```

with the implementations of redirect_push (save the old redirect, set the new one, and add the new destination to the list of outputs) and redirect_pop (restore the previous redirect) defined in the main runtest.py.

TomRoyls: Done in d8bd090 — the section now saves/restores the `redirect` variable exactly as you sketched, with the `###` markers delimiting it, and the individual commands go through the standard `oiiotool()` helper so no hand-concatenated redirects remain. Output is byte-identical to the committed ref, so no ref churn. (The `redirect_push`/`redirect_pop` framework version sounds like a nice follow-up for the main runtest.py, but left out of this PR as suggested.)

lgritz (approved):
LGTM!

testsuite/heif/run.py
lgritz: I'm going to merge this, you've done a good job with it, but just for posterity I will mutter to myself that in a sense, we've done this backwards -- instead of having each new test facet save to yet another output file if it doesn't differ for each version of libheif, the right way to do it is to refactor the test to move JUST THE SMALL VARYING PART to a secondary output that might need to be replicated for each version, and have the vast majority that is stable stain in the main out.txt. Oh well, maybe something to come back to later.

PR #5449, opened 2026-09-05
docs: Convert Sphinx documentation from RST to MyST Markdown

lgritz:
Background: In the early days of OIIO, the documentation was LaTeX, with the primary output being a beautiful book-typeset and -length PDF that was distributed in the repo.

In mid-2019, we switched to Sphinx + Breathe + Doxygen, which allowed for nice online docs that could be hosted on readthedocs.org, made it so that anybody could build the docs, including in CI (the LaTeX system was finicky and I had to build it myself), and of course it was making the repo too big by having a new multi-MB PDF do check in every time we wanted a new version of the docs.

The Sphinx system used rst (ReStructured Text), which is somewhat like Markdown, but way more complicated and ugly. I was never really loving it, but it's what we had to use.

At some point later, Sphinx gained the ability to use Markdown via the "Myst" plugin. But the OIIO docs were SO HUGE that I just couldn't justify the enormous amount of labor to convert it all yet again, plus it's such tedious work that I would probably introduce a lot of new errors along the way. So we never bothered.

Until today -- this is the kind of problem that the robots are great at!

So now the whole thing is Markdown, which I find easier to read and edit, much less ugly in source form, and of course goes along nicely with the many standaline Markdown files that we use in the project.

Pedantic note: It's not plain Markdown, it's Myst-flavored markdown, which gives us a few nonstanard extra controls that are needed to be able to do all the things we did in rst before that are beyond the ability of regular Markdown. Also, one thing it can do is have sections of rst embedded in the markdown, and that's necessary to get some of the things matching how they were before. But they are isolated to those particular things.

Verified that the html produced by this mostly-myst-markdown with a tiny bit of residual rst ALMOST EXACTLY matches the html that was produced by the old rst-only system (literally byte for byte), so we're as certain as we can be that this produces the same results and appearance as before.


jfpanisset (commented):
rst vs md is a bit of a vi vs emacs scenario, but:

- Arguably more people are familiar with Markdown than with RST, potentially lowering the barrier for others to update the documentation
- The "almost byte for byte output" would seem to be a good test that this new infrastructure works

One comment: in some cases, foo.rst was renamed to foo.md and updated, whereas in other cases bar.rst was deleted and bar.md created from scratch, which would potentially lose the git history (which seems still relevant since the .md files aren't a complete rewrite / reformat from the .rst files). Would it make sense to try to retain the history for all renamed files?

src/doc/imagebufalgo.md
```diff
@@ -0,0 +1,3701 @@
...
+> >          # Add color Gaussian noise to an existing image
+> >          oiiotool tahoe.jpg --noise:type=gaussian:stddev=0.1 -o C.exr
+> >
+> >          # Use salt and pepper noise to make occasional random dropouts
+> >          oiiotool tahoe.jpg --noise:type=salt:value=0:portion=0.01:mono=1 -o D.exr
+> > ```
+>
+> %
+
+::::{list-table}
```
luna-y-kim, line 411: Image

Left side is the old one, and right side is this PR.
The example images lose their block quotes, and I found the same happens for the other example images in `imagebufalgo.md`. I'm wondering if this is intended?

lgritz: I wouldn't exactly call it "intended"... but I think it looks better now? I aspire to remove the awkward looking multi-quoting on the rest of it. Ultimately, I just want it indented enough to make it clear that it's part of the noise section. The weird shading that comes along with the block quotes were never really intended and I think it looks terrible, I just haven't had the time to figure out how to pretty it up.

I was hoping to do the rst -> md conversion preserving all of the content and a reasonably close facsimile of the look now, and then return to fix all the things we didn't like before, and the top of my list is this weird multiple nesting appearance.

luna-y-kim: Yeah I agree that the multi-quoting looks terrible! But the cosmetic stuff can wait for later.
As for preserving content, I'm going through the pages, but so far it looks great.

lgritz: Thanks, I really appreciate the double-check!

PR #5451, opened 2026-09-06
feat(simd): raise USE_SIMD default on x86_64 x86-64-v2

lgritz:
Follow-on to PR #5411.

Broaden the x86_64 default from "sse4.2" to "x86-64-v2" (adds POPCNT and CMPXCHG16B), the level every x86-64 CPU since ~2009 has. When the compiler understands -march=x86-64-vN, pass that too, so nothing is missed.

USE_SIMD now takes x86-64-v2/v3/v4 as synonyms for their feature-token sets, and promotes a bare avx/avx2/avx512* token to the level it guarantees, so "USE_SIMD=avx2" also gets the FMA/F16C/BMI that no AVX2 CPU lacks. avx512* promotes to v3 only, since AVX-512 parts vary in which subsets they carry.

Also newly recognized as tokens: cx16, bmi, bmi2, lzcnt, movbe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -556,7 +556,7 @@ jobs:
           - desc: oldest clang11/C++17 py3.9 exr3.1 ocio2.3
             # Oldest clang and versions of the dependencies that we support.
             # Deliberately sets no `simd`, so that this job exercises the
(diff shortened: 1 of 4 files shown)


lgritz:
No objections for 3 weeks and addressed all issues suggested by Copilot review -> merging

PR #5455, opened 2026-09-08
fix(jpeg): Catch corruptions when retrieving thumbnails

lgritz:
When scanning a JPEG file's Exif block for an embedded thumbnail, don't
trust the marker segment fields the file supplies:
    
* Bounds check before reading a segment's length or a SOF0/SOF2 header, so
  a marker near the end of the block can't read past it.
* Reject a segment length below 2 (it can't cover its own length field, and
  subtracting them underflowed the 16 bit value into a ~64 KB skip) or one
  reaching past the end of the block (previously the skip was silently
  dropped and the scan carried on at an arbitrary offset).
* Read the length and the SOF dimensions big endian. The thumbnail is a
  self-contained JPEG stream, so its marker fields are big endian whatever
  byte order the enclosing Exif block uses; keying the byte swap off the
  Exif order meant "II" files reported thumbnail_width/height/nchannels as 0.
    
scan_for_thumbnail() now returns false when it finds the stream corrupt, and
open() turns that into an error under imageinput:strict, as it already did
for Exif and IPTC. Same treatment for a failed XMP decode.
    

lgritz:
Updated with more fixes, including addressing copilot review comments

lgritz:
No objections -> merging

PR #5456, opened 2026-09-08
docs(IB): test perpixel_op documentation example

Yanghaha051218:
## Description

Convert the C++ `ImageBufAlgo::perpixel_op()` example from
`src/doc/imagebuf.md` into a compiling example in the
`docs-examples-cpp` testsuite.

The documentation now references the tested source with `literalinclude`
instead of maintaining a separate inline code example.

The example exercises all three variants shown in the documentation:

- standalone function
- functor
- lambda

This contributes to #3992.

## Tests

- Built `docs-examples-imagebuf` locally
- Ran the executable successfully
- Verified all three variants produce `5 7 9`
- Ran `clang-format`
- Ran `git diff --check`

Assisted-by: ChatGPT / GPT-5.6 Sol


lgritz (approved):
LGTM, thanks!

PR #5457, opened 2026-09-08
fix(psd): bounds-check thumbnail, color data, and layer channel lengths

lgritz:
Fuzzing the psd reader turned up several corrupt-input failures:

- load_resource_thumbnail() computed jpeg_length = length - 28 with no
  lower bound check, so a short resource underflowed into a multi-GB
  allocation.
- load_color_data() allocated from the declared length without checking it
  against the file size.
- load_layer_channel() sized compressed_data from the raw channel
  data_length. Bound it by the bytes remaining from the start of the
  channel, compared unsigned: data_length is 64 bits for PSB, where a
  corrupt value cast to int64_t wraps negative and slips past the guard
  into an uncaught bad_alloc.
- The RLE path subtracted the RLE length table from data_length without
  checking that it fit, underflowing on corrupt input.
- load_layers_16_32() asserted layer_info.length == 0, so a file carrying
  both a normal layer info section and an Lr16/Lr32 block aborted a debug
  build. Return an error instead.

Adds fixtures for each new rejection path.


lgritz:
Updated based on review comments

lgritz:
No objections -> merging

PR #5460, opened 2026-09-09
build: remove FindOpenEXR.cmake module

lgritz:
FindOpenEXR.cmake was a relic from the days when we supported OpenEXR releases from prior to their having reliable exported CMake configs. For a while now, it has forced use of the configs underneath, but then set a bunch of other variables that mimicked those set by the pre-config methodology.

This patch removes this useless artifact, and fixes up all the places that depended on those variables to instead use targets.

Along the way, this fixes a bug wherein OPENEXR_VERSION came out empty for anyone building OpenEXR locally (OpenImageIO_BUILD_LOCAL_DEPS) or setting ALWAYS_PREFER_CONFIG, which silently forced OIIO_OPENEXR_CORE_DEFAULT=0 and left OPENEXR_INCLUDES and IMATH_INCLUDES empty.

Several variables and some code guarded by them were found to be dead, not used since back when we has some non-module versions to contend with. That was all cleaned up as well.


src/cmake/externalpackages.cmake
```diff
@@ -55,12 +55,20 @@ checked_find_package (Imath REQUIRED
     PRINT IMATH_INCLUDES OPENEXR_INCLUDES Imath_VERSION
 )
 
-checked_find_package (OpenEXR REQUIRED
+checked_find_package (OpenEXR REQUIRED CONFIG
     VERSION_MIN 3.1
     NO_FP_RANGE_CHECK
     PRINT IMATH_INCLUDES OPENEXR_INCLUDES Imath_VERSION
```
jinhgkim: FYI, this won't print anything other than Imah_VERIOSN cuz IMATH_INCLUDES and OPENEXR_INCLUDES won't exist yet. Same applies to the Imath a couple line above.

lgritz: oh, you're right, these are stale, I will remove.

lgritz: Updated. In fact, I removed those variables entirely from the codebase, just using the target names in their place.

lgritz: I meant, I removed the _TARGETS variables, not the _INCLUDES variables

jinhgkim: yup, looks good, did you wanna remove line 55 too?

lgritz: Quite right. Fixed. Thanks for catching.

jinhgkim (approved):

PR #5461, opened 2026-09-09
fix(jpeg2000): give OpenJPH error handler static lifetime

lgritz:
OpenJPH's error handling is process-global (configure_error / get_error
share one pointer). open() installed a stack-local handler and only
restored the global on the exception path, so a successful open left the
global dangling at freed stack; a later open captured it and crashed when
OpenJPH next reported an error (stack-use-after-return, seen in fuzzing).

Install a single handler with static lifetime, once, from a thread-safe
static initializer, and never restore it. Doing the install on every
open() would instead race: set_error_stream and configure_error write
plain non-atomic globals that OpenJPH reads from any thread decoding
scanlines. Also drop the UB va_list-through-... forwarding; the handler
just throws.


lgritz:
Updated based on review comments

lgritz:
No objections -> merging

PR #5463, opened 2026-09-10
iv: tolerate partially-written EXR files

### LINKED ISSUE #4713: iv: tolerate/display partially-written exr files (originally #1191).
The idea is to make sure that iv can accept a partial file, meaning: one that opens, but some scanlines/tiles succeed when read and others fail, it should be tolerant of this, and optionally may want a display mode where it's clear which tiles are missing versus black.

The iv side is probably straightforward. The question is how many things you'll bump into upstream (in ImageBuf? ImageCache? ImageInput? OpenEXR's ImageInput implementation?) that might stand in your way and need some adjustment to tolerate the missing parts of the file. Those will need to get fixed.

linsen458-spec:
Fixes #4713

If iv opens an EXR that a renderer crashed on mid-write, the read fails and iv refuses to show anything, even though most of the pixels are actually fine.

As discussed above, everything needed on the OpenEXR side already exists (`oiio:missingcolor`), so this version keeps iv's part to a minimum:

- iv tries the straightforward read first; if it fails, it reopens the file through the ImageBuf config with `oiio:missingcolor` set to black, and the OpenEXR reader fills the unreadable scanlines/tiles instead of failing. In the partial case the status bar notes that the file is only partially readable. For ImageCache-backed reads, a cheap probe of the last scanline/tile still catches the partially-written case before display time.
- The one OpenEXR-side change this PR does carry: a failed multi-scanline chunk decode with `oiio:missingcolor` set used to recurse forever (per-scanline retry -> chunk cache -> retry -> ...), because the retry re-entered the chunk cache it had just failed to fill. The retry now bypasses the chunk cache. This fixes that pre-existing bug for `missingcolor` users too, and it's what makes the iv fallback work at all on compressed multi-scanline files.

Verified locally with a checker EXR truncated at 55%: the straightforward read fails with the expected error, the `missingcolor` read succeeds with the readable portion intact and the truncated tail black, and full files behave identically with and without the attribute.

Thanks for steering this into a much simpler shape — using the existing attribute is obviously the right call.


src/iv/ivutils.h
```diff
@@ -33,6 +40,197 @@ floor2f(float f)
...
+    const int xend   = spec.x + spec.width;
+    const int yend   = spec.y + spec.height;
+    const int zend   = spec.z + spec.depth;
+    bool any_success = false;
+    bool any_failure = false;
+    auto pixel_at    = [&](int x, int y, int z) -> void* {
+        return (char*)buf.localpixels() + (z - spec.z) * buf.z_stride()
+               + (y - spec.y) * buf.scanline_stride()
+               + (x - spec.x) * buf.pixel_stride();
+    };
```
lgritz: There is already `ImageBuf::pixeladdr()`

src/iv/imageviewer.h
```diff
@@ -104,6 +104,16 @@ class IvImage final : public ImageBuf {
     ///
     bool image_valid() const { return m_image_valid; }
 
+    /// True if the last read only partially succeeded, i.e. the image
+    /// specification was readable but some of the pixel data could not be
+    /// read (for example, a file that was only partially written before a
+    /// renderer crashed).
```
lgritz, line 110: I'm not asking you to change this, because you have merely matched the style of the surrounding code, so this is more about me just talking as a note to myself: These `///` doxygen-style comments should never have been in this file, since it's not a public header and we don't automatically extract any documentation from this class.

src/openexr.imageio/exr_pvt.h
```diff
@@ -265,6 +265,11 @@ class OpenEXRInput final : public ImageInput {
     bool read_native_scanlines(int subimage, int miplevel, int ybegin, int yend,
                                int z, int chbegin, int chend,
                                void* data) override;
+    // Internal version with a flag letting the per-scanline retry bypass the
+    // chunk cache (see read_native_scanlines_individually).
+    bool read_native_scanlines(int subimage, int miplevel, int ybegin, int yend,
+                               int z, int chbegin, int chend, void* data,
+                               bool bypass_chunk_cache);
```
lgritz, line 272: I find this a little confusing, with this custom function mixed in among the overrides of the parent class. Maybe you can move the declaration of this function into the private section toward the end of the class declaration? And maybe it should have a different name so it's not easily confused with the regular read_native_scanlines?

src/openexr.imageio/exrinput.cpp
```diff
@@ -1298,6 +1302,15 @@ bool
 OpenEXRInput::read_native_scanlines(int subimage, int miplevel, int ybegin,
                                     int yend, int z, int chbegin, int chend,
                                     void* data)
+{
+    return read_native_scanlines(subimage, miplevel, ybegin, yend, z, chbegin,
+                                 chend, data, /*bypass_chunk_cache=*/false);
+}
```
lgritz, line 1308: Let's add a comment to clarify. Something like:

This is the externally-called read_native_scanlines, which is a wrapper around the internal version that takes an additional parameter saying whether to use the chunk cache or not. When called externally, we always do.


This also raises the question of whether it should be bypass_chunk_cache (defaulting to false), or use_chunk_cache (defaulting to true). Of course, logically they are equivalently useful, but you should think about which one is less awkward to use when explaining how it works and when you'd use it.

src/iv/ivutils.h
```diff
@@ -33,6 +40,56 @@ floor2f(float f)
...
+/// failures are much less gracefully handled.
+///
+/// Returns true if the last scanline/tile could be read (and therefore it
+/// is likely that the entire pixel data block is intact), false if it
+/// could not (meaning that the file is probably truncated or otherwise
+/// partially written).
+inline bool
+image_data_readable(string_view filename, const ImageSpec* config, int subimage,
+                    int miplevel)
+{
```
lgritz, line 60: Is this still needed?

I understand the strategy of "read normally, and only if there's an error, try again using the missingcolor option."

But why not make the "try again" step go through ImageBuf like the first one, instead of needing all this code that drops down to the ImageInput level API?

Also, we should consider the merits of three approaches for iv:
1. Try, and if you fail, try again with missing pixels tolerated. (The current strategy you've implemented.)
2. An option -- perhaps an iv persistent preference, with a command-line override in either direction -- that specifies whether missing pixels should be tolerated and that affects even the "first try."
3. Always succeed. That is, just make iv always tolerate missing pixels and call it a day.

lgritz (commented):
I'm not sure I understand

lgritz:
There are still some CI failures that seem related to this PR. Maybe the newer compilers are giving warnings about things that the older compilers missed?

linsen458-spec:
This is not my native language, I'm using an LLM to restate and fix grammar in my replies and to double-check my work for oversights. Everything you see here has been viewed, reviewed, and approved by me personally before submitting.

Confirmed — gcc 15/16 with `-Woverloaded-virtual` flags the span-based `read_native_scanlines` overloads from the `ImageInput` base class: OpenEXRInput's pointer-based declarations hide them. For what it's worth, this failure predates this PR (the hiding exists on main as well) — our PR just happened to be the one that surfaced it in CI.

Fixed with a `using ImageInput::read_native_scanlines;` declaration in exr_pvt.h, which un-hides the base overloads (their default implementations call the pointer-based ones), pushed as a new commit.

lgritz:
There were several comments/questions I made when I reviewed it a couple days ago that I'd like to hear your opinions about. Please review the earlier discussion above.

linsen458-spec:
Thanks for the detailed review — sorry for the slow response, and thanks for pointing me back to the earlier discussion.

Going through the points:

1. On `ImageBuf::pixeladdr()`: that helper is gone in the current version of the patch, so this one should be moot now.

2. Understood on the `///` comments in imageviewer.h — makes sense that an internal header shouldn't carry doxygen-style docs. Happy to keep new additions consistent with whatever you prefer there.

3. Will do on the declaration: I'll move the internal read_native_scanlines variant out of the overrides into the private section near the end of the class, give it a distinct name, and add the comment you suggested. On the flag name, use_chunk_cache (defaulting to true) reads better to me too — bypass_chunk_cache=false is a double negative that's easy to misread.

4. About the probe and the overall strategy — let me lay out how I see it:

The fallback re-read does go through ImageBuf; the only part that drops to the ImageInput level is the readability probe. That probe exists because when the image is backed by the ImageCache, the normal read "succeeds" without touching pixels, and the failure would only surface awkwardly later at display time — so I used it to detect truncated files up front.

I did think through the three approaches you listed:

- Option 1 (current): the main weakness, as you're hinting, is all the machinery — probe, double read, config copying. And the probe only checks the last scanline/tile, so it can't strictly guarantee the whole image is readable anyway.
- Option 3 (always tolerate) is the simplest, but my worry is that iv is also used to inspect render output, and silently filling missing pixels with black could hide a problem — the user might not realize the file is incomplete.
- So my preference is option 2: an iv preference (tolerating missing pixels by default) applied from the first read, with a command-line override in either direction. That would let me drop the probe and the double read entirely and leave a single read path, while keeping the "partially readable" status-bar notice (the probe would survive, if at all, only to feed that notice).

It's your call though — if you feel option 3 fits iv better, or you have different ideas about the preference's default or shape, I'm happy to go either way. Once you decide I'll rework the patch.

Thanks again for walking me through this, I'm learning a lot from your reviews.

linsen458-spec:
Quick update on the review points, sorry for the scattered follow-ups:

Point 3 is actually already done — pushed as 00ec6db before my last reply. The internal scanline reader is moved out of the overrides into the private section at the end of the class, I kept the `use_chunk_cache` naming, and the comment you suggested is there too. I should have said "done" instead of "will do" in the reply, my bad.

On point 4 I'll hold off on the rework until you decide between the options. My preference is still option 2 (an iv preference with a command-line override, which lets me drop the probe and the double read), but I'm happy to go with option 3 if you feel that fits iv better. Once you pick, I'll rework the patch and update the PR description to match.

lgritz:
I also prefer option 2:

> So my preference is option 2: an iv preference (tolerating missing pixels by default) applied from the first read, with a command-line override in either direction. That would let me drop the probe and the double read entirely and leave a single read path, while keeping the "partially readable" status-bar notice (the probe would survive, if at all, only to feed that notice).

This is simple (least mechanism), covers both preferred use cases, and there's nothing stopping us from doing something more complex in the future if people turn out to desire it.

src/iv/imageviewer.h
lgritz: I definitely do not want you to change the comment style on any lines you are not already altering in this PR for its primary purpose.

But if you are adding NEW lines, or it is necessary to change existing lines, then if it's not too much trouble, I would like you to use the `//` regular comments, just so we aren't increasing the number of lines that we might need to come back and fix later.

linsen458-spec:
Reworked per option 2 and pushed.

iv now reads through a single path, applying its missing-pixel preference
from the first read (`oiio:missingcolor` set to black in the input config
by default). The probe is no longer part of the read path: it only runs
after a successful read, checks the last scanline/tile strictly, and exists
just to feed the "partially readable" status-bar note. The preference can
be overridden either way with `iv --missing-pixels {tolerate,strict}`,
where "strict" is the old behavior.

The openexr side is unchanged since your last review -- still just the
failed-chunk-decode recursion fix and the reader move.

I verified locally with a checker EXR truncated at 55%: strict fails at
pixel-fetch time, tolerant shows the readable portion with the truncated
tail black, and the status-bar note shows up for the partial file and stays
quiet for the full one. Also switched the comment blocks this PR adds to
plain `//` per your note.

lgritz (approved):
LGTM!

Thanks for working through several iterations to drive this to the finish line.

linsen458-spec:
Thanks for the reviews on these -- good working through them with you.

PR #5464, opened 2026-09-10
testsuite: test the oiiotool 'Channel reordering and padding' doc examples

linsen458-spec:
Part of #3992, reworked on top of the merged #5465.

This makes the oiiotool command lines from the "Channel reordering and padding" section of the docs live verbatim as tests, inside run_commands() blocks in testsuite/oiiotool-copy/run.py (next to the other tests that exercise --ch and related flags, as suggested in the review), each bracketed by BEGIN/END-docs-channels-* marker comments. src/doc/oiiotool.md then literalincludes each command between its markers — so the docs and the test are literally the same lines and can no longer drift apart.

The setup images intentionally use asymmetric per-channel colors: with the default black/white checker pattern, R and B are identical in every pixel, which made the channel-swap and alpha-from-red examples produce outputs indistinguishable by content hash.

Testing: the content hashes of all six output images are pinned in oiiotool-copy's ref/out.txt (text-only comparison, no reference images needed), and the test passes locally.

---

This is not my native language, I'm using an LLM to restate and fix grammar in my replies. All the code was reviewed by me personally, and the description above was reviewed and approved by me as well. Looking forward to contributing more here, and happy to hear any feedback.


linsen458-spec:
Not rambling at all — if anything, it's good to see how much you care about getting this right. This is exactly the kind of project that makes me happy to spend the time.

So I went back through both run.py files and diffed the examples against what oiiotool-copy already has. Two things stood out.

On where they live: the two directories are doing different jobs. oiiotool-copy groups by feature, and a lot of its lines lean on files created earlier in the same run.py — src/rgbaz.exr only exists because of the lines above it. oiiotool-channels follows the docs instead, with each example self-contained and the setup commands at the top making whatever it needs.

That's mostly why I'd leave the examples where they are. A line lifted from the middle of oiiotool-copy's run.py would leave a reader wondering where the input came from. Moving all the channel tests out is also riskier than it's worth: some of it is edge-case testing (the PNG channel-count warning, for one) that reads oddly as a doc example, and I'd rather not churn a passing test for no clear win. One small directory per section also fits the size and runtime limits you mentioned better than growing the big ones.

Then the duplicates. I went through them, and the overlap is smaller than it looks. Of the seven examples, only one repeats what oiiotool-copy already covers (`--ch R,G,B` against the `--ch R,G,B,A` test over in oiiotool-copy). The other six hit things copy never touches: zeroing channels by name, swapping two channels, `-i:ch` at open time, and the three ways of adding alpha/Z.

For that one I'd keep both, honestly. Different inputs, different outputs, and it costs about nothing. If we want a rule going forward, the one you floated earlier works for me: before a docs example becomes a test, check whether an existing test already covers it. Reads fine on its own? Let the docs reuse that line. Depends on earlier setup? The example stays with its section and the old test keeps its job.

I'll apply that to the rest of #3992 unless you object. And if you'd rather consolidate by feature and fold these into oiiotool-copy, that's a quick change too — say the word.

lgritz:
If the individual test lines we primarily use for full test coverage just aren't appropriate for what we want in the docs for clear elucidation, then let's keep them both and not worry about a little redundancy.

However, if you don't mind, I think I would prefer that these few new tests just land in oiiotool-copy, near the others that also exercise `--ch` and related commands, rather than making a separate test directory just for them. 

Sometimes when an individual test directory is seen to be very big or expensive, we do try to split it (into roughly equal portions if possible, and hopefully with some obvious logic to which tests go in which directory). In this case, the new tests are few, and they sure seem logically connected to things in oiiotool-copy, so let's not make a new directory for this.

Sorry to ask you to do the extra work. I believe this is the last thing to change on this PR.

By the way, did you intend this as a DevDays project?

linsen458-spec:
Thanks for explaining, that makes sense. Not a big deal, the move was quick — I folded the few tests into oiiotool-copy, right next to the other tests that exercise --ch, and removed the separate directory. The docs literalincludes just point to the new place now, the expected hashes didn't change. Good to also know when you do split a test directory, I didn't know that convention.

About DevDays — no, I didn't plan it as a DevDays project, I just picked this one from #3992 since I wanted to start with one example there. I'd be happy to join a future DevDays though, if it's ok to just pick another example from the list when the time comes.

lgritz:
> About DevDays — no, I didn't plan it as a DevDays project, I just picked this one from #3992 since I wanted to start with one example there. I'd be happy to join a future DevDays though, if it's ok to just pick another example from the list when the time comes.

You are always welcome to join. But also... you don't need to? It's mostly designed to encourage people to make their first PR, which you already have!

Now, you did this work at the same time as DD, and it is a task that quite often DD participants try, so... if you wanted to just retroactively declare it to be a DD project, I would be happy to include it. Totally up to you.

lgritz (approved):
LGTM, thanks for the several rounds to get this just right

PR #5465, opened 2026-09-11
testing: runtest.py refactor, simplification, new features, easy oiiotool docs

lgritz:
Refactor of runtest.py:

* Enhance the workhorse run_app() that assembles command lines: add a failureok parameter and handle that logic internally; automatically call oiio_app() on the command name if it's one of the OIIO command line apps, so the caller doesn't need to.

* Rewrite many of the other utilities like info_command, diff_command, maketx_command, rw_command, testtex_command, iconvert, oiiotool, to (a) use the new run_app() underneath and push a lot of the redundant logic into this core function; (b) use Python "f-strings" as a more compact notation for assembling the command strings. This combination greatly simpifies these functions into fairly thin wrappers around run_app().

* New run_commands that can take a string of multiple commands, split on newlines, and concatenate the run_app() result of each.  Blank lines or comments (lines whose first non-whitespace charcter is `#`) are safely ignored.

* Remove the "successmessage" parameter to iconvert -- we only used it in a couple places, and those can use "&& echo ..." instead.

* Compare test text output vs reference in a way that tolerates trailing whitespace changes (helped resolve some tricky Windows corner cases related to the "&& echo" idiom mentioned above.

So what does all this buy us, especially the run_commands()?  It can transform a sequence in a testsuite/blah/run.py like this,

```
command += oiiotool("-pattern constant:color=.25,.5,.75 64x64 3 -d half -o rgb64.exr")
command += oiiotool("-pattern constant:color=.25,.5,.75 64x64 3 -pattern constant:color=42 64x64 1 --chnames Z --siappend -d half -o rgb-z-parts64.exr")
```

into this:

```
command += run_commands("""
    oiiotool -pattern constant:color=.25,.5,.75 64x64 3 -d half -o rgb64.exr
    oiiotool -pattern constant:color=.25,.5,.75 64x64 3 -pattern constant:color=42 64x64 1 --chnames Z --siappend -d half -o rgb-z-parts64.exr
    """)
```

Nice, yeah? Within multi-line strings, we can just have shell commands exactly as how we would type them...

... and also exactly how we'd like them to appear in documentation!

So this allows us to do with command line examples in the oiiotool.md documentation chapter what we have been doing with C++ and Python code examples in the other chapters -- have the code actually execute as part of our testing (so we know they always work), and have the docs incorporate specific lines by reference, without duplication. For example, in a test:

```
command += run_commands("""
    oiiotool -pattern constant:color=.25,.5,.75 64x64 3 -d half -o rgb64.exr
    # BEGIN-oiiotool-example
    oiiotool -pattern constant:color=.25,.5,.75 64x64 3 -pattern constant:color=42 64x64 1 --chnames Z --siappend -d half -o rgb-z-parts64.exr
    # END-oiiotool-example
    """)
```

and then in the docs,

    ```{literalinclude} ../../testsuite/oiiotool/run.py
    :language: bash
    :start-after: BEGIN-oiiotool-example
    :end-before: END-oiiotool-example
    ```

and it will insert just that one line between the BEGIN/END comment pair.  The line right from the test, that actually executes in the testsuite, so it never is incorrect or stale.

### DIFF
--- a/testsuite/runtest.py
+++ b/testsuite/runtest.py
@@ -235,7 +235,7 @@ def run_app(app: str, silent: bool=False, failureok: bool=False,
 
 # Take shell `commands`, split at newlines, adorn each with redirects, etc.,
 # then re-join with semicolons to make a single command.
-def run_commands(commands : str, silent: bool=False,
+def run_commands(commands: str, silent: bool=False,
                  failureok: bool=failureok, concat: bool=True) -> str :
     result = ""
     for line in commands.splitlines():
--- a/testsuite/runtest.py
+++ b/testsuite/runtest.py
@@ -320,8 +320,8 @@ def rw_command (dir: str, filename: str, testwrite: bool=True, use_oiiotool: boo
 
 # Construct a command that will testtex
 def testtex_command (file: str, extraargs: str="", silent: bool=False, concat: bool=True) -> str:
-    return run_commands(f"testtex {file} {extraargs}",
-                        silent=silent, failureok=failureok, concat=concat)
+    return run_app(f"testtex {file} {extraargs}",
+                   silent=silent, concat=concat)
 
 
 # Construct a command that will run iconvert and append its output to out.txt
@@ -334,8 +334,8 @@ def iconvert (args: str, silent: bool=False, concat: bool=True,
 # Construct a command that will run oiiotool and append its output to out.txt
 def oiiotool (args: str, silent: bool=False, concat: bool=True,
              failureok: bool=False) -> str:
-    return run_commands(f"oiiotool {args}",
-                        silent=silent, failureok=failureok, concat=concat)
+    return run_app(f"oiiotool {args}",
+                   silent=silent, failureok=failureok, concat=concat)
 
 
 
--- a/testsuite/runtest.py
+++ b/testsuite/runtest.py
@@ -47,6 +47,7 @@ tmpdir = "."
 tmpdir = os.path.abspath (tmpdir)
 redirect = " >> out.txt "
 wrapper_cmd = ""
+oiio_app_list = ("oiiotool", "iinfo", "idiff", "maketx", "iconvert", "igrep", "testtex", "iv")
 
 def make_relpath (path: str, start: str=os.curdir) -> str:
     "Wrapper around os.path.relpath which always uses '/' as the separator."
@@ -170,6 +171,15 @@ else :
 
 # Handy functions...
 
+# Strip trailing spaces/tabs from a line, but leave its line ending (if any)
+# alone. Used to make text_diff tolerant of trailing whitespace, which can
+# vary by platform (e.g. cmd.exe's `echo` bakes in a trailing space that a
+# Unix shell would not) without being a meaningful difference in output.
+def _rstrip_line (line: str) -> str:
+    ending = line[len (line.rstrip ('\r\n')):]
+    return line.rstrip () + ending
+
+
 # Compare two text files. Returns 0 if they are equal otherwise returns
 # a non-zero value and writes the differences to "diff_file".
 # Based on the command-line interface to difflib example from the Python
@@ -179,8 +189,8 @@ def text_diff (fromfile: str, tofile: str, diff_file: str=None) -> int:
     try:
         fromdate = time.ctime (os.stat (fromfile).st_mtime)
         todate = time.ctime (os.stat (tofile).st_mtime)
-        fromlines = open (fromfile, 'r').readlines()
-        tolines   = open (tofile, 'r').readlines()
+        fromlines = [_rstrip_line(l) for l in open (fromfile, 'r').readlines()]
+        tolines   = [_rstrip_line(l) for l in open (tofile, 'r').readlines()]
         # if replace_relative:
         #     tolines = replace_relative(tolines)
     except:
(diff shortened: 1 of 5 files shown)


testsuite/runtest.py
```diff
@@ -291,62 +304,38 @@ def rw_command (dir: str, filename: str, testwrite: bool=True, use_oiiotool: boo
...
-               + " " + idiffextraargs + " " + output_filename + redirect + ";\n")
+        cmd += run_app(f"{tool} {preargs} {fn} {extraargs} -o {output_filename}")
+        cmd += run_app(f"idiff -a {fn} -fail {failthresh} -failpercent {failpercent}"
+                       + f" -hardfail {hardfail} -allowfailures {allowfailures}"
+                       + f" -warn {2*failthresh} {idiffextraargs} {output_filename}")
     return cmd
 
 
 # Construct a command that will testtex
 def testtex_command (file: str, extraargs: str="", silent: bool=False, concat: bool=True) -> str:
```
jinhgkim: failureok arg?

testsuite/runtest.py
```diff
@@ -291,62 +304,38 @@ def rw_command (dir: str, filename: str, testwrite: bool=True, use_oiiotool: boo
...
-    cmd = oiio_app("testtex") + " " + file + " " + extraargs + " "
-    if not silent :
-        cmd += redirect
-    if concat:
-        cmd += " ;\n"
-    return cmd
+    return run_commands(f"testtex {file} {extraargs}",
```
jinhgkim: nit: should we use run_app to follow the idiom? single command + testing an oiio app

testsuite/runtest.py
```diff
@@ -291,62 +304,38 @@ def rw_command (dir: str, filename: str, testwrite: bool=True, use_oiiotool: boo
...
+                        silent=silent, failureok=failureok, concat=concat)
```
jinhgkim: nit: I don't think we should pass failureok here since testtex_command doesn't take it as a param? failureok is default to 0, so it doesn't hurt to have it here, but if we wanna be super strict, I think we remove it.

testsuite/runtest.py
```diff
@@ -291,62 +304,38 @@ def rw_command (dir: str, filename: str, testwrite: bool=True, use_oiiotool: boo
...
              failureok: bool=False) -> str:
-    cmd = (oiio_app("oiiotool") + " " + args)
-    if not silent :
-        cmd += redirect
-    if failureok :
-        cmd += " || true "
-    if concat:
-        cmd += " ;\n"
-    return cmd
+    return run_commands(f"oiiotool {args}",
```
jinhgkim: nit: same for here - should we use run_app instead?

jinhgkim (commented):
Nice! I like this refactor.

testsuite/runtest.py
lgritz: Ah, yeah, ok. What happened is I wrote run_commands first, then ended up refactoring it so that all the new logic was pushed into run_app and run_commands just became "split by newlines and call run_app for each", more or less. I think in these spots, I didn't go back and change run_commands to the simpler run_app, which is ok here because it's just one line.

testsuite/runtest.py
```diff
@@ -206,13 +216,35 @@ def text_diff (fromfile: str, tofile: str, diff_file: str=None) -> int:
...
     if concat:
-        command += " ;\n"
-    return command
+        cmd += " ;\n"
+    return cmd
+
+
+# Take shell `commands`, split at newlines, adorn each with redirects, etc.,
+# then re-join with semicolons to make a single command.
+def run_commands(commands : str, silent: bool=False,
```
nrusch: Out of curiosity, is there any reason to prefer a multi-line string for multiple commands over a sequence of strings? The latter makes it possible to wrap very long commands for readability (if you wanted to), and still allows you to intersperse comments, except they would be actual Python comments.

(also a minor style nit:

```suggestion
def run_commands(commands: str, silent: bool=False,
```

lgritz: > Out of curiosity, is there any reason to prefer a multi-line string for multiple commands over a sequence of strings? The latter makes it possible to wrap very long commands for readability (if you wanted to), and still allows you to intersperse comments, except they would be actual Python comments.

@nrush, I'm not sure what you mean; can you clarify?

nrush: @nrusch

lgritz: so sorry

nrusch: By a "sequence of strings" I mean a sequence-type container like list or tuple.

So, for example, instead of:
```python
command += run_commands(
"""
# Here's a comment on the first command
iconvert short-exif-app1-len4.jpg out.null && echo short-exif-app1-len4-ok
# And here's a note about the second
iconvert short-exif-app1-len5.jpg out.null && echo short-exif-app1-len5-ok
""")
```

You would do something like:
```python
command += run_commands(
    [
        # Here's a comment on the first command
        "iconvert short-exif-app1-len4.jpg out.null && echo short-exif-app1-len4-ok",
        # And here's a note about the second
        "iconvert short-exif-app1-len5.jpg out.null && echo short-exif-app1-len5-ok",
    ]
)
```

lgritz: @nrusch 

The specific goal of this PR was to be able to have a sequence of consecutive lines in the run.py that would be *verbatim identical* to any of (a) what a user would type into a shell, (b) what would appear in a shell script to invoke that set of commands, (c) what should get included/copied by reference into the documentation showing examples of the commands.

I believe that this is a property of the first example you wrote above (verbatim lines within a `"""` block), but not the second, which defeats it by needing the extra quoting and commas.

testsuite/runtest.py
```diff
@@ -206,13 +216,41 @@ def text_diff (fromfile: str, tofile: str, diff_file: str=None) -> int:
...
-    command = app
+def run_app(app: str, silent: bool=False, failureok: bool=False,
+            concat: bool=True) -> str:
+    cmd = app.strip()
+    # If the command starts with the name of an OIIO app, substitute the
+    # full path to the built app.
+    words = cmd.split(maxsplit=1)
+    if not words:
+        return ""
+    if words[0] in oiio_app_list:
```
grdanny, line 227: is it ever ok to be not one of the known oiio_app_list commands?

lgritz: Yes! You can run any command you want. But in that case, we wouldn't want to prepend the relative path to the build we are testing, because something not built by OIIO won't be there.

grdanny (approved):
This looks cool to me as well! had one small questions there, but other than this looks very cool and might be nice to get in before dev days if someone wants to do the oiiotool-->docs stuff...

PR #5467, opened 2026-09-12
fix(dpx): don't add to metadata any userdata that failed to read

lgritz:
ReadUserData() returns false on a short read, having filled only part of the buffer. We previously ignored the return and added the whole thing as the "dpx:UserData" attribute, so the unread tail -- uninitialized bytes of a default_init_vector -- became part of the metadata. That is worse than a pixel leak: --info prints it, and oiiotool copies it into converted output without anyone reading a scanline, and that might contain stale heap data that could be sensitive.

The size guard only compared UserSize against the whole file size, but the block starts after the generic and industry headers, so a file can pass that check and still run off the end. Count the offset, and treat a failed read as an error instead of publishing the result.


lgritz:
No comments for a week, straightforward check of previously-ignored return error status value, plus a more correct check of data size plausibility. Merging.

PR #5469, opened 2026-09-12
docs: strip quote-box styling from indented supplementary content

lgritz:
This has bugged me for a long time, and now it's finally fixed!

Furo's default blockquote style adds a shaded background and left border. Markdown docs use ">" blockquotes purely to indent supplementary material (result-as-parameter variants, examples, usage synopses), not as actual quotations, so this made them look quoted. Custom stylesheet keeps the indentation but drops the quote styling.

### DIFF
--- a/src/doc/.gitignore
+++ b/src/doc/.gitignore
@@ -8,4 +8,6 @@
 /openimageio.ind
 _markdown_openimageio/
 _minted-openimageio/
-_static/
+_static/*
+!_static/css/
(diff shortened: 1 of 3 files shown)


lgritz:
Comments?

lgritz:
Thumbs up from Luna, I'm going to take that as an approval. It's docs only (and style only at that), can't break any code. So merging. It looks WAY BETTER!

PR #5477, opened 2026-09-14
fix(fmath): make float_to_rational work for all float inputs

lgritz:
A variety of float values would have UB or otherwise nonsensical behavior for float_to_rational. Tighten up the logic, and some of the call sites.

One reason this is important is that the float values can come from input files, and therefore may be corrupt or constructed to trigger these problems.


src/include/OpenImageIO/fmath.h
```diff
@@ -1306,24 +1306,42 @@ class EightBitConverter {
...
+/// undefined: zero, negative values, NaN, and magnitudes too small to
+/// represent all yield 0/1, and values too large (including infinity)
+/// yield UINT_MAX/1.  The denominator is never 0.
 inline OIIO_HOSTDEVICE void
 float_to_rational (float f, unsigned int &num, unsigned int &den)
 {
-    if (f <= 0) {   // Trivial case of zero, and handle all negative values
+    const unsigned int maxuint = std::numeric_limits<unsigned int>::max();
+    const float maxuintf       = float(maxuint);
+    if (!(f > 1.0f / maxuintf)) {
```
jinhgkim, line 1319: hm? why not just do`if (f <= 1.0f / maxuintf)`?

src/include/OpenImageIO/fmath.h
```diff
@@ -1306,24 +1306,42 @@ class EightBitConverter {
...
+        // Zero, all negative values, NaN, and anything too small to be
+        // expressed as a ratio of unsigned ints. Note the negated test, so
+        // that NaN lands here rather than falling through to a cast that
```
jinhgkim, line 1322: aha nevermind

jinhgkim (approved):
Looks good

src/include/OpenImageIO/fmath.h
lgritz: NaNs fail all comparisons!

PR #5478, opened 2026-09-14
feat(IBA): `IBA::decorr_stretch()` and `oiiotool --decorrstretch`

lgritz:
Weekend fun project: I read [this Gizmodo article](https://gizmodo.com/this-nasa-color-trick-was-meant-for-mars-now-its-unveiling-rock-art-on-earth-2000809844) and thought "that is a cool technique, I'll implement it." So here we are.

Decorrelation stretch exaggerates the subtle color variation of images whose channels are highly correlated, such as hazy aerial photographs, rock faces, or faded pigments. Diagonalize the channel covariance matrix, scale each principal axis to unit variance, rotate back, then give each channel the mean and standard deviation it started with.

The optional "percentile" contrast stretch instead maps a percentile of each channel and its complement onto [0,1], saturating the tails; that is the finishing step most published decorrelation stretches use, and it is what makes the result fill the display range.

The "mode" option chooses which matrix to diagonalize. Correlation mode standardizes each channel first, so all of them carry equal weight and the result is invariant to rescaling any one channel. Covariance mode weights each channel by its own variance. We default to covariance, unlike the MATLAB and JPL implementations, which default to correlation.

The eigenvectors come from a small cyclic Jacobi solver rather than Imath's 3x3 one, so any number of channels can be stretched jointly, not just RGB.

Example of what this does:

| Input | Output |
| ----- | ------ |
| hazy | decorr |


src/doc/oiiotool.md
```diff
@@ -3416,6 +3416,107 @@ current top image.
...
+    :width: 1.5 in
+    :alt: dcsimage4
+    ```
+* - original
+  - decorrstretch
+  - percentile=1
+  - mode=correlation
+::::
+
+> This command was added in OIIO 3.3.
```
jinhgkim: Perhaps 3.2? Isn't 3.3 for next year?

src/doc/pythonbindings.md
```diff
@@ -3008,6 +3008,37 @@ ImageBufAlgo.paste (Big, 100, 100, 0, 0, Small)
...
+    The output channels keep the mean and standard deviation they had in
+    `src`, unless `scale` (a multiplier on the standard deviation), `sigma`
+    (a standard deviation for every channel), or `mean` (a mean for every
+    channel) say otherwise. A nonzero `percentile` instead finishes with a
+    linear contrast stretch that maps that percentile of each channel and
+    its complement onto the [0,1] range. A `mode` of "correlation" gives
+    every channel equal weight, rather than weighting each by its own
+    variance.
+
+    This function was added in OpenImageIO 3.2.
```
jinhgkim, line 3032: ah yes, I guess 3.2 is your targeted release :)

src/python/py_imagebufalgo.cpp
```diff
@@ -2869,6 +2919,17 @@ declare_imagebufalgo(py_module& m)
                     "firstchannel"_a = 0, "roi"_a = ROI::All(),
                     "nthreads"_a = 0)
 
+        .def_static("decorr_stretch", &IBA_decorr_stretch, "dst"_a, "src"_a,
+                    "firstchannel"_a = 0, "nchannels"_a = 0, "scale"_a = 1.0f,
+                    "sigma"_a = 0.0f, "mean"_a = py::none(),
+                    "percentile"_a = 0.0f, "mode"_a = "", "roi"_a = ROI::All(),
+                    "nthreads"_a = 0)
+        .def_static("decorr_stretch", &IBA_decorr_stretch_ret, "src"_a,
```
jinhgkim: decorr_stretch_ret?

src/python/py_imagebufalgo.cpp
```diff
@@ -1218,6 +1218,56 @@ IBA_saturate_ret(const ImageBuf& src, float scale = 0.0f, int firstchannel = 0,
...
+{
+    ParamValueList params;
+    params["firstchannel"] = firstchannel;
+    params["nchannels"]    = nchannels;
+    params["scale"]        = scale;
+    params["sigma"]        = sigma;
+    params["percentile"]   = percentile;
+    if (mode.size())
+        params["mode"] = mode;
+    if (!mean.is(py::none()))
```
jinhgkim: nit: `.is_none()` reads a bit better?

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
```diff
@@ -1199,6 +1200,436 @@ ImageBufAlgo::saturate(const ImageBuf& src, float scale, int firstchannel,
...
+            std::vector<double> s2(size_t(nchannels) * nchannels, 0.0);
+            std::vector<float> v(nchannels);
+            for (ImageBuf::ConstIterator<S> p(src, subroi); !p.done(); ++p) {
+                bool finite = true;
+                for (int c = 0; c < nchannels; ++c) {
+                    v[c] = p[firstchannel + c];
+                    finite &= std::isfinite(v[c]);
+                }
+                if (!finite)
+                    continue;
```
jinhgkim, line 1231: Question: what happens when there are way too many invalid pixel values though? I guess decorr_stretch wouldn't represent the image very well? I am just curious to know what happens then and if it would be better to replace those invalid values to some other valid numbers. Or we could simply make it fail when the valid pixel count doesn't reach a certain threshold, say 80% of the total count?

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
```diff
@@ -1199,6 +1200,436 @@ ImageBufAlgo::saturate(const ImageBuf& src, float scale, int firstchannel,
...
+// Jacobi rotation. A is destroyed in the process. The eigenvalues are
+// returned in `eval`, and the corresponding eigenvectors are the columns of
+// `evec`. Only meant for the very small n of a channel covariance matrix. The
+// rotation formulas are the classical ones for cyclic Jacobi: Golub & Van
+// Loan, "Matrix Computations", "The Symmetric Eigenvalue Problem", section
+// "Jacobi Methods" (8.5 in the 4th edition).
+// (Fun fact: Charlie Van Loan was LG's undergraduate advisor!)
+static void
+jacobi_eigen_symmetric(span<double> A, int n, span<double> eval,
+                       span<double> evec)
```
jinhgkim, line 1263: I wonder if we want to put this helper into Imath eventually for arbitrary n x n matrix.

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
```diff
@@ -1199,6 +1200,436 @@ ImageBufAlgo::saturate(const ImageBuf& src, float scale, int firstchannel,
...
+            for (int c = roi.chbegin; c < firstchannel; ++c)
+                r[c] = a[c];
+            for (int i = 0; i < nchannels; ++i)
+                v[i] = a[firstchannel + i] - inmean[i];
+            for (int i = 0; i < nchannels; ++i) {
+                const float* m = &M[i * nchannels];
+                float sum      = outmean[i];
+                for (int j = 0; j < nchannels; ++j)
+                    sum += m[j] * v[j];
+                r[firstchannel + i] = sum;
```
jinhgkim, line 1333: Here, `sum` is a float. Is it possible that `OIIO_DISPATCH_COMMON_TYPES2(ok, "decorr_stretch", decorr_stretch_apply_, ...)` resolves dst img datatype to uint8 for formats like JPEG or PNG? If so, then we get clipped values in `r`. And later on in the decorr_stretch_percentile, I think we get incorrect results as we would be working with int instead of float.

jinhgkim (commented):
A few comments with a question... also learned a pretty cool "fun fact" along the way :)

src/doc/oiiotool.md
lgritz: quite right

src/python/py_imagebufalgo.cpp
lgritz: Good suggestion. Curiously, is_none is not used elsewhere in this file, but .is(py::none()) is used in a couple other places (that are unrelated to this PR). I must not have previously known about the existence of .is_none(). Will fix.

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
lgritz: I don't think it's worth worrying about.

Because the color transformation depends on the the contents of the image, it won't be stable across an animation or sequence of frames. Depending on the image contents, sometimes it's almost magically helpful in making near-invisible details really clear (as you can see from the example above), and other times, it doesn't help at all or makes the image look horrible. There are lots of cases where it's just not an appropriate tool to use, and an image riddled with NaN or Inf values is surely one of them.

I think this is an experimental visualization tool. It's not, like, appropriate to generate the view a surgeon would get in the middle of an operation.

Note that almost no IBA functions consider NaNs at all. Our general approach is that it's not worth the extra cost of checking every value, for something that should be caught far upstream, if it's even possible to occur at all. Aggregate stat gathering (as is occurring at this line) is a bit of a special case because a single NaN will pollute the whole image-wide stat, rather than just propagating the NaN to the single corresponding pixel in the output.

If somebody is concerned about lots of NaNs creeping into their process, they can use --fixnan upstream of the --decorrstretch to fix it.

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
lgritz: Since nobody has asked for it in Imath's history 25 year history, I'm not sure it's worth the trouble. Imath has some general math utilities, but almost exclusively geared to the (2, 3, 4) sizes of vectors and matrices that are its bread and butter, not trying at all to be a general linear algebra package.

I would say, though, that if we ever need it in a second place in OIIO, at that time we should move it from being local to this file to being in our fmath.h. But I don't want to put it in a public header until we need it in more than one place.

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
lgritz: Yeah, but if you call this with 8 bit in and 8 bit out, you get what you deserve.

People who want precision in their image math should be reading all images into float buffers internally, and quantizing again only for final output. That's what oiiotool does.

lgritz:
> A few comments with a question... also learned a pretty cool "fun fact" along the way :)

Haha, that was there 50% to see if anybody would ever read it. You are thorough!

I admit that my numerical methods knowledge has gotten rusty over the years, so I let Claude write that function for me, but I told it to give references, and when it cited a book by somebody I studied from, I was too amused to let it go without a comment (it *is*, or at least was for decades, legitimately the universal go-to matrix methods textbook). It was Van Loan's class that taught me FORTRAN and is the only time I ever really had to use it. (FORTRAN77 no less! I'm sure I wouldn't recognize today's modern FORTRAN at all. Though 77 was already getting civilized, you could use types with any variable names by that point. Earlier versions of the language implicitly typed the variable by its first letter! Variables starting with I-N were integers and all others were Real. The early days of programming languages were full of truly absurd ideas.)

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
jinhgkim: Aha okay, makes sense.

src/libOpenImageIO/imagebufalgo_pixelmath.cpp
jinhgkim: gotcha,

jinhgkim (approved):
Very cool feature!

PR #5480, opened 2026-09-17
feat(iv): image wipe compare

### LINKED ISSUE #4712: iv wipe mode
There should be a mouse+modifier that splits the screen horizontally and/or vertically (depending on the mouse movement direction) of the current image and the previous image. (originally #292)

nshelton:
## Description
closes #4712 

Adds an interactive wipe image comparison mode for `iv` The most recent (current image) is on the left(or top) and the last image is on the right (or bottom).  Both images have the same transform and render pixel aligned from top left. 

### Behavior
- new View menu (W) hotkey
- starts with a center left-right split, then follows mouse
    - split direction is determined from dominant movement direction (horizontal or vertical)
- (T) still togges last and current image in wipe mode, which swaps the sides. 
- status bar shows the two filenames being compared

### Implementation
the Renderer now has two slots, with texture objects in each one. This keeps the data for the last image availagle. 
The render shader has both textures available and just picks between them now. 

Texture allocation accounts for the pixel format, selected channel count, channel byte size of each image slot. 

This approach was chosen to make it easy to add new comparison modes in the future, like "diff" "absdiff" "colormap diff magnitude" "blend modes" 

### Non-goals
In order to get a manageable PR ready for dev days, many features were not addressed:
- No non-GLSL fallback
- No other comparison modes implemented, just hardcoded wipe behavior
- All state basically driven by current image. This includes
  -  rotation, orientation, OCIO filter, view transforms, exposure, brightness
- Pixel inspector and area selection only operate on current image
- Different channel layouts only partially supported; color processing uses the image A's channel layout. I found out it would be non-trivial to support color transforms for multiple images with arbitrary channel layouts in the shader, but maybe worth adding later. 
-  Images with negative data window origins may not align properly

### Tests

- [x] Enable and disable comparison using the menu and `W`.
- [x] Switch between vertical and horizontal wipes.
- [x] Move the wipe while panning and zooming.
- [x] Swap images and navigate forward/backward while comparison is active.
- [x] Close images while comparison is active.
- [x] Rotate and flip images while comparison is active.
- [x] Exercise pixel view, probe view, and area sampling.
- [x] Compare images with different resolutions.
- [x] Compare RGB, RGBA, and single-channel images.

I could not find existing tests for `iv` so i did not create any. This is a UX feature primarily anyway. 
I  built and tested this on macOS and Rocky9 linux.

There is also no public documentation for `iv`, though maybe it should be added.

### Checklist:

  behavior.
  on the GitHub runners.)

Assisted-by: GitHub Copilot / GPT-5.6 Sol


src/iv/imageviewer.cpp
```diff
@@ -1165,6 +1174,21 @@ ImageViewer::updateStatusBar()
...
+        std::string first_name  = Filesystem::filename(cur()->name());
+        std::string second_name = Filesystem::filename(
+            m_comparison_image->name());
+        if (first_name == second_name) {
+            first_name  = cur()->name();
+            second_name = m_comparison_image->name();
+        }
+        const char* first_side  = glwin->wipe_horizontal() ? "top" : "left";
+        const char* second_side = glwin->wipe_horizontal() ? "bottom" : "right";
+        message += Strutil::fmt::format("  |  {}: {}  {}: {}  (inspecting {})",
```
lgritz: Just `Strutil::format` is fine. The Strutil::fmt::format was to disambiguate during a transition long ago when the default format used a different convention.

src/iv/imageviewer.cpp
```diff
@@ -1165,6 +1174,21 @@ ImageViewer::updateStatusBar()
...
     message += cur()->shortinfo();
     statusImgInfo->setText(message.c_str());
+    if (m_comparison_image) {
+        std::string first_name  = Filesystem::filename(cur()->name());
+        std::string second_name = Filesystem::filename(
+            m_comparison_image->name());
+        if (first_name == second_name) {
+            first_name  = cur()->name();
+            second_name = m_comparison_image->name();
+        }
```
lgritz, line 1184: I was about to comment that a common idiom is probably going to be
```
iv foo/test.exr bar/test.exr
```
but you actually accounted for this!

src/iv/imageviewer.cpp
```diff
@@ -1307,6 +1327,15 @@ ImageViewer::loadCurrentImage(int subimage, int miplevel)
     return false;
 }
 
+bool
```
lgritz, line 1332: I know this is really nitpicky, but we generally prefer 3 spaces between freestanding functions.

lgritz (changes requested):
Looks great!

I built on my end and tried it out, seems perfect.

I noticed a couple of very minor things. And I believe the CI test for "clang-format" revealed some really minor issues.

Fix those and I think we're ready to merge this!

lgritz (approved):
LGTM!

This is a great feature, I will use it all the time!

lgritz:
I have to say, I rather like being in wipe mode and still having mouse clicks to zoom/unzoom or drag the image work as usual. So I think that when we eventually add some kind of "do a thing to freeze the split position", it should not preclude the other mouse actions that people are used to.

PR #5481, opened 2026-09-18
feat(oiiotool): enhance expressions with string formatting operator

### LINKED ISSUE #5208: feat(oiiotool): enhance expressions with string formatting operator
It would be handy for the command line expression syntax (things like `{2 * TOP.width}`) had a fully general way to do string formatting.

This may be kind of involved, but I think it's worth doing.

We'll use this issue to track progress and flesh out a spec for it.

As a motivating example, see #5207, where we posit the utility of something like

    --offset "{format('{:+}{:+}',X,Y}"

There's a bit of ugliness coming from the fact that we use `{}` in oiiotool for indicating expressions, but also `{}` inside the format specifications, and I'm not sure if it will end up needing special quoting. But the advantage of going for a fully general formatting utility that can be used in expression evaluation means that it not only solves your problem, but likely also unlocks a whole bunch of interesting additional usages.

nafisab:
feat(oiiotool): enhance expressions with string formatting operator
Fixes https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/5208

returns a string formatted according to `val`
(using standard std::format/Python replacement fields, such as `{}`,
`{:+}`, `{:04d}`, or `{:.2f}`) applied to the values that follow. For
example, `format('{:+}{:+}', 3, -4)` returns `"+3-4"`. Note: `val`
values containing literal-brace escapes (`{{` or `}}`) are not currently
supported.

Assisted-by: Claude Sonnet 5

Signed-off-by: Nafisa Baker Rahman <baker.nafisa@gmail.com>


src/doc/oiiotool.md
```diff
@@ -237,6 +237,12 @@ contents of an expression may be any of:
     OIIO 2.4.)
   - `not(val)` : returns 1 if `val` is a false value, 0 if `val` evaluates
     to a true value. (Added in OIIO 2.4.)
+  - `format(val,...)` : returns a string formatted according to `val`
+    (using standard `{fmt}`/Python-style replacement fields, such as `{}`,
```
lgritz: Let's change this to: ...using standard std::format/Python replacement fields...

src/oiiotool/expressions.cpp
```diff
@@ -148,6 +149,42 @@ Oiiotool::express_parse_atom(const string_view expr, string_view& s,
...
+            else if (Strutil::string_is<float>(a))
+                store.push_back(Strutil::from_string<double>(a));
+            else
+                store.push_back(a);
+        }
+        try {
+            result = ::fmt::vformat(val, store);
+        } catch (const ::fmt::format_error& e) {
+            express_error(expr, s,
+                          Strutil::fmt::format("format(): {}", e.what()));
```
lgritz: I think you can just say Strutil::format here. The Strutil::fmt::format is a relic from the transition period when we switched the meaning of format from printf-like to std::format-like. Now it's just the default.

lgritz (changes requested):
Just a couple of extremely tiny nits pointed out.

But otherwise, looks great, and it's going to be a helpful feature!

Get the CLA straightened out, fix those couple comments, and then I believe we are ready to merge this.

src/oiiotool/expressions.cpp
nafisab: Awesome! Updated and tested, things look good on my end. :D

src/doc/oiiotool.md
nafisab: Great! Just updated! I also updated this on my commits too!

lgritz:
I think we have a one-line formatting fix needed, but then are ready to go

nafisab:
Great! Just fixed based on the output!

lgritz (approved):
LGTM!

Thanks for this patch, I think it is a powerful addition to the expression evaluation syntax.

PR #5484, opened 2026-09-19
build: Change default OpenJPH auto-build to 0.32.0

lgritz:


lgritz:
No objections -> merging

PR #5485, opened 2026-09-19
fix(python): Avoid exceptions when decoding non-UTF-8 data from string attrs

### LINKED ISSUE #3856: [BUG] python3.10 spec.get_string_attribute() throws a runtime exception with non-utf-8 string headers
**Describe the bug**

When looping over all key/value pairs in an EXR header using the python bindings to OpenImageIO, a field containing 0xFF values causes an awkward `RuntimeError: Could not allocate string object!`. This is very commonly seen in DI when the upstream writing tool doesn't want to set a value and instead just fills it with some default of `0xFF`.

**To Reproduce**
With the following header (as shown with iinfo -a -v):
```sh
    AudioFramerate: 24
    AudioInfo: "������������������������������������������������������������������������@"
    AudioRunningLTC: "MOS"
```
Where the `�` character is `0xFF`.

Attempting to iterate over headers like so:

```python
>>> import OpenImageIO  as oiio
>>> src = oiio.ImageBuf('./my_file.1000.exr')
>>> spec = src.spec()
>>> for a in spec.extra_attribs:
...   print(f'{a.name} ({a.type}): {a.value} ({type(a.value)})')
```

Causes the following exception:

```python
AudioFileName (string):  (<class 'str'>)
AudioFramerate (float): 24.0 (<class 'float'>)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

The above exception was the direct cause of the following exception:

  File "<stdin>", line 2, in <module>
RuntimeError: Could not allocate string object!
```

**Expected behavior**
It appears that strings are assumed to be in utf-8 unicode, which causes an error trying to decode non-unicode data into a python string.

The workaround is to not access `a.value` and instead ask for the attribute directly like so:

```python
>>> try:
...   x = spec.get_string_attribute('AudioInfo')
... except RuntimeError:
...   pass
```

However, I was hoping to have a more graceful exception of this error. For example, I cannot access the raw bytes of this value to determine if the offending field is actually usable. Perhaps adding an `a.raw_value` could return the information in the field as a bytestring so the above try/except has an opportunity to do something else special with the value. Since the accessing of an attribute here is forcing the coercion to unicode, I cannot get to the original data using python in any way.

**Platform information:**
 - OIIO branch/version:  `openimageio-2.4.5.0-r0`
 - OS:  `Alpine Linux 3.17 under Docker`
 - C++ compiler:  

nrusch:
Add a new `py_str_escaped` helper function to handle conversion of string-like C++ types to `py:str` without raising a `UnicodeDecodeError` for non-UTF-8 input data.

This function is a fairly simple wrapper that calls the `PyUnicode_DecodeUTF8` with the `'surrogateescape'` error handling scheme (described at https://docs.python.org/3/library/codecs.html#error-handlers) and transfers ownership to the binding-specific object type.

Fixes #3856 

### Tests

Only manual tests so far, but I plan to add some test cases to the harness.

### Checklist:

  behavior.
  on the GitHub runners.)


src/python/py_oiio.h
```diff
@@ -469,6 +469,28 @@ py_to_stdvector(std::vector<T>& vals, const py::object& obj)
...
+    if (!py_str) {
+        py::raise_python_error();
+    }
+    return py::steal<py::str>(py_str);
+#else
+    if (!py_str) {
+        throw py::error_already_set();
+    }
+    return py::reinterpret_steal<py::str>(py_str);
+#endif
```
nrusch, line 491: I went with a single function definition containing implementation details for both backends, but it would be easy enough to just have two definitions in the `oiio_py` namespace with the Python API call duplicated if you prefer. Or we could split the difference and alias the exception type and handle stealing functions in `py_backend.h` in order to collapse this into one code path.

src/python/py_paramvalue.cpp
```diff
@@ -158,7 +158,7 @@ declare_paramvalue(py_module& m)
     py::class_<ParamValue>(m, "ParamValue")
         .OIIO_PY_PROP_RO("name",
                          [](const ParamValue& self) {
-                             return oiio_py::str(self.name().string());
+                             return py_str_escaped(self.name().string());
```
nrusch, line 161: Flagging that I've also wrapped the `ParamValue.name` accessor, because there's nothing that normalizes/sanitizes attribute names in the C++ API.

lgritz:
Those wheel failures are unrelated. I just merged a fix that should take care of them. If you rebase on the current main, they should go away.

src/python/py_oiio.h
```diff
@@ -469,6 +469,28 @@ py_to_stdvector(std::vector<T>& vals, const py::object& obj)
 }
 
 
+// Make a UTF-8 `py::str` from a string-like C++ type, using `'surrogateescape'`
+// error handling for any non-UTF-8 byte sequences. Compared to calling
+// `py::str` directly, this avoids raising a `UnicodeDecodeError` if the input
+// is not valid UTF-8.
```
lgritz, line 475: Let's add just a bit more to the comment to explain why and when specifically this should be used:

```suggestion
// is not valid UTF-8.
//
// It is important to do this for string return values that come from
// "untrusted" sources, such as being read from other files, so that we ensure
// that we are returning valid UTF-8 str's to the calling Python so that
// exceptions are not raised.
```

lgritz (approved):
LGTM!

I added a couple lines to a comment, but otherwise, all looks great and I am merging.

nrusch:
Oh OK cool. I can come back and address the spec serialization later, but this should cover most of the potential user-facing issues.

lgritz:
Oh, sorry, did I get ahead of you? If there are more changes you want to make before merging this PR, I can certainly wait.

nrusch:
Well, the outstanding tripping hazards here are just around `ImageSpec` serialization, in particular to XML: It's easy enough to route the `ImageSpec.serialize` and `.to_xml` wrappers through this same string decoding pathway, but even though you'll get back a valid Python string, the generated XML data will not be parseable.

I'd appreciate your thoughts on how to proceed here, since you have the context on how you'd like to dovetail this into the impending 3.2 cut (or whether that matters in practice for the Python bindings):
1. Prioritize expediency by wrapping the serialization methods, accepting that the XML will be "broken" and I will try to fix it in a follow-up.
2. Take some more time to find a solution for the XML serialization. So far, it looks like the best "simple" option for this might be converting invalid bytes to hex strings, but at least we would end up with valid XML (though the major caveat there is that string data would be different than the `surrogateescape` string decoding applied elsewhere). I'm also not quite sure how complicated this would be to implement yet.
4. Leave the XML serialization broken (I can't really think of a good reason to do this).

lgritz:
So the problem is that is we turn the xml serialization into valid UTF-8 for Python's sake, that XML won't correctly encode the *actual* string of char 255's? So there is a discrepancy between what the string contains in the file, what the python str can contain, and how the xml needs to be expressed to properly round-trip, right?

Can't we just express it as hex/oct strings, like "\xff\xff"? I think that is your suggestion 2. I'd be fine with that. If we re-consume the resulting XML, will we end up where we started, with a string full of 0xff's?

Totally your choice whether to do that as part of this PR, or whether to commit what we have and do that in isolation next.

nrusch:
OK. I'll push up the change to wrap those methods with this new helper as-is, and then look at a better solution for the XML data as a follow-up.

nrusch:
Done

lgritz (approved):
LGTM

Follow-up to address the xml issue is expected separately.

PR #5486, opened 2026-09-19
fix(build): internalize fmt/args.h to avoid mixing fmt versions

lgritz:
PR #5481 added `#include <fmt/args.h>` to expressions.cpp, the only raw `<fmt/...>` include in the tree. Everything else includes `<OpenImageIO/detail/fmt/...>`, the copies internalized into the build tree, and args.h was not among the internalized headers.

This args.h header also needs to do the same in order to avoid version mismatches.

Curiously, it was only on the "wheel" workflow, and on Mac, where it happened to fail in CI. That's why we didn't notice when #5481 was in review -- not intentionally touching python, it did not trigger the wheel workflow.


lgritz:
Pre-emptively merging without review because it fixes CI that is broken without it.

PR #5487, opened 2026-09-19
build: Raise default auto-build OpenEXR to 3.4.15 and Imath to 3.2.3

lgritz:


lgritz:
No objections -> merging

